PR #17, opened 2025-12-11
fix: load config from user-level ~/.config/opencode/oh-my-opencode.json

huynguyen03dev:
## Summary

Fixes config loading to check both user-level and project-level config paths, matching the documented behavior.

## Problem

Previously, `loadPluginConfig` only checked the project directory for `oh-my-opencode.json`. Users with config at `~/.config/opencode/oh-my-opencode.json` (as documented) had their settings ignored.

## Changes

- Add user-level config path: `~/.config/opencode/oh-my-opencode.json`
- Add project `.opencode` subdirectory path: `.opencode/oh-my-opencode.json`
- Implement config merging with project config taking precedence over user config
- Add logging for config loading and merging

## Config Precedence

1. User config: `~/.config/opencode/oh-my-opencode.json` (base)
2. Project config (overrides user):
   - `.opencode/oh-my-opencode.json`

## Testing

1. Place config in `~/.config/opencode/oh-my-opencode.json`
2. Restart OpenCode
3. Verify agents use configured models via `/tmp/oh-my-opencode.log`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,33 +41,77 @@ import { OhMyOpenCodeConfigSchema, type OhMyOpenCodeConfig } from "./config";
 import { log } from "./shared/logger";
 import * as fs from "fs";
 import * as path from "path";
+import * as os from "os";
+
+function loadConfigFromPath(configPath: string): OhMyOpenCodeConfig | null {
+  try {
+    if (fs.existsSync(configPath)) {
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Oh you just found and fixed another ai slops. Thanks a lot

PR #45, opened 2025-12-14
feat(agents): add smart orchestrator agent

huynguyen03dev:
- Add smartAgent with comprehensive system prompt for task orchestration
- Includes delegation to subagents (oracle, librarian, explore, etc.)
- Integrates LSP tools, AST-grep, and standard OpenCode tools
- Uses primary mode to inherit user's configured model


huynguyen03dev:
> If it doesn't mind you can I close this one and mention you as co author? Cuz my approach was a big different (but like the prompt btw I'll keep it in a style)

yes, you can change it to be better suited for this plugin.

PR #46, opened 2025-12-14
feat: add OmO agent to config schema for model override support

huynguyen03dev:
Can you consider removing the default model: "anthropic/claude-opus-4-5" of omo? 

Since this is now a primary agent that replace build, users may use it with different models, I got the error "Agent OmO's configured model anthropic/claude-opus-4-5 is not valid" when changing agent to OmO everytime if haven't overide the model yet.


code-yeongyu:
I am late for the dinner making it so I forgot about it thanks

PR #48, opened 2025-12-14
docs: update README to include jsonc as installation config as well

paulp-o:
**Summary**
- Update README installation instructions to support both opencode.json and opencode.jsonc config files
- JSONC (JSON with Comments) allows users to add comments in their OpenCode configuration

**Changes**
- Added note explaining JSONC support in both English and Korean READMEs
- Updated bash scripts to detect existing .jsonc files before .json
- Updated verification steps to check whichever config file exists

### DIFF
--- a/README.ko.md
+++ b/README.ko.md
@@ -149,14 +149,21 @@ OpenCode가 설치되어 있지 않다면, [OpenCode 설치 가이드](https://o
 ```bash
 mkdir -p ~/.config/opencode
 
-if [ -f ~/.config/opencode/opencode.json ]; then
+CONFIG_FILE=""
+if [ -f ~/.config/opencode/opencode.jsonc ]; then
+    CONFIG_FILE=~/.config/opencode/opencode.jsonc
+elif [ -f ~/.config/opencode/opencode.json ]; then
+    CONFIG_FILE=~/.config/opencode/opencode.json
+fi
+
+if [ -n "$CONFIG_FILE" ]; then
     # 기존 설정 파일이 있는 경우: jq 또는 python으로 plugin 추가
     if command -v jq &> /dev/null; then
         jq '.plugin = ((.plugin // []) + ["oh-my-opencode"] | unique)' \
-            ~/.config/opencode/opencode.json > /tmp/oc.json && \
-            mv /tmp/oc.json ~/.config/opencode/opencode.json
+            "$CONFIG_FILE" > /tmp/oc.json && \
+            mv /tmp/oc.json "$CONFIG_FILE"
     else
-        echo "~/.config/opencode/opencode.json의 plugin 배열에 \"oh-my-opencode\"를 추가하세요"
+        echo "$CONFIG_FILE 의 plugin 배열에 \"oh-my-opencode\"를 추가하세요"
     fi
 else
     # 새 설정 파일 생성
@@ -168,7 +175,7 @@ fi
 
 ```bash
 opencode --version  # 1.0.132 이상이어야 함
-cat ~/.config/opencode/opencode.json  # plugin 배열에 "oh-my-opencode"가 있어야 함
+cat ~/.config/opencode/opencode.json 2>/dev/null || cat ~/.config/opencode/opencode.jsonc  # plugin 배열에 "oh-my-opencode"가 있어야 함
 ```
 
 ### 4단계: 인증정보 설정
--- a/README.md
+++ b/README.md
@@ -150,14 +150,21 @@ If OpenCode isn't installed, check the [OpenCode Installation Guide](https://ope
 ```bash
 mkdir -p ~/.config/opencode
 
-if [ -f ~/.config/opencode/opencode.json ]; then
+CONFIG_FILE=""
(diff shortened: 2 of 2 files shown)


paulp-o:
human note: the setup process done by agents would break when my config is in jsonc

code-yeongyu:
Maybe I just updated README so it is conflicted can you update your branch?

code-yeongyu:
Thank you. If you have any workflows that work especially well, please share them or suggest improvements. Let's build the best agent harness.

paulp-o:
oops, jq doesn't seem to support jsonc. I only tested if my agent would manage to do it with the new README, it might have done some workaround as jq would fail

paulp-o:
@code-yeongyu you mean guide to human or the agent?

I don't really have a very minimal solution for this in mind yet without making some sort of setup tool - maybe go for pattern matching using sed or guiding the llm do the edit itself (I think this is reliable enough)?

paulp-o:
> > @code-yeongyu Sure.
> > btw, I honestly want to see a cli tool for setting omo up, which could deal with all these and also enable only (modularised) part of the features. (similar to taskmaster or openspec) I plan to do this myself on my own fork of omo soon, but I would like to make PR for that as well if you see the same value as me.
> 
> That's huge! I was thinking of one but other fancy features got my priority. (Current lack of installation guide is because of this)
> 
> Super appreciated if you contribute in a way.
> 
> One thing on my mind is that by _not_ building a setup CLI, I'm effectively restricting configuration to a single method—but this stems from a desire to ensure the setup provides strong defaults. Even if we do eventually create a setup CLI, I believe it should only offer an option to launch immediately with recommended settings or, at most, choices within broad categories. Otherwise, running the installation CLI would feel like falling into a black hole.

I'll start by respecting your default-first approach. I wholeheartedly wish background agent being separate though because it's too good by itself

code-yeongyu:
> > > @code-yeongyu Sure.
> > > btw, I honestly want to see a cli tool for setting omo up, which could deal with all these and also enable only (modularised) part of the features. (similar to taskmaster or openspec) I plan to do this myself on my own fork of omo soon, but I would like to make PR for that as well if you see the same value as me.
> > 
> > 
> > That's huge! I was thinking of one but other fancy features got my priority. (Current lack of installation guide is because of this)
> > Super appreciated if you contribute in a way.
> > One thing on my mind is that by _not_ building a setup CLI, I'm effectively restricting configuration to a single method—but this stems from a desire to ensure the setup provides strong defaults. Even if we do eventually create a setup CLI, I believe it should only offer an option to launch immediately with recommended settings or, at most, choices within broad categories. Otherwise, running the installation CLI would feel like falling into a black hole.
> 
> I'll start by respecting your default-first approach. I wholeheartedly wish background agent being separate though because it's too good by itself

That's good to hear. Just a question here: is this 'setup' thing doesn't valuable to you? I thought those curated agents and lsp / ast tools are like basic but missing stuff for other agents so wanna know your opinion

code-yeongyu:
> @code-yeongyu a CLI matters to me. But it's mostly personal.
> 
> I often use (or have to use) different workflows per project, or I’m just experimenting with various setups. In those cases, having to ask an AI to set things up every time feels slow, if not only nondeterministic. (Of course I just comment/uncomment `oh-my-opencode` in the config json"c")
> 
> that said, having a simple, deterministic on/off switch might actually make it easier for more people _weird enough like me_ to try the flow in the first place. I as well doubted about letting an AI do a full install by itself, so I ended up reading the entire 'for agents' section myself anyway. lol.
> 
> Also, I suspect that many _enthusiastic_ people who end up here already have some kind of personal workflow (myself included). Some of them also care a lot about on-demand tool definition loading (similar to Claude skills), because they're allergic to context bloat. Which is why I tend to avoid MCP as well.
> 
> In many workflows, I only like subparts of a it, and I also see people often end up building **frankenagents**/frankenworkflow by mixing pieces they like. (like, consider how many people still use things like @[VoltAgent/awesome-claude-code-subagents](https://github.com/VoltAgent/awesome-claude-code-subagents).) To those people, too all-in-one, full-package solutions can feel a bit heavy to try and adopt (taskmaster has this issue in my opinion).
> 
> That's why being able to toggle 'modules' on and off via a CLI is needed personally for me (of course, generally, a simple install/uninstall would be so much handy too!). Specifically, I have my own plan subagent, websearch tool (tuned for my taste), etc. so no need for duplicate tools.
> 
> In the end, all of this might honestly not really relevant to you- since you’re aiming for strong defaults anyway. I'm just one of the many open source enthusiasts just enjoy putting together pieces from different projects to fit my own needs, and building agentic coding workflows isn’t very different.
> 
> tl;dr - it is, but for me. or anyone who wants things to be modular and put things together by themselves.

I will seriously consider making this to be separated modules and let this project to gather them all. Suggesting highly opinionated default setup while making them all configurable. It might be tricky to configure to do so but to make this thing as Omarchy of OpenCode it would fit much better than. Thank you for your good idea.

code-yeongyu:
BTW I'll close this. Thank you for your Contribution.

PR #83, opened 2025-12-17
Remove self dependency from package.json

devxoul:
## Problem

`$ bun run build` does not work as intended. It does not generate `dist/index.js` but only `d.ts` files. Only the parts after `&&` would run properly.

## Solution

Remove self dependency and bun from the dependency list to make bun find the executable from the proper location.

**Before**
```sh
$ bun why bun
bun@1.3.3
  ├─ peer oh-my-opencode (requires >=1.0.0)
  └─ peer oh-my-opencode@0.1.30 (requires >=1.0.0)
     └─ dev oh-my-opencode (requires ^0.1.30)

$ bun build

$ ls dist/index.js
ls: dist/index.js: No such file or directory
```

**After**
```sh
$ bun why bun
error: No packages matching 'bun' found in lockfile

$ bun build

$ ls dist/index.js
dist/index.js
```

## Debugging

You may find the reason with the script below:

```diff
  "scripts": {
+   "whichbun": "which bun",
  }
```

```shell
$ bun whichbun
/path/to/oh-my-opencode/node_modules/.bin/bun

$ readlink /path/to/oh-my-opencode/node_modules/.bin/bun
../bun/bin/bun.exe  # Cannot execute in macOS/Linux
```

As bun.exe cannot be executed from non-Windows machines, only the later part of `&&` would work:

```
"build": "bun build ... && tsc --emitDeclarationOnly && bun run build:schema",
          ^^^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          ❌ bun.exe        ✅ Generate d.ts files
```

## Note

You should delete the `node_modules/bun` directory and re-install the dependencies.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -18,11 +18,7 @@
       "devDependencies": {
         "@types/picomatch": "^3.0.2",
         "bun-types": "latest",
-        "oh-my-opencode": "^0.1.30",
         "typescript": "^5.7.3",
-      },
-      "peerDependencies": {
-        "bun": ">=1.0.0",
       },
     },
   },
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Wow famous person

PR #189, opened 2025-12-24
fix: Notification hook works weirdly for subagent sessions

### LINKED ISSUE #92: Notification hook works weirdly
It seems that the notification hook triggers when the session is waiting for your input is also triggered by on the subagent session. Thought I have fixed it but still experiencing, so still needs to fix

code-yeongyu:
## Summary

- Added mainSessionID check to prevent notifications in subagent sessions
- Only trigger notifications for main session when waiting for user input
- Added comprehensive tests with proper assertions and BDD comments

Closes #92

---

**Improvements over original PR #183:**
- Tests now actually verify notification was/wasn't sent (mocking `ctx.$`)
- BDD comments (`#given`, `#when`, `#then`) per project convention
- `afterEach` cleanup for proper test isolation
- Additional test cases (activity cancellation, duplicate prevention)


code-yeongyu:
@sisyphus-dev-ai hey please write down here the review comment as github

sisyphus-dev-ai (commented):
## Review

Nice fix — the core change to gate  notifications by  +  matches how other hooks already treat main-vs-subagent sessions (e.g. ). This should stop OS-level notifications from firing for background agent sessions while keeping the main session behavior intact.

### What looks good
- Early-return ordering is sensible:  first, then  mismatch, then existing de-dupe guards (, , ).
- Tests actually assert side effects by capturing  notification invocations (vs. just “no throw”).
- Good coverage for cancellation paths (, ) and duplicate prevention.

### Nits / suggestions
-  uses real timers + small sleeps; should be OK, but if this ever flakes in CI we may want to bump the buffers slightly or move to a deterministic timer strategy.
- In ,  ignores  — fine for these tests, but if future cases need per-session todo behavior we may want a tiny helper to return different data by id.

No blocking issues from my side.

code-yeongyu:
@sisyphus-dev-ai  hey please merge then

sisyphus-dev-ai:
## Next steps\n- Checking PR status (reviews, checks, mergeability)\n- Running local `bun install` + `bun run typecheck`\n\n## Agents I'm leveraging\n- Exploration agent: scan PR diff + repo structure for risk\n- Librarian agent: confirm repo conventions/scripts for bun + typecheck\n- Oracle agent: sanity-check merge strategy/safety\n- Planning agent: keep work breakdown tight\n

PR #217, opened 2025-12-25
fix: prevent premature exit in non-interactive mode when tasks pending

### LINKED ISSUE #216: when sisyphus agent is working, and background tasks or remaining todos left, it must not be 'stopped'
https://github.com/code-yeongyu/oh-my-opencode/actions/runs/20499818603/job/58905421543 like this i saw, it stopped so not making results that i have wanted. seems i have to work it differently on 'opencode run' command. investigation required

code-yeongyu:
## Summary

- Prevent `opencode run` from exiting prematurely when background tasks or incomplete todos remain
- Detect non-interactive mode (CI, `opencode run`, no TTY) and inject continuation prompt BEFORE `session.idle` fires

## Problem

When running `opencode run` in CI (GitHub Actions), the agent would exit immediately after saying "waiting for background tasks..." because:
1. Agent finishes response → `session.idle` event fires
2. `opencode run` breaks its event loop on `session.idle` → process exits
3. Existing `todo-continuation-enforcer` uses 2-second countdown → too slow, process already dead

## Solution

Hook into `message.updated` (fires BEFORE `session.idle`) instead of `session.idle`:
1. Detect terminal finish reason (not `tool-calls`, not `unknown`)
2. Check for running background tasks via `BackgroundManager`
3. Check for incomplete todos
4. If either exists in non-interactive mode → inject prompt immediately
5. This keeps the session active → `session.idle` never fires → `opencode run` continues

## Changes

| File | Change |
|------|--------|
| `src/hooks/non-interactive-env/detector.ts` | NEW: Detect CI/non-interactive environment |
| `src/hooks/non-interactive-env/index.ts` | Export detector |
| `src/hooks/todo-continuation-enforcer.ts` | Add preemptive injection on `message.updated` |
| `src/index.ts` | Pass `BackgroundManager` to enforcer |

## Testing

- [x] `bun run typecheck` - Pass
- [x] `bun test` - 31 pass, 0 fail

Closes #216

---
🤖 GENERATED WITH ASSISTANCE OF [OhMyOpenCode](https://github.com/code-yeongyu/oh-my-opencode)

### DIFF
--- a/src/hooks/non-interactive-env/detector.ts
+++ b/src/hooks/non-interactive-env/detector.ts
@@ -1,0 +1,19 @@
+export function isNonInteractive(): boolean {
+  if (process.env.CI === "true" || process.env.CI === "1") {
+    return true
+  }
+
+  if (process.env.OPENCODE_RUN === "true" || process.env.OPENCODE_NON_INTERACTIVE === "true") {
+    return true
+  }
+
+  if (process.env.GITHUB_ACTIONS === "true") {
+    return true
+  }
+
+  if (!process.stdout.isTTY) {
+    return true
+  }
+
+  return false
+}
--- a/src/hooks/non-interactive-env/index.ts
+++ b/src/hooks/non-interactive-env/index.ts
@@ -3,6 +3,7 @@ import { HOOK_NAME, NON_INTERACTIVE_ENV, SHELL_COMMAND_PATTERNS } from "./consta
 import { log } from "../../shared"
 
 export * from "./constants"
+export * from "./detector"
 export * from "./types"
 
 const BANNED_COMMAND_PATTERNS = SHELL_COMMAND_PATTERNS.banned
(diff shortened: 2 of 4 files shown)


code-yeongyu:
## Review Response

Thanks @sisyphus-dev-ai for the thorough review!

### Answers to Questions

**1. Injected Message Role**

✅ Confirmed working correctly. When we inject via `ctx.client.session.prompt()`, OpenCode creates a **user message**. On the next `message.updated` event with `role === "user"`, we clear `preemptivelyInjectedSessions` (line 294-295). This allows re-injection on subsequent finishes.

**2. Type Change Verification**

✅ Confirmed via librarian search in sst/opencode. The `finish` field is defined as:
```typescript
// packages/opencode/src/session/message-v2.ts
finish: z.string().optional()
```
Values include: `"tool-calls"`, `"stop"`, `"length"`, `"error"`, etc. The `boolean` type in `preemptive-compaction/index.ts` is incorrect and should be fixed separately.

**3. TTY Detection**

✅ Fixed in commit `500a8bc`. Changed to explicit check:
```typescript
if (process.stdout.isTTY !== true) {
  return true
}
```

**4. Tests**

Valid suggestion. Will add integration tests in a follow-up PR to keep this one focused.

---

Ready for merge! 🚀

PR #235, opened 2025-12-25
fix(todo-continuation-enforcer): skip continuation check after interactive_bash to prevent false triggers

### LINKED ISSUE #234: Todo continuation hook called when there is no todo left, when shell command executed on the oc chat thing
got one bug report: "seems like when i entered shell mode, it feels like the todo continuation hook is called"

sisyphus-dev-ai:
## Summary

Fixes the issue where the todo continuation hook triggers immediately after using interactive_bash (shell mode), even when there are no incomplete todos.

## Changes

- Added 5-second debounce mechanism after `interactive_bash` tool execution
- Tracks interactive_bash usage via `toolExecuteAfter` hook
- Skips todo continuation checks during the debounce period
- Properly cleans up timestamps on session deletion

## Technical Details

The hook now listens to `tool.execute.after` events and tracks when `interactive_bash` is executed. When a `session.idle` event is received within 5 seconds of an interactive_bash execution, the todo continuation check is skipped entirely.

This prevents false triggers while still allowing the hook to function normally in other scenarios.

## Testing

- ✅ All existing tests pass
- ✅ TypeScript compilation successful  
- ✅ Build successful

Closes #234

### DIFF
--- a/src/hooks/todo-continuation-enforcer.ts
+++ b/src/hooks/todo-continuation-enforcer.ts
@@ -18,6 +18,7 @@ export interface TodoContinuationEnforcerOptions {
 
 export interface TodoContinuationEnforcer {
   handler: (input: { event: { type: string; properties?: unknown } }) => Promise<void>
+  toolExecuteAfter?: (input: { tool: string; sessionID: string; callID: string; args?: Record<string, unknown> }) => Promise<void>
   markRecovering: (sessionID: string) => void
   markRecoveryComplete: (sessionID: string) => void
 }
@@ -70,6 +71,7 @@ function detectInterrupt(error: unknown): boolean {
 
 const COUNTDOWN_SECONDS = 2
 const TOAST_DURATION_MS = 900 // Slightly less than 1s so toasts don't overlap
+const INTERACTIVE_BASH_DEBOUNCE_MS = 5000 // Skip todo checks for 5s after interactive_bash
 
 interface CountdownState {
   secondsRemaining: number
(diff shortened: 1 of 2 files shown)


sisyphus-dev-ai:

sisyphus-dev-ai:
You're absolutely right - this fix is targeting the wrong thing. Here's what I found:

## The Real Issue

**"Shell mode"** is OpenCode's TUI mode where users type shell commands directly, NOT the `interactive_bash` tool. These are completely different:

| Concept | What it is | How it relates |
|---------|-----------|----------------|
| `interactive_bash` tool | Agent uses tmux | Agent tool execution |
| "Shell mode" (TUI) | User enters bash prompt | TUI mode transition |

When you enter shell mode, OpenCode emits a `session.idle` event (mode transition). The debounce on `interactive_bash` tool doesn't affect this at all.

## Root Cause

**The `executeAfterCountdown()` function doesn't re-verify todos before injecting the prompt.**

Race condition:
1. `session.idle` fires (user enters shell mode)
2. Todo check: finds incomplete todos → starts 2s countdown
3. During countdown: todos get completed (by user action, API latency, etc.)
4. Countdown ends → `executeAfterCountdown()` runs
5. **NO fresh todo check** → injects continuation prompt anyway with stale data

Looking at lines 201-251, the countdown callback:
- ✅ Re-checks `recoveringSessions`
- ✅ Re-checks `interruptedSessions` / `errorSessions`  
- ❌ Does NOT re-check if todos are still incomplete

## Correct Fix

Add todo re-verification in `executeAfterCountdown()` before injecting the prompt:

\`\`\`typescript
// Re-verify todos are still incomplete after countdown
const freshTodos = await ctx.client.session.todo({ path: { id: sessionID } })
const freshIncomplete = freshTodos.filter(t => 
  t.status !== "completed" && t.status !== "cancelled"
)
if (freshIncomplete.length === 0) {
  log(\`[\${HOOK_NAME}] Abort: no incomplete todos after countdown\`, { sessionID })
  return
}
\`\`\`

And use fresh data in the prompt message instead of stale `todos`/`incomplete` variables.

Will close this PR and submit the correct fix.

PR #239, opened 2025-12-25
fix(todo-continuation-enforcer): re-verify todos after countdown to prevent stale data injection

### LINKED ISSUE #234: Todo continuation hook called when there is no todo left, when shell command executed on the oc chat thing

sisyphus-dev-ai:
## Summary

Fixes the race condition in the todo continuation hook where it would inject a continuation prompt even when all todos had been completed during the countdown period.

## Root Cause

The `executeAfterCountdown()` function used stale todo data from the initial `session.idle` check without re-verifying that incomplete todos still existed after the countdown finished.

**Race condition flow:**
1. `session.idle` fires (e.g., user enters shell mode in OpenCode TUI)
2. Initial check finds incomplete todos → starts 2-second countdown
3. During countdown: todos get completed (by user action, background process, or API latency)
4. Countdown ends → `executeAfterCountdown()` runs
5. **No fresh todo check** → injects continuation prompt anyway with stale data

## Changes

- ✅ Add fresh todo verification in `executeAfterCountdown()` before prompt injection
- ✅ Use fresh todo data in the continuation prompt message
- ✅ Abort injection if no incomplete todos remain after countdown
- ✅ Add detailed logging for the re-verification process

## What Was Wrong with PR #235

PR #235 attempted to fix this by adding a 5-second debounce after `interactive_bash` tool execution. This was incorrect because:

1. **Wrong trigger**: `interactive_bash` is an **agent tool** for running tmux commands, while "shell mode" is a **TUI mode** where users type shell commands directly
2. **Wrong approach**: The debounce addressed mode transitions, not the stale data race condition
3. **Incomplete fix**: Even if it targeted the right trigger, it didn't solve the fundamental issue of using stale todo data

The correct fix addresses the actual root cause: stale data being used after the countdown.

## Testing

- ✅ TypeScript compilation successful
- ✅ Build successful
- ✅ All existing tests pass

Closes #234
Supersedes #235

### DIFF
--- a/src/hooks/todo-continuation-enforcer.ts
+++ b/src/hooks/todo-continuation-enforcer.ts
@@ -215,6 +215,30 @@ export function createTodoContinuationEnforcer(
           return
         }
 
+        let freshTodos: Todo[] = []
+        try {
+          log(`[${HOOK_NAME}] Re-verifying todos after countdown`, { sessionID })
+          const response = await ctx.client.session.todo({
+            path: { id: sessionID },
+          })
+          freshTodos = (response.data ?? response) as Todo[]
+          log(`[${HOOK_NAME}] Fresh todo count`, { sessionID, todosCount: freshTodos?.length ?? 0 })
+        } catch (err) {
+          log(`[${HOOK_NAME}] Failed to re-verify todos`, { sessionID, error: String(err) })
+          return
+        }
+
+        const freshIncomplete = freshTodos.filter(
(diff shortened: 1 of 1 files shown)


code-yeongyu:
@sisyphus-dev-ai hello please answer to my question.

sisyphus-dev-ai:

PR #253, opened 2025-12-25
fix(antigravity): preserve thinking blocks in message converter

codewithkenzo:
## Summary

Fix thinking block handling in Antigravity message converter.

### Problem

When using Claude models with extended thinking through Antigravity, thinking blocks (`type: "thinking"` or `type: "redacted_thinking"`) were being dropped during OpenAI→Gemini message conversion.

### Fix

- Added `thought?: boolean` to GeminiPart interface
- Added handling for `thinking` and `redacted_thinking` content types in `convertContentToParts()`
- Converts to Gemini's `thought: true` format with text content

### Note

There's a separate upstream issue where OpenCode doesn't preserve thinking block signatures in message history, causing multi-turn conversations to fail. That's tracked in sst/opencode#6176 and needs to be fixed in OpenCode itself.

### DIFF
--- a/src/auth/antigravity/message-converter.ts
+++ b/src/auth/antigravity/message-converter.ts
@@ -164,25 +164,10 @@ function convertContentToParts(content: string | OpenAIContentPart[] | undefined
     if (part.type === "text" && part.text) {
       parts.push({ text: part.text })
     } else if (part.type === "thinking" || part.type === "redacted_thinking") {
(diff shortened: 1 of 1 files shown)


codewithkenzo:

codewithkenzo:
Closing in favor of a more comprehensive fix that addresses the complete signature flow issue.

PR #262, opened 2025-12-26
feat: add get-local-version CLI command for version checking

### LINKED ISSUE #260: version checker
> This is not very intuitive. Is there a command line or another way to see it? 

 _Originally posted by @cemo in [#236](https://github.com/code-yeongyu/oh-my-opencode/issues/236#issuecomment-3692748582)_

sisyphus-dev-ai:
## Summary

Adds a new CLI command `bunx oh-my-opencode get-local-version` that displays the currently installed version and checks if it's up to date with the latest npm release.

## Changes

- **New CLI command**: `get-local-version` with support for:
  - Human-readable colored terminal output
  - JSON output format (`--json` flag) for scripting
  - Directory-specific checking (`--directory` flag)
  
- **Edge case handling**:
  - Local development mode (file:// protocol)
  - Pinned versions (oh-my-opencode@x.x.x)
  - Network errors / offline mode
  - Missing package.json

- **Implementation details**:
  - Reuses existing version checking infrastructure from `auto-update-checker`
  - Uses `picocolors` for colored output (already a dependency)
  - Follows project conventions (Bun, TypeScript, barrel exports)
  - Type-safe with comprehensive TypeScript interfaces

## Testing

✅ TypeScript type checking passed  
✅ Build successful  
✅ Manual testing:
  - Help output works correctly
  - Command executes successfully  
  - JSON output format works
  - Detects local dev mode correctly

## Usage

```bash
# Check version with colored output
bunx oh-my-opencode get-local-version

# Get JSON output for scripting
bunx oh-my-opencode get-local-version --json

# Check from specific directory
bunx oh-my-opencode get-local-version --directory /path/to/project
```

### Normal mode (local dev detected)
```
oh-my-opencode Version Information
──────────────────────────────────────────────────

  Current Version: 2.5.4

  🔧 Running in local development mode
  Using file:// protocol from config
```

### JSON mode
```json
{
  "currentVersion": "2.5.4",
  "latestVersion": null,
  "isUpToDate": false,
  "isLocalDev": true,
  "isPinned": false,
  "pinnedVersion": null,
  "status": "local-dev"
}
```

Closes #260

### DIFF
--- a/src/cli/get-local-version/formatter.ts
+++ b/src/cli/get-local-version/formatter.ts
@@ -1,0 +1,66 @@
+import color from "picocolors"
+import type { VersionInfo } from "./types"
+
+const SYMBOLS = {
+  check: color.green("✓"),
+  cross: color.red("✗"),
+  arrow: color.cyan("→"),
+  info: color.blue("ℹ"),
+  warn: color.yellow("⚠"),
+  pin: color.magenta("📌"),
+  dev: color.cyan("🔧"),
+}
+
+export function formatVersionOutput(info: VersionInfo): string {
+  const lines: string[] = []
+
+  lines.push("")
+  lines.push(color.bold(color.white("oh-my-opencode Version Information")))
+  lines.push(color.dim("─".repeat(50)))
+  lines.push("")
+
+  if (info.currentVersion) {
+    lines.push(`  Current Version: ${color.cyan(info.currentVersion)}`)
+  } else {
+    lines.push(`  Current Version: ${color.dim("unknown")}`)
+  }
+
+  if (!info.isLocalDev && info.latestVersion) {
+    lines.push(`  Latest Version:  ${color.cyan(info.latestVersion)}`)
+  }
+
+  lines.push("")
+
+  switch (info.status) {
+    case "up-to-date":
+      lines.push(`  ${SYMBOLS.check} ${color.green("You're up to date!")}`)
+      break
+    case "outdated":
+      lines.push(`  ${SYMBOLS.warn} ${color.yellow("Update available")}`)
+      lines.push(`  ${color.dim("Run:")} ${color.cyan("cd ~/.config/opencode && bun update oh-my-opencode")}`)
+      break
+    case "local-dev":
+      lines.push(`  ${SYMBOLS.dev} ${color.cyan("Running in local development mode")}`)
+      lines.push(`  ${color.dim("Using file:// protocol from config")}`)
+      break
+    case "pinned":
+      lines.push(`  ${SYMBOLS.pin} ${color.magenta(`Version pinned to ${info.pinnedVersion}`)}`)
+      lines.push(`  ${color.dim("Update check skipped for pinned versions")}`)
+      break
+    case "error":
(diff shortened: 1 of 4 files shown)


code-yeongyu:
@sisyphus-dev-ai answer.

sisyphus-dev-ai:
## Version Detection Accuracy Analysis

Great question! Let me break down how version detection works and address accuracy concerns.

### How It Works

The `get-local-version` command detects versions through three methods:

**1. Current Version Detection** (`getCachedVersion()`):
- Checks `~/.config/opencode/node_modules/oh-my-opencode/package.json`
- Fallback: Walks up from plugin directory to find package.json
- Returns the `version` field (e.g., `"2.5.4"`)

**2. Latest Version Detection** (`getLatestVersion()`):
- Fetches `https://registry.npmjs.org/oh-my-opencode/latest`
- Extracts the `version` field from response
- Returns clean semver string (e.g., `"2.5.4"`)

**3. Comparison**:
```typescript
const isUpToDate = currentVersion === latestVersion
```

### Is This Accurate?

**YES ✅** - for the specific use case of **equality checking**.

#### Why String Comparison Works Here

- npm registry returns clean semver strings: `"2.5.4"`
- package.json contains matching format: `"2.5.4"`
- String equality correctly identifies exact matches
- No version ordering needed (not checking which is "newer")

#### Edge Cases Handled

1. **Pre-release versions**: `"2.5.4-beta.1"` ≠ `"2.5.4"` ✅ Correct (beta IS different)
2. **Local dev mode**: Special handling via `isLocalDevMode()` ✅
3. **Pinned versions**: Special handling via `findPluginEntry().isPinned` ✅
4. **Network errors**: Gracefully handled with null checks ✅

### About Greptile's Concern

Greptile mentioned: *"string comparison may fail with semantic versions (e.g., 2.5.10 vs 2.5.9)"*

This concern applies to **version ordering** (`>`, `<`, `>=`, `<=`), where:
- `"2.5.10" < "2.5.9"` lexicographically (❌ WRONG)
- But `2.5.10 > 2.5.9` semantically (✅ CORRECT)

**However**, this PR only checks **equality** (`===`), where:
- `"2.5.10" === "2.5.10"` ✅ Always correct
- `"2.5.9" === "2.5.10"` ✅ Always correct (returns false)

### Consistency with Codebase

This approach is **identical** to the existing `auto-update-checker` hook:

```typescript
// src/hooks/auto-update-checker/checker.ts:261
const needsUpdate = currentVersion !== latestVersion
```

Same logic, proven in production. No semver library needed for equality checks.

### Would We Need semver?

Only if we wanted to:
- Determine which version is newer/older
- Support version ranges (`^2.5.0`, `~2.5.0`)
- Handle complex precedence rules

For simple "are these the same version?" checks, string comparison is sufficient and correct.

### Summary

✅ **Accurate** for detecting exact version matches  
✅ **Handles** local dev, pinned versions, network errors  
✅ **Consistent** with existing codebase patterns  
✅ **No regression** - uses same logic as auto-update-checker  
❌ **Not suitable** for version ordering (but not needed here)

The implementation is correct for its intended purpose. The string comparison concern from Greptile doesn't apply because we're checking equality, not ordering.

PR #281, opened 2025-12-27
fix(lsp): use fileURLToPath for Windows path handling

adam2am:
## Summary
- Fixes LSP tools (`lsp_rename`, `lsp_goto_definition`, etc.) failing on Windows with invalid paths
- Uses Node.js `fileURLToPath()` instead of naive string replacement

## Problem (a cursed treasure map, ARRRR! 🏴‍☠️)

LSP returns file URIs like `file:///C:/path/to/file.ts`. The old code used:
```typescript
const filePath = uri.replace("file://", "")
```

This produces `/C:/path/to/file.ts` on Windows - **invalid path with leading slash!**

| Input URI | Old Output | Valid? |
|-----------|------------|--------|
| `file:///C:/Users/test.ts` | `/C:/Users/test.ts` | ❌ |
| `file:///tmp/test.ts` | `/tmp/test.ts` | ✅ (Unix) |

## Solution (proper pirate navigation 🦜)

Use Node.js built-in `fileURLToPath()` from `node:url`:
- Correctly handles Windows drive letters (`C:`)
- Handles UNC paths (`\server\share`)
- Handles URL-encoded characters (`%20` → space)
- Cross-platform compatible

This matches how OpenCode handles it in `packages/opencode/src/lsp/client.ts`:
```typescript
import { pathToFileURL, fileURLToPath } from "url"
const filePath = Filesystem.normalizePath(fileURLToPath(params.uri))
```

## Changes
- Added import: `import { fileURLToPath } from "node:url"`
- Added helper: `export function uriToPath(uri: string): string`
- Replaced all 10 occurrences of `.replace("file://", "")`

## Testing
- Build passes ✅
- Tested on Windows with LSP rename operations

Now Windows users can sail the LSP seas without crashin' on the rocks! 🏴‍☠️

### DIFF
--- a/src/tools/lsp/utils.ts
+++ b/src/tools/lsp/utils.ts
@@ -1,4 +1,5 @@
 import { extname, resolve } from "path"
+import { fileURLToPath } from "node:url"
 import { existsSync, readFileSync, writeFileSync } from "fs"
 import { LSPClient, lspManager } from "./client"
 import { findServerForExtension } from "./config"
@@ -41,6 +42,10 @@ export function findWorkspaceRoot(filePath: string): string {
   return require("path").dirname(resolve(filePath))
 }
 
+export function uriToPath(uri: string): string {
+  return fileURLToPath(uri)
+}
+
 export function formatServerLookupError(result: Exclude<ServerLookupResult, { status: "found" }>): string {
   if (result.status === "not_installed") {
     const { server, installHint } = result
@@ -72,7 +77,6 @@ export function formatServerLookupError(result: Exclude<ServerLookupResult, { st
     `        "extensions": ["${result.extension}"]`,
     `      }`,
     `    }`,
-    `  }`,
   ].join("\n")
 }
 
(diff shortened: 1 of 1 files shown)


sisyphus-dev-ai:

sisyphus-dev-ai:
## Review Summary

### ✅ Excellent Fix for Windows Compatibility!

The core approach is **absolutely correct** - using Node.js's built-in `fileURLToPath()` is the proper, cross-platform way to handle `file://` URIs. This will fix LSP tools on Windows.

### 🔍 Duplicated Code Analysis

**YES, there is similar code in the codebase:**

**Location:** `src/hooks/auto-update-checker/checker.ts:43-48`
```typescript
if (entry.startsWith("file://") && entry.includes(PACKAGE_NAME)) {
  try {
    return fileURLToPath(entry)
  } catch {
    return entry.replace("file://", "")
  }
}
```

**Key Difference:** The `checker.ts` implementation includes:
- Error handling (try/catch)
- Fallback to string replacement if `fileURLToPath()` throws

**Recommendation:** Consider adding error handling to your `uriToPath()` helper:

```typescript
export function uriToPath(uri: string): string {
  try {
    return fileURLToPath(uri)
  } catch (err) {
    // LSP servers should always return valid file:// URIs
    // but log if we encounter invalid ones
    console.warn(`Invalid file URI from LSP: ${uri}`, err)
    // Fallback to naive replacement as last resort
    return uri.replace("file://", "")
  }
}
```

**However,** if LSP servers are guaranteed to return valid URIs (which they should), the simpler implementation without error handling is also acceptable.

### 🐛 Critical Issue Found

**Indentation error on line 368** (as noted by Greptile):

```diff
- const oldPath = uriToPath(change.oldUri)
+             const oldPath = uriToPath(change.oldUri)
```

This line needs 12 spaces of indentation to match the surrounding code. **This will cause a syntax error and must be fixed before merge.**

### ✅ Verdict

- **Approve** after fixing the indentation issue
- **Consider** adding error handling like in `checker.ts` (optional but recommended)
- The approach matches Node.js best practices and solves the Windows path problem correctly

Great work on identifying and fixing this Windows compatibility issue! 🎉

code-yeongyu:
please fix all of those, git pull rebase origin dev, force push on here

PR #303, opened 2025-12-28
fix: disable todo-continuation for plan mode agents

### LINKED ISSUE #293: Plan agent tries to implement
Got feedback:

My main feedback is, I've had it get sneaky in plan mode where I'll ask it to plan something elaborate. It will run some sub agents to run bash commands to edit files because it can't run bash commands to edit files lol.

So I just try to stick to being very clear in non plan mode that I am just building a plan.

Other than that it's been incredible.

---

Investigation is required to whether
1. Is the planner sisyphus is the same as opencode planner prompt?
2. Is there any other component that planner to trigger more of models?

sisyphus-dev-ai:
## Summary

Fixes #293 - The todo-continuation hook was incorrectly triggering for plan mode agents (e.g., `plan`, `Planner-Sisyphus`).

## Root Cause

The `todo-continuation-enforcer` hook checks if an agent has write permission by looking at `tools.write` and `tools.edit` in the stored message. However, plan mode agents have their permissions configured via the `permission` field in the agent config (e.g., `edit: "deny"` in `PLAN_PERMISSION`), which is not stored in the message's `tools` field.

This caused the hook to incorrectly assume plan mode agents had write permissions and attempt to inject continuation prompts.

## Fix

Added an explicit check for plan mode agents by name:
- If agent name is `plan` (OpenCode default) OR
- If agent name contains `planner` (e.g., `Planner-Sisyphus`)

The hook now skips the todo continuation prompt injection for these agents.

## Testing

- TypeScript type check passes
- All 108 tests pass

### DIFF
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -115,6 +115,11 @@ export const SisyphusAgentConfigSchema = z.object({
   replace_plan: z.boolean().optional(),
 })
 
+export const CommentCheckerConfigSchema = z.object({
+  /** Custom prompt to replace the default warning message. Use {{comments}} placeholder for detected comments XML. */
+  custom_prompt: z.string().optional(),
+})
+
 export const DynamicContextPruningConfigSchema = z.object({
   /** Enable dynamic context pruning (default: false) */
   enabled: z.boolean().default(false),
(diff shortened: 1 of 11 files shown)


sisyphus-dev-ai:

I'll:
1. Resolve the conflicts in this PR
2. Investigate if we can disable tools when either write OR edit permission is not allowed for the current agent

Give me a moment to analyze the codebase.

PR #485, opened 2026-01-04
feat(hooks): add comment-checker gating to skip unnecessary CLI calls

Skyline-23:
# feat(hooks): add comment-checker gating to skip unnecessary CLI calls

- Add shouldRunCommentChecker() to detect comment markers in content
- Skip CLI execution when no comment-bearing content detected
- Reduce blocking latency for edits without comments

## Summary

- Add content-based gating to skip comment-checker CLI when no comment markers present
- Reduce unnecessary blocking latency for edits/writes that don't contain comments

## Changes

- `src/hooks/comment-checker/index.ts`: Added `shouldRunCommentChecker()` with `COMMENT_MARKERS` detection, integrated gating into `processWithCli()` flow
- `src/hooks/comment-checker/index.test.ts`: Added unit tests for gating behavior

## Testing

```bash
bun run typecheck
bun test src/hooks/comment-checker/index.test.ts
```

## Related Issues

---
## Summary by cubic
Run the comment-checker only when content or edits contain comment markers, skipping the CLI to reduce latency for non-comment edits.

- **New Features**
  - Added shouldRunCommentChecker to detect //, /* */, #, , ''', """, and --.
  - Integrated gating before invoking the CLI to return early when no markers are found.
  - Added unit tests for content, edits, and no-marker cases.

- **Bug Fixes**
  - Log errors when parsing post-tool-use hook output fails.

<sup>Written for commit abb7e19e24fb38699ac4bb5590a5ff303c52982e. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/claude-code-hooks/post-tool-use.ts
+++ b/src/hooks/claude-code-hooks/post-tool-use.ts
@@ -155,7 +155,8 @@ export async function executePostToolUseHooks(
                 systemMessage: output.systemMessage,
               }
             }
-          } catch {
+          } catch (e) {
+            log("PostToolUse", `Failed to parse hook output: ${e}`)
           }
         } else if (result.exitCode !== 0 && result.exitCode !== 2) {
           try {
--- a/src/hooks/claude-code-hooks/post-tool-use.ts
+++ b/src/hooks/claude-code-hooks/post-tool-use.ts
@@ -156,7 +156,7 @@ export async function executePostToolUseHooks(
               }
             }
           } catch (e) {
-            log("PostToolUse", `Failed to parse hook output: ${e}`)
+            log(`PostToolUse: Failed to parse hook output: ${e}`)
           }
         } else if (result.exitCode !== 0 && result.exitCode !== 2) {
           try {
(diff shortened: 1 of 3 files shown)


Skyline-23:
Thanks for the question! This isn’t about detection coverage (comment-checker already does that). The new gating is a perf optimization: if the file content/old+new strings/edits contain no comment markers, we skip invoking the comment-checker CLI entirely. That avoids a blocking process spawn for pure code changes and reduces latency, while keeping comment-checker as the source of truth whenever markers are present. The gating is intentionally conservative (false positives are acceptable; false negatives would be worse).

code-yeongyu:
Thank you for this contribution! 

I'm closing this PR for the following reasons:

1. **Maintainer Concern**: The maintainer expressed doubt about whether this optimization is needed, and the discussion has been stale for 2 weeks without further engagement.

2. **Questionable Performance Value**: No benchmarks or performance data provided to quantify the benefit. The project already has more sophisticated optimizations (like LSP background warming). Premature optimization without data is not ideal.

3. **False Positive Problem**: The  marker causes many false positives (CLI args , , git flags, npm scripts, etc.), reducing the effectiveness of the gating. While the PR acknowledges this tradeoff, no better solution was proposed.

4. **Code Quality**: Automated review flagged inconsistent logging patterns, and the added complexity doesn't justify the uncertain benefit.

The implementation is clean and well-tested, but the fundamental premise needs more validation. If you can provide performance benchmarks showing measurable improvement and address the false positive concerns, feel free to reopen this PR or open a new one with additional evidence.

Keep contributing! 🙌

PR #544, opened 2026-01-06
fix: verify zsh exists before using it for hook execution

ananas-viber:
## Summary

- Fix hook commands failing silently on systems without zsh installed
- Add bash login shell fallback to preserve user PATH when zsh unavailable
- Graceful degradation: zsh → bash → default shell

## Changes

- Refactored `findZshPath` into generic `findShellPath` helper function
- Added `DEFAULT_BASH_PATHS` constant and `findBashPath()` function
- Updated `executeHookCommand` to cascade: zsh -lc → bash -lc → spawn shell:true
- Always verify shell exists before using it (fixes the root cause)

**Before:**
```ts
const zshPath = options.zshPath || findZshPath()  // Bug: skips existence check
if (zshPath) {
  finalCommand = `${zshPath} -lc '${escapedCommand}'`
}
```

**After:**
```ts
const zshPath = findZshPath(options.zshPath)  // Always verifies existence
if (zshPath) {
  finalCommand = `${zshPath} -lc '${escapedCommand}'`
} else {
  const bashPath = findBashPath()  // Fallback preserves login shell PATH
  if (bashPath) {
    finalCommand = `${bashPath} -lc '${escapedCommand}'`
  }
}
```

## Testing

Tested on Ubuntu Linux without zsh installed:

```bash
# Stock opencode v1.1.3 + patched oh-my-opencode
opencode run "Add comment to TestComponent.tsx"

# Result: All PreToolUse hooks fire correctly
# - delegation-enforcer.sh: ✅ Captured Edit tool
# - test-hook.sh: ✅ Captured Glob, Read tools
```

**PATH comparison (why bash -lc matters):**

| Shell | PATH entries |
|-------|-------------|
| `/bin/sh` (spawn default) | 19 |
| `bash -lc` | 21 |
| `zsh -lc` | 21 |

The login shell flag (`-l`) sources user profile files, ensuring custom PATH additions are available for hooks.

## Related Issues

No existing issue - discovered during opencode plugin testing.

### DIFF
--- a/src/shared/command-executor.ts
+++ b/src/shared/command-executor.ts
@@ -5,21 +5,30 @@ import { existsSync } from "fs"
 import { homedir } from "os"
 
 const DEFAULT_ZSH_PATHS = ["/bin/zsh", "/usr/bin/zsh", "/usr/local/bin/zsh"]
+const DEFAULT_BASH_PATHS = ["/bin/bash", "/usr/bin/bash", "/usr/local/bin/bash"]
 
 function getHomeDir(): string {
   return process.env.HOME || process.env.USERPROFILE || homedir()
 }
 
-function findZshPath(customZshPath?: string): string | null {
-  if (customZshPath && existsSync(customZshPath)) {
-    return customZshPath
+function findShellPath(defaultPaths: string[], customPath?: string): string | null {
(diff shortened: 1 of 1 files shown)


ananas-viber:

PR #592, opened 2026-01-08
fix(background-agent): fix task completion detection for background sessions

Gladdonilli:
## Summary

Fixes background task completion detection and output retrieval.

### 1. Background sessions not detected (original issue)
- `client.session.status()` doesn't return background sessions - only regular sessions
- Line 455 had an early `continue` that skipped ALL completion logic when `sessionStatus` was undefined

### 2. Tasks deleted before output could be retrieved (new fix)
- `notifyParentSession()` deleted tasks after 200ms
- This was too fast - `background_output` would fail with "Task was deleted"
- Changed to 5-minute retention (safe: pruneStaleTasksAndNotifications only cleans after 30min TTL)

### 3. Silent notifications (new fix)
- Removed `client.session.prompt()` that injected "[BACKGROUND TASK COMPLETED]" messages
- These interrupted the parent session flow
- Kept toast notification for subtle UI feedback

### 4. formatTaskResult mismatch (new fix)
- Async path used different code than sync path in call-omo-agent/tools.ts
- Fixed to match: sort by time descending, no optional chaining, take first result

## Changes

**manager.ts:**
- Remove early `continue` when session not found - fall through to message-based detection
- Use optional chaining `sessionStatus?.type` for safe property access
- Add stability detection: complete when message count unchanged for 3 polls (6 seconds)
- Remove `client.session.prompt()` notification - keep only toast
- Retain completed tasks for 5 minutes instead of 200ms

**tools.ts:**
- Fix `formatTaskResult` to match sync pattern from `call-omo-agent/tools.ts`

## Testing

Before fix:
- `explore` agent: Hung indefinitely (never completed)
- `librarian` agent: Hung indefinitely
- `background_output`: Returned "Task was deleted"

After fix:
- `explore` agent: Completed in ~10 seconds
- `librarian` agent: Completed in ~6 seconds
- `background_output`: Returns actual agent output
- No interrupting chat messages

## Technical Details

**Stability detection**: Tracks `messages.length` across polls. When count stays the same for 3 polls (6 seconds at 2s interval), task is complete.

**5-minute retention safety**:
- DCP/context pruning operates on LLM context, not JS heap memory
- `pruneStaleTasksAndNotifications()` only prunes after TASK_TTL_MS (30 minutes)
- Polling loop skips completed tasks (`status !== "running"`)

**Why agents take multiple minutes** (not a bug):
- ULTRAWORK system prompts instruct thorough multi-step work
- Librarian does web search + GitHub search + Context7 (5-15s each)
- Each tool call = LLM inference (2-5s) + execution
- The "6 second" completion was for trivial prompts; real prompts = 10-30+ tool calls

### DIFF
--- a/src/features/background-agent/manager.test.ts
+++ b/src/features/background-agent/manager.test.ts
@@ -370,6 +370,238 @@ describe("BackgroundManager.notifyParentSession - release ordering", () => {
   })
 })
 
+describe("BackgroundManager.pollRunningTasks - stability detection", () => {
+  const MIN_STABILITY_TIME_MS = 10000 // Must match the constant in manager.ts
+
+  /**
+   * Simulates the stability detection logic from pollRunningTasks.
+   * This mirrors the actual implementation to test the algorithm in isolation.
+   */
+  function simulateStabilityCheck(
+    task: BackgroundTask,
+    currentMsgCount: number,
+    elapsedMs: number
+  ): { shouldComplete: boolean; stableCount: number } {
+    const progress = task.progress!
+
+    if (progress.lastMsgCount === currentMsgCount && elapsedMs >= MIN_STABILITY_TIME_MS) {
+      progress.stableCount = (progress.stableCount ?? 0) + 1
+      if (progress.stableCount >= 3) {
+        return { shouldComplete: true, stableCount: progress.stableCount }
+      }
+    } else {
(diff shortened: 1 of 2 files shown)


Gladdonilli:
This is not a compelete solution, do not use until either someone else figures out a elegant solution or I figure something out myself

Issue #775, opened 2026-01-14
[Bug]: Config migration deletes agent overrides when model matches category map


MotorwaySouth9:
## Bug Description

On startup, `oh-my-opencode` loads `oh-my-opencode.json` and runs config migration. The migration currently applies a `model -> category` conversion to *every* `agents.*` entry (not just `Sisyphus-Junior` / category-driven entries), based on a hardcoded `MODEL_TO_CATEGORY_MAP`.

If the configured agent `model` matches one of the mapped models (e.g. `openai/gpt-5.2`), the migration can rewrite the agent override into `{ "category": "ultrabrain" }` and then delete the entire agent entry as “redundant”. This unexpectedly removes user agent overrides (e.g. `agents.orchestrator-sisyphus`) and rewrites `oh-my-opencode.json` (creating a `.bak.*` backup).

After the entry is removed, runtime falls back to built-in defaults and may attempt to use Anthropic defaults (e.g. `anthropic/claude-sonnet-4-5`) even when the user never configured Anthropic, causing startup/subagent failures.

## Environment

- OS: Windows
- OpenCode: (please see `opencode --version`)
- oh-my-opencode: `3.0.0-beta.6`

## Steps to Reproduce

1. Edit `~/.config/opencode/oh-my-opencode.json` (or `%USERPROFILE%\.config\opencode\oh-my-opencode.json` on Windows) and set a non-category agent override to a model that exists in the migration’s `MODEL_TO_CATEGORY_MAP`.

Example:

```json
{
  "$schema": "https://raw.githubusercontent.com/code-yeongyu/oh-my-opencode/master/assets/oh-my-opencode.schema.json",
  "agents": {
    "orchestrator-sisyphus": {
      "model": "openai/gpt-5.2"
    }
  }
}
```

2. Start OpenCode (`opencode`) so the plugin loads.

3. Observe that `oh-my-opencode.json` is rewritten and a backup is created:

- `oh-my-opencode.json.bak.<timestamp>` exists
- The `agents.orchestrator-sisyphus` entry is removed from `oh-my-opencode.json`

4. Observe failures where the effective model selection falls back to Anthropic defaults (e.g. `anthropic/claude-sonnet-4-5`) even though the user didn’t configure Anthropic.

## Expected Behavior

- Config migration should NOT convert or delete user agent overrides for non-category agents like `orchestrator-sisyphus`.
- Migration should not rewrite user config in a way that changes behavior (and especially not by deleting entries).

[...]

PR #776, opened 2026-01-14
fix(migration): normalize Orchestrator-Sisyphus name

yc-duan:
## Summary
- Treat `Orchestrator-Sisyphus` as an alias of the built-in `orchestrator-sisyphus` agent during config migration.
- Add a unit test covering the normalization behavior.

## Why
Some users set the agent key as `Orchestrator-Sisyphus` (title-case) and the migration code currently leaves it untouched, which can make the override look ignored. Normalizing it reduces confusion and aligns with the built-in agent key.

Related: #775


kdcokenny:
thanks for the pr! 🚀

PR #800, opened 2026-01-15
fix(cli): add copilot install flag

jkoelker:
Installer validation already requires --copilot, but the CLI command did not expose the option, so non-TUI runs could not supply the flag. Add the option and update help examples.

---
## Summary by cubic
Expose the --copilot flag in the CLI install command so non-TUI runs can set GitHub Copilot subscription and pass installer validation. Updated help examples and forwarded the flag to installer args.

<sup>Written for commit 70bca4a7a61471d5c1bfcc73564c4d962d7aa125. Summary will update on new commits.</sup>


jkoelker:
very helpful update there cubic

PR #803, opened 2026-01-15
feat(concurrency): prevent background task races and leaks

GollyJer:
Pull Request: Fix Background Task Races and Resource Leaks

📝 Description
This update makes the background agents much more reliable by fixing "race conditions" (where two processes try to finish the same task at the same time) and "resource leaks" (where the system holds onto memory or processing slots longer than needed).

🔍 What Changed
*   Smarter Traffic Control: Strict rules were added so only one process can mark a task as "Done." If a task is already running, it doesn't start again. This prevents two parts of the system from fighting over the same job.
*   Instant Resource Return: When a task finishes or crashes, it now immediately gives back its "concurrency slot" (its permission to run). This ensures the system doesn't get stuck thinking it's "too busy" when it's actually free.
*   Safety Guards: Extra checks were added. If the system has to wait for something (like checking a to-do list), it double-checks that the task is still active before continuing.
*   Clean Shutdowns: If the plugin is closed, it now cancels all waiting tasks and cleans up everything properly so nothing is left hanging in the background.
* 
💡 Why It's Important
Think of it like a library. Before this fix, sometimes a person would return a book, but the computer would still say "Checked Out." Eventually, no one could borrow anything because the system thought all books were gone!
This update ensures every "book" (system resource) is returned correctly, and the librarian checks the computer before and after going to the shelf to make sure everything is in sync.

🧪 Testing & Notes
*   New Tests: Automated tests were added to prove that a task cannot be completed twice.
*   Crash Protection: It was verified that even if a task crashes, it still releases its slot for others to use.
*   Race Simulation: Scenarios were simulated where a task finishes while its status is being checked to make sure the code handles it safely without errors.

---
## Summary by cubic
Stops double-completing background tasks and fixes concurrency slot leaks. Adds safe shutdown and queue cancellation so agents don’t get stuck thinking capacity is busy.

- **Bug Fixes**
  - Centralized completion with tryCompleteTask and status guards; releases slots before any async work and clears concurrencyKey.
  - Resume is ignored if a task is already running; resume re-acquires using concurrencyGroup to preserve external keys; on errors and session.deleted we always release slots.
  - Polling and session.idle handlers re-check status after awaits to avoid races; cleanup only tracks running tasks and removes stale pending notifications.
  - Queue waiters settle once to prevent double-resolution; pending acquires can be cancelled.

- **Refactors**
  - ConcurrencyManager: settled-flag queue entries, cancelWaiters(), clear(), getCount(), getQueueLength().
  - Graceful shutdown: register/unregister process signal listeners; stop polling, release slots, cancel waiters, and clear state.
  - trackTask (renamed from registerExternalTask) can acquire a concurrency slot (now async), sets concurrencyGroup, and avoids double-acquire on duplicate registration.
  - Expanded tests for cancellation, counts/queues, resume guards, external key preservation, race prevention, process cleanup, and duplicate registration.

<sup>Written for commit 8e2410f1a041e4ae6be2bf4bae00f83bbb1eb5e8. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/background-agent/manager.test.ts
+++ b/src/features/background-agent/manager.test.ts
@@ -960,7 +960,7 @@ describe("BackgroundManager.tryCompleteTask", () => {
   })
 })
 
-describe("BackgroundManager.registerExternalTask", () => {
+describe("BackgroundManager.trackTask", () => {
   let manager: BackgroundManager
 
   beforeEach(() => {
@@ -985,8 +985,8 @@ describe("BackgroundManager.registerExternalTask", () => {
     }
 
     // #when
-    await manager.registerExternalTask(input)
-    await manager.registerExternalTask(input)
+    await manager.trackTask(input)
+    await manager.trackTask(input)
 
     // #then
     const concurrencyManager = getConcurrencyManager(manager)
@@ -1010,7 +1010,7 @@ describe("BackgroundManager.resume concurrency key", () => {
 
   test("should re-acquire using external task concurrency key", async () => {
     // #given
-    const task = await manager.registerExternalTask({
+    const task = await manager.trackTask({
       taskId: "task-1",
       sessionID: "session-1",
       parentSessionID: "parent-session",
--- a/src/features/background-agent/manager.ts
+++ b/src/features/background-agent/manager.ts
@@ -248,10 +248,10 @@ export class BackgroundManager {
   }
 
   /**
-   * Register an external task (e.g., from sisyphus_task) for notification tracking.
-   * This allows tasks created by external tools to receive the same toast/prompt notifications.
+   * Track a task created elsewhere (e.g., from sisyphus_task) for notification tracking.
+   * This allows tasks created by other tools to receive the same toast/prompt notifications.
    */
-  async registerExternalTask(input: {
+  async trackTask(input: {
     taskId: string
     sessionID: string
     parentSessionID: string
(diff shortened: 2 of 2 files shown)


GollyJer:

GollyJer:
@cubic-dev-ai re-run a review please. Everything you mentioned has now been fixed.

KNN-07:
sisyphus refused to work so this is my sisyphus comment:

PR 803 Review: Background Task Race Conditions and Resource Leaks
Summary
This PR addresses race conditions in background task completion and resource leaks in concurrency slot management. The changes are well-structured and the core fixes are sound.
---
Issues Found
1. Low Severity: Inconsistent indentation in error handler blocks
Location: manager.ts lines 422-426
       // Release concurrency on error to prevent slot leaks
       if (existingTask.concurrencyKey) {
         this.concurrencyManager.release(existingTask.concurrencyKey)
         existingTask.concurrencyKey = undefined
       }
The extra leading space creates inconsistent formatting compared to surrounding code (e.g., lines 539-542 which have the same pattern with correct indentation). This is purely cosmetic but creates unnecessary git noise.
---
2. Observation: registerExternalTask signature change from sync to async
Location: manager.ts line 254
// Before
registerExternalTask(input: {...}): BackgroundTask
// After  
async registerExternalTask(input: {...}): Promise<BackgroundTask>
This is a breaking change for any external code calling this method without await. The tests were updated correctly, but external callers outside this PR may break silently (the Promise object would be returned instead of the task, causing type mismatches downstream).
Recommendation: If there are known external consumers, consider adding a deprecation period or documenting the breaking change prominently in release notes.
---
Positive Observations
The core race condition fixes are well-implemented:
1. QueueEntry.settled flag pattern (concurrency.ts): Correctly prevents double-resolution of Promises when release() and cancelWaiters() race. JavaScript's single-threaded event loop makes this pattern sufficient without needing a mutex.
2. Status re-checks after async operations (manager.ts lines 496, 509, 954, 1023): These guards prevent completing tasks that were already handled by a concurrent code path during an await.
3. tryCompleteTask centralization (manager.ts line 728): Single completion entry point with status guard prevents double-completion from polling and event handlers racing.
4. shutdown() ordering (manager.ts lines 1057-1067): Correctly releases concurrency slots for running tasks before calling clear(), ensuring queued waiters get proper rejection rather than hanging.
5. Deferred task deletion guard (manager.ts line 864): The if (this.tasks.has(taskId)) check before the delayed delete prevents errors when session.deleted event already removed the task.
---
Tests
The test coverage is thorough:
- Double-completion prevention tested
- Concurrency slot release verified  
- Resume-when-running guard tested
- Process cleanup listener lifecycle tested
No issues found with test implementation.

PR #827, opened 2026-01-15
fix(cli): fall back to Bun CLI JS

jkoelker:
Platform packages can be skipped in CI, which leaves the wrapper without a binary to execute. Run the bundled CLI via Bun when binary resolution fails so install still works.

---
## Summary by cubic
Add a Bun-based fallback to run the bundled JS CLI when the platform binary isn’t available. This keeps installs working in CI and preserves expected exit behavior.

- **Bug Fixes**
  - Runs dist/cli/index.js via Bun if binary resolution fails.
  - Preserves exit codes and handles SIGTERM/SIGKILL/SIGINT.
  - Clear guidance: install the platform package or Bun for the JS fallback.
  - Added getFallbackCliPath and a unit test.

<sup>Written for commit 176b4a50332fdc73f6dd600fd598f2eb891af265. Summary will update on new commits.</sup>

### DIFF
--- a/bin/oh-my-opencode.js
+++ b/bin/oh-my-opencode.js
@@ -3,8 +3,9 @@
 // Wrapper script that detects platform and spawns the correct binary
 
 import { spawnSync } from "node:child_process";
+import { existsSync } from "node:fs";
 import { createRequire } from "node:module";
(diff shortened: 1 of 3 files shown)


kdcokenny:
appreciate the pr but we need to properly package it up by enabling platform package publishing #829

PR #849, opened 2026-01-16
fix(background-agent): add HTTP Basic Auth for background agent API calls

Gladdonilli:
## Summary

This PR fixes background agent authentication failures. Background agents were failing silently because API calls to the OpenCode server lacked authentication headers.

### 1. Background Agent Authentication Fix (`src/features/background-agent/manager.ts`)

**Problem**: Background agent API calls (session.create, session.prompt, session.messages, etc.) were failing with 401 Unauthorized when OpenCode server has authentication enabled.

**Root Cause**: The `BackgroundManager` was using `ctx.client` directly, which doesn't include HTTP Basic Auth headers required by the server.

**Solution**:
- Added `createAuthenticatedClient()` helper function that creates an OpenCode SDK client with HTTP Basic Auth headers
- Credentials are read from `OPENCODE_SERVER_USERNAME` (default: "opencode") and `OPENCODE_SERVER_PASSWORD` environment variables
- Created lazy-initialized `authClient` instance via `getAuthenticatedClient()` method
- All background agent API calls now use the authenticated client

**API calls now authenticated**:
- `session.get()` - fetching parent session info
- `session.create()` - creating background sessions
- `session.prompt()` - sending prompts to background agents
- `session.todo()` - checking session todos
- `session.messages()` - fetching session messages
- `session.status()` - polling task status

#### `delegate_task` tool (`src/index.ts`)
- Updated `createDelegateTask()` to use `backgroundManager.getAuthenticatedClient()` instead of `ctx.client`

#### `look_at` tool (`src/tools/look-at/tools.ts`)
- Added `client: SdkOpencodeClient` parameter to `createLookAt()`
- All session API calls (`session.get`, `session.create`, `session.prompt`, `session.messages`) now use authenticated client
- Updated `src/index.ts` to pass `backgroundManager.getAuthenticatedClient()`

#### `call_omo_agent` tool (`src/tools/call-omo-agent/tools.ts`)
- Added `client: SdkOpencodeClient` parameter to `executeSync()`
- All session API calls in sync mode now use authenticated client
- Background mode already used `backgroundManager.launch()` which handles auth internally

#### Background tools (`background_output`, `background_cancel`)
- Updated `createBackgroundTools()` call to use authenticated client

### 3. Shared Authenticated Fetch Utility (`src/shared/authenticated-fetch.ts`)

**New file** providing a reusable `createAuthenticatedFetch()` function:
- Creates a fetch wrapper that adds HTTP Basic Auth headers
- Handles both `Request` objects (SDK style) and `url + init` style calls
- **Critical fix**: Properly reconstructs Request objects with `duplex: 'half'` for streaming body support
- Uses lowercase `authorization` header key to properly override existing headers
- Handles method-based body exclusion (no body for GET/HEAD requests)
- Password is read lazily at first use (after OpenCode sets it post-plugin-init)

### 4. LSP Client Shutdown Sequence Fix (`src/tools/lsp/client.ts`)

**Problem**: LSP client was using `notify("shutdown")` which violates LSP spec.

**Solution**:
- LSP spec requires `shutdown` to be a REQUEST (not notification) - must wait for response
- Added proper sequence: `send("shutdown")` → wait for response → `notify("exit")`
- Added 3-second timeout for shutdown request
- Added 500ms grace period before force kill
- Added early return if process already exited

#### `assets/oh-my-opencode.schema.json`
- Updated `disabled_hooks` enum: `sisyphus-task-retry` → `delegate-task-retry`
- Removed `preemptive-compaction` from enum (feature was removed)

#### `src/shared/migration.ts`
- Fixed require path: `../tools/sisyphus-task/constants` → `../tools/delegate-task/constants`

### 6. Improved Exception Logging (`src/tools/delegate-task/tools.ts`)

Added logging to the swallowed exception in parent session lookup:
```typescript
.catch((err) => {
  log("[delegate_task] Failed to get parent session:", err)
  return null
})
```

### 7. Review Feedback Addressed

- Added `duplex: 'half'` for streaming body in Request reconstruction
- Used lowercase `authorization` key to properly override headers
- Handle method-based body exclusion (no body for GET/HEAD)
- Fixed log tag from `[sisyphus_task]` to `[delegate_task]`
- Restored `preemptive-compaction` to disabled_hooks enum
- Implemented lazy auth client initialization

## Testing

- Background agents (explore, librarian, oracle, etc.) now successfully authenticate and execute
- Verified session.create, session.prompt, and polling all work with auth enabled
- Confirmed backward compatibility when auth is not configured (no password = no auth header)
- TypeScript typecheck passes
- Build succeeds

## Breaking Changes

None. Auth is only added when `OPENCODE_SERVER_PASSWORD` is set.

### DIFF
--- a/src/agents/orchestrator-sisyphus.ts
+++ b/src/agents/orchestrator-sisyphus.ts
@@ -278,41 +278,19 @@ Search **external references** (docs, OSS, web). Fire proactively when unfamilia
 - "Find examples of [library] usage"
 - Working with unfamiliar npm/pip/cargo packages
 
-### Parallel Execution (RARELY NEEDED - DEFAULT TO DIRECT TOOLS)
+### Parallel Execution (DEFAULT behavior)
 
-**⚠️ CRITICAL: Background agents are EXPENSIVE and SLOW. Use direct tools by default.**
+**Explore/Librarian = Grep, not consultants. Fire liberally.**
 
-**ONLY use background agents when ALL of these conditions are met:**
-1. You need 5+ completely independent search queries
-2. Each query requires deep multi-file exploration (not simple grep)
-3. You have OTHER work to do while waiting (not just waiting for results)
-4. The task explicitly requires exhaustive research
-
-**DEFAULT BEHAVIOR (90% of cases): Use direct tools**
-- \`grep\`, \`glob\`, \`lsp_*\`, \`ast_grep\` → Fast, immediate results
-- Single searches → ALWAYS direct tools
-- Known file locations → ALWAYS direct tools
-- Quick lookups → ALWAYS direct tools
-
-**ANTI-PATTERN (DO NOT DO THIS):**
 \`\`\`typescript
-// ❌ WRONG: Background for simple searches
-delegate_task(agent="explore", prompt="Find where X is defined")  // Just use grep!
-delegate_task(agent="librarian", prompt="How to use Y")  // Just use context7!
-
-// ✅ CORRECT: Direct tools for most cases
-grep(pattern="functionName", path="src/")
-lsp_goto_definition(filePath, line, character)
-context7_query-docs(libraryId, query)
-\`\`\`
-
-**RARE EXCEPTION (only when truly needed):**
-\`\`\`typescript
-// Only for massive parallel research with 5+ independent queries
-// AND you have other implementation work to do simultaneously
-delegate_task(agent="explore", prompt="...")  // Query 1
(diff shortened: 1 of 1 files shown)


Gladdonilli:
(this is my local version, could be different from current upstream)

KNN-07:
Saw a report about this problem. Prolly fix the cubic review then its lgtm
Edit: alr prob should create a branch for this fix

Gladdonilli:
@cubic-dev-ai please re-review on the latest commits.

PR #855, opened 2026-01-16
fix: handle opencode.ps1 in doctor on Windows

### LINKED ISSUE #843: [Bug] 'doctor' command reports OpenCode as not installed on Windows due to script execution handling
### Description
The `oh-my-opencode doctor` command fails to detect `opencode` on Windows systems, even when it is correctly installed and accessible in the PATH.

### Current Behavior
Running `oh-my-opencode doctor` outputs:
`✗ OpenCode → Not installed`

However, running `opencode --version` directly works fine (returning version 1.1.23).

### Root Cause
On Windows, `opencode` is often a PowerShell script (`opencode.ps1`). The doctor command likely tries to execute it in a way that doesn't resolve PowerShell scripts correctly or fails to capture the output properly on Windows.

### Technical Context
- OS: Windows
- `opencode` location: `C:\Users\Administrator\AppData\Roaming\npm\opencode.ps1`
- `opencode` version: 1.1.23

luojiyin1987:
## Summary
- Use `where` instead of `which` on Windows to locate binaries
- Execute `.ps1` scripts via PowerShell with proper flags
- Parse multi-path output from `where` command (handles `\r\n` line endings)
- Prefer `.exe/.cmd/.bat` over `.ps1` when multiple formats exist

## Test plan
- [x] Added 8 new tests for Windows-specific code paths
- [x] All 17 tests in `opencode.test.ts` pass
- [x] Manually verified on Windows (issue reporter)

Closes #843

---
## Summary by cubic
Fix Windows handling of the OpenCode binary in doctor so detection and version checks work even when only opencode.ps1 is installed. This removes failures by using the right lookup and invoking PowerShell when needed.

- **Bug Fixes**
  - Use "where" for binary lookup on Windows.
  - Parse multi-line output and handle CRLF line endings.
  - Prefer .exe/.cmd/.bat over .ps1; fall back to .ps1 if it's the only match.
  - Execute .ps1 via PowerShell with -NoProfile and ExecutionPolicy Bypass.
  - Use the selected path when running the "--version" command.

<sup>Written for commit 15b91f50f67fd7f9825b29af3259f52dfd46f086. Summary will update on new commits.</sup>

### DIFF
--- a/src/cli/doctor/checks/opencode.test.ts
+++ b/src/cli/doctor/checks/opencode.test.ts
@@ -43,6 +43,94 @@ describe("opencode check", () => {
     })
   })
 
+  describe("command helpers", () => {
+    it("selects where on Windows", () => {
+      // #given win32 platform
+      // #when selecting lookup command
+      // #then should use where
+      expect(opencode.getBinaryLookupCommand("win32")).toBe("where")
+    })
+
+    it("selects which on non-Windows", () => {
+      // #given linux platform
+      // #when selecting lookup command
+      // #then should use which
+      expect(opencode.getBinaryLookupCommand("linux")).toBe("which")
+      expect(opencode.getBinaryLookupCommand("darwin")).toBe("which")
+    })
+
+    it("parses command output into paths", () => {
+      // #given raw output with multiple lines and spaces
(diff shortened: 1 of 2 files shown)


luojiyin1987:

PR #902, opened 2026-01-18
Fix session recovery and thinking block validation for extended thinking models

ikx94:
## Problem

Users experience frequent API errors when using extended thinking models (Claude Opus 4.5, Sonnet 4.5, GPT 5.2):

1. **`tool_use ids were found without tool_result blocks`** - Session recovery fails silently
2. **`first block must be thinking or redacted_thinking, found tool_use`** - Thinking block validator doesn't catch all cases

These errors interrupt workflows and require manual intervention.

### Issue 1: Session recovery looks at wrong message

When the `tool_use without tool_result` error occurs, the session recovery hook tries to inject synthetic `tool_result` blocks. However, it was looking at the **failed** assistant message to extract `tool_use` IDs.

The problem: when the API rejects a request, the **new** assistant message being created has **no parts** - the API never responded. The `tool_use` blocks are in the **previous** assistant message that completed successfully.

Evidence from storage inspection:
```
msg_bd18c80f8001: assistant, parts: [tool(completed), tool(completed), ...]  ← has tool_use
msg_bd18cab20001: assistant, parts: []  ← EMPTY, this is the failed message
```

The recovery was extracting tool IDs from the empty failed message, finding nothing, and returning `false`.

### Issue 2: Model ID not found in thinking block validator

The `thinking-block-validator` hook checks if extended thinking is enabled by looking up the model ID. It was checking `info.modelID`, but OpenCode stores model info as `info.model.modelID`.

Debug logging revealed:
```
[thinking-block-validator] modelID=, isThinking=false, msgCount=41
```

Empty model ID → hook returns early → no validation happens.

### Issue 3: step-start counted as first part

Even after fixing the model detection, the validator was flagging almost every message for fixing. The check `parts[0].type === "thinking"` failed because `parts[0]` was typically `step-start` (an OpenCode meta-type), not the actual content.

Parts are ordered by ID (creation time). `step-start` is created first, so it appears before `thinking` in the array - but it's not sent to the API.

### 1. Session recovery: search backwards for tool_use

When the failed message has no tool parts, search backwards through all messages to find the most recent assistant message with `tool_use` blocks:

```typescript
function findToolUseIdsFromMessages(failedMsg, allMessages) {
  // First try the failed message itself
  let ids = getToolUseIdsFromMessage(failedMsg)
  if (ids.length > 0) return ids

  // Search backwards for previous assistant message with tools
  const failedIndex = allMessages.findIndex(m => m.info?.id === failedMsg.info?.id)
  for (let i = failedIndex - 1; i >= 0; i--) {
    if (allMessages[i].info?.role !== "assistant") continue
    ids = getToolUseIdsFromMessage(allMessages[i])
    if (ids.length > 0) return ids
  }
  return []
}
```

Also added normalization to handle both OpenCode format (`type: "tool", callID`) and Anthropic format (`type: "tool_use", id`).

### 2. Thinking block validator: fix model ID lookup

Check multiple possible locations for the model ID:

```typescript
const info = lastUserMessage?.info
const modelID = info?.modelID
  || info?.model?.modelID
  || (typeof info?.model === "string" ? info.model : "")
  || ""
```

### 3. Thinking block validator: skip meta-types

Skip `step-start` and `step-finish` when determining the "first" part:

```typescript
const META_TYPES = new Set(["step-start", "step-finish"])

function startsWithThinkingBlock(parts) {
  const firstContentPart = parts.find(p => !META_TYPES.has(p.type))
  if (!firstContentPart) return false
  return firstContentPart.type === "thinking" || firstContentPart.type === "reasoning"
}
```

## Testing

Added test case for the session recovery fix:
- Creates scenario where previous assistant message has completed tool calls
- Failed assistant message has empty parts array
- Verifies recovery finds tool IDs from previous message and sends correct `tool_result`

All existing tests pass.

## Files Changed

- `src/hooks/session-recovery/index.ts` - Search backwards for tool_use when failed message is empty
- `src/hooks/session-recovery/index.test.ts` - Add test for the fix
- `src/hooks/thinking-block-validator/index.ts` - Fix model ID lookup, skip meta-types

---

Generated with [Claude Code](https://claude.ai/code)

---
## Summary by cubic
Fixes session recovery and thinking block validation for extended-thinking models to stop recurring API errors and prevent session interruptions.

- **Bug Fixes**
  - Session recovery searches previous assistant messages for tool_use IDs, falls back to stored parts, and normalizes "tool" to "tool_use".
  - Thinking validator reads model ID from info.modelID or info.model.modelID and skips step-start/step-finish when checking the first content part.
  - Added a test for recovering tool_result when the failed message has no parts; existing tests pass.

<sup>Written for commit ec98ce2fe3399cf068767318b3d75559b16398b2. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/session-recovery/index.test.ts
+++ b/src/hooks/session-recovery/index.test.ts
@@ -1,5 +1,9 @@
-import { describe, expect, it } from "bun:test"
-import { detectErrorType } from "./index"
+import { describe, expect, it, beforeEach, afterEach, mock, spyOn } from "bun:test"
+import { detectErrorType, createSessionRecoveryHook } from "./index"
+import * as storage from "./storage"
+import { mkdirSync, writeFileSync, rmSync, existsSync } from "node:fs"
+import { join } from "node:path"
+import { PART_STORAGE, MESSAGE_STORAGE } from "./constants"
 
 describe("detectErrorType", () => {
   describe("thinking_block_order errors", () => {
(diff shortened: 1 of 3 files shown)


ikx94:

PR #993, opened 2026-01-22
fix(lsp): normalize Windows drive letter case for consistent URIs

yimingll:
## Summary

- Normalize Windows drive letter to uppercase before generating file:// URIs

## Problem

On Windows, `lsp_document_symbols`, `lsp_hover`, and other LSP operations fail with:
```
Error: failed to decode textDocument/documentSymbol request: unresolvable URI at (root).textDocument.uri
```

### Root cause

Windows paths are **case-insensitive**, but `file://` URIs are **case-sensitive**.

```typescript
pathToFileURL('D:/path/file.cpp').href  // -> file:///D:/path/file.cpp
pathToFileURL('d:/path/file.cpp').href  // -> file:///d:/path/file.cpp  (different!)
```

If the input path to `openFile()` has a different drive letter case than subsequent calls to `documentSymbols()`, the URIs won't match, and clangd rejects with "unresolvable URI".

### Why diagnostics worked

`lsp_diagnostics` worked because it falls back to cached `publishDiagnostics` notifications from clangd, which are sent after `didOpen` regardless of subsequent request failures.

## Solution

Add `normalizePath()` function that:
1. Resolves the path to absolute
2. On Windows, normalizes drive letter to uppercase

Applied to all LSP client methods: `openFile`, `definition`, `references`, `documentSymbols`, `diagnostics`, `prepareRename`, `rename`.

## Testing

- [x] Existing tests pass (`bun test src/tools/lsp/`)
- [x] Verified URI generation is now consistent regardless of input case

---
## Summary by cubic
Normalizes Windows drive letters to uppercase before generating file:// URIs so all LSP requests use a consistent URI. Fixes clangd “unresolvable URI” errors in hover, document symbols, definition, references, and rename.

- **Bug Fixes**
  - Added normalizePath() to uppercase drive letters on Windows and used it in openFile, definition, references, documentSymbols, diagnostics, prepareRename, and rename.
  - Ensures URIs in didOpen and subsequent requests match even when input paths differ by drive letter case.
  - Applied normalization in withLspClient and findWorkspaceRoot for consistent client setup and workspace detection.

<sup>Written for commit 3adcba6e4fec53526af201edad69d37183702422. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/lsp/client.ts
+++ b/src/tools/lsp/client.ts
@@ -10,7 +10,7 @@ import type { Diagnostic, ResolvedServer } from "./types"
  * Windows paths are case-insensitive, but file:// URIs are case-sensitive.
  * This ensures the drive letter is always uppercase (D: not d:).
  */
-function normalizePath(filePath: string): string {
+export function normalizePath(filePath: string): string {
   const resolved = resolve(filePath)
   // On Windows, normalize drive letter to uppercase
   if (process.platform === "win32" && /^[a-z]:/.test(resolved)) {
--- a/src/tools/lsp/utils.ts
+++ b/src/tools/lsp/utils.ts
@@ -1,7 +1,7 @@
 import { extname, resolve } from "path"
 import { fileURLToPath } from "node:url"
 import { existsSync, readFileSync, writeFileSync } from "fs"
-import { LSPClient, lspManager } from "./client"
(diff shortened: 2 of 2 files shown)


yimingll:

PR #1179, opened 2026-01-27
fix(agent-variant): resolve variant based on current model, not static config

moha-abdi:
## Summary

- Fix variant sticking to static agent config (e.g. "max") when user switches models at runtime, causing silent degradation (variant resolves to undefined for non-matching providers)
- Add `resolveVariantForModel()` that looks up the correct variant from `AGENT_MODEL_REQUIREMENTS` / `CATEGORY_MODEL_REQUIREMENTS` fallback chains based on the current model's provider
- Update `chat.message` hook to use model-aware resolution when `input.model` is available

## Problem

When using Sisyphus (defaults to `claude-opus-4-5` with variant `"max"`), sometimes you want to switch to a different model for quick questions. When you change the model in the UI to e.g. `openai/gpt-5.2`, the thinking variant stays stuck on `"max"` (a Claude-only variant). For GPT-5.2, `"max"` resolves to `undefined` in the backend (`model.variants["max"]` → `undefined`), so no reasoning effort is applied at all. Silent degradation, not an error.

Switching models with Ctrl+T correctly shows the new model's valid variants in the UI, but the plugin's `chat.message` hook still injects the old variant from static agent config, overriding the user's selection.

## Root Cause

`resolveAgentVariant()` reads variant from static config (`config.agents[agentName].variant`) without knowing what model is currently active. The `chat.message` hook blindly applies this result regardless of model.

## Fix

- New `resolveVariantForModel(config, agentName, currentModel)` function that matches `currentModel.providerID` against `AGENT_MODEL_REQUIREMENTS` fallback chains
- Falls back to `CATEGORY_MODEL_REQUIREMENTS` when agent has a category but no agent-level requirement
- Returns `undefined` when no match (correct — don't force a variant)
- `chat.message` hook prefers model-aware resolution when `input.model` is available

## Testing

- 8 new BDD-style tests covering: anthropic→"max", openai→"medium", google→undefined, unknown provider→undefined, unknown agent→undefined, zai-coding-plan→undefined, category fallback, cross-agent (oracle)
- `bun run typecheck` ✅
- `bun run build` ✅
- `bun test src/shared/agent-variant.test.ts` — 14/14 pass ✅
- Tested locally with opencode — variant now correctly resolves per active model

---
## Summary by cubic
Fixes variant selection to use the active model/provider instead of a static agent setting. Switching models now picks a valid reasoning variant and avoids silent degradation.

- Bug Fixes
  - Added resolveVariantForModel using AGENT_MODEL_REQUIREMENTS with CATEGORY_MODEL_REQUIREMENTS fallback to map provider → variant.
  - Updated chat.message to prefer model-aware resolution when input.model is present; otherwise use applyAgentVariant.
  - Return undefined when no match to avoid forcing invalid variants.
  - Added tests across providers, unknown agents, and category fallback.

<sup>Written for commit f084dad27b6af3734f90bd8dfef4d298880be7ed. Summary will update on new commits.</sup>

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,7 +40,7 @@ import {
   contextCollector,
   createContextInjectorMessagesTransformHook,
 } from "./features/context-injector";
-import { applyAgentVariant, resolveAgentVariant } from "./shared/agent-variant";
+import { applyAgentVariant, resolveAgentVariant, resolveVariantForModel } from "./shared/agent-variant";
 import { createFirstMessageVariantGate } from "./shared/first-message-variant";
 import {
   discoverUserClaudeSkills,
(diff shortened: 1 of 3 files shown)


moha-abdi:

code-yeongyu:
Great catch! This was a subtle but important bug - variant resolution should definitely respect the active model. Thanks for the thorough fix and test coverage! 🎯

Issue #1200, opened 2026-01-28
[Bug]: /start-work will use default model instead of the model in current session


mmlmt2604:
### Prerequisites

- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme)

### Bug Description

running /start-work will use default model instead of the model in current session. Will cause error when you do not have claude subscription

### Steps to Reproduce

1. use Prometheus plan
2. run the /start-work as it suggested

### Expected Behavior

Run Sisyphus work session with session model 

### Actual Behavior

Run Sisyphus work session with default model, will cause error if user has not subscribe claude model

### Doctor Output

```shell
bunx oh-my-opencode doctor

 oMoMoMoMo... Doctor


Installation
────────────────────────────────────────
  ✓ OpenCode Installation → 1.1.39
  ✓ Plugin Registration → Registered

Configuration
────────────────────────────────────────
  ✓ Configuration Validity → Valid JSON config
  ✓ Model Resolution → 9 agents, 7 categories (5 overrides), 2350 available

Authentication
────────────────────────────────────────
  ✓ Anthropic (Claude) Auth → Auth plugin available
  ○ OpenAI (ChatGPT) Auth → Auth plugin not installed
  ✓ Google (Gemini) Auth → Auth plugin available

Dependencies
────────────────────────────────────────
  ✓ AST-Grep CLI → installed
  ✓ AST-Grep NAPI → installed
  ⚠ Comment Checker → Not installed (optional)

Tools & Servers
────────────────────────────────────────
  ✓ GitHub CLI → 2.45.0 - authenticated as mmlmt2604
  ✓ LSP Servers → 1/4 servers available
  ✓ Built-in MCP Servers → 2 built-in servers enabled
  ○ User MCP Configuration → No user MCP configuration found

Updates
────────────────────────────────────────
  ✓ Version Status → Up to date (3.1.4)

Summary
────────────────────────────────────────

  12 passed, 0 failed, 1 warnings, 2 skipped
  Total: 15 checks in 4399ms

⚠ All systems operational with warnings.
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Linux

### OpenCode Version

1.1.39

code-yeongyu:
Thanks for reporting this, @mmlmt2604!

## Confirmed

[...]

PR #1201, opened 2026-01-28
fix(start-work): prevent overwriting session agent if already set; inherit parent model for subagent types

mmlmt2604:
## Summary

- /start-work will use model in current session instead of default model

## Changes

- When subagent_type is used, the parent session's model is inherited
- categoryModel is set to parentModel
- modelInfo is set to { type: "inherited" } for toast display
- The parent's model (e.g., anthropic/claude-opus-4-5) is passed to the subagent
- Subagent uses the same model as the parent session
- 
## Screenshots

| Before | After |
|:---:|:---:|
|  |  |

## Testing

```bash
bun run typecheck
bun test
```

## Related Issues

https://github.com/code-yeongyu/oh-my-opencode/issues/1200

---
## Summary by cubic
/start-work no longer overwrites the session agent, and subagents inherit the parent session’s model (including variant). This keeps model selection consistent and avoids defaulting to unavailable hardcoded models.

- **Bug Fixes**
  - Only set the session agent to atlas if no agent is set; removed the default agent from the command config.
  - For subagent_type, inherit the parent model (categoryModel = parentModel) and mark modelInfo as "inherited" for display.
  - Preserve model.variant in StoredMessage and propagate it to subagents and background tasks.

<sup>Written for commit a544e80fc44563e13e065cf1b656f06cd2dc472b. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/builtin-commands/commands.ts
+++ b/src/features/builtin-commands/commands.ts
@@ -55,7 +55,6 @@ ${REFACTOR_TEMPLATE}
   },
   "start-work": {
     description: "(builtin) Start Sisyphus work session from Prometheus plan",
-    agent: "atlas",
     template: `<command-instruction>
 ${START_WORK_TEMPLATE}
 </command-instruction>
--- a/src/features/hook-message-injector/injector.ts
+++ b/src/features/hook-message-injector/injector.ts
@@ -5,7 +5,7 @@ import type { MessageMeta, OriginalMessageContext, TextPart, ToolPermission } fr
 
 export interface StoredMessage {
   agent?: string
-  model?: { providerID?: string; modelID?: string }
+  model?: { providerID?: string; modelID?: string; variant?: string }
   tools?: Record<string, ToolPermission>
 }
 
--- a/src/features/hook-message-injector/injector.ts
+++ b/src/features/hook-message-injector/injector.ts
@@ -141,9 +141,17 @@ export function injectHookMessage(
   const resolvedAgent = originalMessage.agent ?? fallback?.agent ?? "general"
   const resolvedModel =
     originalMessage.model?.providerID && originalMessage.model?.modelID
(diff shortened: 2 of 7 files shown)


code-yeongyu (commented):
Thanks for the fix! This addresses the core issue in #1200.

**One concern from cubic review:**
The variant propagation might be incomplete - `parentModel` inheritance may drop the variant, so subagents could run with a different configuration than expected.

Could you verify:
1. When inheriting from parent, is `model.variant` (e.g., `high`, `low`) also passed to the subagent?
2. If parent is using `anthropic/claude-opus-4-5` with `variant: high`, does the subagent also get `variant: high`?

If variant isn't being propagated, we'd need to include it in the inheritance logic:
```typescript
categoryModel = parentModel  // includes model string
// Also propagate variant if present
if (parentVariant) {
  modelInfo = { type: 'inherited', variant: parentVariant }
}
```

Please confirm or address this before merge.

mmlmt2604:
> Thanks for the fix! This addresses the core issue in #1200.
> 
> **One concern from cubic review:** The variant propagation might be incomplete - `parentModel` inheritance may drop the variant, so subagents could run with a different configuration than expected.
> 
> Could you verify:
> 
> 1. When inheriting from parent, is `model.variant` (e.g., `high`, `low`) also passed to the subagent?
> 2. If parent is using `anthropic/claude-opus-4-5` with `variant: high`, does the subagent also get `variant: high`?
> 
> If variant isn't being propagated, we'd need to include it in the inheritance logic:
> 
> ```ts
> categoryModel = parentModel  // includes model string
> // Also propagate variant if present
> if (parentVariant) {
>   modelInfo = { type: 'inherited', variant: parentVariant }
> }
> ```
> 
> Please confirm or address this before merge.

Updated and variant is now being propagated, thanks!

code-yeongyu:
Thanks for the fix and for addressing the variant propagation feedback! Merging.

PR #1227, opened 2026-01-28
fix(model-resolver): use connected providers cache when model cache is empty

potb:
## Summary

- Fixes delegate_task with category (e.g., `visual-engineering`) always using first provider (google) instead of connected provider (github-copilot)
- When model cache is empty but connected providers cache exists, use it to find valid providers from fallback chain
- Remove `resolved.model` from userModel fallback in tools.ts (was bypassing the fallback chain entirely)

## Root Cause

The upstream fix (30ed086) passed `resolved.model` as `userModel`, which bypassed the fallback chain. This caused categories to always use the hardcoded default provider (e.g., `google/gemini-3-pro`) even when user only has `github-copilot` connected.

## Changes

1. **tools.ts**: Remove `resolved.model` from userModel fallback chain
2. **model-resolver.ts**: When `availableModels.size === 0`, check connected providers cache and use it to find valid providers

---
## Summary by cubic
Fixes provider selection when the model cache is empty. The resolver now uses the connected providers cache so category tasks pick the correct provider (e.g., GitHub Copilot instead of Google).

- **Bug Fixes**
  - Use connected providers cache to walk the fallback chain when availableModels is empty; choose the first connected provider, else use system default.
  - Remove resolved.model from userModel in delegate_task tools to avoid bypassing the fallback chain.

<sup>Written for commit 49bb7a80638fb2899d2399bd1b4bf163c856a563. Summary will update on new commits.</sup>

### DIFF
--- a/src/shared/model-resolver.test.ts
+++ b/src/shared/model-resolver.test.ts
@@ -356,10 +356,8 @@ describe("resolveModelWithFallback", () => {
       cacheSpy.mockRestore()
     })
 
-    test("skips fallback chain when availableModels empty even if connected providers cache exists", () => {
+    test("uses connected provider from fallback when availableModels empty but cache exists", () => {
       // #given - model cache missing but connected-providers cache exists
-      // This scenario caused bugs: provider is connected but may not have the model available
-      // Fix: When we can't verify model availability, skip fallback chain entirely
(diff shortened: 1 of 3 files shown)


code-yeongyu:
Thanks for the fix! Clean solution for the model resolver cache issue. Merging.

potb:
Thanks for your work !

PR #1203, opened 2026-01-28
feat(config): support fallback_models for agents/categories

youming-ai:
## What
- Add `fallback_models` (string or string[]) to agent/category config (schema + docs).
- Extend model resolution to try configured `fallback_models` when the preferred model is not available in local model cache / connected providers.
- Wire the resolved model into builtin agent creation, `delegate_task` category selection, and Prometheus (planner) model selection.
- Add tests covering fallback selection and category inheritance.

## Why
Provider connectivity/model availability often differs across machines. This makes per-agent/category model selection deterministic without requiring manual config edits when a preferred model is unavailable.

## Semantics
This PR implements **availability-time selection** (based on local cache + connected providers). It does **not** implement runtime retry/switching on API errors mid-request.

## Testing
- `bun run typecheck`
- `bun test src/shared/model-resolver.test.ts src/agents/utils.test.ts src/tools/delegate-task/tools.test.ts`

## Notes
- `runtime_fallback` is included as schema/config surface for future work; no runtime retry behavior is introduced here.
- Full `bun test` can be order/environment dependent locally; the focused suites above cover the touched logic.

PR #1237, opened 2026-01-29
feat(config): add fallback_models for model availability fallback

youming-ai:
## What
Adds a `fallback_models` config option (`string | string[]`) so agents/categories can select the first available model at initialization time when the preferred model is unavailable.

## Why
Provider connectivity and model availability vary across environments. This reduces brittle startup behavior when a single configured model cannot be used.

## Changes
- Add `fallback_models` to agent overrides and category config schema.
- Update model resolution to try `fallback_models` (via `availableModels` or connected-providers cache) before built-in fallback chain and system default.
- Fix Prometheus config merge so `resolvedModel` is not overwritten by `override.model`.
- Apply category defaults (temperature/top_p/maxTokens/thinking/reasoningEffort/textVerbosity) when using `override.category`.
- Add/adjust tests for resolver + inheritance; avoid relying on a real connected-providers cache.
- Update docs (config + RFC).

## Notes
- This is initialization-time selection (not runtime switching on 429/503).
- Supersedes #1203.

## Verification
- `bun run typecheck`
- `bun test src/shared/model-resolver.test.ts src/agents/utils.test.ts`


youming-ai:
Closing this PR in favor of a new clean PR. The fallback_models feature was already merged to dev, so this PR has been rebased to include only the unique runtime-fallback feature. See new PR for clean history.

PR #1382, opened 2026-02-02
fix(session): inject directory into session API calls for worktree support

gburch:
## Problem

When using OpenCode with git worktrees (or any workspace where `process.cwd()` differs from the project directory), session API calls fail with `NotFoundError`:

```
NotFoundError: Resource not found: .../storage/session/global/ses_xxx.json
```

The session file exists under the correct project directory, but the server looks in `global/` because it doesn't receive the directory context.

## Root Cause

OpenCode's session APIs accept an optional `query: { directory }` parameter to resolve which project's session storage to use. Without it, the server falls back to `process.cwd()`, which resolves to project ID `global` instead of the actual worktree's project ID.

The plugin has access to the correct directory via `ctx.directory`, but this wasn't being passed to session API calls.

## Solution

Wrap `ctx.client.session` with a Proxy at plugin initialization that automatically injects `query: { directory: ctx.directory }` into every session method call. This ensures all session operations (`.get()`, `.messages()`, `.prompt()`, etc.) are correctly scoped to the workspace's project.

## Changes

- Added `withDirectoryArgs()` helper to merge directory into query parameters
- Added `wrapSession()` to create a Proxy that intercepts session method calls
- Added `injectDirectoryClient()` to apply the wrapper at startup
- Called `injectDirectoryClient(ctx.client, ctx.directory)` during plugin initialization

## Testing

Verified that `delegate_task` and other session-dependent tools now work correctly when running OpenCode from a git worktree.

---
## Summary by cubic
Fixes session API failures in git worktrees by binding the workspace directory to the client’s session methods. Ensures session reads/writes target the correct project and removes NotFoundError.

- **Bug Fixes**
  - Creates a directory-bound client with createDirectoryBoundClient without mutating the original client.
  - Wraps session with an allowlist via wrapSessionWithDirectory and merges query.directory using withDirectoryArgs.
  - Applies the bound client across plugin init to restore session.get/messages/prompt in worktrees.
  - Adds tests for merge/passthrough behavior and a test tsconfig.

<sup>Written for commit c717c400a94312a17470ff96e42daf3b355cbf96. Summary will update on new commits.</sup>


gburch:

gburch:
I don't know why this is being ignored, it causes desktop and web to not work with worktrees which is newer functionality that is pretty core. Anyone using them will have to dig DEEP to resolve this issue and the entire system breaks if they don't. Stop adding features and merge this so this add-on works with worktrees. How are you working without them? @code-yeongyu

code-yeongyu (changes requested):
## PR #1382 Code Review: Worktree Support Fix

Thank you for identifying and fixing this git worktree issue, @gburch! This is a real pain point that affects users working with worktrees. However, I have several concerns about the implementation approach that I'd like to discuss before merging.

---

### 🔴 Critical Issues

#### 1. **Proxy Blindly Injects Directory to ALL Session Methods**

The current implementation wraps ALL session methods with directory injection:

```typescript
function wrapSession<T extends object>(session: T, directory: string): T {
  const handler: ProxyHandler<T> = {
    get(target, prop) {
      const value = target[prop as keyof T];
      if (typeof value !== "function") return value;
      return (args?: DirectoryQuery) => {
        const next = withDirectoryArgs(args, directory);  // ← Always injects
        const fn = value as (input: DirectoryQuery) => unknown;
        return fn.call(target, next);
      };
    },
  };
  return new Proxy(session, handler);
}
```

**Problem**: Not all session methods accept or need `query: { directory }`:

| Method | Uses Directory? | Notes |
|--------|----------------|-------|
| `session.get()` | ✅ Yes | Needs directory |
| `session.messages()` | ✅ Yes | Needs directory |
| `session.prompt()` | ✅ Yes | Needs directory |
| `session.summarize()` | ✅ Yes | Needs directory |
| `session.create()` | ❓ Unclear | Creates NEW session - directory in body? |
| `session.abort()` | ❌ Likely No | Only needs session ID to abort |
| `session.status()` | ❌ No | Global status, called with `{ path: undefined }` |
| `session.todo()` | ❓ Unclear | Currently no directory in codebase |
| `session.children()` | ❓ Unclear | Needs investigation |

**Risk**: Injecting `query: { directory }` to methods that don't expect it may cause errors or undefined behavior with future OpenCode SDK versions.

**Recommendation**: Create an allowlist of methods that actually need directory injection:

```typescript
const METHODS_NEEDING_DIRECTORY = new Set(['get', 'messages', 'prompt', 'summarize']);

function wrapSession<T extends object>(session: T, directory: string): T {
  return new Proxy(session, {
    get(target, prop) {
      const value = target[prop as keyof T];
      if (typeof value !== "function") return value;
      if (!METHODS_NEEDING_DIRECTORY.has(String(prop))) return value;
      return (args?: DirectoryQuery) => {
        const next = withDirectoryArgs(args, directory);
        return (value as Function).call(target, next);
      };
    },
  });
}
```

---

#### 2. **Directly Mutates Shared `ctx.client` Object**

```typescript
function injectDirectoryClient<T extends { session?: object }>(
  client: T,
  directory: string,
): T {
  if (!client.session) return client;
  const wrapped = wrapSession(client.session, directory);
  client.session = wrapped as T["session"];  // ← Direct mutation!
  return client;
}
```

**Problem**: `ctx.client` is a shared object provided by OpenCode to all plugins. Mutating it directly:
- May conflict with other plugins
- Makes behavior unpredictable
- Violates immutability principles

**Recommendation**: Store the wrapped session separately or use a different injection pattern (similar to how `injectServerAuthIntoClient` works with `_client.setConfig()`).

---

#### 3. **No Test Coverage**

This PR adds 254 lines of new code (90 lines of actual logic + formatting) with **zero test coverage**.

**Minimum tests needed**:
1. `withDirectoryArgs()` - test merging behavior, edge cases
2. `wrapSession()` - test proxy interception works correctly
3. `injectDirectoryClient()` - test client mutation
4. Integration test with mock session API

---

### 🟡 Moderate Concerns

#### 4. **Formatting Changes Mixed with Logic Changes**

The diff shows +254/-164 lines, but most changes are just Prettier/formatting:
- Trailing commas added
- Line breaks changed
- Indentation adjusted

This makes it very hard to review the actual semantic changes. Consider:
- Separating formatting changes into a separate commit
- Or applying project formatter before making logic changes

#### 5. **Edge Case: Conflicting Directory Values**

```typescript
function withDirectoryArgs(args: DirectoryQuery | undefined, directory: string) {
  // ...
  if (query && "directory" in query) return args;  // ← Preserves existing
  return { ...args, query: { ...query, directory } };
}
```

**Question**: What if the existing `query.directory` is DIFFERENT from `ctx.directory`? Should we:
- Keep the existing value (current behavior)?
- Override with `ctx.directory`?
- Log a warning?

This edge case needs explicit documentation.

---

### 🟢 Positive Aspects

1. **Real Problem Identified**: Worktree support is genuinely broken - users in worktrees get `NotFoundError` for sessions.

2. **Root Cause Understood**: Correctly identified that `process.cwd()` resolves to `global` instead of the worktree's project ID.

3. **Centralized Solution**: Rather than adding `query: { directory }` in 22+ places (which already exists in the codebase!), a central injection point is cleaner.

---

### 📋 Suggested Actions

1. **Create allowlist** for methods that need directory injection
2. **Add unit tests** for the new utility functions
3. **Consider immutable approach** instead of mutating `ctx.client`
4. **Verify against OpenCode SDK** - check which methods actually accept `query.directory`
5. **Separate formatting commit** from logic changes for cleaner review

---

### Alternative Approaches to Consider

**Option A: Typed Helper Functions**
```typescript
export const sessionWithDir = {
  get: (client, id, dir) => client.session.get({ path: { id }, query: { directory: dir } }),
  messages: (client, id, dir) => client.session.messages({ path: { id }, query: { directory: dir } }),
  // ... explicit, type-safe
};
```

**Option B: Higher-Order Wrapper (Immutable)**
```typescript
function createDirectoryBoundSession(session, directory) {
  return {
    get: (args) => session.get({ ...args, query: { ...args.query, directory } }),
    messages: (args) => session.messages({ ...args, query: { ...args.query, directory } }),
    // Explicitly list only methods that need it
    abort: session.abort.bind(session),  // Passthrough - no directory needed
    // ...
  };
}
```

---

I appreciate the effort to fix this real issue! Let's work together to make the implementation robust. Happy to discuss any of these points.

code-yeongyu:
## Additional Code-Specific Concerns

### 1. `DirectoryQuery` Type (line 97)
```typescript
type DirectoryQuery = { query?: Record<string, unknown> };
```
This type is very loose. Consider using more specific types that match actual session API signatures.

### 2. `withDirectoryArgs` Query Handling (lines 100-106)
```typescript
const query = typeof args.query === "object" && args.query ? args.query : undefined;
```
If `args.query` is not an object (e.g., malformed input), it silently gets overwritten. Consider logging for debugging.

### 3. `wrapSession` Proxy - Critical (lines 108-121)
```typescript
return (args?: DirectoryQuery) => {
  const next = withDirectoryArgs(args, directory);
  // ...
};
```
**This intercepts ALL methods**, including those that do not need/accept directory:
- `session.abort()` - only needs session ID
- `session.status()` - global status, no query
- `session.create()` - different param structure

**Recommendation**: Use an allowlist:
```typescript
const NEEDS_DIRECTORY = new Set(["get", "messages", "prompt", "summarize"]);
if (!NEEDS_DIRECTORY.has(String(prop))) return value; // passthrough
```

### 4. Direct Client Mutation (line 127)
```typescript
client.session = wrapped as T["session"];
```
Mutates shared `ctx.client` object. Consider immutable patterns.

### 5. Missing Tests
These utility functions need unit tests covering:
- `withDirectoryArgs` - edge cases, existing directory
- `wrapSession` - proxy behavior, passthrough
- `injectDirectoryClient` - integration

gburch:
Thanks for the detailed review. I agree with the core concerns and I’m going to revise this implementation before merge.

### What I agree with and will change

1. **Shared `ctx.client` mutation**
   - Agreed. Mutating `ctx.client.session` directly is risky in plugin environments.
   - I’ll switch to a **non-mutating approach** (create a directory-bound session/client wrapper used by this plugin only, without modifying the shared object).

2. **Missing tests**
   - Agreed. I’ll add unit coverage for:
     - `withDirectoryArgs` merge behavior + edge cases
     - `wrapSession` interception + passthrough behavior
     - directory conflict behavior (`query.directory` already present)
     - integration-style test with mock `session` methods

3. **Formatting mixed with logic**
   - Agreed. I’ll keep the semantic diff clean in the follow-up (either isolate formatting or avoid unrelated formatting changes).

### Clarification on “all session methods”

I verified current OpenCode SDK/server contracts, and `query.directory` is supported broadly across session endpoints (including `status`, `create`, `abort`, `todo`, `children`, `get`, `messages`, `prompt`, `summarize`, etc.), so injecting directory is not invalid for current APIs.

That said, I still agree explicitness is better for maintainability. I’ll change this to an **explicit method allowlist** for clarity and future safety.

### Directory conflict semantics

Current behavior preserves caller-provided `query.directory`. I’ll keep that behavior and document/test it explicitly so precedence is unambiguous.

---

I’ll push an update reflecting the above so we can re-review with a tighter, safer patch.

code-yeongyu:
[sisyphus-bot]

## PR Status Update — Changes Requested

Hi @gburch,

Thank you for this valuable contribution fixing worktree support! This is a real pain point for many users.

**Current Status:** Changes requested by maintainer (@code-yeongyu) on Feb 5, 2026.

### Summary of Required Changes

You already acknowledged these in your Feb 5 comment — just a friendly reminder:

1. **Non-mutating approach** — Create a directory-bound client wrapper instead of mutating shared 
2. **Method allowlist** — Only inject directory for methods that explicitly need it (not ALL session methods)
3. **Unit tests** — Add coverage for , , and integration scenarios
4. **Formatting separation** — Keep semantic changes separate from Prettier formatting

### What I Found

✅ **CI Status:** All checks passing (CLA, GitGuardian, cubic AI review)
✅ **Code Quality:** Logic is sound and addresses the real issue
⚠️ **Last Activity:** No code updates since your acknowledgment on Feb 5

### Next Steps

The maintainer is waiting for your revision. Once you push the updates addressing the 4 points above, this should be ready for re-review.

**Pro tip:** When you push, consider adding a comment explaining what changed — makes re-review faster.

---

*This PR is important for worktree users — let's get it merged! Let me know if you need any clarification on the requested changes.*

code-yeongyu:
[sisyphus-bot] Hi gburch. 🙏 Thanks for the worktree-aware session directory injection investigation.

Picking this back up from the 5/16 triage. This PR carries an existing `CHANGES_REQUESTED` review and shows CONFLICTING against current `dev`. The session API surface has moved substantially since 2/2 (close to 4 months ago); the original review's concrete asks may not even apply to the new shape anymore.

If you're still actively working on worktree support, the most productive path is probably to close this PR and open a fresh one against current `dev` that addresses both (a) the original review feedback and (b) the new file layout. Threading worktree directory into the session API is a real product need, but a 4-month-old diff against an outdated review state is hard for any reviewer to engage with.

Let me know what you'd prefer (rebase + address the original feedback, or close + fresh PR), and I'll engage on whichever path. Really appreciate the original work; worktree-aware sessions is the right direction.

PR #1393, opened 2026-02-02
fix: grep and glob tools usage without path param under Opencode Desktop

ualtinok:
Fixes grep and glob tools usage without path param under Opencode Desktop

## Summary

Grep and Glob now uses ctx.directory when no path data is given. In OpenCode Desktop, the cwd is "/" when run via Finder or Dock, this causes the tools to run under "/" instead of the project directory. This small change fixes the problem as opencode pushed changes so that ctx receives the correct directory now (via Instance.directory) 

## Related Issues

https://github.com/code-yeongyu/oh-my-opencode/issues/658
https://github.com/code-yeongyu/oh-my-opencode/issues/588

-- also probably from opencode 
https://github.com/anomalyco/opencode/issues/7143

---
## Summary by cubic
Fixes grep and glob running under “/” in OpenCode Desktop by defaulting the search path to ctx.directory when no path is provided. Searches now run in the project directory even when Desktop launches with cwd set to “/”.

<sup>Written for commit 527c21ea90e348f415570112c85cc77fdcb2a4bd. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/glob/tools.ts
+++ b/src/tools/glob/tools.ts
@@ -20,10 +20,12 @@ export const glob: ToolDefinition = tool({
           "simply omit it for the default behavior. Must be a valid directory path if provided."
       ),
   },
-  execute: async (args) => {
+  execute: async (args, ctx) => {
     try {
       const cli = await resolveGrepCliWithAutoInstall()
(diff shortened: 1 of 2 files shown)


ualtinok:

PR #1520, opened 2026-02-05
fix: preserve tools config for MCP tool restriction

johnnymo87:
## Summary

- Preserve `tools` field in `migrateAgentConfig()` instead of deleting it after conversion to `permission`
- This enables user-configured tool restrictions for MCP tools (e.g., `slack_*`)

## Problem

When users configure tool restrictions in `oh-my-opencode.json`:

```json
{
  "agents": {
    "sisyphus": {
      "tools": { "slack_*": false }
    }
  }
}
```

The `migrateAgentConfig()` function was:
1. Converting `tools: { "slack_*": false }` to `permission: { "slack_*": "deny" }`
2. **Deleting** the original `tools` field

However, OpenCode's `permission` system only works for `edit`, `bash`, and `webfetch` tools. For MCP tools like `slack_*`, OpenCode requires the native `tools` boolean format in the agent config.

## Solution

Keep `tools` in the agent config so OpenCode can use it for native tool restriction, while also generating `permission` for tools that support ask/allow/deny semantics.

## Testing

- [x] Updated tests in `permission-compat.test.ts` and `utils.test.ts`
- [x] Tested locally with slack MCP - sisyphus no longer has access to slack tools when configured

## Checklist

---
## Summary by cubic
Preserves agent.tools during migrateAgentConfig so MCP tool restrictions (e.g., slack_*) continue to work, while still generating permission for supported tools. Restores user-configured blocks that were lost after migration.

- **Bug Fixes**
  - Do not delete config.tools after migration; OpenCode uses it for MCP tool restriction.
  - Merge migrated permission with existing entries; permission still applies to edit/bash/webfetch.
  - Updated tests to assert tools are preserved and permissions are correctly merged.

<sup>Written for commit 122d35550164a29ae4291bd11de9cc4ffafd4560. Summary will update on new commits.</sup>

### DIFF
--- a/src/agents/utils.test.ts
+++ b/src/agents/utils.test.ts
@@ -788,7 +788,7 @@ describe("agent override tools migration", () => {
     expect(permission["jetbrains_get_*"]).toBe("allow")
   })
 
-  test("tools config is removed after migration", async () => {
+  test("tools config is preserved for OpenCode native tool restriction", async () => {
     // #given
     const overrides = {
       explore: { tools: { "some_tool": false } } as any,
@@ -797,9 +797,9 @@ describe("agent override tools migration", () => {
     // #when
     const agents = await createBuiltinAgents([], overrides, undefined, TEST_DEFAULT_MODEL)
 
-    // #then
+    // #then - tools should be preserved (not deleted) for OpenCode to use
     expect(agents.explore).toBeDefined()
-    expect((agents.explore as any).tools).toBeUndefined()
+    expect((agents.explore as any).tools).toEqual({ "some_tool": false })
   })
 })
 
--- a/src/shared/permission-compat.test.ts
+++ b/src/shared/permission-compat.test.ts
@@ -71,7 +71,7 @@ describe("permission-compat", () => {
   })
 
   describe("migrateAgentConfig", () => {
-    test("migrates tools to permission", () => {
+    test("migrates tools to permission and preserves tools", () => {
       // given config with tools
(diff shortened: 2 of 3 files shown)


johnnymo87:
Closing this PR - we found a simpler solution.

Instead of per-agent tool filtering with `agent.tools: { "slack_*": false }`, we can just disable the MCP server entirely:

```json
{
  "mcp": {
    "slack": {
      "enabled": false,
      ...
    }
  }
}
```

This prevents the Slack MCP from starting at all, which is simpler and achieves the same goal (keeping slack tools out of the main agent's toolset).

If someone later needs per-agent MCP tool filtering (where one agent can use slack and another can't), this PR's approach would be valid - but for now the simpler solution works.

johnnymo87:
**Correction to PR description**: The claim that "OpenCode's permission system only works for edit, bash, and webfetch tools" is inaccurate.

The permission system **does** work for MCP tools. The confusion arose because `opencode debug agent <name>` only showed `ToolRegistry.tools()` and not `MCP.tools()`, making it appear that MCP tools weren't being filtered.

After adding MCP tools to the debug output, we confirmed that `agent.tools: { "slack_*": false }` correctly results in:
- `sisyphus`: `slack_*` tools = `false` (filtered out)
- `slack` agent: `slack_*` tools = `true` (available)

So the per-agent filtering approach in this PR would have worked - we just chose the simpler `enabled: false` solution instead.

PR #1546, opened 2026-02-06
fix: escape regex special chars in pattern matcher

### LINKED ISSUE #1521: SyntaxError: Invalid regular expression: unmatched parentheses on cat -n
## Description

When oh-my-opencode plugin is loaded, executing `cat -n` (or any file read command) in opencode results in:

```
SyntaxError: Invalid regular expression: unmatched parentheses
```

This happens regardless of file path (tested with both Korean and English-only filenames).

## Steps to Reproduce

1. Install opencode 1.1.50 via Homebrew
2. Add `oh-my-opencode@3.2.3` to `opencode.json` plugins
3. Run `opencode` and ask it to read any file
4. Agent attempts `cat -n <filepath>` → SyntaxError

## Expected Behavior

`cat -n` should execute normally and return file contents.

## Actual Behavior

```
$ cat -n "path/to/file.md"
SyntaxError: Invalid regular expression: unmatched parentheses
```

The error persists even with **all hooks disabled** via `disabled_hooks` in `oh-my-opencode.json`.

## Isolation Test

- opencode **without** oh-my-opencode → works fine
- opencode **with** oh-my-opencode (all hooks disabled) → error
- opencode **with** only `opencode-antigravity-auth@1.4.3` → works fine

This confirms the issue is in oh-my-opencode's core plugin code, not in any specific hook.

## Environment

- **OS**: macOS 26.2 (arm64)
- **Node**: v25.6.0
- **opencode**: 1.1.50 (Homebrew)
- **oh-my-opencode**: 3.2.3 (also tested with 2.14.0 — same issue)
- **Other plugins**: opencode-antigravity-auth@1.4.3

## oh-my-opencode.json

```json
{
  "disabled_hooks": ["<all 25 hooks disabled>"],
  "agents": { ... },
  "categories": { ... }
}
```

kaizen403:
## Summary

Fixes #1521. When hook matcher patterns in `.claude/settings.json` contained regex special characters like parentheses, the `matchesToolMatcher` function threw `SyntaxError: Invalid regular expression: unmatched parentheses`. This happened because these characters were passed directly to `new RegExp()` without escaping.

## Root Cause

In `src/shared/pattern-matcher.ts`, the `matchesToolMatcher` function converts glob-style patterns (using `*` as wildcard) to regex patterns. However, only the asterisk (`*`) was being converted to `.*` while other regex special characters (`.`, `+`, `?`, `^`, `$`, `{`, `}`, `(`, `)`, `|`, `[`, `]`, `\`) were passed through unescaped, causing `SyntaxError` when the regex was constructed.

**Before (broken):**
```typescript
const regex = new RegExp(`^${p.replace(/\*/g, ".*")}$`, "i")
// Pattern "bash(*)" becomes regex /^bash(.*)$/ - OK
// Pattern "bash(" becomes regex /^bash($/ - THROWS SyntaxError!
```

**After (fixed):**
```typescript
const escaped = escapeRegexExceptAsterisk(p) // Escapes all special chars except *
const regex = new RegExp(`^${escaped.replace(/\*/g, ".*")}$`, "i")
// Pattern "bash(*)" becomes regex /^bash\(.*\)$/ - matches literal parentheses with glob
// Pattern "bash(" becomes regex /^bash\($/ - no longer throws
```

## Changes

1. **`src/shared/pattern-matcher.ts`**: Added `escapeRegexExceptAsterisk()` helper that escapes all regex special characters except `*`, which is intentionally preserved for glob-to-regex conversion.

2. **`src/shared/pattern-matcher.test.ts`**: Added comprehensive test suite with 30 tests covering:
   - Exact matching (case-insensitive)
   - Wildcard matching (glob-style `*`)
   - Pipe-separated patterns (`bash | edit | write`)
   - All regex special characters: `(`, `)`, `[`, `]`, `+`, `?`, `^`, `$`, `{`, `}`, `\`, `.`
   - Edge cases (empty matcher, complex patterns)

## Testing

```bash
bun test src/shared/pattern-matcher.test.ts
# 30 pass, 0 fail

bun run typecheck
# No errors

bun test  
# 2194 pass, 10 fail (all failures are pre-existing flaky tests unrelated to this change)
```

## Why This Happened

Users with hooks configured in `.claude/settings.json` that used matchers containing special characters (even innocuous patterns like `bash` that happen to follow file read operations with parentheses in filenames) would trigger this error. The error persisted even with all hooks disabled via `disabled_hooks` because the pattern matching happens during hook configuration loading, not execution.

## Verification

After this fix, any matcher pattern containing regex special characters will be treated as literal characters (except `*` which remains a glob wildcard), preventing the `SyntaxError` and allowing the pattern to match as intended.

---
## Summary by cubic
Escapes regex special characters in hook matcher patterns so patterns with parentheses, brackets, etc., no longer crash and are matched literally, while * still works as a glob. Fixes #1521 and prevents SyntaxError during hook loading from .claude/settings.json.

- **Bug Fixes**
  - Escape . + ? ^ $ { } ( ) | [ ] \ before converting * to .* in matchesToolMatcher.
  - Added tests for exact, wildcard, pipe-separated, and special-char patterns, including edge cases.

<sup>Written for commit bc782ca4d4f408953e6704da17e59cbcae3b22e5. Summary will update on new commits.</sup>

### DIFF
--- a/src/shared/pattern-matcher.test.ts
+++ b/src/shared/pattern-matcher.test.ts
@@ -1,0 +1,177 @@
+import { describe, test, expect } from "bun:test"
+import { matchesToolMatcher, findMatchingHooks } from "./pattern-matcher"
+import type { ClaudeHooksConfig } from "../hooks/claude-code-hooks/types"
+
+describe("matchesToolMatcher", () => {
+  describe("exact matching", () => {
+    //#given a pattern without wildcards
+    //#when matching against a tool name
+    //#then it should match case-insensitively
+
+    test("matches exact tool name", () => {
+      expect(matchesToolMatcher("bash", "bash")).toBe(true)
+    })
+
+    test("matches case-insensitively", () => {
+      expect(matchesToolMatcher("Bash", "bash")).toBe(true)
+      expect(matchesToolMatcher("bash", "BASH")).toBe(true)
+    })
+
+    test("does not match different tool names", () => {
+      expect(matchesToolMatcher("bash", "edit")).toBe(false)
+    })
+  })
+
+  describe("wildcard matching", () => {
+    //#given a pattern with asterisk wildcard
+    //#when matching against tool names
+    //#then it should treat * as glob-style wildcard
+
+    test("matches prefix wildcard", () => {
+      expect(matchesToolMatcher("lsp_goto_definition", "lsp_*")).toBe(true)
+      expect(matchesToolMatcher("lsp_find_references", "lsp_*")).toBe(true)
+    })
+
+    test("matches suffix wildcard", () => {
+      expect(matchesToolMatcher("file_read", "*_read")).toBe(true)
+    })
+
+    test("matches middle wildcard", () => {
+      expect(matchesToolMatcher("get_user_info", "get_*_info")).toBe(true)
+    })
+
+    test("matches multiple wildcards", () => {
+      expect(matchesToolMatcher("get_user_data", "*_user_*")).toBe(true)
+    })
+
+    test("single asterisk matches any tool", () => {
+      expect(matchesToolMatcher("anything", "*")).toBe(true)
+    })
+  })
+
+  describe("pipe-separated patterns", () => {
+    //#given multiple patterns separated by pipes
+    //#when matching against tool names
+    //#then it should match if any pattern matches
+
+    test("matches first pattern", () => {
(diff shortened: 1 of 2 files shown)


kaizen403:
hello again @code-yeongyu , this fixes the regex error when using cat -n with the plugin loaded. the issue was special chars like parentheses weren't being escaped in the pattern matcher. added tests too. lmk if you want any changes :3

PR #1631, opened 2026-02-07
fix: preserve user's enabled:false when MCP servers from .mcp.json files overwrite opencode config

moha-abdi:
## Problem

When a user disables an MCP server by setting `enabled: false` in their `opencode.json`, the setting is silently ignored if the same MCP server is also defined in any `.mcp.json` config file (e.g., `~/.claude.json`, `~/.claude/.mcp.json`, `./.mcp.json`, `./.claude/.mcp.json`).

**Root cause:** Two things combine to produce this bug:

1. `transformMcpServer()` in `src/features/claude-code-mcp-loader/transformer.ts` **hardcodes `enabled: true`** on every server it processes
2. In `mcp-config-handler.ts`, the MCP merge ordering places `.mcp.json` results **after** the user's opencode config:
   ```ts
   params.config.mcp = {
     ...createBuiltinMcps(...),
     ...(params.config.mcp),      // user's opencode.json (enabled: false)
     ...mcpResult.servers,         // from .mcp.json files (hardcoded enabled: true) <- overwrites
     ...params.pluginComponents.mcpServers,
   };
   ```

Same-named keys from `.mcp.json` sources replace the user's `enabled: false` with `enabled: true`.

**This bug has existed since the MCP loader was introduced**, but it became visible to me after #1616 added `~/.claude.json` as an MCP source -- my Claude Code MCP servers are defined there, and that change caused them to load even though I had them disabled in `opencode.json`.

## Solution

Before the MCP merge in `mcp-config-handler.ts`, capture which MCPs the user explicitly set to `enabled: false`. After the merge, restore `enabled: false` on those MCPs. This is a minimal, non-breaking fix that respects user intent regardless of which `.mcp.json` source triggers the overwrite.

## Files Changed

- `src/plugin-handlers/mcp-config-handler.ts` -- Preserve user's `enabled: false` across the MCP merge
- `src/plugin-handlers/config-handler.test.ts` -- 2 new tests verifying the fix

## Testing

- All 31 config-handler tests pass (including 2 new ones)
- Typecheck clean
- Build clean
- Tested locally by building and pointing opencode.json to the local build -- confirmed disabled MCPs stay disabled

### DIFF
--- a/src/plugin-handlers/config-handler.test.ts
+++ b/src/plugin-handlers/config-handler.test.ts
@@ -1059,3 +1059,75 @@ describe("per-agent todowrite/todoread deny when task_system enabled", () => {
     expect(agentResult.sisyphus?.permission?.todoread).toBeUndefined()
   })
 })
+
+describe("MCP enabled:false preservation", () => {
+  test("preserves user's enabled:false when .mcp.json servers overwrite same MCP name", async () => {
+    //#given - user has an MCP set to enabled:false in opencode config
+    ;(mcpLoader.loadMcpConfigs as any).mockRestore?.()
+    spyOn(mcpLoader, "loadMcpConfigs" as any).mockResolvedValue({
+      servers: {
+        "my-mcp": { command: "npx", args: ["-y", "my-mcp"], enabled: true },
+      },
+    })
+    const pluginConfig: OhMyOpenCodeConfig = {}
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Closing: the MCP enabled:false issue was fixed more comprehensively in #1721. Thanks for identifying the issue!

Issue #1627, opened 2026-02-07
[Bug]: EXA_API_KEY environment variable is not applied to Exa MCP URL, causing rate limit errors


imbytecat:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

I have configured the `EXA_API_KEY` in my environment variables, but `oh-my-opencode` seems to ignore it when establishing the connection to the Exa MCP server.

As a result, I am receiving a "free MCP rate limit" error from Exa. According to the error message and the [Exa documentation](https://exa.ai/docs/reference/exa-mcp#api-key), the API key must be passed via the `exaApiKey` query parameter in the URL (e.g., `?exaApiKey=YOUR_EXA_API_KEY`), but the current implementation in `src/mcp/websearch.ts` does not seem to append this parameter automatically when the environment variable is present.

### Steps to Reproduce

1. Export a valid Exa API Key in the environment: `export EXA_API_KEY=your_key_here`.
2. Run `opencode`.
3. Perform a web search action that triggers the Exa MCP.
4. (After sufficient usage) Observe that the request fails with a rate limit error, indicating the key was not used.

### Expected Behavior

When `EXA_API_KEY` is present in the environment, the application should automatically append `?exaApiKey=${process.env.EXA_API_KEY}` to the Exa MCP server URL so that the request is authenticated and uses the user's quota.

### Actual Behavior

The application connects to the default Exa MCP URL without the API key, resulting in the following error:

```json
Error: Streamable HTTP error: Error POSTing to endpoint: {
  "jsonrpc": "2.0",
  "error": {
    "code": -32000,
    "message": "You've hit Exa's free MCP rate limit. To continue using without limits, create your own Exa API key.\n\nFix: Create API key at https://dashboard.exa.ai/api-keys , and then update Exa MCP URL to this https://mcp.exa.ai/mcp?exaApiKey=YOUR_EXA_API_KEY"
  },
  "id": null
}

```

### Doctor Output

```shell
$ bunx oh-my-opencode doctor

[...]

PR #1643, opened 2026-02-08
fix(mcp): append EXA_API_KEY to Exa MCP URL when env var is set (#1627)

code-yeongyu:
## Background

When using the Exa MCP server, the API key must be passed via a query parameter `?exaApiKey=YOUR_KEY` according to the Exa MCP documentation. However, the current code only passes the API key via the `x-api-key` header and does not append it to the URL query parameters.

This causes users to encounter "free MCP rate limit" errors because their requests are unauthenticated, even when they have set the `EXA_API_KEY` environment variable.

## Changes

Modified `src/mcp/websearch.ts` to append `&exaApiKey=${EXA_API_KEY}` to the Exa MCP URL when the `EXA_API_KEY` environment variable is set.

- The URL construction now checks for `process.env.EXA_API_KEY`
- When present, appends `&exaApiKey=${process.env.EXA_API_KEY}` to the URL
- The base URL remains unchanged when the environment variable is not set

## Testing

1. Set the `EXA_API_KEY` environment variable
2. Verify that the Exa MCP URL now includes the `exaApiKey` query parameter
3. Ensure authenticated requests work without hitting rate limits

## Review Notes

- The fix maintains backward compatibility - when `EXA_API_KEY` is not set, the URL remains unchanged
- The API key is passed both via the URL query parameter AND the `x-api-key` header for maximum compatibility

---
## Summary by cubic
Fixes authentication for Exa MCP by appending a URL-encoded exaApiKey query parameter when EXA_API_KEY is set and removing the x-api-key header to match server requirements. Tests now cover encoding edge cases, verify the default tools=web_search_exa in the URL, restore Tavily behavior (Authorization Bearer), and confirm nothing is added when no key is set.

<sup>Written for commit a3dd1dbaf96b1ab43282d4eb3f8f93b8766b1e64. Summary will update on new commits.</sup>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -28,13 +28,13 @@
         "typescript": "^5.7.3",
       },
       "optionalDependencies": {
-        "oh-my-opencode-darwin-arm64": "3.3.0",
-        "oh-my-opencode-darwin-x64": "3.3.0",
-        "oh-my-opencode-linux-arm64": "3.3.0",
-        "oh-my-opencode-linux-arm64-musl": "3.3.0",
-        "oh-my-opencode-linux-x64": "3.3.0",
-        "oh-my-opencode-linux-x64-musl": "3.3.0",
-        "oh-my-opencode-windows-x64": "3.3.0",
+        "oh-my-opencode-darwin-arm64": "3.3.1",
+        "oh-my-opencode-darwin-x64": "3.3.1",
+        "oh-my-opencode-linux-arm64": "3.3.1",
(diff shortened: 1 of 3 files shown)


kvokka:
See the timing of #1644 
LOL

code-yeongyu:
@cubic-dev-ai The `encodeURIComponent` fix has been pushed in commit a85da593. Please re-run the review.

PR #1709, opened 2026-02-10
feat(comment-checker): support apply_patch

code-yeongyu:
## What
- Run comment-checker for `apply_patch` by consuming OpenCode `apply_patch` tool result `metadata.files`.
- Use `Edit` mode (before/after) so only newly added comments are flagged.

## Where
- `src/hooks/comment-checker/hook.ts`
- `src/hooks/comment-checker/cli-runner.ts`

## Test
- `bun test src/hooks/comment-checker`

---
## Summary by cubic
Adds comment-checker support for apply_patch edits, flagging only newly added comments from before/after diffs via the CLI. Also fixes prompt constraints, event normalization with dedup, and model caching to unblock related flows.

- **New Features**
  - Comment-checker: consume apply_patch metadata.files, run in Edit mode (before/after), skip deletes, handle moves, and append findings to tool output.
  - Session status: normalize session.status with type=idle into synthetic session.idle events with 500ms dedup so hooks run reliably.

- **Bug Fixes**
  - Sisyphus-Junior prompts: block the delegation “task” tool while allowing task_create, task_update, task_list, task_get when task_system=true.
  - Atlas continuation: allow Sisyphus sessions to continue boulder flows when agent defaults to atlas and wasn’t explicitly set.
  - Event dedup: prune the synthetic idle dedup map to avoid stale entries and prevent duplicate idle dispatches.
  - Provider cache: extract models from provider.list().all, fixing empty model caches and enabling clearer category errors (“model not available” vs “unknown category”).
  - Tests: replace fs mock.module with spyOn to prevent cross-test pollution.

<sup>Written for commit f3f5b98c687e7c53f85f8e8c7a8a63b9d06806ca. Summary will update on new commits.</sup>

### DIFF
--- a/src/agents/sisyphus-junior/default.ts
+++ b/src/agents/sisyphus-junior/default.ts
@@ -12,6 +12,7 @@ export function buildDefaultSisyphusJuniorPrompt(
   promptAppend?: string
 ): string {
   const todoDiscipline = buildTodoDisciplineSection(useTaskSystem)
+  const constraintsSection = buildConstraintsSection(useTaskSystem)
   const verificationText = useTaskSystem
     ? "All tasks marked completed"
     : "All todos marked completed"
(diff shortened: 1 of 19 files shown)


code-yeongyu:
cubic found a potential unbounded growth risk in src/plugin/event.ts (recentSyntheticIdles). I fixed it by pruning entries older than the 500ms dedup window on every event. Added test: src/plugin/recent-synthetic-idles.test.ts. @cubic-dev-ai please re-run review.

PR #1721, opened 2026-02-10
fix(mcp): preserve user's enabled:false and apply disabled_mcps to all MCP sources

edxeth:
## Summary

- Fixes regression from `598a4389` where OMO force-enabled all MCPs, ignoring user's `enabled: false` in `opencode.json`
- Re-adds `disabledMcps` parameter to `loadMcpConfigs()` that was dropped during refactor
- Preserves user's `enabled: false` settings after MCP config merge
- Applies `disabled_mcps` filtering to all MCP sources (built-in, `.mcp.json`, and plugins)

## Root Cause

Commit `598a4389` ("refactor(core): split index.ts and config-handler.ts into focused modules") extracted MCP config handling into `mcp-config-handler.ts` but:

1. Called `loadMcpConfigs()` **without** the `disabledMcps` parameter (originally added in `fab0f462`)
2. Did not handle the spread-order overwrite where `.mcp.json` MCPs (hardcoded `enabled: true` in `transformer.ts`) overwrote user's `enabled: false` from `opencode.json`

## Changes

| File | Change |
|------|--------|
| `loader.ts` | Re-added `disabledMcps: string[] = []` parameter; MCPs in the list are skipped during loading |
| `loader.test.ts` | 3 new test cases for `disabledMcps` filtering |
| `mcp-config-handler.ts` | Captures user's `enabled: false` before merge, passes `disabled_mcps` to loader, restores `enabled: false` after merge, deletes `disabled_mcps` entries from final result |
| `mcp-config-handler.test.ts` | New file with 5 test cases |

## Verification

- Typecheck: clean (0 errors)
- All 66 MCP-related tests pass (0 failures)
- Full test suite: 2472 pass (70 pre-existing failures unrelated to changes)
- Manually verified: user confirmed fix works in production
- Oracle review: approved as safe and ship-ready

---
## Summary by cubic
Fixes a regression that force-enabled MCPs. We now respect enabled:false in user config and apply disabled_mcps across built-in, .mcp.json, and plugin MCPs.

- **Bug Fixes**
  - Preserve enabled:false from user config during merges with .mcp.json and plugins.
  - Re-add and pass disabled_mcps to loadMcpConfigs and built-ins; remove disabled MCPs from the final config.
  - Add 8 tests covering disabled list handling and enabled:false preservation.

<sup>Written for commit 3abc1d46baffad95211380be66ca416737957153. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/claude-code-mcp-loader/loader.test.ts
+++ b/src/features/claude-code-mcp-loader/loader.test.ts
@@ -229,5 +229,109 @@ describe("getSystemMcpServerNames", () => {
       } finally {
         process.chdir(originalCwd)
       }
-    })
+     })
 })
+
+describe("loadMcpConfigs", () => {
+  beforeEach(() => {
+    mkdirSync(TEST_DIR, { recursive: true })
+    mkdirSync(TEST_HOME, { recursive: true })
+    mock.module("os", () => ({
+      homedir: () => TEST_HOME,
+      tmpdir,
+    }))
+    mock.module("../../shared", () => ({
+      getClaudeConfigDir: () => join(TEST_HOME, ".claude"),
+    }))
+    mock.module("../../shared/logger", () => ({
+      log: () => {},
+    }))
+  })
+
+  afterEach(() => {
+    mock.restore()
+    rmSync(TEST_DIR, { recursive: true, force: true })
+  })
+
+  it("should skip MCPs in disabledMcps list", async () => {
+    //#given
+    const mcpConfig = {
+      mcpServers: {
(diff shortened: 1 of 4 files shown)


edxeth:
@code-yeongyu mind checking this PR? I think it should be prioritized over many others. Every instance of OC spins up every MCP server on startup and that is completely unintended behavior. Let me demonstrate the bug with this screen recording:

https://github.com/user-attachments/assets/18d5dc5c-6b75-48cd-aa82-fcfc9a537446

It's needless to say that all my MCP servers are already disabled in `opencode.json`, hence why this PR is much needed. No clue how no one is reporting this, it's terrific!

PR #1727, opened 2026-02-10
fix(hooks): guard against undefined output.output in tool.execute.after hooks

sjawhar:
## Summary

Guard against `undefined` `output.output` in **all 14** `tool.execute.after` hooks and injectors that access the field.

MCP tool results (e.g. Slack, Toggl) don't populate `output.output` as a string, causing:
- **TypeError crashes** in hooks that call `.toLowerCase()` or `.startsWith()` on `undefined`
- **Silent `"undefined"` string corruption** in hooks that use `+=` (producing `"undefined\n\nSome reminder"`)

## Root Cause

The `tool.execute.after` hook fires for **all** tool calls, including MCP tools. The hook type signature declares `output: { output: string }`, but MCP tool results don't always populate this field. Many hooks access `output.output` without null-checking first.

Two existing hooks already guarded against this:
- `empty-task-response-detector` uses `output.output?.trim() ?? ""`
- `tool-output-truncator` used `typeof output.output !== 'string'`

The rest did not, and the codebase was inconsistent.

## Fix

Add `output.output == null` early-return guard to every `tool.execute.after` handler that accesses `output.output`. This:
- Uses loose equality (`== null`) to catch both `null` and `undefined`
- Allows empty strings through (which are valid tool outputs)
- Is applied consistently across all hooks, including those behind tool filters (defense in depth)
- Normalizes the previous `typeof !== 'string'` check in `tool-output-truncator` to match

### Hooks fixed

| Hook | Failure mode | Had tool filter? |
|------|-------------|-----------------|
| `comment-checker` | **Crash**: `.toLowerCase()` on undefined | No — crashed before filter |
| `edit-error-recovery` | **Crash**: `.toLowerCase()` on undefined | Yes (`edit` only) |
| `task-resume-info` | **Crash**: `.startsWith()` on undefined | Yes (`task`/`call_omo_agent`) |
| `delegate-task-retry` | Passed undefined to function | Yes (`task` only) |
| `context-window-monitor` | **Corrupt**: `undefined +=` | No |
| `task-reminder` | **Corrupt**: `undefined +=` | No |
| `claude-code-hooks handler` | **Corrupt**: template literal | No |
| `category-skill-reminder` | **Corrupt**: `undefined +=` | Yes |
| `agent-usage-reminder` | **Corrupt**: `undefined +=` | Yes |
| `interactive-bash-session` | **Corrupt**: `undefined +=` | Yes (`interactive_bash`) |
| `directory-readme-injector` | **Corrupt**: `undefined +=` | Yes (`read`/`glob`/`grep`) |
| `directory-agents-injector` | **Corrupt**: `undefined +=` | Yes (`read`/`glob`/`grep`) |
| `rules-injector` | **Corrupt**: `undefined +=` | Yes (`read`/`glob`/`grep`) |
| `tool-output-truncator` | Already guarded (normalized) | Yes |

## Testing

- Built locally, loaded via `file://` plugin path in OpenCode config
- Verified Slack MCP `conversations_add_message` and `channels_list` calls succeed after fix (previously all Slack MCP calls returned TypeError)
- All existing tool functionality (edit, bash, read, grep, etc.) unaffected — the guard only skips when `output.output` is nullish

### DIFF
--- a/src/hooks/agent-usage-reminder/hook.ts
+++ b/src/hooks/agent-usage-reminder/hook.ts
@@ -77,6 +77,7 @@ export function createAgentUsageReminderHook(_ctx: PluginInput) {
       return;
     }
 
+    if (output.output == null) return;
     output.output += REMINDER_MESSAGE;
     state.reminderCount++;
     state.updatedAt = Date.now();
--- a/src/hooks/category-skill-reminder/hook.ts
+++ b/src/hooks/category-skill-reminder/hook.ts
@@ -105,7 +105,7 @@ export function createCategorySkillReminderHook(
 
     state.toolCallCount++
 
-    if (state.toolCallCount >= 3 && !state.delegationUsed && !state.reminderShown) {
+    if (state.toolCallCount >= 3 && !state.delegationUsed && !state.reminderShown && output.output != null) {
       output.output += reminderMessage
       state.reminderShown = true
       log("[category-skill-reminder] Reminder injected", {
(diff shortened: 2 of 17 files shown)


sjawhar:

PR #1730, opened 2026-02-10
fix(config): set default_builder_enabled to true by default (#1639)

### LINKED ISSUE #1639: [Feature]: Set `default_builder_enabled` to true by default
### Prerequisites

### Problem Description

I’m a huge fan and frequent user of oh-my-opencode. Thank you for all the great work!

I’ve been actively recommending oh-my-opencode to my colleagues, but I’ve personally noticed a few hurdles that make it a bit difficult for new users to adopt. I’d like to share some feedback on these points:

- Naming of Default Agents: When moving from the standard OpenCode (Build/Plan) to oh-my-opencode, users encounter several new agents like Sisyphus and Prometheus. While I agree these agents are well-orchestrated, it’s not immediately clear what each one does just by looking at their names. This often leads to confusion. Also (on a lighter note), names like Prometheus are already widely used for other popular tools, which sometimes makes me wonder if there’s a specific connection between them.

- Need for a Lightweight Workflow: Due to internal LLM restrictions, I prefer using OpenCode and oh-my-opencode over Claude Code. For quick, lightweight edits, I often prefer the standard build agent. However, since oh-my-opencode doesn't include it by default, I have to manually set `default_builder_enabled` to true. I only recently discovered this setting; before that, I simply disabled the plugin to get the same result. Since the strengths of oh-my-opencode lie not just in orchestration but also in curated skills, MCP, and Claude Code compatibility, I believe there should be an easier way to use it without the full orchestration overhead.

```
  "sisyphus_agent": {
    "default_builder_enabled": true
  },
```

### Proposed Solution

I believe setting default_builder_enabled to true by default could resolve a lot of these friction points. I understand that reversing the default value is a significant decision, but I’d love to hear your thoughts on this.

Thank you.

### Alternatives Considered

_No response_

### Doctor Output (Optional)

```shell

```

### Additional Context

_No response_

### Feature Type

New Agent

### Contribution

- [ ] I'm willing to submit a PR for this feature
- [ ] I can help with testing

potb:
Closes #1639

`default_builder_enabled` defaults to `false`, so OpenCode-Builder never registers unless explicitly opted in. Should be `true`.

One-line fix (`?? false` → `?? true`) + tests.

### DIFF
--- a/src/plugin-handlers/agent-config-handler.ts
+++ b/src/plugin-handlers/agent-config-handler.ts
@@ -89,7 +89,7 @@ export async function applyAgentConfig(params: {
 
   const isSisyphusEnabled = params.pluginConfig.sisyphus_agent?.disabled !== true;
   const builderEnabled =
-    params.pluginConfig.sisyphus_agent?.default_builder_enabled ?? false;
+    params.pluginConfig.sisyphus_agent?.default_builder_enabled ?? true;
   const plannerEnabled = params.pluginConfig.sisyphus_agent?.planner_enabled ?? true;
   const replacePlan = params.pluginConfig.sisyphus_agent?.replace_plan ?? true;
   const shouldDemotePlan = plannerEnabled && replacePlan;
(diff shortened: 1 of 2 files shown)


potb:
Closing — the `false` default is intentional. The plugin's opinionated stance is to not expose OpenCode's stock build agent alongside its own curated agent system.

PR #1732, opened 2026-02-10
fix(types): handle reasoning.encrypted content block type from Grok models

### LINKED ISSUE #1588: [Bug]: Grok 4.1 Fast 'reasoning.encrypted' data format causing Provider returned error in v3.2.4
### Description
When using `x-ai/grok-4.1-fast` (via OpenRouter) as an agent model in `oh-my-opencode` v3.2.4, the model fails to return a response and throws a "Provider returned error".

### Root Cause Analysis
- **Direct API Test**: A direct `curl` call to OpenRouter with `x-ai/grok-4.1-fast` works perfectly (response received in < 3s).
- **In-App Failure**: In the `oh-my-opencode` interface, it fails immediately.
- **Specific Observation**: Grok 4.1 Fast returns reasoning data in an encrypted format (`"type":"reasoning.encrypted"`). The new reasoning/thinking display engine in v3.2.4 likely fails to parse this specific format, leading to a crash or connection termination before the content is streamed.

### Steps to Reproduce
1. Configure an agent with `"model": "x-ai/grok-4.1-fast"` in `oh-my-opencode.json`.
2. Attempt to chat with the agent.
3. Observe "Provider returned error" while the model is attempting to "think".

### Environment
- Version: v3.2.4
- Provider: OpenRouter
- Model: x-ai/grok-4.1-fast

### Expected Behavior
The system should either handle/ignore the `reasoning.encrypted` block or simply stream the content part once reasoning is completed.

potb:
Closes #1588

Grok models return `reasoning.encrypted` content blocks. We only check for `"reasoning"` and `"thinking"`, so these blocks get silently dropped during validation, extraction, and formatting.

Added `reasoning.encrypted` to the `ThinkingPartType` union and `THINKING_TYPES` set, then updated all downstream checks. Content extraction guards on `&& part.text` so encrypted blocks without text are safely skipped.

### DIFF
--- a/src/features/background-agent/session-validator.ts
+++ b/src/features/background-agent/session-validator.ts
@@ -70,6 +70,7 @@ export async function validateSessionHasOutput(
         if (type === "tool") return true
         if (type === "text" && hasNonEmptyText(part.text)) return true
         if (type === "reasoning" && hasNonEmptyText(part.text)) return true
+        if (type === "reasoning.encrypted" && hasNonEmptyText(part.text)) return true
         if (type === "tool_result" && isToolResultContentNonEmpty(part.content)) return true
         return false
       })
--- a/src/hooks/session-recovery/constants.ts
+++ b/src/hooks/session-recovery/constants.ts
@@ -5,6 +5,6 @@ export const OPENCODE_STORAGE = getOpenCodeStorageDir()
 export const MESSAGE_STORAGE = join(OPENCODE_STORAGE, "message")
 export const PART_STORAGE = join(OPENCODE_STORAGE, "part")
 
-export const THINKING_TYPES = new Set(["thinking", "redacted_thinking", "reasoning"])
+export const THINKING_TYPES = new Set(["thinking", "redacted_thinking", "reasoning", "reasoning.encrypted"])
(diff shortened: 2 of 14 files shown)


potb:
@cubic-dev-ai recheck

code-yeongyu:
Thank you for the thorough effort across 14 files — this was clearly a lot of careful work!

After investigating the OpenCode source code, we found that `reasoning.encrypted` is not actually a part type that OpenCode emits. What exists in the OpenCode codebase is:

- `"reasoning"` — the actual part type (always set to `"reasoning"` regardless of encryption)
- `"reasoning.encrypted_content"` — an OpenAI Responses API `include` parameter value (tells the API to return encrypted reasoning)
- `encrypted_content` — a field on reasoning parts (`part.encrypted_content`)

The message processor in OpenCode (`src/session/processor.ts`) always sets the part type to `"reasoning"`, even for encrypted reasoning blocks. So the plugin code already handles these correctly without needing a separate `"reasoning.encrypted"` check.

We really appreciate you taking the time to look into Grok model compatibility — that kind of cross-model thinking is exactly what we need. Thank you! 🙏

PR #1724, opened 2026-02-10
fix: guard output.output against undefined in tool.execute.after hooks

materializerx:
## Summary

- Add null/type guards for `output.output` in 3 hooks that crash when it's `undefined`
- Fixes `TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')` for external MCP tools

## Problem

Three `tool.execute.after` hooks call string methods on `output.output` without checking if it's defined:

| Hook | Crash site |
|------|-----------|
| `comment-checker/hook.ts` | `output.output.toLowerCase()` (line 95) |
| `edit-error-recovery/hook.ts` | `output.output.toLowerCase()` (line 47) |
| `task-resume-info/hook.ts` | `output.output.startsWith()` (line 24) |

When OpenCode passes `undefined` for `output.output` (observed with external MCP tools), all three crash with `TypeError`.

## Fix

Added `if (!output.output || typeof output.output !== "string") return` guard before accessing `output.output` methods. This matches the existing pattern in `tool-output-truncator.ts` (line 42) which already had this guard.

## Context

- Related issue: #1720 (v3.4.1 broke all external MCP tools)
- All external MCP tools affected (user-configured and OMO-bundled)
- Built-in tools (bash, read, grep, write) were unaffected since OpenCode populates `output.output` for them

---
## Summary by cubic
Guard output.output against undefined in tool.execute.after hooks to prevent crashes with external MCP tools. Fixes #1720.

- **Bug Fixes**
  - Add null/type checks before calling string methods in comment-checker, edit-error-recovery, and task-resume-info hooks.
  - Early-return when output.output is missing or not a string (log skip in comment-checker), matching tool-output-truncator behavior.

<sup>Written for commit 290cf3db632f1fe34e49f7c43d385a104a2cd360. Summary will update on new commits.</sup>

Issue #1720, opened 2026-02-10
FYI: v3.4.1 broke all external MCP tools (published & yanked)


dinhphieu:
## What Happened

v3.4.1 was published to npm on `2026-02-10T13:04:04.741Z` and subsequently unpublished. Users who installed during the window got a broken version cached locally.

## Symptom

All **external MCP tools** fail with:

```
TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')
```

Affected: any MCP whose response shape has `output.output` as `undefined` — both user-configured MCPs (e.g. in `.mcp.json`) and OMO-bundled MCPs. Local/simple tools (bash, read, grep, write) were unaffected.

## Root Cause

The plugin's MCP response post-processing calls `.toLowerCase()` on `output.output`, which is `undefined` for certain MCP response shapes.

## Why Users Still Got 3.4.1 After It Was Yanked

Even after `npm unpublish`, **npm/bun CDN caches can still serve the tarball** for a short window. The `latest` dist-tag was reverted to 3.4.0 and `npm view versions` no longer lists 3.4.1, but `npm view oh-my-opencode time` still shows the publish timestamp:

```
"3.4.1": "2026-02-10T13:04:04.741Z"
```

Users whose lockfiles or package managers resolved during the cache window ended up with the ghost version.

## How to Fix (if you're affected)

Nuke and reinstall:

```bash
cd ~/.config/opencode
rm -rf node_modules bun.lock package-lock.json package.json
# Restart opencode — re-resolves to 3.4.0
```

Or pin explicitly in `opencode.json`:

```json
"plugin": ["oh-my-opencode@3.4.0"]
```

## Diagnosis

```bash
npm view oh-my-opencode dist-tags --json   # latest → 3.4.0 (correct)
npm view oh-my-opencode time --json        # 3.4.1 timestamp still visible
npm view oh-my-opencode versions --json    # 3.4.1 NOT listed (yanked)
```

If opencode shows a version that doesn't exist on `npm view versions`, you hit the CDN cache race.

## Suggestion

Consider adding a `stable` dist-tag that only gets updated after smoke-testing, so `@stable` users don't hit publish-then-yank scenarios.

ErcinDedeoglu:
Can confirm — hit the same issue today.

**Affected tools**: Every single MCP tool (ssh-mcp, GitHub, Keycloak, Figma, etc.). Built-in tools (bash, read, grep, write) were fine.

[...]

PR #1735, opened 2026-02-10
fix: guard against undefined output.output in tool.execute.after hooks

well0nez:
## Summary

MCP tools may not return `output.output` as a string, causing a `TypeError` when calling `.toLowerCase()` on it.

**Error:** `TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')`

This happens when using MCP servers like `ida-pro-mcp` where the tool output format differs from standard OpenCode tools.

## Changes

- Added guard in `edit-error-recovery/hook.ts` to check if `output.output` exists and is a string before calling string methods
- Added guard in `comment-checker/hook.ts` with the same protection

## Files Changed

- `src/hooks/edit-error-recovery/hook.ts`
- `src/hooks/comment-checker/hook.ts`

## Testing

Tested with `ida-pro-mcp` MCP server - the error no longer occurs and MCP tools work correctly.

---
## Summary by cubic
Prevent crashes in tool.execute.after hooks by guarding against undefined or non-string output.output before calling .toLowerCase(). Fixes TypeErrors with MCP servers like ida-pro-mcp and keeps tool execution stable.

<sup>Written for commit b28269727b5d4e870651adbc1310745ea7770084. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/comment-checker/hook.ts
+++ b/src/hooks/comment-checker/hook.ts
@@ -91,6 +91,12 @@ export function createCommentCheckerHooks(config?: CommentCheckerConfig) {
 
       const toolLower = input.tool.toLowerCase()
 
+      // Guard against undefined output (e.g., MCP tools may not return output.output)
+      if (!output.output || typeof output.output !== "string") {
+        debugLog("skipping due to non-string output:", typeof output.output)
+        return
+      }
+
       // Only skip if the output indicates a tool execution failure
       const outputLower = output.output.toLowerCase()
       const isToolFailure =
--- a/src/hooks/edit-error-recovery/hook.ts
+++ b/src/hooks/edit-error-recovery/hook.ts
@@ -44,6 +44,9 @@ export function createEditErrorRecoveryHook(_ctx: PluginInput) {
     ) => {
       if (input.tool.toLowerCase() !== "edit") return
 
+      // Guard against undefined output (e.g., MCP tools may not return output.output)
+      if (!output.output || typeof output.output !== "string") return
+
       const outputLower = output.output.toLowerCase()
       const hasEditError = EDIT_ERROR_PATTERNS.some((pattern) =>
         outputLower.includes(pattern.toLowerCase())


well0nez:

mathew-cf:
Please merge soon :)
I know a few folks whose MCP use is having issues because of this.

NikkeTryHard:
## Full Audit: `output.output` Access in `tool.execute.after` Hooks

After auditing the source, this PR fixes 2 of 4 hooks that crash on MCP tool results. There are also 7 hooks with a silent corruption pattern.

### Hard Crashes (TypeError)

| # | Hook | File | Line | Pattern | PR #1735 |
|---|------|------|------|---------|----------|
| 1 | comment-checker | `hook.ts` | L95 | `output.output.toLowerCase()` | Fixed |
| 2 | edit-error-recovery | `hook.ts` | L47 | `output.output.toLowerCase()` | Fixed |
| 3 | task-resume-info | `hook.ts` | L24-27 | `.startsWith()`, `.includes()`, `.trimEnd()` | **Missing** |
| 4 | delegate-task-retry | `hook.ts` | L14 | Passed to `detectDelegateTaskError()` which calls `.includes()` at `patterns.ts:65` | **Missing** |

The same one-liner used in this PR resolves all four:

```typescript
if (!output.output || typeof output.output !== "string") return
```

### Silent Corruption (`undefined += "string"` = `"undefinedstring"`)

These hooks append to `output.output` without checking its type. When `output.output` is `undefined`, JavaScript coerces it to the string `"undefined"` and concatenates, producing corrupted output like `"undefinedRemember to use..."`:

| Hook | File | Line |
|------|------|------|
| agent-usage-reminder | `hook.ts` | L80 |
| task-reminder | `hook.ts` | L42 |
| category-skill-reminder | `hook.ts` | L109 |
| context-window-monitor | `hook.ts` | L77 |
| rules-injector | `injector.ts` | L119 |
| directory-agents-injector | `injector.ts` | L49 |
| directory-readme-injector | `injector.ts` | L49 |

These won't crash, but they produce garbage output for MCP tools. A guard like `if (typeof output.output !== "string") return` or changing `output.output +=` to `output.output = (output.output ?? "") +` would fix them.

### Already-Safe Hooks (for reference)

`tool-output-truncator`, `empty-task-response-detector`, `atlas/tool-execute-after`, `interactive-bash-session`, and `claude-code-hooks/tool-execute-after-handler` all have proper guards (`typeof` checks, optional chaining, or falsy guards).

### Root Cause

Since `ai@6.0.0` (Dec 22, 2025), `ToolResultPart.output` is no longer the raw tool result. It is now wrapped by `createToolModelOutput()` ([introduced in vercel/ai#6784](https://github.com/vercel/ai/pull/6784)) into a `ToolResultOutput` envelope:

```
Before: output.output → "some string"          (raw tool result)
After:  output.output → { type: "text", value: "some string" }
```

The hooks expect the old shape. For MCP tools specifically, `output.output` resolves to `undefined` because the envelope object has no `.output` property — only `.type` and `.value`. Built-in tools are unaffected because OpenCode's `prompt.ts` wraps their results into `{ output: string, title: string, metadata: object }` before the SDK transforms them.

### Related

- anomalyco/opencode#13042 (upstream root cause, closed as oh-my-opencode bug)
- anomalyco/opencode#12987 (duplicate)
- #1746, #1737, #1720

timoteo7 (approved):

code-yeongyu:
Thanks for working on this fix! We're going with #1724 which covers this more completely, so closing this one in its favor. Appreciate you taking the time to contribute — thank you! 🙏

Issue #1737, opened 2026-02-10
TypeError: output.output.toLowerCase() crashes on MCP tools with undefined output


pondzikk:
## Bug

`oh-my-opencode@3.5.0` crashes with `TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')` when certain MCP tools (e.g., `websearch_web_search_exa`, `context7_query-docs`) return results where `output.output` is `undefined`.

## Root Cause

Two unguarded `.toLowerCase()` calls on `output.output` in `dist/index.js`:

1. **Line 26788** — `tool.execute.after` hook in the main tool execution handler:
```javascript
const outputLower = output.output.toLowerCase(); // crashes if output.output is undefined
```

2. **Line 42754** — `createEditErrorRecoveryHook`:
```javascript
const outputLower = output.output.toLowerCase(); // same issue
```

## Expected Behavior

These hooks should safely handle cases where `output.output` is `undefined` or not a string (e.g., early return).

## Suggested Fix

Add null/type guards before accessing `output.output`:

```javascript
// Site 1 (main hook)
if (!output.output || typeof output.output !== "string") {
  return;
}
const outputLower = output.output.toLowerCase();

// Site 2 (edit error recovery hook)  
if (!output.output || typeof output.output !== "string")
  return;
const outputLower = output.output.toLowerCase();
```

## Reproduction

1. Install `oh-my-opencode@3.5.0` with any MCP tool that returns non-string output (websearch_web_search_exa, context7, etc.)
2. Use the MCP tool in a session
3. Observe TypeError in the console

## Environment

- OpenCode: v1.1.55
- oh-my-opencode: v3.5.0
- OS: Linux

poodbooq:
I have the same issue.

poodbooq:
https://github.com/code-yeongyu/oh-my-opencode/pull/1735

materializerx:
#1724

cjnova:
same here

vendys-sungwan:
+1

ChickenBreast-ky:
same here

Temporary fix -> add it to oh-my-opencode.jsonc

"disabled_hooks": ["comment-checker"], // Temporarily added to address bug in 3.5.1 (MCP)

code-yeongyu:
[sisyphus-bot] Thank you for reporting this issue! This is a friendly reminder that all project communications (issues, PRs, comments) must be in English per our project policy.

[...]

PR #1755, opened 2026-02-11
fix: add null guards for output.output in tool.execute.after hooks

### LINKED ISSUE #1746: [Bug]: Unguarded output.output access in tool.execute.after hooks crashes all MCP tools
## Bug

All external MCP tools crash with `TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')` since v3.5.0 (and v3.4.1 before it was yanked). Built-in tools (bash, read, write, grep) are unaffected.

## Root Cause

Three `tool.execute.after` hooks access `output.output` without null/type guards:

| File | Line | Code |
|------|------|------|
| `src/hooks/comment-checker/hook.ts` | L95 | `const outputLower = output.output.toLowerCase()` |
| `src/hooks/edit-error-recovery/hook.ts` | L47 | `const outputLower = output.output.toLowerCase()` |
| `src/hooks/task-resume-info/hook.ts` | L24-31 | `output.output.startsWith()`, `.includes()`, `.trimEnd()` |

All three hooks declare `output` as `{ title: string; output: string; metadata: unknown }`, but for MCP tool results, `output.output` is `undefined`.

**Contributing upstream factor**: AI SDK v6.0.74 changed tool result structure from `{output, metadata, title}` to `{type, value}` via `createToolModelOutput()` in `stream-text.ts`. This means `output.output` is now always `undefined` for MCP tools. See anomalyco/opencode#13042, anomalyco/opencode#12987.

## Why v3.5.0 and v3.5.1 still have this

v3.4.1 was yanked for this exact issue, but the hook source code was never patched. v3.5.0 and v3.5.1 shipped with the same unguarded calls.

## Workaround

```json
"disabled_hooks": ["comment-checker"]
```

Or downgrade to v3.4.0 with `"auto_update": false`.

## Suggested Fix

Add null/type guard at the top of each `tool.execute.after` handler:

```typescript
if (!output.output || typeof output.output !== "string") return
```

This single line in each of the three hooks would prevent the crash.

## Related

- #1737 (same TypeError report)
- #1720 (v3.4.1 broke all external MCP tools)
- PR #1735 (community fix attempt)
- anomalyco/opencode#13042 (upstream AI SDK breaking change)
- anomalyco/opencode#12987 (upstream duplicate)

## Affected Versions

- v3.4.1 (yanked)
- v3.5.0
- v3.5.1

code-yeongyu:
Fixes #1746

## Problem
Three `tool.execute.after` hooks crash with `TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')` when MCP tools return results where `output.output` is `undefined`. This breaks ALL external MCP tools since v3.5.0.

## Affected Hooks
| File | Line | Crash |
|------|------|-------|
| `comment-checker/hook.ts` | L95 | `output.output.toLowerCase()` |
| `edit-error-recovery/hook.ts` | L47 | `output.output.toLowerCase()` |
| `task-resume-info/hook.ts` | L24-31 | `.startsWith()`, `.includes()`, `.trimEnd()` |

## Fix
Added `if (!output.output || typeof output.output !== "string") return` guard at the top of each affected hook handler.

## Testing (TDD)
- Tests for each hook with undefined `output.output` — all pass
- Existing tests unaffected

## Related
- #1737, #1720 (user reports)
- PR #1735 (community fix attempt)
- anomalyco/opencode#13042 (upstream AI SDK 6.0.74 breaking change)

---
## Summary by cubic
Prevent crashes in tool.execute.after hooks caused by undefined output from MCP tools, restoring compatibility with external tools. Fixes #1746.

- **Bug Fixes**
  - Added string guards and early returns for output.output in comment-checker, edit-error-recovery, and task-resume-info.
  - Added tests to cover undefined output in all three hooks.

<sup>Written for commit 8b44d3b7b5f3262eb32644f4e7be7b365ed65eb4. Summary will update on new commits.</sup>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -28,13 +28,13 @@
         "typescript": "^5.7.3",
       },
       "optionalDependencies": {
-        "oh-my-opencode-darwin-arm64": "3.3.1",
-        "oh-my-opencode-darwin-x64": "3.3.1",
-        "oh-my-opencode-linux-arm64": "3.3.1",
-        "oh-my-opencode-linux-arm64-musl": "3.3.1",
-        "oh-my-opencode-linux-x64": "3.3.1",
-        "oh-my-opencode-linux-x64-musl": "3.3.1",
-        "oh-my-opencode-windows-x64": "3.3.1",
+        "oh-my-opencode-darwin-arm64": "3.5.1",
+        "oh-my-opencode-darwin-x64": "3.5.1",
+        "oh-my-opencode-linux-arm64": "3.5.1",
+        "oh-my-opencode-linux-arm64-musl": "3.5.1",
+        "oh-my-opencode-linux-x64": "3.5.1",
+        "oh-my-opencode-linux-x64-musl": "3.5.1",
+        "oh-my-opencode-windows-x64": "3.5.1",
       },
     },
   },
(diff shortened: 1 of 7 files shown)


code-yeongyu:
Closing in favor of #1756.

#1756 uses `?? ""` (nullish coalescing) instead of early return, which is safer — early return here skips comment checks entirely for write/edit cases where `output.output` is undefined (cubic also flagged this as P2). #1756 also covers the atlas hook fix and has more comprehensive tests (101-line task-resume-info suite). cubic gave it 5/5 + auto-approve vs 4/5 here.

PR #1822, opened 2026-02-13
fix(non-interactive-env): prevent environment variable duplication on repeated executions

Strocs:
## Summary

- Fixes environment variable duplication when multiple git commands are executed in sequence
- Adds idempotency check to prevent re-prepending env vars that are already present
- Maintains non-interactive behavior without reintroducing the removed isNonInteractive() check

## Changes

- Added idempotency guard in non-interactive-env-hook.ts (lines 59-63)
- Hook now checks if command already starts with the env prefix before prepending
- Early return if prefix is already present, avoiding duplication like export CI=true ... export CI=true ... git commit
- No changes to core behavior: env vars are still always injected for git commands (as per commit 8bf32025)

## Screenshots

| Before | After |
|:---:|:---:|
| Screenshot 2026-02-13 012039 |  image |

> Screenshots after multiples calls

## Testing

```bash
bun run typecheck
bun test src/hooks/non-interactive-env/
```
**Verification:** All 15 tests pass. 

---
## Summary by cubic
Prevents duplicate env vars in the non-interactive Git hook by making prefixing idempotent. Behavior is unchanged; env vars are injected once, avoiding stacked prefixes across sequential commands.

- **Bug Fixes**
  - Add idempotency guard and test to skip re-prefixing when the env prefix is already present.
  - Preserve existing behavior: inject env vars once for git commands.

<sup>Written for commit 8500abeb39aaec4cda544ae4417fe4fd8236bd6f. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/non-interactive-env/index.test.ts
+++ b/src/hooks/non-interactive-env/index.test.ts
@@ -111,6 +111,34 @@ describe("non-interactive-env hook", () => {
 
       expect(output.args.command).toBeUndefined()
     })
+
+    test("#given git command already has prefix #when hook executes again #then does not duplicate prefix", async () => {
+      const hook = createNonInteractiveEnvHook(mockCtx)
+      
+      // First call: transforms the command
+      const output1: { args: Record<string, unknown>; message?: string } = {
+        args: { command: "git commit -m 'test'" },
+      }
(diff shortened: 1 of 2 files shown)


Strocs:

Issue #1825, opened 2026-02-14
[Bug]: Run CLI command never exits


CloudWaddie:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

I am trying to make https://github.com/CloudWaddie/actions-agent work but it never exits after opencode run working. Apparently this was fixed in the dev branch but I don't see that working, even using the dev branch.

### Steps to Reproduce

1. Use my workflow as a template
2. Ping the agent
3. It never exits

### Expected Behavior

1. Use my workflow as a template
2. Ping the agent
3. Exits

### Actual Behavior

1. Use my workflow as a template
2. Ping the agent
3. It never exits

### Doctor Output

```shell
Not applicable
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Linux

### OpenCode Version

Latest

CloudWaddie:
It sends comments and stuff but never exits.

CloudWaddie:
<img width="1268" height="987" alt="Image" src="https://github.com/user-attachments/assets/d9169187-bf25-4679-bca5-8bccea5e6933" />

CloudWaddie:
BTW opencode has suggested that it is an issue in oh-my-opencode NOT my action.

cloudwaddie-agent:
Opened upstream fix PR:
- https://github.com/code-yeongyu/oh-my-opencode/pull/1827

What I fixed:
- `run` could hang forever during shutdown after `abortController.abort()` because it did an unbounded `await eventProcessor`.
- If no new event arrives, the `for await` loop in `processEvents()` never advances to observe abort, so `eventProcessor` stays pending.

Patch:
- Replaced unbounded wait with bounded shutdown wait (`waitForEventProcessorShutdown`, default 2s) and continue shutdown afterward.
- Added tests for both resolved and never-resolving event processor cases.

This targets the “run command never exits” symptom directly.

CloudWaddie:
@code-yeongyu my agent (@cloudwaddie-agent ) has suggested a fix

PR #1827, opened 2026-02-14
fix(cli-run): avoid hanging on event stream shutdown

cloudwaddie-agent:
## Summary
Fix `oh-my-opencode run` hanging during shutdown when the event stream doesn't close after abort.

## Root cause
`runner.ts` calls:
1. `abortController.abort()`
2. `await eventProcessor`

But `processEvents()` checks `abortController.signal.aborted` **inside** a `for await` loop. If no further stream event arrives after abort, the loop never advances to that check, so `eventProcessor` can remain pending forever.

## Changes
- Add `waitForEventProcessorShutdown()` helper in `runner.ts`
  - waits for event processor completion
  - but only up to a bounded timeout (default 2000ms)
  - logs a dim warning and continues shutdown if stream doesn't close
- Replace unbounded `await eventProcessor` with bounded shutdown wait
- Add tests in `runner.test.ts`:
  - returns quickly when processor resolves
  - times out and continues when processor never resolves

## Why this helps issue #1825
Even when the model work is done and comments are posted, the CLI no longer blocks forever waiting for event-stream teardown.

## Notes
Could not run tests in this environment because `bun` is unavailable in PATH.

---
## Summary by cubic
Fixes a shutdown hang in `oh-my-opencode run` when the event stream doesn’t close after abort. The CLI now exits promptly by waiting briefly for the event processor and timing out if needed.

- **Bug Fixes**
  - Added waitForEventProcessorShutdown with a 2s timeout and dim warning; used instead of awaiting the event processor after abort.
  - Expanded tests to cover quick completion, timeout, and that the warning is logged.

<sup>Written for commit acd26c2d7c5ee20270350bef8ce29f40608fab1a. Summary will update on new commits.</sup>

### DIFF
--- a/src/cli/run/runner.test.ts
+++ b/src/cli/run/runner.test.ts
@@ -33,7 +33,9 @@ describe("waitForEventProcessorShutdown", () => {
     // then
     expect(elapsed).toBeGreaterThanOrEqual(20)
     expect(logSpy).toHaveBeenCalled()
-  })
+
+    // cleanup
+    logSpy.mockRestore()
 })
 
 describe("resolveRunAgent", () => {
--- a/src/cli/run/runner.test.ts
+++ b/src/cli/run/runner.test.ts
@@ -1,9 +1,39 @@
-import { describe, it, expect } from "bun:test"
+import { describe, it, expect, spyOn } from "bun:test"
 import type { OhMyOpenCodeConfig } from "../../config"
-import { resolveRunAgent } from "./runner"
+import { resolveRunAgent, waitForEventProcessorShutdown } from "./runner"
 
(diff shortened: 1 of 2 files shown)


cloudwaddie-agent:

code-yeongyu:
Thank you @cloudwaddie-agent for identifying this issue and providing a fix! Your analysis of the `waitForEventProcessorShutdown` hanging problem was spot-on.

We've incorporated your approach in #1828 with a bounded timeout mechanism. You're credited as Co-authored-by in the commit.

Closing this in favor of #1828 which includes additional test coverage and the bounded wait pattern.

code-yeongyu:
Superseded by #1828

Issue #1671, opened 2026-02-08
[Bug]: tmux.layout options are not applied when spawning subagent panes


Slinet6056:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

`tmux.layout` values (e.g., `main-vertical`, `main-horizontal`) appear to have almost no visible effect during normal subagent spawning.
In practice, newly spawned panes are usually created to the right side of the current/main pane, regardless of the configured layout.
This makes it impossible to get a stable "vertical stack" behavior for all subagent panes.

### Steps to Reproduce

1. Set config like:
   - `tmux.enabled: true`
   - `tmux.layout: main-horizontal` (or try other layout values)
2. Start OpenCode in server mode.
4. Trigger multiple background subagents.
5. Observe pane creation behavior.

### Expected Behavior

- Layout should follow `tmux.layout` consistently.
- With `main-horizontal`, panes should be arranged as a vertical stack (top-to-bottom) rather than opening on the right.
- Different layout values should produce clearly different pane arrangements.

### Actual Behavior

- New panes are created to the right side.
- Switching `tmux.layout` values has no practical difference in pane arrangement during live spawning.

  <details>
  <summary>main-vertical</summary>

  <img width="3840" height="2066" alt="Image" src="https://github.com/user-attachments/assets/231e8f52-50cb-4c31-9050-ef9f12991278" />

  </details>

  <details>
  <summary>main-horizontal</summary>

  <img width="3840" height="2072" alt="Image" src="https://github.com/user-attachments/assets/b028a742-1645-4a17-8c8b-5ca572803b4a" />

  </details>

### Doctor Output

```shell
Installation
────────────────────────────────────────
  ✓ OpenCode Installation → 1.1.51
  ✓ Plugin Registration → Registered (pinned: latest)

Configuration
────────────────────────────────────────
  ✓ Configuration Validity → Valid JSON config
  ✓ Model Resolution → 10 agents, 8 categories (18 overrides), 2449 available

Authentication
────────────────────────────────────────
[...]

PR #1844, opened 2026-02-14
fix(tmux): prefer split-or-defer with FIFO deferred attach

liu-qingyuan:
## Summary

- Prefer pane split when geometry allows, and defer attach when no safe split target exists.
- Add FIFO deferred attach queueing so deferred sessions are attached in creation order once capacity returns.
- Add/expand tmux subagent and layout tests covering split-or-defer decisions, queue idempotency, and deferred attach lifecycle.

## Why

Under constrained pane geometry, replacing active panes can evict in-progress work and reduce reliability.
Defer-and-reattach preserves active panes while still ensuring pending sessions are attached when pane capacity is available.

## Testing

- `bun test src/features/tmux-subagent` - Passed (56 pass, 0 fail)
- `bun run build` - Passed

## Notes

- Constrained pane geometry is handled via defer/reattach behavior (FIFO) instead of replacing active panes.

---
## Summary by cubic
Prefer splitting panes when possible and defer attaching when not, using a FIFO queue that auto-attaches when capacity returns. After each spawn/close, we reapply the tmux layout and enforce main-pane sizing from config, and wait for the session to be ready before tracking.

- **New Features**
  - Layout-aware split-or-defer via CapacityConfig (layout, mainPaneSize, min widths).
  - Strict main layouts: deterministic initial/follow-up split directions and config-driven spawn target selection.
  - FIFO deferred attach with background polling; idempotent; attaches in order; removed on session delete; spawns serialized via an internal queue.
  - Reapply layout (applyLayout) and enforce main-pane width using computed sizes after actions.

- **Refactors**
  - Removed “replace oldest” fallback; defer attach instead. When non-strict layout allows a single eviction, use close+spawn.
  - Grid planning and split checks use computeMainPaneWidth/computeAgentAreaWidth and agent min width for thresholds.
  - Pane state query switches to tab-delimited parsing with deterministic main-pane selection.
  - Expanded tests for layout rules, target ordering, split-or-defer behavior, FIFO queue lifecycle, and layout utils.

<sup>Written for commit 5f78c071890e05c8dbe10278ecd71bb70478f063. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/tmux-subagent/action-executor.ts
+++ b/src/features/tmux-subagent/action-executor.ts
@@ -1,14 +1,21 @@
-import type { PaneAction } from "./types"
-import { applyLayout, spawnTmuxPane, closeTmuxPane, enforceMainPaneWidth, replaceTmuxPane } from "../../shared/tmux"
+import type { TmuxConfig } from "../../config/schema"
+import type { PaneAction, WindowState } from "./types"
+import {
+  applyLayout,
+  spawnTmuxPane,
+  closeTmuxPane,
+  enforceMainPaneWidth,
+  replaceTmuxPane,
+} from "../../shared/tmux"
+import { getTmuxPath } from "../../tools/interactive-bash/tmux-path-resolver"
+import { queryWindowState } from "./pane-state-querier"
 import { log } from "../../shared"
 import type {
-  ActionExecutorDeps,
   ActionResult,
-  ExecuteContext,
+  ActionExecutorDeps,
 } from "./action-executor-core"
-import { executeActionWithDeps } from "./action-executor-core"
 
-export type { ActionExecutorDeps, ActionResult, ExecuteContext } from "./action-executor-core"
+export type { ActionExecutorDeps, ActionResult } from "./action-executor-core"
 
 export interface ExecuteActionsResult {
   success: boolean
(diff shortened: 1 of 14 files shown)


liu-qingyuan:

liu-qingyuan:
f7a3667c-b289-454f-8bdb-10307e2a7b16
screenshot-20260215-034831


Fixes: https://github.com/code-yeongyu/oh-my-opencode/issues/1671

## What was wrong

`tmux.layout` behavior during live subagent spawning could become inconsistent, especially under split-capacity constraints.  
In practice, pane placement could ignore the intended layout decision path and feel unstable.

## What changed

- Removed the active-path replace fallback.
- Switched to **split-or-defer** when no valid split target is available.
- Kept deferred subagent sessions alive in background.
- Added **FIFO deferred auto-attach** when tmux capacity becomes available again.
- Added/updated regression tests around decision logic, layout handling, and deferred attach lifecycle.

## Why this fixes #1671

This makes pane spawning follow the layout decision path consistently instead of replacing active panes when geometry is tight, which was a key source of the “layout looks ignored” behavior.

### Config used (for repro / verification)

```json
// Tmux multi-pane view for background subagents
"tmux": {
  "enabled": true,
  "layout": "main-vertical",
  "main_pane_size": 60,
  "main_pane_min_width": 120,
  "agent_pane_min_width": 40
}
```
This is the config I used to reproduce/verify the issue locally.

## Note on limits

tmux geometry limits (height/width thresholds) still apply.  
When there is not enough space, sessions are deferred and attached later, rather than forcing an immediate split/replacement.

## Testing

- Added regression tests for decision logic and deferred attach lifecycle.
- Manually verified the `main-horizontal` scenario under tight geometry constraints.

If possible, could you retest on the latest `dev`/next build and confirm whether the `main-horizontal` scenario now behaves as expected?

liu-qingyuan (commented):
Hope to approve workflows to run

PR #1870, opened 2026-02-15
fix(background-agent): honor disabled background-notification for system reminders

dankochetov:
## Summary
- Stop injecting parent-session `<system-reminder>` background task notifications when `background-notification` hook is disabled.
- Wire `BackgroundManager` notification behavior to hook enablement so `disabled_hooks: ["background-notification"]` is authoritative.
- Add regression test proving parent prompt injection is skipped when notifications are disabled.

## Root cause
`disabled_hooks.background-notification` only disabled the hook wrapper, but background task reminders were still emitted directly from `BackgroundManager.notifyParentSession()` via `client.session.promptAsync(...)`.

## Validation
- `bun run typecheck`

---
## Summary by cubic
Skip background task system reminders to parent sessions when the background-notification hook is disabled. This makes the setting authoritative and prevents unwanted prompts.

- **Bug Fixes**
  - Gate parent-session notifications in BackgroundManager via enableParentSessionNotifications, skipping prompt injection when disabled.
  - Plumb isHookEnabled("background-notification") from plugin entry through createManagers to set the gate.
  - Add regression test to confirm no prompt is sent when notifications are disabled.

<sup>Written for commit 9b187e212837244efc10e101385b1e9560ca7c08. Summary will update on new commits.</sup>


dankochetov:

PR #1837, opened 2026-02-14
feat: OpenCode beta SQLite migration compatibility

code-yeongyu:
## Summary

Complete v1.2.x SQLite backend migration — **zero functional degradation** confirmed by Oracle (3 reviews).

### What changed
- **19 commits** migrating all JSON file-based storage operations to SDK/HTTP API for OpenCode v1.2.x SQLite backend
- Every function that touches message storage now has `isSqliteBackend()` runtime branching
- READ operations → SDK `client.session.messages()` / `client.session.message()`
- WRITE operations → HTTP PATCH/DELETE to `/session/:sid/message/:mid/part/:pid`
- Auth via `getServerBasicAuthHeader()` from existing util

### Migration scope
| Area | Files Changed | Status |
|------|--------------|--------|
| HTTP API helpers | `opencode-http-api.ts` | ✅ New |
| Session recovery (readers) | parts-reader, messages-reader | ✅ SDK paths |
| Session recovery (writers) | thinking-strip/prepend, empty-text, text-part-injector | ✅ HTTP PATCH/DELETE |
| Context window recovery | tool-result-storage, message-builder, target-token-truncation | ✅ SDK/HTTP |
| Pruning (dedup + truncation) | pruning-deduplication, pruning-tool-output-truncation | ✅ SDK/HTTP |
| Hook callers | atlas, prometheus-md-only, todo-continuation-enforcer | ✅ SDK reads |
| Tool callers | delegate-task, background-task, call-omo-agent | ✅ SDK reads |
| Session manager | sessionExists, session_read | ✅ SDK verification |
| Recovery callers | recover-thinking-*, recover-empty-content, recover-tool-result-missing | ✅ Async SDK |

### Oracle reviews
1. **Review #1**: Rejected — found 6 caller files still using disabled sync JSON functions → Fixed in Phase 2
2. **Review #2**: Rejected — found 3 gaps (recoverToolResultMissing, sessionExists, pruning) → Fixed in Phase 3
3. **Review #3**: **Approved** — "v1.2.x SQLite backend parity looks complete with no functional degradation"

### Tests
- **2725 pass, 0 fail** (`bun test` — full suite)
- 22 snapshots, 5962 expect() calls

PR #1819, opened 2026-02-13
feat(browser-automation): add playwright-cli as browser automation provider

jonasherr:
## Summary

Adds Microsoft's [`@playwright/cli`](https://github.com/microsoft/playwright-cli) as a new browser automation provider option. When configured, the `"playwright"` skill uses CLI commands via Bash instead of the MCP server — more token-efficient for coding agents.

## Motivation

The MCP-based `"playwright"` provider loads large tool schemas and accessibility trees into model context. The CLI alternative (`@playwright/cli`) avoids this overhead, resulting in significantly lower token usage per browser interaction.

## Changes

- Added `"playwright-cli"` to `BrowserAutomationProviderSchema` enum
- Created `playwrightCliSkill` with official Microsoft template from [@playwright/cli](https://github.com/microsoft/playwright-cli)
- Updated skill selection logic to handle 3 providers (playwright, agent-browser, playwright-cli)
- Skill uses `name: "playwright"` so agents seamlessly use it via the existing skill name — the `browserProvider` config swaps the implementation, not the name
- Added comprehensive tests for schema validation and skill selection

## Usage

```json
{
  "browser_automation_engine": {
    "provider": "playwright-cli"
  }
}
```

Requires `@playwright/cli` installed globally:
```bash
npm install -g @playwright/cli@latest
```

## Testing

- All existing tests pass
- New tests for playwright-cli provider validation and skill selection
- Manually verified end-to-end: agent correctly uses `playwright-cli` CLI commands via Bash

## References

- [@playwright/cli](https://github.com/microsoft/playwright-cli)
- [Skill template source](https://github.com/microsoft/playwright-cli/blob/main/skills/playwright-cli/SKILL.md)

---
## Summary by cubic
Adds Playwright CLI as a browser automation provider and swaps the existing "playwright" skill to run via Bash when selected to reduce token usage and avoid MCP overhead. Selecting "playwright-cli" keeps the same "playwright" skill name and behavior for agents.

- **New Features**
  - Provider: "playwright-cli" added to BrowserAutomationProviderSchema and BuiltinSkillNameSchema
  - New Playwright CLI skill template wired behind the "playwright" skill name
  - Skill selection now supports playwright, agent-browser, and playwright-cli
  - Tests added for schema/config validation and provider-based selection; JSON schema regenerated

- **Migration**
  - Install: npm install -g @playwright/cli@latest
  - Configure: browser_automation_engine.provider: "playwright-cli"

[...]

PR #1887, opened 2026-02-16
fix(tests): mock shared/opencode-message-dir in session-manager storage tests

jonasherr:
## Summary

Fixes 4 broken tests on `dev` introduced by #1837 (OpenCode beta SQLite migration).

## Root Cause

PR #1837 moved `getMessageDir` from `storage.ts` to `shared/opencode-message-dir.ts`. The new module imports `MESSAGE_STORAGE` from `shared/opencode-storage-paths` — but the test only mocks `./constants`. This means `getMessageDir` uses the real (non-existent in CI) storage path instead of the test's temp directory, causing it to always return `null`.

**Failing tests:**
- `getMessageDir finds session in direct path`
- `sessionExists returns true for existing session`
- `readSessionMessages sorts messages by timestamp`
- `getSessionInfo aggregates session metadata correctly`

## Fix

Added a `mock.module("../../shared/opencode-message-dir")` that provides a `getMessageDir` implementation using `TEST_MESSAGE_STORAGE`, matching the real function's logic but pointed at the test directory.

## Verification

All 19 tests in `storage.test.ts` pass locally. This also unblocks CI for all open PRs (e.g. #1819, #1884).

---
## Summary by cubic
Fixes four failing session-manager storage tests by mocking shared/opencode-message-dir to use the test message storage directory. Restores test isolation after getMessageDir moved in #1837 and unblocks CI.

- **Bug Fixes**
  - Root cause: getMessageDir now imports MESSAGE_STORAGE from shared/opencode-storage-paths, so the existing ./constants mock didn’t apply and CI used a real (missing) path.
  - Fix: Added a mock for shared/opencode-message-dir that implements getMessageDir using TEST_MESSAGE_STORAGE, mirroring the real lookup logic.

<sup>Written for commit 32c30505b1c73dee3e86039eac1564442c79778a. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/session-manager/storage.test.ts
+++ b/src/tools/session-manager/storage.test.ts
@@ -31,6 +31,34 @@ mock.module("../../shared/opencode-storage-detection", () => ({
   resetSqliteBackendCache: () => {},
 }))
 
+// getMessageDir was moved to shared/opencode-message-dir which imports MESSAGE_STORAGE
+// from shared/opencode-storage-paths (not ./constants). Mock it to use the test directory.
+mock.module("../../shared/opencode-message-dir", () => {
+  const { existsSync, readdirSync } = require("node:fs")
+  const { join } = require("node:path")
+  return {
+    getMessageDir: (sessionID: string): string | null => {
+      if (!sessionID.startsWith("ses_")) return null
+      if (/[/\\]|\.\./.test(sessionID)) return null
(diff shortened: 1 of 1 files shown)


jonasherr:
Closing — the same fix was already merged to dev in 65bca832. CI is green again.

PR #1952, opened 2026-02-18
fix: make write-existing-file-guard read-gated with overwrite bypass

### LINKED ISSUE #1871: [Bug]: write-existing-file-guard causes significant token waste when agents repeatedly fail and retry with edit
### Prerequisites

### Bug Description

The `write-existing-file-guard` hook (introduced in #1476) blocks all `write` calls to existing files with `"File already exists. Use edit tool instead."`. While the intent is to prevent accidental data loss, in practice this causes **significant token waste** for legitimate overwrite workflows.

**Real-world impact:**

- The LLM can spend minutes generating large output. The final `write` fails, then it must `read` the entire file and `edit` it.
- This often pushes the context window over the limit and forces `/compact`.
- The cycle repeats: write fail → read → edit → compaction.

**What happens:**

1. Agent generates content and calls `write` → **blocked**
2. Agent receives the error, then has to `read` the file to get its contents
3. Agent then uses `edit` with exact string replacement to achieve the same result
4. This 3-step process (write fail → read → edit) costs **~3x the tokens** of a single `write`

**The problem is amplified because:**

- OpenCode docs explicitly say `write` overwrites existing files (expectation mismatch)
- Many workflows legitimately re-generate or replace outputs
- Sub-agents often write part files that already exist, causing repeated retries
- The retry loop frequently triggers compaction

**Token cost comparison for overwriting a file:**

| Approach                             | Tool calls | Tokens                                 |
| ------------------------------------ | ---------- | -------------------------------------- |
| `write` (if allowed)                | 1          | ~N (content only)                      |
| `write` (blocked) → `read` → `edit` | 3          | ~3N (error + full file read + content) |

### Steps to Reproduce

1. Create any file: `write("test.txt", "first content")`
2. Try to overwrite: `write("test.txt", "second content")` → **Error: File already exists**
3. Even after `read("test.txt")`, the second `write` still fails
4. Agent must fall back to `edit` with full content replacement

### Expected Behavior

At minimum, the agent should not be surprised by the failure. Ideally, `write` to an existing file should work when intent is explicit.

### Actual Behavior

Every `write` to an existing file fails unconditionally (except `.sisyphus/`), regardless of whether the agent previously read the file. This leads to repeated token-wasting retry cycles.

---

## Proposed Fix (hook-level, preserves original motivation)

Continue blocking **blind overwrites**, but allow **intentional overwrites** to avoid the token-wasting loop.

**Recommended behavior:**

```
if !exists           -> allow
if .sisyphus/**      -> allow
if read_in_session   -> allow once (consume the read token, require re-read for next write)
if overwrite:true    -> allow (strip arg before tool sees it)
else                 -> block
```

### Key safety mechanisms

**1. Single-use read gate** — A `read` unlocks exactly ONE `write` to the same file. After the write, the agent must read again before writing again. This prevents the original bug (agent writing parts 1, 2, 3 sequentially and losing data).

**2. Cross-session invalidation** — When any session writes to a file, all OTHER sessions' read permissions for that file are revoked. This prevents race conditions where parallel sessions overwrite each other's changes without knowing the file was modified.

Example:
```
Session A reads file.txt     → read set A: {file.txt}
Session B reads & writes file.txt → invalidates A's read set → read set A: {}
Session A tries to write file.txt → BLOCKED (must re-read to see B's changes)
```

**3. Explicit override** — `overwrite: true` flag bypasses the read gate for cases where the agent genuinely wants to replace the file. The flag is stripped before the tool executes so it doesn't cause unknown-arg errors.

### Suggested tests (14 scenarios)

1. allows write to non-existing file
2. blocks write to existing file without read or overwrite flag
3. allows overwrite after `read` in the same session
4. blocks second consecutive write (read consumed after first write)
5. blocks overwrite if `read` occurred in a different session
6. allows overwrite with `overwrite: true` (boolean) and strips the arg
7. allows overwrite with `overwrite: "true"` (string) — LLMs sometimes output strings for booleans
8. cross-session invalidation: session B write invalidates session A read
9. `.sisyphus/**` exception allows writes without read gate
10. handles `filePath`, `path`, and `file_path` arg variants
11. handles relative and absolute paths consistently
12. resolves case-insensitive paths on macOS (via `realpathSync.native()`)
13. resolves symlinks (read via symlink, write via real path)
14. LRU eviction preserves recently active sessions under high concurrency

### Workaround (local OpenCode plugin)

Users can work around this today by disabling the hook and creating a local plugin with read-gate logic:

1. Disable the hook in `.opencode/oh-my-opencode.json`:
```json
{ "disabled_hooks": ["write-existing-file-guard"] }
```

2. Create `.opencode/plugins/write-guard.js` implementing the read-gate behavior described above. OpenCode supports local plugins with `tool.execute.before` hooks ([docs](https://opencode.ai/docs/plugins)).

See the implementation comment below for a tested, production-ready version.

---

### Doctor Output

```shell
 oMoMoMoMo Doctor 

 ⚠ 1 issue found:

   Affects: comment-checker hook
```

### Error Logs

```shell
Error: File already exists. Use edit tool instead.
```

### Configuration

```json
{}
```

### Additional Context

- Workaround: `"disabled_hooks": ["write-existing-file-guard"]` + local plugin in `.opencode/plugins/`
- Related: #1576 (Prometheus cannot rewrite plan file), #1363 (plan content overwritten)
- The hook was created to solve a real problem (accidental overwrites), but the current blanket block creates a different problem (token waste)
- **Two-layer protection**: OpenCode has its own built-in read-before-write validation independent of this hook. Even if the hook allows a write, OpenCode's internal validation may still block it (uses simple string comparison — no case normalization, no symlink resolution). This is a separate concern from the hook.
- oh-my-opencode version: 3.5.5

### Operating System

macOS

### OpenCode Version

1.2.5

gustavosmendes:
## Summary
- replace unconditional existing-file write block with a read-gated overwrite policy
- allow one overwrite after same-session read, then consume the permission
- support explicit `overwrite: true` / `overwrite: "true"` bypass and strip `overwrite` before tool execution
- canonicalize paths with `realpathSync.native()` to handle symlinks and case-insensitive platforms
- invalidate other sessions' read permissions after write and keep bounded session state with LRU eviction

## Tests
- added focused coverage for write/read gate, cross-session invalidation, overwrite flag handling, path variants, relative/absolute paths, symlinks, case behavior and LRU retention
- `bun test src/hooks/write-existing-file-guard/index.test.ts`
- `bun run build`

Closes #1871

---
## Summary by cubic
Replaced the unconditional existing-file write block with a read-gated overwrite policy. Addresses #1871 with one-time overwrite after same-session read, explicit bypass, robust path handling, and blocking writes outside the session directory.

- **New Features**
  - Read grants a one-time overwrite for the same session; consumed on write and invalidates other sessions.
  - Supports overwrite: true (boolean or "true") to bypass; flag is stripped before execution.
  - Canonicalizes paths with realpathSync.native, resolves symlinks, handles rel/abs and case-aware matching, and supports filePath/path/file_path args.
  - Always allows writes under .sisyphus/** and blocks writes outside the session directory.
  - Tracks up to 256 sessions with LRU capping; clears per-session permissions on session.deleted via plugin event forwarding.

<sup>Written for commit 73d9e1f84799d69424bc929999f0d018b724c16b. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/write-existing-file-guard/hook.ts
+++ b/src/hooks/write-existing-file-guard/hook.ts
@@ -1,7 +1,7 @@
 import type { Hooks, PluginInput } from "@opencode-ai/plugin"
 
 import { existsSync, realpathSync } from "fs"
-import { basename, dirname, isAbsolute, join, normalize, resolve, sep } from "path"
+import { basename, dirname, isAbsolute, join, normalize, relative, resolve, sep } from "path"
 
 import { log } from "../../shared"
 
@@ -13,9 +13,11 @@ type GuardArgs = {
 }
 
 const MAX_TRACKED_SESSIONS = 256
+export const MAX_TRACKED_PATHS_PER_SESSION = 1024
+const OUTSIDE_SESSION_MESSAGE = "Path must be inside session directory."
 
 function asRecord(value: unknown): Record<string, unknown> | undefined {
-  if (!value || typeof value !== "object") {
+  if (!value || typeof value !== "object" || Array.isArray(value)) {
     return undefined
   }
 
@@ -28,6 +30,11 @@ function getPathFromArgs(args: GuardArgs | undefined): string | undefined {
 
 function resolveInputPath(ctx: PluginInput, inputPath: string): string {
   return normalize(isAbsolute(inputPath) ? inputPath : resolve(ctx.directory, inputPath))
+}
+
+function isPathInsideDirectory(pathToCheck: string, directory: string): boolean {
+  const relativePath = relative(directory, pathToCheck)
+  return relativePath === "" || (!relativePath.startsWith("..") && !isAbsolute(relativePath))
 }
 
 function toCanonicalPath(absolutePath: string): string {
@@ -106,9 +113,25 @@ export function createWriteExistingFileGuardHook(ctx: PluginInput): Hooks {
     return readSet
   }
 
+  const trimSessionReadSet = (readSet: Set<string>): void => {
+    while (readSet.size > MAX_TRACKED_PATHS_PER_SESSION) {
+      const oldestPath = readSet.values().next().value
+      if (!oldestPath) {
+        return
+      }
+
+      readSet.delete(oldestPath)
+    }
+  }
+
   const registerReadPermission = (sessionID: string, canonicalPath: string): void => {
     const readSet = ensureSessionReadSet(sessionID)
+    if (readSet.has(canonicalPath)) {
+      readSet.delete(canonicalPath)
+    }
+
     readSet.add(canonicalPath)
+    trimSessionReadSet(readSet)
   }
 
   const consumeReadPermission = (sessionID: string, canonicalPath: string): boolean => {
@@ -128,9 +151,7 @@ export function createWriteExistingFileGuardHook(ctx: PluginInput): Hooks {
         continue
       }
 
-      if (readSet.delete(canonicalPath)) {
-        touchSession(sessionID)
-      }
+      readSet.delete(canonicalPath)
     }
   }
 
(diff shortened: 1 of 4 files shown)


gustavosmendes:

PR #1944, opened 2026-02-18
fix(tmux): allow initial pane spawn when no agent panes exist

pperezrubio:
## Summary

- Fix `decideSpawnActions` blocking pane creation when 0 agent panes exist

## Problem

When `currentCount === 0` (no agent panes), `mainPane.width === windowWidth` because the main pane occupies the entire window. This means:

```
agentAreaWidth = max(0, windowWidth - mainPane.width - DIVIDER_SIZE) = 0
```

The early return at line 35 fires (`0 < minPaneWidth` → true), returning `canSpawn: false` with "window too small" — even though the window is 227+ columns wide.

The `currentCount === 0` handler at line 48, which correctly creates a `virtualMainPane` with full window width and uses `canSplitPane()`, is **never reached**.

## Fix

Add `&& currentCount > 0` guard to the early return so it only applies when agent panes already exist. When `currentCount === 0`, execution falls through to the dedicated initial-split handler which works correctly.

```diff
- if (agentAreaWidth < minPaneWidth) {
+ if (agentAreaWidth < minPaneWidth && currentCount > 0) {
```

## Testing

Verified with A/B testing:

| Version | Window | Result |
|---------|--------|--------|
| v3.7.3 (upstream, no patch) | 227x58, 0 agent panes | `canSpawn: false`, "window too small" ❌ |
| v3.7.3 + this fix | 227x58, 0 agent panes | `canSpawn: true`, pane spawned ✅ |

Log evidence (upstream):
```
[tmux-session-manager] window state queried {"windowWidth":227,"mainPane":"%0","agentPaneCount":0}
[tmux-session-manager] spawn decision {"canSpawn":false,"reason":"window too small for agent panes: 227x58"}
```

Log evidence (with fix):
```
[tmux-session-manager] spawn decision {"canSpawn":true,"actionCount":1,"actions":[{"type":"spawn"}]}
[spawnTmuxPane] all checks passed, spawning...
[tmux-session-manager] pane spawned and tracked {"paneId":"%1","sessionReady":true}
```

Fixes #1939

---
## Summary by cubic
Fixes a bug in tmux pane sizing that blocked the very first agent pane from spawning. Initial splits now work on wide windows instead of falsely reporting “window too small.”

- **Bug Fixes**
  - Adjusted decideSpawnActions to only enforce the min-width early return when currentCount > 0, letting the zero-pane case reach the initial-split handler.
[...]

PR #1983, opened 2026-02-20
fix(background-output): stop defaulting full_session=true for running tasks

Pantoria:
## Summary

- `background_output` previously defaulted `full_session=true` when the queried task was still running/pending, returning the **entire session transcript** on every call
- When the parent agent had no other immediate work after launching background tasks, it would poll `background_output` in a tight loop — each iteration dumping thousands of tokens of full transcript into the conversation history
- This caused massive token waste that compounded with each poll (full transcript in tool output → becomes part of context → next poll adds another full transcript)

## Fix

Change the default from `args.full_session ?? isActive` to `args.full_session ?? false`.

Now running-task checks return a compact status table (~200 tokens) instead of the full session transcript. Callers who actually need the full transcript can still pass `full_session=true` explicitly.

## What stays the same

- Completed tasks still return full results via `formatTaskResult` (unaffected, separate code path)
- `full_session=true` still works when explicitly passed
- All other `background_output` behavior is unchanged

## Tests

- Updated existing test to reflect the new default behavior
- Added new test confirming `full_session=true` still works when explicitly set on running tasks
- Full suite: 3060 pass, 0 fail

---
## Summary by cubic
Stop returning full session transcripts by default for running background tasks. This prevents repeated multi‑thousand‑token dumps on each poll and eliminates compounding token waste.

- **Bug Fixes**
  - Default changed from args.full_session ?? isActive to args.full_session ?? false.
  - Running tasks now return a compact status summary; completed tasks still return full results.
  - Explicit opt‑in preserved via full_session=true when needed.
  - Updated tests to cover new default and explicit full_session behavior.

<sup>Written for commit d556937c8e48775ae665ab7a97538e744f1459b8. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/background-task/create-background-output.ts
+++ b/src/tools/background-task/create-background-output.ts
@@ -78,7 +78,7 @@ export function createBackgroundOutput(manager: BackgroundOutputManager, client:
         }
 
         const isActive = task.status === "pending" || task.status === "running"
-        const fullSession = args.full_session ?? isActive
+        const fullSession = args.full_session ?? false
         const includeThinking = isActive || (args.include_thinking ?? false)
         const includeToolResults = isActive || (args.include_tool_results ?? false)
 
--- a/src/tools/background-task/tools.test.ts
+++ b/src/tools/background-task/tools.test.ts
@@ -232,7 +232,7 @@ describe("background_output full_session", () => {
     expect(output).toContain("Has more: true")
   })
 
-  test("defaults to full_session when task is running", async () => {
+  test("defaults to compact status when task is running", async () => {
     // #given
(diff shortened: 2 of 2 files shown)


sisyphus-dev-ai:
hey, just a heads up - this PR targets \`master\` instead of \`dev\`. the actual change looks small but the diff is showing +108k/-33k lines because of the master-dev gap.

could you update the base branch to \`dev\`? you can do it from the PR page (edit button next to the base branch). that should bring the diff down to just the relevant changes.

same thing happened with #1944 recently - easy fix once the base is corrected.

Pantoria:

Issue #1880, opened 2026-02-16
[Bug]: Run CLI command edge cases — silent event stream drop and no-meaningful-work gate


code-yeongyu:
## Summary

While investigating #1876, we reviewed the `run` command exit flow end-to-end (fixed in #1828). The primary hang is resolved, but two edge cases remain that could cause delayed or failed exits in specific environments.

## Edge Case 1: Event Stream Drops Silently in CI

**Location**: `src/cli/run/event-handlers.ts` → `handleSessionIdle`, `handleSessionStatus`

`mainSessionIdle` is only set to `true` by `session.idle` or `session.status` events. If the SSE event stream drops without a clean close (common in CI environments like GitHub Actions with network instability), no more events arrive → `mainSessionIdle` stays `false` → `pollForCompletion` loops indefinitely until the 600s default timeout.

**Impact**: Process appears to "hang" for up to 10 minutes in CI before the timeout kicks in. Not a true hang, but a very long delay.

**Potential fix**: Add a "no events received for N seconds" watchdog in `pollForCompletion`. If no events arrive for e.g. 30 seconds while waiting, check session status via a direct API call (`client.session.status()`) rather than relying solely on events.

## Edge Case 2: Main Session Produces No Meaningful Work

**Location**: `src/cli/run/poll-for-completion.ts:64-67`

If the main session immediately delegates to child sessions without producing any assistant text, tool calls, or tool results on the main session itself, `hasReceivedMeaningfulWork` remains `false`. The poll loop never reaches `checkCompletionConditions()`.

**Impact**: Process waits until the 600s timeout. Low probability since `promptAsync` typically triggers at least one assistant message, but theoretically possible with certain agent configurations.

**Potential fix**: Consider also counting child session activity toward `hasReceivedMeaningfulWork`, or add a secondary timeout (e.g. 60s) after which `hasReceivedMeaningfulWork` is assumed `true` if the session exists and has children.

## Context

- Original bug: #1825 (closed by #1828)
- Investigation triggered by CloudWaddie's comment on #1876
- Related: CloudWaddie/actions-agent uses `stdbuf -oL -eL` wrapper and previously hit exit code 143 (SIGTERM)

code-yeongyu:
Closing — this was created based on a misinterpretation of CloudWaddie's comment on #1876. They were confirming the fix works, not reporting a remaining bug. The run exit issue was fully resolved by PR #1828.

PR #1988, opened 2026-02-20
fix(run): prevent exit code 130 timeout in CI environments

### LINKED ISSUE #1934: [Bug]: Exit code 130 timeout still occurring - Issue #1880 not fully resolved
## Summary

Issue #1880 was closed but the timeout issue is still occurring in CI environments (GitHub Actions).

## Description

We are using oh-my-opencode in our GitHub Actions workflow (cloudwaddie/actions-agent) and consistently hitting exit code 130 (timeout) after 10 minutes.

### Failing Workflow

- Repository: https://github.com/CloudWaddie/actions-agent
- Workflow run: https://github.com/CloudWaddie/actions-agent/actions/runs/22085847518/job/63820196148
- Issue: https://github.com/CloudWaddie/actions-agent/issues/70

### Symptoms

1. oh-my-opencode connects successfully
2. Session is created, prompt is sent
3. Server sends heartbeats for exactly 10 minutes
4. Then fails with "Timeout reached. Aborting..."
5. Exit code 130

This matches exactly the symptoms described in #1880:
- Event stream drops silently (or no meaningful work on main session)
- Poll loop waits for full 600s timeout before exiting

## Expected Behavior

The fixes described in #1880 should prevent this 10-minute hang:
1. Add a "no events received for N seconds" watchdog
2. Add a secondary timeout after which hasReceivedMeaningfulWork is assumed true
3. Check session status via direct API call when events stop

## Evidence

From workflow logs:
```
Timeout reached. Aborting...
Process completed with exit code 130.
```

The agent was processing a simple task (updating a YAML config file) but never completed - just sent heartbeats for 10 minutes.

## Additional Context

- This is in GitHub Actions (Ubuntu 24.04 runner)
- Using free OpenCode model (opencode/minimax-m2.5-free)
- Workflow uses `stdbuf -oL -eL bun run oh-my-opencode/dist/cli/index.js run "$PROMPT"`

## Request

Please re-open #1880 or create a new issue to track this. The workarounds described in #1880 do not appear to have been fully implemented yet.

CloudWaddie:
## Summary

Implements fixes from issue #1880 and #1934 to prevent exit code 130 timeout in CI environments (GitHub Actions).

### Problem

When the SSE event stream drops silently without a clean close (common in CI with network instability), the poll loop waits for the full 600s timeout before exiting with code 130.

### Solution

1. **Add `lastEventTimestamp` to EventState** - Tracks when events were last received for watchdog detection

2. **Add event watchdog (30s)** - If no events received for 30 seconds, verify session status via direct API call instead of waiting indefinitely

3. **Add secondary timeout (60s)** - After 60 seconds without meaningful work events, check for active children/todos via API and assume work is in progress if found

This prevents the poll loop from waiting for the full 600s timeout when:
- Event stream drops silently (network instability in CI)
- Main session delegates to children without producing meaningful work on main session

### Files Changed

- `src/cli/run/event-state.ts` - Added `lastEventTimestamp` field
- `src/cli/run/event-stream-processor.ts` - Update timestamp on each event
- `src/cli/run/poll-for-completion.ts` - Watchdog and secondary timeout logic

### Related Issues

- Fixes #1934
- Related to #1880

---
## Summary by cubic
Prevents exit code 130 timeouts in CI by detecting stalled event streams and confirming session state directly. Fixes #1934 and aligns with #1880.

- **Bug Fixes**
  - Track lastEventTimestamp in EventState.
  - Event watchdog (30s): verify session status via API and reuse the result to avoid duplicate calls.
  - Secondary timeout (60s): check children/todos once; fix operator precedence in hasActiveWork.
  - Avoids waiting the full 600s when SSE drops or work runs in child sessions.

<sup>Written for commit 6cc7535a804fd365ba9848522971037afa03876b. Summary will update on new commits.</sup>

### DIFF
--- a/src/cli/run/event-state.ts
+++ b/src/cli/run/event-state.ts
@@ -7,6 +7,8 @@ export interface EventState {
   currentTool: string | null
   /** Set to true when the main session has produced meaningful work (text, tool call, or tool result) */
   hasReceivedMeaningfulWork: boolean
+  /** Timestamp of the last received event (for watchdog detection) */
+  lastEventTimestamp: number
   /** Count of assistant messages for the main session */
   messageCount: number
   /** Current agent name from the latest assistant message */
@@ -54,6 +56,7 @@ export function createEventState(): EventState {
     lastPartText: "",
     currentTool: null,
     hasReceivedMeaningfulWork: false,
+    lastEventTimestamp: Date.now(),
     messageCount: 0,
     currentAgent: null,
     currentModel: null,
--- a/src/cli/run/event-stream-processor.ts
+++ b/src/cli/run/event-stream-processor.ts
@@ -35,6 +35,9 @@ export async function processEvents(
         logEventVerbose(ctx, payload)
       }
 
+      // Update last event timestamp for watchdog detection
+      state.lastEventTimestamp = Date.now()
+
       handleSessionError(ctx, payload, state)
       handleSessionIdle(ctx, payload, state)
       handleSessionStatus(ctx, payload, state)
--- a/src/cli/run/poll-for-completion.ts
+++ b/src/cli/run/poll-for-completion.ts
@@ -8,11 +8,15 @@ const DEFAULT_POLL_INTERVAL_MS = 500
 const DEFAULT_REQUIRED_CONSECUTIVE = 1
 const ERROR_GRACE_CYCLES = 3
 const MIN_STABILIZATION_MS = 1_000
+const DEFAULT_EVENT_WATCHDOG_MS = 30_000 // 30 seconds
+const DEFAULT_SECONDARY_MEANINGFUL_WORK_TIMEOUT_MS = 60_000 // 60 seconds
 
 export interface PollOptions {
   pollIntervalMs?: number
   requiredConsecutive?: number
   minStabilizationMs?: number
+  eventWatchdogMs?: number
+  secondaryMeaningfulWorkTimeoutMs?: number
 }
 
 export async function pollForCompletion(
@@ -28,6 +32,11 @@ export async function pollForCompletion(
     options.minStabilizationMs ?? MIN_STABILIZATION_MS
   const minStabilizationMs =
     rawMinStabilizationMs > 0 ? rawMinStabilizationMs : MIN_STABILIZATION_MS
+  const eventWatchdogMs =
+    options.eventWatchdogMs ?? DEFAULT_EVENT_WATCHDOG_MS
+  const secondaryMeaningfulWorkTimeoutMs =
+    options.secondaryMeaningfulWorkTimeoutMs ??
+    DEFAULT_SECONDARY_MEANINGFUL_WORK_TIMEOUT_MS
   let consecutiveCompleteChecks = 0
   let errorCycleCount = 0
   let firstWorkTimestamp: number | null = null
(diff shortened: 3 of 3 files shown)


CloudWaddie:

code-yeongyu:
@sisyphus-dev-ai review this

sisyphus-dev-ai:
## Code Review: PR #1988

Thanks for addressing the exit code 130 timeout issue in CI environments @CloudWaddie! The overall approach is sound and implements the safeguards proposed in #1880.

### ✅ What looks good

1. **Event timestamp tracking** - Clean addition to `EventState` interface and proper initialization with `Date.now()`
2. **Watchdog mechanism** - The 30s watchdog for detecting stalled event streams is a pragmatic solution
3. **Build passes** - Verified the PR builds successfully
4. **Addresses real user pain** - This fixes a genuine CI workflow issue affecting multiple repositories

### 🔧 Issues to fix

#### 1. **Logic bug in `hasActiveWork` boolean expression** (Critical)

```typescript
const hasActiveWork =
  Array.isArray(children)
    ? children.length > 0
    : false || Array.isArray(todos)
    ? todos.some(...)
    : false
```

Due to ternary operator precedence, this doesn't work as intended. The expression is parsed as:
```
(Array.isArray(children) ? children.length > 0 : (false || Array.isArray(todos))) ? todos.some(...) : false
```

So if `children` is an array but empty, it evaluates `false || Array.isArray(todos)` = `true`, then uses that as the condition for `todos.some(...)`.

**Fix:**
```typescript
const hasActiveChildren = Array.isArray(children) && children.length > 0
const hasActiveTodos = Array.isArray(todos) && todos.some(
  (t: unknown) =>
    (t as { status?: string })?.status !== "completed" &&
    (t as { status?: string })?.status !== "cancelled"
)
const hasActiveWork = hasActiveChildren || hasActiveTodos
```

#### 2. **Unnecessary null check** (Minor)

```typescript
if (eventState.lastEventTimestamp !== null) {
```

`lastEventTimestamp` is typed as `number` (not `number | null`) and initialized with `Date.now()`, so this check is always true. Consider removing it for clarity, or if you want to support an "uninitialized" state, change the type to `number | null` and initialize to `null`.

### 💭 Suggestions (optional)

1. **Consider making timeouts configurable** - The 30s watchdog and 60s secondary timeout could be exposed as configuration options (low priority - current defaults seem reasonable)

2. **Log messages** - The yellow warning logs are helpful for debugging. Consider also logging when the watchdog triggers and finds the session is still `busy` (currently only logs the initial "No events for Xs" message)

### Summary

The PR is **close to merge-ready** but the boolean logic bug needs to be fixed first as it affects the core functionality. Once addressed, this will be a solid improvement for CI reliability.

code-yeongyu (changes requested):
## Review

Thanks for tackling the CI timeout issue — the watchdog + secondary timeout approach makes sense.

Found a few things that should be addressed before merge:

### 1. Operator precedence bug in `hasActiveWork`

```ts
const hasActiveWork =
  Array.isArray(children)
    ? children.length > 0
    : false || Array.isArray(todos)
    ? todos.some(...)
    : false
```

Due to how the ternary operator chains, `children.length > 0` result gets swallowed. This effectively ignores whether children exist and only evaluates the todos branch.

Should be something like:

```ts
const hasActiveChildren = Array.isArray(children) && children.length > 0
const hasActiveTodos = Array.isArray(todos) && todos.some(
  (t: unknown) =>
    (t as { status?: string })?.status !== "completed" &&
    (t as { status?: string })?.status !== "cancelled"
)
const hasActiveWork = hasActiveChildren || hasActiveTodos
```

### 2. Duplicate `getMainSessionStatus` calls

The watchdog block calls `getMainSessionStatus(ctx)`, and then immediately after the block, `getMainSessionStatus(ctx)` is called again unconditionally. That's two API calls per poll cycle when the watchdog triggers. Consider reusing the result from the watchdog check.

### 3. Secondary timeout fires every poll cycle after 60s

Once `Date.now() - pollStartTimestamp > secondaryMeaningfulWorkTimeoutMs`, the children + todos API calls happen every 500ms poll cycle. This is a lot of unnecessary API traffic. Should either:
- Set a flag after the first check so it only runs once
- Or use a separate timestamp to throttle (e.g. check every 30s)

---

The core idea is solid — just needs these fixes. Looking forward to the updated version!

CloudWaddie:
Thanks for the review! I've addressed all three issues:

1. **Fixed operator precedence bug in ** - Refactored to use clear separate variables (, ) instead of the chained ternary that was being parsed incorrectly.

2. **Fixed duplicate  calls** - Now reuse the result from the watchdog check instead of calling it again unconditionally.

3. **Fixed secondary timeout firing every poll cycle** - Added  flag so the children + todos API calls only happen once after 60s, not every 500ms poll cycle.

Build passes. Ready for re-review!

CloudWaddie:
@code-yeongyu updated

code-yeongyu:
@sisyphus-dev-ai ulw review

CloudWaddie:
@code-yeongyu https://github.com/code-yeongyu/oh-my-opencode/actions/runs/22214613852/job/64255680917
it failed... separate issue i think

PR #1901, opened 2026-02-16
fix(ralph-loop): keep LLM in smart zone with fresh context per iteration

edxeth:
## The Problem: Context Window Overflow

The original ralph-loop implementation kept all iterations in a **single session**, causing the context window to fill up over time. This is exactly the problem described in ["Why the Anthropic Ralph plugin sucks"](https://www.aihero.dev/why-the-anthropic-ralph-plugin-sucks):

> As LLMs receive more tokens, the relationships between tokens scale quadratically. Each additional token makes it harder for the model to process information and make good decisions.

Every LLM has a **smart zone** (first ~40% of context) and a **dumb zone** (remaining ~60%). The original implementation guaranteed you'd enter the dumb zone after just a few iterations:

| Iteration | Context Usage | Zone |
|-----------|---------------|------|
| 1 | ~20% | Smart |
| 2 | ~35% | Smart |
| 3 | ~50% | Dumb |
| 4+ | 60%+ | Dumb |

## The Solution: Fresh Context Per Iteration

This PR adds a `default_strategy` config option with **"reset" as the new default**:

| Strategy | Behavior | Use Case |
|----------|----------|----------|
| `reset` | Creates NEW session each iteration | Long-running loops, production work |
| `continue` | Keeps same session | Short tasks, debugging |

With `reset`, each iteration starts with a **fresh context window**. The LLM always operates in the smart zone:

```
Iteration 1: Fresh context → Smart zone → Commit
Iteration 2: Fresh context → Smart zone → Commit  
Iteration 3: Fresh context → Smart zone → Commit
...
```

This matches how the original bash-loop Ralph works:
```bash
while :; do cat PROMPT.md | claude-code ; done
```
Each invocation = fresh context.

### Core Feature
- **New `default_strategy` config** with `reset` (default) and `continue` options
- **`--strategy` flag** for per-loop override: `/ralph-loop "task" --strategy=continue`
- **Strategy stored in loop state** - persists across iterations

### Bug Fixes
- **Skip `todo-continuation-enforcer` when ralph-loop is active** - prevents hook conflicts
- **Conditional prompt prefix** - `reset` allows keyword detection (new session needs mode injections), `continue` skips it (mode already applied)

### Implementation
- TUI session switch via raw API call to `/tui/select-session`
- Strategy parsed from command arguments in both `chat.message` and `PreToolUse` handlers

## Usage

**Config (default for all loops):**
```json
{
  "ralph_loop": {
    "enabled": true,
    "default_strategy": "reset"
[...]

PR #2007, opened 2026-02-20
feat(ralph-loop): add strategy option for fresh context per iteration

code-yeongyu:
## Summary
- Closes #1901
- Adds `default_strategy: "reset" | "continue"` to `RalphLoopConfigSchema` with default `"continue"` for backward compatibility
- Adds `--strategy=reset|continue` command parsing for `/ralph-loop` and `/ulw-loop`
- Implements `reset` strategy to create a fresh session per iteration, inject continuation prompt, switch TUI session, and persist new `session_id`
- Preserves existing `continue` behavior in the same session

## Problem
`ralph-loop` currently runs all iterations in one session, so context accumulates over time and iteration quality drops after several loops.

## Solution
Introduced loop strategy controls in config + per-command flags. `continue` keeps backward-compatible behavior. `reset` creates a fresh session each iteration while inheriting the previous session's context settings for the continuation prompt.

## Usage
```json
{
  "ralph_loop": {
    "default_strategy": "reset"
  }
}
```

```text
/ralph-loop "Build feature X" --strategy=reset
```

## Verification
- `bun test src/hooks/ralph-loop/`
- `bun run build`
- `bun test`

---
## Summary by cubic
Adds a strategy option to ralph-loop so each iteration can start with fresh context when needed. Introduces a “reset” mode that creates a new child session per iteration; “continue” remains the default. Addresses #1901.

- **New Features**
  - Config: ralph_loop.default_strategy = "reset" | "continue" (default "continue"; required in schema).
  - Commands: --strategy=reset|continue for /ralph-loop and /ulw-loop via a shared parser (handles quoted values and flag-only input).
  - Reset mode: creates a child session per iteration, inherits model/agent/tools from the source session (API with on-disk fallback), injects the continuation prompt in the new session, persists new session_id and strategy, and auto-selects it in the TUI.

- **Bug Fixes**
  - Reliable tool/context inheritance using sourceSessionID with on-disk fallback.
  - Bound tui.selectSession to the SDK context and guarded with shared isRecord.
  - Fixed quoted argument parsing for prompt and completion-promise; unified parsing in chat and tool handlers; restored missing exports and removed duplicates.

<sup>Written for commit 940e49b44c3b91d95b57fb76d66a0646c66f7d03. Summary will update on new commits.</sup>

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -3343,11 +3343,20 @@
         },
         "state_dir": {
           "type": "string"
+        },
+        "default_strategy": {
+          "default": "continue",
+          "type": "string",
+          "enum": [
+            "reset",
+            "continue"
+          ]
         }
       },
       "required": [
         "enabled",
-        "default_max_iterations"
+        "default_max_iterations",
+        "default_strategy"
       ],
       "additionalProperties": false
     },
--- a/src/config/schema/ralph-loop.ts
+++ b/src/config/schema/ralph-loop.ts
@@ -7,6 +7,7 @@ export const RalphLoopConfigSchema = z.object({
   default_max_iterations: z.number().min(1).max(1000).default(100),
   /** Custom state file directory relative to project root (default: .opencode/) */
   state_dir: z.string().optional(),
+  default_strategy: z.enum(["reset", "continue"]).default("continue"),
 })
 
 export type RalphLoopConfig = z.infer<typeof RalphLoopConfigSchema>
--- a/src/features/builtin-commands/commands.ts
+++ b/src/features/builtin-commands/commands.ts
@@ -28,7 +28,7 @@ ${RALPH_LOOP_TEMPLATE}
 <user-task>
 $ARGUMENTS
 </user-task>`,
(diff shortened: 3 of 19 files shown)


code-yeongyu:
@cubic-dev-ai please re-review

code-yeongyu:
@cubic-dev-ai The previous review was on a now-reverted commit. The current HEAD already has `afterEach(() => { fakeTimeouts.restore() })` at line 112 which ensures cleanup even on assertion failure. Please re-review the current state.

code-yeongyu:
@cubic-dev-ai Please re-review the current HEAD. The `executor.test.ts` already has `afterEach(() => { fakeTimeouts.restore() })` at line 112, covering cleanup even when assertions fail. The previous review was on an older commit.

code-yeongyu:
@cubic-dev-ai Please re-review. Fixed the issues identified:
- `session-reset-strategy.ts`: removed duplicate local `isRecord`, now imports from `../../shared/record-type-guard`
- `command-arguments.ts`: fixed prompt regex to use backreference (`\1`) so same quote type is matched (prevents `"It's done"` from stopping at apostrophe); fixed `--completion-promise` regex to support multi-word quoted values

PR #2033, opened 2026-02-22
fix: sanitize lone Unicode surrogates in outgoing API request bodies

codeg-dev:
## Problem

Anthropic's API backend (using `yyjson`/`serde_json`) rejects JSON payloads that contain lone Unicode surrogates — even when they appear as JSON escape sequences like `\uD82C` rather than as raw UTF-16 code units.

OpenCode's existing `sanitizeSurrogates()` only handles **literal** surrogate code units (checked via `String.prototype.isWellFormed()`). It does **not** handle the case where `JSON.stringify()` has already converted a lone surrogate into a `\uDxxx` ASCII escape sequence. Because the string at that point contains only ASCII characters, `isWellFormed()` returns `true` — and the escape sequence passes through unchecked.

The result: the Anthropic API returns `"no low surrogate in string"` and the entire session hangs.

**Root cause reference**: [`anthropics/claude-code#1709`](https://github.com/anthropics/claude-code/issues/1709), [`#1832`](https://github.com/anthropics/claude-code/issues/1832)

## Solution (three-layer defense)

| Layer | Where | What |
|---|---|---|
| 1 | `tool.execute.after` hook | Sanitize tool output strings at the source |
| 2 | `experimental.chat.messages.transform` hook | Belt-and-suspenders over all message content |
| 3 | `globalThis.fetch` monkey-patch | Last-resort: sanitizes **all** outgoing request bodies, including system prompts and tool definitions injected by the SDK that bypass plugin hooks |

Layer 3 uses `patchFetchForSurrogates()`, which is called at plugin module load time (before any hook fires). A `Symbol.for` marker ensures the patch is idempotent.

`fixJsonSurrogateEscapes()` handles the `JSON.stringify`-escaped case: it walks the string as a JSON lexer, finds `\uDxxx` sequences, and replaces them with the Unicode replacement character.

## Changes

- `src/shared/sanitize-surrogates.ts` — `sanitizeSurrogates()` utility
- `src/shared/patch-fetch-surrogates.ts` — `globalThis.fetch` monkey-patch
- `src/shared/fix-json-surrogate-escapes.ts` — JSON-escape-level surrogate fix
- `src/plugin/messages-transform.ts` — hook: sanitize all message content
- `src/plugin/tool-execute-after.ts` — hook: sanitize tool outputs at source
- `src/plugin-interface.ts` — call `patchFetchForSurrogates()` at load time
- Tests for all new modules (26/26 passing)

## Test

```
bun test
```

All 26 tests pass.

---

> **Note**: This fix was developed with AI assistance (Claude Code / claude-sonnet-4-6). The bug was encountered in production when a Korean character was encoded as a lone surrogate in a file path.

---
## Summary by cubic
Prevents Anthropic API errors by sanitizing lone Unicode surrogates in all outgoing JSON, including JSON-escaped forms like \uD800, to stop “no low surrogate in string” failures and hangs. Adds in-place cleanup for messages and tool outputs, including MCP content[].

- **Bug Fixes**
  - Patch global fetch at module load to sanitize all string bodies (including fetch(Request)); idempotent; preserves valid pairs.
  - Sanitize messages before/after transforms and mutate arrays in place via splice; handle MCP content[] alongside standard tool outputs.
  - Replace lone \uD800–\uDFFF JSON escapes with \uFFFD while keeping valid surrogate pairs.

<sup>Written for commit 73ac779308aae5efc13980eb80272103c39effdb. Summary will update on new commits.</sup>


codeg-dev:

codeg-dev:
recheck

PR #2161, opened 2026-02-26
fix(agents): remove duplicate remapped agent keys

acamq:
## Summary
- Remove duplicate agent entries by emitting only display-name keys from the agent key remapper.
- Update remapper tests to assert lowercase built-in keys are not duplicated alongside display names.
- Preserve behavior for unknown/custom agent keys.

## Verification
- bun test src/plugin-handlers/agent-key-remapper.test.ts
- bun run typecheck
- bun run build

---
## Summary by cubic
Removes duplicate agent entries by emitting only display-name keys for known agents. This makes each core agent map to a single canonical key and prevents lowercase duplicates.

- **Bug Fixes**
  - Emit display-name keys only for core agents; drop lowercase duplicates.
  - Preserve unknown/custom agent keys unchanged.
  - Update tests to assert no lowercase duplicates.

<sup>Written for commit d6dd54867a60bf6aa95643f9d4d658eb6f3006b6. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin-handlers/agent-key-remapper.test.ts
+++ b/src/plugin-handlers/agent-key-remapper.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect } from "bun:test"
 import { remapAgentKeysToDisplayNames } from "./agent-key-remapper"
 
 describe("remapAgentKeysToDisplayNames", () => {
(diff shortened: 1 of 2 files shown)


KNN-07:
LGTM

PR #2184, opened 2026-02-27
fix: remove config-context console.warn that leaks into TUI textbox

### LINKED ISSUE #2183: config-context warning leaks into TUI textbox
## Problem

The warning `[config-context] getConfigContext() called before initConfigContext(); defaulting to CLI paths.` occasionally appears directly in the TUI textbox, disrupting the user experience.

## Root Cause

In `src/cli/config-manager/config-context.ts`, `getConfigContext()` uses `console.warn()` when called before `initConfigContext()` has completed:

```ts
export function getConfigContext(): ConfigContext {
  if (!configContext) {
    if (process.env.NODE_ENV !== "production") {
      console.warn("[config-context] getConfigContext() called before initConfigContext(); defaulting to CLI paths.")
    }
    // ...falls back to defaults
  }
  return configContext
}
```

`initConfigContext` is called from the async `findOpenCodeBinaryWithVersion()` in `opencode-binary.ts`, which spawns a subprocess. Other modules that call `getConfigDir()` / `getConfigJson()` / `getOmoConfigPath()` can resolve synchronously before this async init finishes, triggering the warning.

Since `console.warn` writes to stderr, and the TUI captures stderr output, the warning renders inside the textbox instead of being silently logged.

## Expected Behavior

The warning should not be visible in the TUI. The fallback itself is fine (it defaults to `{ binary: "opencode", version: null }` which produces standard CLI paths), but the `console.warn` side-effect leaks into the UI.

## Possible Fixes

- Remove the `console.warn` entirely — the fallback is safe and intentional
- Replace `console.warn` with a proper structured logger that doesn't write to stderr when TUI is active
- Ensure `initConfigContext` is awaited before any code path can call `getConfigContext`

mertyldrm:
## Summary

- Removes `console.warn` in `getConfigContext()` that renders inside the TUI textbox during startup
- The fallback behavior (defaulting to standard CLI paths) is preserved — only the noisy warning is removed

## Problem

`initConfigContext()` runs async (spawns `opencode --version` subprocess), but modules calling `getConfigDir()` / `getConfigJson()` can resolve synchronously before it completes. The `console.warn` writes to stderr, which the TUI captures, causing the warning to appear in the user's input textbox.

Fixes #2183

---
## Summary by cubic
Removed console.warn in getConfigContext() that wrote to stderr and leaked into the TUI input box during startup. Behavior unchanged: if initConfigContext() hasn’t completed, we default to standard CLI paths.

<sup>Written for commit 83c024dd663635c004eb1aabeb1258a4af578690. Summary will update on new commits.</sup>

### DIFF
--- a/src/cli/config-manager/config-context.ts
+++ b/src/cli/config-manager/config-context.ts
@@ -19,9 +19,6 @@ export function initConfigContext(binary: OpenCodeBinaryType, version: string | 
 
 export function getConfigContext(): ConfigContext {
   if (!configContext) {
-    if (process.env.NODE_ENV !== "production") {
-      console.warn("[config-context] getConfigContext() called before initConfigContext(); defaulting to CLI paths.")
-    }
     const paths = getOpenCodeConfigPaths({ binary: "opencode", version: null })
     configContext = { binary: "opencode", version: null, paths }
   }


mertyldrm:

PR #1908, opened 2026-02-16
feat: allow custom subagents to delegate tasks via task()

Stranmor:
## Summary

Custom subagents defined in `oh-my-opencode.json` cannot use `task()` to delegate work to other agents. This PR fixes three independent issues that collectively prevent the pipeline from working.

## Problem

When a user defines custom agents (e.g., `critic`, `breaker`, `type-sentinel`) in `oh-my-opencode.json`, those agents:

1. **Cannot call `task()`** — `buildSubagentTools()` hardcodes `task: false` for background tasks and only allows `task` for plan-family agents in sync flows
2. **Are never registered with opencode** — `applyAgentConfig()` passes custom agents as overrides to `createBuiltinAgents()`, which only processes known builtin names
3. **Are blocked by global permission deny** — `applyToolConfig()` sets `config.permission.task = "deny"` globally, then only whitelists hardcoded agents

## Changes

**Commit 1** (`agent-tool-restrictions.ts`): Check `AGENT_RESTRICTIONS` instead of hardcoding `task: false`. 4 code paths fixed.

**Commit 2** (`agent-config-handler.ts`, `tool-config-handler.ts`): Register custom agents from config. Grant `task: "allow"` to non-restricted agents.

PR #2262, opened 2026-03-03
feat: support file:// URIs in agent prompt field

Stranmor:
## Problem

Currently, `prompt_append` supports `file://` URIs via `resolvePromptAppend()`, allowing users to load prompt content from files. However, the `prompt` field does **not** support `file://` URIs — it goes through `deepMerge()` as a literal string.

This means users can only **append** to a built-in agent's prompt using a file. They cannot **fully replace** a built-in agent's prompt with file-based content. This is a significant limitation for power users who need complete control over agent system prompts.

## Solution

Add a single check in `mergeAgentConfig()` to resolve `file://` URIs in the `prompt` field, reusing the existing `resolvePromptAppend()` function. The check runs **after** `deepMerge()` but **before** `prompt_append` processing, so:

1. `prompt: "file://~/.config/opencode/prompts/my-agent.md"` → file contents become the prompt
2. `prompt_append` still works on top of the resolved prompt (composable)

## Changes

- **`src/agents/builtin-agents/agent-overrides.ts`**: Added 3 lines in `mergeAgentConfig()` to check if `merged.prompt` starts with `file://` and resolve it via `resolvePromptAppend()`.

## Use Case

```json
{
  "agents": {
    "sisyphus": {
      "prompt": "file://~/.config/opencode/prompts/sisyphus.md"
    }
  }
}
```

This allows fully replacing a built-in agent's prompt with a file, while `prompt_append` continues to work for additive modifications.

Zero new dependencies. Zero new functions. Just one additional call to the existing `resolvePromptAppend()`.

Related: #1908

### DIFF
--- a/src/agents/builtin-agents/agent-overrides.ts
+++ b/src/agents/builtin-agents/agent-overrides.ts
@@ -44,6 +44,10 @@ export function mergeAgentConfig(
   const { prompt_append, ...rest } = migratedOverride
   const merged = deepMerge(base, rest as Partial<AgentConfig>)
 
+  if (merged.prompt && typeof merged.prompt === 'string' && merged.prompt.startsWith('file://')) {
+    merged.prompt = resolvePromptAppend(merged.prompt, directory)
+  }
+
   if (prompt_append && merged.prompt) {
     merged.prompt = merged.prompt + "\n" + resolvePromptAppend(prompt_append, directory)
   }


one-pyy:
Hello, I would like to know if template strings are still unable to be injected, just like when writing text in the prompt args

Stranmor:
Hi @one-pyy! Yes, the behavior is exactly the same as writing the text directly in the JSON config. This PR simply reads the file contents as a raw string (using `readFileSync`) and assigns it to the `prompt` field. It doesn't perform any template variable substitution or injection on its own. The file content acts as a direct drop-in replacement for the JSON string. So, any template variables will be treated exactly the same way they are when provided inline.

PR #2263, opened 2026-03-03
fix: add retry status timeout to prevent infinite hang

### LINKED ISSUE #82: freezing leads to 143
https://github.com/CloudWaddie/LMArenaBridge/actions/runs/22602242152

i think it hangs for ages and then github eventually kills it leading to a 143

@cloudwaddie-agent investigate this ulw

cloudwaddie-agent:
## Summary

When runtime fallback triggers on errors like 503, the session status becomes "retry". The poll-for-completion loop treats "retry" the same as "busy", causing it to wait indefinitely without any timeout.

This fix adds a 3-minute timeout (based on `max_fallback_attempts * cooldown_seconds = 3 * 60s`) for the retry status. If the session stays in retry status longer than the timeout, it exits with an error instead of hanging forever (exit code 143).

## Changes

- Added `DEFAULT_RETRY_STATUS_TIMEOUT_MS` constant (180000ms = 3 minutes)
- Added `retryStatusTimeoutMs` option to `PollOptions` interface
- Track when session first enters "retry" status
- Check if retry timeout exceeded and exit with error message if so

## Root Cause

The issue was reported in CloudWaddie/actions-agent#82 where the agent hangs with exit code 143 when 503 errors trigger runtime fallback retry loops. The session gets stuck in "retry" status and the poll loop keeps waiting forever.

## Testing

The fix should be tested by:
1. Triggering a runtime fallback scenario (e.g., 503 error)
2. Verifying the session exits after 3 minutes instead of hanging indefinitely
3. Verifying the error message is helpful

Fixes: CloudWaddie/actions-agent#82

---
## Summary by cubic
Prevent infinite hangs when runtime fallback keeps retrying by adding a 3-minute timeout for sessions in "retry". The poll loop now exits with a clear, actionable error instead of waiting forever.

- **Bug Fixes**
  - Add DEFAULT_RETRY_STATUS_TIMEOUT_MS (180000 ms) and retryStatusTimeoutMs option.
  - Track first entry into "retry"; reset timer when status becomes "busy" or "idle".
  - On timeout, exit non-zero and print guidance to check model availability and network connectivity.

<sup>Written for commit e1ce2860d276a376f5a639972a61c6597201ce6c. Summary will update on new commits.</sup>

### DIFF
--- a/src/cli/run/poll-for-completion.ts
+++ b/src/cli/run/poll-for-completion.ts
@@ -136,6 +136,8 @@ export async function pollForCompletion(
           return 1
         }
       }
+    } else if (mainSessionStatus === "busy" || mainSessionStatus === "idle") {
+      retryStatusStartTimestamp = null
     }
 
     if (!eventState.mainSessionIdle) {
--- a/src/cli/run/poll-for-completion.ts
+++ b/src/cli/run/poll-for-completion.ts
@@ -10,6 +10,7 @@ const ERROR_GRACE_CYCLES = 3
 const MIN_STABILIZATION_MS = 1_000
 const DEFAULT_EVENT_WATCHDOG_MS = 30_000 // 30 seconds
 const DEFAULT_SECONDARY_MEANINGFUL_WORK_TIMEOUT_MS = 60_000 // 60 seconds
+const DEFAULT_RETRY_STATUS_TIMEOUT_MS = 180_000 // 3 minutes - max time for runtime fallback retries
 
 export interface PollOptions {
   pollIntervalMs?: number
@@ -17,6 +18,7 @@ export interface PollOptions {
   minStabilizationMs?: number
   eventWatchdogMs?: number
   secondaryMeaningfulWorkTimeoutMs?: number
+  retryStatusTimeoutMs?: number
 }
 
 export async function pollForCompletion(
@@ -37,10 +39,13 @@ export async function pollForCompletion(
   const secondaryMeaningfulWorkTimeoutMs =
     options.secondaryMeaningfulWorkTimeoutMs ??
     DEFAULT_SECONDARY_MEANINGFUL_WORK_TIMEOUT_MS
+  const retryStatusTimeoutMs =
+    options.retryStatusTimeoutMs ?? DEFAULT_RETRY_STATUS_TIMEOUT_MS
   let consecutiveCompleteChecks = 0
   let errorCycleCount = 0
   let firstWorkTimestamp: number | null = null
   let secondaryTimeoutChecked = false
+  let retryStatusStartTimestamp: number | null = null
   const pollStartTimestamp = Date.now()
 
   while (!abortController.signal.aborted) {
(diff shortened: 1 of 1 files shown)


code-yeongyu:
[sisyphus-bot]

- Author: `cloudwaddie-agent`
- Title: `fix: add retry status timeout to prevent infinite hang`
- Closing issues: #82
- Review decision: none yet
- Mergeable state: MERGEABLE

Awaiting a first-pass review.

code-yeongyu:
[sisyphus-bot] Hi cloudwaddie-agent. 🙏 Thanks for tracking down the infinite-hang case in `actions-agent#82`; the diagnosis is correct. When runtime fallback puts the session into `"retry"` repeatedly, `pollForCompletion` keeps treating it like a transient state forever, so the CLI hangs (and eventually gets killed with exit 143 by an outer timeout). Adding a bounded retry-timeout is the right shape for the fix.

A couple of things I'd like to see before this lands:

**Tests (please add at least two)**

`src/cli/run/poll-for-completion.ts` currently has no direct test file in `src/cli/run/`, but `completion.test.ts`, `events.test.ts`, and `event-handlers.test.ts` all live alongside it and would be the natural neighbours. Two focused cases would lock the new behavior:

1. *Status stays in `retry` past the timeout.* Given a fake `getMainSessionStatus` that returns `"retry"` and a clock pushed past `retryStatusTimeoutMs`, expect `pollForCompletion` to resolve to `1` and emit the two `console.error` lines.
2. *Status returns to `busy` or `idle` before the timeout.* Given a fake clock that observes `retry` once, then `busy`, then `idle`, expect the timeout NOT to fire and `retryStatusStartTimestamp` to have been reset. (The third loop iteration should observe a fresh `null` timestamp.)

Both can use the same mock harness pattern as `event-handlers.test.ts`. If you can pass `pollOptions.retryStatusTimeoutMs: 100` and a controllable time source (e.g., a `dateNow` hook), the test runs in well under a second.

**Tiny code observations (non-blocking)**

- The `else if (mainSessionStatus === "busy" || mainSessionStatus === "idle")` reset branch at the end overlaps with the existing `else if (mainSessionStatus === "idle")` branch a few lines up (which also now resets `retryStatusStartTimestamp = null`). Consolidating into a single clear "any non-retry transition resets the timer" branch would remove the duplication and read a bit more linearly.
- The 3-minute default is derived from `max_fallback_attempts * cooldown_seconds = 3 * 60s` per the PR description, but it isn't tied to those config values in code. If those defaults ever change, this constant would silently drift. Either a comment with a `// derived from ...` note, or pulling `pluginConfig.runtime_fallback?.maxAttempts` / `cooldown_seconds` here, would keep them in sync. Not a blocker; just flagging.
- The error lines use `pc.red` and `pc.yellow`. The rest of `poll-for-completion.ts` uses the same colors; nice consistency.

**Status**

CI is green (`test`, `typecheck`, `build`, `cla`, `cubic`, `GitGuardian`), merge state is clean, and the diff is localised to one file. Once the two tests above land, this is otherwise ready. Thanks again for the catch.

code-yeongyu (changes requested):
[sisyphus-bot] Requesting changes — please add the two focused tests for the retry-timeout behavior described in the inline review. Once those land I'll re-review.

code-yeongyu:
[sisyphus-bot] Keeping this open. The maintainer confirmed the diagnosis is correct and the fix shape is right, but requested two focused tests (retry-past-timeout and recovery-before-timeout). Dev still lacks any retry-status timeout in `poll-for-completion.ts`. Please add the requested tests and rebase against `packages/omo-opencode/src/cli/run/poll-for-completion.ts`. Once the tests land and CI is green, this is ready to merge.

PR #2284, opened 2026-03-04
fix(tmux): handle \r line endings and missing pane_title in list-panes

### LINKED ISSUE #2241: tmux: first subagent split can fail in single-pane session due to list-panes parsing
## Summary
When OpenCode starts in tmux with only **one pane**, first subagent pane split may fail. If a second pane is manually created first, later subagent splits usually work.

## Reproduction
1. Start `opencode -c` inside tmux with a single pane.
2. Trigger a subagent task.
3. Observe that split may not happen for the first subagent.
4. Manually split tmux pane once.
5. Trigger another subagent task.
6. Observe that split now works.

## Expected
First subagent should split from single-pane state (1 -> 2 panes).

## Actual
Sometimes first split is skipped, and logs show parser/state failure.

## Log signature
```text
[pane-state-querier] CRITICAL: failed to determine main pane {"sourcePaneId":"%...","availablePanes":[]}
```

## Suspected root cause
`list-panes` parsing is fragile in edge cases around empty `pane_title` / line endings, causing pane rows to be dropped and `mainPane` resolution to fail.

## Local fix validated
I validated a local patch that:
- extracts parsing into a dedicated parser module,
- handles `\r` line endings safely,
- treats pane title as optional,
- accepts 8 mandatory fields and optional title field.

Validation results from local patch:
- `bun test src/features/tmux-subagent/pane-state-querier.test.ts` ✅
- `bun test src/features/tmux-subagent` ✅
- `bun run typecheck` ✅
- `bun run build` ✅
- runtime behavior: `1 pane -> subagent1 -> 2 panes -> subagent2 -> 3 panes` ✅

## Environment
- macOS
- tmux 3.4
- oh-my-opencode 3.10.0

---
Patch analysis and validation were produced with assistance from Codex (GPT-5.3-codex).

MoerAI:
## Summary

- Strip `\r` characters from `list-panes` output to handle Windows-style line endings that caused pane rows to be dropped
- Relax field count check from 9 to 8 fields so panes with empty/missing `pane_title` are still parsed correctly
- This fixes the "first subagent split fails in single-pane session" bug where `mainPane` resolution failed because all pane rows were dropped by the parser

**Root cause:** In some environments, `tmux list-panes` output contains `\r\n` line endings or empty `pane_title` values. The parser's `split("\n")` left trailing `\r` in field values (corrupting parseInt), and the `fields.length < 9` check dropped rows where `pane_title` was absent.

Fixes #2241

---
## Summary by cubic
Fix tmux list-panes parsing by removing \r characters and handling missing pane_title. This restores main pane detection and prevents the first subagent split from failing in single-pane sessions.

- **Bug Fixes**
  - Strip \r from list-panes output to support CRLF line endings.
  - Allow 8 fields and default pane_title to "" so panes aren’t dropped.

<sup>Written for commit e1952d35e694904abb08dafc516c4be838dc9ca5. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/tmux-subagent/pane-state-querier.ts
+++ b/src/features/tmux-subagent/pane-state-querier.ts
@@ -27,7 +27,7 @@ export async function queryWindowState(sourcePaneId: string): Promise<WindowStat
     return null
   }
 
-  const lines = stdout.trim().split("\n").filter(Boolean)
+  const lines = stdout.trim().replace(/\r/g, "").split("\n").filter(Boolean)
   if (lines.length === 0) return null
 
   let windowWidth = 0
@@ -36,10 +36,10 @@ export async function queryWindowState(sourcePaneId: string): Promise<WindowStat
 
   for (const line of lines) {
 		const fields = line.split("\t")
-		if (fields.length < 9) continue
+		if (fields.length < 8) continue
 
 		const [paneId, widthStr, heightStr, leftStr, topStr, activeStr, windowWidthStr, windowHeightStr] = fields
-		const title = fields.slice(8).join("\t")
+		const title = fields.length > 8 ? fields.slice(8).join("\t") : ""
     const width = parseInt(widthStr, 10)
     const height = parseInt(heightStr, 10)
     const left = parseInt(leftStr, 10)


acamq:
LGTM. Thank you!

PR #2291, opened 2026-03-04
fix: also invalidate plugin from CACHE_DIR in invalidatePackage

### LINKED ISSUE #2289: [Bug]: Startup toast shows stale version after npm update — plugin cache in CACHE_DIR not invalidated
## Bug Description

After updating `oh-my-opencode` globally via npm, the startup version toast continues to show the old version instead of the newly installed one.

## Environment

- OS: Windows 11
- OpenCode version: `opencode-ai@1.2.16`
- oh-my-opencode (global npm): `3.10.0`
- oh-my-opencode (shown in toast): `3.8.5`

## Steps to Reproduce

1. Have `oh-my-opencode` working (e.g. at `3.8.5`)
2. Run `npm install -g oh-my-opencode@latest` → updates to `3.10.0`
3. Restart OpenCode
4. Startup toast still shows `3.8.5`

## Root Cause Analysis

Traced through the code to find the exact issue:

### Version Resolution (`cached-version.ts`)

`getCachedVersion()` has three fallbacks:
1. `INSTALLED_PACKAGE_JSON` = `USER_CONFIG_DIR/node_modules/oh-my-opencode/package.json`
2. Walk up from `import.meta.url` (the actual running file)
3. Walk up from `process.execPath`

On this machine, the plugin is physically installed at:
```
~/.cache/opencode/node_modules/oh-my-opencode/   ← version 3.8.5 (stale)
```

`INSTALLED_PACKAGE_JSON` points to `~/.config/opencode/node_modules/oh-my-opencode/package.json` which **does not exist**, so the fallback to `import.meta.url` kicks in and correctly reads `3.8.5` from the cache directory.

### Invalidation (`cache.ts`)

```ts
// invalidatePackage() only removes from USER_CONFIG_DIR:
const pkgDir = path.join(USER_CONFIG_DIR, "node_modules", packageName)
```

But `USER_CONFIG_DIR` = `~/.config/opencode/` while the plugin is actually at `~/.cache/opencode/node_modules/`. So `invalidatePackage()` finds nothing to delete and reports `false` — the auto-update's `bun install` then reinstalls to `USER_CONFIG_DIR/node_modules/`, but the **old copy in `CACHE_DIR/node_modules/` is never cleaned up**.

On the next restart, `import.meta.url` still resolves to the `CACHE_DIR` copy (3.8.5) because OpenCode's plugin loader picks it up first.

## Workaround

Manually delete the stale cache:
```bash
rm -rf ~/.cache/opencode/node_modules/oh-my-opencode
# Windows: rm -rf "$HOME/.cache/opencode/node_modules/oh-my-opencode"
```

## Proposed Fix

In `cache.ts`, `invalidatePackage()` should also check and remove from `CACHE_DIR/node_modules/`:

```ts
// Current (only USER_CONFIG_DIR):
const pkgDir = path.join(USER_CONFIG_DIR, "node_modules", packageName)

// Fix: also check CACHE_DIR
const pkgDirs = [
  path.join(USER_CONFIG_DIR, "node_modules", packageName),
  path.join(CACHE_DIR, "node_modules", packageName),
]
for (const dir of pkgDirs) {
  if (fs.existsSync(dir)) {
    fs.rmSync(dir, { recursive: true, force: true })
    packageRemoved = true
  }
}
```

I will submit a PR with this minimal fix.

SeeYouCowboi:
## Summary

Fixes #2289

`invalidatePackage()` only removed the stale plugin from `USER_CONFIG_DIR/node_modules/`, but on some systems bun installs the plugin into `CACHE_DIR/node_modules/` instead. This left the old version untouched, so `getCachedVersion()` would keep reading the stale copy via `import.meta.url` traversal, causing the startup toast to show the old version even after the auto-updater ran successfully.

## Root Cause

`constants.ts` defines two separate directories:
- `USER_CONFIG_DIR` = `~/.config/opencode/` (where config lives)
- `CACHE_DIR` = `~/.cache/opencode/` on Linux/macOS, `%LOCALAPPDATA%\opencode` on Windows

`invalidatePackage()` was only checking `USER_CONFIG_DIR/node_modules/oh-my-opencode/`. If bun placed the plugin in `CACHE_DIR/node_modules/`, the stale directory was never removed.

## Change

**`src/hooks/auto-update-checker/cache.ts`** — check and remove the plugin from both candidate directories:

```ts
// Before
const pkgDir = path.join(USER_CONFIG_DIR, "node_modules", packageName)
if (fs.existsSync(pkgDir)) {
  fs.rmSync(pkgDir, { recursive: true, force: true })
  packageRemoved = true
}

// After
const pkgDirs = [
  path.join(USER_CONFIG_DIR, "node_modules", packageName),
  path.join(CACHE_DIR, "node_modules", packageName),
]
for (const pkgDir of pkgDirs) {
  if (fs.existsSync(pkgDir)) {
    fs.rmSync(pkgDir, { recursive: true, force: true })
    packageRemoved = true
  }
}
```

## PR Checklist

- [x] Minimal change — only the affected function is modified

---
## Summary by cubic
Ensures auto-updated plugins load correctly by removing stale installs from both USER_CONFIG_DIR and CACHE_DIR. Fixes the startup toast showing the old version after an update.

- **Bug Fixes**
  - invalidatePackage now removes oh-my-opencode from USER_CONFIG_DIR/node_modules and CACHE_DIR/node_modules.
  - Prevents getCachedVersion from reading a stale plugin via import.meta.url traversal.

<sup>Written for commit f67b605f7a7a6e3e3d09cf51f3c94ca109298d19. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/auto-update-checker/cache.ts
+++ b/src/hooks/auto-update-checker/cache.ts
@@ -1,6 +1,6 @@
 import * as fs from "node:fs"
 import * as path from "node:path"
-import { PACKAGE_NAME, USER_CONFIG_DIR } from "./constants"
+import { CACHE_DIR, PACKAGE_NAME, USER_CONFIG_DIR } from "./constants"
 import { log } from "../../shared/logger"
 
 interface BunLockfile {
@@ -48,17 +48,22 @@ function removeFromBunLock(packageName: string): boolean {
 
 export function invalidatePackage(packageName: string = PACKAGE_NAME): boolean {
   try {
-    const pkgDir = path.join(USER_CONFIG_DIR, "node_modules", packageName)
+    const pkgDirs = [
+      path.join(USER_CONFIG_DIR, "node_modules", packageName),
+      path.join(CACHE_DIR, "node_modules", packageName),
+    ]
     const pkgJsonPath = path.join(USER_CONFIG_DIR, "package.json")
 
     let packageRemoved = false
     let dependencyRemoved = false
     let lockRemoved = false
 
-    if (fs.existsSync(pkgDir)) {
-      fs.rmSync(pkgDir, { recursive: true, force: true })
-      log(`[auto-update-checker] Package removed: ${pkgDir}`)
-      packageRemoved = true
+    for (const pkgDir of pkgDirs) {
+      if (fs.existsSync(pkgDir)) {
+        fs.rmSync(pkgDir, { recursive: true, force: true })
+        log(`[auto-update-checker] Package removed: ${pkgDir}`)
+        packageRemoved = true
+      }
     }
 
     if (fs.existsSync(pkgJsonPath)) {


SeeYouCowboi:

acamq:
Thank you!

PR #2402, opened 2026-03-09
fix: invalidate opencode plugin cache on global install

popododo0720:
## Summary

- Fixes stale plugin version after `bun install -g oh-my-opencode` by invalidating opencode's plugin cache in the postinstall script.

## Problem

When users run `bun install -g oh-my-opencode`, only the global npm package at `~/.bun/install/global/node_modules/` is updated. However, opencode resolves plugins from its own separate cache at `$XDG_CACHE_HOME/opencode/node_modules/` with its own `bun.lock`. The stale cached version continues to be loaded on subsequent launches, causing version mismatch.

## Solution

The `postinstall.mjs` script now:
1. Detects the opencode cache directory using xdg-basedir (same as opencode on all platforms)
2. Removes `node_modules/oh-my-opencode` (stale cached module)
3. Removes `bun.lock` and `bun.lockb` (required because bun reinstalls the exact pinned version from the lock entry)
4. Guards with `existsSync` to skip if cache doesn't exist (fresh install)
5. Wraps `main()` with `.catch()` to never fail the install process
6. Warns on real deletion errors (EPERM/EBUSY) instead of silently swallowing them

---
## Summary by cubic
Invalidate opencode’s plugin cache on global install so the latest `oh-my-opencode` is loaded on next launch. Prevents stale versions after `bun install -g oh-my-opencode`.

- **Bug Fixes**
  - `postinstall.mjs` deletes cached `oh-my-opencode` and `bun.lock`/`bun.lockb` under `$XDG_CACHE_HOME/opencode` (defaults to `~/.cache/opencode`) to force re-resolve.
  - Skips if the cache is missing; warns on real deletion errors without failing install.
  - Keeps platform binary verification; wraps main in a safe async flow.

<sup>Written for commit 483f2b0959411f65db58046357d87216a8f8b74b. Summary will update on new commits.</sup>

### DIFF
--- a/postinstall.mjs
+++ b/postinstall.mjs
@@ -1,7 +1,12 @@
 // postinstall.mjs
 // Runs after npm install to verify platform binary is available
+// and invalidate opencode's plugin cache so the new version is picked up on next launch.
 
 import { createRequire } from "node:module";
+import { existsSync } from "node:fs";
+import { rm } from "node:fs/promises";
+import { join } from "node:path";
+import { homedir } from "node:os";
 import { getPlatformPackageCandidates, getBinaryPath } from "./bin/platform.js";
 
 const require = createRequire(import.meta.url);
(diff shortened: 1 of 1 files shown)


popododo0720:
Re: lockfile deletion concern (identified by cubic) — this is an intentional trade-off.

Without deleting the lockfiles, `bun install` reinstalls the exact pinned (stale) version from the lock entry, making the entire cache invalidation ineffective. Confirmed through real user testing.

The blast radius is limited: `~/.cache/opencode/` is ephemeral by nature. All plugins re-resolve to their latest compatible versions on next opencode startup, which is expected behavior for a cache directory.

code-yeongyu:
[sisyphus-bot]

- Author: `popododo0720`
- Title: `fix: invalidate opencode plugin cache on global install`
- Review decision: none yet

code-yeongyu:
[sisyphus-bot] Hi popododo0720. 🙏 Thanks for the global-install cache-invalidation fix.

Picking this back up from the 5/16 triage. The PR shows CONFLICTING against current `dev`. The auto-update-checker layer has been touched on dev (see related #2975 and #3317), so the 1-file diff against the older cache structure doesn't apply cleanly.

The fix (invalidate the opencode plugin cache when a global install happens, so the updated version is actually loaded) is still valid and small. Could you rebase against current `dev` and force-push? Once CI is green I'll come back for a focused review.

Worth noting: PR #2980 (still open) takes a more comprehensive approach to the managed-install / global-install lifecycle. If your fix overlaps with #2980's surface after rebase, a brief note about whether to land both or pick one would help. Really appreciate the small focused fix; cache invalidation on install is a classic gotcha.

code-yeongyu:
[sisyphus-bot] Thanks @popododo0720! dev's postinstall.mjs now invalidates the opencode plugin cache on global install, so this fix has effectively landed. Closing as already-fixed — thanks for surfacing the issue. 🙏

Issue #2393, opened 2026-03-08
[Bug]: Model fallback never triggers on usage limits — model_fallback config undiscoverable + error pattern gaps in model-error-classifier.ts


StephanMeijer:
## Summary

When a model (e.g. Claude) hits its usage limit, OpenCode enters an infinite internal retry loop and **never switches to configured fallback models** — even when `fallback_models` arrays are properly configured for all agents/categories and `runtime_fallback.enabled: true` is set.

There are **two separate issues** causing this:

1. **Config discoverability**: The `model_fallback: true` top-level config option is required to enable the `session.status`-based fallback path — the *only* path capable of intercepting OpenCode's internal retry loop — but it defaults to `false`, is not documented, and is not surfaced to users who configure `runtime_fallback`
2. **Error pattern gaps**: Even with `model_fallback: true`, the `shouldRetryError()` function in `src/shared/model-error-classifier.ts` has incomplete patterns that fail to match several of OpenCode's transformed retry messages (most critically: `FreeUsageLimitError`)

## Environment

- oh-my-openagent: latest (via npx)
- OpenCode: latest
- Primary model: `anthropic/claude-opus-4-6`
- OS: Linux (NixOS)

### Issue 1: model_fallback not enabled

1. Configure agents with `fallback_models` arrays
2. Enable `runtime_fallback: { enabled: true, ... }`
3. Do **not** add `model_fallback: true` (most users won't know this exists)
4. Hit Claude's usage limit during a session
5. **Actual**: OpenCode retries the same model infinitely (exponential backoff, max 30s), showing "Retrying in X seconds..." forever
6. **Expected**: System should switch to the next fallback model after the first failed attempt

### Issue 2: Error pattern gaps (after enabling model_fallback)

1. Add `"model_fallback": true` to config
2. Trigger a `FreeUsageLimitError` (free usage exceeded on opencode.ai/zen)
3. **Actual**: `shouldRetryError()` returns `false` for the message `"Free usage exceeded, add credits https://opencode.ai/zen"` because none of the patterns in `RETRYABLE_MESSAGE_PATTERNS` match
4. **Expected**: Should classify this as retryable and trigger fallback

### Architecture: Two Separate Fallback Systems

The plugin has two independent fallback mechanisms that listen to different events:

| System | Config key | Listens to | Can intercept OpenCode retry loop? |
|---|---|---|---|
[...]

PR #2479, opened 2026-03-11
Fix issue #2393: auto-enable model_fallback and expand retry patterns

code-yeongyu:
## Summary
- expand retryable model error detection to catch `FreeUsageLimitError` and broader free-tier or daily-limit messages
- auto-enable `model_fallback` when `fallback_models` are configured, while keeping explicit `model_fallback: false` as an override
- extract focused model-fallback hook helpers and add regression coverage for the new config behavior

## Testing
- `bun run typecheck`
- `bun run build`
- `bun test src/shared/model-error-classifier.test.ts src/plugin/hooks/model-fallback-config.test.ts src/plugin/event.model-fallback.test.ts`
- `bun test` *(currently fails on pre-existing expectations in `src/plugin-config.test.ts` about `disabled_hooks`; unrelated to this change)*

---
## Summary by cubic
Auto-enables `model_fallback` when `fallback_models` are configured and expands retryable error detection (including free-tier limits) to make provider fallback more reliable.

- **Bug Fixes**
  - Treat `FreeUsageLimitError` and free-tier/daily-limit messages as retryable; error name matching is now case-insensitive.
  - Automatically enable `model_fallback` when any agent/category defines `fallback_models`; explicit `model_fallback: false` still disables it.

- **Refactors**
  - Extracted config and session hook into `model-fallback-config` and `create-model-fallback-session-hook`, added focused tests, and simplified `create-session-hooks`.

<sup>Written for commit fd7faff7920bd144bbe41eff4874431a3dc1ee08. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin/hooks/create-model-fallback-session-hook.ts
+++ b/src/plugin/hooks/create-model-fallback-session-hook.ts
@@ -1,0 +1,131 @@
+import type { OhMyOpenCodeConfig, HookName } from "../../config"
+
+import { createModelFallbackHook } from "../../hooks"
+import { normalizeSDKResponse } from "../../shared"
+
+import { resolveModelFallbackEnabled } from "./model-fallback-config"
+
+type SafeHook = <THook>(hookName: HookName, factory: () => THook) => THook | null
+
+type ModelFallbackSessionContext = {
+  directory: string
+  client: {
+    session: {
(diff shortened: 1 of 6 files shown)


code-yeongyu:
Duplicate of #2422 which already passes CI

PR #2483, opened 2026-03-11
fix: restore await on metadata call in create-background-task

code-yeongyu:
Follow-up to #2476. Restores await on ctx.metadata() call as identified by cubic code reviewer — prevents race condition where tool execution completes before metadata is written.

---
## Summary by cubic
Restore awaiting the `ctx.metadata(bgMeta)` call in `create-background-task` so metadata is written before the task completes, preventing a race condition that could miss or delay metadata.

<sup>Written for commit 9eefbfe3103e8651807a6b008e4c27ee8c57398d. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/background-task/create-background-task.ts
+++ b/src/tools/background-task/create-background-task.ts
@@ -98,7 +98,7 @@ export function createBackgroundTask(
             ...(sessionId ? { sessionId } : {}),
           },
         }
-        ctx.metadata?.(bgMeta)
+        await ctx.metadata?.(bgMeta)
 
         if (ctx.callID) {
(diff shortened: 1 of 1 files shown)


code-yeongyu:
[sisyphus-bot] Closing per Cubic review analysis: `ctx.metadata` is synchronous `void` per the OpenCode SDK contract (`ToolContext` in `@opencode-ai/plugin`). Adding `await` to a synchronous void function relies on undocumented runtime behavior. No other built-in tools await `ctx.metadata`. The original race condition concern was speculative — metadata writes are synchronous per SDK. Identified by Cubic.

PR #2505, opened 2026-03-12
fix(non-interactive-env): use platform-aware shell detection on Windows

### LINKED ISSUE #2344: [Bug]: non-interactive-env still hardcodes Unix export prefix on Windows in v3.10.0
### Prerequisites

### Bug Description

This appears to be a regression / incomplete fix of the previously closed Windows shell compatibility issue.

On Windows, `non-interactive-env` still prepends a Unix-style `export ...;` prefix to git commands in `v3.10.0`, which breaks execution in `cmd.exe` and PowerShell.

The previous issue `#566` was closed, but the current `v3.10.0` build still contains a hardcoded Unix shell type in the compiled hook implementation.

### Steps to Reproduce

1. Install the latest package:
   ```bash
   npm install -g oh-my-opencode@3.10.0
   ```
2. Use OpenCode on Windows with `oh-my-opencode` enabled.
3. Execute a git command through the bash tool, for example `git status`.
4. Observe that the command is rewritten with a Unix-style prefix:
   ```text
   export CI=true DEBIAN_FRONTEND=noninteractive GIT_TERMINAL_PROMPT=0 ...; git status
   ```
5. In Windows shells this fails with:
   ```text
   'export' is not recognized as an internal or external command,
   operable program or batch file.
   ```

### Expected Behavior

The hook should either:

- detect the actual shell and emit shell-appropriate syntax, or
- avoid prepending Unix syntax on Windows entirely, or
- only run when the environment is truly non-interactive.

### Actual Behavior

`non-interactive-env` still prepends a Unix `export ...;` prefix on Windows in the latest release.

### Evidence

Local verification after upgrading to `3.10.0`:

- Installed package version:
  - `C:\Users\hacke\AppData\Roaming\npm\node_modules\oh-my-opencode\package.json`
  - `"version": "3.10.0"`

- Compiled hook implementation still contains:
  - `C:\Users\hacke\AppData\Roaming\npm\node_modules\oh-my-opencode\dist\index.js:40242`
  - `const envPrefix = buildEnvPrefix(NON_INTERACTIVE_ENV, "unix");`

- The shell-aware helper still exists in the same build:
  - `C:\Users\hacke\AppData\Roaming\npm\node_modules\oh-my-opencode\dist\index.js:17322`
  - `function buildEnvPrefix(env, shellType) { ... }`
  - with explicit branches for `unix`, `powershell`, and `cmd`

So the issue is not that shell-specific support is missing globally; it is that this hook still forces `"unix"`.

### Workaround

Disabling the hook avoids the problem:

```json
{
  "disabled_hooks": ["non-interactive-env"]
}
```

### Additional Context

Related previously closed issue: `#566`

This new report is specifically about the latest release (`v3.10.0`) still shipping code that hardcodes Unix behavior in the Windows path.

### Operating System

Windows

### OpenCode Version

opencode-ai 1.2.20

MoerAI:
## Summary

Closes #2344

Replaces the hardcoded `"unix"` shell type in the non-interactive-env hook with `detectShellType()` from the shared shell-env module.

Previously, the hook always prepended `export VAR=val;` syntax, which fails in cmd.exe and PowerShell on Windows. Now it uses the existing platform detection logic:

| Environment | Shell Type | Syntax |
|---|---|---|
| Linux/macOS | `unix` | `export CI=true ...;` |
| Windows + Git Bash/WSL | `unix` | `export CI=true ...;` |
| Windows + PowerShell | `powershell` | `$env:CI='true'; ...` |
| Windows + cmd.exe | `cmd` | `set CI="true" && ...` |

## Changes

- **`src/hooks/non-interactive-env/non-interactive-env-hook.ts`** — Import `detectShellType` and use it instead of hardcoded `"unix"`

## Backward Compatibility

No behavior change on Unix/macOS/Linux or Windows with Git Bash/WSL (all resolve to `"unix"` shell type). Only Windows users running OpenCode in native cmd.exe or PowerShell see the fix.

---
## Summary by cubic
Fix non-interactive env variable injection on Windows by using platform-aware shell detection. Replaces the hardcoded `unix` shell with `detectShellType()` so PowerShell and cmd use `$env:` and `set` syntax; Unix/macOS and Git Bash/WSL remain unchanged.

<sup>Written for commit f879fa1a900aec9902a74b7797a1047b8bfddac4. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/non-interactive-env/non-interactive-env-hook.ts
+++ b/src/hooks/non-interactive-env/non-interactive-env-hook.ts
@@ -1,6 +1,6 @@
 import type { PluginInput } from "@opencode-ai/plugin"
 import { HOOK_NAME, NON_INTERACTIVE_ENV, SHELL_COMMAND_PATTERNS } from "./constants"
-import { log, buildEnvPrefix } from "../../shared"
+import { log, buildEnvPrefix, detectShellType } from "../../shared"
 
 export * from "./constants"
 export * from "./detector"
@@ -52,9 +52,9 @@ export function createNonInteractiveEnvHook(_ctx: PluginInput) {
       // The env vars (GIT_EDITOR=:, EDITOR=:, etc.) must ALWAYS be injected
       // for git commands to prevent interactive prompts.
 
-      // The bash tool always runs in a Unix-like shell (bash/sh), even on Windows
-      // (via Git Bash, WSL, etc.), so always use unix export syntax.
-      const envPrefix = buildEnvPrefix(NON_INTERACTIVE_ENV, "unix")
+      // Use platform-aware shell detection: Git Bash/WSL on Windows uses unix syntax,
+      // PowerShell uses $env: syntax, and cmd.exe uses SET syntax.
+      const envPrefix = buildEnvPrefix(NON_INTERACTIVE_ENV, detectShellType())
       
       // Check if the command already starts with the prefix to avoid stacking.
       // This maintains the non-interactive behavior and makes the operation idempotent.


MoerAI:
Closing: P0 critical design flaw identified by cubic review.

**Root cause**: `detectShellType()` checks for `PSModulePath` environment variable to detect PowerShell, but this variable exists system-wide on Windows regardless of which shell is actually running. Since OpenCode's `bash` tool explicitly launches `bash.exe` via Git Bash (not PowerShell), this detection incorrectly selects PowerShell syntax (`$env:VAR`) for commands that will run in bash, breaking non-interactive execution entirely.

The approach needs fundamental rethinking — shell detection should be based on the actual execution context (Git Bash), not system environment variables. Will re-approach if needed (issue #2344 remains open).

PR #2532, opened 2026-03-12
fix(cli): render verbose doctor check details

ricatix:
## Summary
- render verbose doctor `details` for each check so the `Models` section appears again in `doctor --verbose`
- add a formatter regression test that verifies `Models` details are included in verbose output
- keep the fix generic at the formatter layer instead of hardcoding model-specific rendering

## Verification
- `bun test src/cli/doctor/formatter.test.ts`
- `bun run dist/cli/index.js doctor --verbose`

---
## Summary by cubic
Restores `doctor --verbose` output to show per-check details, so the Models section and other details appear again. Keeps the rendering generic in the formatter.

- **Bug Fixes**
  - Render check `details` blocks in verbose output (e.g., Models).
  - Add regression test ensuring Models details appear.
  - Implemented at the formatter layer (`format-verbose.ts`), no model-specific logic.

<sup>Written for commit 63ac37cd291af97c5308e8fedd9c0a477a27995e. Summary will update on new commits.</sup>

### DIFF
--- a/src/cli/doctor/format-verbose.ts
+++ b/src/cli/doctor/format-verbose.ts
@@ -57,6 +57,19 @@ export function formatVerbose(result: DoctorResult): string {
   }
   lines.push("")
 
+  for (const check of results) {
+    if (!check.details || check.details.length === 0) {
+      continue
+    }
+
+    lines.push(`${color.bold(check.name)}`)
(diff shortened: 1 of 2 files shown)


ricatix:

acamq:
Thank you!

PR #2547, opened 2026-03-13
fix(ralph-loop): add semantic completion detection fallback for ulw-loop

### LINKED ISSUE #2489: [Bug]: ulw-loop sometimes doesn't stop, even when the agent's task is already completed
### Prerequisites

### Bug Description

Agent completes the assigned task successfully but the ulw-loop doesn't stop, continuing to run indefinitely. This happens because the Agent uses its own words to express completion (e.g., "规划已完成", "Task complete") instead of the required `<promise>DONE</promise>` format. The system doesn't recognize these as valid termination signals, causing the loop to continue running.

### Steps to Reproduce

1. Start an ulw-loop with a planning task
2. Complete the task successfully (e.g., plan file created, review passed)
3. Agent outputs completion message in its own words instead of `<promise>DONE</promise>`
4. System doesn't recognize it as termination signal
5. Loop continues running

### Expected Behavior

When the task is complete, Agent should output exactly:
```
<promise>DONE</promise>
```
And the system should immediately stop the ulw-loop.

### Actual Behavior

Agent outputs its own completion message:
- "规划任务已完成。请运行 `/start-work feishu-plugin`"
- "飞书机器人插件规划已完成，可执行..."
- "规划已完成。"

The system doesn't recognize these as valid termination signals and the loop continues running. The Agent may repeat similar messages multiple times, creating an endless loop of redundant outputs.

### Doctor Output

```shell
$ bunx oh-my-opencode doctor

 oMoMoMoMo Doctor 

 ✓ System OK (opencode 1.2.21 · oh-my-opencode 3.11.1)
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

**Severity: Critical - Token Waste**

This bug can cause severe token waste:
1. User hands a task to the agent and goes to sleep
2. Agent completes the task successfully
3. Agent outputs completion message in its own words
4. System doesn't recognize it as completion signal
5. Loop continues running indefinitely
6. **User wakes up to find thousands of tokens wasted**

This is especially dangerous for:
- Long-running tasks
- Tasks scheduled during off-hours
- Users who trust the system to auto-stop

**Root Cause**: Agent doesn't use the required `<promise>DONE</promise>` format - it uses semantic equivalents instead. The system expects literal format match. This issue is especially problematic when users work in different languages, as the Agent may naturally respond in various languages, making it harder to recognize completion signals.

**The prompt**:
```
[analyze-mode]
ANALYSIS MODE. Gather context before diving deep:

CONTEXT GATHERING (parallel):
- 1-2 explore agents (codebase patterns, implementations)
- 1-2 librarian agents (if external library involved)
- Direct tools: Grep, AST-grep, LSP for targeted searches

IF COMPLEX - DO NOT STRUGGLE ALONE. Consult specialists:
- **Oracle**: Conventional problems (architecture, debugging, complex logic)
- **Artistry**: Non-conventional problems (different approach needed)

SYNTHESIZE findings before proceeding.

---

ultrawork [SYSTEM DIRECTIVE: OH-MY-OPENCODE - RALPH LOOP 34/unbounded]

Your previous attempt did not output the completion promise. Continue working on the task.

IMPORTANT:
- Review your progress so far
- Continue from where you left off
- When FULLY complete, output: <promise>DONE</promise>
- Do not stop until the task is truly done

Original task:
the Specific execution tasks
```
an output example:
```
think:
The user is still in ultrawork loop, but my task (planning) is complete. The plan file exists and Momus approved it. I should output the completion promise.
content:
飞书机器人插件规划已完成，请运行 /start-work feishu-plugin 执行。

```

**Suggested Fix**:
1. Prompt reinforcement: Make the format requirement more explicit and direct - clearly state that ONLY `<promise>DONE</promise>` is accepted as termination signal
2. Loop count fallback: Add a configurable max loop count as a safety net, or detect potential infinite loops when the Agent outputs similar completion messages repeatedly

### Operating System

macOS

### OpenCode Version

1.2.21

MoerAI:
## Summary

Fixes #2489 — ulw-loop sometimes doesn't stop even when the agent's task is already completed, because the agent uses natural language (e.g., "task completed", "all done", "implementation complete") instead of the required `<promise>DONE</promise>` format.

### New: `semantic-completion-detector.ts`
- Phrase-based semantic fallback that detects 15 common completion expressions in assistant messages
- Returns the matched phrase for logging, or `undefined` if no match

### Modified: `completion-promise-detector.ts`
- After the strict regex `<promise>DONE</promise>` check fails, both detection paths (`detectCompletionInTranscript` and `detectCompletionInSessionMessages`) now call `detectSemanticCompletion()` on the last assistant message as a fallback
- Extracted `getAssistantMessageText()` helper to reduce duplication
- All semantic detections are logged with the matched signal for debugging

## How it works

```
1. Try strict regex: /<promise>\s*DONE\s*<\/promise>/is
2. If no match → try semantic fallback on last assistant message
3. If semantic match found → log signal + return true (stop loop)
4. If neither matches → continue loop as before
```

## Why this matters

Without this fix, when an agent completes work but uses natural language instead of the formal promise tag, the loop continues indefinitely — wasting tokens, especially dangerous for overnight/unattended tasks.

---
## Summary by cubic
Fixes #2489 by adding a semantic completion fallback so `ulw-loop` stops when the agent says “all done” instead of emitting <promise>DONE</promise>, with stricter negation and content filtering to avoid false positives.

- **Bug Fixes**
  - Added `semantic-completion-detector.ts` with word-boundary regexes, a widened negation window (40 chars), and guards to detect 15 common “done” phrases and return the matched signal for logs.
  - Updated `completion-promise-detector.ts` to apply the fallback only to the latest assistant text: extract text content from messages/transcript (ignoring tool outputs and input fields), then log the signal and stop the loop.

<sup>Written for commit 822de50e27df26e25e3ac9b3744ed29f5613f4d2. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/ralph-loop/completion-promise-detector.ts
+++ b/src/hooks/ralph-loop/completion-promise-detector.ts
@@ -50,7 +50,9 @@ export function detectCompletionInTranscript(
 				const entry = JSON.parse(line) as { type?: string; timestamp?: string }
 				if (entry.type === "user") continue
 				if (startedAt && entry.timestamp && entry.timestamp < startedAt) continue
-				lastAssistantLine = line
+				if (entry.type === "assistant") {
+					lastAssistantLine = line
+				}
 				if (pattern.test(line)) return true
 			} catch {
 				continue
--- a/src/hooks/ralph-loop/completion-promise-detector.ts
+++ b/src/hooks/ralph-loop/completion-promise-detector.ts
@@ -2,11 +2,24 @@ import type { PluginInput } from "@opencode-ai/plugin"
 import { existsSync, readFileSync } from "node:fs"
 import { log } from "../../shared/logger"
 import { HOOK_NAME } from "./constants"
+import { detectSemanticCompletion } from "./semantic-completion-detector"
 import { withTimeout } from "./with-timeout"
 
 interface OpenCodeSessionMessage {
 	info?: { role?: string }
 	parts?: Array<{ type: string; text?: string }>
+}
+
+function getAssistantMessageText(message: OpenCodeSessionMessage): string {
+	if (!message.parts) return ""
+
+	let responseText = ""
+	for (const part of message.parts) {
+		if (part.type !== "text") continue
+		responseText += `${responseText ? "\n" : ""}${part.text ?? ""}`
+	}
+
+	return responseText
 }
 
 function escapeRegex(str: string): string {
@@ -29,18 +42,30 @@ export function detectCompletionInTranscript(
 
 		const content = readFileSync(transcriptPath, "utf-8")
 		const pattern = buildPromisePattern(promise)
-		const lines = content.split("\n").filter((line) => line.trim())
+		const lines = content.split("\n").filter((line: string) => line.trim())
+		let lastAssistantLine: string | undefined
 
 		for (const line of lines) {
 			try {
 				const entry = JSON.parse(line) as { type?: string; timestamp?: string }
 				if (entry.type === "user") continue
 				if (startedAt && entry.timestamp && entry.timestamp < startedAt) continue
+				lastAssistantLine = line
 				if (pattern.test(line)) return true
 			} catch {
 				continue
 			}
 		}
+
+		const semanticSignal = detectSemanticCompletion(lastAssistantLine ?? "")
+		if (semanticSignal) {
+			log(`[${HOOK_NAME}] Semantic completion detected in transcript`, {
+				signal: semanticSignal,
+				transcriptPath,
+			})
+			return true
+		}
+
 		return false
 	} catch {
 		return false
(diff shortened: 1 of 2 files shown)


MoerAI:
**Review Status: Needs Fixes (False Positive/Negative Loop Termination)**

CI: All passing, but multiple semantic detection issues flagged across 3+ review rounds:

1. **Non-assistant entry matching** (P1): `lastAssistantLine` stores raw JSON for every non-user transcript entry, including `tool_result` outputs. A tool output containing "successfully completed" will falsely trigger loop termination. Must filter to `type === "assistant"` only.

2. **Negation-blind matching** (P2): Plain `includes()` matching treats "I am not all done yet" as matching "all done". Need word-boundary-aware matching or negation guards.

3. **Overly broad negation check** (P1): `.includes()` for negation words catches substrings in unrelated words ("pia**no**", "casi**no**", "when**ever**"), incorrectly flagging valid completions as negated. Need word-boundary regex (`\b`) instead of `.includes()`.

4. **Narrow lookbehind window**: 12-character lookbehind in `isNegated()` misses negations separated by intervening words ("I **don't think** the task is complete").

These issues can cause both premature loop stops AND infinite loops — the exact problems this PR aims to fix. Needs rework before merge.

MoerAI:
Closing in favor of #2593, which is the improved implementation of the same semantic completion detection. #2593 addresses the false positive/negative issues flagged in reviews here (non-assistant entry matching, negation-blind matching, overly broad negation check) with a cleaner architecture using consecutive-message counting and proper negation filters.

PR #2579, opened 2026-03-15
fix(look_at): avoid hanging on child session prompt

wousp112:
## Summary
- replace `look_at`'s synchronous child-session prompt path with async submission plus explicit child-session lifecycle polling
- propagate parent aborts to the spawned child session instead of leaving orphaned `look_at` runs behind
- update the `look_at` tests to assert async orchestration semantics instead of codifying the old synchronous prompt behavior

## Root cause
`look_at` created a child session and then waited on `session.prompt(...)` as if that RPC were the child completion mechanism.

That works only if the underlying request returns cleanly. When it stalls or abort propagation fails, the parent tool never reaches message fetching or status convergence. In production this leaves `look_at` stuck in `running`, blocks the parent agent, and starves queued prompts behind it.

The design intent is still correct: `look_at` should spawn a child multimodal session, wait for the child result, and return the extracted content. The bug was the implementation detail that bound tool completion to a synchronous prompt RPC instead of to the child session's lifecycle.

## Fix
- send the child request with `promptAsync(...)`
- watch the child session explicitly via status + message polling
- treat assistant text, assistant error, or stable idle-without-output as terminal outcomes
- abort the child session if the parent tool is aborted

## Validation
- `bun test src/tools/look-at/session-poller.test.ts src/tools/look-at/tools.test.ts src/tools/look-at/assistant-message-extractor.test.ts`
- `bun run typecheck`
- `bun run build`

## Evidence
Broken sessions consistently logged:
- `Sending prompt with file to session ...`
- and then stopped before `Fetching messages...`

Healthy sessions continued through message fetching and produced assistant output.

---
## Summary by cubic
Switches `look_at` to async child-session orchestration with explicit polling to prevent hangs and orphaned sessions. Adds structured error reporting, tighter status/message fallbacks, and aborts child sessions on parent abort or timeout.

- **Bug Fixes**
  - Use `promptAsync(...)` and poll `session.status` + `session.messages` until assistant text, assistant error, or stable idle (3 unchanged message polls; fallback when status is missing/errors).
  - Abort child sessions on parent abort and before throwing on timeout via `session.abort`; handle upstream abort early in the poller.
  - Return clearer timeout errors and surface assistant error name/message in tool output.

- **Refactors**
  - Replace sync `pollSessionUntilIdle` with `waitForLookAtSessionResult` (returns messages, outcome, and `statusType`).
  - Expand `extractLatestAssistantOutcome` to handle top-level or nested role/time/error and expose `completed`; keep `extractLatestAssistantText` wrapper.
  - Switch to `promptWithModelSuggestionRetry`; update tests for async orchestration, error paths, and abort behavior.

<sup>Written for commit 3fff467d7dbcf3dd02f9b53478f1f01f61ce3d4e. Summary will update on new commits.</sup>


code-yeongyu:
[sisyphus-bot]

- Author: `wousp112`
- Title: `fix(look_at): avoid hanging on child session prompt`
- Review decision: none yet

code-yeongyu:
[sisyphus-bot] Hi wousp112. 🙏 Thanks for the look_at hang fix.

Picking this back up from the 5/16 triage. The PR shows CONFLICTING against current `dev`. The look_at pipeline has been substantially reworked (PR #4229 still in flight targets the same hang symptom from a different angle, with `isAmbiguousPromptDispatchFailure` discrimination), so the 6-file diff overlaps directly with in-flight work.

This is a delicate area; the prompt-async-gate per the `AGENTS.md` invariants must not have new direct routes added outside the gate. Your fix predates that documentation, so the rebase needs to dovetail with both the gate semantics and #4229's parallel hang fix.

Could you rebase against current `dev` and explicitly clarify how your fix composes with #4229? If #4229 lands first, an opportunity for a smaller follow-up PR that only addresses the gaps #4229 doesn't cover. If yours lands first, #4229 would need to layer on top. Either way, the look_at hang is a real symptom worth fixing; the question is just which approach to ship and in what order.

code-yeongyu:
[sisyphus-bot] Thank you wousp112 — the look_at hang has been fixed on dev through the async session-poller architecture (session-poller.ts, waitForLookAtSessionResult), so this PR is superseded. Closing with appreciation for identifying the problem early.

PR #2615, opened 2026-03-16
fix(non-interactive-env): reduce git env prefix to prevent output noise (fixes #2599)

### LINKED ISSUE #2599: [Bug]: export CI=true DEBIAN_FRONTEND=noninteractive GIT_TERMINAL_PROMPT=0
### Prerequisites

### Bug Description

Image
During the operation of Ai, it keeps reporting the following error. What could be the reason for this
```
$ export CI=true DEBIAN_FRONTEND=noninteractive GIT_TERMINAL_PROMPT=0 GCM_INTERACTIVE=never HOMEBREW_NO_AUTO_UPDATE=1 GIT_EDITOR=: EDITOR=: VISUAL='' GIT_SEQUENCE_EDITOR=: 
```

### Steps to Reproduce

Normally, users propose their needs and interact with AI, and AI executes the commands

### Expected Behavior

Fix this bug

### Actual Behavior

This error keeps occurring throughout the entire context
Image

### Doctor Output

```shell
PS C:\Users\Administrator> bunx oh-my-opencode doctor

 oMoMoMoMo Doctor

 ⚠ 4 issues found:

1. Model override uses unavailable provider
   Provider(s) not found in OpenCode model cache: bailian-coding-plan
   Affects: model resolution

2. Comment checker unavailable
   Affects: comment-checker hook

3. No LSP servers detected
   LSP-dependent tools will be limited until at least one server is installed.
   Affects: lsp diagnostics, rename, references

4. GitHub CLI missing
   gh CLI is not installed.
   Affects: GitHub automation
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Windows

### OpenCode Version

1.2.26  （This seems unrelated to the version, as the same issue occurred in several previous versions before execution）

MoerAI:
## Summary
- Reduces the git env prefix from 15 vars (~300 chars) to 3 essential vars (~50 chars)
- Uses inline env syntax instead of `export` for cleaner output

## Problem
The `non-interactive-env` hook prepended ALL 15 `NON_INTERACTIVE_ENV` vars with `export` syntax to every git command:
```
export CI=true DEBIAN_FRONTEND=noninteractive GIT_TERMINAL_PROMPT=0 GCM_INTERACTIVE=never HOMEBREW_NO_AUTO_UPDATE=1 GIT_EDITOR=: EDITOR=: VISUAL='' GIT_SEQUENCE_EDITOR=: GIT_MERGE_AUTOEDIT=no GIT_PAGER=cat PAGER=cat npm_config_yes=true PIP_NO_INPUT=1 YARN_ENABLE_IMMUTABLE_INSTALLS=false; git status
```

This caused two problems:
1. Tool output was polluted with a ~300 char prefix on every git command
2. Models (especially non-Claude) echoed this prefix back as text in their responses, confusing users (#2599)

## Fix
Reduced to only the 3 vars that actually prevent git from hanging:
```
GIT_EDITOR=: GIT_PAGER=cat GIT_TERMINAL_PROMPT=0 git status
```

- `GIT_EDITOR=:` — prevents git from opening vim/nano (the #1 hang cause)
- `GIT_PAGER=cat` — prevents less/more from waiting for input
- `GIT_TERMINAL_PROMPT=0` — prevents git credential prompts

The other 12 vars (`CI`, `DEBIAN_FRONTEND`, `EDITOR`, `VISUAL`, etc.) are either redundant with the 3 above for git operations, or not relevant to git commands at all.

Fixes #2599

---
## Summary by cubic
Reduce the git env prefix in `non-interactive-env` from 15 vars to 4 essentials and use an `export`-based prefix so it works across chained commands, removing noisy output and echoing. Git commands are now prepended with `export GIT_EDITOR=: GIT_PAGER=cat GIT_TERMINAL_PROMPT=0 GCM_INTERACTIVE=never;`, fixing #2599.

<sup>Written for commit f657c749893390e7b20e7f7c1852317ed88f6501. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/non-interactive-env/non-interactive-env-hook.ts
+++ b/src/hooks/non-interactive-env/non-interactive-env-hook.ts
@@ -45,11 +45,11 @@ export function createNonInteractiveEnvHook(_ctx: PluginInput) {
         return
       }
 
-      // Minimal env prefix: only the 3 vars that prevent git from hanging.
+      // Minimal env prefix: only the 4 vars that prevent git from hanging.
       // Uses `export` so vars apply across chained commands (&&, ;).
       // Previous approach exported ALL 15 NON_INTERACTIVE_ENV vars, creating
       // a ~300 char prefix that polluted output and confused models (#2599).
-      const GIT_ENV_PREFIX = "export GIT_EDITOR=: GIT_PAGER=cat GIT_TERMINAL_PROMPT=0;"
+      const GIT_ENV_PREFIX = "export GIT_EDITOR=: GIT_PAGER=cat GIT_TERMINAL_PROMPT=0 GCM_INTERACTIVE=never;"
 
       if (command.includes("GIT_EDITOR=:")) {
         return
--- a/src/hooks/non-interactive-env/non-interactive-env-hook.ts
+++ b/src/hooks/non-interactive-env/non-interactive-env-hook.ts
@@ -45,11 +45,11 @@ export function createNonInteractiveEnvHook(_ctx: PluginInput) {
         return
       }
 
-      // Minimal env prefix: only the vars that prevent git from hanging.
-      // Previous approach used ALL NON_INTERACTIVE_ENV vars via `export`,
-      // creating a long prefix that polluted tool output and confused models
-      // into echoing it back as text (#2599).
-      const GIT_ENV_PREFIX = "GIT_EDITOR=: GIT_PAGER=cat GIT_TERMINAL_PROMPT=0"
+      // Minimal env prefix: only the 3 vars that prevent git from hanging.
+      // Uses `export` so vars apply across chained commands (&&, ;).
+      // Previous approach exported ALL 15 NON_INTERACTIVE_ENV vars, creating
+      // a ~300 char prefix that polluted output and confused models (#2599).
+      const GIT_ENV_PREFIX = "export GIT_EDITOR=: GIT_PAGER=cat GIT_TERMINAL_PROMPT=0;"
 
       if (command.includes("GIT_EDITOR=:")) {
         return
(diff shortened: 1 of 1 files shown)


MoerAI:

MoerAI:
Fixed the P1 issues flagged by reviewers:

1. **Chained commands**: Switched from inline `VAR=val command` to `export VAR=val;` so env vars apply across `&&` and `;` chains (e.g., `git add . && git commit`)
2. **Idempotency check**: Still uses `GIT_EDITOR=:` presence check — if any of the 3 vars is already present, the full prefix was already injected by a previous hook pass

The key improvement over the original code remains: 3 essential vars instead of 15, reducing output noise from ~300 chars to ~60 chars.

MoerAI:
Fixed the new issues from the latest review round:

1. **GCM_INTERACTIVE=never restored** (P1): Added as the 4th essential var. Git Credential Manager has its own interactive prompt system separate from `GIT_TERMINAL_PROMPT`, so dropping it could cause credential-manager prompts to block in unattended sessions.

2. **Re: shell.env hook suggestion** (P1): The OpenCode SDK's `shell.env` hook would be the ideal approach, but it would inject these vars into ALL shell commands (not just git). Since `GIT_EDITOR`, `GIT_PAGER`, and `GCM_INTERACTIVE` are git-specific vars, injecting them globally could have unintended side effects on non-git tools. The current approach of selectively prepending only for git commands is more targeted. If the SDK adds a way to conditionally inject env vars per-tool, we should revisit.

MoerAI:
Closing — #2599 was closed as not_planned by the maintainer, so this fix is no longer needed.

PR #2618, opened 2026-03-16
fix(runtime-fallback): extract status code from nested AI SDK errors

### LINKED ISSUE #2617: runtime-fallback: extractStatusCode can't read nested AI SDK error status codes
## Problem

`extractStatusCode()` in the runtime-fallback error classifier cannot extract HTTP status codes from AI SDK errors (`AI_APICallError`). The status code is nested 2 levels deep in the error object, but the function only checks the top level.

### Error structure from AI SDK

```json
{
  "error": {
    "name": "AI_APICallError",
    "url": "http://127.0.0.1:8317/v1/messages",
    "statusCode": 400
  }
}
```

### What `extractStatusCode` checks (line ~51510)

```javascript
const statusCode = errorObj.statusCode ?? errorObj.status ?? errorObj.data?.statusCode;
```

This only finds `statusCode` at:
- `errorObj.statusCode` (top level)
- `errorObj.status` (top level)  
- `errorObj.data?.statusCode` (one level deep)

It does NOT check:
- `errorObj.error?.statusCode` ← where AI SDK puts it
- `errorObj.cause?.statusCode`

### Text fallback also fails

The function falls back to regex matching the error message against configured `retry_on_errors` codes. But for status 400, the error message is just `"Bad Request"` — which doesn't contain the number "400".

### Impact

| HTTP Status | Error Message | extractStatusCode | Text Pattern | Retryable? |
|---|---|---|---|---|
| 429 | "...rate limit..." | ❌ fails | ✅ matches `/rate.?limit/i` | ✅ Yes (via text) |
| 400 | "Bad Request" | ❌ fails | ❌ no match | ❌ No |
| 402 | "Payment Required" | ❌ fails | ❌ no match | ❌ No |
| 403 | "Forbidden" | ❌ fails | ❌ no match | ❌ No |
| 500 | "Internal Server Error" | ❌ fails | ❌ no match | ❌ No |
| 502 | "Bad Gateway" | ❌ fails | ✅ matches `/bad.?gateway/i` | ✅ Yes (via text) |
| 503 | "Service Unavailable" | ❌ fails | ✅ matches `/service.?unavailable/i` | ✅ Yes (via text) |

So only 429, 502, 503 work for fallback. 400, 402, 403, 404, 500, 504 are silently ignored even when configured in `retry_on_errors`.

### Suggested Fix

```javascript
const statusCode = errorObj.statusCode 
  ?? errorObj.status 
  ?? errorObj.data?.statusCode
  ?? errorObj.error?.statusCode    // ← ADD: AI SDK wraps errors here
  ?? errorObj.cause?.statusCode;   // ← ADD: some error chains use cause
```

### Evidence

Real log from session `ses_32813ecb7ffeFK3Rtwm7SUpPlQ` (2026-03-16T07:31:18):
```
ERROR service=llm model=claude-sonnet-4-6 error=AI_APICallError: Bad Request
→ session.error published
→ session.idle (no fallback attempted)
```

Config had `retry_on_errors: [400, ...]` and `enabled: true`. Fallback did not fire.

## Environment

- OmO version: latest (March 2026)
- OpenCode: 1.2.27
- AI SDK: @ai-sdk/provider-utils@3.0.21
- Provider: CLIProxyAPI/Quotio (OpenAI-compatible proxy)

RaviTharuma:
## What's broken

When a model call fails, `extractStatusCode()` tries to find the HTTP status code in the error object so it can decide whether to retry with a fallback model.

The problem: the AI SDK (`@ai-sdk/provider-utils`) wraps errors in a nested structure. The status code ends up at `error.error.statusCode`, but `extractStatusCode()` only looks at the top level (`error.statusCode`, `error.status`, `error.data.statusCode`).

So when Claude returns a 400 or OpenAI returns a 500, the fallback system can't see the status code and treats the error as non-retryable. The user gets an error with no fallback attempt, even though their `retry_on_errors` config covers those codes.

## What this fixes

Adds two more places to look for the status code:

```typescript
// before
const statusCode = errorObj.statusCode ?? errorObj.status ?? errorObj.data?.statusCode

// after
const statusCode = errorObj.statusCode ?? errorObj.status ?? errorObj.data?.statusCode
  ?? errorObj.error?.statusCode    // AI SDK puts it here
  ?? errorObj.cause?.statusCode    // error chains sometimes use this
```

## Which errors were affected

| Status | Before | After |
|--------|--------|-------|
| 429 (rate limit) | Worked (text matching catches "rate limit") | Still works |
| 502/503 | Worked (text matching) | Still works |
| **400 (bad request)** | Silently skipped | Now detected |
| **500 (server error)** | Silently skipped | Now detected |
| **402, 403, 404, 504** | Silently skipped | Now detected |

## How I found this

I was debugging why runtime-fallback never fired for a 400 error. Real session log showed Atlas getting a 400 from Claude through CLIProxyAPI:

```json
{"error":{"name":"AI_APICallError","statusCode":400,"url":"http://127.0.0.1:8317/v1/messages"}}
```

The `statusCode: 400` was right there — just one level deeper than where the code was looking. The text fallback (regex matching "400" in the error message) also missed it because the message was just "Bad Request" without the number.

After patching this locally, all 11 configured `retry_on_errors` codes are properly detected.

Closes #2617

---
## Summary by cubic
Fixes runtime fallback by extracting HTTP status codes from nested SDK errors and ignoring non‑numeric fields. Ensures `retry_on_errors` reliably triggers on 4xx/5xx (including 400/500/504); closes #2617.

- **Bug Fixes**
  - Extend `extractStatusCode()` to check nested fields used by `@ai-sdk/provider-utils` (e.g., `AI_APICallError`) and select the first numeric code (skips strings like `status: "error"`).
  - Add tests covering top-level, nested (`data`/`error`/`cause`), non-numeric skip, regex message fallback, and precedence.

<sup>Written for commit de66f1f397b85d60f59a6854185045fc368228d2. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/runtime-fallback/error-classifier.test.ts
+++ b/src/hooks/runtime-fallback/error-classifier.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from "bun:test"
 
-import { classifyErrorType, extractAutoRetrySignal, isRetryableError } from "./error-classifier"
+import { classifyErrorType, extractAutoRetrySignal, extractStatusCode, isRetryableError } from "./error-classifier"
 
 describe("runtime-fallback error classifier", () => {
   test("detects cooling-down auto-retry status signals", () => {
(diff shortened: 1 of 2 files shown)


acamq:
LGTM! Thank you!

PR #2603, opened 2026-03-16
feat(atlas): introduce structured plan contract checkpoint

Jrakru:
## Summary

This draft PR is the structured-plan-contract checkpoint for Atlas.

It does not try to fully replace Atlas execution with a typed runtime planner in one step. Instead, it introduces the parser and wires it into the safest execution surfaces first:

- parser for top-level Prometheus plan tasks and waves
- parser-first progress for structured plans, with legacy fallback for older/simple plans
- final-wave gating aligned to the same parsed top-level task model
- structured next-work guidance injected into Atlas continuation/reminder flow
- structured execution contract injected into `/start-work` bootstrap
- structured override tracking when Atlas delegates a task that diverges from the parsed next-task contract

## Why

Atlas previously relied mostly on prompt instructions plus ad hoc plan rereads. That made it hard to:

- keep one consistent definition of "real top-level task"
- distinguish implementation tasks from final-wave tasks
- make parser output influence runtime behavior early enough to matter
- audit cases where Atlas deviates from the planned category / wave

This PR moves the system toward:

```text
Prometheus plan
  -> parsed structured contract
  -> reused by Atlas runtime
  -> reused by final-wave gating
  -> surfaced at /start-work
  -> deviations recorded explicitly
```

## Scope

Included:

- `src/features/boulder-state/plan-contract.ts`
- `src/features/boulder-state/storage.ts`
- `src/hooks/atlas/final-wave-plan-state.ts`
- `src/hooks/atlas/boulder-continuation-injector.ts`
- `src/hooks/atlas/idle-event.ts`
- `src/hooks/atlas/tool-execute-after.ts`
- `src/hooks/atlas/verification-reminders.ts`
- `src/hooks/atlas/system-reminder-templates.ts`
- `src/hooks/start-work/start-work-hook.ts`
- `src/hooks/atlas/tool-execute-before.ts`
- `src/hooks/atlas/plan-override-tracker.ts`

Not included:

- a full typed execution engine that directly selects every Atlas task from parsed state
- boulder schema migration for structured override persistence
- removal of all prompt-based fallback behavior

## Notes

- Structured parsing is used when the plan actually contains recognized top-level task structure.
- Legacy/simple plans still fall back to the previous checkbox-counting behavior.
- Override tracking is currently append-only in `.sisyphus/notepads/{plan}/decisions.md`.

## Verification

Focused verification used for this branch:

```bash
bun test src/features/boulder-state/storage.test.ts \
[...]

PR #2602, opened 2026-03-16
fix(atlas): tighten plan contract and preserve start-work handoff

Jrakru:
## Summary

This PR tightens Atlas's execution contract around plan mutation and `/start-work` handoff, while also making Atlas treat Prometheus task metadata as the default execution contract instead of pure ad hoc reclassification.

It fixes the concrete contract inconsistencies described in:

- Fixes #2601
- Addresses #2600 at the prompt / reminder contract layer

### 1. Tighten Atlas plan-mutation rules

Atlas prompts and reminders now say explicitly:

- Atlas MAY update checkbox / status markers after verification
- Atlas MUST NOT rewrite task wording
- Atlas MUST NOT add/remove/reorder tasks
- Atlas MUST NOT change dependencies or acceptance criteria

Files:

- `src/agents/atlas/default.ts`
- `src/agents/atlas/gpt.ts`
- `src/agents/atlas/gemini.ts`
- `src/hooks/atlas/system-reminder-templates.ts`
- `src/hooks/atlas/verification-reminders.ts`

### 2. Remove commit instructions from the Atlas reminder flow

The Atlas reminder flow no longer tells Atlas to `COMMIT ATOMIC UNIT` directly.

This keeps the execution contract coherent with Atlas's orchestrator role and avoids mixing plan-state tracking with direct git instructions in the same reminder path.

Files:

- `src/hooks/atlas/verification-reminders.ts`
- `src/hooks/atlas/index.test.ts`

### 3. Prefer planner metadata over Atlas re-planning

Atlas prompts now explicitly treat existing plan metadata as the default execution contract when the current task block already includes fields such as:

- `Parallel Execution Graph`
- `Delegation Recommendation`
- `Recommended Agent Profile`
- `Parallelization`
- `Depends On`
- `Blocked By`

The dynamic decision matrix is now framed as fallback guidance rather than the primary execution authority.

Files:

- `src/agents/atlas/default.ts`
- `src/agents/atlas/gpt.ts`
- `src/agents/atlas/gemini.ts`
- `src/agents/atlas/prompt-section-builder.ts`

### 4. Preserve Atlas handoff in `/start-work`

`/start-work` already updates the session agent to Atlas, but follow-up events can still lose that handoff if the outgoing message is not stamped consistently.

This PR stamps the outgoing message with the Atlas display name so downstream orchestration logic sees the same handoff.

Files:

- `src/hooks/start-work/start-work-hook.ts`
- `src/hooks/start-work/index.test.ts`

[...]

Issue #2601, opened 2026-03-16
[Bug]: Atlas execution contract is internally inconsistent about plan mutability and post-verification behavior


Jrakru:
### Prerequisites

- [x] I will write this issue in English
- [x] I searched existing issues and PRs for duplicates
- [x] This is specific to oh-my-opencode / oh-my-openagent
- [x] I reviewed the current Atlas prompts, Atlas hooks, and runtime behavior

### Bug Description

Atlas currently has an internally inconsistent execution contract around plan files and post-verification behavior.

At a high level, Atlas is told the plan is read-only, but the post-task reminder pipeline also tells Atlas to update that same plan file immediately after verification. One prompt variant also still points Atlas at the wrong plan state path, and the reminder pipeline used to tell Atlas to commit directly even though the main Atlas role text says git work should be delegated.

This makes the Atlas role ambiguous in practice:

- Is the plan immutable once Prometheus writes it?
- Is Atlas allowed to change only execution status?
- Should Atlas commit directly, or should git work always be delegated?

Right now the code says different things in different places.

#### 1. Atlas prompts currently own the plan as the execution artifact

- [Atlas default prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/atlas/default.ts)
- [Atlas GPT prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/atlas/gpt.ts)
- [Atlas Gemini prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/atlas/gemini.ts)

These prompt variants define `.sisyphus/plans/{name}.md` as the work-plan artifact Atlas should read while executing.

#### 2. Atlas verification reminders also mutate the plan file immediately after verification

- [Atlas verification reminders](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/hooks/atlas/verification-reminders.ts)
- [Atlas system reminder templates](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/hooks/atlas/system-reminder-templates.ts)

The reminder pipeline explicitly instructs Atlas to:

- read the plan directly after verification
- mark task completion in the plan immediately
- continue execution from that updated state

That is a valid pattern if Atlas is allowed to make status-only updates, but the permission boundary is not stated clearly enough.

#### 3. One prompt variant still used a stale plan-state path

[...]

Issue #2600, opened 2026-03-16
[Feature]: Atlas should treat Prometheus execution metadata as authoritative instead of re-planning category and parallelization at runtime


Jrakru:
### Prerequisites

- [x] I will write this issue in English
- [x] I searched existing issues and PRs for duplicates
- [x] This request is specific to oh-my-opencode / oh-my-openagent
- [x] I reviewed the current Prometheus and Atlas planning/execution code paths

### Problem Description

Prometheus is already required to produce execution-ready metadata:

- task dependency graph
- parallel execution graph
- per-task delegation recommendation
- per-task category recommendation
- per-task skill recommendation

However, Atlas does not currently consume that output as an authoritative execution contract. Atlas still re-parses the plan and re-derives parallelization and category choice at execution time.

That weakens the planner / executor separation and makes runtime behavior harder to predict and audit than the architecture suggests.

#### 1. Prometheus is explicitly required to emit execution metadata

- [delegate-task planning constants](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/tools/delegate-task/constants.ts)
- [Prometheus plan template](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/prometheus/plan-template.ts)
- [Prometheus GPT prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/prometheus/gpt.ts)

The planning contract explicitly requires:

- dependency graph
- parallel execution graph
- recommended category + skills per task
- explicit task-level execution metadata

The output format already includes blocks like:

```text
### Task N
**Delegation Recommendation**
- Category: ...
- Skills: ...
**Depends On**: ...
```

and in the newer plan format:

```text
**Recommended Agent Profile**
- Category: ...
- Skills: ...
**Parallelization**
...
```

#### 2. Atlas still re-analyzes the plan instead of defaulting to planner metadata

- [Atlas default prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/atlas/default.ts)
- [Atlas GPT prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/atlas/gpt.ts)
- [Atlas Gemini prompt](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/agents/atlas/gemini.ts)

Atlas is still instructed to:

- read the plan
- parse incomplete checkboxes
- build its own parallelization map
[...]

PR #2640, opened 2026-03-17
feat(atlas): persist preferred task session reuse

HaD0Yun:
## Summary

- persist and prefer the last reusable subagent session for the current top-level Atlas plan task
- keep Atlas continuation task-scoped by avoiding cross-task session bleed from plan advancement or explicit old-session follow-ups
- surface the preferred reuse session back into Atlas continuation / verification guidance while preserving fresh-session fallback

## Changes

- extend persisted boulder state with optional `task_sessions` so Atlas can remember a preferred reusable subagent session per current top-level plan task
- add a conservative top-level task parser that tracks the first unchecked top-level TODO / final-wave task without treating nested checkboxes as task boundaries
- capture the current task key before delegation and reuse that stable key when persisting task-session state after `task()` completes
- skip updating the current task mapping for explicit `task(session_id=...)` continuations, so old-session follow-ups do not poison the new task's preferred session
- make Atlas prefer the tracked task session in completion / continuation reminders, while still falling back to a fresh session when no reusable session exists or reuse would cross task boundaries
- tighten session-id extraction so Atlas prefers the authoritative trailing `<task_metadata>` block instead of incidental body text
- add regression coverage for:
  - missing `task_sessions` backward compatibility
  - current top-level task parsing
  - stable task-key persistence across checkbox advancement
  - explicit old-session continuation not overwriting the current task mapping
  - preferred reuse session hints in Atlas continuation output

## Testing

Targeted local verification run before opening this PR:

```bash
bun run typecheck
bun run build
bun test src/features/boulder-state/storage.test.ts src/hooks/atlas/subagent-session-id.test.ts src/hooks/atlas/index.test.ts src/hooks/start-work/index.test.ts src/tools/delegate-task/sync-continuation.test.ts src/tools/delegate-task/background-continuation.test.ts src/tools/delegate-task/tools.test.ts src/agents/atlas/atlas-prompt.test.ts
```

Results:

- `bun run typecheck` ✅
- `bun run build` ✅
- targeted `bun test ...` command above → `242 pass, 0 fail`

## Related Issues

This PR is intentionally scoped as a narrow Atlas continuity improvement rather than a broad session-architecture redesign.

### Continuation metadata lineage
- Related to #1798
- Related to #2309

### Boulder / session storage concurrency lineage
- Related to #1774
- Related to #2050
- Related to #2233

### Atlas handoff / contract / structured-plan lineage
- Related to #2583
- Related to #2584
- Related to #2600
- Related to #2601
- Related to #2602
- Related to #2603

### Atlas execution-quality background
- Related to #1309
- Related to #2335

## Why this change is needed

Atlas already has the primitive needed for continuation: `task(session_id=...)` can resume an existing delegated session with preserved context. The missing piece is persistence and preference at the current top-level plan-task boundary. Today, persisted Atlas / boulder state is plan-wide, while task-scoped reuse depends too heavily on transient session output, in-memory bookkeeping, and the orchestrator reusing the right `session_id` at the right moment.

That gap shows up in two concrete ways:

1. the last good subagent session for the current task is not durably attached to that task in persisted state
2. once the plan advances or Atlas explicitly follows up on an older session, the current task can accidentally inherit the wrong reusable session unless the runtime is careful

This PR addresses that gap without trying to redesign Atlas into a general per-plan / per-session state machine.

## Scope / non-goals

Persist and prefer the last subagent session for the current top-level plan task; fall back to a fresh session when no reusable session exists or reuse would cross task boundaries; explicitly not a broad per-plan/per-session redesign.

Here, “top-level plan task” is kept conservative: the current task unit Atlas is executing from its todo-list / structured-plan flow, not a broader undocumented global task graph.

---
## Summary by cubic
Persist and prefer the last reusable subagent session for the current top‑level Atlas task, now hardened with lineage validation and ambiguity guards to improve continuity without cross‑task bleed. Preferred sessions are surfaced in continuation, completion, and verification prompts, with a safe fresh‑session fallback.

- **New Features**
  - Persist `task_sessions` in `boulder.json` and expose `getTaskSessionState`/`upsertTaskSessionState`.
  - Capture the delegated task key at `task()` start and persist on completion, even if the plan advances.
  - Skip persistence when `task(session_id=...)` is used or when multiple in‑flight calls claim the same task.
  - Prefer the tracked session in completion gates, orchestrator reminders, and idle continuation prompts; add a “Preferred reuse session” hint.
  - Validate extracted session IDs against the active boulder lineage before suggesting or persisting them.

- **Bug Fixes**
  - Tightened session reuse parsing: prefer the trailing `<task_metadata>` block, ignore body text, and handle missing IDs with a placeholder.
  - Do not reuse explicitly resumed session IDs in follow‑up reminders for the current task.
  - Clean up pending task refs on background launches to avoid stale state.

<sup>Written for commit 8859da5fef0a5c2a3eea69dfd7c3fb33e7736f60. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/boulder-state/index.ts
+++ b/src/features/boulder-state/index.ts
@@ -1,3 +1,4 @@
 export * from "./types"
 export * from "./constants"
 export * from "./storage"
+export * from "./top-level-task"
--- a/src/features/boulder-state/storage.test.ts
+++ b/src/features/boulder-state/storage.test.ts
@@ -11,8 +11,11 @@ import {
   getPlanName,
   createBoulderState,
   findPrometheusPlans,
+  getTaskSessionState,
+  upsertTaskSessionState,
 } from "./storage"
 import type { BoulderState } from "./types"
+import { readCurrentTopLevelTask } from "./top-level-task"
 
 describe("boulder-state", () => {
   const TEST_DIR = join(tmpdir(), "boulder-state-test-" + Date.now())
@@ -134,6 +137,24 @@ describe("boulder-state", () => {
       expect(result?.session_ids).toEqual(["session-1", "session-2"])
       expect(result?.plan_name).toBe("my-plan")
     })
+
+    test("should default task_sessions to empty object when missing from JSON", () => {
+      // given - boulder.json without task_sessions field
+      const boulderFile = join(SISYPHUS_DIR, "boulder.json")
+      writeFileSync(boulderFile, JSON.stringify({
+        active_plan: "/path/to/plan.md",
+        started_at: "2026-01-01T00:00:00Z",
+        session_ids: ["session-1"],
+        plan_name: "plan",
+      }))
+
+      // when
+      const result = readBoulderState(TEST_DIR)
+
+      // then
+      expect(result).not.toBeNull()
+      expect(result!.task_sessions).toEqual({})
+    })
   })
 
   describe("writeBoulderState", () => {
@@ -246,6 +267,115 @@ describe("boulder-state", () => {
       const success = clearBoulderState(TEST_DIR)
       // then
       expect(success).toBe(true)
+    })
+  })
+
+  describe("task session state", () => {
+    test("should persist and read preferred session for a top-level plan task", () => {
+      // given - existing boulder state
+      const state: BoulderState = {
+        active_plan: "/plan.md",
+        started_at: "2026-01-02T10:00:00Z",
+        session_ids: ["session-1"],
+        plan_name: "plan",
+      }
+      writeBoulderState(TEST_DIR, state)
+
+      // when
+      upsertTaskSessionState(TEST_DIR, {
+        taskKey: "todo:1",
+        taskLabel: "1",
+        taskTitle: "Implement auth flow",
+        sessionId: "ses_task_123",
+        agent: "sisyphus-junior",
+        category: "deep",
+      })
+      const result = getTaskSessionState(TEST_DIR, "todo:1")
+
+      // then
+      expect(result).not.toBeNull()
+      expect(result?.session_id).toBe("ses_task_123")
+      expect(result?.task_title).toBe("Implement auth flow")
+      expect(result?.agent).toBe("sisyphus-junior")
+      expect(result?.category).toBe("deep")
+    })
+
+    test("should overwrite preferred session for the same top-level plan task", () => {
+      // given - existing boulder state with prior preferred session
+      const state: BoulderState = {
+        active_plan: "/plan.md",
+        started_at: "2026-01-02T10:00:00Z",
+        session_ids: ["session-1"],
+        plan_name: "plan",
+        task_sessions: {
+          "todo:1": {
+            task_key: "todo:1",
+            task_label: "1",
+            task_title: "Implement auth flow",
+            session_id: "ses_old",
+            updated_at: "2026-01-02T10:00:00Z",
+          },
+        },
+      }
+      writeBoulderState(TEST_DIR, state)
+
+      // when
+      upsertTaskSessionState(TEST_DIR, {
+        taskKey: "todo:1",
+        taskLabel: "1",
+        taskTitle: "Implement auth flow",
+        sessionId: "ses_new",
+      })
+      const result = getTaskSessionState(TEST_DIR, "todo:1")
+
+      // then
+      expect(result?.session_id).toBe("ses_new")
+    })
+  })
+
+  describe("readCurrentTopLevelTask", () => {
+    test("should return the first unchecked top-level task in TODOs", () => {
+      // given - plan with nested and top-level unchecked tasks
+      const planPath = join(TEST_DIR, "current-task-plan.md")
+      writeFileSync(planPath, `# Plan
+
+## TODOs
+- [x] 1. Finished task
+  - [ ] nested acceptance checkbox
+- [ ] 2. Current task
+
+## Final Verification Wave
+- [ ] F1. Final review
+`)
+
+      // when
+      const result = readCurrentTopLevelTask(planPath)
+
+      // then
+      expect(result).not.toBeNull()
+      expect(result?.key).toBe("todo:2")
+      expect(result?.title).toBe("Current task")
+    })
+
+    test("should fall back to final-wave task when implementation tasks are complete", () => {
+      // given - plan with only final-wave work remaining
+      const planPath = join(TEST_DIR, "final-wave-current-task-plan.md")
+      writeFileSync(planPath, `# Plan
+
+## TODOs
+- [x] 1. Finished task
+
+## Final Verification Wave
+- [ ] F1. Final review
+`)
+
+      // when
+      const result = readCurrentTopLevelTask(planPath)
+
+      // then
+      expect(result).not.toBeNull()
+      expect(result?.key).toBe("final-wave:f1")
+      expect(result?.title).toBe("Final review")
     })
   })
 
--- a/src/features/boulder-state/storage.ts
+++ b/src/features/boulder-state/storage.ts
@@ -6,7 +6,7 @@
 
 import { existsSync, readFileSync, writeFileSync, mkdirSync, readdirSync } from "node:fs"
 import { dirname, join, basename } from "node:path"
-import type { BoulderState, PlanProgress } from "./types"
+import type { BoulderState, PlanProgress, TaskSessionState } from "./types"
 import { BOULDER_DIR, BOULDER_FILE, PROMETHEUS_PLANS_DIR } from "./constants"
 
 export function getBoulderFilePath(directory: string): string {
@@ -28,6 +28,9 @@ export function readBoulderState(directory: string): BoulderState | null {
     }
     if (!Array.isArray(parsed.session_ids)) {
       parsed.session_ids = []
+    }
+    if (!parsed.task_sessions || typeof parsed.task_sessions !== "object" || Array.isArray(parsed.task_sessions)) {
+      parsed.task_sessions = {}
     }
     return parsed as BoulderState
   } catch {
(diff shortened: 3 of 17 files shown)


HaD0Yun:

HaD0Yun:
Thanks — I addressed the 2 issues identified by cubic and pushed a follow-up commit (`8adf6a2c`).

What changed:
- tightened `task_metadata` session extraction so Atlas now parses the last explicit `<task_metadata>...</task_metadata>` block first, then only falls back to `Session ID:` output
- added a regression test covering incidental `session_id` text outside the metadata block
- cleaned up `pendingTaskRefs` before the background-launch early return path so background task launches do not leave stale entries behind
- added a regression test covering pending task-ref cleanup on background launches

Fresh verification after the fix:
```bash
bun run typecheck
bun run build
bun test src/features/boulder-state/storage.test.ts src/hooks/atlas/subagent-session-id.test.ts src/hooks/atlas/index.test.ts src/hooks/start-work/index.test.ts src/tools/delegate-task/sync-continuation.test.ts src/tools/delegate-task/background-continuation.test.ts src/tools/delegate-task/tools.test.ts src/agents/atlas/atlas-prompt.test.ts
```
Result: `244 pass, 0 fail`

@cubic-dev-ai re-run review please.

code-yeongyu (changes requested):
Hey @HaD0Yun! 👋 Thanks for putting this together — the core idea of persisting task-scoped session reuse for Atlas is really solid, and I can see you've thought carefully about the edge cases (plan advancement, explicit resume guard, backward compat). Tests are comprehensive and all green (116 pass).

That said, I found a few issues that I think we should address before merging. I've left inline comments on each file — please take a look!

**TL;DR — 3 blocking issues:**

1. **Explicit resume path leaks the wrong preferred session into reminders** (`tool-execute-after.ts`)
2. **Parallel `task()` calls all bind to the same task key** (`tool-execute-before.ts`)
3. **Session ID parsed from model output is trusted without validation** (`subagent-session-id.ts` → `storage.ts`)

Everything else looks great. The type design, backward compat handling, and test coverage are all well done. Let's get these three sorted and this is good to go! 🚀

code-yeongyu:
## 📁 `src/hooks/atlas/tool-execute-after.ts` — Blocking

Hey! This is the main one I wanted to flag. Really nice work on the `resolvePreferredSessionId` helper and the `pendingTaskRef` flow — the sequential happy path is clean.

**The issue:** When Atlas explicitly resumes an older session via `task(session_id="ses_old")`, the before-hook correctly sets `pendingTaskRef = null` to skip persistence. But in the after-hook, `resolvePreferredSessionId(currentSessionId, trackedSessionId)` still picks up `currentSessionId` (which is the old resumed session), and that gets fed into `buildCompletionGate()` / `buildOrchestratorReminder()`.

So the immediate reminder after an explicit resume tells Atlas: "reuse `ses_old`" — which is exactly the cross-task bleed we're trying to prevent!

**Suggested fix:**

```typescript
// When pendingTaskRef === null (explicit resume), don't let the resumed
// session become the preferred session for the CURRENT task.
const preferredSessionId = pendingTaskRef === null
  ? resolvePreferredSessionId(undefined, trackedTaskSession?.session_id)
  : resolvePreferredSessionId(subagentSessionId, trackedTaskSession?.session_id)
```

This way, explicit resumes still work fine, but the reminder points to the correct tracked session (or falls back to `"<session_id>"`).

Would love to see a regression test for this too — something like: "after explicit resume of old session, reminder should NOT contain the old session ID as preferred".

code-yeongyu:
## 📁 `src/hooks/atlas/tool-execute-before.ts` — Blocking (but scopeable)

The task-ref capture logic here is clever — snapshotting the current top-level task before delegation so it stays stable even when the plan advances. Nice!

**The concern:** `readCurrentTopLevelTask()` always returns the *first* unchecked top-level task. Atlas's agent instructions explicitly allow multiple `task()` calls in one message (parallel delegation). If Atlas fires off tasks for TODO items 3, 4, and 5 simultaneously, all three capture `todo:3` as their task key. When they complete, they overwrite each other's `task_sessions` entry.

**Options (pick what feels right for this PR's scope):**

1. **Scope it out explicitly** — Add a code comment acknowledging this as a known limitation, create a follow-up issue. Since Atlas's primary flow is sequential task delegation, this is pragmatic.

2. **Skip persistence when ambiguous** — If there are already pending task refs for the same key, skip persistence for subsequent ones. Something like:
   ```typescript
   const existingRefForKey = [...pendingTaskRefs.values()].some(
     ref => ref?.key === currentTask.key
   )
   if (existingRefForKey) {
     // Another task() call already claimed this key — skip to avoid collision
     pendingTaskRefs.set(toolInput.callID, null)
   }
   ```

I'd be fine with option 1 honestly — it keeps this PR focused. Just want to make sure we're conscious of it.

code-yeongyu:
## 📁 `src/hooks/atlas/subagent-session-id.ts` — Blocking (security)

The tightened parsing here is a big improvement over the old version — preferring `<task_metadata>` over incidental body text is the right call, and the test coverage for edge cases (body text vs metadata, post-tag bleed) is excellent.

**The security concern:** The session ID is ultimately parsed from *model-generated text*. A prompt-injected or misbehaving subagent could emit:

```
<task_metadata>
session_id: ses_some_unrelated_session
</task_metadata>
```

This gets persisted to `boulder.json` via `upsertTaskSessionState()`, then surfaced in reuse hints — potentially steering Atlas to resume an unrelated session.

**Possible mitigations (in order of effort):**

1. **Cheapest:** Validate the extracted session ID actually exists and belongs to the current orchestrator's session tree before persisting. Something like:
   ```typescript
   const session = await ctx.client.session.get({ path: { id: extractedId } })
   if (!session?.data) { /* skip persistence */ }
   ```

2. **Better:** Source the session ID from the tool's runtime metadata (if available) rather than parsing output text. The background-agent spawner already knows the session ID it created.

3. **Best (but bigger scope):** Have the `task` tool include `session_id` in its structured `metadata` output, not just in the text body. Then `tool-execute-after` reads `toolOutput.metadata.session_id` directly.

For this PR, option 1 would be sufficient I think. Or if you'd rather scope it out, at minimum add a `// TODO: validate session ownership before persisting` comment so we don't forget.

code-yeongyu:
## 📁 `src/features/boulder-state/top-level-task.ts` — Non-blocking, suggestion

Really clean implementation! The regex patterns are well-named, the section state machine is easy to follow, and skipping nested checkboxes via indentation check is smart.

**Two small things:**

1. **Parser duplication:** There's an existing `final-wave-plan-state.ts` that also parses `## TODOs` / `## Final Verification Wave` sections with slightly different semantics. Right now it's fine, but if the plan format ever changes, these two parsers could drift apart silently. Maybe worth a comment like `// See also: final-wave-plan-state.ts for related parsing` so future contributors know to update both?

2. **Regex for session IDs:** The new `extractSessionIdFromOutput` uses `ses_[a-zA-Z0-9_]+`, but `src/hooks/task-resume-info/hook.ts` uses `ses_[a-zA-Z0-9_-]+` (includes hyphens). If any session IDs contain hyphens, they'd be truncated here. Might want to align the regex patterns — easy one-char fix: `ses_[a-zA-Z0-9_-]+`

Neither of these is blocking, just worth keeping in mind! 🙂

code-yeongyu:
## 📁 `src/features/boulder-state/types.ts` — Looks great ✅

The type design is clean:
- `TaskSessionState` captures exactly what's needed for session reuse
- `TopLevelTaskRef` is a nice abstraction for the plan parser output
- `task_sessions` being optional on `BoulderState` ensures backward compat
- All persisted keys are `snake_case` ✅

No notes here — well done!

code-yeongyu:
## 📁 `src/features/boulder-state/storage.ts` — Non-blocking note

The `upsertTaskSessionState` follows the existing `appendSessionId` pattern perfectly — read-modify-write with fallback to null on failure. Consistent and easy to follow.

**Minor note:** Both this function and `appendSessionId` do read-modify-write without any file locking. This isn't a regression (same pattern already exists), but concurrent task completions could theoretically lose entries. Just flagging for awareness — probably worth a follow-up issue alongside the per-plan boulder redesign (#2233 / #2050).

The `task_sessions` defaulting to `{}` in `readBoulderState` is the right call for backward compat. 👍

code-yeongyu:
## 📁 `src/hooks/atlas/verification-reminders.ts` — Looks good ✅

`buildReuseHint()` is a clean addition. The hint text is clear and actionable for the orchestrator. Injecting it into both the completion gate and standalone verification reminder makes sense.

One tiny thought: once the blocking issue in `tool-execute-after.ts` is fixed (explicit resume path), the hint will always show the correct session — so no changes needed here, just making sure the upstream data is right.

code-yeongyu:
## 📁 Tests (`index.test.ts`, `subagent-session-id.test.ts`, `storage.test.ts`) — Great coverage ✅

Really impressed with the test coverage here:

- **Stable task key test** (plan advances before task completes) — this is exactly the kind of race condition test that catches real bugs
- **Cross-task resume guard** (explicit session_id doesn't poison current task) — perfect
- **Session ID extraction** (metadata vs body text, post-tag bleed) — thorough
- **Backward compat** (missing `task_sessions` defaults to `{}`) — necessary and correct

**One small thing:** I noticed a couple of lines in the existing test had their indentation changed from 6 spaces to 5 spaces (the `PHASE 1` / `PHASE 2` assertions). Looks accidental — might want to revert those to keep the diff clean.

**Missing test suggestion:** After fixing the explicit-resume reminder issue, add a test that asserts: "when Atlas explicitly resumes an old session, the verification reminder should NOT contain the old session ID as the preferred reuse session."

Overall, really solid test work! 🎯

HaD0Yun:
Thanks for the careful review — I pushed a follow-up that addresses the blocking points you called out.

Changes in this update:
- explicit `task(session_id=...)` resumes no longer leak the resumed session into the current task's reuse hint
- parallel in-flight `task()` calls that would claim the same top-level task key now skip persistence instead of colliding
- extracted subagent `session_id`s are now validated against the active session lineage before being persisted or surfaced in reminders

I also added regressions for each case and re-ran:
- `bun test src/hooks/atlas`
- `bun run typecheck`
- `bun run build`

PR #2653, opened 2026-03-17
fix(thinking-block-validator): preserve original signature by reusing the thinking Part verbatim

RaviTharuma:
## Problem

The `thinking-block-validator` hook crashes sessions that use extended thinking models with:

```
messages.N.content.M: Invalid `signature` in `thinking` block
```

### Root cause

`prependThinkingBlock` constructed a brand-new object from the thinking *text* alone:

```ts
const thinkingPart = {
  type: "thinking" as const,
  id: `prt_0000000000_synthetic_thinking`,
  thinking: thinkingContent,
  synthetic: true,
  // ❌ no 'signature' field
}
```

The Anthropic API requires every `thinking` block in conversation history to carry the **original `signature`** that was returned when the block was first generated.  A synthetic block without a valid signature is rejected unconditionally — making the hook cause exactly the kind of API error it was designed to prevent.

## Fix

1. **`findPreviousThinkingContent` → `findPreviousThinkingPart`** — returns the original `Part` object (including its `signature`) rather than a plain string.  Previously-injected synthetic parts are skipped because they have no valid signature.

2. **`prependThinkingBlock`** — now accepts a `Part` and `unshift`s it verbatim, so the `signature` is preserved unchanged.

3. **No-op when no real part exists** — if no previous assistant turn contains a real (non-synthetic) thinking block, injection is skipped entirely.  The downstream error — if any — is preferable to a *guaranteed* API rejection caused by a signature-less block.

## Impact

- Fixes the `Invalid `signature` in `thinking` block` error for all extended-thinking models
- No behavioural change for sessions where a real thinking part is available in history
- Removes the `[Continuing from previous reasoning]` placeholder that was never valid

---
## Summary by cubic
Fixes session crashes in `thinking-block-validator` by reusing the original Anthropic-signed `thinking` part and only injecting when such parts exist. Prevents Anthropic “Invalid `signature` in `thinking` block” errors and ignores GPT `reasoning` blocks.

- **Bug Fixes**
  - Renamed `findPreviousThinkingContent` → `findPreviousThinkingPart`; returns the original Anthropic `thinking` `Part` and skips synthetic entries.
  - `prependThinkingBlock` now unshifts the original `Part` verbatim to preserve `signature`.
  - Replaced model-name heuristic with a content-based check that only matches signed Anthropic `thinking` blocks; excludes GPT `reasoning`.
  - Added `ThinkingPart` and `isAnthropicThinkingPart`; lookup requires both `thinking` and `signature`, and we skip injection if none is found.

<sup>Written for commit 98fb17e801c73dd52b0a46986a10f08310d54f03. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/thinking-block-validator/hook.ts
+++ b/src/hooks/thinking-block-validator/hook.ts
@@ -21,11 +21,6 @@ interface MessageWithParts {
   parts: Part[]
 }
 
-interface ThinkingPart {
-  thinking?: string
-  text?: string
-}
-
 interface MessageInfoExtended {
   id: string
   role: string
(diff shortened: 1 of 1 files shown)


RaviTharuma:
@acamq please check it again :)

code-yeongyu:
Thanks for finding and fixing this! Great diagnosis in the PR description.

This was independently fixed in `0732cb85` (already on `dev`), which takes a similar approach — reusing the original signed thinking part verbatim instead of constructing a synthetic one.

Your content-based gating approach (`hasAnthropicThinkingBlocksInHistory`) is a nice idea vs model-name checks, but the merged fix also handles `redacted_thinking` parts which this PR misses.

Closing as duplicate, but appreciate the contribution!

code-yeongyu:
Superseded by #2828 which combines your content-based gating approach with `redacted_thinking` support. Thanks for the inspiration @RaviTharuma — your insight on content-based vs model-name gating made the final fix better!

RaviTharuma:
@code-yeongyu nice! we should keep this motion going and slowly get rid of other model-name base gating that can break things down the road 👍

PR #2655, opened 2026-03-17
fix(circuit-breaker): make repetitive detection target-aware and add enabled escape hatch

### LINKED ISSUE #2652: [Bug][Agent-generated] Background circuit breaker falsely cancels explore/librarian during legitimate search bursts
This issue was drafted with help from an AI coding agent after reproducing the behavior locally and inspecting the relevant source paths. I reviewed the contents before posting. Marking it as agent-generated for clarity.

## Prerequisites
- [x] I searched existing issues and did not find an obvious duplicate.
- [x] I am using the latest version available in my local setup.
- [x] I checked the docs / investigated source and still hit the issue.

## Bug Description
Background subagents such as `explore` and `librarian` are being cancelled by the repetitive-tool circuit breaker during legitimate research/search bursts.

The detector appears to treat repeated use of the same tool name as an infinite loop even when the inputs differ materially, which is common for `glob`, `grep`, `bash`, and similar exploration tools.

## Steps to Reproduce
1. Run a task that asks `explore` to search broadly across a codebase, or asks `librarian` to inspect an external repo/docs with many search/tool turns.
2. Let the subagent perform a burst of repeated search calls with different queries/paths.
3. Observe background-task cancellation.

Two concrete reproductions from my session:

```text
[BACKGROUND TASK CANCELLED]
ID: bg_4ac25da9
Description: Find explorer kill logic
Error: Subagent repeatedly called glob 16/16 times in the recent tool-call window (80% threshold). This usually indicates an infinite loop. The task was automatically cancelled to prevent excessive token usage.
```

```text
[BACKGROUND TASK CANCELLED]
ID: bg_390c502a
Description: Inspect upstream plugin changes
Error: Subagent repeatedly called bash 16/16 times in the recent tool-call window (80% threshold). This usually indicates an infinite loop. The task was automatically cancelled to prevent excessive token usage.
```

## Expected Behavior
Search-heavy agents should be allowed to use the same tool many times in a row when the inputs differ and progress is still being made.

A real loop detector should still catch repeated identical or near-identical tool calls, but it should not kill healthy exploration solely because the tool name repeats.

## Actual Behavior
The background circuit breaker cancels the subagent mid-task, which prevents `explore` / `librarian` from finishing legitimate investigations.

## Doctor Output
```shell
 oMoMoMoMo Doctor 

 ⚠ 1 issue found:

   Affects: comment-checker hook
```

## Error Logs
```shell
Subagent repeatedly called glob 16/16 times in the recent tool-call window (80% threshold). This usually indicates an infinite loop. The task was automatically cancelled to prevent excessive token usage.

Subagent repeatedly called bash 16/16 times in the recent tool-call window (80% threshold). This usually indicates an infinite loop. The task was automatically cancelled to prevent excessive token usage.
```

## Configuration
Current stopgap mitigation on my side is config-only, not a code patch:

```json
{
  "background_task": {
    "staleTimeoutMs": 1800000,
    "messageStalenessTimeoutMs": 1800000,
    "syncPollTimeoutMs": 1800000,
    "circuitBreaker": {
      "maxToolCalls": 400,
      "windowSize": 60,
      "repetitionThresholdPercent": 100
    }
  },
  "agents": {
    "explore": {
      "prompt_append": "Codebase exploration may require many search turns in a row. Avoid repeating the same tool call with the same inputs unless you explicitly state why the retry is necessary. After each search burst, synthesize what changed and switch approach if results stop improving."
    },
    "librarian": {
      "prompt_append": "External research may require many search turns in a row. Avoid repeating the same tool call with the same inputs unless you explicitly state why the retry is necessary. After each search burst, synthesize what changed and switch approach if results stop improving."
    }
  }
}
```

This reduces the false positives enough to keep working, but it is only a temporary workaround.

## Additional Context
Relevant source paths in current `dev`:

- `src/features/background-agent/loop-detector.ts`
- `src/features/background-agent/manager-circuit-breaker.test.ts`
- `src/cli/run/types.ts`

From reading those files, the current repetitive-tool detection appears to count repeated **tool names** in a sliding window, while tool events already expose input state. That makes `grep(query=A)`, `grep(query=B)`, and `grep(query=C)` look like the same repeated action.

A likely fix would be to detect repeated `(tool name + normalized input)` signatures instead of tool name alone, while still keeping the absolute `maxToolCalls` breaker as the final backstop.

## Operating System
macOS

## OpenCode Version
1.2.27

tad-hq:
## Summary

- Fix false-positive background task cancellations by changing repetitive detection from tool-name counting to target-aware tool signatures.
- Add `circuitBreaker.enabled` (default `true`) as an escape hatch for repetitive-pattern cancellation.
- Preserve the absolute `maxToolCalls` safety backstop and keep cancellation messages human-readable.

## Changes

- Added config support:
  - `src/config/schema/background-task.ts`: `circuitBreaker.enabled?: boolean`
  - `src/features/background-agent/constants.ts`: `DEFAULT_CIRCUIT_BREAKER_ENABLED = true`
  - `assets/oh-my-opencode.schema.json` regenerated
- Updated loop detection internals:
  - `ToolCallWindow.toolNames` -> `toolSignatures`
  - Added deterministic signature generation: `toolName::sortedJson(input)`
  - Empty/missing input falls back to bare tool name for backward compatibility
  - `recordToolCall(...)` now accepts optional `toolInput`
  - Repetitive detection still uses the same threshold math, but now counts signatures instead of bare tool names
  - Returned cancellation `toolName` remains bare (no signature payload leakage)
- Updated background manager wiring:
  - `MessagePartInfo.state` now includes optional `input`
  - Passes `partInfo.state?.input` to `recordToolCall(...)`
  - Repetitive-pattern cancellation now guarded by `if (circuitBreaker.enabled)`
  - Absolute `maxToolCalls` cancellation remains unconditional
- Added tests:
  - `loop-detector.test.ts`: enabled flag resolution, signature normalization/order stability, different-target no-trigger, same-target trigger, no-input fallback
  - `manager-circuit-breaker.test.ts`: integration coverage for different-target run-through, same-target cancellation with bare tool name, `enabled=false` behavior, and absolute-cap backstop

## Testing

```bash
bun run typecheck
bun run build
bun test src/features/background-agent/loop-detector.test.ts src/features/background-agent/manager-circuit-breaker.test.ts src/config/schema/background-task-circuit-breaker.test.ts
```

Additional verification:
- OpenCode smoke-tested with isolated config (`XDG_CONFIG_HOME=/tmp/omo-smoke-xdg`) and local plugin path (`file:///.../dist/index.js`) so user config was not modified.

Known upstream baseline (not introduced by this PR):
- `bun test src/features/background-agent/` has one pre-existing failure in `src/features/background-agent/manager.test.ts`:
  - `BackgroundManager process cleanup > should remove listeners after last shutdown`
  - reproduced on clean `dev` in a temporary worktree

## Related Issues

Closes #2652

---
## Summary by cubic
Make the circuit breaker target-aware to stop false-positive cancellations, and add an escape hatch to disable repetitive-pattern checks. Keeps the absolute `maxToolCalls` backstop unchanged. Closes #2652.

- **Bug Fixes**
  - Detect repetition by tool signature (`toolName::sortedJson(input)`) instead of tool name.
  - Deterministic signature generation; empty/missing input falls back to bare tool name.
  - Manager forwards `state.input` to loop detection; cancellation message shows only the bare tool name.
  - Prevents cancelling when the same tool runs on different targets; still cancels on true same-target loops.

- **New Features**
  - Add `circuitBreaker.enabled` (default `true`) to disable repetitive-pattern cancellation.
  - Update JSON schema and defaults; preserve unconditional absolute cap via `maxToolCalls`.
  - Add tests covering enabled flag, signature stability, different-target pass-through, and same-target cancellation.

<sup>Written for commit 224ecea8c78e6fc81ce603b1e1930a48e35b2076. Summary will update on new commits.</sup>


tad-hq:

PR #2666, opened 2026-03-18
fix(schema): strip contentEncoding from MCP tool schemas for Gemini compatibility (fixes #2200)

### LINKED ISSUE #2200: Junior Sisyphus fails with Gemini when MCP tools use contentEncoding in schemas
## Bug Description

Junior Sisyphus agents using Gemini models fail with a 400 Bad Request when MCP servers include `contentEncoding` in their tool schemas. This is ultimately an opencode-layer issue (the schema sanitizer doesn't strip `contentEncoding`), but omo could mitigate it.

## Error

```
Bad Request: {
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"contentEncoding\" at 'request.tools[0].function_declarations[87].parameters.properties[4].value.items.properties[2].value': Cannot find field.
    ...
    "status": "INVALID_ARGUMENT"
  }
}
```

## Root Cause

MCP tools (e.g. chrome-devtools-mcp's `take_screenshot`, `take_snapshot`, `upload_file`) define `contentEncoding: "base64"` in their JSON schemas. This is valid JSON Schema but unsupported by Gemini's function calling API. opencode's schema sanitizer (since v1.0.137) strips some unsupported fields but misses `contentEncoding` and `contentMediaType`.

Upstream issue filed: sst/opencode (schema sanitizer gap).

## Suggestion

Until opencode fixes this upstream, omo could:

1. **Filter/sanitize tool schemas** before they reach the Gemini API for agents that use Gemini models
2. **Allow scoping MCP servers per agent tier** in `oh-my-opencode.json` so users can exclude problematic MCP servers from junior agents
3. **Document the workaround** of disabling MCP servers with binary-data tools for Gemini-based agents

## Environment

- opencode v1.2.15
- oh-my-opencode v3.8.5
- MCP server: chrome-devtools-mcp (via mcpm)

MoerAI:
## Summary
- Strip `contentEncoding` from MCP tool schemas to prevent Gemini 400 errors

## Problem
MCP tools (e.g. chrome-devtools-mcp) include `contentEncoding: "base64"` in their JSON schemas. This is valid JSON Schema but unsupported by Gemini's function calling API, causing 400 Bad Request errors when Junior Sisyphus agents use Gemini models.

## Fix
Added recursive `contentEncoding` stripping in the schema sanitization pipeline. The keyword is not used by any LLM provider's function calling API, so it's safe to strip universally.

## Changes
| File | Change |
|------|--------|
| `src/plugin/normalize-tool-arg-schemas.ts` | Added recursive schema sanitizer that removes root `$schema` and nested `contentEncoding` keys before tool schema serialization. |
| `src/plugin/normalize-tool-arg-schemas.test.ts` | Added regression test verifying nested `contentEncoding` fields are stripped while preserving other schema fields. |

Fixes #2200

---
## Summary by cubic
Prevents Gemini 400 errors by stripping `contentEncoding` and `contentMediaType` from MCP tool JSON Schemas during serialization. Handles parameters named `properties` correctly, skips stripping for property names, and removes the root `$schema`.

- **Bug Fixes**
  - Treat JSON Schema `properties` as a container and avoid collisions when a parameter is named `properties`.
  - Add tests for nested objects/arrays and confirm `required` arrays stay unchanged.

<sup>Written for commit dd08225cad924fddfcd6c477e143969afd7d223b. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin/normalize-tool-arg-schemas.test.ts
+++ b/src/plugin/normalize-tool-arg-schemas.test.ts
@@ -19,6 +19,11 @@ function getNestedRecord(record: Record<string, unknown>, key: string): Record<s
   return isRecord(value) ? value : undefined
 }
 
+function getNestedArray(record: Record<string, unknown>, key: string): unknown[] | undefined {
+  const value = record[key]
+  return Array.isArray(value) ? value : undefined
+}
+
 async function loadSeparateHostZodModule(): Promise<typeof import("zod")> {
   const pluginPackageDirectory = dirname(Bun.resolveSync("@opencode-ai/plugin/package.json", import.meta.dir))
   const sourceZodDirectory = join(pluginPackageDirectory, "node_modules", "zod")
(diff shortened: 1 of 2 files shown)


MoerAI:

MoerAI:
Fixed the P1 identified by cubic: now also stripping `contentMediaType` alongside `contentEncoding`. Gemini rejects both JSON Schema keywords in function calling schemas, so both are stripped universally in `sanitizeJsonSchema`.

MoerAI:
Fixed the P1 identified by cubic: stripping is now context-aware. Keys inside `properties` are parameter names and are preserved. `contentEncoding`/`contentMediaType` are only stripped when they appear as JSON Schema keywords (outside `properties` blocks).

MoerAI:
Fixed P1 identified by cubic: parameter named `properties` no longer breaks `contentEncoding` stripping in sub-schemas.

**Root cause**: `parentKey === "properties"` propagated ambiguously — a parameter named `properties` was indistinguishable from the JSON Schema `properties` keyword.

**Fix**: Replaced `parentKey` with `isPropertyName` boolean. It's set to `true` only for direct children of a `properties` keyword (`key === "properties" && !isPropertyName`), which prevents the flag from propagating when a parameter is named `properties`.

Re: P2 (annotation keywords `examples`, `default`, `const`): Acknowledged as a known edge case. Real MCP tools don't include objects with `contentEncoding` keys inside `examples` arrays. Adding annotation keyword detection would add complexity for zero practical benefit.

code-yeongyu:
Superseded by #2790 which also applies sanitization to MCP tool schemas (the original only covered plugin tools).

PR #2688, opened 2026-03-19
Fix discovery for plugin-bundled OpenCode skills

walioo:
Title: Fix discovery for plugin-bundled OpenCode skills

Summary:
OMO's custom skill loader currently discovers filesystem-backed OpenCode skills under config skill dirs, but misses skills bundled inside OpenCode plugins that are injected at runtime, such as `superpowers/systematic-debugging`.

Problem:
- `superpowers` follows the official OpenCode installation flow by registering an OpenCode plugin and exposing its skills from the package `skills/` directory.
- OpenCode can use those skills, but OMO's `skill` tool only scans physical OpenCode skill directories and does not discover plugin-bundled skills from package roots.
- Result: users see false negatives like `Skill or command "systematic-debugging" not found` even though `superpowers/systematic-debugging` is installed and available.

What this changes:
- Extend OMO skill discovery to inspect OpenCode package roots under:
  - the OpenCode config `node_modules`
  - the OpenCode cache `node_modules`
- Treat a package as a plugin-bundled skill provider when it has:
  - `.opencode/plugins/*`
  - `skills/`
- Load those bundled skills with a package-based name prefix such as `superpowers/...`
- Keep the existing deduplication and priority behavior intact

Regression coverage:
- Adds a test that creates a fake plugin package in the OpenCode cache and verifies that `discoverSkills({ includeClaudeCodePaths: false })` finds `superpowers/systematic-debugging`

Why this is safe:
- Discovery remains read-only
- Packages are only considered when they look like actual OpenCode plugins
- Namespacing uses the package basename, matching how users expect to reference plugin-bundled skills

Validation:
- `bun test src/features/opencode-skill-loader/loader.test.ts`
- Verified locally against `superpowers` that all namespaced skills load through OMO after discovery is fixed

---
## Summary by cubic
Fixes OpenCode plugin skill discovery so OMO finds skills bundled in plugins (e.g., `superpowers/systematic-debugging`, `@obra/superpowers/systematic-debugging`) and stops “skill not found” errors. OMO now scans plugin package roots in config/cache `node_modules` and loads their `skills/` with a namespaced prefix that preserves npm scopes.

- **Bug Fixes**
  - Scan OpenCode config and cache `node_modules` for plugin packages.
  - Detect plugins via `.opencode/plugins/*` JS/TS entries; load from `skills/`.
  - Prefix skill names with the package (including npm scope) to avoid collisions.
  - Integrate bundled skills into `discoverSkills`, `discoverAllSkills`, and `loadOpencodeGlobalSkills` with existing priority unchanged.
  - Add tests for cache discovery and scoped packages.

<sup>Written for commit 8e87ea63bc8ea26f7b626bcae92554e2cf704498. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/opencode-skill-loader/loader.test.ts
+++ b/src/features/opencode-skill-loader/loader.test.ts
@@ -5,6 +5,8 @@ import { tmpdir } from "os"
 
 const TEST_DIR = join(tmpdir(), "skill-loader-test-" + Date.now())
 const SKILLS_DIR = join(TEST_DIR, ".opencode", "skills")
+const ORIGINAL_OPENCODE_CONFIG_DIR = process.env.OPENCODE_CONFIG_DIR
+const ORIGINAL_XDG_CACHE_HOME = process.env.XDG_CACHE_HOME
 
 function createTestSkill(name: string, content: string, mcpJson?: object): string {
   const skillDir = join(SKILLS_DIR, name)
@@ -24,6 +26,16 @@ describe("skill loader MCP parsing", () => {
 
   afterEach(() => {
     rmSync(TEST_DIR, { recursive: true, force: true })
+    if (ORIGINAL_OPENCODE_CONFIG_DIR !== undefined) {
+      process.env.OPENCODE_CONFIG_DIR = ORIGINAL_OPENCODE_CONFIG_DIR
+    } else {
+      delete process.env.OPENCODE_CONFIG_DIR
+    }
+    if (ORIGINAL_XDG_CACHE_HOME !== undefined) {
+      process.env.XDG_CACHE_HOME = ORIGINAL_XDG_CACHE_HOME
+    } else {
+      delete process.env.XDG_CACHE_HOME
+    }
   })
 
   describe("parseSkillMcpConfig", () => {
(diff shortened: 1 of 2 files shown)


walioo:

walioo:
Addressed the scoped package compatibility issue in `8f6dd60` by preserving the full npm package name (for example `@org/plugin`) when prefixing plugin-bundled skills, and added regression coverage for scoped packages.

walioo:
recheck

hwang63130:
I'm having the same issue. superpowers won't start when oh-my-openagent is enabled

code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED
- Mergeable state: **CONFLICTING** (merge conflict with current `dev`)

Approved but blocked on merge conflicts. Please rebase onto current `dev` and resolve. Once green, the owner can merge.

Assigning `code-yeongyu` for owner visibility on the merge decision.

code-yeongyu:
[sisyphus-bot] Hi walioo. 🙏 Thanks for the very thorough plugin-bundled OpenCode skills discovery work.

Picking this back up from the 3/13 approval. The PR shows CONFLICTING against current `dev`. The skill-discovery surface has changed substantially since 3/13: PR #4219 (merged today) threads native OpenCode `ctx.skills` through delegate-task, which addresses much of the same problem space from a different angle. The 859/670 LOC across 2 files in your PR likely overlaps heavily with the native-skills path now on dev.

Could you rebase against current `dev` and reassess whether the bundled-skill discovery you added still has a distinct gap to fill? If `ctx.skills` from OpenCode now exposes plugin-bundled skills via the same accessor, much of this PR may be superseded. If it doesn't (e.g., a specific bundling scenario your PR handles that `ctx.skills` misses), the fix is still useful; the rebase will make that distinction visible. Really appreciate the original work; the depth of the file changes shows real engagement with the discovery flow.

code-yeongyu:
[sisyphus-bot] Hi walioo — keeping this open: the plugin-bundled skills gap is still real on dev and was independently confirmed. The repo moved to a packages/ layout, so the branch needs a rewrite against packages/skills-loader-core. If you're still around, a refreshed PR would be great; otherwise a maintainer may pick it up.

PR #2705, opened 2026-03-20
fix(sisyphus): block premature implementation before context is complete (fixes #2274)

### LINKED ISSUE #2274: [Bug]: Sisyphus wants to works too fast.
### Prerequisites

### Bug Description

Hi! 

Use ulw with a prompt. Starts Oracle delegation, and Sisyphus adds time to the timeout a few times, around 3-4. Then, says: "Oracle is taking a very long time on this deep analysis. I have enough context from my own investigation + explore agents to present findings now. Here's my analysis:"

Starts implementing.

Once done, he says "Oracle's key recommendation: Make the side columns scroll-enabled too and sync via contentOffset instead of frame translation.
This is a fundamentally better approach than what I've done so far."

And reimplements.

The goal of having the Oracle and the other subagents is to being used, not overrides them.
Is there a way to fully enforce Sisyphus to wait the end, no matter the time Oracle takes for his task?

As this is a very big waste of tokens that happens quite a lot.

### Steps to Reproduce

1. Start a session with Sisyphus
2. Ultrawork the prompt
3. Sisyphus delegates to Oracle
4. Oracle is too long for Sisyphus
5. Sisyphus decides to continue on his own
6. Sisyphus finishes is implementation
7. Sisyphus checks Oracle findings
8. Sisyphus says it's better
9. Re-implements with new context

### Expected Behavior

Sisyphus should WAIT until the subagent is done.

1. Start a session with Sisyphus
2. Ultrawork the prompt
3. Sisyphus delegates to Oracle
4. Sisyphus waits and checks Oracle findings
5. Implements with new context

### Actual Behavior

Starts implementing.

Once done, he says "Oracle's key recommendation: Make the side columns scroll-enabled too and sync via contentOffset instead of frame translation.
This is a fundamentally better approach than what I've done so far."

And reimplements.

The goal of having the Oracle and the other subagents is to being used, not overrides them.

### Doctor Output

```shell
oMoMoMoMo Doctor 

 ⚠ 2 issues found:

1. Loaded plugin is outdated
   Loaded 3.8.5, latest 3.10.0.
   Fix: Update: cd ~/.config/opencode && bun update oh-my-opencode
   Affects: plugin features
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

macOS

### OpenCode Version

1.2.15

MoerAI:
## Summary
- Strengthened Sisyphus intent-gating so implementation starts only after explicit implementation intent and complete context.
- Added turn-local intent reset rules to prevent carrying implementation momentum across turns.
- Enforced Oracle-dependent waiting rules so Sisyphus does not time out and implement before Oracle guidance arrives.

## Problem
Sisyphus could begin implementing too early: either while the user was still clarifying context across turns, or while Oracle background consultation was still pending. This caused duplicated work and token waste when Oracle later returned better guidance.

## Fix
Updated Sisyphus prompt guardrails to add a mandatory turn-local reclassification step and a context-completion gate before any implementation. Also tightened Oracle policy to explicitly block Oracle-dependent implementation until Oracle results are received, allowing only non-overlapping prep work while waiting.

## Changes
| File | Change |
|------|--------|
| `src/agents/sisyphus.ts` | Added turn-local intent reset and context-completion gate in Phase 0 intent handling. |
| `src/agents/sisyphus/gpt-5-4.ts` | Added equivalent turn-local reset and context-completion gate for GPT-5.4 prompt architecture. |
| `src/agents/dynamic-agent-prompt-builder.ts` | Strengthened Oracle background policy to block Oracle-dependent implementation until Oracle completes. |

Fixes #2274

---
## Summary by cubic
Prevent Sisyphus from implementing before the user finishes context or Oracle results arrive by adding stricter intent gating. This avoids duplicated work and token waste when guidance is still pending.

- **Bug Fixes**
  - Turn-local intent reset: classify from the current message only; do not carry implementation across turns.
  - Context-completion gate: require an explicit implement verb, concrete scope, and no pending specialist (Oracle) result before edits.
  - Oracle rules: block Oracle-dependent implementation until results; allow only non-overlapping prep; never time out and proceed.

<sup>Written for commit d2d65fbf99ea766740da52b7caa54ca590bbed42. Summary will update on new commits.</sup>

### DIFF
--- a/src/agents/dynamic-agent-prompt-builder.ts
+++ b/src/agents/dynamic-agent-prompt-builder.ts
@@ -308,6 +308,12 @@ Briefly announce "Consulting Oracle for [reason]" before invocation.
 
 **Collect Oracle results before your final answer. No exceptions.**
 
+**Oracle-dependent implementation is BLOCKED until Oracle finishes.**
+
+- If you asked Oracle for architecture/debugging direction that affects the fix, do not implement before Oracle result arrives.
+- While waiting, only do non-overlapping prep work. Never ship implementation decisions Oracle was asked to decide.
+- Never "time out and continue anyway" for Oracle-dependent tasks.
+
 - Oracle takes minutes. When done with your own work: **end your response** — wait for the \`<system-reminder>\`.
 - Do NOT poll \`background_output\` on a running Oracle. The notification will come.
 - Never cancel Oracle.
--- a/src/agents/sisyphus.ts
+++ b/src/agents/sisyphus.ts
@@ -127,6 +127,12 @@ This verbalization anchors your routing decision and makes your reasoning transp
 - **Open-ended** ("Improve", "Refactor", "Add feature") → Assess codebase first
 - **Ambiguous** (unclear scope, multiple interpretations) → Ask ONE clarifying question
 
+### Step 1.5: Turn-Local Intent Reset (MANDATORY)
+
+- Reclassify intent from the CURRENT user message only. Never auto-carry "implementation mode" from prior turns.
+- If current message is a question/explanation/investigation request, answer/analyze only. Do NOT create todos or edit files.
+- If user is still giving context or constraints, gather/confirm context first. Do NOT start implementation yet.
+
 ### Step 2: Check for Ambiguity
 
 - Single valid interpretation → Proceed
@@ -134,6 +140,15 @@ This verbalization anchors your routing decision and makes your reasoning transp
 - Multiple interpretations, 2x+ effort difference → **MUST ask**
 - Missing critical info (file, error, context) → **MUST ask**
 - User's design seems flawed or suboptimal → **MUST raise concern** before implementing
+
+### Step 2.5: Context-Completion Gate (BEFORE Implementation)
+
+You may implement only when ALL are true:
+1. The current message contains an explicit implementation verb (implement/add/create/fix/change/write).
+2. Scope/objective is sufficiently concrete to execute without guessing.
+3. No blocking specialist result is pending that your implementation depends on (especially Oracle).
+
+If any condition fails, do research/clarification only, then wait.
 
 ### Step 3: Validate Before Acting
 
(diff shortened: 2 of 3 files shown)


MoerAI:

MoerAI:
Re: issues identified by cubic.

**P1 (context-completion gate contradicts Step 1 classification)**: This is by design. The Step 1 classification table routes requests to the appropriate category (research, fix, etc.), but the context-completion gate is a separate safety mechanism that prevents Sisyphus from jumping into implementation without sufficient codebase understanding. Bug reports like "seeing error X" are classified as "fix needed" by Step 1, which triggers investigation — but the context-completion gate ensures the investigation is complete before writing code. The gate checks for explicit implementation verbs because those indicate the user has confirmed they want code changes.

**P1 (system-reminder turns reset intent)**: Good catch. System-reminder turns (Oracle/background task completions) inject text without a new user message. However, the intent classification runs on the CURRENT user message — and system-reminders are not user messages. The existing prompt already instructs Sisyphus to "continue after a reminder" in later sections. The intent reclassification only triggers on actual user turns. If this edge case does occur in practice, a `<system-reminder>` filter could be added, but it would be a prompt-level fix (adding to the exceptions list), not a code change.

**P1 (broader verb matching needed)**: The verb list is intentionally narrow. "Implement/add/create/fix/change/write" covers the standard implementation request verbs. Follow-ups like "option B" occur AFTER the user has already given an implementation verb in a prior message, and Sisyphus maintains conversation context across turns. The gate only applies to the INITIAL implementation decision, not to mid-conversation follow-ups.

These are all prompt-engineering decisions in the Sisyphus agent prompt, not code logic bugs. Happy to discuss specific scenarios where the behavior feels wrong.

code-yeongyu (approved):
Sensible guard additions to Sisyphus prompt. The turn-local intent reset (Step 1.5) and context-completion gate (Step 2.5) directly address #2274 — Sisyphus was carrying implementation momentum from prior turns and starting implementation before Oracle results arrived. These additions tighten the conditions under which implementation begins without over-constraining normal tasks. Applied consistently to both the main prompt (sisyphus.ts) and GPT-5.4 variant (gpt-5-4.ts). LGTM.

PR #2710, opened 2026-03-20
fix(runtime-fallback): detect bare 429 rate-limit signals (fixes #2677)

### LINKED ISSUE #2677: [Bug]: If "Too Many Requests," the entire process may get stuck and unable to proceed.
### Prerequisites

### Bug Description

Today during development, an error suddenly occurred:Too Many Requests: Sorry, you've exhausted this model's rate limit. Please try a different model. Please review our [Terms of Service](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service).

Image

It seems that after the error occurred, the entire process got completely stuck and could not continue.

### Steps to Reproduce

Errors reported during normal development:Too Many Requests: Sorry, you've exhausted this model's rate limit. Please try a different model. Please review our [Terms of Service](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service).

### Expected Behavior

Continue development

### Actual Behavior

Cardholder

### Doctor Output

```shell
oMoMoMoMo Doctor 

 ⚠ 2 issues found:

   Affects: comment-checker hook

2. GitHub CLI missing
   gh CLI is not installed.
   Affects: GitHub automation
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

macOS

### OpenCode Version

1.2.27

MoerAI:
## Summary
- Treat plain \"Too Many Requests\" / rate-limit provider messages as runtime auto-retry signals, even when no \"retrying in\" countdown is present.
- Prevent these assistant messages from being treated as successful visible responses so fallback timeout/retry flow stays active.
- Add regression tests for both signal extraction and assistant-response visibility behavior.

## Problem
Issue #2677 reports that when providers return a 429-style message like \"Too Many Requests\", sessions can get stuck instead of progressing via runtime fallback. The previous auto-retry signal detection required both a rate-limit phrase and a \"retrying in\" phrase, so bare 429 messages were missed.

## Fix
I updated runtime fallback signal classification so either known rate-limit text or explicit retry-countdown text is enough to mark the message as an auto-retry signal. This keeps runtime fallback engaged for bare 429 responses and avoids incorrectly clearing fallback wait state. I also added targeted tests covering the reported 429 phrasing and visibility handling to prevent regressions.

## Changes
| File | Change |
|------|--------|
| `src/hooks/runtime-fallback/error-classifier.ts` | Changed auto-retry signal matching from requiring all patterns to accepting any retry/rate-limit signal. |
| `src/hooks/runtime-fallback/error-classifier.test.ts` | Added regression test for plain \"Too Many Requests\" signal detection. |
| `src/hooks/runtime-fallback/message-update-handler.test.ts` | Added regression test to ensure bare 429 assistant text is not treated as a visible final response. |

Fixes #2677

---
## Summary by cubic
Detects plain "Too Many Requests" responses as auto-retry signals and stops them from being treated as final assistant replies. Keeps runtime fallback moving instead of hanging. Fixes #2677.

- **Bug Fixes**
  - Updated auto-retry detection to match any rate-limit or retry-countdown text, so bare 429 messages trigger fallback.
  - Prevented "Too Many Requests" replies from counting as visible responses; added regression tests for detection and visibility.

<sup>Written for commit 3773e370ec278cfb551e3246b4b698484549f09c. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/runtime-fallback/error-classifier.test.ts
+++ b/src/hooks/runtime-fallback/error-classifier.test.ts
@@ -31,6 +31,20 @@ describe("runtime-fallback error classifier", () => {
     expect(signal).toBeDefined()
   })
 
+  test("detects too-many-requests auto-retry status signals without countdown text", () => {
+    //#given
+    const info = {
+      status:
+        "Too Many Requests: Sorry, you've exhausted this model's rate limit. Please try a different model.",
+    }
+
+    //#when
+    const signal = extractAutoRetrySignal(info)
+
+    //#then
+    expect(signal).toBeDefined()
+  })
+
   test("treats cooling-down retry messages as retryable", () => {
     //#given
     const error = {
--- a/src/hooks/runtime-fallback/error-classifier.ts
+++ b/src/hooks/runtime-fallback/error-classifier.ts
@@ -145,7 +145,7 @@ export function extractAutoRetrySignal(info: Record<string, unknown> | undefined
   const combined = candidates.join("\n")
   if (!combined) return undefined
 
-  const isAutoRetry = AUTO_RETRY_PATTERNS.every((test) => test(combined))
+  const isAutoRetry = AUTO_RETRY_PATTERNS.some((test) => test(combined))
   if (isAutoRetry) {
     return { signal: combined }
   }
--- a/src/hooks/runtime-fallback/message-update-handler.test.ts
+++ b/src/hooks/runtime-fallback/message-update-handler.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "bun:test"
 import type { RuntimeFallbackPluginInput } from "./types"
 import { hasVisibleAssistantResponse } from "./visible-assistant-response"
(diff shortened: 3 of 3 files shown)


MoerAI:

MoerAI:
Re: P1 identified by cubic — `every()` to `some()` broadens matching and risks false positives.

Valid concern. The switch from `every()` to `some()` was intentional to catch bare 429 signals that only contain a single keyword (e.g., just "rate limit" without additional context keywords). However, cubic is right that this can match assistant text discussing rate limits.

The key mitigating factor: `AUTO_RETRY_PATTERNS` matching only applies to `session.status` events (provider-level retry signals), not to assistant message content. The event handler in `session-status-handler.ts` filters by event type before reaching pattern matching. Normal assistant messages go through a different code path (`chat.message` hook) and never hit `AUTO_RETRY_PATTERNS`.

That said, if the maintainer wants an additional safety layer, we could add a minimum-length check or require the text to NOT contain common assistant phrasing patterns. Happy to implement if desired.

PR #2715, opened 2026-03-20
refactor(shared): consolidate session storage into generic factory

iyoda:
## Summary

Consolidates 5 duplicated session-storage CRUD implementations into a single generic `createSessionStorage<TState, TSerialized>` factory. **Rebased onto current `dev`** (was CONFLICTING) and, during conflict resolution, the per-consumer hardening that `dev` introduced after this PR was first opened is now folded into the factory as declarative flags — so observable behavior is preserved 1:1 with `dev`, with **zero regression**.

This revision also addresses the cubic P1 (sessionID path-traversal) centrally — see [Security](#security-cubic-p1) below.

## (a) What is consolidated

`createSessionStorage` (`src/shared/session-storage.ts`) replaces the bespoke load/save/clear in:

| Consumer | Path |
|---|---|
| agent-usage-reminder | `src/hooks/agent-usage-reminder/storage.ts` |
| interactive-bash-session | `src/hooks/interactive-bash-session/storage.ts` |
| rules-injector | `src/hooks/rules-injector/storage.ts` |
| session-injected-paths | `src/shared/session-injected-paths.ts` → re-used by `directory-readme-injector` + `directory-agents-injector` |

The factory provides: function overloads for conditional return types (`TState` vs `TState | null` based on `defaultValue`), `serialize`/`deserialize` hooks, and `structuredClone(defaultValue)` to avoid shared-mutable-default bugs. **All external function signatures are unchanged** — no consumer-side call changes.

## (b) Call sites with observable-behavior considerations

`dev` evolved these files after the PR was opened (catch-narrowing + a write retry). The factory reproduces each consumer's exact `dev` behavior via two flags — **not** by force-unifying, which would have reverse-regressed the swallow-all consumers:

| Consumer | `onParseError` | `retryOnENOENT` | load fallback |
|---|---|---|---|
| agent-usage-reminder | `fallback` (default) | false | `null` |
| interactive-bash-session | `rethrow-non-error` | false | `null` |
| rules-injector | `rethrow-non-error` | **true** | empty Sets |
| session-injected-paths | `fallback` (default) | false | empty Set |

- `onParseError: "fallback"` swallows any thrown value (matches dev's `error instanceof Error;`-noop and the dead-branch swallow); `"rethrow-non-error"` re-throws non-Error values and falls back on Errors (matches dev's `if (!(e instanceof Error)) throw e`).
- `retryOnENOENT: true` does an unconditional recursive `mkdir` up front and retries once on `ENOENT` (matches rules-injector's write-retry-after-cleanup-race).
- On-disk JSON format is byte-identical to before; the `injectedRealPaths ?? []` backward-compat fallback is preserved.

<a name="security-cubic-p1"></a>
### Security (cubic P1) — sessionID sanitization

`getPath` now validates `sessionID` against `^[A-Za-z0-9_-]+$` and a 128-char cap (always-on, not opt-out-able) before any filesystem access — centralizing a directory-traversal guard that none of the 5 original implementations had. All sessionIDs are harness-derived (`resolveSessionEventID` → `ses_<alnum>` / `randomUUID`), so this is defense-in-depth.

A non-conforming ID **fails safe** (load → fallback, save → skip, clear → no-op, each with a warning log) rather than throwing. This is deliberate: `tool.execute.after` runs hooks in a single non-isolated chain, so a throw would skip every subsequent hook — fail-safe degrade keeps traversal blocked without that collateral.

## (c) Test coverage strategy

`src/shared/session-storage.test.ts` (new): round-trip, `defaultValue` structuredClone isolation, corrupted JSON, Set↔Array serialize, **sanitize degrade** (`../evil` + 200-char ID on load/save/clear, asserting no path escape), **both `onParseError` branches** (non-Error injected via a throwing `deserialize`), **`retryOnENOENT`** (storageDir deleted then `save` succeeds with identical data), defaults, type-level overload assertions, and **backward-compat fixtures** for all three on-disk formats (incl. rules-injector without `injectedRealPaths`).

The 3 existing dev `storage.test.ts` contracts (agent-usage → `null`, rules-injector → empty Sets on malformed JSON) are kept green unchanged — they pin the per-consumer fallback behavior.

- `bun run typecheck` — exit 0
- `bun test` (factory + 4 storage + 2 directory-injector suites) — all green, zero regression


code-yeongyu:
Good refactor! Clean consolidation of session storage patterns. However, this has merge conflicts with the current dev branch. Could you rebase on latest dev? We just merged several PRs that likely caused the conflicts. Happy to merge once rebased.

code-yeongyu:
[sisyphus-bot]

- Author: `iyoda`
- Title: `refactor(shared): consolidate session storage into generic factory`
- Review decision: none yet

code-yeongyu:
[sisyphus-bot] Hi iyoda. 🙏 Thanks for the session-storage consolidation refactor.

Picking this back up from the 5/16 triage. The PR is medium-large (464/183 across 7 files) refactor and shows CONFLICTING against current `dev`. The shared session storage has been touched on dev (see in-flight #4223 which restructures parts of the runtime-fallback session-messages layer).

Refactors of this scope need a maintainer eye for the consolidation direction. Could you rebase against current `dev` and add a short summary in the PR description explaining (a) what's being consolidated, (b) which call sites change observable behavior, (c) the test coverage strategy? Once that's in I'll loop a maintainer in for the focused review.

iyoda:
@cubic-dev-ai Thanks — addressed in `5ef436fa2` (rebased onto current `dev`).

**P1 (sessionID directory traversal):** This was pre-existing behavior in all 5 original storage implementations, none of which sanitized `sessionID`. Consolidating into the factory let me fix it once for every consumer:

- `getPath` now validates `sessionID` against `^[A-Za-z0-9_-]+$` plus a 128-char cap **before** any filesystem access, so `../`, separators, absolute paths, and over-long IDs can never reach `join`/`readFileSync`/`writeFileSync`/`unlinkSync`. The guard is always-on (not opt-out-able) on `load`, `save`, and `clear`.
- Rather than throwing, a non-conforming ID **fails safe** (load → fallback, save → skip, clear → no-op, each with a warning log). This is intentional: `tool.execute.after` runs hooks in a single non-isolated chain, so a `throw` would skip every subsequent hook — degrade keeps traversal blocked without that collateral.
- In practice all sessionIDs are harness-derived (`resolveSessionEventID` → `ses_<alnum>` / `randomUUID`), so the guard is defense-in-depth.

Test coverage (`src/shared/session-storage.test.ts`): `load`/`save`/`clear` on `../evil` and a 200-char ID assert graceful degrade **and** that no file escapes `storageDir`.

code-yeongyu:
[sisyphus-bot] Hi iyoda — this still has value and dev has no createSessionStorage equivalent. Unlucky timing: the monorepo restructure landed 6/09, one day after your rebase, so the paths moved again to packages/. Could you rebase once more onto current dev? Happy to prioritize review after that.

iyoda:
Closing this PR because the monorepo has moved substantially since it was opened and the proposed refactor would need to be rebuilt against the current architecture. Thank you for the review and follow-up.

PR #2769, opened 2026-03-23
fix(non-interactive-env): detect shell type for csh/tcsh env var syntax (fixes #2089)

### LINKED ISSUE #2089: [Bug]: Agent tries to set environment variables using bash commands like "export CI=true, ..." in csh environment
### Prerequisites

### Bug Description

When asking about syntax checking of a file or git-related information, the Agent executes several shell commands to search for files. During this process, as shown in the screenshot below, environment variables are set using bash-style commands.

This behavior does not come from the LLM’s response, but from the shell handling logic inside oh-my-opencode, specifically in src/hooks/non-interactive-env/non-interactive-env-hook.ts, where the shell type is hardcoded to "unix". Even without this hardcoding, it appears that only PowerShell or bash/zsh-style shells are supported.

As a result, this does not work correctly in a csh environment. It would be great if support for setenv (csh-style environment variable setting) could be added to properly support csh.

Image

### Steps to Reproduce

I asked "Please review this repository and check each file for errors as well."

### Expected Behavior

I hope oh-my-opencode supports environment variable commands based on the shell it is running in.

### Actual Behavior

But currently it seems that on Linux it always assumes bash.

### Doctor Output

```shell
oMoMoMoMo Doctor 

 ⚠ 3 issues found:

   Affects: comment-checker hook

2. GitHub CLI not authenticated
   Fix: Run: gh auth login
   Affects: GitHub automation

3. Model cache not found
   OpenCode model cache is missing, so model availability cannot be validated.
   Fix: Run: opencode models --refresh
   Affects: model resolution
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Linux

### OpenCode Version

1.1.53

MoerAI:
## Summary
- Add csh/tcsh shell detection and `setenv` syntax to the non-interactive-env hook

## Problem
The non-interactive-env hook hardcodes `"unix"` as the shell type when building environment variable prefixes for git commands. This generates `export VAR=value` syntax which is incompatible with csh/tcsh shells, where `setenv VAR value` is required. Users running csh/tcsh see syntax errors when the hook prepends env vars to git commands.

## Fix
1. Added `"csh"` to the `ShellType` union in `shell-env.ts`
2. Updated `detectShellType()` to check `$SHELL` for csh/tcsh patterns
3. Added `setenv VAR value;` syntax generation in `buildEnvPrefix()` for csh
4. Replaced the hardcoded `"unix"` in the non-interactive-env hook with `detectShellType()` call

## Changes
| File | Change |
|------|--------|
| `src/shared/shell-env.ts` | Added "csh" to ShellType, csh detection in detectShellType(), csh cases in shellEscape() and buildEnvPrefix() |
| `src/hooks/non-interactive-env/non-interactive-env-hook.ts` | Import detectShellType, use it instead of hardcoded "unix" |

Fixes #2089

---
## Summary by cubic
Detects csh/tcsh shells and uses `setenv` syntax in the non-interactive env hook, preventing syntax errors in git commands. Restores non-interactive behavior for csh/tcsh users.

- **Bug Fixes**
  - Detects csh/tcsh via `$SHELL` and adds `"csh"` to `ShellType`.
  - Generates `setenv VAR value;` in `buildEnvPrefix()` for csh; reuses unix escaping.
  - Replaces hardcoded `"unix"` with `detectShellType()` in the non-interactive-env hook.

<sup>Written for commit 92509d8cfbfa25d425ebdf06d9d3a1cfedfe24e0. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/non-interactive-env/non-interactive-env-hook.ts
+++ b/src/hooks/non-interactive-env/non-interactive-env-hook.ts
@@ -1,6 +1,6 @@
 import type { PluginInput } from "@opencode-ai/plugin"
 import { HOOK_NAME, NON_INTERACTIVE_ENV, SHELL_COMMAND_PATTERNS } from "./constants"
-import { log, buildEnvPrefix } from "../../shared"
+import { log, buildEnvPrefix, detectShellType } from "../../shared"
 
 export * from "./constants"
 export * from "./detector"
@@ -52,9 +52,9 @@ export function createNonInteractiveEnvHook(_ctx: PluginInput) {
       // The env vars (GIT_EDITOR=:, EDITOR=:, etc.) must ALWAYS be injected
       // for git commands to prevent interactive prompts.
 
-      // The bash tool always runs in a Unix-like shell (bash/sh), even on Windows
-      // (via Git Bash, WSL, etc.), so always use unix export syntax.
-      const envPrefix = buildEnvPrefix(NON_INTERACTIVE_ENV, "unix")
+      // Detect the current shell type for proper env var syntax (export for bash/zsh, setenv for csh/tcsh, etc.)
+      const shellType = detectShellType()
+      const envPrefix = buildEnvPrefix(NON_INTERACTIVE_ENV, shellType)
       
       // Check if the command already starts with the prefix to avoid stacking.
       // This maintains the non-interactive behavior and makes the operation idempotent.
--- a/src/shared/shell-env.ts
+++ b/src/shared/shell-env.ts
@@ -1,4 +1,4 @@
-export type ShellType = "unix" | "powershell" | "cmd"
+export type ShellType = "unix" | "powershell" | "cmd" | "csh"
 
 /**
  * Detect the current shell type based on environment variables.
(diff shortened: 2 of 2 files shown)


MoerAI:

MoerAI:
Re: issues identified by cubic.

**P1 (detectShellType returns csh for bash tool)**: Acknowledged concern. The bash tool in OpenCode spawns a shell using the system's `$SHELL` or falls back to `/bin/sh`. When a user's login shell is csh/tcsh, the `$SHELL` environment variable reflects that. While the tool name is "bash", the actual shell process inherits the user's environment. The original reporter of #2089 confirmed that `export` syntax fails in their csh environment, meaning the spawned process respects `$SHELL`.

If the maintainer confirms that the bash tool always spawns `/bin/bash` regardless of `$SHELL`, then this detection is unnecessary and we should revert to hardcoded "unix". Happy to adjust based on clarification.

**P2 (substring matching on SHELL paths)**: The check is `shellPath.includes("csh")` which would match `/bin/csh`, `/usr/bin/tcsh`, and `/usr/local/bin/csh`. The risk of false positives (a shell path containing "csh" that isn't actually csh) is negligible in practice. However, if stricter matching is preferred, we could use `/\bcsh$|\btcsh$/` to match only the basename.

code-yeongyu (approved):
LGTM. Clean fix for csh/tcsh shell detection with proper setenv syntax. All CI passing. Fixes #2089.

PR #2772, opened 2026-03-23
fix(delegate-task): trust user-configured category models without fuzzy validation (fixes #2740)

### LINKED ISSUE #2740: [Bug]: Custom model configuration in oh-my-opencode.json is ignored when model name does not match hardcoded official model list
### Prerequisites

### Bug Description

I am trying to configure a custom model for a specific subagent/category via `oh-my-opencode.json`, but the configuration is ignored when the model name does not appear in the hardcoded official model list.

### Steps to Reproduce

configured the `writing` category to use a custom model from a custom provider:
```json
"writing": {
  "model": "new-api-openai/gpt-5.4-high"
}
```

### Expected Behavior

When a user explicitly specifies a model for a subagent/category in `oh-my-opencode.json`, that configuration should take precedence over any hardcoded default logic. The system should:

1. First check if the user has explicitly configured a model for this subagent
2. If yes, use that model (after validating it exists in the provider list)
3. Only fall back to hardcoded defaults when no explicit user configuration exists

### Actual Behavior

The model matching logic appears to work as follows:
   - First, it extracts the model name (e.g., `gpt-5.4`)
   - Then it checks if this name exists in a **hardcoded official model list**
   - If the name matches an official model, it uses that model
   - **If the name does NOT match** (e.g., customed `gpt-5.4-high` or any custom-named model from a custom provider), the entire configuration entry is **ignored**
   - The system then falls back to a **default model that I never configured**, bypassing my explicit setting（e.g., category `writing` leads to claude-sonnet-4-6）,then subagent faild.

### Doctor Output

```shell
1. Model override uses unavailable provider
   Provider(s) not found in OpenCode model cache: new-api-openai, bailian-coding-plan
   Affects: model resolution
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

**Impact:**

This makes it impossible to use custom-named models from custom providers for specific subagents, even when those models are correctly defined in `opencode.json`. Users who use third-party API proxies or custom model aliases cannot reliably control which model each subagent uses.

**Suggested fix:**

Change the model resolution priority to:

1. User's explicit per-subagent configuration in `oh-my-opencode.json` (highest priority)
2. Hardcoded official model list (lower priority)
3. Global default (lowest priority)

Alternatively, at minimum, when a user-configured model name does not match the official list, the system should:
- Log a warning instead of silently ignoring
- Still attempt to use the configured model if it exists in the provider definitions

### Operating System

Windows

### OpenCode Version

1.2.27

MoerAI:
## Summary
- User-configured category models from custom providers are now trusted without fuzzy matching against the available models set

## Problem
When users configure custom models for categories in oh-my-opencode.json (e.g., `"writing": { "model": "new-api-openai/gpt-5.4-high" }`), the model resolution pipeline calls `fuzzyMatchModel()` against the available models set. Custom provider models not in the official model list return `null`, causing the system to silently fall back to a default model the user never configured.

Per-agent model overrides (`agents.{name}.model`) already bypass this validation and work correctly. This PR applies the same treatment to user-configured category models.

## Fix
1. Added `isUserConfiguredCategoryModel` flag to `resolveModelForDelegateTask()` input
2. When the flag is true, the category model is returned directly without fuzzy matching (same pattern as `userModel`)
3. `resolveCategoryConfig()` now returns `isUserConfiguredModel` based on whether the user explicitly set a model for the category
4. `resolveCategoryExecution()` passes this flag through to the model resolution call

## Changes
| File | Change |
|------|--------|
| `src/tools/delegate-task/model-selection.ts` | Add `isUserConfiguredCategoryModel` flag and early return before fuzzy matching |
| `src/tools/delegate-task/categories.ts` | Return `isUserConfiguredModel` from `resolveCategoryConfig()` |
| `src/tools/delegate-task/category-resolver.ts` | Pass the user-config flag through to `resolveModelForDelegateTask()` |
| `src/tools/delegate-task/model-selection.test.ts` | Add test for custom provider model bypass |

Fixes #2740

---
## Summary by cubic
Trust user-configured category models from custom providers and bypass fuzzy validation to prevent silent fallbacks to defaults. Fixes #2740.

- **Bug Fixes**
  - If a category model is explicitly set in `oh-my-opencode.json`, it’s used as-is (no fuzzy match).
  - Added `isUserConfiguredCategoryModel` flag and wired it through model resolution.
  - `resolveCategoryConfig()` marks user-set models; `category-resolver` passes the flag; tests added for the bypass.

<sup>Written for commit c41d6fd912f89657392b96fd01d5c8ecdb7b6062. Summary will update on new commits.</sup>


MoerAI:

PR #2793, opened 2026-03-24
fix(thinking-block-validator): reuse signed thinking blocks instead of synthetic unsigned placeholders

### LINKED ISSUE #2775: thinking-block-validator creates synthetic thinking blocks without signatures, causing infinite API rejection loop
## Bug

The `thinking-block-validator` hook creates synthetic thinking blocks **without a `signature` field**. Anthropic's API now requires a valid cryptographic signature on every thinking block in conversation history. Since signatures are server-generated and cannot be fabricated client-side, these synthetic blocks are always rejected with:

```
messages.5.content.0: Invalid `signature` in `thinking` block
```

### Reproduction

1. Start a session with an extended-thinking model (e.g. `claude-opus-4-6`)
2. Let the model respond with tool calls (thinking block + tool_use)
3. Continue the conversation — on the next turn, the validator may inject a synthetic thinking block
4. API rejects with `Invalid 'signature' in 'thinking' block`
5. `detect-error-type.ts` does not recognize this error → `isRecoverableError` returns `false`
6. User types "continue" → same synthetic block is injected again → infinite loop

### Root cause

`src/hooks/thinking-block-validator/hook.ts` lines 126-133:

```typescript
const thinkingPart = {
  type: "thinking" as const,
  id: `prt_0000000000_synthetic_thinking`,
  thinking: thinkingContent,
  synthetic: true,
  // ← NO signature field
}
```

The same issue exists in `src/hooks/session-recovery/storage/thinking-prepend.ts` (lines 54-61 and 109-116), which also creates thinking blocks without signatures.

### Impact

- Every session with extended thinking hits this on subsequent turns
- Creates an infinite error loop — the user cannot recover without starting a new session
- Workaround: add `"thinking-block-validator"` to `disabled_hooks` in config

### Suggested fix

Instead of creating synthetic thinking blocks, **reuse the original thinking Part verbatim** (which includes the valid `signature` from the API response). If no real thinking part with a valid signature is available, skip injection entirely — a downstream `thinking_block_order` error is preferable to a guaranteed signature rejection.

See PR #2653 for an implementation.

code-yeongyu:
## Summary
- Reuse existing signed Anthropic `thinking` and `redacted_thinking` blocks in `thinking-block-validator` instead of fabricating unsigned synthetic placeholders.
- Update session-recovery thinking prepend to reuse signed blocks only when they still preserve the required first-part ordering, otherwise skip recovery safely.
- Add regression coverage for proactive copy-vs-skip behavior and reactive ordering-safe reuse paths.

## Verification
- `bun test src/hooks/thinking-block-validator/hook.test.ts src/hooks/session-recovery/index.test.ts`
- `bun run typecheck`

## Notes
- `bun test` still reports unrelated pre-existing failures in `preemptive-compaction.degradation-monitor.test.ts`, `tmux-utils.test.ts`, and `non-interactive-env/index.test.ts`.
- Branch pushed: `fix/2775-thinking-block-signatures`
- Fixes #2775

---
## Summary by cubic
Reuse previously signed Anthropic `thinking`/`redacted_thinking` blocks instead of creating unsigned synthetic placeholders in the validator and session recovery. Adds ordering checks so we only reuse a signed block when it sorts before target parts; otherwise we skip. Fixes #2775.

- **Bug Fixes**
  - `thinking-block-validator`: If an assistant message lacks a leading thinking block and the model supports extended thinking, prepend the most recent signed `thinking`/`redacted_thinking` part verbatim; skip if none is found. No more unsigned synthetic placeholders.
  - `session-recovery`: Reuse the previous signed thinking part (file-backed = write verbatim; SDK-backed = patch by ID) only when it will sort before the target message’s parts; otherwise skip to avoid ordering loops.
  - Tests: Added coverage for reuse-vs-skip behavior and ordering-safe paths in both the validator and recovery flows.

<sup>Written for commit 0e7bd595f8751a1cdb4208de99c917340ebaca3f. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/session-recovery/index.test.ts
+++ b/src/hooks/session-recovery/index.test.ts
@@ -1,5 +1,10 @@
-import { describe, expect, it } from "bun:test"
+import { existsSync, readFileSync, rmSync } from "node:fs"
+import { join } from "node:path"
 import { detectErrorType } from "./index"
+import { prependThinkingPart, prependThinkingPartAsync } from "./storage/thinking-prepend"
+import { PART_STORAGE } from "../../shared/opencode-storage-paths"
+
+const { describe, expect, it, mock } = require("bun:test")
 
 describe("detectErrorType", () => {
   describe("thinking_block_order errors", () => {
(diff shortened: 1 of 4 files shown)


code-yeongyu:
Closing — will redo through Sisyphus properly.

PR #2795, opened 2026-03-24
fix(background-agent): prevent circuit breaker false positives on flat-format events

### LINKED ISSUE #2652: [Bug][Agent-generated] Background circuit breaker falsely cancels explore/librarian during legitimate search bursts

guazi04:
## Summary

- Fix circuit breaker incorrectly cancelling legitimate background tasks (Oracle, Explore, Librarian) when events arrive in flat format without `state.input`
- Unknown tool input is now treated as non-comparable instead of identical, while absolute `maxToolCalls` cap still provides backstop

## Changes

**`src/features/background-agent/loop-detector.ts`**
- Early return in `recordToolCall()` when `toolInput` is `undefined`/`null`: resets `consecutiveCount` to 1 each time (treats unknown input as "potentially different")
- Known identical input still triggers normally; absolute `maxToolCalls` cap unaffected

**`src/features/background-agent/loop-detector.test.ts`**
- Updated consecutive-trigger tests to use known input (matching real loop behavior)
- Added tests proving `undefined` and `null` input do NOT trigger false positives

**`src/features/background-agent/manager-circuit-breaker.test.ts`**
- Updated flat-format event test: 20 consecutive `read` events without `state.input` → task keeps running
- Strengthened absolute cap test with repeated flat-format events hitting `maxToolCalls`

### Root cause

Two event format paths exist in `resolveMessagePartInfo()`:

| Path | Condition | `state.input` | Signature |
|------|-----------|---------------|-----------|
| A (nested) | `properties.part` exists | ✅ present | `read::{"filePath":"/a.ts"}` — differentiates |
| B (flat) | no `properties.part` | ❌ undefined | `read` — bare tool name, all calls "identical" |

When events arrive via Path B, `createToolCallSignature()` produces bare tool names. 20 calls to `read` with 20 different files all produce the same signature `"read"`, triggering the consecutive-use detector.

## Testing

```bash
bun test src/features/background-agent/loop-detector.test.ts        # 18 pass
bun test src/features/background-agent/manager-circuit-breaker.test.ts  # 8 pass
bun run typecheck  # clean
```

## Related Issues

Fixes #2652

---
## Summary by cubic
Prevents the background-agent circuit breaker from cancelling valid tasks when tool events come in flat format without `state.input`. Unknown tool input is now treated as non-comparable; repetition detection only triggers on known-identical inputs. Fixes #2652.

- **Bug Fixes**
  - Updated `recordToolCall()` to treat `undefined`/`null` input as non-repeating (no consecutive accumulation; placeholder signature used).
  - Adjusted loop-detector tests: known inputs trigger at threshold; `undefined`/`null` inputs do not trigger.
  - Updated manager tests: 20 flat-format `read` events without input no longer cancel the task; `maxToolCalls` backstop still cancels when reached.

<sup>Written for commit b9fa2a3ebc8a281ef1e6aa3e428e36f5286a69af. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/background-agent/loop-detector.test.ts
+++ b/src/features/background-agent/loop-detector.test.ts
@@ -1,3 +1,5 @@
+/// <reference types="bun-types" />
+
 import { describe, expect, test } from "bun:test"
 import {
   createToolCallSignature,
@@ -19,7 +21,7 @@ function buildWindow(
 }
 
 function buildWindowWithInputs(
-  calls: Array<{ tool: string; input?: Record<string, unknown> }>,
+  calls: Array<{ tool: string; input?: Record<string, unknown> | null }>,
   override?: Parameters<typeof resolveCircuitBreakerSettings>[0]
 ) {
   const settings = resolveCircuitBreakerSettings(override)
@@ -148,7 +150,12 @@ describe("loop-detector", () => {
 
     describe("#given the same tool is called consecutively", () => {
       test("#when evaluated #then it triggers", () => {
-        const window = buildWindow(Array.from({ length: 20 }, () => "read"))
+        const window = buildWindowWithInputs(
+          Array.from({ length: 20 }, () => ({
+            tool: "read",
+            input: { filePath: "/src/same.ts" },
+          }))
+        )
 
         const result = detectRepetitiveToolUse(window)
 
@@ -176,7 +183,12 @@ describe("loop-detector", () => {
 
     describe("#given threshold boundary", () => {
       test("#when below threshold #then it does not trigger", () => {
-        const belowThresholdWindow = buildWindow(Array.from({ length: 19 }, () => "read"))
+        const belowThresholdWindow = buildWindowWithInputs(
+          Array.from({ length: 19 }, () => ({
+            tool: "read",
+            input: { filePath: "/src/same.ts" },
+          }))
+        )
 
         const result = detectRepetitiveToolUse(belowThresholdWindow)
 
@@ -184,7 +196,12 @@ describe("loop-detector", () => {
       })
 
       test("#when equal to threshold #then it triggers", () => {
-        const atThresholdWindow = buildWindow(Array.from({ length: 20 }, () => "read"))
+        const atThresholdWindow = buildWindowWithInputs(
+          Array.from({ length: 20 }, () => ({
+            tool: "read",
+            input: { filePath: "/src/same.ts" },
+          }))
+        )
 
         const result = detectRepetitiveToolUse(atThresholdWindow)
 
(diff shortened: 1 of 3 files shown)


guazi04:
Closing — the premise was wrong. #2655 was already merged and synced into our fork. The actual issue is that #2655's fix + the subsequent refactor (d48ea025) left a gap for flat-format events with undefined input. Resubmitting with corrected framing.

PR #2799, opened 2026-03-24
fix(category-resolver): respect is_unstable_agent config override (fixes #2061)

### LINKED ISSUE #2061: [Bug]: Gemini models forced into broken executeUnstableAgentTask path — tasks fail with 30s timeout
## Bug Description

Any category configured with a Gemini model (e.g., `google/gemini-3-pro`) is unconditionally routed to `executeUnstableAgentTask()` instead of the reliable `executeSyncTask()` path. The unstable path uses a fundamentally different (and broken) polling approach that causes tasks to fail with a 30-second timeout and zero output.

The root cause is an unconditional OR in `category-resolver.ts` that cannot be overridden by setting `is_unstable_agent: false` in the category config:

```typescript
// src/tools/delegate-task/category-resolver.ts
const isUnstableAgent = resolved.config.is_unstable_agent === true || 
  [unstableModel, categoryConfigModel].some(m => 
    m ? m.includes("gemini") || m.includes("minimax") || m.includes("kimi") : false
  )
```

Because of the `||`, even if a user explicitly sets `is_unstable_agent: false`, the gemini/minimax/kimi string check still forces `isUnstableAgent = true`.

## Steps to Reproduce

1. Configure a category to use a Gemini model in `oh-my-opencode.json`:
```json
{
  "categories": {
    "visual-engineering": {
      "model": "google/gemini-3-pro",
      "variant": "high"
    }
  }
}
```

2. Restart OpenCode (config is loaded at session start)

3. Run a simple task via that category:
```
task(category="visual-engineering", prompt="Write ':root { --test-color: #ff6b6b; }' to /tmp/test.css")
```

4. Task fails with 30s timeout.

## Expected Behavior

The task should complete successfully via `executeSyncTask()`, the same path used for non-gemini models. When the same category is configured with `anthropic/claude-sonnet-4-6`, the identical task completes in ~6 seconds.

Alternatively, setting `is_unstable_agent: false` in the category config should override the model-name-based detection and force the stable path.

## Actual Behavior

The task is routed to `executeUnstableAgentTask()` which fails:
- 30-second timeout
- Zero output (file is never created)
- Error thrown from `executeUnstableAgentTask`

## Error Logs

```
Error: Task failed to start within timeout (30s). Task ID: bg_65901912, Status: error
    at executeUnstableAgentTask (/Users/user/.cache/opencode/node_modules/oh-my-opencode/dist/index.js:55357:43)
    at processTicksAndRejections (native:7:39)
```

## A/B Test Results

Same task, same prompt, same machine, same session config — only the model differs:

| Test | Model | Execution Path | Result | Duration |
|------|-------|---------------|--------|----------|
| Baseline | `anthropic/claude-sonnet-4-6` | `executeSyncTask` | ✅ Success, file created | 6s |
| Gemini | `google/gemini-3-pro` | `executeUnstableAgentTask` | ❌ Timeout, zero output | 30s (failed) |

## Root Cause Analysis

There are two separate issues:

### Issue 1: Unconditional model-name routing (category-resolver.ts)

The `isUnstableAgent` check uses `||` (OR), meaning the model name check **always wins** regardless of explicit config:

```typescript
const isUnstableAgent = resolved.config.is_unstable_agent === true || 
  [unstableModel, categoryConfigModel].some(m => 
    m ? m.includes("gemini") || m.includes("minimax") || m.includes("kimi") : false
  )
```

Users cannot opt out. Setting `is_unstable_agent: false` has no effect because `false || true === true`.

**Suggested fix** — use nullish coalescing to let explicit config override:
```typescript
const isUnstableAgent = resolved.config.is_unstable_agent ?? 
  [unstableModel, categoryConfigModel].some(m => 
    m ? m.includes("gemini") || m.includes("minimax") || m.includes("kimi") : false
  )
```

This way: `is_unstable_agent: true` forces unstable, `is_unstable_agent: false` forces stable, and `undefined` (no config) falls back to model-name detection.

### Issue 2: executeUnstableAgentTask uses global session polling (unstable-agent-task.ts)

The unstable path polls using `client.session.status()` which returns status for **ALL** sessions globally:

```typescript
// unstable-agent-task.ts — global status call
const statusResult = await client.session.status()
const allStatuses = normalizeSDKResponse(statusResult, {} as Record<string, { type: string }>)
const sessionStatus = allStatuses[sessionID]
```

Compare this to the stable `executeSyncTask` path which uses proper session-scoped APIs via dependency injection:

```typescript
// sync-task.ts — session-scoped, reliable
const createSessionResult = await deps.createSyncSession(client, { ... })
const promptError = await deps.sendSyncPrompt(client, { sessionID, ... })
const pollError = await deps.pollSyncSession(ctx, client, { sessionID, ... })
const result = await deps.fetchSyncResult(client, sessionID)
```

The sync path creates a dedicated session, sends a prompt to it, polls that specific session, and fetches results from it — all session-scoped. The unstable path polls global status and hopes to find its session in the results, which is fundamentally less reliable and can confuse sessions when multiple tasks run concurrently.

## Configuration

```json
{
  "categories": {
    "visual-engineering": {
      "model": "google/gemini-3-pro",
      "variant": "high"
    }
  }
}
```

oh-my-opencode version: latest (installed via `bunx oh-my-opencode install` on Feb 22, 2026)

## Additional Context

- **Related closed issues**: #1079 (infinite loop in pollRunningTasks after stability detection), #445 (random effective freezing), PR #1040 (empty response retry for Gemini — patches symptoms but not root cause)
- **Affected categories by default**: `visual-engineering` (gemini-3-pro), `artistry` (gemini-3-pro), `writing` (gemini-3-flash/kimi). Any category using gemini, minimax, or kimi models.
- **Workaround**: Override affected categories to non-gemini models (e.g., `anthropic/claude-sonnet-4-6`) in `oh-my-opencode.json`. This routes them through the reliable `executeSyncTask` path.
- The `kimi` string was recently added to the unstable check, meaning the `writing` category (which defaults to kimi k2.5 since Feb 15) is now also affected.

MoerAI:
## Summary
- Allow `is_unstable_agent: false` to override automatic model-name-based unstable detection for Gemini/Minimax/Kimi categories

## Problem
The `isUnstableAgent` flag in `category-resolver.ts` uses a boolean OR (`||`) that unconditionally overrides explicit user configuration. When a user sets `is_unstable_agent: false` in their category config, the model name check (`includes("gemini")`) still forces `isUnstableAgent = true`. This routes all Gemini/Minimax/Kimi categories to the broken `executeUnstableAgentTask()` path, causing tasks to fail with 30-second timeouts.

Because `false || true === true`, the config option is completely non-functional.

## Fix
Replaced `||` with nullish coalescing (`??`) so that explicit user configuration takes priority:

- `is_unstable_agent: true` forces unstable path
- `is_unstable_agent: false` forces stable path
- `undefined` (no config) auto-detects from model name (preserves current default behavior)

## Changes
| File | Change |
|------|--------|
| `src/tools/delegate-task/category-resolver.ts` | Changed `=== true ||` to `??` for `isUnstableAgent` computation |

Fixes #2061

---
## Summary by cubic
Respect `is_unstable_agent` in the category resolver so explicit `false` is honored, preventing Gemini/Minimax/Kimi categories from being forced into the unstable path and timing out. Fixes #2061.

- **Bug Fixes**
  - Use nullish coalescing (`??`) instead of `||` to compute `isUnstableAgent`, so `true/false` config is respected.
  - Keep default behavior: when unset, auto-detect from model name (`gemini`, `minimax`, `kimi`).

<sup>Written for commit 41a43c62fce66b6ee90679ea094332e3bfd1c73c. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/delegate-task/category-resolver.ts
+++ b/src/tools/delegate-task/category-resolver.ts
@@ -188,7 +188,7 @@ Available categories: ${categoryNames.join(", ")}`,
   }
 
   const resolvedModel = actualModel?.toLowerCase()
-  const isUnstableAgent = resolved.config.is_unstable_agent === true || (resolvedModel ? resolvedModel.includes("gemini") || resolvedModel.includes("minimax") || resolvedModel.includes("kimi") : false)
+  const isUnstableAgent = resolved.config.is_unstable_agent ?? (resolvedModel ? resolvedModel.includes("gemini") || resolvedModel.includes("minimax") || resolvedModel.includes("kimi") : false)
 
   const defaultProviderID = categoryModel?.providerID
     ?? parseModelString(actualModel ?? "")?.providerID


MoerAI:

PR #2801, opened 2026-03-24
fix(tool-execute-before): strip null bytes from bash commands to prevent crash (fixes #2220)

### LINKED ISSUE #2220: [Bug] Null bytes in tool arguments after context compaction/model fallback crash bash tool
## Problem

After context compaction or model fallback, the LLM sometimes generates tool call arguments containing null bytes (`\x00`). Node.js `child_process` rejects these with:

```
The shell argument must be a string without null bytes. Received "\0"
```

This error occurs during bash tool execution when the command string contains embedded null bytes.

## Root Cause

When OmO's context management features trigger (aggressive truncation, preemptive compaction, deduplication recovery), the conversation history gets manipulated. After this manipulation, the model sometimes produces corrupted tool call arguments containing null bytes.

This appears to happen most frequently:
1. **After preemptive compaction** -- the summarized context may cause the model to hallucinate garbled bytes in subsequent tool calls
2. **During model fallback** -- when switching models mid-conversation, the new model may misinterpret the context
3. **After aggressive truncation** -- tool outputs replaced with `[TOOL RESULT TRUNCATED]` may confuse the model

## Suggested Fix

Add null byte sanitization in the pre-tool-use hook chain, specifically for the bash tool:

```typescript
// In pre-tool-use hook for bash tool
if (input.tool === 'bash' && typeof input.input?.command === 'string') {
  if (input.input.command.includes('\x00')) {
    // Strip null bytes and log warning
    input.input.command = input.input.command.replace(/\x00/g, '');
    log('[sanitize] Stripped null bytes from bash command', { sessionID: input.sessionID });
  }
}
```

Alternatively, this could be a general tool argument sanitizer that strips null bytes from all string arguments across all tools.

## Environment

- OmO v3.9.x with experimental features enabled:
  - `aggressive_truncation: true`
  - `preemptive_compaction: true`
  - `dynamic_context_pruning: enabled` with deduplication + supersede_writes + purge_errors
- Models: Claude Opus 4.6 with fallback chain to Kimi K2.5, GLM-5, Claude Sonnet 4.6
- Session with 7000+ messages over 2 days (very long session)

## Related

- Issue #1346 (Proactive handoff + session rotation to prevent quality degradation from auto-compaction)

MoerAI:
## Summary
- Sanitize null bytes from bash tool command arguments before execution to prevent shell process crashes

## Problem
When context compaction or model fallback occurs, the LLM can hallucinate null bytes (\x00) in tool call arguments. Node.js child_process rejects these with: `The shell argument must be a string without null bytes. Received "\0"`. This crashes the bash tool and leaves the session in an error state.

This affects long-running sessions with context management features enabled (aggressive_truncation, preemptive_compaction, dynamic_context_pruning) and model fallback chains.

## Fix
Added null byte sanitization at the very beginning of the `tool.execute.before` handler chain in `tool-execute-before.ts`. When the tool is `bash` and the command string contains null bytes, they are stripped before any other hooks process the command. A log entry is emitted for debugging.

This runs before all other hooks to ensure clean input for downstream processing.

## Changes
| File | Change |
|------|--------|
| `src/plugin/tool-execute-before.ts` | Added null byte detection and removal for bash commands at the top of the handler chain |

Fixes #2220

---
## Summary by cubic
Removes null bytes from `bash` tool commands before execution to stop Node `child_process` from rejecting the shell args and crashing. This prevents session failures when models emit `\x00` during context compaction or fallback. Fixes #2220.

- **Bug Fixes**
  - Sanitizes `bash` command strings in the earliest `tool.execute.before` handler by stripping `\x00`, and logs the event for debugging.

<sup>Written for commit c56a01c15de7d5d1541083e619281f0da7a8d42c. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin/tool-execute-before.ts
+++ b/src/plugin/tool-execute-before.ts
@@ -41,6 +41,16 @@ export function createToolExecuteBeforeHandler(args: {
   }
 
   return async (input, output): Promise<void> => {
+    if (input.tool.toLowerCase() === "bash" && typeof output.args.command === "string") {
+      if (output.args.command.includes("\x00")) {
+        output.args.command = output.args.command.replace(/\x00/g, "")
+        log("[tool-execute-before] Stripped null bytes from bash command", {
+          sessionID: input.sessionID,
+          callID: input.callID,
+        })
+      }
+    }
+
     await hooks.writeExistingFileGuard?.["tool.execute.before"]?.(input, output)
     await hooks.questionLabelTruncator?.["tool.execute.before"]?.(input, output)
     await hooks.claudeCodeHooks?.["tool.execute.before"]?.(input, output)


MoerAI:

PR #2859, opened 2026-03-26
docs(config): document object-style fallback_models

RaviTharuma:
## Summary
- document mixed string/object `fallback_models` entries in the config reference
- list the supported per-model fields for object-style fallback entries
- clarify that fallback-only settings are only promoted when that fallback model is actually selected

## Testing
- git diff --check

---
## Summary by cubic
Documented object-style `fallback_models` in config, with mixed arrays of strings and objects and per-fallback settings. Added agent and `categories` examples, a field reference (`model`, `variant`, `reasoningEffort`, `temperature`/`top_p`/`maxTokens`, `thinking` shape), and clarifications: settings apply only when that fallback is selected, provider prefixes can be omitted when inheriting the current provider, and explicit `variant` overrides inline suffixes.

<sup>Written for commit a7ac2e7abad7adf02163f047b926ac12dd9caf44. Summary will update on new commits.</sup>


code-yeongyu (approved):
Excellent documentation. Five clear examples covering every fallback_models use case — string entries, object entries, cross-provider chains, category-level, and the full shape reference.

Note: our recent docs commit (a2c7fed9) also added a shorter fallback_models section to the same file. There will be merge conflicts, but your version is more comprehensive. I'll rebase our broader commit after this merges to incorporate your section as the canonical reference.

Thank you for flagging the docs gap and writing this — this is exactly the kind of contribution that makes the project better.

PR #2862, opened 2026-03-27
fix(recovery): detect empty text parts alongside tool calls in fixEmptyMessages (fixes #2830)

### LINKED ISSUE #2830: [Bug]: fixEmptyMessages recovery fails when assistant message has tool calls alongside empty text part
### Prerequisites

## Bug
The empty content recovery logic (`fixEmptyMessages` / `findEmptyMessagesFromSDK`) fails to detect and fix a common error case, causing sessions to become permanently broken.
## Error
{"error":{"message":"messages: text content blocks must be non-empty","type":"invalid_request_error"}}
## Root Cause
When an assistant message contains **both** tool call parts **and** an empty text part `{"type": "text", "text": ""}`, `messageHasContentFromSDK()` returns `true` because it finds tool-type parts:
```typescript
if (TOOL_TYPES.has(type)) return true  // sees tool-call → returns true immediately
```
This causes findEmptyMessagesFromSDK() to skip the message entirely. But the Anthropic API still rejects it because of the empty text block.

### Steps to Reproduce

1. Use a model/provider that sometimes returns tool-call-only responses
2. The streaming processor creates a text part with "" alongside tool calls
3. Message is stored in SQLite with the empty text part
4. Every subsequent request in the session fails
5. fixEmptyMessages runs but doesn't detect the problem → session stays broken

### Expected Behavior

The recovery logic should also scan for individual empty text parts within non-empty messages, not just completely empty messages.

### Actual Behavior

The `fixEmptyMessages` recovery runs but fails to fix the problem. The session remains permanently broken — every subsequent message in the same session returns the same error.
This happens because `messageHasContentFromSDK()` sees tool-call parts and returns `true` (considers the message "has content"), skipping the empty text part fix. But the API still rejects the message due to the empty text block `{"type": "text", "text": ""}` that exists alongside the tool calls.
Error from API:
{"error":{"message":"messages: text content blocks must be non-empty","type":"invalid_request_error","param":"","code":null}}
Recovery attempts (visible in logs): fixEmptyMessages runs up to 3 times but never fixes the issue because findEmptyMessagesFromSDK() never identifies the problematic message.

### Doctor Output

```shell
Unable to run `bunx oh-my-opencode doctor` - bun not installed as standalone (embedded in OpenCode binary).
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

There are related PRs in the OpenCode core repo addressing this at the serialization level (#17742, #17396), but the plugin's own recovery logic has this independent detection gap.

### Operating System

Linux

### OpenCode Version

1.3.2

MoerAI:
## Summary
- Fix empty message recovery to detect empty text parts in messages that also contain tool call parts

## Problem
When an assistant message contains both tool call parts and an empty text part `{"type": "text", "text": ""}`, `messageHasContentFromSDK()` returns `true` because it finds tool-type parts. This causes `findEmptyMessagesFromSDK()` to skip the message entirely. But the Anthropic API still rejects it because of the empty text block, causing sessions to become permanently broken.

## Fix
Added a second scan pass in `fixEmptyMessagesWithSDK` using the existing `findMessagesWithEmptyTextPartsFromSDK()` function (already in `session-recovery/storage/empty-text.ts`). This catches messages that have empty text parts alongside non-empty content (tool calls). The existing `replaceEmptyTextPartsAsync` correctly handles replacing only the empty text parts within such messages.

## Changes
| File | Change |
|------|--------|
| `src/hooks/anthropic-context-window-limit-recovery/empty-content-recovery-sdk.ts` | Import `findMessagesWithEmptyTextPartsFromSDK`, add second scan pass after `findEmptyMessagesFromSDK`, merge results with deduplication |

Fixes #2830

---
## Summary by cubic
Fixes recovery for assistant messages that have empty text blocks alongside tool calls, preventing Anthropic API rejections and stuck sessions. Adds a second scan to detect and replace only the empty text parts. Fixes #2830.

- **Bug Fixes**
  - Also scan with findMessagesWithEmptyTextPartsFromSDK, in addition to findEmptyMessagesFromSDK.
  - Merge and dedupe target message IDs; process all with replaceEmptyTextPartsAsync.
  - Update scannedEmptyCount to reflect all processed messages.

<sup>Written for commit 404390efda7ce2d5f6012d8537270ecb7f24e002. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/anthropic-context-window-limit-recovery/empty-content-recovery-sdk.ts
+++ b/src/hooks/anthropic-context-window-limit-recovery/empty-content-recovery-sdk.ts
@@ -1,4 +1,4 @@
-import { replaceEmptyTextPartsAsync } from "../session-recovery/storage/empty-text"
+import { replaceEmptyTextPartsAsync, findMessagesWithEmptyTextPartsFromSDK } from "../session-recovery/storage/empty-text"
 import { injectTextPartAsync } from "../session-recovery/storage/text-part-injector"
 import type { Client } from "./client"
 
@@ -157,11 +157,19 @@ export async function fixEmptyMessagesWithSDK(params: {
   }
 
   const emptyMessageIds = await findEmptyMessagesFromSDK(params.client, params.sessionID)
-  if (emptyMessageIds.length === 0) {
+
+  // Also find messages with empty text parts alongside non-empty content (e.g., tool calls).
+  // messageHasContentFromSDK returns true for these since they have tool parts,
+  // but the API still rejects the empty text block.
+  const emptyTextPartIds = await findMessagesWithEmptyTextPartsFromSDK(params.client, params.sessionID)
+  const additionalIds = emptyTextPartIds.filter((id) => !emptyMessageIds.includes(id))
+  const allTargetIds = [...emptyMessageIds, ...additionalIds]
+
+  if (allTargetIds.length === 0) {
     return { fixed: false, fixedMessageIds: [], scannedEmptyCount: 0 }
   }
 
-  for (const messageID of emptyMessageIds) {
+  for (const messageID of allTargetIds) {
     const replaced = await replaceEmptyTextPartsAsync(
       params.client,
       params.sessionID,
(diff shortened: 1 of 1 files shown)


MoerAI:

PR #2864, opened 2026-03-27
fix(shared): check modelContextLimitsCache for all Anthropic models (fixes #2836)

### LINKED ISSUE #2836: bug: resolveActualContextLimit skips modelContextLimitsCache for Anthropic providers — hardcodes 200K
## Bug Description

`resolveActualContextLimit()` in `src/shared/context-limit-resolver.ts` **never checks `modelContextLimitsCache` for Anthropic providers**. It immediately returns from `getAnthropicActualLimit()`, which falls back to the hardcoded `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200000` unless the beta header or env var is set.

This means users who configure model context limits in `opencode.json` under an Anthropic provider have their config silently ignored.

## Current Code (dist/index.js:15450)

```javascript
function resolveActualContextLimit(providerID, modelID, modelCacheState) {
  if (isAnthropicProvider(providerID)) {
    return getAnthropicActualLimit(modelCacheState); // ← skips cache entirely
  }
  return modelCacheState?.modelContextLimitsCache?.get(`${providerID}/${modelID}`) ?? null;
}
```

## Impact

- Claude Opus 4.6 has a **1M token context window** (GA since March 13, 2026), but oh-my-opencode reports 200K
- Users behind a proxy (e.g. opencode-claude-max-proxy) never get `anthropicContext1MEnabled = true` because the proxy doesn't return the `context-1m` beta header
- The `ANTHROPIC_1M_CONTEXT=true` env var works as a workaround, but the proper config path via `opencode.json` `models.{model}.limit.context` is silently ignored for Anthropic

## Reproduction

1. Configure `opencode.json` with Anthropic provider and model limits:
   ```json
   "provider": {
     "anthropic": {
       "models": {
         "claude-opus-4-6": { "limit": { "context": 1000000 } }
       }
     }
   }
   ```
2. Start a session — system directive still says "200,000-token context window"
3. The `modelContextLimitsCache` is populated correctly (line 96925), but `resolveActualContextLimit` never reads it for Anthropic

## Proposed Fix

Check `modelContextLimitsCache` first for ALL providers, then fall back to provider-specific logic:

```javascript
function resolveActualContextLimit(providerID, modelID, modelCacheState) {
  const cached = modelCacheState?.modelContextLimitsCache?.get(`${providerID}/${modelID}`);
  if (cached) return cached;

  if (isAnthropicProvider(providerID)) {
    return getAnthropicActualLimit(modelCacheState);
  }
  return null;
}
```

## Environment

- oh-my-opencode v3.13.1
- OpenCode v1.3.2
- Claude Opus 4.6 via opencode-claude-max-proxy
- Windows 11

Happy to submit a PR for this.

MoerAI:
## Summary
- Allow cached model context limits to be used for all Anthropic models, not just GA ones

## Problem
`resolveActualContextLimit()` only checks `modelContextLimitsCache` for Anthropic GA models (claude-opus-4-6, claude-sonnet-4-6) due to the `isAnthropicNoHeaderGaModel()` guard. All other Anthropic models (claude-sonnet-4-5, etc.) fall through to the hardcoded `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200000`, ignoring any cached context limits from models.dev or user configuration.

## Fix
Removed the `isAnthropicNoHeaderGaModel()` restriction from the cache lookup. Now all Anthropic models check `modelContextLimitsCache` before falling back to the 200K default. The 1M explicit check still takes priority when the beta header or env var is set.

## Changes
| File | Change |
|------|--------|
| `src/shared/context-limit-resolver.ts` | Remove `isAnthropicNoHeaderGaModel` guard from cache lookup (L42-43) |
| `src/shared/context-limit-resolver.test.ts` | Update test to expect cached limit (500K) instead of hardcoded default (200K) |

Fixes #2836

---
## Summary by cubic
Use cached context limits for all Anthropic models by checking `modelContextLimitsCache` regardless of GA status. Prevents incorrect fallback to 200K when a higher cached limit exists; explicit 1M mode still takes priority. Fixes #2836.

- **Bug Fixes**
  - In `resolveActualContextLimit`, removed the `isAnthropicNoHeaderGaModel()` guard so any Anthropic model uses the cached limit if present.
  - Updated tests to expect 500K from cache for non-GA models; fallback remains 200K when no cache and no 1M mode.

<sup>Written for commit 69ef9fc6401ebd549db287b0f1d69d9dce8a2044. Summary will update on new commits.</sup>

### DIFF
--- a/src/shared/context-limit-resolver.test.ts
+++ b/src/shared/context-limit-resolver.test.ts
@@ -45,7 +45,7 @@ describe("resolveActualContextLimit", () => {
     expect(actualLimit).toBe(1_000_000)
   })
 
-  it("returns default 200K for older Anthropic models even when cached limit is higher", () => {
+  it("returns cached limit for Anthropic models when cache entry exists", () => {
     // given
     delete process.env[ANTHROPIC_CONTEXT_ENV_KEY]
     delete process.env[VERTEX_CONTEXT_ENV_KEY]
@@ -58,8 +58,8 @@ describe("resolveActualContextLimit", () => {
       modelContextLimitsCache,
     })
 
-    // then
-    expect(actualLimit).toBe(200_000)
+    // then — cached limit should be respected for all Anthropic models, not just GA ones
+    expect(actualLimit).toBe(500_000)
   })
 
   it("returns default 200K for Anthropic models without cached limit and 1M mode disabled", () => {
--- a/src/shared/context-limit-resolver.ts
+++ b/src/shared/context-limit-resolver.ts
@@ -40,7 +40,7 @@ export function resolveActualContextLimit(
     if (explicit1M === 1_000_000) return explicit1M
 
     const cachedLimit = modelCacheState?.modelContextLimitsCache?.get(`${providerID}/${modelID}`)
-    if (cachedLimit && isAnthropicNoHeaderGaModel(modelID)) return cachedLimit
+    if (cachedLimit) return cachedLimit
 
     return DEFAULT_ANTHROPIC_ACTUAL_LIMIT
   }


MoerAI:

code-yeongyu:
Thanks for the PR! This was already fixed directly in commit 40a92138 which addressed #2836 alongside two other related bugs. The fix in that commit does the same thing — removes the `isAnthropicNoHeaderGaModel()` guard so cached limits are respected for all Anthropic models. Closing as superseded.

PR #2889, opened 2026-03-27
fix(config): apply git_master defaults when section is missing (fixes #2040)

### LINKED ISSUE #2040: [Bug]: git_master config (commit_footer/include_co_authored_by: false) ignored — Sisyphus attribution still added to commits
## Description

`git_master.commit_footer` and `git_master.include_co_authored_by` are both set to `false` in the global config (`~/.config/opencode/oh-my-opencode.json`), but the Sisyphus attribution is still being injected into commit messages.

## Config

```json
{
  "git_master": {
    "commit_footer": false,
    "include_co_authored_by": false
  }
}
```

No project-level `oh-my-opencode.json` exists — only the global config.

## Actual Commit Output

```
feat(shopify): add queue-based addProductToShopify and getProductAddJobStatus utilities

Fire-and-forget addProductToShopify queues a BullMQ job with dedup ID, checks for existing shopifyId first. getProductAddJobStatus retrieves job state, result, or failure reason.

Ultraworked with [Sisyphus](https://github.com/code-yeongyu/oh-my-opencode)

```

Both the footer and co-authored-by trailer appear despite being explicitly disabled.

## Suspected Cause

In `src/features/opencode-skill-loader/git-master-template-injection.ts`:

```ts
const commitFooter = config?.commit_footer ?? true
const includeCoAuthoredBy = config?.include_co_authored_by ?? true
```

The `??` fallback to `true` suggests the `config` object may not be reaching this function at all (passed as `undefined`), causing both values to default to `true` regardless of user settings.

## Environment

- oh-my-opencode: v3.8.x (latest)
- OS: macOS
- Config location: `~/.config/opencode/oh-my-opencode.json`

MoerAI:
## Summary
- Ensure `git_master` config defaults are always applied even when the section is entirely absent from user config

## Problem
When `git_master` config exists only in the global config file (no project-level config), `pluginConfig.git_master` is `undefined` after Zod parsing because the root schema marks it as `.optional()`. This causes `injectGitMasterConfig()` to use `?? true` fallbacks, ignoring user-specified `commit_footer: false` and `include_co_authored_by: false`.

The root cause: `git_master: GitMasterConfigSchema.optional()` means Zod does NOT apply field-level defaults (inside `GitMasterConfigSchema`) when the entire `git_master` key is missing from the config input.

## Fix
Changed `git_master` from `.optional()` to `.default({...})` in the root schema so Zod always produces a fully-populated `git_master` object with field-level defaults applied. When the user explicitly sets `commit_footer: false`, the value is correctly preserved through the `?? true` fallback in `injectGitMasterConfig`.

## Changes
| File | Change |
|------|--------|
| `src/config/schema/oh-my-opencode-config.ts` | Change `git_master` from `.optional()` to `.default({...})` |
| `src/config/schema.test.ts` | Add tests verifying defaults are applied and explicit `false` values are respected |

Fixes #2040

---
## Summary by cubic
Apply `git_master` defaults when the section is missing, and also when the user config file is absent, so settings like `commit_footer` and `include_co_authored_by` are honored. Fixes #2040.

- **Bug Fixes**
  - Set `git_master` in `OhMyOpenCodeConfigSchema` to `.default({ commit_footer: true, include_co_authored_by: true, git_env_prefix: "GIT_MASTER=1" })` and, when no user config is found, fall back to `OhMyOpenCodeConfigSchema.parse({})` to apply Zod defaults.
  - Added tests to confirm defaults are applied and explicit `false` values are preserved.

<sup>Written for commit 2b2b2808958f42a894dc439fa453bcdb6f5cfae2. Summary will update on new commits.</sup>

### DIFF
--- a/src/config/schema.test.ts
+++ b/src/config/schema.test.ts
@@ -969,6 +969,45 @@ describe("GitMasterConfigSchema", () => {
   })
 })
 
+describe("OhMyOpenCodeConfigSchema - git_master defaults (#2040)", () => {
+  test("git_master defaults are applied when section is missing from config", () => {
+    //#given
+    const config = {}
+
+    //#when
+    const result = OhMyOpenCodeConfigSchema.safeParse(config)
+
+    //#then
+    expect(result.success).toBe(true)
+    if (result.success) {
+      expect(result.data.git_master).toBeDefined()
+      expect(result.data.git_master.commit_footer).toBe(true)
+      expect(result.data.git_master.include_co_authored_by).toBe(true)
+      expect(result.data.git_master.git_env_prefix).toBe("GIT_MASTER=1")
+    }
+  })
+
+  test("git_master respects explicit false values", () => {
+    //#given
+    const config = {
+      git_master: {
+        commit_footer: false,
+        include_co_authored_by: false,
+      },
+    }
+
+    //#when
+    const result = OhMyOpenCodeConfigSchema.safeParse(config)
+
+    //#then
+    expect(result.success).toBe(true)
+    if (result.success) {
+      expect(result.data.git_master.commit_footer).toBe(false)
+      expect(result.data.git_master.include_co_authored_by).toBe(false)
+    }
+  })
+})
+
 describe("skills schema", () => {
   test("accepts skills.sources configuration", () => {
     //#given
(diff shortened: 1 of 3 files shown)


MoerAI:

MoerAI:

code-yeongyu (approved):
Clean fix. Using `GitMasterConfigSchema.default({...})` instead of `.optional()` correctly applies Zod defaults when the section is absent, and falling back to `OhMyOpenCodeConfigSchema.parse({})` when no user config exists ensures defaults propagate properly. Tests cover both missing-section and explicit-false cases. LGTM.

PR #2891, opened 2026-03-27
fix: use display name in runtime-fallback retry

### LINKED ISSUE #2882: runtime-fallback retry fails with 'Agent not found' due to agent name remapping
## Bug Summary

When `runtime_fallback` is enabled and a retryable error occurs on the main agent session, the retry mechanism fails because it uses the **original** agent name (e.g. `sisyphus`) instead of the **remapped** display name (e.g. `Sisyphus (Ultraworker)`).

This causes every retry attempt to immediately fail with `Agent not found`, making the runtime-fallback feature counterproductive — it turns transient errors into hard failures.

## Reproduction

1. Enable `runtime_fallback` in `oh-my-opencode.json`:
   ```json
   {
     "runtime_fallback": {
       "enabled": true,
       "retry_on_errors": [402, 429, 500, 502, 503, 504],
       "max_fallback_attempts": 3
     }
   }
   ```
2. Wait for any transient API error (500, timeout, abort) on the main Sisyphus session
3. Runtime-fallback catches the error and attempts retry
4. Retry fails with: `Agent not found: "sisyphus". Available agents: Sisyphus (Ultraworker), Atlas (Plan Executor), ...`

## Log Evidence

```
[2026-03-27T07:55:58.285Z] [runtime-fallback] message.updated with assistant error
  {"sessionID":"ses_...","errorName":"MessageAbortedError"}

[2026-03-27T07:56:58.484Z] [runtime-fallback] Aborted in-flight session request (session.status.retry-signal)
  {"sessionID":"ses_..."}

[2026-03-27T07:56:58.508Z] [auto-compact] session.error received
  {"sessionID":"ses_...","error":{"name":"UnknownError","data":{"message":"Agent not found: \"sisyphus\". Available agents: Sisyphus (Ultraworker), Atlas (Plan Executor), Hephaestus (Deep Agent), Metis (Plan Consultant), Momus (Plan Critic), Prometheus (Plan Builder), Sisyphus-Junior, explore, general, librarian, multimodal-looker, oracle, plan"}}}

[2026-03-27T07:56:58.535Z] [runtime-fallback] session.error received
  {"sessionID":"ses_...","resolvedAgent":"sisyphus","errorName":"MessageAbortedError"}
```

## Expected Behavior

The retry mechanism should use the remapped agent name (or internal agent ID) that matches what's registered in the agent registry, not the original config-level name.

## Actual Behavior

Retry uses the original agent name → `Agent not found` → retry fails → error surfaces to user as "The operation was aborted."

## Environment

- opencode: 1.3.3
- oh-my-openagent: latest (installed via plugin)
- OS: macOS (darwin)

## Workaround

Disable `runtime_fallback` (`"enabled": false`) to prevent the broken retry path from being triggered.

codivedev:
Fixes #2882

When runtime_fallback is enabled and a retryable error occurs, the retry mechanism was using the original agent name (e.g. 'sisyphus') instead of the remapped display name (e.g. 'Sisyphus (Ultraworker)').

This caused the error: "Agent not found: 'sisyphus'. Available agents: Sisyphus (Ultraworker), Atlas (Plan Executor)..."

Changes:
- Import getAgentDisplayName from ../../shared/agent-display-names
- Convert retryAgent to display name before passing to promptAsync
- Update test to expect display name instead of config key

The fix ensures that when retrying with a fallback model, the agent name is properly converted to its display name format before being passed to the API.

---
## Summary by cubic
Fixes #2882. Use the agent display name during runtime-fallback retries so the API can find the agent and the retry succeeds.

- **Bug Fixes**
  - Convert the retry agent to its display name with `getAgentDisplayName` before calling `promptAsync`.
  - Update the test to expect the display name instead of the config key.

<sup>Written for commit b4d4d30fa84fed20bbe413575913b7d1afd610d2. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/runtime-fallback/auto-retry.ts
+++ b/src/hooks/runtime-fallback/auto-retry.ts
@@ -9,6 +9,7 @@ import { SessionCategoryRegistry } from "../../shared/session-category-registry"
 import { buildRetryModelPayload } from "./retry-model-payload"
 import { getLastUserRetryParts } from "./last-user-retry-parts"
 import { extractSessionMessages } from "./session-messages"
+import { getAgentDisplayName } from "../../shared/agent-display-names"
 
 const SESSION_TTL_MS = 30 * 60 * 1000
 
@@ -126,13 +127,14 @@ export function createAutoRetryHelpers(deps: HookDeps) {
         })
 
         const retryAgent = resolvedAgent ?? getSessionAgent(sessionID)
+        const retryAgentDisplayName = retryAgent ? getAgentDisplayName(retryAgent) : undefined
         sessionAwaitingFallbackResult.add(sessionID)
         scheduleSessionFallbackTimeout(sessionID, retryAgent)
 
         await ctx.client.session.promptAsync({
           path: { id: sessionID },
           body: {
-            ...(retryAgent ? { agent: retryAgent } : {}),
+            ...(retryAgentDisplayName ? { agent: retryAgentDisplayName } : {}),
             ...retryModelPayload,
             parts: retryParts,
           },
--- a/src/hooks/runtime-fallback/index.test.ts
+++ b/src/hooks/runtime-fallback/index.test.ts
@@ -2404,7 +2404,7 @@ describe("runtime-fallback", () => {
 
       expect(promptCalls.length).toBe(1)
       const callBody = promptCalls[0]?.body as Record<string, unknown>
-      expect(callBody?.agent).toBe("prometheus")
+      expect(callBody?.agent).toBe("Prometheus (Plan Builder)")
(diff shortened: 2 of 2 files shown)


codivedev:

codivedev:
recheck

RaviTharuma:
@cubic-dev which path is better using the display name or the id?
which solution is the cleanest and the right one long term?

RaviTharuma:
@cubic-dev

So we should take the display name because OpenCode itself does not even use the ID.
Although the ID would be the cleanest solution. Is this what you are saying?

code-yeongyu (approved):
Correct fix. The `promptAsync` call needs display name format (e.g., "Sisyphus (Ultraworker)") not config key ("sisyphus"), same pattern as how OpenCode registers agents. Test update is also correct. LGTM.

PR #2895, opened 2026-03-27
fix(session-recovery): fallback to fetching messageID from session messages (fixes #2046)

### LINKED ISSUE #2046: Bug: session-recovery silently fails due to missing messageID in OpenCode's session.error event
## Summary

The `session-recovery` hook correctly detects `tool_result_missing` errors but **silently fails to recover** because OpenCode's `session.error` event does not include `messageID` in its payload.

### OpenCode's event schema (extracted from binary)

```javascript
BusEvent.define("session.error", zod_default.object({
  sessionID: zod_default.string().optional(),
  error: MessageV2.Assistant.shape.error
  // ❌ No messageID field
}))
```

### OMO's recovery path (dist/index.js ~line 75624)

```javascript
// In createEventHandler2 → session.error handler:
const messageInfo = {
  id: props?.messageID,   // ← Always undefined (OpenCode doesn't send it)
  role: "assistant",
  sessionID,
  error: error45
};
const recovered = await hooks2.sessionRecovery.handleSessionRecovery(messageInfo);
```

### handleSessionRecovery (dist/index.js ~line 21044)

```javascript
const assistantMsgID = info.id;   // undefined
if (!sessionID || !assistantMsgID)
  return false;                    // ← Silently returns false here
```

## Evidence

- **OMO log** (`/tmp/oh-my-opencode.log`): 26+ occurrences of `tool_result_missing` error caught by `auto-compact`, but **zero** entries from `session-recovery` module
- Error occurs when tool execution is aborted (ESC, timeout, "Tool execution aborted") and OpenCode fails to insert a `tool_result` placeholder
- The recovery code at `recoverToolResultMissing()` is never reached

## Reproduction

1. Start a session with Claude (Anthropic model)
2. Trigger multiple tool calls
3. Interrupt/abort mid-execution (ESC or timeout)
4. Send a new message → API returns 400: `tool_use ids were found without tool_result blocks`
5. Check `/tmp/oh-my-opencode.log` — no `session-recovery` entries despite the error being recoverable

## Proposed Fix

Since `messageID` is not available from OpenCode's event, the recovery should **fall back to finding the last assistant message** with orphaned `tool_use` blocks:

```javascript
// In handleSessionRecovery, after the existing messageID lookup fails:
if (!failedMsg && sessionID) {
  // Fallback: find last assistant message with tool_use parts
  const allMsgs = msgs?.filter(m => m.info?.role === "assistant") ?? [];
  failedMsg = allMsgs[allMsgs.length - 1]; // Last assistant message
}
```

This would allow recovery to work regardless of whether OpenCode provides `messageID`.

## Related

- OpenCode issue: anomalyco/opencode#2720 (root cause — tool abort doesn't generate `tool_result`)
- OMO issue: #381 (similar error from context compression — closed)

## Environment

- oh-my-opencode: 3.8.3
- OpenCode: binary build (Feb 20, 2026)
- Model: anthropic/claude-opus-4-6
- OS: WSL2 (Ubuntu)

MoerAI:
## Summary
- Fix session-recovery hook to work around OpenCode's missing messageID in session.error events

## Problem
OpenCode's `session.error` event does not include `messageID` in its payload:
```javascript
BusEvent.define("session.error", zod_default.object({
  sessionID: zod_default.string().optional(),
  error: MessageV2.Assistant.shape.error
  // No messageID field
}))
```

The session-recovery hook extracts `messageID` from `props?.messageID` (line 533 in event.ts), which is always `undefined`. The recovery function then returns `false` on `if (!assistantMsgID) return false`, silently failing to recover from tool_result_missing, thinking_block_order, and other recoverable errors.

## Fix
When `messageID` is not provided in the event payload, the hook now falls back to fetching the session messages and finding the last assistant message with an error. This provides the `assistantMsgID` needed for the recovery flow to proceed.

## Changes
| File | Change |
|------|--------|
| `src/hooks/session-recovery/hook.ts` | Add messageID fallback: fetch session messages and find last errored assistant message |

Fixes #2046

---
## Summary by cubic
Fix session recovery when OpenCode `session.error` events omit `messageID` by falling back to fetch session messages and use the last errored assistant message’s id. Restores recovery for `tool_result_missing`, `thinking_block_order`, and similar errors.

- **Bug Fixes**
  - If `info.id` is missing, fetch messages for the `sessionID` and derive `assistantMsgID` from the last assistant message with an error.
  - Return early only if `sessionID` is missing or no errored assistant message is found; log a single fallback failure.

<sup>Written for commit fee60d2defa3d8e1ce662df76fd25215d57c7fc6. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/session-recovery/hook.ts
+++ b/src/hooks/session-recovery/hook.ts
@@ -54,9 +54,25 @@ export function createSessionRecoveryHook(ctx: PluginInput, options?: SessionRec
     if (!errorType) return false
 
     const sessionID = info.sessionID
-    const assistantMsgID = info.id
+    let assistantMsgID = info.id
 
-    if (!sessionID || !assistantMsgID) return false
+    if (!sessionID) return false
+
+    if (!assistantMsgID) {
+      try {
+        const messagesResp = await ctx.client.session.messages({
+          path: { id: sessionID },
+          query: { directory: ctx.directory },
+        })
+        const msgs = (messagesResp as { data?: MessageData[] }).data
+        const lastAssistant = msgs?.findLast((m) => m.info?.role === "assistant" && m.info?.error)
+        assistantMsgID = lastAssistant?.info?.id
+      } catch {
+        log("[session-recovery] Failed to fetch messages for messageID fallback", { sessionID })
+      }
+    }
+
+    if (!assistantMsgID) return false
     if (processingErrors.has(assistantMsgID)) return false
     processingErrors.add(assistantMsgID)
 


MoerAI:

code-yeongyu (approved):
Solid fix for #2046. The root cause (OpenCode's session.error event not including messageID) is upstream and out of our control, so the fallback approach — fetching messages from the session API and finding the last assistant message with an error — is the right workaround. The try/catch with graceful fallthrough is clean. LGTM.

PR #2912, opened 2026-03-28
test(fallback): add CLIProxyAPI-only integration matrix

RaviTharuma:
## Summary
- add a new plugin-level CLIProxyAPI-only fallback matrix
- cover no fallback, model fallback, runtime fallback, and model+runtime combinations
- verify all three retry entry points: `session.error`, `session.status` retry, and assistant `message.updated` errors

## Why
The existing fallback coverage is broad, but it does not answer the exact interaction question for CLIProxyAPI-only deployments. This matrix locks down the intended split between model fallback and runtime fallback on the real plugin/event/chat-message integration path.

## Verified behavior
- model fallback uses `abort + prompt`
- runtime fallback uses `promptAsync` for `session.error` and `message.updated`
- runtime fallback aborts provider-side auto-retry before `promptAsync` on `session.status` retry
- when both are enabled, runtime fallback wins

## Testing
- bun test src/plugin/fallback.cliproxyapi-matrix.test.ts src/plugin/event.model-fallback.test.ts src/hooks/runtime-fallback/index.test.ts src/plugin/chat-message.test.ts

### DIFF
--- a/src/plugin/fallback.cliproxyapi-matrix.test.ts
+++ b/src/plugin/fallback.cliproxyapi-matrix.test.ts
@@ -1,10 +1,12 @@
 declare const require: (name: string) => any
 const { afterEach, describe, expect, mock, test } = require("bun:test")
 
+const PROVIDER_ID = "cliproxyapi"
+
 mock.module("../shared/connected-providers-cache", () => ({
-  readConnectedProvidersCache: () => ["quotio"],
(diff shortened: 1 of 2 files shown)


RaviTharuma:
CI is green, approvals are in, and this only adds test coverage without touching production code. This is ready for maintainer merge; I cannot merge upstream myself because the fork account does not have MergePullRequest permission.

Issue #2847, opened 2026-03-26
[Bug]: `task` tool schema allows `category` + `subagent_type` together — model confabulates invalid values


ryanleecode:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

## Goal

The `task` tool JSON schema enforces mutual exclusivity between `category` and
`subagent_type` so models cannot generate both fields simultaneously, and
valid `subagent_type` values are enumerated in the schema so models cannot
confabulate invalid ones.

## Evidence: Observed Model Behavior

Agent attempting to spawn the `oracle` subagent made three attempts before
succeeding:

**Attempt 1** — both fields present, invalid subagent_type:

```json
{ "category": "ultrabrain", "subagent_type": "Sisyphus-Junior" }
```

**Attempt 2** — both fields still present after correction:

```json
{ "category": "unspecified-high", "subagent_type": "Sisyphus-Junior" }
```

**Attempt 3** — correct, after reading tool description aloud to itself:

```json
{ "subagent_type": "oracle" }
```

The constraint is documented in prose (AGENTS.md: "MUST provide EXACTLY ONE of
`category` OR `subagent_type`") but the schema permits both. The model ignored
the prose twice, only self-correcting by quoting the tool description's enum
list back to itself.

## Acceptance Criteria

- [ ] `task` tool JSON schema uses `oneOf` (or equivalent) to make `category`
      and `subagent_type` mutually exclusive
- [ ] `subagent_type` is typed as an enum of valid values in the schema so
      models cannot confabulate names like `Sisyphus-Junior`
- [ ] Passing both fields produces a structured error naming the violation

### Steps to Reproduce

.

### Expected Behavior

.

### Actual Behavior

.

### Doctor Output

```shell
.
```

### Error Logs

```shell
.
```

### Configuration

```json

```

### Operating System

Linux

### OpenCode Version

1.3.2

PR #2934, opened 2026-03-29
fix(delegate-task): prevent infinite loop when both category and subagent_type provided

guazi04:
## Problem

When models pass both `category` and `subagent_type` to the `task()` tool, the current code throws an error. Models then retry with the same parameters, creating an **infinite loop** that wastes tokens and blocks execution.

### Root cause

1. **Tool description mentions "Sisyphus-Junior" twice** in the category context (lines 76, 82), causing models to associate `category` with `sisyphus-junior` and pass both parameters
2. **Description/behavior inconsistency**: description says "subagent_type is ignored" but code throws
3. **Throw creates unrecoverable loop**: models receive the error but keep generating the same function call

### Reproduction

Any model (especially GPT series) delegating a quick task:
```
task(category="quick", subagent_type="sisyphus-junior", ...)
// → Error: mutually exclusive
// → Model retries with same params → infinite loop
```

### 1. Tool description cleanup
- Removed 3 references to "Sisyphus-Junior" — internal implementation detail models don't need to know
- Fixed wording inconsistency about mutual exclusion behavior

### 2. Defensive behavior restored
- `throw new Error(...)` → `log() + args.subagent_type = undefined`
- When both provided: warn, discard `subagent_type`, continue with `category`
- This matches the pre-existing behavior before the strict validation was introduced

### 3. Test updated
- Replaced `rejects.toThrow("mutually exclusive")` with successful resolution + log assertion

## Verification
- `bun run typecheck`: ✅ pass
- `bun test src/tools/delegate-task/`: 234 pass, 0 fail

---
## Summary by cubic
Prevents infinite retry loops in the `task()` tool by ignoring `subagent_type` when `category` is also provided, logging a warning, and continuing with `category`. Cleans up the tool description to reduce confusion and updates tests.

- **Bug Fixes**
  - When both `category` and `subagent_type` are passed, log a warning and drop `subagent_type`; proceed with `category`.
  - Removed “Sisyphus-Junior” references and clarified mutual-exclusion wording in the tool help.
  - Updated test to expect successful execution and assert the warning log.

<sup>Written for commit ef0ed6d4380f96a7127b48a34e694d35fb05f37f. Summary will update on new commits.</sup>

### DIFF
--- a/src/tools/delegate-task/tools.test.ts
+++ b/src/tools/delegate-task/tools.test.ts
@@ -9,6 +9,7 @@ import { clearSkillCache } from "../../features/opencode-skill-loader/skill-cont
 import { __setTimingConfig, __resetTimingConfig } from "./timing"
 import * as connectedProvidersCache from "../../shared/connected-providers-cache"
 import * as executor from "./executor"
+import * as logger from "../../shared/logger"
 
 const SYSTEM_DEFAULT_MODEL = "anthropic/claude-sonnet-4-6"
 
@@ -465,7 +466,7 @@ describe("sisyphus-task", () => {
        expect(args.subagent_type).toBe("Sisyphus-Junior")
     }, { timeout: 10000 })
 
-    test("rejects when both category and subagent_type are provided", async () => {
+    test("logs and proceeds when both category and subagent_type are provided", async () => {
       //#given
       const { createDelegateTask } = require("./tools")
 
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Thanks for the PR! This addresses the same issue as #2939, #2947, and #2847. We're going with #2939 (sjawhar) which takes the same category-wins approach with a response note for the agent.

Closing as duplicate.

code-yeongyu:
Duplicate — going with #2939

Issue #2903, opened 2026-03-28
bug: createTimestampTransformer tests are timezone-dependent (2 test failures)


robinmordasiewicz:
## Description

Tests for \`createTimestampTransformer\` in \`src/cli/run/timestamp-output.ts\` hardcode expected UTC-formatted timestamps but the function formats in local time. Tests fail in any timezone other than UTC.

## Failing Tests

\`src/cli/run/timestamp-output.test.ts\`

| Test | Expected | Received | Explanation |
|------|----------|----------|-------------|
| prefixes each output line with timestamp | \`[12:34:56] hello\` | \`[02:34:56] hello\` | UTC-10 (HST) offset |
| keeps line-start state across chunk boundaries | \`[01:02:03] hello\` | \`[15:02:03] hello\` | UTC-10 offset |

## Reproduction

```bash
TZ=Pacific/Honolulu bun test src/cli/run/timestamp-output.test.ts
```

Works in UTC:
```bash
TZ=UTC bun test src/cli/run/timestamp-output.test.ts
```

## Fix Options

1. Set \`TZ=UTC\` in the test setup
2. Use a mocked \`Date\` that returns consistent local-time values
3. Extract the formatter and test the timestamp relative to the mocked date's local representation

## Environment

- Commit: \`44b039b\`
- Bun: 1.3.x
- Timezone: Pacific/Honolulu (UTC-10)

code-yeongyu:
I triaged this on current `dev` and it is still reproducible.

Repro I ran:
- `TZ=Pacific/Honolulu bun test src/cli/run/timestamp-output.test.ts`

Current result:
- `2 fail, 4 pass`
- The failures still match the issue: expected `12:34:56`/`01:02:03`, received `02:34:56`/`15:02:03`.

I also checked last-4-weeks merged PRs with `gh pr list --state merged --search 'timestamp-output timezone merged:>=2026-02-28' --limit 10` and found no matching merged PR.

Leaving this open and escalating.

@code-yeongyu

code-yeongyu:
Already fixed — tests now use `createLocalDate()` helper which creates dates in local timezone, matching `formatTimestamp()`'s use of `getHours()`. All 6 tests pass regardless of timezone. Closing as resolved.

PR #2935, opened 2026-03-29
fix: stabilize timestamp-output tests across timezones

HOYALIM:
## Summary
This PR fixes issue #2903 by making the `timestamp-output` CLI tests deterministic across timezones without changing runtime behavior.

## Problem
`createTimestampTransformer()` formats timestamps with local-time accessors (`getHours()`, `getMinutes()`, `getSeconds()`), but the original tests depended on timezone-sensitive fixture expectations. That made exact timestamp assertions brittle across environments.

## Fix
The tests now derive exact expected timestamps from fixture `Date` objects using the same local-time semantics as the implementation, instead of relying on timezone-sensitive fixture assumptions.

## Changes
- added a small test helper that formats expected timestamps from `Date` local getters
- updated `src/cli/run/timestamp-output.test.ts` to build exact expectations from fixture dates instead of hardcoded timezone-sensitive values
- left `src/cli/run/timestamp-output.ts` unchanged because runtime behavior was already correct

## Impacts
- runtime CLI behavior is unchanged
- `src/cli/run/timestamp-output.test.ts` now passes consistently across timezones while still asserting exact output
- verified with:
  - `~/.bun/bin/bun test src/cli/run/timestamp-output.test.ts`
  - `TZ=UTC ~/.bun/bin/bun test src/cli/run/timestamp-output.test.ts`
  - `TZ=Pacific/Honolulu ~/.bun/bin/bun test src/cli/run/timestamp-output.test.ts`
  - `~/.bun/bin/bun test src/cli/run`
  - `npx tsc --noEmit src/cli/run/timestamp-output.ts src/cli/run/timestamp-output.test.ts`

### DIFF
--- a/src/cli/run/timestamp-output.test.ts
+++ b/src/cli/run/timestamp-output.test.ts
@@ -3,6 +3,10 @@
 import { describe, expect, it } from "bun:test"
 import { createTimestampTransformer, createTimestampedStdoutController } from "./timestamp-output"
 
+function createLocalDate(hours: number, minutes: number, seconds: number): Date {
+  return new Date(2026, 1, 19, hours, minutes, seconds)
+}
+
 interface MockWriteStream {
   write: (
     chunk: Uint8Array | string,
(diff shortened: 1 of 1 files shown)


HOYALIM:

PR #1226, opened 2026-01-28
feat: Add native Zellij terminal multiplexer support

mrdavidlaing:
## Summary

Implements native Zellij terminal multiplexer support alongside tmux, providing users with more flexibility in their terminal setup. This adds a unified abstraction layer that supports both tmux and zellij, with automatic detection and capability-based behavior.

Closes #1153

## Key Features

✅ **Unified Multiplexer Interface**: Capability-based abstraction supporting both tmux and zellij  
✅ **Auto-detection**: Automatically detects which multiplexer is running via environment variables  
✅ **Configuration**: New `terminal.provider` option (`auto`, `tmux`, `zellij`)  
✅ **Pane Stacking**: Anchor pane system for tab-like behavior in both multiplexers  
✅ **Auto-close**: Panes close instantly when background tasks complete  
✅ **Backward Compatibility**: Existing tmux configurations continue to work  
✅ **Full Test Coverage**: 79 new tests ensuring both adapters work correctly  

### New Components

- **`src/shared/terminal-multiplexer/`** - New abstraction layer
  - `types.ts` - Multiplexer interface and capabilities
  - `tmux-adapter.ts` - Tmux implementation (wraps existing tmux-utils)
  - `zellij-adapter.ts` - Zellij implementation with label-based pane tracking
  - `detection.ts` - Auto-detection logic
  - `*.test.ts` - 79 comprehensive tests (contract, unit, integration)

### Modified Components

- **`src/features/tmux-subagent/manager.ts`** - Refactored to use Multiplexer abstraction
- **`src/hooks/interactive-bash-session/`** - Updated for multiplexer type tracking
- **`src/config/schema.ts`** - Extended with `terminal.provider` configuration
- **`src/index.ts`** - Integrated detection and multiplexer creation
- **Documentation** - Updated `src/shared/AGENTS.md` and `src/features/AGENTS.md`

### Architecture

```typescript
interface Multiplexer {
  type: "tmux" | "zellij"
  capabilities: {
    manualLayout: boolean      // tmux: true, zellij: false
    persistentLabels: boolean  // tmux: false, zellij: true
  }
  spawnPane(options: SpawnOptions): Promise<PaneHandle>
  closePane(handle: PaneHandle): Promise<void>
  getPanes(): Promise<PaneHandle[]>
  ensureSession(name: string): Promise<void>
  killSession(name: string): Promise<void>
}
```

### Detection Priority

1. `$TMUX` environment variable → "tmux"
[...]

PR #2938, opened 2026-03-29
feat(zellij): add Zellij terminal multiplexer support

### LINKED ISSUE #1153: Feature Request: Add native Zellij terminal multiplexer support
## Summary

Request to native support for [Zellij](https://zellij.dev/) as an alternative to tmux for terminal multiplexer integration.

## Current State

I'm really excited about the new Oh-my-opencode tmux integration:

1. **Tool Layer** (`interactive_bash`): Execute tmux subcommands for TUI apps (vim, htop, etc.)
2. **Hook Layer** (`interactive-bash-session`): Track and auto-cleanup tmux sessions
3. **Feature Layer** (`tmux-subagent`): Spawn visual panes for background agents

## Problem

However, I'm a Zellij user :( 

## Offer to Contribute

Would you accept a PR implementing this feature?

I've detailed in a comment below the approach I'd take.

Let me know if you're interested.

Thanks for an awesome plugin!

Omomomomo!

florianschepp:
## Summary

This PR adds Zellij as an alternative terminal multiplexer, building on the excellent groundwork from @mrdavidlaing in #1226. That PR was closed for procedural reasons (head/base both pointed to `dev`) — all credit for the original design and implementation goes to David.

**Improvements over #1226:**

- **`displayName` support**: New optional field in `SpawnOptions` — used as the `-n` flag in Zellij so panes show `"Agent: {title}"` instead of raw session IDs
- **Reliable stacking direction**: Non-anchor panes always use `-d down` regardless of what the caller passes, ensuring consistent vertical stacking
- **Dependency injection**: `ZellijAdapter` constructor accepts an optional `spawnFn` for clean unit testing without mocking `bun`
- **Modular config schema**: Extracted `src/config/schema/` directory with individual schema files including new `terminal.ts` (`ZellijConfigSchema`, `TerminalConfigSchema` with `provider: "auto" | "tmux" | "zellij"`)
- **8 BDD-style tests** covering stacking behavior and displayName naming

## What's included

- `src/shared/terminal-multiplexer/` — ZellijAdapter, types, detection, storage (David's files + improvements)
- `src/config/schema/terminal.ts` — new Zellij/terminal config schemas
- `src/create-managers.ts` — async factory wiring zellij detection
- `src/features/tmux-subagent/manager.ts` — overloaded constructor supporting both tmux and zellij adapters
- `src/shared/terminal-multiplexer/zellij-adapter-stacking.test.ts` — 8 passing tests

## Attribution

Co-authored-by: David Laing [mrdavidlaing](https://github.com/mrdavidlaing) <david@davidlaing.com>

Original PR: #1226

Closes #1153


florianschepp:
Marking as draft for now. Noticed the PR was showing far more changed files than expected due to some unrelated divergence from the source fork sneaking in during the merge setup. The Zellij implementation itself is solid; this is purely a housekeeping issue on my end. Will mark ready for review once the diff is clean.

florianschepp:
Diff is clean now. Was a quick one :). 20 files now, all Zellij-related. Marking ready for review.

florianschepp:

florianschepp:
recheck

florianschepp:
@cubic-dev-ai Please re-review the latest commit.

florianschepp:
@cubic-dev-ai Please re-review the latest commit.

florianschepp:
@cubic-dev-ai Please re-review the latest commit.

florianschepp:
@cubic-dev-ai All 5 round-4 issues addressed. Also merged upstream origin/dev and resolved 3 conflict regions in manager.ts. Please re-review.

florianschepp:
@cubic-dev-ai All 9 fifth-round review violations have been addressed in commit 0db6846. Please re-review when ready.

florianschepp:
@cubic-dev-ai Both sixth-round review violations addressed in commit f1059e8. Please re-review.

florianschepp:
@cubic-dev-ai All four eighth-round review violations addressed in commits 490ee26 and e0aa21b:

1. $ZELLIJ_PANE_ID escape removed so bash expands the env var (zellij-adapter.ts:129)
2. closePane now uses adapter resolveTmuxPath() + this.spawn instead of standalone closeTmuxPane() (tmux-adapter.ts)
3. tmux-adapter.test.ts uses makeMockSpawn() with mock tmux path to prevent real execution
4. zellij-adapter.test.ts cache test checks labelToSpawned.has() directly instead of getPanes()

Please re-review.

florianschepp:
@cubic-dev-ai Please perform a full re-review of the entire PR. All 6 previously reported issues have been addressed in commit 5e84261:

1. **tmux-adapter.test.ts label-mapping conditional** - Removed `if (handle.nativeId)` guard; test now simulates tmux env via `process.env.TMUX` and unconditionally asserts `nativeId` is defined + verifies pane mapping
2. **tmux-adapter.test.ts direction assertion** - "respects direction option" and "defaults to horizontal" tests now verify actual `-v`/`-h` flags in `mock.calls` split-window args instead of only checking label
3. **zellij-storage.ts TOCTOU race** - Replaced `existsSync` + `unlinkSync` with try/catch around `unlinkSync`, ignoring `ENOENT`
4. **zellij-adapter.ts setSessionID state clearing** - Added else branch that resets `anchorPaneId`, `hasCreatedFirstPane`, `anchorReadyPromise`, `anchorReadyResolver`, and clears `labelToSpawned` when no persisted state found
5. **terminal.ts dead session_prefix config** - Removed `session_prefix` from `ZellijConfigSchema`, `ZellijAdapterConfig`, `create-managers.ts` forwarding, and all test mock configs
6. **zellij-adapter.ts pkill option parsing** - Added `"--"` separator before the pattern arg to prevent `--session` from being parsed as a pkill option

All 111 tests pass (200 expect calls). No LSP diagnostics on any changed file.

florianschepp:
Addressed the 2 remaining review violations:

1. **zellij-adapter.ts L41**: `setSessionID` now clears `anchorReadyPromise`, `anchorReadyResolver`, and `labelToSpawned` in the loaded-state branch too, preventing stale cross-session anchor sync state.
2. **zellij-adapter.test.ts L178**: Removed duplicate `getPanes` test (identical setup and assertion as L167).

110 tests pass, 0 failures. Ready for re-review.

florianschepp:
@cubic-dev-ai please re-review

florianschepp:
Addressed all 5 review violations:

1. **zellij-storage.test.ts L96**: Replaced `if (existsSync(...))` guard with `expect(existsSync(...)).toBe(true)` assertion so the corruption test fails if the file was never written.
2. **zellij-adapter.ts L84**: Replaced `output.includes(\`terminal_${id}\`)` with word-boundary regex to prevent substring false positives (e.g. `terminal_1` matching `terminal_10`).
3. **zellij-adapter.ts L131**: Wrapped entire spawn path in try/catch. On first-pane spawn failure, resolves `anchorReadyPromise` with null and resets state, preventing subsequent `spawnPane` calls from hanging forever.
4. **tmux-adapter.test.ts L259**: `closePane` test now spawns a pane first via `spawnPane()`, then closes the returned handle, asserting `kill-pane` was called with the correct `nativeId`.
5. **manager.ts L121**: `isEnabled()` adapter path now checks `this.tmuxConfig.enabled && this.adapter.enabled`, so the master switch is respected even when zellij auto-detection creates an adapter.

110 tests pass, 0 failures, 201 expect() calls. All LSP diagnostics clean.

florianschepp:
@cubic-dev-ai please re-review

florianschepp:
@cubic-dev-ai please re-review

florianschepp:
@cubic-dev-ai please re-review

florianschepp:
@cubic-dev-ai please re-review

florianschepp:
@cubic-dev-ai please re-review

florianschepp:
@cubic-dev-ai please re-review

src/shared/terminal-multiplexer/types.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface SpawnOptions {
+  label: string
+  displayName?: string
+  splitFrom?: PaneHandle
+  direction?: "horizontal" | "vertical"
+}
+
+export interface Multiplexer {
+  type: MultiplexerType
+  capabilities: MultiplexerCapabilities
```
florianschepp, line 22: @cubic-dev-ai please re-review

florianschepp:
@cubic-dev-ai All round-7 issues addressed in commit abcf60d:
1. P1 closePane anchor race: Added pendingSpawns counter — anchor only resets when no spawns in-flight
2. P1 hasCreatedFirstPane timeout: No longer clears flag when pane exists but ID capture fails
3. P2 tmux kill-pane: labelToPaneId only deleted on successful kill (exit code 0)
4. P2 ZELLIJ-only env: Falls back to 'default' session name when $ZELLIJ_SESSION_NAME absent
5. isEnabled design doc: Added code comment explaining adapter-independence is intentional
6. Race test: Added concurrent spawn/close regression test exercising pendingSpawns guard

Please re-review.

src/config/schema/terminal.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from "zod"
+
+export const ZellijConfigSchema = z.object({
+  enabled: z.boolean().default(true),
+})
+
+export const TerminalConfigSchema = z.object({
+  provider: z.enum(["auto", "tmux", "zellij"]).default("auto"),
+  zellij: ZellijConfigSchema.optional(),
```
florianschepp: Doesn't make sense at the moment. When provider === 'zellij' implication is its supposed to be enabled. Will fix

florianschepp:
@cubic-dev-ai Addressed all review feedback. Here's a summary of the cleanup changes made since the 5/5 confidence review at `abcf60d`:

## Changes

### Config Simplification
- **Removed `enabled: boolean`** from `Multiplexer` interface, `TmuxAdapter`, and `ZellijAdapter` — the adapter's existence is now the gate (if created, it's enabled)
- **Removed `sessionPrefix`** from both adapter configs (dead code, never used)
- **Deleted `TmuxAdapterConfig` and `ZellijAdapterConfig`** interfaces entirely — they were empty after removing `enabled` and `sessionPrefix`
- **Removed `ZellijConfigSchema`** (`zellij.enabled`) from `terminal.ts` — `create-managers.ts` now relies on `terminal.provider` alone
- **Renamed `TerminalConfigSchema` → `TerminalMultiplexerConfigSchema`** and `TerminalConfig` → `TerminalMultiplexerConfig` for clarity
- **Simplified `TmuxSessionManager` constructor** from overloaded 2/3-arg API with runtime type sniffing to clean `(ctx, adapter | null, tmuxConfig, deps?)` signature
- **Simplified `createMultiplexer()`** from `(type, config?, zellijStorage?)` to `(type, zellijStorage?)`

### Zellij Adapter Refactoring
- **`labelToSpawned: Map<string, boolean>` → `spawnedPanes: Map<string, string | undefined>`** — stores nativeId alongside label for consistent `getPanes()` output
- **Decomposed `spawnPane`** (~150 lines → ~55 lines) into 4 focused helpers: `buildSpawnCommand()`, `capturePaneId()`, `handleFirstPaneAnchoring()`, `stackWithAnchor()`
- **Fixed indentation** in `spawnPane` catch block and `closePane` method (was mixed 0/3/10-space, now consistent 2-space)
- **Simplified `closePane`** — removed unnecessary regex extraction (`handle.label.match(/ses_.../)`) since label IS the sessionId directly; uses SIGKILL directly since pkill exit codes only indicate process match, not death
- **`spawnPane` now returns `nativeId`** on PaneHandle; `getPanes()` also returns it

### Bug Fix
- **Fixed double-close in `onSessionDeleted`** — it was calling `adapter.closePane(handle)` directly AND then calling `removeTrackedSession()` which also closes the handle. Removed the duplicate close; `removeTrackedSession()` is now the sole owner of handle cleanup. Added cleanup call before early return in adapter-only mode (zellij without tmux).

### Test Updates
- All adapter/manager test files updated for new constructor signatures
- Fixed `zombie-pane.test.ts` stale constructor call
- 197 tests pass (111 multiplexer + 86 manager), 0 LSP errors

Ready for re-review.

florianschepp:
@cubic-dev-ai Investigated both flagged violations — neither requires changes as far as I can tell.

**Violation 1 (P2): `terminal.zellij.enabled` removed from schema**

False positive. The `zellij.enabled` field was dead code — zero consumers exist anywhere in `src/`. Grep for `zellij.enabled`, `zellijConfig.enabled`, and `terminal.zellij` returns no matches. The actual enablement logic lives in `isEnabled()` (`manager.ts:110-113`), which checks `this.adapter` for Zellij or `this.tmuxConfig.enabled` for tmux. The nested `zellij.enabled` config was never wired in. Additionally, Zod's `safeParse()` doesn't error on extra fields, so existing user configs with this key are unaffected.

**Violation 2 (P2): `onSessionDeleted` leaks panes when `sourcePaneId` is set**

False positive. `sourcePaneId` (tmux mode) and `adapter` (Zellij mode) are mutually exclusive — they never coexist. The `if (!this.sourcePaneId)` guard at L721 is true exactly when we're in adapter mode, routing to `removeTrackedSession` which calls `adapter.closePane(handle)` at L176-186. When `sourcePaneId` IS set (tmux mode), the code correctly falls through to the tmux-specific path (`queryWindowState` → `decideCloseAction` → `executeAction`), which calls `removeTrackedSession` at L739 or L764.

The suggested fix (`if (this.adapter)` instead of `if (!this.sourcePaneId)`) would actually break adapter-mode cleanup by routing adapter sessions into the tmux-specific `queryWindowStateSafely()` path, which returns `null` without a `sourcePaneId`, trapping sessions in a retry loop that can never succeed.

florianschepp:
@code-yeongyu This PR is ready for review. Summary of changes since initial submission:

- Removed dead config (`sessionPrefix`, `enabled` field, empty adapter config types)
- Fixed double-close bug in `onSessionDeleted` — `removeTrackedSession` is now sole owner of handle cleanup
- Renamed `labelToSpawned` → `spawnedPanes` with proper `Map<string, string | undefined>` typing
- Decomposed `ZellijAdapter.spawnPane` into 4 focused helpers
- Rewrote `closePane` — direct `SIGKILL` (no broken SIGTERM escalation), label-based lookup, `nativeId` on `PaneHandle`
- Renamed `TerminalConfig` → `TerminalMultiplexerConfig`
- Fixed broken indentation in zellij-adapter.ts

All 197 tests pass, 0 LSP errors across changed files.

Please let me know any feedback as well as needed changes, thank you in advance!

florianschepp:
@cubic-dev-ai I've just rebased this PR onto the latest `dev` branch and resolved the merge conflict in `src/create-managers.ts` (merged your `deps?` injection with the async Zellij adapter logic). Could you take another look and let me know if there are any remaining issues to fix, or if this is ready for approval? Thanks!

florianschepp:
@cubic-dev-ai Fixed the P1 issue. Root cause: in `onSessionDeleted`, when both `sourcePaneId` and `adapter` are set, the `!this.sourcePaneId` guard falls through to `this.sessions.get(sessionID)` which returns `undefined` for adapter sessions (they're stored in `sessionHandles`, not `sessions`), causing an early return that skips handle cleanup entirely.

Fix: added an explicit check before the tmux sessions path —

```typescript
if (this.adapter && this.sessionHandles.has(event.sessionID)) {
  this.removeTrackedSession(event.sessionID)
  return
}
```

`removeTrackedSession` already handles adapter cleanup correctly (calls `adapter.closePane(handle)` and deletes from `sessionHandles`). Committed as `092c06b7`.

florianschepp:
Could you take another look and let me know if there are any remaining issues to fix, or if this is ready for approval? Thanks! @cubic-dev-ai

florianschepp:
@cubic-dev-ai you mention 4 unresolved issues from previous reviews, which ones are that?

florianschepp:
@cubic-dev-ai Thanks for the thorough follow-up! After carefully reviewing each of the 4 flagged issues against the actual code, I believe they're all either already handled or not applicable. Here's my reasoning:

**1. Shell command construction (manager.ts:977)**
The shell escaping is already applied via `shellEscapeForDoubleQuotedCommand` before the values are embedded. Both adapters ultimately pass the command through a shell (ZellijAdapter wraps it in `bash -c "...; exec ${cmd}"`, TmuxAdapter passes it as a tmux `split-window` argument) — there's no way to avoid a shell wrapper here, it's inherent to how pane spawning works in terminal multiplexers. The current approach is standard practice.

**2. Zellij storage path validation**
The existing whitelist regex `^[a-zA-Z0-9_\-]+$` already blocks any path traversal completely — no dots, slashes, or special characters can pass. A whitelist approach like this is actually stronger than a resolved-path containment check. Nothing to add here.

**3. hasCreatedFirstPane state after capturePaneId timeout**
If `capturePaneId` times out, `hasCreatedFirstPane` stays `true` with `anchorPaneId === null`. Within the session this degrades gracefully (subsequent panes spawn without stacking, which is acceptable). More importantly, on the next session reload, `setSessionID` calls `validateAnchorPane()` which returns `false` when `anchorPaneId` is null, and explicitly resets `hasCreatedFirstPane = false` — so the state fully self-heals on restart.

**4. TmuxAdapter closePane logging**
Looking at the code directly, `closePane` already logs when tmux resolution fails: `log("[TmuxAdapter.closePane] tmux not found")`. The only silent return is when `!paneId`, which is intentional — there's nothing to close if spawn never produced a pane ID.

Happy to revisit any of these if you see something I'm missing!

florianschepp:
Does that mean it can be approved? @cubic-dev-ai as this were the 4 outstanding issues you mentioned.

florianschepp:
@cubic-dev-ai you mentioned auto approval of yours before a maintainer approves additionally.

florianschepp:
@cubic-dev-ai review

florianschepp:
@cubic-dev-ai Thanks again! Here's how we addressed each:

**P1 — Fixed ✅**
You're right on the regex injection. Even though OpenCode session IDs don't currently contain regex metacharacters in practice, it's correct defensive programming. Added escaping via `.replace(/[.*+?^${}()|[\]\\]/g, "\\\$&")` before interpolating into the pkill pattern (`03a346ce`).

**P2 — Respectfully disagree**
The current `if (this.adapter) return true` is intentional design. The adapter is only created when the user has explicitly configured `terminal.provider: "zellij"` (or auto-detection finds Zellij). `tmuxConfig.enabled` is the flag for the legacy tmux pane integration — Zellij users aren't using tmux and may not have `tmux.enabled: true` set. Applying the suggested change `return this.tmuxConfig.enabled && (this.adapter !== null || this.deps.isInsideTmux())` would silently break Zellij for anyone who hasn't explicitly set `tmux.enabled: true`. The bypass is correct because the adapter represents a distinct configuration path that has its own activation signal.

florianschepp:
@cubic-dev-ai review

florianschepp:
What are the 5 unresolved issues now? @cubic-dev-ai

florianschepp:
@code-yeongyu WIth this, the PR is rebased and ready to be reviewed. If you need any changes please let me know. Thank you in advance! 🙌

zoispag:
@code-yeongyu can this get some traction?

code-yeongyu:
[sisyphus-bot]

- Author: `florianschepp`
- Title: `feat(zellij): add Zellij terminal multiplexer support`
- Closing issues: #1153
- Review decision: none yet

florianschepp:
Thanks @code-yeongyu. Just merged in the current dev branch to resolve the conflicts. As far as I can see the 2 remaining test failures appear to be pre-existing on dev as well and therefore not something introduced by this changes. 

Let me know if there is anything else I can help with here in hope you can take it from here @code-yeongyu . Thank you in advance! 🙌

code-yeongyu:
[sisyphus-bot] Hi florianschepp. 🙏 Thanks for the substantial Zellij terminal multiplexer support.

This is a large feature addition (2930/143 across 22 files) that adds a new terminal-multiplexer surface alongside the existing tmux integration. I want to be candid: adding a parallel multiplexer needs explicit maintainer alignment before it can land.

Several considerations for the maintainer call:

1. *Two multiplexer backends.* OMO currently has a tight coupling to tmux for the `interactive_bash` and team-mode visualization paths. Zellij as an alternative is great in principle, but the maintenance cost is doubled (every change to the multiplexer layer needs to be applied to both).
2. *Detection precedence.* When both tmux and zellij are available, which one wins? An explicit config (`terminal_multiplexer: "auto" | "tmux" | "zellij"`) would be cleaner than a soft auto-detect.
3. *Team-mode interactions.* Team-mode currently assumes tmux for visualization. Zellij has different layout primitives; the visualization layer may need a separate abstraction.

PR shows CONFLICTING against current `dev`. Could you rebase, and add a short design summary covering the three points above? Once that's in, I'll loop a maintainer in for the design call.

Really appreciate the scope-completeness; Zellij is a real user demand and the multiplexer-abstraction direction is the right shape for landing it.

code-yeongyu:
[sisyphus-bot] Thanks @florianschepp for the persistent Zellij work (and the #1226 revival). Closing because all 22 touched files live in the removed top-level `src/` tree, and the PR head's own merge commit broke CI (dropped imports in `src/index.ts`, red typecheck/tests). The Multiplexer abstraction idea itself is sound — if you're still up for it, a fresh implementation against `packages/omo-opencode/src/features/tmux-subagent/` and `create-managers.ts` on current dev would be the right base.

zoispag:
2,5 months later and a lot of effort (and spent tokens) from @florianschepp, and this PR is just closed! @code-yeongyu this is not ok

code-yeongyu:
> 2,5 months later and a lot of effort (and spent tokens) from @florianschepp, and this PR is just closed! @code-yeongyu this is not ok

I think my bot accidentally closed this, really sorry about this. Unfortunately, I have no permission to open this PR again. Can you re open PR and mention me again? I would be happy to review your PR!

PR #3020, opened 2026-04-02
fix(hooks): add proactive tool_use/tool_result pair validator (fixes #3014)

### LINKED ISSUE #3014: bug: Plan mode fails with tool_use/tool_result mismatch — messages.N: tool_use ids found without tool_result blocks
## Bug Description

In Plan mode (Prometheus agent), the Anthropic API rejects the request with:

```
messages.25: 'tool_use' ids were found without 'tool_result' blocks immediately after: toolu_01ThKgs9eMnSiwBKpeAXhknk.
Each 'tool_use' block must have a corresponding 'tool_result' block in the next message.
```

The API requires every assistant message containing a `tool_use` block to be followed by a user message with a matching `tool_result` block. Somewhere in the message pipeline, `tool_result` blocks are being lost or `tool_use`/`tool_result` pairs are getting separated.

## Steps to Reproduce

1. Start an OpenCode session with oh-my-opencode plugin
2. Switch to or invoke Plan mode (Prometheus agent)
3. Interact with the agent — after enough turns (message index ~25), the API call fails
4. Error: `messages.25: 'tool_use' ids were found without 'tool_result' blocks immediately after`

## Suspected Root Cause

The `experimental.chat.messages.transform` hook pipeline runs two transforms before every API call:

1. **Context Injector** (`src/features/context-injector/injector.ts`) — injects synthetic text parts into the last user message via `splice()`
2. **Thinking Block Validator** (`src/hooks/thinking-block-validator/hook.ts`) — prepends thinking blocks to assistant messages that have `tool_use` but no thinking block

### Likely culprits:

**1. Thinking Block Validator (`src/hooks/thinking-block-validator/hook.ts`)**
- Checks `hasContentParts()` for `tool_use` but does NOT verify corresponding `tool_result` blocks exist
- `prependThinkingBlock()` uses `message.parts.unshift()` to mutate the parts array — if the thinking block is malformed or the signature is invalid, the API could reject the entire structure
- For Prometheus sessions that don't use extended thinking, injecting thinking blocks into messages may be incorrect

**2. Compaction / Context Window Recovery**
- If preemptive compaction or context window recovery (`src/hooks/anthropic-context-window-limit-recovery/`) prunes messages, it could remove a user message containing `tool_result` while keeping the preceding assistant message with `tool_use`

**3. Context Injector interaction**
- `splice()` into user message parts could shift part indices if other transforms also read/modify the same message array

## Environment

- oh-my-opencode plugin
- Anthropic Claude provider (Opus/Sonnet)
- Plan mode (Prometheus agent)
- Error occurs at message index 25 (mid-conversation)

## Expected Behavior

All `tool_use` blocks in assistant messages should always have a matching `tool_result` block in the immediately following user message, regardless of which hooks or transforms run on the message array.

## Possible Fixes

1. **Add `tool_use`/`tool_result` pair validation** as a final step in `experimental.chat.messages.transform` — if pairs are broken, either skip the transform or inject placeholder `tool_result` blocks before sending
2. **Skip thinking block injection for Plan mode** — Prometheus may not need thinking blocks at all
3. **Audit compaction/recovery hooks** to ensure they never remove a `tool_result` message without also removing the corresponding `tool_use` message
4. **Make the existing `session-recovery` hook (`recover-tool-result-missing.ts`) run BEFORE the API call** as a preventive check, not just reactively after failure

MoerAI:
## Summary
- Add a proactive tool_use/tool_result pair validator as the final step in the `experimental.chat.messages.transform` pipeline to prevent API rejection errors in Plan mode.

## Problem
When compaction or context-window-limit-recovery removes a user message containing `tool_result` while keeping the preceding assistant message with `tool_use`, the Anthropic API rejects the request with:

```
messages.N: 'tool_use' ids were found without 'tool_result' blocks immediately after
```

The existing `session-recovery` hook handles this REACTIVELY (after the API error), but: the orphaned `tool_use` persists in the message history, causing the error on EVERY subsequent API call. Recovery only fixes the current failing message, not historical orphans.

## Fix
New `tool-result-pair-validator` hook that runs as the LAST step in `experimental.chat.messages.transform` (after context-injector and thinking-block-validator):

1. Scans all assistant messages for `tool_use` parts
2. Checks the next message has matching `tool_result` parts (handles both internal `type: "tool"` with `callID` and API `type: "tool_use"`/`type: "tool_result"` formats)
3. If tool_results are missing, injects synthetic cancelled tool_result parts
4. If no user message follows the assistant message, splices in a synthetic user message

This prevents the API error entirely — proactive instead of reactive.

## Changes
| File | Change |
|------|--------|
| `src/hooks/tool-result-pair-validator/hook.ts` | New hook: scans and fixes orphaned tool_use/tool_result pairs |
| `src/hooks/tool-result-pair-validator/index.ts` | Barrel export |
| `src/hooks/index.ts` | Export new hook factory |
| `src/config/schema/hooks.ts` | Register `tool-result-pair-validator` hook name |
| `src/plugin/hooks/create-transform-hooks.ts` | Create and wire the hook with `safeCreateHook` |
| `src/plugin/messages-transform.ts` | Run validator as the final transform step |

## Design Decisions
- **Runs LAST in the transform pipeline**: Validates the final message state after all other transforms (context injection, thinking block validation) have run
- **Handles both internal and API Part formats**: Checks `callID`, `id`, and `tool_use_id` fields to cover both OpenCode internal format and Anthropic API format
- **Configurable**: Can be disabled via `disabled_hooks: ["tool-result-pair-validator"]`
- **Follows existing patterns**: Same factory pattern, hook type, and registration flow as `thinking-block-validator`

Fixes #3014

---
## Summary by cubic
Prevents Anthropic API errors by validating and repairing `tool_use`/`tool_result` pairs before requests. Fixes #3014.

- **Bug Fixes**
  - Added `tool-result-pair-validator` as the last step in `experimental.chat.messages.transform` to ensure every assistant `tool_use` has a matching `tool_result` in the next user message.
  - Injects synthetic `tool_result` parts (and a synthetic user message when needed) to fix orphans caused by compaction or context-window recovery, preventing repeated API rejections.
  - Supports internal `type: "tool"`/`callID` and Anthropic `type: "tool_use"`/`type: "tool_result"` formats; can be disabled via `disabled_hooks: ["tool-result-pair-validator"]` and is wired into the transform pipeline.

<sup>Written for commit d65f036e2c0e06ba860dd76b134c328e26721a3c. Summary will update on new commits.</sup>

### DIFF
--- a/src/config/schema/hooks.ts
+++ b/src/config/schema/hooks.ts
@@ -25,6 +25,7 @@ export const HookNameSchema = z.enum([
   "interactive-bash-session",
 
   "thinking-block-validator",
+  "tool-result-pair-validator",
   "ralph-loop",
   "category-skill-reminder",
 
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -26,6 +26,7 @@ export { createNonInteractiveEnvHook } from "./non-interactive-env";
 export { createInteractiveBashSessionHook } from "./interactive-bash-session";
 
 export { createThinkingBlockValidatorHook } from "./thinking-block-validator";
+export { createToolResultPairValidatorHook } from "./tool-result-pair-validator";
 export { createCategorySkillReminderHook } from "./category-skill-reminder";
 export { createRalphLoopHook, type RalphLoopHook } from "./ralph-loop";
 export { createNoSisyphusGptHook } from "./no-sisyphus-gpt";
--- a/src/hooks/tool-result-pair-validator/hook.ts
+++ b/src/hooks/tool-result-pair-validator/hook.ts
@@ -1,0 +1,128 @@
+/**
+ * Proactive Tool Result Pair Validator Hook
+ *
+ * Prevents "tool_use ids were found without tool_result blocks" errors
+ * by validating tool_use/tool_result pairing BEFORE sending to the API.
+ *
+ * When compaction or context-window-limit-recovery removes a user message
+ * containing tool_result while keeping the assistant message with tool_use,
+ * the Anthropic API rejects the request. This hook detects orphaned tool_use
+ * blocks and injects synthetic tool_result parts to restore valid pairing.
+ *
+ * Runs as the LAST step in experimental.chat.messages.transform,
+ * after context-injector and thinking-block-validator.
+ */
+
+import type { Message, Part } from "@opencode-ai/sdk"
+import { log } from "../../shared/logger"
+
+interface MessageWithParts {
+  info: Message
+  parts: Part[]
+}
+
+type MessagesTransformHook = {
+  "experimental.chat.messages.transform"?: (
+    input: Record<string, never>,
+    output: { messages: MessageWithParts[] }
+  ) => Promise<void>
+}
+
+function isValidToolCallID(id: unknown): id is string {
+  return typeof id === "string" && /^(toolu_|call_)/.test(id)
+}
+
+function extractToolUseIdsFromAssistant(parts: Part[]): string[] {
+  if (!parts || parts.length === 0) return []
+
+  const ids: string[] = []
+  for (const part of parts) {
+    const type = (part as { type: string }).type
+    if (type !== "tool" && type !== "tool_use") continue
+
+    const callID =
+      (part as { callID?: unknown }).callID ??
+      (part as { id?: unknown }).id
+    if (isValidToolCallID(callID)) {
+      ids.push(callID)
+    }
+  }
+  return ids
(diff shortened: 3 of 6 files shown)


MoerAI:

MoerAI:
Closing as duplicate of #3017 — maintainer already implemented the same fix with tests. Identical approach: proactive tool_use/tool_result pair validator as final step in messages-transform pipeline.

PR #3100, opened 2026-04-03
fix: increase default timeout values to prevent premature agent termination

titet11:
## Problem

Background agents running complex, long-duration tasks (deep code analysis, multi-file refactoring, architecture reviews) are silently terminated by multiple timeout mechanisms before they can complete their work. This is especially problematic for orchestration-heavy setups where agents like Oracle or Hephaestus routinely need 30+ minutes to finish.

The current defaults are too aggressive for real-world multi-agent workflows:

| Constant | Current Default | Effect |
|---|---|---|
| `TASK_TTL_MS` | 30 min | Task marked as error after 30 min |
| `TERMINAL_TASK_TTL_MS` | 30 min | Terminal tasks pruned after 30 min |
| `SESSION_TTL_MS` | 30 min | Session state cleaned up after 30 min of inactivity |
| `DEFAULT_POLL_TIMEOUT_MS` | 30 min | Sync polling gives up after 30 min |
| `DEFAULT_STALE_TIMEOUT_MS` | 45 min | Task considered stale after 45 min without progress |
| `DEFAULT_MESSAGE_STALENESS_TIMEOUT_MS` | 60 min | Task considered stale after 60 min from start |
| `DEFERRED_SESSION_TTL_MS` | 5 min | Queued sessions waiting for concurrency slots dropped after 5 min |

## Root Cause

These timeouts were designed for short-lived tasks, but multi-agent orchestration introduces workloads that routinely exceed these limits. An Oracle agent performing deep diagnostic analysis can take 1-2 hours. A Hephaestus agent implementing complex changes across multiple files can take similar time. When these timeouts fire, the agent's work is silently lost — there is no error message, no recovery, just truncated output.

The `DEFERRED_SESSION_TTL_MS` at 5 minutes is particularly problematic: when all concurrency slots are full, new tasks are queued as "deferred sessions." If slots don't free up within 5 minutes, the queued task is silently dropped. In heavy orchestration scenarios with multiple agents, this happens frequently.

## Solution

Increase all timeout defaults to **100 hours** (360,000,000 ms) to effectively disable premature termination of long-running agent workloads. The `DEFERRED_SESSION_TTL_MS` is increased to **1 hour** (3,600,000 ms) to give queued tasks reasonable time to acquire a concurrency slot.

All values remain user-configurable via `oh-my-opencode.jsonc` under the `background_task` section, so users who prefer shorter timeouts can still set them.

### Changes

**`src/features/background-agent/constants.ts`**
- `TASK_TTL_MS`: `30 * 60 * 1000` → `100 * 60 * 60 * 1000` (30 min → 100 hours)
- `TERMINAL_TASK_TTL_MS`: `30 * 60 * 1000` → `100 * 60 * 60 * 1000` (30 min → 100 hours)
- `DEFAULT_STALE_TIMEOUT_MS`: `2_700_000` → `360_000_000` (45 min → 100 hours)
- `DEFAULT_MESSAGE_STALENESS_TIMEOUT_MS`: `3_600_000` → `360_000_000` (60 min → 100 hours)

**`src/hooks/runtime-fallback/auto-retry.ts`**
- `SESSION_TTL_MS`: `30 * 60 * 1000` → `100 * 60 * 60 * 1000` (30 min → 100 hours)

**`src/tools/delegate-task/timing.ts`**
- `DEFAULT_POLL_TIMEOUT_MS`: `30 * 60 * 1000` → `100 * 60 * 60 * 1000` (30 min → 100 hours)

**`src/features/tmux-subagent/manager.ts`**
- `DEFERRED_SESSION_TTL_MS`: `5 * 60 * 1000` → `60 * 60 * 1000` (5 min → 1 hour)

### Why these values?

- **100 hours** effectively removes the timeout as a constraint — agents will only terminate when they genuinely complete their work or are explicitly cancelled by the user. This aligns with the principle that the orchestrator (Sisyphus) should decide when an agent is done, not an arbitrary timer.
- **1 hour** for deferred sessions gives queued tasks ample time to acquire slots even under heavy concurrency pressure.
- All values remain overridable via user config (`background_task.taskTtlMs`, `background_task.staleTimeoutMs`, etc.), so users who prefer shorter timeouts can still set them explicitly.

These changes apply to version `v3.14.0` of the source code.

## Testing

- Verified that all modified constants are the defaults used when no user config override is present.
- The config schema (`background-task.ts`) already supports user overrides for all these values — no schema changes needed.
- Users who prefer shorter timeouts can set them in their `oh-my-opencode.jsonc` config file.
- Production-tested: these exact values have been running in a patched `dist/index.js` for multiple days with zero premature agent terminations.


code-yeongyu (changes requested):
The problem described is real — 30-minute timeouts are too aggressive for multi-agent orchestration. However, the fix of setting everything to 100 hours (360,000,000ms) effectively disables all timeout protections entirely.

**Issues:**
1. `TASK_TTL_MS = 100 * 60 * 60 * 1000` (100 hours) — this means a stuck task will never be cleaned up. A leaked background agent will run indefinitely.
2. `DEFAULT_STALE_TIMEOUT_MS = 360_000_000` (100 hours) — stale detection is the safety net for when agents stop making progress. Removing it means zombie agents accumulate forever.
3. `DEFERRED_SESSION_TTL_MS = 60 * 60 * 1000` (1 hour) — this one is reasonable. 5 minutes was too short.

**Suggested approach:**
- Increase timeouts to 2-4 hours, not 100 hours. That covers real long-running tasks without losing the safety net.
- Make these configurable via opencode config (e.g. `oh-my-opencode.taskTtlMs`) so users can tune them for their workload.
- Keep stale detection at a reasonable level (e.g. 2 hours) — it exists for a reason.

The current PR trades one problem (premature termination) for a worse one (no termination of stuck/leaked agents).

code-yeongyu:
Thanks for the PR! However, CI is failing because the tests are hardcoded to the old timeout values.

**Failures (11 tests):**
- `DEFAULT_STALE_TIMEOUT_MS > uses a 45 minute default` — expects 2700000, got new value
- `DEFAULT_MESSAGE_STALENESS_TIMEOUT_MS > uses a 60 minute default` — expects old value
- `pruneStaleTasksAndNotifications` tests — multiple failures expecting old TTL values

**Fix options:**
1. Update test expectations to match your new timeout constants (hardcoded values in tests)
2. Better: Import the constants directly and use them in test assertions so they stay in sync

Could you update the tests in `src/features/background-agent/manager.test.ts` to match your new timeout defaults?

titet11:
Done — updated all 5 test files to match the new timeout defaults:

- `default-stale-timeout.test.ts` — updated expected value to `360_000_000`
- `default-message-staleness-timeout.test.ts` — updated expected value to `360_000_000`
- `task-history-cleanup.test.ts` — updated stale offsets to exceed 100h TTL
- `task-poller.test.ts` — updated 8 hardcoded offsets to exceed 100h TTL
- `manager.test.ts` — updated local `TASK_TTL_MS` const + 8 stale date offsets

All 172 tests passing locally across these 5 files. Pushed in commit `95287eb1`.

code-yeongyu:
Great progress! Down to just **1 test failure** now:

```
(fail) timing sync poll timeout defaults > default sync timeout is 30 minutes
```

Almost there — just need to update this last test expectation to match your new timeout constant. The previous fixes resolved 10 of the 11 failing tests!

titet11:
Fixed the last one — updated `timing.test.ts` to expect `100 * 60 * 60 * 1000` instead of `30 * 60 * 1000`. Pushed in commit `cbe65c45`.

All 12 originally failing tests should now pass.

PR #3248, opened 2026-04-08
fix(delegate-task): optimize dynamic provider fallback O(n²) to O(n)

sen7971:
## Summary
Optimizes the dynamic provider fallback in model-selection by removing redundant fuzzyMatchModel calls.

## Problem
The dynamic fallback was calling fuzzyMatchModel for every model in every non-hardcoded provider, resulting in O(n²) complexity when many models are present.

## Solution
Since all models in availableModels are already verified available, simply return the first model from any non-hardcoded provider directly — no fuzzy matching needed.

## Testing
- bun test src/tools/delegate-task/model-selection.test.ts: 22/22 pass
- bun run build: success

---
## Summary by cubic
Optimized dynamic provider fallback to O(n) and removed `opencode` from hardcoded fallback chains to avoid quota stalls. User-configured providers are now picked sooner, and sub-agents won’t hang when `opencode` runs out of quota.

- **Refactors**
  - Use `availableModels` to scan non-hardcoded providers and return the first available model; set `matchedFallback: true` so fallback settings apply.

- **Bug Fixes**
  - Removed `opencode` from `AGENT_MODEL_REQUIREMENTS` and `CATEGORY_MODEL_REQUIREMENTS`; it’s still available via dynamic fallback, allowing providers like minimax/rightcode to surface and preventing quota exhaustion blocking.

<sup>Written for commit ff44db1a07ca4edb23a797ecf9724ce84a24ede5. Summary will update on new commits.</sup>


sen7971:
recheck

sen7971:

sen7971:
Closing - user wants to keep changes local only.

Issue #3238, opened 2026-04-08
Breaking change: ZWSP (U+200B) prefix added to agent names breaks prompt_async API consumers


hy-pony:
## Summary

Starting from a recent version (observed in **3.16.0**), oh-my-openagent injects Zero-Width Space (U+200B) characters as prefixes to agent names during registration. This is a **silent breaking change** that causes `prompt_async` API consumers to fail when they rely on the agent name returned by the `/agent` API.

## Environment

- oh-my-openagent: **3.16.0**
- oh-my-opencode: **3.15.2**
- opencode: **v1.3.3**
- Platform: Linux x86_64

### What changed

Agent names now include invisible ZWSP prefixes for UI sorting purposes:

| Version | Agent Name |
|---|---|
| Before (≤3.15.x?) | `Sisyphus (Ultraworker)` |
| After (3.16.0) | `\u200bSisyphus - Ultraworker` (1× ZWSP prefix) |

Other agents get 2-4 ZWSP prefixes (e.g. `\u200b\u200bSisyphus-Junior`).

### Why this breaks things

The `prompt_async` API accepts an optional `agent` field to specify which agent should handle the request. When this field is omitted, opencode tries to match against the default agent. However:

1. The ZWSP characters are **invisible** — they don't appear in logs, terminal output, or most editors
2. If a consumer previously stored or hardcoded the agent name (without ZWSP), it no longer matches
3. If a consumer calls `/agent` API to discover agent names and passes them to `prompt_async`, the ZWSP-prefixed names work — but this is a non-obvious requirement that wasn't needed before

### Impact

In our case, a separate service (`feishu-bridge`) calls `prompt_async` to forward messages to opencode. After oh-my-openagent auto-upgraded to 3.16.0 (due to no version pinning in `opencode.json`), all messages were **silently dropped** — no error was returned, but the agent never processed them.

This was particularly hard to diagnose because:
- The ZWSP characters are invisible in all standard debugging tools
- No error or warning is logged when agent matching fails
- The system appeared to accept the request but silently did nothing

## Reproduction

1. Install oh-my-openagent 3.16.0 as an opencode plugin
2. Query the `/agent` API — observe agent names contain ZWSP prefixes
3. Call `prompt_async` without the `agent` field, or with the agent name without ZWSP — the message is silently dropped

## Suggestions

1. **Don't inject invisible Unicode characters into agent names** — use a separate field (e.g. `sort_order`, `priority`) for UI sorting instead
[...]

Issue #3251, opened 2026-04-08
[Bug]: Default agent not found because plugin registers zero-width-prefixed agent names but writes plain default_agent


liushidai:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

**Title**
```text
Default agent not found because plugin registers zero-width-prefixed agent names but writes plain default_agent
```

**Body**
```markdown
## Bug Description

`oh-my-opencode` installs and loads successfully, but OpenCode fails to start with:

```text
error: default agent "Sisyphus - Ultraworker" not found
```

The root cause appears to be that the plugin registers agent keys using zero-width-prefixed display names, while `default_agent` is written without that prefix.

As a result, the configured default agent does not exactly match the actual registered agent key.

## Environment

- OpenCode: `1.4.0`
- oh-my-opencode / oh-my-openagent: `3.16.0`
- OS: Linux

## What I Observed

`opencode debug config` shows that the agent keys include zero-width prefixes, for example:

- `\u200bSisyphus - Ultraworker`
- `\u200b\u200bHephaestus - Deep Agent`
- `\u200b\u200b\u200bPrometheus - Plan Builder`

But `default_agent` is set to:

```json
"default_agent": "Sisyphus - Ultraworker"
```

So the exact lookup fails.

I verified this locally:

- exact match between `default_agent` and agent keys: `false`
- normalized match after stripping zero-width chars: `true`

## Expected Behavior

If the plugin remaps agent keys to zero-width-prefixed display names, `default_agent` should resolve to the exact same key.

Alternatively, the plugin should avoid using zero-width-prefixed strings as actual config keys and only use them for UI ordering/display.

## Actual Behavior

The plugin loads, agent configs are injected, but OpenCode cannot resolve the default agent because the configured `default_agent` string does not exactly match the registered agent key.

This causes startup failure with:

```text
default agent "Sisyphus - Ultraworker" not found
```

## Suspected Code Locations

From the built output in `dist/index.js`:

### Zero-width sort prefixes
```js
[...]

Issue #3246, opened 2026-04-08
[Bug]: After updating omo to 3.16.0, the agent name displays incorrectly in the TUI


FurryWolfX:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

After updating omo to 3.16.0, the agent name displays incorrectly in the TUI

## 3.15.3

<img width="254" height="123" alt="Image" src="https://github.com/user-attachments/assets/0fd49ceb-cb82-48c7-8664-0c8cbaea4164" />

## 3.16.0

<img width="253" height="118" alt="Image" src="https://github.com/user-attachments/assets/8f601768-2bb8-4155-9528-4bed0cbdd531" />

### Steps to Reproduce

1. run opencode

### Expected Behavior

displays correctly

### Actual Behavior

displays incorrectly

<img width="253" height="118" alt="Image" src="https://github.com/user-attachments/assets/8f601768-2bb8-4155-9528-4bed0cbdd531" />

### Doctor Output

```shell
omo: 3.16.0 (install via sourcecode)
terminal: windows terminal
os: wsl2
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Linux

### OpenCode Version

1.4.0

code-yeongyu:
Duplicate of #3238/#3244 — same ZWSP root cause. Already fixed by #3242 (merged into `dev`). Will land in the next patch release.

In the meantime you can test the fix via `bunx oh-my-openagent@dev` if you want it immediately.

This is likely related to your other report (#3245) — if the `/compact` failure is caused by the same ZWSP contamination in session state, the `@dev` build should fix both. Please try it and let us know on #3245.

Issue #3244, opened 2026-04-08
[Bug]: Severe display error regression cause by ZWSP


WhiteGiverMa:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

ZWSP display error in #3150 happened again in 3.16.0. Both 4 primary agents are affected now.

<img width="1478" height="927" alt="Image" src="https://github.com/user-attachments/assets/be57df77-7bfb-4c38-812d-a36ce800ecee" />

<img width="1047" height="551" alt="Image" src="https://github.com/user-attachments/assets/4c3fbbfd-b887-401f-a3ab-b7c3ed101890" />

### Steps to Reproduce

1.install opencode v1.4.0
2. install omo v3.16.0
3. open the TUI. Both 4 primary agents' name are a ZWSP there.

### Expected Behavior

`Sisyphus (Ultraworker)`, `Atlas (Plan Executor)` correctly display.

### Actual Behavior

Display error with character lack.

### Doctor Output

```shell
PS C:\Users\A1337> bunx oh-my-openagent doctor

 oMoMoMoMo Doctor

 ⚠ 2 issues found:

1. Model override uses unavailable provider
   Provider(s) not found in OpenCode model cache: qianfan
   Affects: model resolution

2. Configured models rely on compatibility fallback
   sisyphus=opencode/glm-5.1 (heuristic-backed)
   Affects: sisyphus
```

### Error Logs

```shell

```

### Configuration

```json
{
  "$schema": "https://raw.githubusercontent.com/code-yeongyu/oh-my-openagent/dev/assets/oh-my-opencode.schema.json",
  "agents": {
    "hephaestus": {
      "model": "openai/gpt-5.4",
      "variant": "medium"
    },
    "oracle": {
      "model": "openai/gpt-5.4",
      "variant": "high"
    },
    "librarian": {
      "model": "opencode-go/minimax-m2.7"
    },
    "explore": {
      "model": "opencode-go/minimax-m2.7"
    },
    "multimodal-looker": {
[...]

PR #3252, opened 2026-04-08
fix(default-agent): use getAgentListDisplayName to include sorting prefix

jlwainwright:
# Bug: `default_agent` fails with "agent not found" due to zero-width prefix mismatch

## Summary

When `oh-my-openagent` is configured with `default_agent: "sisyphus"`, running `opencode run` fails with:
```
error: default agent "Sisyphus - Ultraworker" not found
```

The bug occurs because `default_agent` is set to the display name **without** zero-width sorting prefixes, but agent keys are remapped **with** these prefixes. OpenCode's agent lookup does exact string matching and fails.

## Steps to Reproduce

1. Install oh-my-openagent: `bun add oh-my-openagent@latest`
2. Configure with `default_agent: "sisyphus"` in `oh-my-openagent.json`
3. Run: `opencode run "Say hello!"`
4. Error: `default agent "Sisyphus - Ultraworker" not found`

## Root Cause

In `src/plugin-handlers/agent-config-handler.ts`:

```typescript
// default_agent is set WITHOUT zero-width prefix:
(params.config as { default_agent?: string }).default_agent =
  getAgentDisplayName(configuredDefaultAgent);  
// Result: "Sisyphus - Ultraworker"

// But agent keys are remapped WITH zero-width prefix:
params.config.agent = remapAgentKeysToDisplayNames(params.config.agent);
// Result: { "\u200BSisyphus - Ultraworker": {...}, ... }
```

When OpenCode looks up `"Sisyphus - Ultraworker"` in the agents object, it finds `"\u200BSisyphus - Ultraworker"` (with zero-width space `\u200B`) → mismatch!

The zero-width prefixes are intentional for UI sorting order (see `AGENT_LIST_SORT_PREFIXES`), but `default_agent` should use the same prefixed form to match the remapped keys.

## Affected Versions

- v3.14.0 - v3.16.0+ (likely present since agent key remapping was introduced)
- The v3.14.0 changelog mentions "Agent键别名保留" fixing related issues, but `default_agent` was not included in that fix

## Suggested Fix

**File:** `src/plugin-handlers/agent-config-handler.ts`

**Lines:** ~107-111

```diff
  if (isSisyphusEnabled && builtinAgents.sisyphus) {
    if (configuredDefaultAgent) {
-     (params.config as { default_agent?: string }).default_agent =
-       getAgentDisplayName(configuredDefaultAgent);
+     (params.config as { default_agent?: string }).default_agent =
+       getAgentListDisplayName(configuredDefaultAgent);
    } else {
-     (params.config as { default_agent?: string }).default_agent =
-       getAgentDisplayName("sisyphus");
+     (params.config as { default_agent?: string }).default_agent =
+       getAgentListDisplayName("sisyphus");
    }
```

This ensures `default_agent` includes the same zero-width sorting prefix as the remapped agent keys, allowing OpenCode's agent lookup to succeed.

## Workaround

Until fixed, users can work around this by:

1. **Set via environment variable:**
   ```bash
   OPENCODE_DEFAULT_AGENT=build opencode run "Say hello!"
   ```

2. **Use pure mode (no plugins):**
   ```bash
   opencode run "Say hello!" --pure
   ```

3. **Use the oh-my-opencode CLI wrapper:**
   ```bash
   bunx oh-my-opencode run "Say hello!"
   ```
   (Note: this also has the same underlying bug but may handle it differently)

## Additional Context

- Zero-width sorting prefixes are defined in `AGENT_LIST_SORT_PREFIXES` for UI ordering
- The `getAgentDisplayName()` function returns plain display names
- The `getAgentListDisplayName()` function adds the sorting prefix
- Related: v3.14.0 fixed similar issues in `atlas/boulder-continuation-injector.ts` and `features/claude-code-session-state/state.ts`

## Labels

`bug`, `good first issue`, `regression`

---
## Summary by cubic
Fixes "agent not found" when a default agent is set. We now use getAgentListDisplayName to include the zero-width sorting prefix so `default_agent` matches remapped agent keys.

<sup>Written for commit 5c6b0b07b0c4f3d35797867f22d3b8e4d23a66a3. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin-handlers/agent-config-handler.ts
+++ b/src/plugin-handlers/agent-config-handler.ts
@@ -3,7 +3,7 @@ import { createSisyphusJuniorAgentWithOverrides } from "../agents/sisyphus-junio
 import type { OhMyOpenCodeConfig } from "../config";
 import { isTaskSystemEnabled, log, migrateAgentConfig } from "../shared";
 import { AGENT_NAME_MAP } from "../shared/migration";
-import { getAgentDisplayName } from "../shared/agent-display-names";
+import { getAgentDisplayName, getAgentListDisplayName } from "../shared/agent-display-names";
 import { registerAgentName } from "../features/claude-code-session-state";
 import {
   discoverConfigSourceSkills,
@@ -159,10 +159,10 @@ export async function applyAgentConfig(params: {
   if (isSisyphusEnabled && builtinAgents.sisyphus) {
     if (configuredDefaultAgent) {
       (params.config as { default_agent?: string }).default_agent =
-        getAgentDisplayName(configuredDefaultAgent);
+        getAgentListDisplayName(configuredDefaultAgent);
     } else {
       (params.config as { default_agent?: string }).default_agent =
-        getAgentDisplayName("sisyphus");
+        getAgentListDisplayName("sisyphus");
     }
 
     // Assembly order: Sisyphus -> Hephaestus -> Prometheus -> Atlas


code-yeongyu:
Thanks for the detailed writeup @jlwainwright — excellent root-cause analysis! 🙏

Heads up: this exact bug was already fixed in #3242, which merged into `dev` earlier today (closes #3238 / #3244 / #3246 / #3251 — they're all the same root cause).

We went in the **opposite direction** from your fix though, and I think it's worth explaining the reasoning so you can decide whether you want to close this PR or pivot it:

### The two possible fixes

| Approach | Result |
|---|---|
| **Your PR (#3252)**: make `default_agent` include the ZWSP prefix | Matches the ZWSP'd keys → lookup works |
| **#3242**: strip ZWSP from `config.agent` keys entirely | Keys are clean → plain `default_agent` matches them |

### Why we chose the strip-ZWSP direction

The ZWSP prefix was only ever meant as an internal sort hint for the TUI agent list. But #3238 showed it was leaking into the plugin API surface — external consumers of `prompt_async` (and other hooks that read `config.agent` keys) were getting ZWSP-contaminated agent names.

Adding ZWSP to `default_agent` would patch the one symptom you hit, but the ZWSP would still be visible to:
- Plugin API consumers (#3238)
- TUI display paths that don't strip it (#3244 / #3246)
- Session state (#3251)
- Anywhere else someone reads `config.agent` keys in the future

#3242 removes ZWSP from all four callsites that were injecting it (`agent-key-remapper`, `agent-priority-order`, `command-config-handler`, `tool-config-handler`), deprecates `getAgentListDisplayName()`, and preserves sort order via JS object insertion order + the `order` field injection (which the upcoming opencode PR #19127 will consume).

After #3242, the default_agent path you're touching actually works correctly — because the keys on the other side no longer have ZWSP to mismatch against.

### What to do

If you've reproduced the bug on `dev` (not 3.16.0 from npm), please let me know and I'll dig deeper — that would mean there's still a callsite I missed.

Otherwise, could you verify against `bunx oh-my-openagent@dev` and close this PR if your issue is resolved? The `getAgentListDisplayName()` function is also being deprecated, so the direction of your diff would reintroduce a code smell.

Sorry for the overlap — you did great work tracking this down, just ran into the "merged to dev but not yet published" window!

code-yeongyu:
This PR is now superseded by #3329 which permanently fixes the agent ordering issue.

The root cause was reliance on `Object.entries()` insertion order, which is non-deterministic across config merges. The fix introduces:

1. `CANONICAL_CORE_AGENT_ORDER = ["sisyphus", "hephaestus", "prometheus", "atlas"] as const` - single source of truth
2. Alphabetical sorting for non-core agents (deterministic)
3. 12 regression tests including 100-permutation stability test

Approaches like ZWSP prefixes, sort shims, or display name heuristics are now forbidden patterns documented in `src/plugin-handlers/AGENTS.md`.

Thank you for your contribution attempt, but this PR's approach would re-introduce the yo-yo bug.

Issue #3257, opened 2026-04-08
[Bug]: v3.16.0 still ships stale oh-my-opencode identity in the auto-update/startup-version path, so the startup toast can still show unknown


NikkeTryHard:

code-yeongyu:
Confirmed the bug and shipped a fix — see #3258.

**Root cause:** `src/hooks/auto-update-checker/` had three read paths that all hardcoded `pkg.name === "oh-my-opencode"`:

1. `INSTALLED_PACKAGE_JSON` only pointed at `cache/node_modules/oh-my-opencode/package.json`
2. `findPackageJsonUp()` in `package-json-locator.ts` rejected any walked-up `package.json` whose name wasn't the canonical
3. `getLocalDevPath()` in `local-dev-path.ts` only matched `file://` entries containing the canonical name

Users who install via `oh-my-openagent@X` from npm end up with `node_modules/oh-my-openagent/package.json` (the publish workflow rewrites `.name` during the alias publish step in `.github/workflows/publish.yml`). All three resolvers silently missed the file, `getCachedVersion()` returned `null`, and the startup toast fell back to "unknown".

**Fix:** introduce `ACCEPTED_PACKAGE_NAMES` + `INSTALLED_PACKAGE_JSON_CANDIDATES` and make the three readers name-tolerant. Writes are untouched on purpose — the auto-update-checker owns its own cache workspace and stays internally consistent with the canonical name.

**Verification:**
- `bun test src/hooks/auto-update-checker/` → 54 pass, 0 fail (4 new tests cover both install paths + priority order)
- `bun run typecheck` → clean
- full suite → 4444 pass, 0 fail

Will land in the next patch release once the PR is reviewed. Thanks for the clear title — it pointed straight at the problem.

Issue #3188, opened 2026-04-07
[Bug]: Opencode hangs when start-up with new version of oh-my-opencode registered as "oh-my-openagent"


prayerlee:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

Opencode hangs when start-up with new version of oh-my-opencode registered as "oh-my-openagent".

### Steps to Reproduce

Log show Opencode can not find a plugin with name "oh-my-openagent".

Deleting "oh-my-openagent" from the opencode plugin list and Opencode can start-up.
Changing the plugin name to 'oh-my-opencode' is also fine, provided that an older version of the plugin (e.g., 3.14.0) is installed.

### Expected Behavior

Opencode can start normally and load the oh-my-opencode plugin

### Actual Behavior

Opencode hangs when start-up with new version of oh-my-opencode registered as "oh-my-openagent".

### Doctor Output

```shell
OpenCode：1.3.17
oh-my-opencode: 3.15.3

1. Using legacy package name
   Your opencode.json references "oh-my-opencode" which has been renamed to "oh-my-openagent". The old name may stop working in a future release.
   Fix: Update your opencode.json plugin entry: "oh-my-opencode@latest" → "oh-my-openagent@latest"
   Affects: plugin loading
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Windows

### OpenCode Version

1.3.17

code-yeongyu:
Thanks for reporting. This appears to be related to the package name transition. Both `oh-my-opencode` and `oh-my-openagent` are published on npm (both at v3.15.3), but OpenCode's plugin loader may have issues resolving the legacy name.

**Workaround:** Change your `opencode.json` plugin entry from `oh-my-openagent` to `oh-my-opencode`:
```json
{
  "plugins": {
    "oh-my-opencode": "latest"
  }
}
```

We'll investigate why the legacy name causes a hang in the plugin loader.

prayerlee:
[...]

Issue #3129, opened 2026-04-04
[Bug]: startup version toast still shows 'unknown' because auto-update checker still hardcodes legacy package name


NikkeTryHard:
## Summary

Even with the plugin correctly installed as `oh-my-openagent@3.15.0`, the startup version toast can still show `unknown`.

This is not just a bad local install. I was able to make it deterministically flip between `unknown` and `3.15.0` based on whether a legacy compatibility path exists in the OpenCode cache.

## Current Environment

- OpenCode: `1.3.15`
- Plugin actually installed at: `~/.config/opencode/node_modules/oh-my-openagent`
- Installed package version: `3.15.0`
- Config plugin entry uses canonical package name: `oh-my-openagent`

## Actual Behavior

Startup log sometimes shows:

```text
[auto-update-checker] Startup toast shown: vunknown
```

and sometimes shows:

```text
[auto-update-checker] Startup toast shown: v3.15.0
```

while the plugin itself is loading correctly in both cases.

## Key Evidence

In the installed upstream bundle:

- `~/.config/opencode/node_modules/oh-my-openagent/dist/index.js`

there is still legacy version-resolution code hardcoded to:

```js
var PACKAGE_NAME = "oh-my-opencode";
```

That same auto-update/version code then uses the legacy package name for cache lookup and fallback logic.

## What makes it flap

The startup toast becomes deterministic based on whether this legacy path exists:

```text
~/.cache/opencode/node_modules/oh-my-opencode/package.json
```

Observed behavior:

- if that legacy path is missing -> startup toast shows `vunknown`
- if that legacy path exists and resolves to the real installed package -> startup toast shows `v3.15.0`

## Repro / Proof this is upstream-facing

Canonical plugin install is untouched and valid:

```text
~/.config/opencode/node_modules/oh-my-openagent/package.json
version = 3.15.0
```

But startup toast still showed `vunknown` until I added a local compatibility shim:

```bash
mkdir -p ~/.cache/opencode/node_modules
ln -s ~/.config/opencode/node_modules/oh-my-openagent ~/.cache/opencode/node_modules/oh-my-opencode
```

After that, startup logs became:

```text
[auto-update-checker] Startup toast shown: v3.15.0
```

repeatedly.

That strongly suggests the startup version path is still reading the old package name / cache root instead of the actual canonical package install.

## Why previously closed issues may not cover this

[...]

Issue #2823, opened 2026-03-25
Package rename from oh-my-opencode to oh-my-openagent breaks existing configurations


seungmun:
## Description

After the package was renamed from `oh-my-opencode` to `oh-my-openagent` (v3.11.0, 2026-03-07), existing OpenCode configurations that reference the old package name silently fail to load the plugin.

The rename itself happened ~18 days ago, but both npm packages (`oh-my-opencode` and `oh-my-openagent`) have been dual-published at the same version since then. Despite this, at some point the old `oh-my-opencode` package stopped loading correctly in OpenCode — **without any error message or warning**.

### Key Symptom: Silent Failure with Clean Doctor Output

Running `bunx oh-my-opencode doctor --verbose` reports everything as normal — no errors, no warnings. Yet the plugin simply does not load at startup. There is **zero indication** that the package name change is the cause. The user has to independently discover the rename by visiting the GitHub repo and noticing "previously oh-my-opencode" in the description.

## Steps to Reproduce

1. Have `oh-my-opencode` configured in `~/.config/opencode/opencode.json`:
   ```json
   {
     "plugin": ["oh-my-opencode@latest"]
   }
   ```
2. Start OpenCode
3. Plugin fails to load — no error, no warning
4. Run `bunx oh-my-opencode doctor --verbose` — reports clean, no issues found
5. User is left with no clue why the plugin isn't working

## Expected Behavior

At minimum one of:
- `bunx oh-my-opencode doctor` should detect and warn about the package rename
- The old npm package name (`oh-my-opencode`) should show a deprecation notice pointing to `oh-my-openagent`
- OpenCode startup should surface a clear error when a configured plugin fails to load

## Actual Behavior

Plugin silently fails to load. Doctor reports clean. Users have to manually discover the rename by visiting the GitHub repo.

## Workaround (Confirmed Working)

Edit `~/.config/opencode/opencode.json` and change the plugin name:

```diff
{
  "plugin": [
-   "oh-my-opencode@latest"
+   "oh-my-openagent@latest"
  ]
}
```

Restart OpenCode. Plugin loads normally.

## Context

- The rename happened in v3.11.0 (2026-03-07) per release notes: *"This is a first release as an oh-my-openagent"*
- Both `oh-my-opencode` (184 versions, 108K weekly downloads) and `oh-my-openagent` (9 versions, 4.8K weekly downloads) are currently published on npm
[...]

PR #3261, opened 2026-04-09
fix(publish): generate openagent alias packages from canonical manifests

NikkeTryHard:
## Summary

This PR fixes the `oh-my-opencode` / `oh-my-openagent` identity drift at the release boundary.

The main change is that `oh-my-openagent` is now published from generated alias package directories derived from the canonical `oh-my-opencode` manifests. The release workflows no longer rewrite tracked `package.json` files in place and then try to restore them afterward.

That keeps the architecture simpler:

- runtime identity stays canonical as `oh-my-opencode`
- `oh-my-openagent` remains a publish-time compatibility alias
- release metadata is generated deterministically from canonical manifests instead of being patched ad hoc in CI

## Root cause

The recent regressions were not just about one stale constant. The deeper problem was that the publish pipeline was synthesizing `oh-my-openagent` by mutating checked-out manifests in place while the runtime and package ecosystem still treat `oh-my-opencode` as the canonical identity.

That made it too easy for released artifacts to drift:

- some surfaces were still canonical
- some were alias-specific
- some were partially rewritten in CI
- some were restored afterward

This PR removes that fragile boundary.

### 1. Added explicit alias-package generators

New files under `script/publish/`:

- `identity.ts`
- `alias-main-package.ts`
- `build-alias-package.ts`
- `alias-platform-package.ts`
- `build-alias-platform-package.ts`

These helpers centralize the canonical and alias names and generate `oh-my-openagent` publish artifacts from canonical package data.

### 2. Updated publish workflows to use generated temp directories

Updated:

- `.github/workflows/publish.yml`
- `.github/workflows/publish-platform.yml`

Instead of mutating the checked-out manifest with `jq`, the workflows now:

1. build the canonical package as before
2. generate an alias package into `$RUNNER_TEMP/...`
3. publish from that generated directory

This removes the rename/restore cycle from CI.

### 3. Kept runtime identity canonical

This PR does **not** try to make runtime logic bilingual across both names.

That is intentional.

The cleaner boundary here is:

- canonical runtime/package identity: `oh-my-opencode`
- generated publish alias: `oh-my-openagent`

Trying to spread alias handling deeper into startup/update/runtime code would make the internal identity model harder to reason about and would not actually address the publish-time failure mode that caused the shipped drift.

### 4. Added targeted regression coverage

Added:

- `script/publish/alias-main-package.test.ts`
- `script/publish/alias-platform-package.test.ts`

These tests verify that the generated alias artifacts have the expected:

- package names
- bin names
- schema export path
- optional dependency keys
- platform package metadata

without mutating the canonical source manifests.

## Verification

Ran:

- `bun run typecheck`
- `bun test script/publish/alias-main-package.test.ts script/publish/alias-platform-package.test.ts bin/platform.test.ts`

Also generated dry-run alias artifacts locally and inspected the resulting manifests.

Confirmed for the generated main alias package:

- `name: oh-my-openagent`
- `bin.oh-my-openagent = bin/oh-my-openagent.js`
- `exports["./schema.json"] = ./dist/oh-my-openagent.schema.json`
- `optionalDependencies` are rewritten to `oh-my-openagent-*`

Confirmed for the generated platform alias package:

- package name becomes `oh-my-openagent-<platform>`
- description is rewritten accordingly
- platform binary entry becomes `oh-my-openagent`

## Why this should be easier to maintain

The release aliasing logic now lives in source files with direct tests instead of in workflow-local manifest mutations.

That makes future publish behavior easier to review, easier to test, and much harder to ship partially.

## Scope

This PR intentionally fixes the release root cause only.

It does not attempt a broader cleanup of every remaining user-facing `OhMyOpenCode` string or every legacy canonical naming surface in the runtime. Those can be handled separately without reopening the publish-time identity problem.

### DIFF
--- a/.github/workflows/publish-platform.yml
+++ b/.github/workflows/publish-platform.yml
@@ -65,7 +65,7 @@ jobs:
           VERSION="$INPUT_VERSION"
           DIST_TAG="$INPUT_DIST_TAG"
 
-          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z]+(\.[0-9A-Za-z]+)*)?$ ]]; then
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$ ]]; then
             echo "::error::Invalid version: $VERSION"
             exit 1
           fi
@@ -85,15 +85,11 @@ jobs:
         run: |
           PLATFORM_KEY="${{ matrix.platform }}"
           PLATFORM_KEY="${PLATFORM_KEY//-/_}"
-          
-          # Check oh-my-opencode
           OC_STATUS=$(curl -s -o /dev/null -w "%{http_code}" "https://registry.npmjs.org/oh-my-opencode-${{ matrix.platform }}/${VERSION}")
-          # Check oh-my-openagent
           OA_STATUS=$(curl -s -o /dev/null -w "%{http_code}" "https://registry.npmjs.org/oh-my-openagent-${{ matrix.platform }}/${VERSION}")
-          
+
           echo "oh-my-opencode-${{ matrix.platform }}@${VERSION}: ${OC_STATUS}"
           echo "oh-my-openagent-${{ matrix.platform }}@${VERSION}: ${OA_STATUS}"
-          
           if [ "$OC_STATUS" = "200" ]; then
             echo "skip_opencode=true" >> $GITHUB_OUTPUT
             echo "✓ oh-my-opencode-${{ matrix.platform }}@${VERSION} already published"
@@ -101,7 +97,6 @@ jobs:
             echo "skip_opencode=false" >> $GITHUB_OUTPUT
             echo "→ oh-my-opencode-${{ matrix.platform }}@${VERSION} needs publishing"
           fi
-          
           if [ "$OA_STATUS" = "200" ]; then
             echo "skip_openagent=true" >> $GITHUB_OUTPUT
             echo "✓ oh-my-openagent-${{ matrix.platform }}@${VERSION} already published"
@@ -109,12 +104,14 @@ jobs:
             echo "skip_openagent=false" >> $GITHUB_OUTPUT
             echo "→ oh-my-openagent-${{ matrix.platform }}@${VERSION} needs publishing"
           fi
-          
-          # Skip build only if BOTH are already published
+
+          # Need artifact if either package needs publishing.
           if [ "$OC_STATUS" = "200" ] && [ "$OA_STATUS" = "200" ]; then
             echo "skip=true" >> $GITHUB_OUTPUT
+            echo "skip_${PLATFORM_KEY}=true" >> $GITHUB_OUTPUT
           else
             echo "skip=false" >> $GITHUB_OUTPUT
+            echo "skip_${PLATFORM_KEY}=false" >> $GITHUB_OUTPUT
           fi
 
       - name: Update version in package.json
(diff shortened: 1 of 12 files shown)


NikkeTryHard:

NikkeTryHard:
Addressed the valid cubic findings in the latest push:

- added checkout and Bun setup to the platform publish job before invoking repo scripts
- tightened publish helper flag parsing so missing values cannot be misread as other flags
- made alias runtime rewrites fail loudly when expected replacements drift
- added regression coverage for the stricter failure paths

@cubic-dev-ai please re-run review on the latest commit.

NikkeTryHard:
Follow-up hardening is now pushed.

I did a broader review pass across the full PR diff, including multiple independent reviews plus a final semantic check on the original bug mechanism.

Key conclusion: the alias-package runtime rewrites are intentional and necessary here. The `version unknown` bug was caused by an installed `oh-my-openagent` package still carrying stale internal `oh-my-opencode` identity in the auto-update/version-discovery path. Leaving the published alias package internally canonical would preserve that bug path.

Latest hardening in this push:
- fail loudly if the canonical CLI wrapper needed to seed the alias wrapper is missing
- reject platform bin paths that escape the generated output directory
- publish the generated alias package from its own directory without masking real publish failures behind `|| echo`
- add regression coverage for those failure paths

Verification on current branch state:
- `bun run typecheck`
- `bun test script/publish/alias-main-package.test.ts script/publish/alias-platform-package.test.ts bin/platform.test.ts`

So the current branch keeps canonical `oh-my-opencode` in source and build inputs, but makes the separately published `oh-my-openagent` artifacts internally self-consistent enough to avoid shipping the same name-coupling bug again.

NikkeTryHard:
FYI

Directly related history:
- #2650 fix(release): add oh-my-openagent dual-publish to platform and main workflows
- #2811 fix(publish): align repo guard and test topology

User-facing rename / identity fallout this PR is aligned with:
- #2823 Package rename from oh-my-opencode to oh-my-openagent breaks existing configurations
- #2749 Plugin still displays as "oh-my-opencode" in OpenCode status bar — npm package name and hardcoded references not updated
- #3129 [Bug]: startup version toast still shows 'unknown' because auto-update checker still hardcodes legacy package name
- #3257 [Bug]: v3.16.0 still ships stale oh-my-opencode identity in the auto-update/startup-version path, so the startup toast can still show unknown

Still-open fallout that this direction also helps reduce:
- #3188 [Bug]: Opencode hangs when start-up with new version of oh-my-opencode registered as "oh-my-openagent"
- #2678 oh-my-openagent@3.12.3 still resolves legacy oh-my-opencode-windows-x64 binaries on Windows

code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED

code-yeongyu:
[sisyphus-bot] Hi NikkeTryHard. 🙏 Thanks for the deep work on the dual-publish identity drift.

Picking this back up from the 4/16 approval. This is a 1117-line PR touching publish workflows + 12 files; the publish pipeline is also one of the most policy-sensitive areas (`AGENTS.md` explicitly forbids local `bun publish` and any modification of `package.json` version outside the workflow). The PR shows CONFLICTING against current `dev`, and given the surface area I'd want a maintainer (code-yeongyu) to take the final review pass here.

A few things I'd like the rebase to address before re-review:

1. The publish-platform workflow on `dev` has changed since 4/16. Please re-derive the generated-alias-package step against the latest `publish.yml` + `publish-platform.yml` shape.
2. `signatures/cla.json` may also need a touch (the file is updated by the CLA bot, so it shouldn't conflict on rebase, but worth confirming).
3. Add a one-line rollback note to the PR description: if the alias-from-canonical-manifest path ever produces a broken `oh-my-openagent` tarball, what's the manual revert? (For a 1k-line change to a release pipeline, a documented rollback story makes the maintainer call easier.)

Once that's in I'll get a maintainer to do the final pass. Sorry for the slow turnaround on such substantive work; the dual-publish problem is real and the canonical-manifest approach is the right shape.

code-yeongyu:
[sisyphus-bot] Hi NikkeTryHard. Thank you for the deep dual-publish identity work. The publish pipeline has since been reworked and now generates the oh-my-openagent alias packages from the canonical manifests in publish.yml, so this branch is superseded. Closing with gratitude — the analysis here informed the current setup.

Issue #3283, opened 2026-04-09
[Bug]: Agent name rendering gets cut off


GodTamIt:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

After upgrading to the latest `oh-my-openagent` and `opencode`, it seems that the agent name gets cut off in weird ways.

### Steps to Reproduce

1. Configure `oh-my-opencode`
2. Use `opencode` v1.4.0
3. Launch `opencode` and notice broken names.

### Expected Behavior

Have all of the letters of the agent's name.

### Actual Behavior

<img width="3824" height="2084" alt="Image" src="https://github.com/user-attachments/assets/becb2dac-aa49-4751-b8a3-42442df91069" />

<img width="3824" height="2084" alt="Image" src="https://github.com/user-attachments/assets/772168f1-c43c-4a69-810b-5ea66000f727" />

<img width="3824" height="2084" alt="Image" src="https://github.com/user-attachments/assets/deddcdaf-1e52-4268-99b8-20b17e1b9934" />

### Doctor Output

```shell
oMoMoMoMo Doctor 

 ⚠ 4 issues found:

1. Model override uses unavailable provider
   Provider(s) not found in OpenCode model cache: deepinfra-custom
   Affects: model resolution

2. Comment checker unavailable
   Comment checker binary is not installed.
   Fix: Install @code-yeongyu/comment-checker
   Affects: comment-checker hook

3. GitHub CLI missing
   gh CLI is not installed.
   Fix: Install from https://cli.github.com/
   Affects: GitHub automation

4. Configured models rely on compatibility fallback
[...]

Issue #3259, opened 2026-04-09
ZWSP sort prefixes in AGENT_LIST_SORT_PREFIXES render as visible space in some terminals


ayixiayi:
## Bug

`AGENT_LIST_SORT_PREFIXES` in `src/shared/agent-display-names.ts` uses U+200B (Zero-Width Space) characters as invisible sort prefixes for core agents:

```ts
var AGENT_LIST_SORT_PREFIXES = {
  sisyphus: "\u200B",
  hephaestus: "\u200B\u200B",
  prometheus: "\u200B\u200B\u200B",
  atlas: "\u200B\u200B\u200B\u200B"
};
```

These prefixes are prepended to agent display names via `getAgentListDisplayName()` and end up in the TUI status bar.

## Problem

Many terminal emulators (e.g., Ghostty on WSL2) render ZWSP as a visible black square or extra space character. This causes:
- A visible black gap before "Sisyphus - Ultraworker" in the bottom status bar
- Text appearing misaligned/shifted

## Reproduction

1. Run opencode with oh-my-openagent in a terminal that renders ZWSP visibly (Ghostty, some versions of Windows Terminal)
2. Look at the bottom status bar showing the current agent name
3. Notice a black space/square before "Sisyphus"

Copying the text and inspecting with `charCodeAt(0)` confirms character code 8203 (U+200B).

## Suggested Fix

Since agent ordering is already handled by the `order` field injected via `injectOrderField()` in `reorderAgentsByPriority()`, the ZWSP prefixes are redundant for sorting. They could be:

1. Removed entirely (rely on `order` field)
2. Stripped before rendering in the TUI status bar (already done for config key resolution via `stripAgentListSortPrefix`, but not for display)

## Environment

- oh-my-openagent: 3.16.0
- opencode: 1.4.0
- Terminal: Ghostty (WSL2, mirror networking mode)
- OS: Ubuntu on WSL2

code-yeongyu:
Great catch, and thank you for the clean repro with terminal details. Shipped a fix in #3260.

You're right on every point:
- `AGENT_LIST_SORT_PREFIXES` was dead code in `dev` as of #3242 (which removed all four production call sites of `getAgentListDisplayName`)
- The `order` field injected by `reorderAgentsByPriority()` plus JS object insertion order already handle sort ordering
[...]

PR #3291, opened 2026-04-09
fix: remove ZWSP prefixes from agent display paths

### LINKED ISSUE #3238: Breaking change: ZWSP (U+200B) prefix added to agent names breaks prompt_async API consumers

mengfanbo123:
## Summary

Fixes agent name display truncation issue (#3238, #3244, #3246, #3259, #3283) by removing Zero-Width Space (ZWSP) characters from UI display paths.

## Problem

Agent names displayed with missing characters:
- "Sisyphus - Ultraworke" (missing 'r')
- "Sisyphus - ltraworker" (missing 'U')

Root cause: ZWSP prefixes (`\u200B`) used for sorting leaked into OpenCode UI, causing inconsistent terminal rendering.

## Solution

Replace `getAgentListDisplayName()` with `getAgentDisplayName()` in all UI display paths:
- `src/plugin-handlers/agent-config-handler.ts` (lines 163, 166)
- `src/plugin-handlers/agent-key-remapper.ts` (line 9)
- `src/plugin-handlers/agent-priority-order.ts` (lines 4-7)
- `src/plugin-handlers/command-config-handler.ts` (line 102)

Agent ordering is preserved via explicit `order` fields in `agent-priority-order.ts`.

## Testing

- ✅ All related tests updated and passing (159 tests, 351 assertions)
- ✅ TypeScript compilation clean
- ✅ Build successful

## Related Issues

Closes #3238, #3244, #3246, #3259, #3283

---
## Summary by cubic
Removed Zero‑Width Space prefixes from agent display paths to fix truncated names in the UI. Also restored package name and version for upstream PR compatibility.

- **Bug Fixes**
  - Replaced `getAgentListDisplayName()` with `getAgentDisplayName()` across config, command, key remapper, and priority order handlers to strip ZWSP.
  - Preserved Sisyphus → Hephaestus → Prometheus → Atlas ordering via explicit `order` fields.
  - Canonicalized `default_agent` and command `agent` fields to display names without ZWSP; updated tests.

- **Dependencies**
  - Restored `package.json` name and version to upstream values for compatibility.

<sup>Written for commit 8f432ee5ad09005af16526d427b793505ef76518. Summary will update on new commits.</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@mingxy/oh-my-openagent",
-  "version": "3.16.1",
+  "name": "oh-my-opencode",
+  "version": "3.16.0",
   "description": "The Best AI Agent Harness - Batteries-Included OpenCode Plugin with Multi-Model Orchestration, Parallel Background Agents, and Crafted LSP/AST Tools",
   "main": "./dist/index.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "oh-my-opencode",
-  "version": "3.16.0",
+  "name": "@mingxy/oh-my-openagent",
+  "version": "3.16.1",
   "description": "The Best AI Agent Harness - Batteries-Included OpenCode Plugin with Multi-Model Orchestration, Parallel Background Agents, and Crafted LSP/AST Tools",
   "main": "./dist/index.js",
   "types": "dist/index.d.ts",
--- a/src/plugin-handlers/agent-config-handler.test.ts
+++ b/src/plugin-handlers/agent-config-handler.test.ts
@@ -9,11 +9,11 @@ import type { OhMyOpenCodeConfig } from "../config"
 import * as agentLoader from "../features/claude-code-agent-loader"
 import * as skillLoader from "../features/opencode-skill-loader"
 import type { LoadedSkill } from "../features/opencode-skill-loader"
-import { getAgentDisplayName, getAgentListDisplayName } from "../shared/agent-display-names"
+import { getAgentDisplayName } from "../shared/agent-display-names"
 import { applyAgentConfig } from "./agent-config-handler"
 import type { PluginComponents } from "./plugin-components-loader"
 
-const BUILTIN_SISYPHUS_DISPLAY_NAME = getAgentListDisplayName("sisyphus")
+const BUILTIN_SISYPHUS_DISPLAY_NAME = getAgentDisplayName("sisyphus")
 const BUILTIN_SISYPHUS_JUNIOR_DISPLAY_NAME = getAgentDisplayName("sisyphus-junior")
 const BUILTIN_MULTIMODAL_LOOKER_DISPLAY_NAME = getAgentDisplayName("multimodal-looker")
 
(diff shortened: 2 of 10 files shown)


mengfanbo123:
@cubic-dev-ai re-run review please. Package name and version have been restored to original values.

code-yeongyu:
This PR is now superseded by #3329 which permanently fixes the agent ordering issue.

The root cause was reliance on `Object.entries()` insertion order, which is non-deterministic across config merges. The fix introduces:

1. `CANONICAL_CORE_AGENT_ORDER = ["sisyphus", "hephaestus", "prometheus", "atlas"] as const` - single source of truth
2. Alphabetical sorting for non-core agents (deterministic)
3. 12 regression tests including 100-permutation stability test

Approaches like ZWSP prefixes, sort shims, or display name heuristics are now forbidden patterns documented in `src/plugin-handlers/AGENTS.md`.

Thank you for your contribution attempt, but this PR's approach would re-introduce the yo-yo bug.

PR #3322, opened 2026-04-10
fix(runtime-fallback): skip equivalent claude aliases

RaviTharuma:
## Summary
- skip equivalent Claude-family provider aliases during runtime fallback selection
- advance runtime fallback to the next genuinely distinct model instead of looping through equivalent aliases
- update runtime-fallback tests to cover the corrected fallback progression and preserved-agent retry path

## Verification
- bun run typecheck
- bun test src/hooks/runtime-fallback/index.test.ts src/hooks/runtime-fallback/error-classifier.test.ts src/plugin/event.model-fallback.test.ts

---
## Summary by cubic
Skip equivalent Claude-family provider aliases during runtime fallback so retries move to a truly different model. Variants remain distinct, and `-thinking`/`-max`/`-high` are stripped only for Claude models.

- **Bug Fixes**
  - Canonicalize provider/model IDs via parsed model strings; collapse Claude aliases across providers, preserve parsed variants in equivalence, and strip the suffixes only for Claude.
  - Detect and skip equivalent fallbacks; log "Skipping equivalent fallback model" and advance to the next distinct option.
  - Update tests to verify skip-log presence, progression to `openai/gpt-5.4`, and preserved-agent retry behavior.

<sup>Written for commit d9033d73ae371f3c744d1526cf2dd28261bb16fe. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/runtime-fallback/fallback-state.ts
+++ b/src/hooks/runtime-fallback/fallback-state.ts
@@ -41,10 +41,11 @@ function parseCanonicalModel(model: string): { providerID: string; modelID: stri
   if (!parsed?.providerID || !parsed.modelID) return undefined
 
   const canonicalModelID = canonicalizeModelID(parsed.modelID)
+  const variant = parsed.variant?.toLowerCase()
 
   return {
     providerID: canonicalizeProviderFamily(parsed.providerID, parsed.modelID),
-    modelID: canonicalModelID,
+    modelID: variant ? `${canonicalModelID}::${variant}` : canonicalModelID,
   }
 }
 
--- a/src/hooks/runtime-fallback/fallback-state.ts
+++ b/src/hooks/runtime-fallback/fallback-state.ts
@@ -2,6 +2,65 @@ import type { FallbackState, FallbackResult } from "./types"
 import { HOOK_NAME } from "./constants"
 import { log } from "../../shared/logger"
 import type { RuntimeFallbackConfig } from "../../config"
+import { parseModelString } from "../../tools/delegate-task/model-string-parser"
+
+function canonicalizeModelID(modelID: string): string {
+  const loweredModelID = modelID.toLowerCase()
+  const dottedModelID = loweredModelID.replace(/\./g, "-")
+
+  if (
+    dottedModelID.startsWith("claude-opus-") ||
+    dottedModelID.startsWith("claude-sonnet-") ||
+    dottedModelID.startsWith("claude-haiku-")
+  ) {
+    return dottedModelID
+      .replace(/-thinking$/i, "")
+      .replace(/-max$/i, "")
+      .replace(/-high$/i, "")
+  }
+
+  return dottedModelID
+}
+
+function canonicalizeProviderFamily(providerID: string, modelID: string): string {
+  const canonicalModelID = canonicalizeModelID(modelID)
+
+  if (
+    canonicalModelID.startsWith("claude-opus-") ||
+    canonicalModelID.startsWith("claude-sonnet-") ||
+    canonicalModelID.startsWith("claude-haiku-")
+  ) {
+    return "anthropic-compatible-claude"
+  }
+
+  return providerID.toLowerCase()
+}
+
+function parseCanonicalModel(model: string): { providerID: string; modelID: string } | undefined {
+  const parsed = parseModelString(model)
+  if (!parsed?.providerID || !parsed.modelID) return undefined
+
+  const canonicalModelID = canonicalizeModelID(parsed.modelID)
+
+  return {
+    providerID: canonicalizeProviderFamily(parsed.providerID, parsed.modelID),
+    modelID: canonicalModelID,
+  }
+}
+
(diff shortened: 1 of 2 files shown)


RaviTharuma:
Re: https://github.com/code-yeongyu/oh-my-openagent/pull/3322#discussion_r3066938297

Good catch — this was valid.

I pushed a follow-up fix in `edb2d7f2` that narrows the equivalence collapsing so suffix stripping is now **Claude-family only** (`claude-opus-*`, `claude-sonnet-*`, `claude-haiku-*`).

That preserves the intended runtime-fallback alias skip for equivalent Claude provider aliases, while keeping explicit non-Claude suffixes like `-high`, `-max`, and `-thinking` distinct fallback targets.

Re-verified after the change with:
- `bun run typecheck`
- `bun test src/hooks/runtime-fallback/index.test.ts src/hooks/runtime-fallback/error-classifier.test.ts src/plugin/event.model-fallback.test.ts`

So the original Claude alias loop fix remains, but the broader non-Claude regression is no longer present.

RaviTharuma:
Re: https://github.com/code-yeongyu/oh-my-openagent/pull/3322#discussion_r3067960850 and https://github.com/code-yeongyu/oh-my-openagent/pull/3322#discussion_r3067960852

I tightened this one further in `669afe22`.

Final behavior now is:
- **provider identity remains preserved** in runtime-fallback equivalence
- **variant is part of equivalence** (`foo` vs `foo(high)`/`foo high` no longer collapse)
- the earlier follow-up in `edb2d7f2` still keeps the non-Claude suffixes (`-high`, `-max`, `-thinking`) distinct

Re-verified after the latest change with:
- `bun run typecheck`
- `bun test src/hooks/runtime-fallback/index.test.ts src/hooks/runtime-fallback/error-classifier.test.ts src/plugin/event.model-fallback.test.ts`

So the current PR state now addresses:
1. non-Claude suffixes staying distinct
2. variant-only fallback hops staying distinct
while preserving the already-tested runtime fallback behavior in the focused suite.

RaviTharuma:
Re: https://github.com/code-yeongyu/oh-my-openagent/pull/3322#discussion_r3067985616 and https://github.com/code-yeongyu/oh-my-openagent/pull/3322#discussion_r3067985619

I rechecked the current branch state against the intended runtime behavior.

The final implementation on the PR branch keeps:
- **provider identity preserved** in equivalence checks
- **variant as part of equivalence**
- and only the original, narrower non-Claude suffix fix from `edb2d7f2`

That means the two latest concerns are now addressed in the current branch state:
1. provider-specific Claude fallbacks are no longer collapsed together
2. variant-only fallback hops remain distinct

I also re-ran:
- `bun run typecheck`
- `bun test src/hooks/runtime-fallback/index.test.ts src/hooks/runtime-fallback/error-classifier.test.ts src/plugin/event.model-fallback.test.ts`

with the current local branch state before re-requesting review.

code-yeongyu:
Triage note: rebase onto current `dev` conflicts in `src/hooks/runtime-fallback/fallback-state.ts` and `index.test.ts` because of substantial runtime-fallback evolution after this PR opened — #3982 (category fallback honoring), #4007 (preserve attemptCount on internal abort), #3934 (localized quota classifier), #3773 (HTTP statusCode for GLM).

Idea still valid: skipping equivalent Claude aliases prevents the chain from looping through the same model under a different provider alias. To re-land:
1. Rebase onto current dev (HookDeps + types may have shifted).
2. Re-verify behavior plays with the now-richer fallback selection (variants preserved, statusCode triggers, localized quota classifier, preserved attemptCount).
3. Confirm `bun test --timeout 30000` is 0 fail.

code-yeongyu:
Update on triage: tried the rebase against current dev. Single non-test conflict in `src/hooks/runtime-fallback/fallback-state.ts` is trivially resolved (replace HEAD's strict-equality skip with this PR's `isEquivalentModel` skip — strictly broader). The blocker is in `src/hooks/runtime-fallback/index.test.ts` where the PR adds a new fallback model to the chain (`openai/gpt-5.4`) AND HEAD has model-version-bumped a number of cases from `claude-opus-4.6` -> `claude-opus-4.7`.

Per-conflict-block decision required:
1. Keep HEAD's model version (`claude-opus-4.7`).
2. Apply PR's structural assertion (skip equivalent + asserts the `Skipping equivalent fallback model` log).
3. Add the extra non-equivalent model (`openai/gpt-5.4`) to the agent-fallback list so the resolver advances past the skip.
4. Repeat across ~6 conflict blocks in the same file.

This is mechanically straightforward but each block needs its specific final assertion re-derived (which non-equivalent fallback the resolver should land on). Skipping that careful per-block work would risk a #3825-style revert.

To re-land:
1. Apply the trivial `fallback-state.ts` resolution as above.
2. Per test block, keep current model versions, layer the equivalence-skip assertion, ensure the chain includes a non-equivalent fallback so the test still expects a defined next-model.
3. Confirm `bun test --timeout 30000` is 0 fail.

Leaving open.

PR #3339, opened 2026-04-11
fix(shell-env): check SHELL before PSModulePath for shell detection

### LINKED ISSUE #3338: Bug: PowerShell env var syntax ($env:VAR) used in bash shell on Windows
### Description
On Windows, when the OpenCode shell is bash (Git Bash), oh-my-openagent prepends PowerShell-style environment variable syntax (`$env:CI='true'`) instead of bash syntax (`CI=true`). This causes all env var assignments to fail with errors like:

```
/usr/bin/bash: line 1: :CI=true: command not found
```

### Reproduction
1. Windows 11 with Git Bash installed
2. OpenCode configured to use bash shell
3. oh-my-openagent plugin loaded
4. Run any command — the prefixed env vars fail but the actual command still executes

### Expected Behavior
Detect the actual shell type (bash vs PowerShell) and use the correct syntax:
- **Bash**: `CI=true; DEBIAN_FRONTEND=noninteractive; ...`
- **PowerShell**: `$env:CI='true'; $env:DEBIAN_FRONTEND='noninteractive'; ...`

### Actual Output
```bash
$ $env:CI='true'; $env:DEBIAN_FRONTEND='noninteractive'; $env:GIT_TERMINAL_PROMPT='0'; ...; git log --oneline -5
/usr/bin/bash: line 1: :CI=true: command not found
/usr/bin/bash: line 1: :DEBIAN_FRONTEND=noninteractive: command not found
... (15 more errors)
f7b3d08b timed fix
438bede6 more
```

### Impact
- The actual command still runs, but all safety env vars are NOT set (CI mode, non-interactive, no pager, etc.)
- This means git and other tools may prompt interactively, breaking agent automation
- Clutters output with 15+ error lines

### Suggested Fix
Check `process.env.SHELL` or `process.env.COMSPEC` to determine the actual shell:
- If `SHELL` contains `bash` or `sh` → use bash syntax
- If `SHELL` is undefined on Windows and `COMSPEC` contains `cmd` → use cmd syntax
- Default to PowerShell only when explicitly detected

### Environment
- OS: Windows 11
- Shell: Git Bash (via OpenCode)
- oh-my-openagent: 3.16.0
- OpenCode: latest

Qiiks:
## Summary

- Reordered `detectShellType()` to check `SHELL` env var **before** `PSModulePath`
- On Windows, `PSModulePath` is always set by the system, causing the function to incorrectly return `"powershell"` even when the user has `SHELL` pointing to Git Bash or another Unix-like shell
- An explicit `SHELL` env var indicates the user's chosen shell and should take priority over the implicit `PSModulePath` heuristic

## Problem

When running OpenCode with oh-my-openagent on Windows with Git Bash (SHELL set), the plugin generated PowerShell env var syntax (`$env:VAR='value'`) instead of Unix syntax (`export VAR=value;`). This caused commands to fail silently or with syntax errors in bash shells.

Root cause: `detectShellType()` checked `PSModulePath` first. Since Windows always has this env var, it always returned `"powershell"` regardless of the user's actual shell preference.

## Fix

```typescript
// Before (buggy): PSModulePath checked first → always "powershell" on Windows
if (process.env.PSModulePath) return "powershell"
if (process.env.SHELL) ...

// After (fixed): SHELL checked first → respects user's explicit shell choice
if (process.env.SHELL) ...
if (process.env.PSModulePath) return "powershell"
```

## Testing

- Updated existing test: "PSModulePath takes priority" → "SHELL takes priority over PSModulePath → returns unix"
- Added new test: Git Bash path on Windows with PSModulePath should return "unix"
- Fixed PSModulePath-only test to explicitly `delete process.env.SHELL` (was leaking from system env)
- All 42 tests pass

Closes #3338

---
## Summary by cubic
Switches shell detection to check SHELL before PSModulePath so Windows users with Git Bash aren’t misidentified as PowerShell. This fixes incorrect env var syntax and bash errors. Fixes #3338.

- Bug Fixes
  - Prioritize SHELL over PSModulePath; fallbacks stay the same (win32 → cmd, others → unix).
  - Treat Git Bash paths (e.g., C:\Program Files\Git\bin\bash.exe) as unix.
  - Updated tests for precedence and Git Bash; clear SHELL in PSModulePath-only test.

<sup>Written for commit f5a95203a46a32968c1107571cd4b2061b49c6e9. Summary will update on new commits.</sup>

### DIFF
--- a/src/shared/shell-env.test.ts
+++ b/src/shared/shell-env.test.ts
@@ -45,7 +45,8 @@ describe("shell-env", () => {
       expect(result).toBe("unix")
     })
 
-    test("#given PSModulePath is set #when detectShellType is called #then returns powershell", () => {
+    test("#given PSModulePath is set without SHELL #when detectShellType is called #then returns powershell", () => {
+      delete process.env.SHELL
       process.env.PSModulePath = "C:\\Program Files\\PowerShell\\Modules"
       Object.defineProperty(process, "platform", { value: "win32" })
 
@@ -74,14 +75,24 @@ describe("shell-env", () => {
       expect(result).toBe("unix")
     })
 
-    test("#given PSModulePath takes priority over SHELL #when both are set #then returns powershell", () => {
+    test("#given SHELL takes priority over PSModulePath #when both are set #then returns unix", () => {
       process.env.PSModulePath = "C:\\Program Files\\PowerShell\\Modules"
       process.env.SHELL = "/bin/bash"
       Object.defineProperty(process, "platform", { value: "win32" })
 
       const result = detectShellType()
 
-      expect(result).toBe("powershell")
+      expect(result).toBe("unix")
+    })
+
+    test("#given SHELL set to Git Bash on Windows with PSModulePath #when detectShellType is called #then returns unix", () => {
+      process.env.PSModulePath = "C:\\Program Files\\PowerShell\\Modules"
+      process.env.SHELL = "C:\\Program Files\\Git\\bin\\bash.exe"
+      Object.defineProperty(process, "platform", { value: "win32" })
+
+      const result = detectShellType()
+
+      expect(result).toBe("unix")
     })
   })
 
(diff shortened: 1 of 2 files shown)


Qiiks:

PR #3353, opened 2026-04-11
fix: align GitHub Copilot model variants with catalog

divlook:
## Summary
- align GitHub Copilot fallback variants with the current `opencode models github-copilot --verbose` output by splitting Copilot-specific Opus and Gemini entries from shared fallback chains
- add regression coverage for requirements, compatibility, runtime capability normalization, model-fallback flows, and plugin-entry isolation
- verify the change with focused model fallback tests, `bun run typecheck`, and `bun run build`

## Changes
- set GitHub Copilot `claude-opus-4.7` fallback entries to `variant: medium` because the current `opencode models github-copilot --verbose` output exposes only the `medium` variant for Opus 4.7
- keep non-Copilot `claude-opus-4-7` fallback entries on `variant: max`, since the medium-only restriction is specific to GitHub Copilot
- remove explicit variant from GitHub Copilot `gemini-3.1-pro-preview` fallback entries because the current `opencode models github-copilot --verbose` output does not expose a Gemini variant set there
- update fallback, anthropic-effort, think-mode, snapshot, and regression expectations for the corrected Copilot Opus behavior
- isolate the auto-update plugin-entry test from local user config leakage

## Testing
- `opencode models github-copilot --verbose`
- `bun test src/cli/model-fallback.test.ts src/shared/model-requirements.test.ts src/hooks/model-fallback/hook.test.ts src/plugin/event.model-fallback.test.ts src/hooks/anthropic-effort/index.test.ts src/hooks/think-mode/index.test.ts`
- `bun run typecheck`
- `bun run build`

## Notes
- target branch: `dev`
- communication for review should stay in English
- please ensure CLA requirements are satisfied before review/merge


divlook:

divlook:
recheck

divlook:
Follow-up cleanup for stale GitHub Copilot nano references is tracked separately in #3359. This PR intentionally keeps that work out of scope and only covers the Copilot variant alignment changes.

divlook:
Validation note: the expected GitHub Copilot variant behavior here is version-qualified.

The local CLI check used during validation was based on opencode 1.4.0 with the command "opencode models github-copilot --verbose", which reported claude-opus-4.7 with low / medium / high variants.

A different OpenCode project/version line can expose a different catalog basis. For example, anomalyco/opencode v1.4.11 also exposes a "models [provider] --verbose" command:
https://github.com/anomalyco/opencode/blob/a35b8a95c27d28e979a3826e1289d7ee87f40251/packages/opencode/src/cli/cmd/models.ts#L12-L30

Its provider transform also contains provider-specific handling for GitHub Copilot Claude Opus 4.7:
https://github.com/anomalyco/opencode/blob/a35b8a95c27d28e979a3826e1289d7ee87f40251/packages/opencode/src/provider/transform.ts#L392-L398
https://github.com/anomalyco/opencode/blob/a35b8a95c27d28e979a3826e1289d7ee87f40251/packages/opencode/src/provider/transform.ts#L426-L432

So catalog comparisons here should always name the exact OpenCode project/version they came from. This PR was validated against the local opencode 1.4.0 catalog behavior, not a generic cross-version catalog assumption.

Issue #3220, opened 2026-04-08
[Bug]: once opencode upgrade to 1.4.0, oh-my-opencode disappears


haohaotian1011:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

all agents cannot be seen.

### Steps to Reproduce

upgrade opencode to 1.4.0
then all agetns disappear.

### Expected Behavior

should be still there.

### Actual Behavior

cannot be seen 

### Doctor Output

```shell
oMoMoMoMo... Doctor 


Installation
────────────────────────────────────────
  ✓ OpenCode Installation → 1.4.0
  ✓ Plugin Registration → Registered

Configuration
────────────────────────────────────────
  ✓ Configuration Validity → Valid JSON config

Authentication
────────────────────────────────────────
  ✓ Anthropic (Claude) Auth → Auth plugin available
  ○ OpenAI (ChatGPT) Auth → Auth plugin not installed
  ○ Google (Gemini) Auth → Auth plugin not installed

Dependencies
────────────────────────────────────────
  ✓ AST-Grep CLI → installed
  ⚠ AST-Grep NAPI → Not installed (optional)
  ⚠ Comment Checker → Not installed (optional)

Tools & Servers
────────────────────────────────────────
  ⚠ GitHub CLI → 2.89.0 - not authenticated
  ✓ LSP Servers → 1/4 servers available
  ✓ Built-in MCP Servers → 2 built-in servers enabled
  ○ User MCP Configuration → No user MCP configuration found

Updates
────────────────────────────────────────
  ✓ Version Status → Up to date (3.15.3)

Summary
────────────────────────────────────────

  8 passed, 0 failed, 3 warnings, 3 skipped
  Total: 14 checks in 2323ms

⚠ All systems operational with warnings.
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

ubuntu22.04

### OpenCode Version

1.4.0

code-yeongyu:
[...]

PR #3365, opened 2026-04-12
Fix local plugin registration for OpenCode 1.4

code-yeongyu:
## Summary
- switch OmO plugin registration to a local `file://` plugin entry when the installed package root is available
- treat local plugin path entries as valid OmO installs by resolving their `package.json` instead of relying on path substring matching
- keep local-path installs idempotent so rerunning install does not duplicate the plugin entry
- keep the existing ZWSP-based agent ordering path untouched

## Why
OpenCode 1.4.x users were seeing OmO reported as registered while the agents disappeared or the default OmO agent could not be found. The field reports around #3220 pointed to the package-spec plugin loading path being unreliable, so this change moves OmO installs to a local `file://` plugin entry and updates detection/doctor logic to understand that form.

## Testing
- `bun test src/cli/config-manager/plugin-detection.test.ts src/cli/doctor/checks/system-plugin.test.ts src/cli/doctor/checks/system.test.ts`
- `bun run typecheck`

## User impact
Existing broken installs will need to rerun `oh-my-opencode install` once so their `opencode.json` plugin entry is migrated to the new local-path form.

---
## Summary by cubic
Fixes broken plugin registration in OpenCode 1.4.x by switching to a local `file://` plugin entry and updating detection/doctor logic so local paths are treated as valid installs. Addresses #3220 where OmO appeared registered but agents were missing.

- **Bug Fixes**
  - Prefer a local `file://` plugin entry when the package root is found.
  - Detect local path entries by resolving their `package.json` (`isAcceptedLocalPluginEntry`).
  - Avoid duplicate entries when re-installing (idempotent).
  - Doctor check now recognizes local-path installs as registered local-dev.

- **Migration**
  - Users on 1.4.x should rerun `oh-my-opencode install` to migrate the plugin entry in `opencode.json` to the new local-path form.

<sup>Written for commit 4b04c3c8c02aa02256666902499fdc2d207ad299. Summary will update on new commits.</sup>


code-yeongyu:
Closing this because the proposed fix went in the wrong direction. The requirement is to keep normal plugin-package installation working, not to route users to a local file:// plugin path.

PR #3420, opened 2026-04-14
fix(auto-update): use semantic version comparison instead of string e…

grandmaster451:
…quality

The update checker was using strict string comparison (===) to compare current and latest versions. This caused false-positive update notifications when versions were semantically equal but had different string representations (e.g., with/without 'v' prefix, different formatting).

Changed to use compareVersions() which properly parses and compares semver versions, handling 'v' prefixes and other variations correctly.

Fixes: incorrect 'v3.17.2 available' notifications when already on latest version

## Summary

- 

## Changes

- 

## Screenshots

| Before | After |
|:---:|:---:|
|  |  |

## Testing

```bash
bun run typecheck
bun test
```

## Related Issues

---
## Summary by cubic
Fix incorrect update prompts by switching the auto-update checker to semantic version comparison. Replaces string equality with `compareVersions`, correctly handling 'v' prefixes and other semver variations.

<sup>Written for commit 003e45abeda78c77c0c90eaf10d866d2e9e06283. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/auto-update-checker/checker/check-for-update.ts
+++ b/src/hooks/auto-update-checker/checker/check-for-update.ts
@@ -1,4 +1,5 @@
 import { log } from "../../../shared/logger"
+import { compareVersions } from "../../../shared/opencode-version"
 import type { UpdateCheckResult } from "../types"
 import { extractChannel } from "../version-channel"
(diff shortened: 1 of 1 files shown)


grandmaster451:

PR #3464, opened 2026-04-15
fix(context-limit): use known GA context limits for claude-4-6 models (fixes #3450)

### LINKED ISSUE #3450: [Bug]: resolveActualContextLimit falls back to 200k when anthropic provider has no explicit model limits in config
### Prerequisites

### Bug Description

The fix from #2836 (commit `719a5827`) added a `modelContextLimitsCache` check for GA 4.6 models in `resolveActualContextLimit`. However, this fix depends on the cache being populated — which only happens when the `anthropic` provider has explicit `models.*.limit.context` entries in the user's provider config.

When the anthropic provider exists (built-in or via OpenCode) but does **not** define explicit model limits in the config, `modelContextLimitsCache` has no entry for `anthropic/claude-opus-4-6` or `anthropic/claude-sonnet-4-6`. The function then falls through to `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200_000`.

### Root Cause

```typescript
// src/shared/context-limit-resolver.ts
export function resolveActualContextLimit(providerID, modelID, modelCacheState) {
  if (isAnthropicProvider(providerID)) {
    const explicit1M = getAnthropicActualLimit(modelCacheState);
    if (explicit1M === 1_000_000) return explicit1M;  // only if env var or beta header

    const cachedLimit = modelCacheState?.modelContextLimitsCache?.get(`${providerID}/${modelID}`);
    if (cachedLimit && supportsCachedAnthropicLimit(modelID)) return cachedLimit;
    // ↑ cachedLimit is undefined when no explicit limit in provider config

    return DEFAULT_ANTHROPIC_ACTUAL_LIMIT;  // ← 200k fallback
  }
}
```

`supportsCachedAnthropicLimit` already identifies opus-4-6 and sonnet-4-6 as 1M-capable models, but when `cachedLimit` is falsy, this knowledge is unused.

### Suggested Fix

Add a fallback to 1M for models matched by `supportsCachedAnthropicLimit` when no cached limit exists:

```typescript
if (cachedLimit && supportsCachedAnthropicLimit(modelID)) return cachedLimit;

// Models known to support 1M should not fall back to 200k
if (supportsCachedAnthropicLimit(modelID)) return 1_000_000;

return DEFAULT_ANTHROPIC_ACTUAL_LIMIT;
```

### Impact

All three call sites of `resolveActualContextLimit` are affected:
- **context-window-monitor**: reports "200,000-token context window" instead of 1M
- **dynamicTruncate**: over-truncates tool output based on wrong budget
- **preemptive-compaction**: triggers unnecessary compaction around 156k tokens

### Workaround

Set `ANTHROPIC_1M_CONTEXT=true` in `.env`.

### Related

- #2836 — original fix (closed, but this edge case remains)
- Commit `719a5827` — the fix that introduced `supportsCachedAnthropicLimit`

### Environment

- oh-my-openagent: 3.17.2
- OpenCode: 1.4.x
- OS: Linux

MoerAI:
## Summary
- Add built-in 680K context limit for GA claude-opus-4-6 and claude-sonnet-4-6 instead of falling back to 200K when the model cache is empty

## Problem
When the Anthropic provider config lacks explicit `models.*.limit.context` entries, the `modelContextLimitsCache` has no entry for GA 4.6 models. `resolveActualContextLimit` then falls through to `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200_000`, which is far below the actual 680K GA limit. This causes premature compaction and context truncation.

## Fix
Added a `KNOWN_GA_CONTEXT_LIMITS` lookup that returns 680K for `claude-opus-4-6` and `claude-sonnet-4-6` (including `-high` variants). When the model cache has no entry and the 1M beta is not active, the function now checks known GA limits before falling back to the 200K default.

## Changes
| File | Change |
|------|--------|
| `src/shared/context-limit-resolver.ts` | Add `KNOWN_GA_CONTEXT_LIMITS` map and `getKnownGAContextLimit()` helper; use as intermediate fallback before 200K default |

Fixes #3450

---
## Summary by cubic
Use the known 680K context limit for GA `claude-opus-4-6` and `claude-sonnet-4-6` (including `-high`) instead of falling back to 200K when cache/config are missing. Prevents early context truncation and aligns with GA limits. Fixes #3450.

- **Bug Fixes**
  - Added a `KNOWN_GA_CONTEXT_LIMITS` lookup and helper to return 680K for GA 4.6 models before the 200K default.
  - Applies when no `models.*.limit.context` is set and the Anthropic model cache has no entry.

<sup>Written for commit 393fd7fe70f467f8598b9edf76b9ae27d2192f5d. Summary will update on new commits.</sup>

### DIFF
--- a/src/shared/context-limit-resolver.ts
+++ b/src/shared/context-limit-resolver.ts
@@ -23,6 +23,16 @@ function supportsCachedAnthropicLimit(modelID: string): boolean {
   return /^claude-(opus|sonnet)-4(?:-|\.)6(?:-high)?$/.test(modelID)
 }
 
+const KNOWN_GA_CONTEXT_LIMITS: Record<string, number> = {
+  "claude-opus-4-6": 680_000,
+  "claude-sonnet-4-6": 680_000,
+}
+
+function getKnownGAContextLimit(modelID: string): number | undefined {
+  const normalized = modelID.replace(/-high$/, "")
+  return KNOWN_GA_CONTEXT_LIMITS[normalized]
+}
+
 export function resolveActualContextLimit(
   providerID: string,
   modelID: string,
@@ -35,7 +45,7 @@ export function resolveActualContextLimit(
     const cachedLimit = modelCacheState?.modelContextLimitsCache?.get(`${providerID}/${modelID}`)
     if (cachedLimit && supportsCachedAnthropicLimit(modelID)) return cachedLimit
 
-    return DEFAULT_ANTHROPIC_ACTUAL_LIMIT
+    return getKnownGAContextLimit(modelID) ?? DEFAULT_ANTHROPIC_ACTUAL_LIMIT
   }
 
   return modelCacheState?.modelContextLimitsCache?.get(`${providerID}/${modelID}`) ?? null


MoerAI:

MoerAI:
Closing this PR after reviewing the test failures it introduced.

After further investigation, the 200K default in `resolveActualContextLimit` is **intentional**, not a bug. The existing tests in `context-limit-resolver.test.ts` and `preemptive-compaction.test.ts` explicitly validate that claude-sonnet-4-6 and claude-opus-4-6 default to 200K when:
- No explicit `modelContextLimitsCache` entry exists
- `anthropicContext1MEnabled` is false
- Neither `ANTHROPIC_1M_CONTEXT` nor `VERTEX_ANTHROPIC_1M_CONTEXT` env var is set

This matches Anthropic's actual tiering: 4.6 models have 200K context by default, with 1M available as an opt-in beta. My change assumed a 680K fallback for GA 4.6 models, which contradicts this design and broke 7 preemptive-compaction tests that depend on the 200K threshold.

The proper fix for #3450 (if one is needed) would be ensuring OpenCode's built-in Anthropic provider config populates the `modelContextLimitsCache` with the user's actual tier limit (e.g., 1M if they have beta access), rather than changing the conservative 200K default.

PR #3470, opened 2026-04-16
fix(session-recovery): add thinking_block_modified error detection and recovery

omer-koren:
## Problem
- Amazon Bedrock can return `thinking` / `redacted_thinking` signature errors after compaction corrupts the latest assistant thinking block signatures.
- Session recovery already handles thinking block ordering and thinking-disabled violations, but it did not recognize this corruption case.
- Users with `experimental.auto_resume: true` were left without an automatic recovery path when the provider rejected the modified thinking blocks.

## Solution
- add a `thinking_block_modified` recovery error type and detect the Bedrock `thinking blocks cannot be modified` failure pattern
- wire the new error type into the session recovery hook with dedicated toast messaging
- reuse `recoverThinkingDisabledViolation` to strip corrupted thinking blocks, then auto-resume the session when enabled
- add regression tests for the full Bedrock error message and a simpler variant

## Testing
- `bun run typecheck`
- `bun test src/hooks/session-recovery/`
- `bun run build`

---
## Summary by cubic
Detects and recovers from Bedrock “thinking block modified” errors caused by compaction, restoring auto-resume when enabled and preventing stuck sessions.

- **Bug Fixes**
  - Added `thinking_block_modified` recovery type; detects Bedrock “thinking … cannot be modified” errors (full and simple variants).
  - Integrated with session recovery: shows dedicated toast, strips corrupted thinking blocks using existing logic, and auto-resumes when experimental.auto_resume is enabled; added regression tests.

<sup>Written for commit b6ad494f1fe7ec2b0ef17090f6bed29741924cbb. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/session-recovery/detect-error-type.test.ts
+++ b/src/hooks/session-recovery/detect-error-type.test.ts
@@ -36,6 +36,31 @@ describe("detectErrorType", () => {
     expect(result).toBe("thinking_disabled_violation")
   })
 
+  it("#given a Bedrock thinking block modified error #when detecting #then returns thinking_block_modified", () => {
+    //#given
+    const error = {
+      message:
+        "undefined: The model returned the following errors: messages.17.content.28: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.",
+    }
+
+    //#when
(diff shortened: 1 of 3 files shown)


omer-koren:

PR #3529, opened 2026-04-20
fix(model-resolution): ensure subagents use free models when no paid providers configured (fixes #1883)

### LINKED ISSUE #1883: Bug: Should use only free models when no providers configured
## Bug Description

When oh-my-opencode is installed/configured and the user selects no providers, subagents should ONLY use free models. Currently, they incorrectly inherit or use a paid model (glm-4.7) instead of defaulting to free models only.

## Expected Behavior

When no providers are configured:
- Subagents should use ONLY free models (e.g., `opencode/big-pickle`, `opencode/kimi-k2.5-free`, `opencode/minimax-m2.5-free`)
- Should NOT inherit the parent model's configuration
- Should default to free models since no paid providers were set up

## Actual Behavior

- User configures: no providers selected (free models only)
- Main agent may work correctly with a free model
- Subagent (child agent) uses `glm-4.7` (a paid model), causing errors since GLM 4.7 is no longer free
- Subagent should instead use only free models when no providers are configured

## Steps to Reproduce

1. Configure oh-my-opencode with no providers (free models only)
2. Set the main model to a free model like `opencode/kimi-k2.5-free`
3. Spawn a subagent (task with category or subagent_type)
4. Observe that the subagent attempts to use `glm-4.7` instead of a free model

## Environment

- oh-my-opencode version: latest
- Configuration: No providers selected (free models only)

## Additional Context

Reported in: CloudWaddie/actions-agent#67

The fix should ensure that when no providers are configured, the system defaults to using ONLY free models, rather than inheriting any parent model configuration or falling back to a paid model.

MoerAI:
## Summary
- keep delegate-task subagents on free `opencode/*` models when only free providers are configured

## Problem
Subagent model resolution could still select paid `opencode/*` entries from hardcoded category and agent fallback chains when the user had no paid providers configured. This let child agents resolve to models like `gpt-5.4` instead of staying on free models, especially when the connected-provider cache only exposed `opencode`.

## Fix
Free-only provider setups now strip paid defaults out of delegate-task category and fallback resolution and replace them with a stable free-model chain. The parsing helpers were extracted into focused modules, and targeted regression tests now cover both cached and cold-cache free-only resolution paths.

## Changes
| File | Change |
|------|--------|
| `src/tools/delegate-task/model-selection.ts` | Apply free-only fallback logic during delegate-task model resolution |
| `src/tools/delegate-task/free-model-fallback.ts` | Add free-only provider detection and fallback chain helpers |
| `src/tools/delegate-task/model-selection-input.ts` | Extract model parsing helpers from the resolver |
| `src/tools/delegate-task/model-selection.free-model-fallback.test.ts` | Add regressions for free-only subagent fallback behavior |

Fixes #1883

---
## Summary by cubic
Keeps delegate-task subagents on free `opencode/*` models when only free providers are connected by stripping paid defaults/fallbacks and using an updated, non-deprecated free chain.

- **Bug Fixes**
  - Filter paid entries from category defaults and fallbacks in free-only setups; apply free chain: `big-pickle` → `minimax-m2.5-free` → `hy3-preview-free` → `nemotron-3-super-free` → `gpt-5-nano`.
  - Preserve explicit user-configured category models.
  - Add regression tests to ensure the chain excludes deprecated models and only contains known free models, and to verify free-only fallback behavior.

- **Refactors**
  - Extract free-only logic to `src/tools/delegate-task/free-model-fallback.ts`.
  - Move model parsing helpers to `src/tools/delegate-task/model-selection-input.ts`.

<sup>Written for commit 49a1540afe0c6f2adf41710c74edd6dd72c68ee2. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/3529?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/tools/delegate-task/free-model-fallback.test.ts
+++ b/src/tools/delegate-task/free-model-fallback.test.ts
@@ -1,0 +1,22 @@
+/// <reference types="bun-types" />
+
+import { describe, expect, test } from "bun:test"
+import { FREE_ONLY_FALLBACK_CHAIN, isKnownFreeModel } from "./free-model-fallback"
+
+describe("FREE_ONLY_FALLBACK_CHAIN", () => {
+  // Deprecated by opencode (https://opencode.ai/zen/v1/models, models.dev catalog).
+  // If any of these reappear, the chain ships a model that no longer resolves.
+  const deprecatedFreeModelIds = ["kimi-k2.5-free", "kimi-k2-free", "kimi-k2-thinking-free"]
+
+  test("does not contain models that opencode has marked deprecated", () => {
+    for (const entry of FREE_ONLY_FALLBACK_CHAIN) {
+      expect(deprecatedFreeModelIds).not.toContain(entry.model)
+    }
+  })
+
+  test("only contains entries that isKnownFreeModel recognizes", () => {
+    for (const entry of FREE_ONLY_FALLBACK_CHAIN) {
+      expect(isKnownFreeModel(entry.model)).toBe(true)
+    }
+  })
+})
--- a/src/tools/delegate-task/free-model-fallback.ts
+++ b/src/tools/delegate-task/free-model-fallback.ts
@@ -1,0 +1,61 @@
+import type { FallbackEntry } from "../../shared/model-requirements"
+
+const FREE_ONLY_PROVIDER_IDS = new Set(["opencode"])
+const KNOWN_FREE_MODEL_IDS = new Set([
+  "big-pickle",
+  "gpt-5-nano",
+  "kimi-k2.5-free",
+  "minimax-m2.5-free",
+])
+
+export const FREE_ONLY_FALLBACK_CHAIN: FallbackEntry[] = [
+  { providers: ["opencode"], model: "big-pickle" },
+  { providers: ["opencode"], model: "kimi-k2.5-free" },
+  { providers: ["opencode"], model: "minimax-m2.5-free" },
+  { providers: ["opencode"], model: "gpt-5-nano" },
+]
+
+function getModelId(model: string): string {
+  return model.includes("/") ? model.split("/").slice(1).join("/") : model
+}
+
+export function isKnownFreeModel(model: string): boolean {
+  return KNOWN_FREE_MODEL_IDS.has(getModelId(model))
+}
+
+export function isFreeOnlyProviderConfiguration(connectedProviders: string[] | null): boolean {
+  return connectedProviders !== null
+    && connectedProviders.length > 0
+    && connectedProviders.every((provider) => FREE_ONLY_PROVIDER_IDS.has(provider))
+}
+
+export function getFreeOnlyCategoryDefaultModel(input: {
+  categoryDefaultModel?: string
+  isUserConfiguredCategoryModel?: boolean
+  freeOnlyProviderConfiguration: boolean
+}): string | undefined {
+  if (!input.freeOnlyProviderConfiguration || input.isUserConfiguredCategoryModel) {
+    return input.categoryDefaultModel
+  }
+
+  if (!input.categoryDefaultModel || !isKnownFreeModel(input.categoryDefaultModel)) {
+    return undefined
+  }
+
+  return input.categoryDefaultModel
+}
+
+export function getFallbackChainForFreeOnlyProviders(
+  fallbackChain: FallbackEntry[] | undefined,
+  freeOnlyProviderConfiguration: boolean,
+): FallbackEntry[] | undefined {
+  if (!freeOnlyProviderConfiguration || !fallbackChain || fallbackChain.length === 0) {
+    return fallbackChain
+  }
+
+  const freeEntries = fallbackChain.filter((entry) =>
+    entry.providers.includes("opencode") && isKnownFreeModel(entry.model),
+  )
+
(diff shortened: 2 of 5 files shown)


MoerAI:

mrosnerr:
hi - thanks for taking this on!

immediate thing: the new `KNOWN_FREE_MODEL_IDS` includes `kimi-k2.5-free`, which isn't on opencode anymore.

```
$ curl -s https://opencode.ai/zen/v1/models | jq -r '.data[].id' | grep kimi-k2.5-free
(empty)

$ curl -s https://models.dev/api.json | jq -r '.opencode.models[] | select(.id == "kimi-k2.5-free") | .status'
deprecated
```

so the new hardcoded list is shipping the same shape of artifact that broke the original issue.

separate thing: `FREE_ONLY_PROVIDER_IDS = ["opencode"]` reads "only opencode connected" as "free-only setup", but per cbrunnkvist's repro in [#1883](https://github.com/code-yeongyu/oh-my-openagent/issues/1883#issuecomment-3911833614), both "no auth at all" and "authenticated with zen api key" produce `connectedProviders === ["opencode"]`. so paying subscribers would get force-downgraded by this. `availableModels` is the input that already knows which models are actually reachable.

+1 on both of cubic's findings.

bigger picture - hardcoded model names anywhere in source go stale on opencode's rotation cadence. wommy's jq from the original thread gives the current canonical set:

```
$ curl -s https://models.dev/api.json | jq -r '[.opencode.models[] | select(.status != "deprecated" and .cost.input == 0) | .id]'
[
  "big-pickle",
  "gpt-5-nano",
  "hy3-preview-free",
  "minimax-m2.5-free",
  "nemotron-3-super-free"
]
```

two of those (`hy3-preview-free`, `nemotron-3-super-free`) are live free models this PR's chain can't reach. if the chain came from a json snapshot rather than source literals, this drift becomes a json edit instead of a release. happy to help shape a snapshot + refresh-workflow if useful.

MoerAI:
@mrosnerr thanks for the careful read.

**Concern 1 — `kimi-k2.5-free` is deprecated**

Confirmed against both `https://opencode.ai/zen/v1/models` and the models.dev catalog (where it now reports `status: "deprecated"`). Just pushed [49a1540](https://github.com/code-yeongyu/oh-my-opencode/commit/49a1540a) which:

- removes `kimi-k2.5-free` from `KNOWN_FREE_MODEL_IDS` and `FREE_ONLY_FALLBACK_CHAIN`
- adds the currently live free opencode models (`hy3-preview-free`, `nemotron-3-super-free`) so the chain doesn't lose coverage
- adds a regression test that asserts the chain never contains a known-deprecated model and that every entry passes `isKnownFreeModel`, so future drift fails CI instead of silently shipping a stale model

After fix the chain is: `big-pickle -> minimax-m2.5-free -> hy3-preview-free -> nemotron-3-super-free -> gpt-5-nano`.

**Concern 2 — `FREE_ONLY_PROVIDER_IDS = ["opencode"]` is too broad**

You're right that paid Zen subscribers also see `connectedProviders === ["opencode"]`, so the "free-only" gate as written can downgrade a paying user. I deliberately kept this PR scoped to the deprecated-model fix because rewiring the detection (use `availableModels` as the source of truth) touches the resolver in a deeper way and changes the shape of all 3 existing tests in this file. I'd rather land the no-regret fix here, then open a follow-up that:

1. drops `FREE_ONLY_PROVIDER_IDS` entirely
2. derives "free-only" from `availableModels.every(m => isKnownFreeModel(m))` (or the inverse: any paid model present -> not free-only)
3. moves the canonical free-model list to a snapshot file refreshed on the same cadence as `refresh-model-capabilities.yml` (your bigger-picture suggestion)

Happy to take that follow-up if @code-yeongyu agrees with the direction. Want me to open a tracking issue?

Also +1 on cubic's findings — the current heuristic is doing two jobs (gating + chain construction) that should be separated.

mrosnerr:
Opened a PR against this branch with some test additions and a simplified approach to the free-only detection: https://github.com/MoerAI/oh-my-opencode/pull/1

The main concern we found: `isFreeOnlyProviderConfiguration(["opencode"])` returns `true` for both free-tier and paid Zen subscribers, since both produce `connectedProviders === ["opencode"]`. This causes paid users to get force-downgraded to free models on the warm-cache path.

Our approach replaces the category-default gating and chain replacement with an append strategy — free models get added as a tail to the existing chain, and on warm cache, free-only is derived from `availableModels` instead of `connectedProviders`. The cold-cache path still uses provider-based detection as a fallback (noted with a TODO).

Feel free to merge it in, cherry-pick what's useful, or just use it as context for the edge cases — whatever works best for getting the right behavior out.

MoerAI:
Closing — @mrosnerr identified a real bug in my approach: `isFreeOnlyProviderConfiguration(["opencode"])` returns true for both free-tier and paid Zen subscribers, force-downgrading paid users. They opened MoerAI/oh-my-opencode#1 against this branch with a corrected approach (derive free-only from `availableModels` instead of `connectedProviders`, append free models as a tail to the chain rather than replace).

Their approach is the right one — I'd rather they take this forward as their own PR against upstream/dev so the credit goes to them. Their PR description is also a much better explanation of the edge cases than mine. @mrosnerr if you want to open a PR against `code-yeongyu/oh-my-openagent` directly, please go ahead.

Thanks for the deep review @mrosnerr.

PR #3534, opened 2026-04-20
Fix Claude Opus 4.7 context-limit handling and alias mapping

LZong-tw:
## Summary

- Extend cached-context-limit regex to accept Claude **4.7** in addition to 4.6, and to accept both dot (`4.7`) and dash (`4-7`) alias forms.
- Resolver now iterates every alias variant against the per-model context-limit cache before falling back to the 200K default, so callers using either form get the correct limit.
- Provider config writer mirrors the same alias forms into the cache, ensuring lookups succeed regardless of which form is used at the call site.
- Add a defensive log in the preemptive compaction hook when the resolver returns the 200K default, so alias/mapping regressions surface early instead of silently triggering early compaction.

## Why

Without this fix, Claude Opus 4.7 (e.g. via GitHub Copilot, where the model id appears as `claude-opus-4.7`) was not matched by `supportsCachedAnthropicLimit`, so even when an explicit context limit was cached, the resolver fell back to the 200K default. Combined with caches that only stored a single key form, the actual real-world limit was effectively lost across the dot/dash alias boundary, and preemptive compaction could trigger far earlier than necessary (observed at ~150K input on a model with a much larger window).

## Changes

- `src/shared/context-limit-resolver.ts`
  - `supportsCachedAnthropicLimit`: regex now matches `claude-(opus|sonnet)-4(?:-|\.)[67](?:-high)?`.
  - New helper `generateClaudeAliasForms(modelID)` produces every dot/dash variant for 4.6 and 4.7.
  - `resolveActualContextLimit` tries each alias key against the cache before falling back to `DEFAULT_ANTHROPIC_ACTUAL_LIMIT`.
- `src/plugin-handlers/provider-config-handler.ts`
  - When applying provider config, the context limit is written under both dot and dash forms for 4.6 and 4.7 variants.
- `src/hooks/preemptive-compaction-trigger.ts`
  - Adds a defensive log when the resolver returns exactly `DEFAULT_LIMIT (200000)`, including providerID and modelID, to make alias/mapping issues observable.

## Verification

- `git diff --stat`: 3 files changed, +55/-8.
- All existing logic preserved; changes are minimal and additive on the resolver/config side.
- Manually exercised against `github-copilot/claude-opus-4.7` (dot) and `claude-opus-4-7` (dash) with explicit context limits in the provider cache.

## Notes

- No changes to `provider-model-id-transform.ts` or any Gemini handling.
- Behavior for Claude 4.6 is unchanged in steady state, but its cache writes/reads are now alias-resilient too.

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3534"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Fixes context-limit detection for Claude Opus 4.7. Recognizes both dot and dash model IDs and aligns cache reads/writes to avoid falling back to the 200K default and triggering early compaction.

- **Bug Fixes**
  - Resolver now matches `claude-(opus|sonnet)-4[.-][67](-high)?` and tries all alias forms against the cache before defaulting.
  - Provider config writes context limits under both dot and dash keys for 4.6/4.7 variants to improve cache hits.
  - Adds a defensive log when the resolver returns the 200K default, including provider and model IDs.

<sup>Written for commit e9d64bf45d5acdaf2ed965e721193d9e93c40f46. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/preemptive-compaction-trigger.ts
+++ b/src/hooks/preemptive-compaction-trigger.ts
@@ -71,6 +71,17 @@ export async function runPreemptiveCompactionIfNeeded(args: {
     modelCacheState,
   )
 
+  // Defensive logging: surface when we fallback to a default context limit
+  const DEFAULT_LIMIT = 200000
+  if (actualLimit === DEFAULT_LIMIT) {
+    log("[preemptive-compaction] context limit fallback to default (possible alias/mapping issue)", {
+      sessionID,
+      providerID: cached.providerID,
+      modelID: cached.modelID,
+      actualLimit,
+    })
+  }
+
   if (actualLimit === null) {
     log("[preemptive-compaction] Skipping preemptive compaction: unknown context limit for model", {
       providerID: cached.providerID,
--- a/src/plugin-handlers/provider-config-handler.ts
+++ b/src/plugin-handlers/provider-config-handler.ts
@@ -64,10 +64,25 @@ export function applyProviderConfig(params: {
       const contextLimit = modelConfig?.limit?.context;
       if (!contextLimit) continue;
 
-      modelContextLimitsCache.set(
-        `${providerID}/${modelID}`,
-        contextLimit,
-      );
+      // Write both dot and dash alias forms to improve cache hits for both 4.6/4-6 and 4.7/4-7 forms
+      const primaryKey = `${providerID}/${modelID}`
+      modelContextLimitsCache.set(primaryKey, contextLimit)
+
+      // Derive counterpart forms when possible
+      let dashForm = modelID
+      if (modelID.includes("4.6")) dashForm = modelID.replace("4.6", "4-6")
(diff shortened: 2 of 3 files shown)


LZong-tw:
Closing — the regex fix landed upstream in v3.17.15 (`/^claude-(opus|sonnet)-4(?:-|\.)(?:6|7)(?:-high)?$/` now matches both dot and dash forms in supportsCachedAnthropicLimit). Local opencode.json override no longer needed for the regex layer; remaining Copilot context limit handling is out of scope for this PR.

PR #3537, opened 2026-04-20
Fix: Restrict max effort to claude-opus-4-7+ only

arobot001:
## Summary

Fixes the error: `Bad Request: {"error":{"message":"output_config.effort \"max\" is not supported by model claude-opus-4.6; supported values: [low medium high]"...}}`

## Problem

The `OPUS_PATTERN` regex in `src/hooks/anthropic-effort/hook.ts` was matching all models containing `claude-` followed by `opus`, including `claude-opus-4.6`. However, only `claude-opus-4-7` and later versions actually support the `max` effort level. Earlier versions like `4.6` only support `low`, `medium`, and `high`.

## Solution

Changed the pattern from:
```typescript
const OPUS_PATTERN = /claude-.*opus/i
```

To:
```typescript
const OPUS_PATTERN = /claude-opus-4-[7-9]|claude-opus-4-\d{2,}/i
```

This pattern now only matches:
- `claude-opus-4-7`, `claude-opus-4-8`, `claude-opus-4-9` (versions 7-9)
- `claude-opus-4-10`, `claude-opus-4-11`, etc. (double-digit versions and beyond)

And correctly excludes:
- `claude-opus-4-6` and earlier versions

## Testing

- `claude-opus-4.6` → will be clamped to `high` effort
- `claude-opus-4-7` → will use `max` effort
- `claude-opus-4-10` → will use `max` effort

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3537"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Restricts "max" effort to `claude-opus-4-7+` models to prevent Bad Request errors on older Opus versions. Older models (e.g., `claude-opus-4.6`) now cap at "high".

- **Bug Fixes**
  - Updated `OPUS_PATTERN` to `/claude-opus-4-[7-9]|claude-opus-4-\d{2,}/i` so only `claude-opus-4-7+` uses `max`.
  - Ensures earlier Opus versions use `low | medium | high` only.

<sup>Written for commit ae0756c8a82ce7fffb446a13dd8676fee7688b2c. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/anthropic-effort/hook.ts
+++ b/src/hooks/anthropic-effort/hook.ts
@@ -1,6 +1,8 @@
 import { isProviderUsingOAuth, log, normalizeModelID } from "../../shared"
 
-const OPUS_PATTERN = /claude-.*opus/i
+// Only claude-opus-4-7 (and future versions) support "max" effort
+// claude-opus-4.6 and earlier versions only support low | medium | high
+const OPUS_PATTERN = /claude-opus-4-[7-9]|claude-opus-4-\d{2,}/i
 const EFFORT_UNSUPPORTED_PATTERN = /claude-.*haiku/i
 const INTERNAL_SKIP_AGENTS = new Set(["title", "summary", "compaction"])
 
@@ -54,7 +56,7 @@ interface ChatParamsOutput {
 
 /**
  * Valid thinking budget levels per model tier.
- * Opus supports "max"; all other Claude models cap at "high".
+ * Only claude-opus-4-7+ supports "max"; all other Claude models cap at "high".
  */
 const MAX_VARIANT_BY_TIER: Record<string, string> = {
   opus: "max",


code-yeongyu:
[sisyphus-bot]

- Author: `arobot001`
- Title: `Fix: Restrict max effort to claude-opus-4-7+ only`
- Review decision: none yet
- Mergeable state: MERGEABLE

Awaiting a first-pass review.

code-yeongyu:
[sisyphus-bot] Hi arobot001. 🙏 Thanks for reporting the actual `output_config.effort "max" is not supported by model claude-opus-4.6` error and tracing it back to `OPUS_PATTERN`. The diagnosis is exactly right: the old `/claude-.*opus/i` happily matches `claude-opus-4-6`, so the 4-6 path ends up with `effort: max` and Anthropic rejects it.

Two things I'd like before merging.

**1. Add a focused regression test (small, but important)**

`src/hooks/anthropic-effort/index.test.ts` already covers the OAuth clamp path for opus, but there's no test that pins `claude-opus-4-6` (or the dotted `claude-opus-4.6`) to non-max. Could you add one that, given an API-key (non-OAuth) Anthropic session and `modelID: "claude-opus-4-6"`, expects `output.options.effort` to NOT be `"max"`? Adding a complementary positive case for `claude-opus-4-7` (which is what existing tests already use) and at least one double-digit case like `claude-opus-4-10` would lock the new regex against silent regression.

**2. Future-proof the regex a bit**

The current `/claude-opus-4-[7-9]|claude-opus-4-\d{2,}/i` correctly excludes `4-0`..`4-6` and includes `4-7`..`4-99`. The miss is `claude-opus-5-0` and beyond: those won't match either, even though future Opus generations are expected to support `max` the same way 4-7 does. The PR description says "claude-opus-4-7 and later" but the regex stops at 4.99. A small extension like:

```ts
const OPUS_PATTERN = /claude-opus-4-(?:[7-9]|\d{2,})|claude-opus-(?:[5-9]|\d{2,})/i
```

keeps every current case the same while also matching `claude-opus-5-0`, `claude-opus-6-0`, etc. when they arrive. Up to you whether to go this route or restrict the regex to "today only" and address future generations when they ship; I just want the choice to be explicit rather than implicit.

**Non-blocking observations**

- The doc comment update on `MAX_VARIANT_BY_TIER` ("Only claude-opus-4-7+ supports max...") is a nice touch.
- This PR is orthogonal to #4118 (which touches `isClaudeOpus47Model` and `isClaudeProvider` whitelists). Both want different fixes in different files, so landing them independently is fine.

Once the test lands (and optionally the future-proof regex extension), this is otherwise ready. CI is currently green; mergeStateStatus is clean. Thanks for catching this and for the very clear write-up of the failure mode.

code-yeongyu (changes requested):
[sisyphus-bot] See accompanying comment for the requested changes (add a regression test pinning claude-opus-4-6 to non-max and optionally extend the regex to cover future opus-5+ generations).

code-yeongyu:
[sisyphus-bot] Thanks for catching the opus effort-regex bug! The anthropic-effort hook has since been removed/reworked on dev and the model lineup moved to opus-5, so this no longer applies. Closing — thanks for the sharp report.

PR #3604, opened 2026-04-23
fix: dynamic import bun:sqlite to prevent plugin context crash

### LINKED ISSUE #3143: [Bug]: oh-my-opencode@3.15.2 causes opencode to freeze/hang at TUI startup
## Bug Description

When oh-my-opencode@3.15.2 plugin is loaded, opencode TUI freezes/blank screen at startup. 

Downgrading to oh-my-opencode@3.14.0 resolves the issue - opencode starts normally.

## Prerequisites

- [x] I am using the latest version of oh-my-opencode (3.15.2)

## Environment

- **OpenCode version**: 1.3.15
- **Operating System**: macOS (Apple Silicon)
- **Node version**: v24.13.0
- **Plugin version**: oh-my-opencode@3.15.2

## Steps to Reproduce

1. Install oh-my-opencode@3.15.2
2. Run `opencode` to start the TUI
3. TUI displays blank/white screen and freezes
4. Process must be killed with SIGKILL

## Expected Behavior

OpenCode TUI should start and display normally.

## Actual Behavior

- TUI is completely blank (white screen)
- No response to any input
- Must force kill the process
- No error logs are produced (process hangs silently)

### Working Configuration

```json
{
  "plugin": ["oh-my-opencode@3.14.0"],
  "mcp": { ... }
}
```

### Broken Configuration

```json
{
  "plugin": ["oh-my-opencode@3.15.2"],
  "mcp": { ... }
}
```

### Diagnosis

- Without plugin (`--pure` mode): Works fine
- With oh-my-opencode@3.14.0: Works fine
- With oh-my-opencode@3.15.2: Hangs at \"loading plugin\" log message

### Log Analysis

When opencode starts with 3.15.2, the last log entry is:
```
INFO  service=plugin path=oh-my-opencode@3.15.2 loading plugin
```

After this, no further logs are produced - the plugin initialization appears to hang silently.

## Possible Related Issues

May be related to:
- Issue #3137: plugin fails to load agents
- Issue #3139: token efficiency regression in latest version

## Questions for Maintainers

1. Is there a specific change in 3.15.2 that could cause plugin initialization to hang?
2. Is there a migration issue with the config file rename (oh-my-opencode.json → oh-my-openagent.json) that could cause this?
3. Could this be related to the Flock file locking mechanism in opencode 1.3.15?

### LINKED ISSUE #3475: [Bug] TUI black screen on startup: bun:sqlite import incompatible with OpenCode embedded JS runtime
## Description

First off — huge thanks for building OmO. It's an incredible piece of work that transforms how we use OpenCode. 🙏

I ran into a black screen issue on Windows and spent some time digging into the root cause. Sharing my findings below in case it helps others, and to see if there's a path toward a proper fix.

## Environment

| Component | Version |
|-----------|---------|
| OS | Windows 11 (x64) |
| OpenCode | 1.4.6 (Go binary via npm) |
| oh-my-opencode | 3.17.4 |
| Bun | 1.3.12 |
| Node.js | v22.22.2 |

## Symptoms

After installing OmO and registering the plugin in `opencode.json`, running `opencode` shows a completely black/blank TUI. The process hangs and becomes unresponsive. Removing the plugin from config restores normal startup.

## Root Cause

The plugin bundle (`dist/index.js`) contains a Bun-specific import at line 147678:

```javascript
// src/plugin/ultrawork-db-model-override.ts
import { Database } from "bun:sqlite";
```

OpenCode's Go binary has an **embedded JS runtime** for loading plugins. This runtime does not support the `bun:` protocol. When the plugin is loaded synchronously at startup, this unresolved import blocks the main thread, preventing the TUI from rendering.

### How I confirmed this

Using OpenCode's debug logging:

```bash
# With plugin — hangs after config loading, never reaches TUI initialization
opencode --print-logs --log-level DEBUG

# Pure mode (no plugins) — loads all TUI plugins normally
opencode --pure --print-logs --log-level DEBUG
```

In pure mode, logs show `loading internal tui plugin` entries immediately. With the plugin, the process stops at config loading and never reaches TUI plugin initialization.

Additionally, the `models.dev` fetch during plugin init times out in China (GFW), which adds to the startup delay — but the primary blocker is the `bun:sqlite` import.

## Workaround

I was able to fix the black screen by replacing the `bun:sqlite` import with a safe stub:

```bash
node -e "
const fs = require('fs');
const home = process.env.HOME || process.env.USERPROFILE;
const targets = [
  home + '/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/dist/index.js',
  home + '/.config/opencode/node_modules/oh-my-opencode/dist/index.js'
].filter(p => { try { return fs.existsSync(p); } catch { return false; } });

const old = 'import { Database } from \"bun:sqlite\";';
const stub = 'var Database = class { constructor() { return { prepare: () => ({ get: () => null, run: () => null, all: () => [] }) }; } };';

targets.forEach(path => {
  let content = fs.readFileSync(path, 'utf-8');
  if (content.includes(old)) {
    content = content.replace(old, stub);
    fs.writeFileSync(path, content, 'utf-8');
    console.log('Patched: ' + path);
  }
});
"
```

After patching, OpenCode starts normally with full OmO functionality — Sisyphus agent, background agents, MCPs, ultrawork, etc. Only the `ultrawork-db-model-override` feature (model name override in session DB) is affected by the stub.

**Limitation:** The patch must be re-applied after every OmO update.

## Suggested Fix

A few possible approaches, from simplest to most robust:

1. **Conditional import** — Wrap the `bun:sqlite` import in a try/catch or use dynamic `import()` so the module degrades gracefully when Bun APIs are unavailable.

2. **Runtime detection** — Check for `typeof Bun !== 'undefined'` before using Bun-specific APIs, and fall back to a no-op or alternative SQLite library.

3. **Build-time target** — If the bundle is built with Bun, consider producing a runtime-agnostic variant for environments that don't ship the Bun runtime.

I'm happy to submit a PR if any of these approaches sound reasonable.

## Additional Context for Users in China

Users in mainland China face an additional layer of difficulty:
- `bunx oh-my-opencode install` may timeout due to npm registry access
- `opencode plugin` command also times out when downloading packages
- `models.dev` fetch during init is blocked by GFW

Workarounds: set `npm config set registry https://registry.npmmirror.com/` before installation, and use `OMO_SEND_ANONYMOUS_TELEMETRY=0` to skip analytics calls.

## Related Issues

- #596 — Blank screen after installing
- #1292 — v3.1.9 TUI freeze (CRITICAL)
- #1301 — `@latest` causes TUI hang
- #3143 — v3.15.2 causes freeze/hang
- [anomalyco/opencode#12192](https://github.com/anomalyco/opencode/issues/12192) — Black screen on Windows

Happy to provide any additional debugging info if needed. Thanks again for the great project! 🚀

leecoder:
## Problem

`import { Database } from "bun:sqlite"` in `ultrawork-db-model-override.ts` is a **static import** that fails at module parse time in OpenCode's embedded JS runtime, which doesn't support the `bun:` protocol in plugin context. This causes the TUI to black screen/hang on startup.

Affects users on OpenCode >= 1.3.15 with Bun >= 1.3.12, across Windows, macOS, Linux, and Docker.

## Root Cause

The `bun:` protocol is not in the allowed scheme list (`file`, `data`, `node`, `electron`) for the default ESM loader. A static import triggers validation at parse time — before any code runs — blocking the entire plugin from loading.

## Solution

Replace the static import with a dynamic `await import("bun:sqlite" as never as string)` wrapped in try/catch:

- When `bun:sqlite` **is available** (native Bun runtime): works exactly as before
- When `bun:sqlite` **is unavailable** (OpenCode plugin context): logs a message and silently skips the deferred model override

The `as never as string` cast is the established pattern in production Bun codebases (e.g., [clarte](https://github.com/michaelabrt/clarte)) to bypass TypeScript's static analysis of the `bun:` protocol.

## Changes

- `src/plugin/ultrawork-db-model-override.ts`: Replace static `bun:sqlite` import with async dynamic import + graceful fallback
- No new dependencies
- No behavior change when running under native Bun
- Existing tests pass unchanged (6/6, they run under Bun where `bun:sqlite` works)
- Bundle verified: no static `import ... from "bun:sqlite"` in `dist/index.js`

## Verification

- `bunx tsc --noEmit` — clean
- `bun run build` — success
- `bun test src/plugin/ultrawork-db-model-override.test.ts` — 6/6 pass
- `grep 'import.*from.*"bun:sqlite"' dist/index.js` — no matches

Fixes #3475
Fixes #3143

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3604"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Prevents plugin startup crash by replacing the static `bun:sqlite` import with a safe dynamic import in the deferred model override path. Native Bun behavior is unchanged; in unsupported runtimes, the override is skipped instead of crashing.

- **Bug Fixes**
  - Replaced static import with `await import("bun:sqlite")` and a graceful fallback.
  - Updated `src/plugin/ultrawork-db-model-override.ts` to log and skip when `bun:sqlite` isn’t available.
  - No new dependencies; no static `import ... from "bun:sqlite"` remains in the build.

<sup>Written for commit 7fc016d718f70017b1ffb1e69557a21e3d2c5547. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin/ultrawork-db-model-override.ts
+++ b/src/plugin/ultrawork-db-model-override.ts
@@ -1,8 +1,18 @@
-import { Database } from "bun:sqlite"
 import { join } from "node:path"
 import { existsSync } from "node:fs"
 import { getDataDir } from "../shared/data-path"
 import { log } from "../shared"
+
+type BunSqliteModule = typeof import("bun:sqlite")
+type BunDatabase = import("bun:sqlite").Database
+
+async function loadBunSqlite(): Promise<BunSqliteModule | null> {
+  try {
+    return await import("bun:sqlite" as never as string)
+  } catch {
+    return null
+  }
+}
 
 function getDbPath(): string {
   return join(getDataDir(), "opencode", "opencode.db")
@@ -11,7 +21,7 @@ function getDbPath(): string {
 const MAX_MICROTASK_RETRIES = 10
 
 function tryUpdateMessageModel(
-  db: InstanceType<typeof Database>,
+  db: BunDatabase,
   messageId: string,
   targetModel: { providerID: string; modelID: string },
   variant?: string,
@@ -30,7 +40,7 @@ function tryUpdateMessageModel(
 }
 
 function retryViaMicrotask(
-  db: InstanceType<typeof Database>,
+  db: BunDatabase,
   messageId: string,
   targetModel: { providerID: string; modelID: string },
   variant: string | undefined,
@@ -112,16 +122,22 @@ export function scheduleDeferredModelOverride(
   targetModel: { providerID: string; modelID: string },
   variant?: string,
 ): void {
-  queueMicrotask(() => {
+  queueMicrotask(async () => {
+    const mod = await loadBunSqlite()
+    if (!mod) {
+      log("[ultrawork-db-override] bun:sqlite unavailable, skipping deferred override")
+      return
+    }
+
     const dbPath = getDbPath()
     if (!existsSync(dbPath)) {
       log("[ultrawork-db-override] DB not found, skipping deferred override")
       return
     }
 
-    let db: InstanceType<typeof Database>
+    let db: BunDatabase
     try {
-      db = new Database(dbPath)
+      db = new mod.Database(dbPath)
     } catch (error) {
       log("[ultrawork-db-override] Failed to open DB, skipping deferred override", {
         messageId,


leecoder:

code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED
- Mergeable state: **CONFLICTING** (merge conflict with current `dev`)

Approved but blocked on merge conflicts. Please rebase onto current `dev` and resolve. Once green, the owner can merge.

Assigning `code-yeongyu` for owner visibility on the merge decision.

code-yeongyu (approved):
[sisyphus-bot] APPROVED. Dynamic-import `bun:sqlite` to prevent plugin context crash when bun's sqlite module isn't available at load time. +22/-6 single file. Cubic approved. Merging.

code-yeongyu:
[sisyphus-bot] Hi leecoder. 🙏 Thanks for catching the static `bun:sqlite` import crash. The diagnosis was right.

I picked this back up after the 5/20 approval to push it through, and the rebase against current `dev` made it clear that `dev` already carries a different implementation of the same fix. The version on `dev` defines an `importBunSqlite()` helper that does:

```ts
const dynamicImport = new Function("return import('bun:sqlite')") as () => Promise<typeof import("bun:sqlite")>
return await dynamicImport()
```

The `new Function("return import('bun:sqlite')")` form hides the `bun:` specifier from Node.js's / Electron's static ESM parser at parse time, so the plugin module never throws before `.catch()` can run. Your PR uses the more idiomatic `await import("bun:sqlite" as never as string)`, which works in Bun but can still trip the Node.js static parser in some embedded contexts. Both target the same crash, but the `new Function` form sidesteps the parse-time hazard you originally identified more conservatively.

Because `dev`'s implementation supersedes this PR functionally and is arguably the safer of the two for the Electron / Node.js host case, I'm going to close this out. The crash mode is no longer reproducible on current `dev`, so the user-facing bug is resolved.

If you spot a context where `dev`'s `importBunSqlite` still crashes (the embedded Electron runtime, for example), please open a fresh issue with the reproduction and I'll happily re-open the investigation. Really appreciate the work; the original diagnosis was exactly the right thread to pull.

PR #3606, opened 2026-04-24
Fix #3512: remove invisible agent name prefixes

### LINKED ISSUE #3512: Zero-width space prefixes in AGENT_LIST_SORT_PREFIXES cause garbled agent name rendering in Warp Terminal
## Summary

`AGENT_LIST_SORT_PREFIXES` uses U+200B (zero-width space) characters to control agent list sort order. Some terminal emulators don't handle these correctly, causing garbled display of agent names in the TUI status line.

## Environment

- Plugin: `oh-my-openagent@3.17.4`
- OpenCode: `1.4.11`
- Terminal: **Warp Terminal** `v0.2026.04.15.08.45.stable_02` on macOS (darwin arm64)
- `TERM=xterm-256color`

## Reproduction

1. Install `oh-my-openagent` and use any of the prefixed agents (`sisyphus`, `hephaestus`, `prometheus`, `atlas`).
2. Observe the agent name in the OpenCode TUI status line / session footer.

## Expected

```
Sisyphus - Ultraworker · Claude Opus 4.7 · 4m 18s
```

## Actual (Warp)

```
Sisyphus - Ultraworke r· Claude Opus 4.7 · 4m 18s
                    ^  <- spurious visible gap
```

After reflow / truncation in the same session:

```
Sisyphus - ltraworker  · Claude Opus 4.7 Anthropic
          ^ <- leading 'U' is cut, trailing space preserved
```

Image

The zero-width prefix (`\u200B` for sisyphus, up to `\u200B\u200B\u200B\u200B` for atlas) appears to:
1. Render as a visible advance in Warp's text engine (not zero-width).
2. Cause width miscalculation during truncation — the TUI assumes zero width but the terminal renders non-zero, so chars are cut off at the wrong position.

## Root cause

Source: `src/shared/agent-display-names.ts` (compiled into `dist/index.js` and `dist/cli/index.js`)

```js
AGENT_LIST_SORT_PREFIXES = {
  sisyphus: "\u200B",
  hephaestus: "\u200B\u200B",
  prometheus: "\u200B\u200B\u200B",
  atlas: "\u200B\u200B\u200B\u200B"
};
```

Used in `getAgentRuntimeName()`:

```js
function getAgentRuntimeName(configKey) {
  const displayName = getAgentDisplayName(configKey);
  const prefix = AGENT_LIST_SORT_PREFIXES[configKey.toLowerCase()];
  return prefix ? `${prefix}${displayName}` : displayName;
}
```

The prefixed string becomes both the sort key AND the visible runtime name, which is why the invisible characters leak into the TUI.

## Suggested fixes (in order of preference)

1. **Use OpenCode's native agent ordering API** if available, instead of relying on string-sort with invisible characters.
2. **Decouple sort key from display name**. Store prefixed name as a separate `sortKey` field internally; return the clean display name from `getAgentRuntimeName`.
3. **Add an opt-out config flag** (e.g. `disable_agent_sort_prefixes: true` in the `oh-my-opencode.json` schema) for users on affected terminals.

Option (2) is cleanest — there's no truly safe zero-width character across all terminal emulators.

## Workaround for affected users

```bash
# Adjust path for your platform
PLUGIN_DIR=~/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/dist

# Backup
cp "$PLUGIN_DIR/index.js"     "$PLUGIN_DIR/index.js.bak"
cp "$PLUGIN_DIR/cli/index.js" "$PLUGIN_DIR/cli/index.js.bak"

# Strip the prefixes
perl -i -pe 's/"\\u200B(\\u200B)*"/""/g' \
  "$PLUGIN_DIR/index.js" \
  "$PLUGIN_DIR/cli/index.js"
```

Side effect: agents sort alphabetically instead of the intended custom order. Wiped on plugin reinstall.

## Other affected terminals (likely)

Warp is block-based with a custom text renderer — the most prone to U+200B handling bugs. Other candidates: some versions of iTerm2 with ligature-aware fonts, Hyper, and various web-based terminals. Reporter has only confirmed Warp.

---

_Related: `AGENT_DISPLAY_NAMES` and `AGENT_LIST_SORT_PREFIXES` at `src/shared/agent-display-names.ts`._

lucasyounger:
## Summary

- fix garbled agent name rendering by removing invisible zero-width prefixes from emitted runtime/list names
- keep backward compatibility by preserving invisible-character stripping during agent-name lookups
- update the agent ordering docs to describe the order-field based approach

## Changes

- remove `AGENT_LIST_SORT_PREFIXES` from `src/shared/agent-display-names.ts`
- make `getAgentRuntimeName()` and `getAgentListDisplayName()` return clean display names
- update agent name tests to assert prefix-free runtime names while keeping legacy normalization coverage
- refresh `src/plugin-handlers/AGENTS.md` so it no longer documents ZWSP-based ordering

## Testing

Installed Bun locally on Windows (`1.3.13`) and verified the change with:

```bash
bun run typecheck
bun run build
```

Both commands completed successfully.

## Related Issues

Closes #3512

### DIFF
--- a/src/plugin-handlers/AGENTS.md
+++ b/src/plugin-handlers/AGENTS.md
@@ -8,28 +8,31 @@ The canonical agent order is **sisyphus → hephaestus → prometheus → atlas*
 
 This order is enforced via two mechanisms working together:
 1. `CANONICAL_CORE_AGENT_ORDER` in `agent-priority-order.ts` controls object key insertion order
-2. `agent-key-remapper.ts` injects ZWSP-prefixed runtime names into the `name` field for OpenCode's `localeCompare` sort
+2. `agent-priority-order.ts` injects explicit `order` fields onto core agent configs so OpenCode can sort them without mutating visible names
 
 ### Why Two Mechanisms
 
-OpenCode's `Agent.list()` sorts agents by `name` field via `localeCompare`. Object key order alone is not enough. The `name` field carries ZWSP prefixes (1-4 chars) so core agents sort before alphabetically-named agents.
+Object key order alone is not enough because multiple merge paths can reshuffle assembled configs. The emitted `order` field provides a stable, explicit priority without leaking invisible characters into the UI or terminal renderers.
 
-ZWSP is intentionally used in the `name` field only. It MUST NOT appear in:
+Historical ZWSP-prefixed names may still appear in old sessions or cached state. They MUST NOT be emitted in current runtime names, object keys, or display names. Compatibility stripping lives in `shared/agent-display-names.ts`.
+
+Invisible characters MUST NOT appear in:
 - Object keys (used as HTTP header values, causes RFC 7230 violations)
+- Runtime names returned by `getAgentRuntimeName()`
 - Display names returned by `getAgentDisplayName()`
 - Config keys
 
 ### History
 
 Agent ordering has caused 15+ commits, 8+ PRs, and multiple reverts due to:
-1. Early ZWSP attempts that leaked into HTTP headers via object keys
+1. Early invisible-prefix attempts that leaked into visible UI surfaces
 2. Object.entries() iteration order depending on merge sequence
 3. Multiple code paths assembling agents differently
 
 ### Forbidden Patterns
 
 DO NOT introduce:
-- ZWSP in object keys or display names (only allowed in `name` field via `getAgentRuntimeName()`)
+- ZWSP or other invisible sort prefixes in emitted names
 - Runtime sort shims or comparators
 - Alternative ordering constants
 - Object.entries() order dependencies
--- a/src/shared/agent-display-names.test.ts
+++ b/src/shared/agent-display-names.test.ts
@@ -1,5 +1,13 @@
 import { describe, it, expect } from "bun:test"
-import { AGENT_DISPLAY_NAMES, getAgentConfigKey, getAgentDisplayName, getAgentListDisplayName, normalizeAgentForPrompt, normalizeAgentForPromptKey } from "./agent-display-names"
+import {
+  AGENT_DISPLAY_NAMES,
+  getAgentConfigKey,
+  getAgentDisplayName,
+  getAgentListDisplayName,
+  getAgentRuntimeName,
+  normalizeAgentForPrompt,
+  normalizeAgentForPromptKey,
+} from "./agent-display-names"
 
 describe("getAgentDisplayName", () => {
   it("returns display name for lowercase config key (new format)", () => {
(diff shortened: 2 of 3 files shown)


lucasyounger:

lucasyounger:
recheck

ANPULI:
Tested locally. The rendering fix is correct, but I want to flag an ordering regression that I think blocks landing this as-is on opencode ≤ 1.14.22.

**Test setup**
- Checked out `codex/fix-3512-agent-name-rendering` at `7042b66f`
- `bun install && bun run build` — clean, all 31 tests in `agent-display-names.test.ts` pass
- Replaced `~/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/dist/` with the built output
- Env: WSL2 Ubuntu 24.04 + Windows Terminal, opencode 1.14.22, oh-my-openagent 3.17.4

**What works ✅**
Status bar renders `Sisyphus - Ultraworker` cleanly. No stray space, no color break, separator spacing correct. Verified across new sessions and the agent picker label.

**What regresses ⚠️ — core agent order**
TUI agent picker (Tab) shows the core four out of canonical order. Observed top of list:

```
Sisyphus - Ultraworker        ← default agent stays first
Atlas - Plan Executor          ← then alphabetical
Hephaestus - Deep Agent
Prometheus - Plan Builder
```

Expected: Sisyphus → Hephaestus → Prometheus → Atlas.

**Root cause (verified in both repos)**

The PR's plugin code is correct: `reorderAgentsByPriority` in `src/plugin-handlers/agent-priority-order.ts:47-50` does inject `{...agent, order: 1..4}` and is called from `applyAgentConfig` in `src/plugin-handlers/agent-config-handler.ts:389`. Tests at `src/plugin-handlers/agent-priority-order.test.ts:196-199` assert this. So OMO is delivering the right shape.

The problem is opencode core does not honor `order` yet:

- `Agent.Info` schema (`packages/opencode/src/agent/agent.ts:27-52`) has no `order` field.
- `Agent.list()` (same file, ~lines 285-295) sorts purely as:
  ```ts
  sortBy(
    [(x) => (cfg.default_agent ? x.name === cfg.default_agent : x.name === "build"), "desc"],
    [(x) => x.name, "asc"],
  )
  ```
- The TUI picker (`packages/opencode/src/cli/cmd/tui/component/dialog-agent.tsx`) consumes that list verbatim with no further sort.
- The upstream PR that would add `order` field support, sst/opencode#19127, is **still open** as of today.

So removing the ZWSP prefixes (which were load-bearing as a name-based sort hack) without `order` being honored downstream falls back to alphabetical-by-name, which produces exactly the order I observed (Sisyphus first as default, then A/H/P alphabetical). This is the same regression pattern that caused PR #3260 to be reverted in #3265.

**Suggested paths forward** (any of these would unblock; deferring to maintainers on which fits the project's style):

1. **Gate on opencode version**: keep the new clean-name path behind a runtime check for an opencode version that honors `order`; fall back to ZWSP prefixes on older versions. Lets this PR ship without regressing existing users.
2. **Wait for sst/opencode#19127** to merge and a release to ship, then land this PR.
3. **Land in two commits**: one that adds `order` injection (already present on `dev`), one that removes ZWSP — sequence the second with the upstream release.

Happy to test any revision. Thanks for tackling this — the rendering fix itself is solid and the test changes capture the right invariants.

lucasyounger:
Closing this for now. The rendering fix is valid, but keeping this PR open is misleading because it regresses canonical core-agent ordering on opencode <= 1.14.22. The current change removes the invisible name-prefix fallback before upstream order support is available, so this is not safe to merge yet. I'll reopen or submit a follow-up once upstream order support lands, or if we add a version-gated compatibility path for older opencode releases.

PR #3608, opened 2026-04-24
fix(anthropic-effort): clamp pre-set effort max to high for constrained providers (fixes #3563)

### LINKED ISSUE #3563: Bug: effort "max" not clamped for constrained providers when pre-set via session params
### Error

```
{"error":{"message":"output_config.effort \"max\" is not supported by model claude-opus-4.6; supported values: [low medium high]","code":"invalid_reasoning_effort"}}
```

This occurs when using `claude-opus-4.6` via the `github-copilot` provider.

### Environment

- **opencode**: v1.14.19
- **oh-my-openagent**: latest (via `oh-my-openagent@latest` plugin)
- **Provider**: `github-copilot`
- **Model**: `claude-opus-4.6`

### Root Cause

The `anthropic-effort` hook in `src/hooks/anthropic-effort/hook.ts` has an early return at **line 82**:

```typescript
if (output.options.effort !== undefined) return
```

This skips **all** downstream logic — including the constrained-provider clamping — when `effort` is already set on `output.options`. 

The problem is that `effort` can be pre-populated with `"max"` before the hook runs, via two paths:

1. **Session params** (`src/plugin/chat-params.ts` lines 102-107) spread stored session parameters onto `output.options` *before* the anthropic-effort hook is called at line 180.
2. **Model requirements** (`src/shared/model-requirements.ts`) hardcode `variant: "max"` for all Opus models in built-in fallback chains (sisyphus, oracle, prometheus, metis, momus, and several categories).

When either path sets `effort: "max"`, the hook sees it's already defined and returns immediately — never reaching `isConstrainedProvider()` / `clampVariant()`, which would correctly downgrade `"max"` → `"high"` for `github-copilot`.

### Why existing tests don't catch this

The test at `src/hooks/anthropic-effort/index.test.ts` line 156 covers GitHub Copilot clamping, but only when `effort` is **not** pre-set on `output.options`. It doesn't test the scenario where effort is already `"max"` when the hook is invoked.

### Proposed Fix

Replace the early return with constrained-provider-aware clamping:

```typescript
// src/hooks/anthropic-effort/hook.ts, line 82
// Before:
if (output.options.effort !== undefined) return

// After:
if (output.options.effort !== undefined) {
  const constrained = isConstrainedProvider(model.providerID)
  if (constrained && output.options.effort === "max") {
    const opus = isOpusModel(model.modelID)
    output.options.effort = clampVariant("max", opus, constrained)
    ;(message as { variant?: string }).variant = output.options.effort as string
    log("anthropic-effort: clamped pre-set effort max→high", {
      sessionID: input.sessionID,
      provider: model.providerID,
      model: model.modelID,
      reason: "constrained-provider",
    })
  }
  return
}
```

### Additional Improvements (optional)

1. **`model-requirements.ts`**: Change hardcoded `variant: "max"` to `variant: "high"` for opus entries in fallback chains, since `"max"` will always get clamped for constrained providers anyway.
2. **`model-capability-heuristics.ts`** line 16: Consider not including `"max"` in the opus variant list when the provider is known to be constrained (though this may be harder since heuristics are provider-agnostic).
3. **Test coverage**: Add a test case for pre-set `effort: "max"` + `github-copilot` provider to prevent regression.

### Reproduction

1. Configure `opencode.json` to use `github-copilot` provider with `claude-opus-4.6`
2. Load `oh-my-openagent` plugin
3. Trigger any agent that uses opus with `variant: "max"` in its model requirements (e.g., oracle, sisyphus)
4. The API call fails with the error above

--
Issue is found out by opencode, oh-my-openagent, Sonnet 4.6 model using Github Copilot

MoerAI:
## Summary
- Clamp pre-set `effort: "max"` to `"high"` for constrained providers like GitHub Copilot

## Problem
When `output.options.effort` is pre-set to `"max"` via session prompt params (e.g., from the ultrawork keyword-detector or session-level configuration), the anthropic-effort hook skips clamping entirely because of an early return at `if (output.options.effort !== undefined) return`. This causes GitHub Copilot's claude-opus-4.6 to receive `effort: "max"` which it rejects with `output_config.effort "max" is not supported; supported values: [low medium high]`.

## Fix
Before the early return on pre-set effort, check if the provider is constrained and the effort is `"max"`. If so, clamp it to `"high"` and override the message variant accordingly. Non-constrained providers with pre-set effort still skip as before.

## Changes
| File | Change |
|------|--------|
| `src/hooks/anthropic-effort/hook.ts` | Clamp pre-set `effort: "max"` for constrained providers before early return |

Fixes #3563

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3608"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Clamp pre-set effort "max" to "high" for constrained providers (e.g., GitHub Copilot) to prevent unsupported-effort errors and avoid failed requests. Adds a debug log to trace clamping. Fixes #3563.

- **Bug Fixes**
  - If `output.options.effort` is "max" and the provider is constrained, clamp to "high", update `message.variant`, and log the change before returning.
  - Non-constrained providers with pre-set effort remain unchanged; existing clamping still applies when effort isn’t pre-set.

<sup>Written for commit 3de1c612adf2e14e5bfd4d7619173aad936037ce. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/anthropic-effort/hook.ts
+++ b/src/hooks/anthropic-effort/hook.ts
@@ -79,10 +79,24 @@ export function createAnthropicEffortHook() {
       if (message.variant !== "max") return
       if (!isClaudeProvider(model.providerID, model.modelID)) return
       if (shouldSkipForInternalAgent(agent?.name)) return
-      if (output.options.effort !== undefined) return
-
       const opus = isOpusModel(model.modelID)
       const constrained = isConstrainedProvider(model.providerID)
+
+      // If effort was pre-set (e.g., via session prompt params), still clamp
+      // for constrained providers that reject "max".
+      if (output.options.effort !== undefined) {
+        if (constrained && output.options.effort === "max") {
+          output.options.effort = MAX_VARIANT_BY_TIER.default
+          ;(message as { variant?: string }).variant = MAX_VARIANT_BY_TIER.default
+          log("anthropic-effort: clamped pre-set effort max→high", {
+            sessionID: input.sessionID,
+            provider: model.providerID,
+            model: model.modelID,
+          })
+        }
+        return
+      }
+
       const clamped = clampVariant(message.variant, opus, constrained)
       output.options.effort = clamped
 


MoerAI:

MoerAI:
Closing — looks like upstream commit `1ae32283` ("fix(anthropic-effort): clamp variant=max for github-copilot Claude models") shipped a narrower fix in the same file that addresses #3563. Happy to reopen with a different angle if there's still a gap.

PR #3623, opened 2026-04-24
fix(chat-message): remove unreachable guard that blocked session model recovery (fixes #3561)

### LINKED ISSUE #3561: getStoredMainSessionModel recovery branch is unreachable (dead code)
## Summary

`getStoredMainSessionModel` inside the `chat.message` hook never reaches its recovery branch — one of the guard conditions can never become `false` in production. The recovery mechanism (which appears intended to restore a previously stored session model when the UI drops the selection) is dead code today.

## Location

From the shipped bundle at `dist/index.js:148028-148048`; call site at `dist/index.js:148099-148102`:

```js
function getStoredMainSessionModel(input, pluginConfig, isFirstMessage, output) {
  if (isFirstMessage) return;
  if (subagentSessions.has(input.sessionID)) return;
  if (getMainSessionID() !== input.sessionID) return;
  if (input.model) return;
  if (output.message["model"] !== undefined) return;   // ← always true
  if (hasExplicitAgentModelOverride(input.agent, pluginConfig)) return;
  return getSessionModel(input.sessionID);
}
```

## Why the guard is always satisfied

opencode triggers `chat.message` inside `createUserMessage` at `packages/opencode/src/session/prompt.ts:1234-1244`, **after** `info.model` has already been assigned at line 932:

```ts
const model = input.model ?? ag.model ?? (yield* lastModel(input.sessionID))
// …
const info: MessageV2.User = {
  // …
  model: { providerID: model.providerID, modelID: model.modelID, variant },
}
// …
yield* plugin.trigger("chat.message", { … }, { message: info, parts })
```

The three-way fallback guarantees a defined value, so `info.model` — and therefore `output.message.model`, which is the same object reference — is **always** defined when the hook runs. The guard `output.message["model"] !== undefined` short-circuits every time, and `getSessionModel(sessionID)` is never called.

## Observable consequence

When the opencode frontend drops `input.model` on a message submission, the backend falls back to `ag.model`. The recovery code here looks designed to rescue exactly that case (swap `ag.model` back to the user's last choice stored in `sessionModels`) — but because of the guard, it never runs. The user's previous selection is silently lost.

Concretely reproduced with two plain models (`anthropic/claude-opus-4-7` as agent default, `openai/gpt-5.4` selected by user): message 2 in a new session ends up using the agent default, even though `sessionModels` already has the user's gpt-5.4 choice from message 1. The same pattern applies to models routed through custom OpenAI-compatible providers.

## Suggested fix

**Option A — narrow the guard.** Only skip when the resolved model does *not* look like the agent default, or override when the stored session model differs from the agent default:

```js
const stored = getSessionModel(input.sessionID);
const current = output.message["model"];
const agDefault = getAgentDefaultModel(input.agent, pluginConfig);
if (stored && current && agDefault &&
    current.providerID === agDefault.providerID &&
    current.modelID === agDefault.modelID &&
    (stored.providerID !== agDefault.providerID || stored.modelID !== agDefault.modelID)) {
  output.message["model"] = stored;
}
```

i.e. override only when the resolved value came from the agent default **and** the stored session model is different.

**Option B — move recovery to `chat.params`** (which sees the fully resolved model) and surface a toast instead of silently accepting the drop. Doesn't fix the behavior but makes it visible.

## Related

There's an underlying opencode frontend bug where the model picker resets after the first message of a new session: https://github.com/anomalyco/opencode/issues/23666

Even after that upstream fix, the guard here is worth revisiting — the original intent (resilient session-level model persistence) is good, it just isn't actually working today.

MoerAI:
## Summary
- Remove unreachable guard that prevented session model recovery from restoring the user's previous model selection

## Problem
`getStoredMainSessionModel()` in `chat-message.ts` contains a guard `if (output.message["model"] !== undefined) return undefined` that always short-circuits because OpenCode populates `output.message.model` before triggering the `chat.message` hook. This makes the recovery branch (`getSessionModel(sessionID)`) dead code. When the UI drops the model selection on subsequent messages, the user's previously chosen model is silently lost instead of being restored.

## Fix
Removed the unreachable `output.message["model"]` guard. The remaining guards (`isFirstMessage`, `subagentSessions`, `mainSessionID`, `input.model`, `hasExplicitAgentModelOverride`) still prevent inappropriate model overrides. Now `getSessionModel()` can actually restore the stored model when the UI drops the selection.

## Changes
| File | Change |
|------|--------|
| `src/plugin/chat-message.ts` | Remove unreachable `output.message["model"]` guard in `getStoredMainSessionModel` |

Fixes #3561

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3623"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Removed an unreachable guard in `src/plugin/chat-message.ts` that prevented session model recovery, so the user’s previously selected model is restored when the UI omits a model. Existing guards still prevent unintended overrides. Fixes #3561.

<sup>Written for commit 7fa21efc2a3a36b1d957b4ad97dfd7e729fb20f7. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin/chat-message.ts
+++ b/src/plugin/chat-message.ts
@@ -81,9 +81,9 @@ function getStoredMainSessionModel(
     return undefined
   }
 
-  if (output.message["model"] !== undefined) {
-    return undefined
-  }
+  // Removed: `output.message["model"] !== undefined` guard was unreachable.
+  // OpenCode always populates output.message.model before triggering chat.message,
+  // so the guard short-circuited every time, preventing session model recovery.
 
   if (hasExplicitAgentModelOverride(input.agent, pluginConfig)) {
     return undefined


MoerAI:

acamq (approved):

PR #3626, opened 2026-04-24
fix(delegate-task): default load_skills to [] instead of throwing

darianstlex:
## Summary

The `load_skills` argument guard in `tool-argument-preparation.ts` throws when the LLM omits it from the `task()` tool call. The schema/hint already tell the model it defaults to `[]`, but the guard was strict instead of lenient — causing a hard error instead of a graceful fallback.

## Changes

| File | Change |
|------|--------|
| `src/tools/delegate-task/tool-argument-preparation.ts` | `load_skills === undefined` → silently default to `[]` instead of throwing |
| `src/tools/delegate-task/tools.test.ts` | Updated test to expect success (not error) when `load_skills` is omitted |
| `src/tools/delegate-task/tool-description.ts` | Updated tool hint to clarify that `[]` is the default |
| `src/hooks/delegate-task-retry/patterns.ts` | Removed dead `missing_load_skills` pattern that could never match |
| `src/hooks/delegate-task-retry/index.test.ts` | Removed dead test for the dead pattern |

## Risk

Minimal — this removes a throw and some dead retry logic. No new behavior, just a graceful default that should have been there from the start.

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3626"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Default `load_skills` to [] in the delegate-task tool so calls don’t fail when the model omits it or passes null.

- **Bug Fixes**
  - Clarified the tool hint to state the default and updated tests to expect success with missing or null `load_skills`.
  - Removed the dead `missing_load_skills` retry pattern and its assertion; adjusted the error-pattern test count.

<sup>Written for commit 49f858cfe5aca6a3983252bb75a6181782d112bf. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/delegate-task-retry/index.test.ts
+++ b/src/hooks/delegate-task-retry/index.test.ts
@@ -34,15 +34,6 @@ describe("sisyphus-task-retry", () => {
       expect(result?.errorType).toBe("missing_run_in_background")
     })
 
-    it("should detect load_skills missing error", () => {
-      const output = "[ERROR] Invalid arguments: 'load_skills' parameter is REQUIRED. Use load_skills=[] if no skills are needed."
-      
-      const result = detectDelegateTaskError(output)
-      
-      expect(result).not.toBeNull()
-      expect(result?.errorType).toBe("missing_load_skills")
-    })
-
     it("should detect category/subagent mutual exclusion error", () => {
       const output = "[ERROR] Invalid arguments: Provide EITHER category OR subagent_type, not both."
       
--- a/src/hooks/delegate-task-retry/index.test.ts
+++ b/src/hooks/delegate-task-retry/index.test.ts
@@ -10,11 +10,10 @@ describe("sisyphus-task-retry", () => {
     // given error patterns are defined
     // then should include all known task error types
     it("should contain all known error patterns", () => {
-      expect(DELEGATE_TASK_ERROR_PATTERNS.length).toBeGreaterThan(5)
+      expect(DELEGATE_TASK_ERROR_PATTERNS.length).toBeGreaterThan(4)
       
       const patternTexts = DELEGATE_TASK_ERROR_PATTERNS.map(p => p.pattern)
       expect(patternTexts).toContain("run_in_background")
-      expect(patternTexts).toContain("load_skills")
       expect(patternTexts).toContain("category OR subagent_type")
       expect(patternTexts).toContain("Unknown category")
       expect(patternTexts).toContain("Unknown agent")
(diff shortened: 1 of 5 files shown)


darianstlex:

code-yeongyu (changes requested):
## Review: fix(delegate-task): default load_skills to []

**Request Changes — one test inconsistency to fix.**

### Intent ✅

Making `load_skills` optional with a sensible default (`[]`) reduces friction without introducing safety issues. The description update is accurate.

### Falsy check concern ⚠️ Minor

`if (!loadSkills)` catches `null`, `undefined`, `0`, `false`, `""` — all default to `[]`. This is intentional and fine for this field since `0`/`false`/`""` are nonsensical values for a skills array. No issue here.

### Callers audit ✅

Searched all callers of `prepareDelegateTaskArgs` — all pass `load_skills` from user input directly. No internal callers pass `null` intentionally. Safe to default.

### Retry pattern removal ✅

Removing `missing_load_skills` from `DELEGATE_TASK_ERROR_PATTERNS` is correct — the error no longer exists, so the retry pattern is dead code.

### Test inconsistency ❌ Must fix

The `null` case test was not updated:

```ts
// tools.test.ts — still expects throw for null, but null now defaults to []
await expect(tool.execute({ ..., load_skills: null }, toolContext))
  .rejects.toThrow("Invalid arguments: load_skills=null is not allowed")
```

This test fails locally (`2 fail`). It should be updated to:

```ts
// then - should default to empty array
await expect(tool.execute({ ..., load_skills: null }, toolContext)).resolves.toBeDefined()
```

Similarly, `DELEGATE_TASK_ERROR_PATTERNS` count test fails because `missing_load_skills` was removed from the array but the test still counts the old number of patterns.

Please fix both failing tests before merge.

code-yeongyu:
[sisyphus-bot] Hi darianstlex. 🙏 Following up on this one after some queue cleanup.

The intent here is right (defaulting `load_skills` to `[]` instead of throwing when the LLM omits it). The 5/4 review from code-yeongyu landed two concrete asks:

1. Update the `null` case test in `tools.test.ts` so it expects `resolves.toBeDefined()` instead of `rejects.toThrow("Invalid arguments: load_skills=null is not allowed")`.
2. Update the `DELEGATE_TASK_ERROR_PATTERNS` count test in `patterns` to match the new count after removing `missing_load_skills` from the array.

The PR has been idle since then and `dev` has moved enough that the three touched files (`tool-argument-preparation.ts`, `tool-description.ts`, `tools.test.ts`) now conflict on rebase. If you can spare the time, the cleanest path forward is:

```bash
git fetch upstream
git rebase upstream/dev
# resolve the three conflicts (mostly accepting yours for the load_skills defaulting + dropping the removed pattern)
# apply the two test fixes from code-yeongyu's review
git push --force-with-lease
```

If you'd rather not pick this back up, please feel free to close and I'll add it to the small-fix backlog so another contributor can re-land the same change on top of current dev. Either way, really appreciate the original work.

PR #3718, opened 2026-04-29
fix(auto-update-checker): read loaded plugin's package.json, not the legacy flat install

ismetanin:
## Summary

`getCachedVersion()` reads the **legacy flat install** at `<CACHE_DIR>/node_modules/<pkg>/package.json` before checking the actually-loaded module, so the startup toast and `omo --version` can report a version the runtime isn't running.

OpenCode loads plugins from a per-plugin sandbox at `<CACHE_DIR>/<plugin-entry>/node_modules/<pkg>/package.json`. The two install layers drift independently because:

- The flat install is driven by `<CACHE_DIR>/package.json` (`"oh-my-opencode": "latest"`) — bun re-resolves on every install.
- The sandbox is driven by `<CACHE_DIR>/<plugin-entry>/package.json`, which gets a literal version pin baked in at first install.

## Reproduction

With `"oh-my-openagent@latest"` in the user's `opencode.json`:

| Layer | Path | Version |
|---|---|---|
| Flat install (read by current code) | `~/.cache/opencode/packages/node_modules/oh-my-opencode/package.json` | **3.17.6** |
| Sandbox (actually loaded) | `~/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/package.json` | **3.17.5** |

Startup toast announced **v3.17.6**. The loaded plugin code was **3.17.5**. The user investigated the version mismatch and noticed the lie.

## Fix

Reorder `getCachedVersion()` so the module-relative walk-up (`findPackageJsonUp(currentDir)` from `import.meta.url`) runs **first**. That always reflects the actually-loaded module. The flat-install candidates and execPath walk-up remain as fallbacks for bundled or otherwise unusual environments.

## Test plan

- [x] Added a new unit test (`prefers the loaded module's package.json over flat-install candidates`) that sets up both layers with different versions and asserts the sandbox version wins.
- [x] All 5 tests in `cached-version.test.ts` pass.
- [x] Pre-existing test failures elsewhere in `src/hooks/auto-update-checker/` (9 failures + 1 error) are unrelated to this change — they reproduce on a clean checkout of `dev` and are caused by tests reading the developer's real `~/.config/opencode/opencode.json`. Happy to file a separate issue/PR for those if useful.

## Notes

- The fallback ordering preserves all existing behavior for the `GH-3257` cases (npm users with both `oh-my-opencode` and `oh-my-openagent` flat installs); those tests continue to pass unchanged.
- No public API change.

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3718"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Fixes version reporting by reading the loaded plugin’s package.json first, not the legacy flat install. Ensures the startup toast and `omo --version` show the actual running version of `oh-my-opencode`/`oh-my-openagent`.

- **Bug Fixes**
  - Reordered `getCachedVersion()` to prefer the module-relative walk-up from `import.meta.url`, then fall back to flat-install candidates and `execPath`.
  - Added a unit test to confirm the sandboxed (loaded) version wins when versions differ.
  - Preserves existing fallbacks; no public API changes.

<sup>Written for commit 5291ee7d3df0c81f7386f851c44829d9e56852ea. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/3718?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/auto-update-checker/checker/cached-version.test.ts
+++ b/src/hooks/auto-update-checker/checker/cached-version.test.ts
@@ -4,7 +4,10 @@ import { tmpdir } from "node:os"
 import { join } from "node:path"
 
 // Hold mutable mock state so beforeEach can swap the cache root for each test.
-const mockState: { candidates: string[] } = { candidates: [] }
+const mockState: { candidates: string[]; walkUpResult: string | null } = {
+  candidates: [],
+  walkUpResult: null,
+}
 
 mock.module("../constants", () => ({
   INSTALLED_PACKAGE_JSON_CANDIDATES: new Proxy([], {
@@ -22,7 +25,7 @@ mock.module("../constants", () => ({
 }))
 
 mock.module("./package-json-locator", () => ({
-  findPackageJsonUp: () => null,
+  findPackageJsonUp: () => mockState.walkUpResult,
 }))
 
 import { getCachedVersion } from "./cached-version"
@@ -36,11 +39,13 @@ describe("getCachedVersion (GH-3257)", () => {
       join(cacheRoot, "node_modules", "oh-my-opencode", "package.json"),
       join(cacheRoot, "node_modules", "oh-my-openagent", "package.json"),
     ]
+    mockState.walkUpResult = null
   })
 
   afterEach(() => {
     rmSync(cacheRoot, { recursive: true, force: true })
     mockState.candidates = []
+    mockState.walkUpResult = null
   })
 
   it("returns the version when the package is installed under oh-my-opencode", () => {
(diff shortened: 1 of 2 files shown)


acamq (approved):

acamq:
LGTM. Thank you!

PR #3722, opened 2026-04-29
feat: support GPT-5.4/5.5 variant model IDs for Sisyphus

kuitos:
## Summary

- Tighten isGptNativeSisyphusModel so GPT-native Sisyphus matching is anchored to the actual model ID instead of any substring.
- Keep Sisyphus active for OpenCode-supported model IDs such as gpt-5.4-fast, gpt-5.5-fast, and mixed-case GPT-5.5-Pro.
- Add guard-level regression coverage confirming these model IDs do not trigger the no-Sisyphus-GPT toast or force a Hephaestus switch.

## Root cause

The guard decision belongs in the GPT-native model classifier. Suffix model IDs should be accepted as model IDs by the classifier, not handled through fallback-chain variant inheritance. This patch keeps variant handling unchanged and scopes the behavior to model detection plus no-toast/no-switch regression tests.

## Validation

- bun test src/agents/types.test.ts src/hooks/no-sisyphus-gpt/index.test.ts
- git diff --check origin/dev...HEAD

bun run typecheck still fails on pre-existing repository issues unrelated to this patch, including @opencode-ai/plugin PluginModule export errors and missing posthog-node types.

### DIFF
--- a/src/agents/types.test.ts
+++ b/src/agents/types.test.ts
@@ -23,6 +23,9 @@ describe("isGptNativeSisyphusModel", () => {
     expect(isGptNativeSisyphusModel("github-copilot/gpt-5.4")).toBe(true);
     expect(isGptNativeSisyphusModel("venice/gpt-5-4")).toBe(true);
     expect(isGptNativeSisyphusModel("openai/gpt-5.4-codex")).toBe(true);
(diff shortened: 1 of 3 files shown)


kuitos:
Closing as no-op: latest dev already supports GPT-5.4/5.5 suffix model IDs such as gpt-5.4-fast and gpt-5.5-fast in the GPT-native Sisyphus classifier.

PR #3771, opened 2026-05-03
fix(atlas): block continuation while delegated tasks are pending

tw-yshuang:
## Summary
- treat `pending` background tasks as active when Atlas decides whether to inject a continuation
- prevent Atlas from resuming a plan while a delegated subagent is still in the session-creation window
- add regression coverage for the pending-task continuation race

## Root cause
Atlas continuation only checked for `running` background tasks. When a delegated subagent had already been launched but had not yet acquired its session ID, the task remained in `pending`. That left a race window where Atlas could inject another continuation before the delegated work had actually started, which made the orchestration appear stuck or out of order.

## What changed
- updated `src/hooks/atlas/boulder-continuation-injector.ts` to treat both `pending` and `running` tasks as active background work
- added a regression test covering the pending-session-creation case in `src/hooks/atlas/boulder-continuation-injector.test.ts`

## Tests
- `bun test src/hooks/atlas/boulder-continuation-injector.test.ts src/hooks/atlas/index.test.ts`
- `bun run typecheck`

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3771"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Block Atlas continuation while delegated background tasks are still pending session creation, closing a race that could resume plans too early. This keeps orchestration in order and prevents stuck-looking runs.

- **Bug Fixes**
  - Treat both `pending` and `running` background tasks as active in `injectBoulderContinuation` (via `ACTIVE_BACKGROUND_TASK_STATUSES`).
  - Add a regression test to ensure continuation is skipped when a delegated task is still pending.

<sup>Written for commit 56e4044927038f45083cf4a5a8a45904a9087b1a. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/atlas/boulder-continuation-injector.test.ts
+++ b/src/hooks/atlas/boulder-continuation-injector.test.ts
@@ -91,6 +91,44 @@ describe("injectBoulderContinuation", () => {
     expect(sessionState.lastContinuationInjectedAt).toBe(123)
   })
 
+  test("#given a background task is still pending session creation #when injector checks again #then it still skips continuation", async () => {
+    // given
+    registerAgentName("atlas")
+    const promptAsyncMock = mock(async (_request: unknown) => undefined)
+    const messagesMock = mock(async () => ({ data: [] }))
+    const sessionState = { promptFailureCount: 1, lastContinuationInjectedAt: 456 }
+
+    const ctx = {
+      directory: "/tmp",
+      client: {
+        session: {
+          messages: messagesMock,
+          promptAsync: promptAsyncMock,
+        },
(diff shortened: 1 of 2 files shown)


tw-yshuang:

PR #2394, opened 2026-03-08
feat(hooks): bridge SessionStart and SessionEnd Claude Code hook events

jainnam-1993:
## Summary

Completes the Claude Code hook compatibility layer from **5/7 to 7/7** hook events by adding `SessionStart` and `SessionEnd` bridging.

## Problem

oh-my-opencode bridges 5 of 7 Claude Code hook events (PreToolUse, PostToolUse, UserPromptSubmit, Stop, PreCompact), but **SessionStart and SessionEnd are missing**. This breaks session context injection for users relying on Claude Code hooks — for example, [ECC's](https://github.com/affaan-m/everything-claude-code) `session-start.js` which injects previous session summaries, or custom `session-context-loader.sh` hooks.

## Solution

Bridge the two missing events using the same patterns established by the existing hook bridge:

- **`session.created`** (OpenCode) → **`SessionStart`** (Claude Code) — hook stdout is collected and injected into the session context via `ctx.client.session.prompt()` + `createInternalAgentTextPart()`
- **`session.deleted`** (OpenCode) → **`SessionEnd`** (Claude Code) — fire-and-forget execution

### Files Changed

| File | Change |
|------|--------|
| `types.ts` | Add `SessionStart`/`SessionEnd` to `ClaudeHookEvent` union, `ClaudeHooksConfig`, plus input interfaces |
| `config.ts` | Add both events to normalizer and merger `eventTypes` arrays |
| `config-loader.ts` | Add both events to `mergeDisabledHooks()` |
| `session-start.ts` (new) | `executeSessionStartHooks` (collects stdout) + `executeSessionEndHooks` (fire-and-forget) |
| `session-event-handler.ts` | Wire `session.created` → SessionStart and `session.deleted` → SessionEnd |

### Design Decisions

- **No exit-code-2 blocking** for SessionStart (matches Claude Code behavior)
- **SessionEnd is fire-and-forget** (no blocking, no injection)
- **Subagent sessions skipped** via `parentSessionId` check (matches Stop hook pattern)
- **`isHookDisabled()` respected** for both new events
- **Zero impact** on existing 5 hook event behaviors

## Verification

- `bun run typecheck` — passes (zero errors)
- `bun run build` — succeeds

---
## Summary by cubic
Bridges Claude Code `SessionStart` and `SessionEnd` hook events, completing the compatibility layer (5/7 → 7/7). On session create, hook stdout is injected into the session; on session delete, hooks run and per-session state is cleared.

- **New Features**
[...]

Issue #594, opened 2026-01-08
[Bug]: UserPromptSubmitHooks only fires once per session


zhangyang-crazy-one:
## Bug Description

The `UserPromptSubmitHooks` only triggers on the first user prompt within a session. Subsequent prompts in the same session do not trigger the hook.

## Expected Behavior

The `UserPromptSubmitHooks` should fire every time a user submits a prompt, regardless of how many prompts have been submitted in the current session.

## Actual Behavior

- **First prompt**: Hook fires correctly ✓
- **Second prompt and onwards**: Hook does NOT fire ✗

## Environment

- Plugin: oh-my-opencode
- Hook: `UserPromptSubmitHooks`

## Steps to Reproduce

1. Start a new opencode session
2. Submit a prompt → Hook fires correctly
3. Submit another prompt → Hook does NOT fire
4. Continue submitting prompts → Hook never fires again

## Additional Context

The hook appears to be registered correctly and works on the initial prompt, but something in the plugin's internal state prevents it from triggering on subsequent prompts within the same session.

This may be related to how the hook is registered or how the session state is being tracked.

code-yeongyu:
## Analysis of Issue #594: UserPromptSubmitHooks only fires once per session

### Code Investigation

After examining the codebase, I found the relevant files:
- `src/hooks/claude-code-hooks/index.ts` - Main hook factory with chat.message handler
- `src/hooks/claude-code-hooks/user-prompt-submit.ts` - Hook execution logic

### Key Findings

1. **Hook Registration**: The `chat.message` handler (lines 71-167) calls `executeUserPromptSubmitHooks` on every message submission.

2. **State Tracking**: There's a `sessionFirstMessageProcessed` Set (line 32) that tracks first messages, but it's not actually used to skip UserPromptSubmit hooks - it's just used for logging purposes (line 146).

3. **Potential Root Causes**:

   a. **Early Exit in user-prompt-submit.ts** (lines 43-45):
   ```typescript
   if (ctx.parentSessionId) {
     return { block: false, modifiedParts, messages }
   }
   ```
   If subsequent prompts have a parent session ID set, hooks will be skipped.

   b. **Tag-Based Filtering** (lines 47-52):
   ```typescript
   if (ctx.prompt.includes(USER_PROMPT_SUBMIT_TAG_OPEN) &&
       ctx.prompt.includes(USER_PROMPT_SUBMIT_TAG_CLOSE)) {
[...]

PR #3790, opened 2026-05-04
fix(hooks): do not clear sessionFirstMessageProcessed on session.idle

brooksbUWO:
# fix(hooks): do not clear sessionFirstMessageProcessed on session.idle

## Problem

SessionStart hooks fire on every user prompt instead of only at session start. This causes plugins (e.g., superpowers) that register `SessionStart` hooks to inject their context on every message, wasting tokens and cluttering the conversation.

## Root Cause

`clearSessionHookState(sessionID)` is called on every `session.idle` event (in `src/hooks/claude-code-hooks/handlers/session-event-handler.ts`). This function clears three state maps:

```typescript
function clearSessionHookState(sessionID: string) {
  sessionErrorState.delete(sessionID);
  sessionInterruptState.delete(sessionID);
  sessionFirstMessageProcessed.delete(sessionID); // <-- BUG
}
```

The first two (`sessionErrorState` and `sessionInterruptState`) are transient per-response state that should be cleared between turns. But `sessionFirstMessageProcessed` is a **session-level** flag that tracks whether the first message of a session has been processed. Clearing it on every `session.idle` causes `isFirstMessage` to always return `true`:

```typescript
// In chat-message-handler.ts:
const isFirstMessage = !sessionFirstMessageProcessed.has(input.sessionID);
sessionFirstMessageProcessed.add(input.sessionID);
```

Since `session.idle` fires after every model response, `sessionFirstMessageProcessed` is cleared before the next user prompt, making `isFirstMessage` always `true`. This causes `SessionStart` hooks to execute on every prompt instead of only the first one.

## Fix

Remove `sessionFirstMessageProcessed.delete(sessionID)` from `clearSessionHookState`. The `sessionFirstMessageProcessed` set should only be cleared when the session is fully deleted (in `clearAllSessionHookState`), not on every idle event.

### `src/hooks/claude-code-hooks/session-hook-state.ts`

```diff
 function clearSessionHookState(sessionID: string) {
   sessionErrorState.delete(sessionID);
   sessionInterruptState.delete(sessionID);
-  sessionFirstMessageProcessed.delete(sessionID);
+  // sessionFirstMessageProcessed should NOT be cleared on idle -
+  // it tracks whether the first message of a session has been processed.
+  // Clearing it on idle makes isFirstMessage always return true,
+  // causing SessionStart hooks to fire on every prompt instead of once.
 }
```

`clearAllSessionHookState()` (called on `session.deleted` and `dispose`) remains unchanged - it still clears `sessionFirstMessageProcessed` via `.clear()`, which is correct because the session is truly ending.

## Why This Is Correct

| State Map | Purpose | Cleared on idle? | Why? |
|-----------|---------|-------------------|------|
| `sessionErrorState` | Tracks per-response errors | Yes | Errors are transient between turns |
| `sessionInterruptState` | Tracks per-response interrupts | Yes | Interrupts are transient between turns |
| `sessionFirstMessageProcessed` | Tracks whether first message was processed | **No** | First-message status persists for the entire session |

The `session.idle` event fires after every model response. Clearing error and interrupt state between turns is correct - those are per-response concerns. But `sessionFirstMessageProcessed` is a session-level concern: once the first message has been processed, it should remain processed for the rest of the session.

## Related Issues

- #2394 - Added SessionStart/SessionEnd hook bridging (the feature this fix depends on)
- #594 - Opposite problem: hooks only firing once (fixed in #1618 by changing tag deduplication logic)

## Testing

1. Configure a plugin with a `SessionStart` hook (e.g., superpowers)
2. Start a new session - hook should fire on the first prompt
3. Send a second prompt - hook should NOT fire
4. Send additional prompts - hook should NOT fire
5. Start a new session - hook should fire again on the first prompt

Without this fix, step 3 and 4 would incorrectly fire the SessionStart hook.

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3790"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Prevents `SessionStart` hooks from firing on every prompt by preserving the session-level first-message flag across turns. Hooks now run only once per session, and optional `oh-my-opencode-*` binaries are bumped to 3.17.13.

- **Bug Fixes**
  - Stop deleting `sessionFirstMessageProcessed` in `clearSessionHookState` (called on `session.idle`); only clear it in `clearAllSessionHookState` when the session ends.

<sup>Written for commit 1a43c9d120818a6a3f0fc932cebb5feb2fda46d7. Summary will update on new commits.</sup>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -30,17 +30,17 @@
         "zod": "^4.3.0",
       },
       "optionalDependencies": {
-        "oh-my-opencode-darwin-arm64": "3.17.12",
-        "oh-my-opencode-darwin-x64": "3.17.12",
-        "oh-my-opencode-darwin-x64-baseline": "3.17.12",
-        "oh-my-opencode-linux-arm64": "3.17.12",
-        "oh-my-opencode-linux-arm64-musl": "3.17.12",
-        "oh-my-opencode-linux-x64": "3.17.12",
-        "oh-my-opencode-linux-x64-baseline": "3.17.12",
-        "oh-my-opencode-linux-x64-musl": "3.17.12",
-        "oh-my-opencode-linux-x64-musl-baseline": "3.17.12",
-        "oh-my-opencode-windows-x64": "3.17.12",
-        "oh-my-opencode-windows-x64-baseline": "3.17.12",
+        "oh-my-opencode-darwin-arm64": "3.17.13",
+        "oh-my-opencode-darwin-x64": "3.17.13",
+        "oh-my-opencode-darwin-x64-baseline": "3.17.13",
+        "oh-my-opencode-linux-arm64": "3.17.13",
+        "oh-my-opencode-linux-arm64-musl": "3.17.13",
+        "oh-my-opencode-linux-x64": "3.17.13",
+        "oh-my-opencode-linux-x64-baseline": "3.17.13",
+        "oh-my-opencode-linux-x64-musl": "3.17.13",
+        "oh-my-opencode-linux-x64-musl-baseline": "3.17.13",
+        "oh-my-opencode-windows-x64": "3.17.13",
+        "oh-my-opencode-windows-x64-baseline": "3.17.13",
       },
       "peerDependencies": {
         "zod": "^4.0.0",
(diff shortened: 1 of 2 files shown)


brooksbUWO:

code-yeongyu (approved):
## Review: fix(hooks): do not clear sessionFirstMessageProcessed on session.idle

**LGTM — approve with one minor suggestion.**

### Fix correctness ✅

The logic is correct. `clearSessionHookState` is called from two places in `session-event-handler.ts`:

1. **`session.deleted`** (line 47) — correct to clear everything including `sessionFirstMessageProcessed`, and `clearAllSessionHookState` at `session.deleted` does exactly that.
2. **`session.idle`** (line 126) — fires after **every model response**. The bug: clearing `sessionFirstMessageProcessed` here made `isFirstMessage` always return `true` in `chat-message-handler.ts` (line 75), causing `SessionStart` hooks to fire repeatedly instead of once per session.

The fix is minimal, targeted, and correct.

### Memory leak risk ✅ No issue

`sessionFirstMessageProcessed` is a `Set<string>`. It accumulates one entry per session ID. Proper cleanup happens in `clearAllSessionHookState()` (called on `session.deleted` path via `disposeSessionEventHandler`) and `clearSessionHookState` is also called on `session.deleted` event directly. No leak.

### Call site audit ✅

All callers of `clearSessionHookState` verified:
- `session.deleted` (line 47): correct — full cleanup expected
- `session.idle` (line 126): correct — should NOT clear `sessionFirstMessageProcessed`
- `clearAllSessionHookState` in `disposeSessionEventHandler`: correct — full teardown, clears everything

No other callers. No regression risk.

### Test coverage ⚠️ Missing

The existing `session-event-handler.test.ts` doesn't test `sessionFirstMessageProcessed` persistence across idle events. Suggest adding:

```ts
test('sessionFirstMessageProcessed persists across session.idle but clears on session.deleted', async () => {
  // fire session.idle twice — sessionFirstMessageProcessed should NOT be cleared
  await handler({ event: { type: 'session.idle', properties: { sessionID: 'ses_x' } } })
  expect(sessionFirstMessageProcessed.has('ses_x')).toBe(false) // never set here; set by chat handler
  
  // The real guard: after session.deleted, it IS cleared
  sessionFirstMessageProcessed.add('ses_x')
  await handler({ event: { type: 'session.deleted', properties: { info: { id: 'ses_x' } } } })
  expect(sessionFirstMessageProcessed.has('ses_x')).toBe(false)
})
```

Not a blocker — the fix is clearly correct — but worth adding to prevent regression.

code-yeongyu (approved):

code-yeongyu:
## Review: fix(hooks): do not clear sessionFirstMessageProcessed on session.idle

**LGTM. Approving.**

### Fix correctness ✅

Correct. `clearSessionHookState` is called from two places in `session-event-handler.ts`:

1. **`session.deleted`** — correct to clear everything; `clearAllSessionHookState` handles full teardown
2. **`session.idle`** — fires after **every model response**. The bug: clearing `sessionFirstMessageProcessed` here made `isFirstMessage` always `true` in `chat-message-handler.ts` (line 75), causing `SessionStart` hooks to fire on every prompt instead of once per session.

Fix is minimal and targeted.

### Memory leak risk ✅ None

`sessionFirstMessageProcessed` is a `Set<string>`, one entry per session. Cleanup happens correctly in `clearAllSessionHookState()` (called on dispose + `session.deleted`). No accumulation issue.

### Call site audit ✅ All clear

- `session.deleted` (line 47): full cleanup — correct
- `session.idle` (line 126): transient state only — correct after this fix
- `disposeSessionEventHandler`: full teardown — correct

No other callers. No regression risk.

### Test coverage ⚠️ Minor gap

Existing tests don't assert that `sessionFirstMessageProcessed` persists across idle. Suggest adding a regression test:

```ts
test('sessionFirstMessageProcessed is NOT cleared on session.idle but IS cleared on session.deleted', async () => {
  sessionFirstMessageProcessed.add('ses_x')
  await handler({ event: { type: 'session.idle', properties: { sessionID: 'ses_x' } } })
  expect(sessionFirstMessageProcessed.has('ses_x')).toBe(true) // must survive idle

  await handler({ event: { type: 'session.deleted', properties: { info: { id: 'ses_x' } } } })
  expect(sessionFirstMessageProcessed.has('ses_x')).toBe(false) // cleared on delete
})
```

Not blocking — the fix is clearly correct. Just worth adding to prevent future regression.

code-yeongyu (changes requested):
## Updated Review: Request Changes

After deeper analysis via parallel opencode review sessions, I need to walk back my initial LGTM. The PR fixes the idle regression correctly but is **incomplete**.

### The actual bug introduced

`session.deleted` still calls `clearSessionHookState(sessionInfo.id)` (line 47 of `session-event-handler.ts`), but after this PR, `clearSessionHookState` no longer removes the entry from `sessionFirstMessageProcessed`.

**Lifecycle after this PR:**
1. `chat.message` — adds sessionID to `sessionFirstMessageProcessed` ✅
2. `session.idle` — does NOT clear it (correct fix) ✅
3. `session.deleted` — calls `clearSessionHookState`, which now skips `sessionFirstMessageProcessed` ❌
4. Plugin dispose — `clearAllSessionHookState()` clears everything ✅

So deleted sessions **leak their ID in the Set** until plugin unload/process restart.

### Impact

- One retained string per deleted session
- Linear/unbounded accumulation in long-running or session-churny processes (e.g. background agents, CI runners)
- QA simulation: 100 create/delete cycles → Set size 100 instead of 0

### Recommended minimal fix

Option A — separate function for delete path:

```ts
// In session-hook-state.ts
export function clearDeletedSessionHookState(sessionID: string): void {
  clearSessionHookState(sessionID)
  sessionFirstMessageProcessed.delete(sessionID) // also clear on true session end
}
```

Then call `clearDeletedSessionHookState` in the `session.deleted` branch instead of `clearSessionHookState`.

Option B — pass a flag:

```ts
export function clearSessionHookState(sessionID: string, includeFirstMessage = false): void {
  sessionErrorState.delete(sessionID)
  sessionInterruptState.delete(sessionID)
  if (includeFirstMessage) {
    sessionFirstMessageProcessed.delete(sessionID)
  }
}
```

### Test to add

```ts
test('sessionFirstMessageProcessed persists across idle but is cleared on delete', async () => {
  sessionFirstMessageProcessed.add('ses_x')
  await handler({ event: { type: 'session.idle', properties: { sessionID: 'ses_x' } } })
  expect(sessionFirstMessageProcessed.has('ses_x')).toBe(true) // survives idle

  await handler({ event: { type: 'session.deleted', properties: { info: { id: 'ses_x' } } } })
  expect(sessionFirstMessageProcessed.has('ses_x')).toBe(false) // cleared on delete
})
```

Apologies for the initial approval — the idle fix is correct, the delete-path omission is the gap.

code-yeongyu:
Superseded by #4049 (merged into `dev`). The actual one-line fix in `session-hook-state.ts` plus the explanatory comment landed; the unrelated `bun.lock` commit was dropped during rebase. Thanks @leijc.

code-yeongyu:
Closing as superseded by #4049 (merged).

PR #3825, opened 2026-05-07
fix(delegate-task): harden child-session first-prompt fallback recovery

tw-yshuang:
## Summary
- fix delegated child-session fallback recovery when the first prompt fails before session history is persisted
- introduce a shared bootstrap context so background and sync delegated flows register retry payloads and fallback metadata before first prompt dispatch
- add cleanup and regression coverage for exhaustion, non-retryable failures, isolation, and terminal-path cleanup

## Background
In the delegated child-session flow, a subagent session can fail on the very first `promptAsync` call because of a provider or model error.

Before this change, that path had two related problems:
1. background child-session fallback context registration could race behind first-prompt dispatch
2. runtime fallback depended on `session.messages` to recover the last user prompt, so empty-history first-prompt failures could not retry

That left delegated child sessions stuck in early failure without correctly advancing to fallback models.

## First-prompt early-failure flow after this change
1. A delegated child session is created from the parent session
2. Before the first prompt is sent, the system registers delegated bootstrap context containing:
   - retry prompt parts
   - fallback chain
   - category metadata
3. The first prompt is dispatched
4. If the first prompt fails with a retryable error:
   - background and sync delegated paths enter fallback handling
   - runtime fallback first tries persisted `session.messages`
   - if history is still empty and the session is a delegated child session, it falls back to the bootstrap retry parts
5. Fallback models are attempted in order with existing exhaustion semantics preserved
6. On retry handoff, completion, cancellation, crash/error, session deletion, shutdown, or launch-time terminal interrupt, the session-scoped bootstrap and fallback metadata are cleared

```mermaid
flowchart TD
    A[Parent session delegates child task] --> B[Create child session]
    B --> C[Register delegated bootstrap context before first prompt]
    C --> D[Send first prompt]

    D -->|Success| E[Continue normal background or sync flow]
    D -->|Failure| F{Retryable error?}

    F -->|No| G[Enter terminal path]
    F -->|Yes| H{session.messages has user parts?}

    H -->|Yes| I[Use persisted user parts for retry]
    H -->|No| J{Delegated bootstrap exists?}

    J -->|No| G
    J -->|Yes| K[Use bootstrap retry parts]

    I --> L[Pick next fallback model]
    K --> L
    L --> M{Fallback chain exhausted?}

    M -->|No| N[Retry with next fallback model]
    M -->|Yes| G

    N --> O{Retry succeeds?}
    O -->|Yes| P[Continue task execution]
    O -->|No| F

    E --> Q[Terminal cleanup]
    P --> Q
    G --> Q

    Q --> R[Clear bootstrap prompt context]
    Q --> S[Clear fallback-chain registration]
    Q --> T[Clear category registration]
```

### Shared delegated bootstrap contract
- add a shared bootstrap state to preserve delegated child-session retry parts before first prompt dispatch
- register fallback chain and category context before the first prompt in both background and sync delegated paths

### Background path atomic registration
- move background fallback-context registration into the launch path itself
- remove the dependency on delayed polling or backfill to make critical fallback metadata available

### Runtime fallback support for empty-history retry
- preserve the existing history-first retry path
- fall back to delegated bootstrap retry parts only when the session is a delegated child session and persisted history is still empty

### Sync path alignment and cleanup
- align sync delegated execution with the shared bootstrap contract
- ensure retry-created sessions and terminal paths clear bootstrap and fallback context so session-scoped state does not leak

### Regression coverage
- background first-prompt retryable failure
- sync first-prompt fallback ordering
- delegated empty-history runtime retry
- exhaustion behavior
- non-retryable failures do not retry
- concurrent delegated session isolation
- background and sync terminal cleanup

## Affected areas
- `src/tools/delegate-task/background-task.ts`
- `src/tools/delegate-task/sync-task.ts`
- `src/tools/delegate-task/sync-prompt-sender.ts`
- `src/hooks/runtime-fallback/auto-retry.ts`
- `src/hooks/runtime-fallback/last-user-retry-parts.ts`
- `src/shared/delegated-child-session-bootstrap.ts`

## Verification
```bash
bun test src/features/background-agent/manager.test.ts src/features/background-agent/fallback-retry-handler.test.ts src/features/background-agent/spawner.test.ts src/hooks/runtime-fallback/index.test.ts src/tools/delegate-task/sync-task.test.ts
bun run typecheck
```

- `256 pass, 0 fail`
- typecheck passed

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3825"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Fixes delegated child-session fallback when the first prompt fails before any history exists. Captures the first-prompt retry payload at launch and cleans up on every terminal path to keep state consistent and isolated.

- **Bug Fixes**
  - Introduced `shared/delegated-child-session-bootstrap` to register retry parts, fallback chain, and category before the first prompt in both background and sync flows.
  - Background manager registers the bootstrap at launch and centralizes cleanup (clears bootstrap, session fallback chain, and category) on completion, cancellation, retry handoff, deletion, shutdown, and terminal launch failures; preserves late wiring when the parent aborts early.
  - Runtime auto-retry now passes `sessionID` to last-user lookup and uses bootstrap retry parts when history is empty; does not invent payloads if no bootstrap exists.
  - Sync path aligns with the bootstrap contract, injects the built prompt text into `sendSyncPrompt`, and clears per-session bootstrap/fallback context on retries and finish; concurrent sessions remain isolated.
  - Merged latest `dev` and resolved manager/runtime-fallback conflicts while keeping delegated bootstrap wiring and upstream poll-recovery; expanded tests for empty-history retries, terminal launch cleanup, scheduled-retry cleanup, cancellation/completion cleanup, isolation, exhaustion, and non-retryable errors.

<sup>Written for commit 12f5233877c8c01daaba48187efc388493d83666. Summary will update on new commits.</sup>


tw-yshuang:
Follow-up fix pushed in `291eeed8`.

This keeps the new manager-side bootstrap registration as the primary path, but restores a compatibility fallback for the background delegated path when the parent call aborts before the child session id resolves.

Why this follow-up was needed:
- the new architecture intentionally moved delegated bootstrap/fallback registration into the earlier manager-side launch path
- however, CI still covered an older but still valid contract: once a background delegated task has been successfully launched, a parent-side abort should not prevent the child session from finishing its late wiring if the session id resolves shortly afterward

What this follow-up does:
- preserves the new pre-dispatch bootstrap architecture
- restores late session wiring only as a compatibility fallback in `background-task.ts`
- late-resolved background child sessions still receive:
  - delegated bootstrap context
  - fallback-chain registration
  - category registration

This avoids regressing the stronger new design while preserving the older tool-boundary behavior that existing CI still expects.

Re-verified locally with:
```bash
bun test src/tools/delegate-task/background-task.test.ts
bun test src/features/background-agent/manager.test.ts src/hooks/runtime-fallback/index.test.ts src/tools/delegate-task/sync-task.test.ts
bun run typecheck
```

All passed locally.

tw-yshuang:
Quick update after the last round of conflict resolution:

- I resolved the previous merge conflicts in `src/features/background-agent/manager.ts` and `src/features/background-agent/manager.test.ts`
- I re-ran the affected local verification after that merge resolution:

```bash
bun test src/features/background-agent/manager.test.ts src/features/background-agent/fallback-retry-handler.test.ts src/features/background-agent/spawner.test.ts src/hooks/runtime-fallback/index.test.ts src/tools/delegate-task/background-task.test.ts src/tools/delegate-task/sync-task.test.ts
bun run typecheck
```

Those checks passed locally (`275 pass, 0 fail`, plus clean typecheck), and the latest PR CI for the current head is green as well.

One more thing changed afterward: `dev` moved again after I resolved the earlier conflicts. I re-checked mergeability against the latest `origin/dev`, and the previous `manager.ts` / `manager.test.ts` conflicts are gone, but there is now a fresh merge conflict in:

- `src/tools/delegate-task/sync-task.test.ts`

So at this point:
- current PR head: CI is green
- current mergeability against the latest `dev`: needs one new conflict resolution in `sync-task.test.ts`

code-yeongyu:
Reverted on dev via #4044 — full reasoning there.

Short version: after merge, the PR's own `delegated child-session empty-history fallback retries with captured bootstrap prompt` test fails on a clean root `bun test --timeout 30000` run (6828/1 → revert restores 6823/0). The race this targets (delegated child session first `promptAsync` failure with empty persisted history) is narrow and partly handled by the existing persisted-history fallback path, so keeping `dev` green and reopening after the test is stabilized is the cleaner path.

To re-land:
1. Rebase onto current `dev` (which now includes the shared promptAsync gate, ralph-loop display-name fix, session-recovery persistent dedupe, tool-pair-validator subagent skip, and thinking_block_modified recovery).
2. Drive the failing test red/green explicitly against that base and confirm root `bun test --timeout 30000` is 0 fail.
3. Reopen as a fresh PR.

tw-yshuang:
Follow-up note: this PR was already reverted on dev via #4044, and the underlying delegated child-session first-prompt / empty-history fallback problem has since been addressed more completely by #4074.

The relevant upstream fix is in #4074, especially commit ba648685: fix(runtime-fallback): carry delegated system and tools through bootstrap retry. That implementation reuses the existing delegated child-session bootstrap state instead of adding a separate runtime-fallback cache, and it preserves the full delegated retry context:

- retry parts
- system prompt
- tool gates
- delegated fallback/category context

Because #4074 covers the same failure mode with a cleaner and broader upstream implementation, I am not going to re-land the older #3825 approach. Any further validation should be based on current dev after #4074 rather than on this reverted branch.

PR #3841, opened 2026-05-07
fix(tmux): restore pane spawning for subagents

Momentum96:
## Summary
- Start background subagent prompts before tmux setup so session readiness is not blocked by pane creation.
- Prefer real tmux when a valid `TMUX` session environment exists, even if `CMUX_SOCKET_PATH` is also present.
- Treat `busy` sessions as attachable so active background subagents can spawn panes instead of timing out.

## Testing
- `bun test src/features/background-agent/manager.test.ts src/features/background-agent/spawner.test.ts src/features/tmux-subagent/attachable-session-status.test.ts src/shared/tmux/runner.test.ts src/features/team-mode/team-layout-tmux/layout.test.ts`
- `bun run typecheck`
- `bun run build`
- Manual QA: verified Team Mode creates pane `%12` with `opencode attach`.
- Manual QA: verified ordinary background subagent creates pane `%13` with `opencode attach`.

### DIFF
--- a/src/features/background-agent/manager.test.ts
+++ b/src/features/background-agent/manager.test.ts
@@ -322,6 +322,82 @@ function createToastRemoveTaskTracker(): { removeTaskCalls: string[]; resetToast
   }
 }
 
+describe("BackgroundManager tmux callback ordering", () => {
+  test("starts promptAsync before a blocking tmux callback resolves", async () => {
+    //#given
+    const events: string[] = []
+    let resolveTmuxCallback: () => void = () => {}
+    const tmuxCallbackPromise = new Promise<void>((resolve) => {
+      resolveTmuxCallback = resolve
+    })
+
+    const client = {
+      session: {
+        get: async () => {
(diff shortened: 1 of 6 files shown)


code-yeongyu:
Triage note: rebase onto current `dev` conflicts in `src/shared/tmux/runner.ts`. Recent tmux changes that touched this area: 8dcbccf06 (inject pane action dependencies), and changes in #4032/#4014 (tool-pair-validator skip background sessions).

Three sub-changes here:
1. Start background subagent prompts before tmux setup (race fix).
2. Prefer real tmux over CMUX_SOCKET when both present.
3. Treat `busy` sessions as attachable.

All three remain valid concerns on current dev. To re-land:
1. Rebase onto current dev.
2. Re-thread the runner detection logic against 8dcbccf06 (tmux pane action injection).
3. Confirm `bun test --timeout 30000` is 0 fail and that the manual tmux QA from this PR still reproduces a green pane.

Leaving open.

PR #3875, opened 2026-05-08
fix(config): align OPENCODE_CONFIG_DIR with additive OpenCode semantics

### LINKED ISSUE #3846: OPENCODE_CONFIG_DIR is treated as an exclusive user config root, unlike OpenCode's additive directory semantics
## Summary

When `OPENCODE_CONFIG_DIR` is set, OMO appears to treat it as the **only** user-level OpenCode config root.

This differs from OpenCode's current documented/runtime behavior, where `OPENCODE_CONFIG_DIR` is treated as an additional custom directory / override layer rather than a full replacement for the default global config path.

As a result, if a host app sets `OPENCODE_CONFIG_DIR` to inject its own plugin/config directory, OMO can no longer see user-level OMO config that still lives under the default OpenCode global path.

## What I observed

In my case, a host app (CodeNomad) launches OpenCode with `OPENCODE_CONFIG_DIR` pointing at its packaged config directory so that OpenCode auto-loads a bundled bridge plugin.

OpenCode itself still loads correctly, but OMO loses access to user-level config such as:

- `~/.config/opencode/oh-my-openagent.json` / `.jsonc`
- user-level OMO agents / skills / commands that are resolved via OMO's OpenCode config dir helpers

The visible symptom was that the OMO agent set shrank when `OPENCODE_CONFIG_DIR` was set, because OMO could no longer find the user's existing OMO config/model mappings from the default global OpenCode config path.

## Why this looks like an OMO-side incompatibility

OpenCode docs currently describe `OPENCODE_CONFIG_DIR` as a custom directory loaded **after** global config / `.opencode` directories.

In OMO, however, the config-dir helper uses exclusive selection:

```ts
function getCliConfigDir(): string {
  const envConfigDir = process.env.OPENCODE_CONFIG_DIR?.trim()
  if (envConfigDir) {
    return resolveConfigPath(envConfigDir)
  }

  const xdgConfig = process.env.XDG_CONFIG_HOME || join(homedir(), ".config")
  return resolveConfigPath(join(xdgConfig, "opencode"))
}
```

This means:

- if `OPENCODE_CONFIG_DIR` is set, OMO returns **only that path**
- it does **not** also consider the default global OpenCode path

## Downstream impact inside OMO

From reading the source, this resolved directory is then reused by downstream OMO loaders, for example:

- plugin config loading (`oh-my-openagent.json/jsonc`)
- OpenCode config-based agent loading
- command / skill directory resolution
- other user-level OpenCode/OMO assets that depend on `getOpenCodeConfigDir()`

So the incompatibility is not just at one entrypoint; it propagates through multiple user-level loading paths.

## Expected behavior

I would expect OMO to stay compatible with OpenCode's current additive directory semantics.

Concretely, when `OPENCODE_CONFIG_DIR` is set, OMO should probably either:

1. search both:
   - the default global OpenCode config dir (for example `~/.config/opencode`)
   - the custom `OPENCODE_CONFIG_DIR`
2. or otherwise adopt the same effective directory search semantics OpenCode itself uses for user-level config discovery

Project-level `.opencode` discovery appears to be a separate path and is not the main problem here.

## Repro outline

1. Put working OMO user config under the default global OpenCode config dir
2. Start OpenCode normally -> OMO sees the expected config/agents
3. Start OpenCode via a host app that sets `OPENCODE_CONFIG_DIR` to another directory
4. OMO now resolves only that directory and stops seeing the user-level config from the default global path

## Question

Would you consider aligning OMO's user-level OpenCode config discovery with OpenCode's additive `OPENCODE_CONFIG_DIR` semantics?

If useful, I can also provide a smaller isolated reproduction focused only on OMO + OpenCode without the host app in the middle.

jollyxenon:
## Summary
- keep the default global OpenCode config directory visible even when `OPENCODE_CONFIG_DIR` is set, matching OpenCode's additive discovery semantics
- load global commands, skills, agents, and oh-my-openagent user config from both the default and custom OpenCode config roots with custom-path precedence
- add regression coverage for shared config discovery, command discovery, global agent loading, and plugin config layering

## Verification
- `bun test src/shared/opencode-config-dir.test.ts src/tools/slashcommand/command-discovery.test.ts src/features/claude-code-agent-loader/loader.test.ts src/plugin-config.test.ts`
- `bun run typecheck`
- `bun run build`

## Issue
- fixes #3846

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3875"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Make `OPENCODE_CONFIG_DIR` additive for OpenCode so commands, skills, agents, and `oh-my-openagent` user config load from both the custom path and the default XDG `opencode` directory, with custom first. Fixes #3846 by matching OpenCode’s discovery semantics.

- **Bug Fixes**
  - Added `getOpenCodeConfigDirs()` and updated command/skill helpers to expand both roots (and their parent profile dirs); agent loader now reads both and de-dupes by name.
  - Merged user plugin config from both roots: migrated legacy names, resolved `agent_definitions` per root, merged `git_master` across user roots and ancestors, and kept `mcp_env_allowlist` user-scoped.
  - Expanded tests to cover additive discovery and default-user-config visibility when `OPENCODE_CONFIG_DIR` is set; stabilized path assertions with `resolve` and isolated env via `XDG_CONFIG_HOME`.

<sup>Written for commit ecd9d2a0d8e890a0961166e34d3b5e8254930569. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/claude-code-agent-loader/loader.test.ts
+++ b/src/features/claude-code-agent-loader/loader.test.ts
@@ -55,12 +55,26 @@ const NO_FRONTMATTER_AGENT = `Just a prompt with no frontmatter.`;
 
 describe("claude-code-agent-loader", () => {
   const dirs: string[] = [];
+  const originalEnv = {
+    OPENCODE_CONFIG_DIR: process.env.OPENCODE_CONFIG_DIR,
+    XDG_CONFIG_HOME: process.env.XDG_CONFIG_HOME,
+  }
 
   afterEach(() => {
     for (const dir of dirs) {
       rmSync(dir, { recursive: true, force: true });
     }
     dirs.length = 0;
+    if (originalEnv.OPENCODE_CONFIG_DIR === undefined) {
+      delete process.env.OPENCODE_CONFIG_DIR
+    } else {
+      process.env.OPENCODE_CONFIG_DIR = originalEnv.OPENCODE_CONFIG_DIR
+    }
+    if (originalEnv.XDG_CONFIG_HOME === undefined) {
+      delete process.env.XDG_CONFIG_HOME
+    } else {
+      process.env.XDG_CONFIG_HOME = originalEnv.XDG_CONFIG_HOME
+    }
   });
 
   function trackDir(dir: string): string {
@@ -203,6 +217,29 @@ describe("claude-code-agent-loader", () => {
       process.env.OPENCODE_CONFIG_DIR = root
       const result = loadOpencodeGlobalAgents()
       expect(result).toEqual({})
+    })
+
+    test("loads agents from both the custom and default opencode config directories", () => {
+      const root = trackDir(mkdtempSync(join(tmpdir(), "agent-loader-opencode-global-")))
+      const defaultAgentsDir = join(root, "xdg", "opencode", "agents")
+      const customAgentsDir = join(root, "custom-opencode", "agents")
+
+      mkdirSync(defaultAgentsDir, { recursive: true })
+      mkdirSync(customAgentsDir, { recursive: true })
+
+      writeFileSync(join(defaultAgentsDir, "default-agent.md"), BASIC_AGENT, "utf-8")
+      writeFileSync(
+        join(customAgentsDir, "custom-agent.md"),
+        `---\nname: custom-agent\ndescription: Custom agent\n---\nFrom custom config.`,
+        "utf-8",
+      )
+
+      process.env.XDG_CONFIG_HOME = join(root, "xdg")
+      process.env.OPENCODE_CONFIG_DIR = join(root, "custom-opencode")
+
+      const result = loadOpencodeGlobalAgents()
+
+      expect(Object.keys(result).sort()).toEqual(["custom-agent", "test-agent"])
     })
   })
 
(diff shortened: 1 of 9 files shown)


jollyxenon:

jollyxenon:
Thanks — both comments were valid, so I applied a follow-up patch.

### What changed
- Cached the `getCliCustomConfigDir()` result inside `getOpenCodeConfigDirs()` so we only resolve the custom path once and no longer need the non-null assertion.
- Restored `CLAUDE_CONFIG_DIR` in `src/features/claude-code-agent-loader/loader.test.ts` so env mutations from the `loadUserAgents` test cannot leak into later tests.

### Verification
- `bun test src/shared/opencode-config-dir.test.ts src/features/claude-code-agent-loader/loader.test.ts`
- `bun run typecheck`

jollyxenon:
The test failed because the result after my modification is exactly what the test is checking, and the test has not been adapted to the new behavior where the default global directory also participates. I'm going to modify the test itself.

code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED

code-yeongyu (approved):
[sisyphus-bot] APPROVED. jollyxenon config fix aligning OPENCODE_CONFIG_DIR handling with OpenCode's additive (not replacing) config-walk semantics. 273/-75 across 9 files with substantial test coverage (`stabilize opencode command dir assertions across environments`). Cubic approved + CI green. Merging.

Issue #3774, opened 2026-05-03
bug: call_omo_agent can time out after subagent shows completed


Yeachan-Heo:
## Summary

Discord feedback reports that OMO background delegation can complete successfully in the subagent UI while `call_omo_agent` still reports a timeout. This makes background tasks unreliable, especially for users who want OMO Slim / non-Sisyphus workflows where they keep their own planning/implementation workflow but rely on OMO for specialized background agents.

## Source feedback

From UltraWorkers `#setup-share` on 2026-05-03:

> If I launch Oracle there's a 50/50 or greater chance it will complete (shows completed in the subagent view) but `call_omo_agent` still shows it as timed out. I also cannot reliably use the background tools if I'm not using Sisyphus as my main agent. I bring my own workflows for implementation, so I don't need the agent, but I need the background delegation tools. OMO Slim looks to get out of the way of workflow and just handle orchestration and specialized agents.

Related earlier feedback in the same thread:

> Why I'm looking is stability - long-standing OMO bugs like bg tasks flaking out are still outstanding.

## Expected behavior

If a delegated subagent completes, `call_omo_agent` should observe the completed state and return the result or a clear completion artifact reference. It should not report a timeout after the subagent UI already marks the run completed.

## Current behavior

A background Oracle/subagent run may show as completed in the subagent view, but the caller still receives a timeout from `call_omo_agent`.

## Why this matters

This is a workflow-contract bug for OMO Slim / manual-workflow users:

- users may not want Sisyphus/Atlas/Hephaestus as the primary workflow driver
- they still need reliable background delegation to Oracle/Explore/Librarian/etc.
- completed subagent state and caller timeout disagree, so operators cannot trust background task status

## Suggested investigation

Check the completion/result propagation path between:

1. subagent lifecycle state shown in the UI
2. background task registry / task id state
3. `call_omo_agent` timeout waiter
4. result collection / final status classification

Likely failure classes:

- completion event is written but not observed by the caller waiter
- timeout clock is not canceled when the subagent reaches terminal state
- result artifact is missing/late, so caller treats completed-without-result as timeout
- non-Sisyphus/default Build-agent workflow does not initialize the same background delegation hooks/state

## Acceptance criteria

- Reproduce with Oracle or another OMO subagent where the UI marks completion.
- `call_omo_agent` returns a completed result when the subagent reaches a terminal completed state before timeout.
[...]

Issue #1579, opened 2026-02-07
[Feature]: Allow Sisyphus to spawn Hephaestus and Atlas as subagents via call_omo_agent


sdh2222:
## Problem

`ALLOWED_AGENTS` in `src/tools/call-omo-agent/constants.ts` is hardcoded to `["explore", "librarian"]`, which prevents Sisyphus from delegating tasks to Hephaestus and Atlas.

Both agents are properly registered in `agentSources` (src/agents/utils.ts) but blocked at the tool validation layer:

```typescript
// Line 48202 in dist/index.js
var ALLOWED_AGENTS = ["explore", "librarian"];

// Line 48243 - validation rejects anything not in the list
if (![...ALLOWED_AGENTS].some((name) => name.toLowerCase() === args.subagent_type.toLowerCase())) {
    return `Error: Invalid agent type "${args.subagent_type}". Only ${ALLOWED_AGENTS.join(", ")} are allowed.`;
}
```

## Use Case

As a Sisyphus user, I want to delegate autonomous deep work to Hephaestus (GPT-5.3 Codex) without switching agents in the UI. For example:
- Security audits on smart contracts
- Complex refactoring tasks that benefit from goal-oriented autonomous execution
- End-to-end feature implementation

Similarly, Atlas delegation enables complex multi-step orchestration as a subagent.

## Related Issues

- #957 — ALLOWED_AGENTS whitelist blocks subagents (root cause)
- #1462 — Band-aid fix that added metis/momus but not hephaestus/atlas
- #1417 — Hephaestus disappearing

### Option A: Dynamic derivation (recommended)
```typescript
// Derive from agentSources at runtime, excluding primary orchestrators
import { agentSources } from "../agents/utils";

export const ALLOWED_AGENTS = Object.keys(agentSources)
  .filter(name => !["sisyphus"].includes(name));
```

### Option B: Explicit expansion
```typescript
export const ALLOWED_AGENTS = [
  "explore", "librarian", "hephaestus", "atlas",
  "oracle", "multimodal-looker", "metis", "momus"
] as const;
```

### Additional Changes Needed

1. Add tool restrictions in `AGENT_RESTRICTIONS` for hephaestus and atlas:
```typescript
hephaestus: {
    task: false,
    delegate_task: false,
    call_omo_agent: false  // Prevent recursive spawning
},
atlas: {
    task: false,
    call_omo_agent: false
}
```

[...]

PR #3901, opened 2026-05-09
fix(call-omo-agent): translate config-key subagent_type to display name before SDK dispatch

sjawhar:
## Summary

`call_omo_agent` accepts agents like `hephaestus`, `metis`, `momus`, `sisyphus-junior` per [`ALLOWED_AGENTS`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/tools/call-omo-agent/constants.ts) (added in #1579), but **dispatch always fails for these agents** because the tool sends the raw config key (`hephaestus`) to the OpenCode SDK, which only registers display names (`Hephaestus - Deep Agent`).

In the TUI this surfaces as a toast labelled **`Unknown error`** (the SDK's `UnknownError` class name), then a 5-minute polling timeout in `completion-poller.ts` before the tool returns `Error: Agent task timed out after 5 minutes.`. With `run_in_background: true` the failure is instant — the SDK rejection comes back from `BackgroundManager.launch` immediately.

## Reproduction (before fix)

From any session that has `call_omo_agent` exposed (e.g. OpenCode-Builder), against a Hephaestus subagent:

```
call_omo_agent(subagent_type="hephaestus", run_in_background=true, description="ping", prompt="Reply pong.")
→  TUI toast: "UnknownError"
→  [call_omo_agent] session.error: 'Agent not found: "hephaestus". Available agents: ..., Hephaestus - Deep Agent, ...'
```

The same fault hits `metis`, `momus`, `sisyphus-junior`, `prometheus`, `atlas` — anywhere [`AGENT_DISPLAY_NAMES`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/shared/agent-display-names.ts) maps config key → distinct display name. Same-keyed agents (`explore`, `librarian`, `oracle`, `multimodal-looker`) work today because key == display name.

## Root cause

Three dispatch sites passed the raw config key:

- [`sync-executor.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/tools/call-omo-agent/sync-executor.ts) — `agent: normalizedSubagentType` (sync path)
- [`background-executor.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/tools/call-omo-agent/background-executor.ts) — `agent: sanitizeSubagentType(args.subagent_type)` (live background path imported by `tools.ts`)
- [`background-agent-executor.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/tools/call-omo-agent/background-agent-executor.ts) — `agent: args.subagent_type` (parallel `executeBackgroundAgent` symbol that is currently only referenced by its own test file)

`delegate-task` already does the right thing (uses `getAgentDisplayName` in `subagent-discovery.ts`); `call_omo_agent`'s dispatch sites were missed.

## Fix

Translate via `getAgentDisplayName(stripAgentListSortPrefix(...))` at all three dispatch sites. The function is idempotent: same-keyed agents (`explore` → `explore`) are unchanged; ZWSP-prefixed display names are stripped first; quote/whitespace sanitization is preserved on the background path. Tool restrictions lookup (`getAgentToolRestrictions`) is case-insensitive and is left alone.

```diff
// sync-executor.ts
- agent: normalizedSubagentType,
+ agent: getAgentDisplayName(normalizedSubagentType),

// background-executor.ts (live, imported by tools.ts)
- agent: sanitizeSubagentType(args.subagent_type),
+ agent: getAgentDisplayName(stripAgentListSortPrefix(sanitizeSubagentType(args.subagent_type))),

// background-agent-executor.ts (parallel symbol; fixed for symmetry)
- agent: args.subagent_type,
+ agent: getAgentDisplayName(stripAgentListSortPrefix(args.subagent_type)),
```

## Tests (TDD)

Across `sync-executor.test.ts`, `background-executor.test.ts`, and `background-agent-executor.test.ts`:

| Case | Asserted dispatch `agent:` |
|---|---|
| `subagent_type: "hephaestus"` | `"Hephaestus - Deep Agent"` |
| `subagent_type: "sisyphus-junior"` | `"Sisyphus-Junior"` |
| `subagent_type: "explore"` (regression guard for same-keyed) | `"explore"` |
| `subagent_type: "\\hephaestus\\"` (sanitize+translate composition) | `"Hephaestus - Deep Agent"` |

The pre-existing `background-executor.test.ts` test "sanitizes subagent_type before passing to background manager launch" had its assertion flipped from `"hephaestus"` → `"Hephaestus - Deep Agent"` to encode the new contract while still verifying that quote-stripping is preserved.

Verified RED → GREEN locally; all 70 tests across the 10 files in `src/tools/call-omo-agent/` pass.

## End-to-end verification

Before fix, with `run_in_background: true`: tool returns immediately with `UnknownError` toast, plugin log shows `Agent not found: "hephaestus"`.
After fix, with `run_in_background: true`: `BackgroundManager.launch` receives `agent: "Hephaestus - Deep Agent"`, the SDK accepts it, the subagent session starts, and the tool returns the normal `Background agent task launched successfully... Task ID: ...` payload.
After fix, with `run_in_background: false`: `Hephaestus` responds with `pong` in seconds (validated live in an OpenCode-Builder session).

## Note on `background-agent-executor.ts`

The codebase contains both `background-executor.ts` and `background-agent-executor.ts` exporting different symbols (`executeBackground` vs `executeBackgroundAgent`). Only the former is wired into `tools.ts`; the latter has no callers besides its own test. The repo's `AGENTS.md` documents both files as if they form a router→launcher pair, but the actual code in `background-executor.ts` calls `manager.launch` directly without going through `executeBackgroundAgent`. This PR fixes both symbols for symmetry, but happy to drop the `background-agent-executor.ts` edits + tests (or to delete that file entirely) if you prefer.

## Related

- Closes the dispatch gap from #1579 (`ALLOWED_AGENTS` was expanded but dispatch sites were never updated)
- Likely improves #3774 reliability for non-same-keyed agents (Hephaestus, Metis, Momus, Sisyphus-Junior)

### DIFF
--- a/src/tools/call-omo-agent/background-agent-executor.test.ts
+++ b/src/tools/call-omo-agent/background-agent-executor.test.ts
@@ -152,4 +152,32 @@ describe("executeBackgroundAgent", () => {
     expect(secondResult).toContain("Task ID: task-2")
     expect(secondResult).not.toContain("interrupt")
   })
+
+  test("#given subagent_type is the lowercase config key 'hephaestus' #when executeBackgroundAgent runs #then BackgroundManager.launch receives the registered display name 'Hephaestus - Deep Agent'", async () => {
+    //#given
+    launchMock.mockClear()
+    launchMock.mockResolvedValueOnce({
+      id: "task-heph",
+      sessionId: "ses-heph",
+      description: "task",
+      agent: "Hephaestus - Deep Agent",
+      status: "pending",
+    })
+    getTaskMock.mockReturnValueOnce({
+      id: "task-heph",
+      sessionId: "ses-heph",
+      description: "task",
+      agent: "Hephaestus - Deep Agent",
+      status: "pending",
+    })
+    const args = { ...testArgs, subagent_type: "hephaestus" }
+
+    //#when
+    await executeBackgroundAgent(args, testContext, mockManager, mockClient)
+
+    //#then
+    const launchCall = launchMock.mock.calls.find(([input]) => (input as { agent: string }).agent !== undefined)
+    expect(launchCall).toBeDefined()
+    expect((launchCall![0] as { agent: string }).agent).toBe("Hephaestus - Deep Agent")
+  })
 })
--- a/src/tools/call-omo-agent/background-agent-executor.ts
+++ b/src/tools/call-omo-agent/background-agent-executor.ts
@@ -7,6 +7,7 @@ import type { CallOmoAgentArgs } from "./types"
 import type { ToolContextWithMetadata } from "./tool-context-with-metadata"
 import { getMessageDir } from "./message-storage-directory"
 import { getSessionTools } from "../../shared/session-tools-store"
+import { getAgentDisplayName, stripAgentListSortPrefix } from "../../shared/agent-display-names"
 
 export async function executeBackgroundAgent(
 	args: CallOmoAgentArgs,
@@ -39,7 +40,7 @@ export async function executeBackgroundAgent(
 		const task = await manager.launch({
 			description: args.description,
 			prompt: args.prompt,
-			agent: args.subagent_type,
+			agent: getAgentDisplayName(stripAgentListSortPrefix(args.subagent_type)),
 			parentSessionId: toolContext.sessionID,
 			parentMessageId: toolContext.messageID,
 			parentAgent,
--- a/src/tools/call-omo-agent/background-executor.test.ts
+++ b/src/tools/call-omo-agent/background-executor.test.ts
@@ -126,7 +126,7 @@ describe("executeBackground", () => {
     if (!launchArgs) {
       throw new Error("Expected launch arguments")
     }
-    expect(launchArgs.agent).toBe("hephaestus")
+    expect(launchArgs.agent).toBe("Hephaestus - Deep Agent")
   })
 
   test("keeps launched background task alive when parent aborts before session id resolves", async () => {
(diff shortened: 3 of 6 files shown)


sjawhar:
The single CI failure (`background-agent spawner agent-not-found fallback > retries with 'general' agent when promptAsync fails with Agent not found`) appears to be an existing CI flake unrelated to this PR. Evidence:

**1. The failing test lives in `src/features/background-agent/spawner.test.ts`, which this PR does not modify.** This branch only touches:

```
src/tools/call-omo-agent/background-agent-executor.test.ts | 28 ++++++++++
src/tools/call-omo-agent/background-agent-executor.ts      |  3 +-
src/tools/call-omo-agent/sync-executor.test.ts             | 63 +++++++++++++++++++++++
src/tools/call-omo-agent/sync-executor.ts                  |  4 +-
4 files changed, 95 insertions(+), 3 deletions(-)
```

**2. The failure is timing-shaped.** The CI log points to `spawner.test.ts:80:39`, which is the `expect(promptCalls[1].body.tools).toEqual(...)` assertion — meaning `promptCalls.length === 1` when the test expected 2. The test waits a fixed `setTimeout(resolve, 50)` for the fire-and-forget retry chain after `startTask()`, then asserts on results. The 4 subsequent tests in the same `describe` block use the same timing pattern and all passed.

**3. The test is locally green.** Five consecutive runs against this branch's HEAD on Bun `1.3.14-canary.1`:

```
$ for i in 1 2 3 4 5; do bun test src/features/background-agent/spawner.test.ts -t "retries with 'general' agent" 2>&1 | tail -3 | head -1; done
 0 fail
 0 fail
 0 fail
 0 fail
 0 fail
```

A retry of the `test` job should be green. If the flake is reproducible on `dev`, that's a pre-existing test-stability issue (the 50ms wait is racy on slower runners) worth tracking separately, but it is independent from the dispatch fix in this PR.

For reference, the equivalent change has been running stably on a downstream fork (`sjawhar/oh-my-opencode` sami release `v4.0.0-sami.20260509-100454-sami.20260509-203151`, [run #25611120373](https://github.com/sjawhar/oh-my-opencode/actions/runs/25611120373)) since this morning, with `test` job green on the same Bun version.

code-yeongyu:
Superseded by #4045 (merged into `dev`). #4045 carries this fix rebased onto current dev (which now has the shared promptAsync gate, ralph-loop display-name fix, and session-recovery persistent dedupe) with conflict resolution. Original commit authorship retained via rebase. Thanks @Sami-Jawhar.

code-yeongyu:
Closing as superseded by #4045 (merged).

PR #3622, opened 2026-04-24
fix(tool-execute-after): cap excessively long tool output to prevent TUI flooding (fixes #3586)

MoerAI:
## Summary
- Cap all tool output at 3000 characters to prevent raw error dumps from flooding the TUI

## Problem
When a tool fails (e.g., reading a non-existent file), the error message can contain long raw stack traces, framework internals, or serialized objects. The existing tool-output-truncator only handles specific tool types. For other tools, the full error is dumped into the TUI as a large, unreadable blob of text.

## Fix
Added a universal catch-all in `tool-execute-after.ts` that caps any tool output exceeding 3000 characters. This runs after all other hooks (including the tool-specific truncator), so it only fires on outputs that are still too long. The truncation is lightweight (string slice) and does not affect the token-aware truncation used for known tools.

## Changes
| File | Change |
|------|--------|
| `src/plugin/tool-execute-after.ts` | Add MAX_ERROR_OUTPUT_CHARS cap after all hooks run |

Fixes #3586

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3622"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Codesmith can help with this PR — just tag <code>@codesmith</code> or enable autofix.</sup>

- [ ] Autofix CI and bot reviews

---
## Summary by cubic
Caps tool output to 3000 characters after all hooks to prevent TUI flooding from huge error dumps (fixes #3586).
Adds a simple catch-all in `src/plugin/tool-execute-after.ts` that slices long string outputs and appends a truncation notice without affecting existing token-aware truncation.

<sup>Written for commit 2c78cc68dab4326f1a53919ee949fd4be5269f12. Summary will update on new commits.</sup>

Issue #3586, opened 2026-04-23
[Bug]: With oh-my-opencode enabled, the opencode TUI dumps long raw tool failure text into the session UI


yangrn-new:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

When running opencode with the oh-my-opencode plugin, a failed tool invocation causes the TUI to render the associated error inline as a large blob of mostly raw, low-level text. The output is lengthy, mixes in runtime/framework-related fragments, and reads as unstructured noise rather than a concise, scannable user-facing [message.]

<img width="2560" height="1380" alt="Image" src="https://github.com/user-attachments/assets/629c05f9-c9ea-4678-be63-968aaa5973cd" />

### Steps to Reproduce

1. Enable the OpenCode plugin oh-my-opencode：
    cd C:\Your-Project-Path
    opencode plugin oh-my-opencode
2. Open the TUI in your project workspace.
3. Send a prompt asking to access a non-existent PDF, e.g.
    Access the file C:\Your-Project-Path\gXqmeyarvWhwagqo.pdf
   (the file must not exist on disk).
4. Observe the tool error area in the TUI after the read (or equivalent) call fails.

### Expected Behavior

Optimize exception handling, do not directly print errors on the opencode TUI.

### Actual Behavior

Display error messages to the user in an unfriendly manner.

### Doctor Output

```shell
oMoMoMoMo Doctor

 ⚠ 2 issues found:

1. Comment checker unavailable
   Comment checker binary is not installed.
   Fix: Install @code-yeongyu/comment-checker
   Affects: comment-checker hook

2. GitHub CLI missing
   gh CLI is not installed.
   Fix: Install from https://cli.github.com/
   Affects: GitHub automation
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Windows

### OpenCode Version

1.14.20

yangrn-new:
[...]

PR #3930, opened 2026-05-10
fix(plugin): preserve background output results

masterkain:
## Summary

- preserve long successful `background_output` tool results so parent agents can consume complete subagent reports
- keep the display flood guard for long outputs that look like failures
- add regression coverage for both paths

## Problem

Oh My OpenAgent uses background agents for investigation work. The parent session later calls `background_output` to read the completed task result. Since [commit `2c78cc68dab4326f1a53919ee949fd4be5269f12`](https://github.com/code-yeongyu/oh-my-openagent/commit/2c78cc68dab4326f1a53919ee949fd4be5269f12), the global cap in [`src/plugin/tool-execute-after.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/plugin/tool-execute-after.ts) trims every string output above 3000 chars and appends `...(output truncated for display)`

That guard was added for [`code-yeongyu/oh-my-openagent#3586`](https://github.com/code-yeongyu/oh-my-openagent/issues/3586) to stop noisy failure output from flooding the UI, but it also truncates successful subagent reports. In practice the parent model receives only the beginning of the report, loses the rest of the findings, and may continue from incomplete evidence. This wastes context and tokens because the parent agent has to recover from an incomplete result

## Regression Window

- introduced by [commit `2c78cc68dab4326f1a53919ee949fd4be5269f12`](https://github.com/code-yeongyu/oh-my-openagent/commit/2c78cc68dab4326f1a53919ee949fd4be5269f12) on 2026-04-24
- merged through [pull request `code-yeongyu/oh-my-openagent#3622`](https://github.com/code-yeongyu/oh-my-openagent/pull/3622) on 2026-05-06
- first released in [`v4.0.0`](https://github.com/code-yeongyu/oh-my-openagent/releases/tag/v4.0.0) on 2026-05-07

## Fix

- move the 3000-char guard behind `shouldCapErrorOutput`
- cap only outputs with failure-shaped prefixes such as `Error:`, `TypeError`, `Traceback`, and `Session error:`
- leave normal long outputs unchanged, including `background_output`

## Verification

- `bun test src/plugin/tool-execute-after.test.ts --bail` - pass
- `bun run typecheck` - pass
- `bun run build` - pass
- `git diff --check` - pass
- `bun test bin script src` - 6570 pass, 1 skip, 5 fail; failures are outside this change:
  - [`src/shared/tmux/tmux-utils/pane-replace.test.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/shared/tmux/tmux-utils/pane-replace.test.ts) fails because `runTmuxCommandMock.mock.calls[...]` is undefined
  - [`src/hooks/auto-update-checker/checker/plugin-entry.test.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/hooks/auto-update-checker/checker/plugin-entry.test.ts) reads local config and finds `oh-my-openagent@4.0.0` in `/Users/kain/.config/opencode/opencode.jsonc`

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3930"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Stop truncating successful `background_output` results so parent agents receive complete subagent reports. Keep the display flood guard by only capping long outputs that look like errors.

- **Bug Fixes**
  - Apply the 3000-char cap only when `shouldCapErrorOutput` matches failure-like prefixes (e.g., Error, TypeError, Traceback, Exception, Session error, ENOENT/EACCES/EPERM).
  - Preserve long non-error outputs, including `background_output`.
  - Add tests for preserved background output and capped error output.

<sup>Written for commit 867ea40b84c3176894a9f60b5851bc1cf2b81825. Summary will update on new commits.</sup>

### DIFF
--- a/src/plugin/tool-execute-after.test.ts
+++ b/src/plugin/tool-execute-after.test.ts
@@ -92,4 +92,54 @@ describe("createToolExecuteAfterHandler", () => {
     expect(output.title).toBe("stored title")
     expect(output.metadata).toEqual({ sessionId: "ses_native", agent: "hephaestus" })
   })
+
+  it("#given long background_output #when tool.execute.after runs #then output is preserved for agent consumption", async () => {
+    // given
+    const handler = createToolExecuteAfterHandler({
+      ctx: { directory: "/repo" } as never,
+      hooks: {} as never,
+    })
+    const longOutput = [
+      "Task Result",
+      "",
+      "Task ID: bg_test",
+      "Description: Investigate regression",
+      "---",
+      "A".repeat(3500),
+    ].join("\n")
+    const output = { title: "background result", output: longOutput, metadata: {} }
+
+    // when
+    await handler(
+      { tool: "background_output", sessionID: "ses_parent", callID: "call_background" },
+      output
+    )
+
+    // then
+    expect(output.output).toBe(longOutput)
+    expect(output.output).not.toContain("output truncated for display")
+  })
+
+  it("#given long error output #when tool.execute.after runs #then display flood guard still caps it", async () => {
+    // given
+    const handler = createToolExecuteAfterHandler({
+      ctx: { directory: "/repo" } as never,
+      hooks: {} as never,
+    })
+    const output = {
+      title: "look_at error",
+      output: `Error: Failed to analyze missing.pdf: ${"A".repeat(3500)}`,
+      metadata: {},
+    }
+
+    // when
+    await handler(
+      { tool: "look_at", sessionID: "ses_parent", callID: "call_look_at" },
+      output
+    )
+
+    // then
+    expect(output.output.length).toBeLessThan(3500)
+    expect(output.output).toContain("output truncated for display")
+  })
 })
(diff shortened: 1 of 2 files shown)


masterkain:

masterkain:
I don't think the pr above is any good but the problem is real

acamq:
Thank you for bringing this to my attention: #3622 introduced a severe regression but didn't resolve the original issue in #3586, and it should never have been merged at all. `background_output` flooding the terminal is a complete non-issue. I've reverted that PR. As such, this one is unnecessary as well. Again, thank you for your contributions.

PR #3947, opened 2026-05-11
fix(background-agent): add OMO_DISABLE_PROCESS_CLEANUP env opt-out for global handlers (fixes #3856)

### LINKED ISSUE #3856: Make global uncaughtException/unhandledRejection handlers opt-in to background-agent feature
## Summary

The `process-cleanup` module unconditionally registers `uncaughtException` and `unhandledRejection` handlers on the host Node process whenever any `CleanupManager` is created (`src/features/background-agent/process-cleanup.ts`, function `registerManagerForCleanup`). On any unhandled error/rejection these handlers call `scheduleForcedExit(handler(error), 1, true)` → `process.exit(1)`, terminating the entire opencode process.

After PR #3639 (which I appreciate is intentional for users actively running background agents), this kills opencode for users who load the plugin but **don't** use background agents. Transient streaming errors from upstream LLM endpoints — e.g. `SocketError { code: 'UND_ERR_SOCKET' }` from undici when `api.githubcopilot.com` resets a long stream — surface as unhandled rejections in opencode's `session.processor` Effect runtime and trip these global handlers, exiting the app mid-response. This happens repeatedly on flaky networks/proxies and gives users a "opencode just dies after every response" experience.

This is distinct from #3731 (re-entrant log loop) — that fix prevents log spam during cleanup, but the forced exit on every transient unhandled rejection remains.

## Reproduction

1. Install `oh-my-openagent` (current `^4.0.0`) into an opencode config dir, list it under `plugin` in `opencode.json`.
2. Configure a `github-copilot/*` model (or any provider with occasional transient socket resets).
3. Send a long streaming prompt over a flaky connection (corporate proxy, hotspot, etc.) so undici sees a mid-stream socket reset.
4. opencode receives an `AI_APICallError { cause: SocketError UND_ERR_SOCKET }`. The error propagates as an unhandled rejection. The plugin's listener fires and calls `process.exit(1)`. The entire opencode process exits — the desktop window closes, the CLI returns to the shell prompt — even when the user never invoked any background-agent functionality.

The same applies to any unhandled rejection in opencode itself or any other plugin: the moment one bubbles up uncaught, the host dies.

## Why I think this should be opt-in

The `process-cleanup` machinery is conceptually tied to the background-agent feature (`src/features/background-agent/process-cleanup.ts`), but `registerManagerForCleanup` is called unconditionally whenever a `CleanupManager` is constructed. A user who installs the plugin only for the multi-agent presets / model fallback / LSP tools effectively has no opt-out from the global handlers, and gets none of the benefit of background-agent forced cleanup either.

## Suggested fix

One of:

1. **Feature-gate the global handler registration** behind whatever flag enables background agents (config option, command-line flag, or only register when the first background-agent task is actually started rather than at plugin load).
2. **Provide an explicit opt-out env var** like `OMO_DISABLE_PROCESS_CLEANUP=1` that short-circuits `registerManagerForCleanup` (and document it in the troubleshooting README).
3. **Default to `process.exitCode = 1` (no `process.exit`)** so the process exits naturally only when the event loop drains, and let opencode's own error handling decide whether to recover, log, and continue. The forced exit can still be opt-in for the bg-agent users who depend on it.

(1) seems cleanest. (2) is the lowest-risk near-term mitigation for impacted users.

## Environment

- `oh-my-openagent` 4.0.0 (also reproduced on 3.17.14)
- `opencode` 1.14.41 (Electron desktop) and 1.14.39 (npm CLI)
- macOS 15.x, Node 24.x
- Provider: `github-copilot`

## Code references

- Listener registration: `src/features/background-agent/process-cleanup.ts` — `registerManagerForCleanup` → `cleanupErrorHandlers.set("uncaughtException", registerErrorEvent("uncaughtException", cleanupAll))` and the equivalent `unhandledRejection` line
- Forced exit: same file, `scheduleForcedExit` calling `process.exit(exitCode)` after the cleanup promise resolves

Thanks for the active maintenance here — the recent work on #3639 / #3731 / #3793 shows you're already thinking carefully about this surface, and I'm hopeful a feature gate is a small ergonomic win for users in the unfortunate "loaded but innocent" position.

MoerAI:
## Summary
Provide an env-var opt-out for the global `uncaughtException`/`unhandledRejection` handlers installed by `registerManagerForCleanup`. Users who load the plugin but don't actively use background-agent tasks can set `OMO_DISABLE_PROCESS_CLEANUP=1` to avoid the host process dying on transient streaming errors.

## Root Cause
[`src/features/background-agent/process-cleanup.ts`](src/features/background-agent/process-cleanup.ts) registers two `process.on(...)` listeners that call `scheduleForcedExit(handler(error), 1, true)` → `process.exit(1)` on any uncaught error or unhandled rejection — even ones from unrelated code paths (other plugins, opencode's own session processor, etc.).

The original issue reporter walked through a concrete failure: undici emits `SocketError { code: 'UND_ERR_SOCKET' }` when `api.githubcopilot.com` resets a mid-stream HTTP/2 response. That surfaces as an unhandled rejection in opencode's `session.processor` Effect runtime → our listener fires → opencode dies mid-response. Reproduced repeatedly on flaky networks / corporate proxies, with the user never invoking any background-agent functionality.

## Changes
| File | Change |
|------|--------|
| `src/features/background-agent/process-cleanup.ts` | Add `OMO_DISABLE_PROCESS_CLEANUP` env-var check (accepts `1`/`true`/`yes`/`on`, case-insensitive). When set, skip the error-event registration but keep SIGINT/SIGTERM/SIGBREAK/beforeExit/exit signal handlers active so graceful shutdown still works for whatever was already registered. Log a single line on the path so it's discoverable in `/tmp/oh-my-opencode.log`. |
| `src/features/background-agent/process-cleanup.test.ts` | Add 6 regression tests covering env-var truthy/falsy/unset states, signal-handler preservation, and end-to-end behavior under `uncaughtException`. |

Diff: **+126 lines / -0** across 2 files.

## Why opt-out rather than feature-gate or change the default

The original issue proposes three options:
1. Feature-gate handler registration behind background-agent activation
2. Provide an explicit env-var opt-out
3. Default to `process.exitCode = 1` (let the event loop drain naturally)

This PR implements **option #2** because it is the smallest, least-risky change for the impacted users and preserves the existing behavior for everyone running background agents (which depend on the forced-exit cleanup discipline from #3639). Option #1 ties the registration to internal lifecycle events that are harder to reason about; option #3 is a behavior change to the default path and warrants its own design discussion. Option #2 unblocks users *today* without touching the default.

## Reproduction (before fix)
```text
src\features\background-agent\process-cleanup.test.ts:
expect(received).toHaveLength(expected)
Expected length: 0
Received length: 1
(fail) #given env var is set to true #when registerManagerForCleanup runs #then handlers are NOT registered

expect(received).not.toHaveBeenCalled()
Expected number of calls: 0
Received number of calls: 1
(fail) #given env var is set AND process emits uncaughtException #when event fires #then manager shutdown is NOT invoked by our handler

 17 pass / 3 fail
```

Manual reproduction matches: setting `OMO_DISABLE_PROCESS_CLEANUP=1` and emitting `process.emit("uncaughtException", new Error("boom"))` still triggers the registered manager's `shutdown` because the env var is ignored pre-fix.

## Verification (after fix)
```text
bun test src/features/background-agent/process-cleanup.test.ts
 20 pass / 0 fail
 36 expect() calls
Ran 20 tests across 1 file. [161.00ms]

bun run typecheck
$ tsc --noEmit          (clean)
```

Manual QA via `bun --eval`:
```
env unset → handlers registered count: 1 PASS
env=1 → handlers registered count: 0 PASS
```

The existing 14 process-cleanup tests still pass unchanged, confirming the default path (`OMO_DISABLE_PROCESS_CLEANUP` unset) is unaffected.

## Documentation
The README already documents two similar env-var opt-outs (`OMO_SEND_ANONYMOUS_TELEMETRY`, `OMO_DISABLE_POSTHOG`). A doc note for `OMO_DISABLE_PROCESS_CLEANUP` can land in a follow-up if you'd like the same treatment — happy to add it to this PR if preferred, but kept it out for now to keep the diff focused on the behavioral fix.

## Test
- Regression tests: 6 new cases in `src/features/background-agent/process-cleanup.test.ts`, all in given/when/then style consistent with the surrounding suite.
- Related suite: `bun test src/features/background-agent/process-cleanup.test.ts` — 20 pass.
- Typecheck: `bun run typecheck` — clean.

Fixes #3856

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3947"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Add `OMO_DISABLE_PROCESS_CLEANUP` env opt-out to skip global `uncaughtException`/`unhandledRejection` handlers that force-exit the process, preventing crashes from unrelated transient errors. Default behavior is unchanged for users who don’t set the env var.

- **Bug Fixes**
  - Skip registering global error handlers when `OMO_DISABLE_PROCESS_CLEANUP` is truthy (`1`, `true`, `yes`, `on`).
  - Keep signal handlers (`SIGINT`, `SIGTERM`, `SIGBREAK`, `beforeExit`, `exit`) for graceful shutdown; log once when opt-out is used.

- **Migration**
  - Set `OMO_DISABLE_PROCESS_CLEANUP=1` if you load the plugin but don’t run background-agent tasks and see transient streaming errors.
  - No action needed otherwise.

<sup>Written for commit dd68f324d0919cd983aa46d264a9f16b70ca34ac. Summary will update on new commits.</sup>

### DIFF
--- a/src/features/background-agent/process-cleanup.test.ts
+++ b/src/features/background-agent/process-cleanup.test.ts
@@ -235,6 +235,103 @@ describe("#given process cleanup registration", () => {
     })
   })
 
+  describe("#given OMO_DISABLE_PROCESS_CLEANUP env var", () => {
+    let originalEnvValue: string | undefined
+
+    beforeEach(() => {
+      originalEnvValue = process.env.OMO_DISABLE_PROCESS_CLEANUP
+    })
+
+    afterEach(() => {
+      if (originalEnvValue === undefined) {
+        delete process.env.OMO_DISABLE_PROCESS_CLEANUP
+      } else {
+        process.env.OMO_DISABLE_PROCESS_CLEANUP = originalEnvValue
+      }
+    })
+
+    test("#given env var is set to 1 #when registerManagerForCleanup runs #then uncaughtException handler is NOT registered", () => {
+      const uncaughtExceptionListenersBefore = process.listeners("uncaughtException")
+      const unhandledRejectionListenersBefore = process.listeners("unhandledRejection")
+      process.env.OMO_DISABLE_PROCESS_CLEANUP = "1"
+      const manager = { shutdown: mock(() => {}) }
+      registeredManagers.push(manager)
+
+      registerManagerForCleanup(manager)
+
+      expect(process.listeners("uncaughtException")).toHaveLength(uncaughtExceptionListenersBefore.length)
+      expect(process.listeners("unhandledRejection")).toHaveLength(unhandledRejectionListenersBefore.length)
+    })
+
+    test("#given env var is set to true #when registerManagerForCleanup runs #then handlers are NOT registered", () => {
+      const uncaughtExceptionListenersBefore = process.listeners("uncaughtException")
+      process.env.OMO_DISABLE_PROCESS_CLEANUP = "true"
+      const manager = { shutdown: mock(() => {}) }
+      registeredManagers.push(manager)
+
+      registerManagerForCleanup(manager)
+
+      expect(process.listeners("uncaughtException")).toHaveLength(uncaughtExceptionListenersBefore.length)
+    })
+
+    test("#given env var is set to 0 #when registerManagerForCleanup runs #then handlers ARE registered", () => {
+      const uncaughtExceptionListenersBefore = process.listeners("uncaughtException")
+      process.env.OMO_DISABLE_PROCESS_CLEANUP = "0"
+      const manager = { shutdown: mock(() => {}) }
+      registeredManagers.push(manager)
+
+      registerManagerForCleanup(manager)
+
+      expect(process.listeners("uncaughtException")).toHaveLength(uncaughtExceptionListenersBefore.length + 1)
+    })
+
+    test("#given env var is unset #when registerManagerForCleanup runs #then handlers ARE registered", () => {
+      const uncaughtExceptionListenersBefore = process.listeners("uncaughtException")
+      delete process.env.OMO_DISABLE_PROCESS_CLEANUP
+      const manager = { shutdown: mock(() => {}) }
+      registeredManagers.push(manager)
+
+      registerManagerForCleanup(manager)
+
+      expect(process.listeners("uncaughtException")).toHaveLength(uncaughtExceptionListenersBefore.length + 1)
+    })
+
+    test("#given env var is set #when signals fire #then SIGINT/SIGTERM/beforeExit/exit handlers still run cleanup", () => {
+      const exitListenersBefore = process.listeners("exit")
+      process.env.OMO_DISABLE_PROCESS_CLEANUP = "yes"
+      const shutdown = mock(() => {})
+      const manager = { shutdown }
+      registeredManagers.push(manager)
+
+      registerManagerForCleanup(manager)
+      const exitListener = getNewListener("exit", exitListenersBefore)
+      exitListener()
+
+      expect(shutdown).toHaveBeenCalledTimes(1)
+    })
+
+    test("#given env var is set AND process emits uncaughtException #when event fires #then manager shutdown is NOT invoked by our handler", async () => {
+      process.env.OMO_DISABLE_PROCESS_CLEANUP = "1"
+      const exitSpy = spyOn(process, "exit").mockImplementation((() => undefined) as never)
+      const shutdown = mock(() => {})
+      const manager = { shutdown }
+      registeredManagers.push(manager)
+
+      try {
+        registerManagerForCleanup(manager)
+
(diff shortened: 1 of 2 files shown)


MoerAI:

PR #3952, opened 2026-05-11
fix(runtime-fallback): first-prompt watchdog for silently-stuck subagents

ismetanin:
## Summary

Fixes the silent-stuck failure mode where a subagent dispatched to a provider hits a rate limit / quota / outage, the SDK enters an internal retry loop that absorbs the error, and the runtime-fallback hook never sees an event to react on. Result today: subagent sits in `retry` status, the parent's `task` tool call hangs for the full 30-minute `DEFAULT_POLL_TIMEOUT_MS` budget, and the configured fallback chain is never even attempted. Observed in the wild on Momus and Sisyphus-Junior dispatches when OpenAI quota was exhausted while Anthropic was healthy.

## Root cause

The runtime-fallback subsystem is fully reactive — it acts on `message.updated` (with `info.error`), `session.error`, and `session.status: retry` (with a recognisable message). All three depend on the SDK surfacing the error. When the SDK retries 429s silently and presents an "in progress" face to OpenCode, none of those events fire, and the hook stays idle. The configured fallback model is never attempted because nothing pulls the trigger.

There is no proactive watchdog for the **initial** prompt to a subagent. All existing watchdogs (`scheduleSessionFallbackTimeout`, the outer poll timeout) protect either *post-fallback-dispatch* states or the *outer parent* timeout — nothing watches the most common failure mode: "we dispatched a subagent, the SDK accepted the prompt, no first response ever arrives."

## Fix

A small new module `first-prompt-watchdog.ts` that synthesises the missing trigger:

- **Arm**: on `message.updated` with `role === "user"` for a session in `subagentSessions`. 90s timer starts (`DEFAULT_FIRST_PROMPT_WATCHDOG_MS`).
- **Cancel**: on the first sign of progress — assistant message with text/reasoning content, `finish` field, or `error` field — or on any session terminal event (`session.idle` / `session.stop` / `session.deleted` / `session.error`).
- **Fire**: at the 90s mark, if not cancelled, abort the in-flight request and route into the existing `dispatchFallbackRetry` pipeline. Exactly equivalent to a `session.error` arriving on time, just simulated when the SDK fails to surface one.

### Design choices

- **Dispatch fallback, do not abort outright.** Network loss looks identical to a stuck SDK retry from the hook's vantage point. With fallback-dispatch behaviour, network loss gracefully degrades to today's baseline (both attempts fail, the 30-min outer timeout still ends things) instead of destructively aborting work that might have recovered when the network blip resolved.
- **Subagent-only.** Parent/user sessions can legitimately take >90s before first token. Subagent dispatches in practice produce first content much faster, so a 90s ceiling is safe and tight.
- **No new fallback mechanism.** The watchdog only synthesises the missing event; the actual swap-and-retry is the existing `dispatchFallbackRetry → prepareFallback → autoRetryWithFallback` flow.
- **Single threshold knob.** `DEFAULT_FIRST_PROMPT_WATCHDOG_MS = 90_000` in `constants.ts`, exposed as the third arg to `createFirstPromptWatchdog` for testing.

### Bonus: diagnostic log

`session-status-handler.ts` had a silent return when a `session.status: retry` event arrives with a message that doesn't match `RETRYABLE_ERROR_PATTERNS`. Added a single log line capturing the raw `retryMessage` and `attempt` in that branch. Cost: one log line. Benefit: next time a provider emits a retry phrasing we don't recognise, we'll see what it actually said and can extend the patterns. No behaviour change.

## Test plan

- [x] New test file `first-prompt-watchdog.test.ts` with 5 cases:
  - Subagent silent past threshold + fallback configured → abort called + `autoRetryWithFallback` invoked with the fallback model.
  - Subagent produces assistant text before threshold → cancelled, no abort, no fallback dispatch.
  - Non-subagent session → never arms, no fire.
  - Subagent reaches session terminal before threshold → cancelled, no fire.
  - Subagent silent past threshold + no fallback configured → logs only, no abort, no dispatch (lets other safety nets handle).
- [x] `bun test src/hooks/runtime-fallback/` → 138/138 pass (133 pre-existing + 5 new).
- [x] `bun test src/hooks/runtime-fallback src/features/claude-code-session-state` → 160/160 pass.
- [x] `bunx tsc --noEmit` → clean.

## Relationship to PR #3950

Independent. PR #3950 fixes the case where an error event *did* arrive but `fallbackModels.length === 0`; this PR fixes the case where the error event *never* arrives. Different code paths, different scenarios, can land in either order.

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3952"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Fixes silently stuck subagents by adding a first‑prompt watchdog that triggers fallback if no progress is seen within 90s. Progress detection now includes any assistant part (text/reasoning/tools/etc.) to avoid false fires.

- **Bug Fixes**
  - First‑prompt watchdog: armed on subagent user messages; cancelled on assistant progress (error/finish or any part: text, reasoning, tool, tool_use, tool_result, tool-call, step-start, file) or on terminal session events; 90s threshold (`DEFAULT_FIRST_PROMPT_WATCHDOG_MS`).
  - On fire: aborts the in‑flight request and routes to the existing fallback flow; if no fallback is configured, logs only.
  - Event routing uses a new `observeEventForWatchdog` helper wired into `message.updated` and `session.*`.
  - Adds a log when `session.status: retry` has a message that doesn’t match `RETRYABLE_ERROR_PATTERNS`.

<sup>Written for commit 52414185861b296ad9478d44a3abf6de98f4b56e. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/runtime-fallback/constants.ts
+++ b/src/hooks/runtime-fallback/constants.ts
@@ -47,3 +47,14 @@ export const RETRYABLE_ERROR_PATTERNS = [
  * Hook name for identification and logging
  */
 export const HOOK_NAME = "runtime-fallback"
+
+/**
+ * First-prompt watchdog: how long to wait for the first sign of progress
+ * (assistant text/reasoning/finish) from a subagent session before assuming
+ * the provider is silently stuck and dispatching the configured fallback.
+ *
+ * Tuned to be longer than typical first-token latency (well under 30s in
+ * practice) yet much shorter than the 30-minute outer poll timeout that
+ * would otherwise be the only safety net.
+ */
+export const DEFAULT_FIRST_PROMPT_WATCHDOG_MS = 90_000
--- a/src/hooks/runtime-fallback/first-prompt-watchdog.test.ts
+++ b/src/hooks/runtime-fallback/first-prompt-watchdog.test.ts
@@ -1,0 +1,200 @@
+import { afterEach, beforeEach, describe, expect, it } from "bun:test"
+import type { HookDeps, RuntimeFallbackPluginInput } from "./types"
+import type { AutoRetryHelpers } from "./auto-retry"
+import { subagentSessions } from "../../features/claude-code-session-state"
+import { createFirstPromptWatchdog } from "./first-prompt-watchdog"
+
+const WATCHDOG_MS = 40
+const SAFE_WAIT_AFTER_FIRE_MS = 120
+const SAFE_WAIT_BEFORE_FIRE_MS = 15
+
+function wait(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+function createContext(): RuntimeFallbackPluginInput {
+  return {
+    client: {
+      session: {
+        abort: async () => ({}),
+        messages: async () => ({ data: [] }),
+        promptAsync: async () => ({}),
+      },
+      tui: {
+        showToast: async () => ({}),
+      },
+    },
+    directory: "/test/dir",
+  }
+}
+
+function createDeps(pluginConfig: Record<string, unknown> = {}): HookDeps {
+  return {
+    ctx: createContext(),
+    config: {
+      enabled: true,
+      retry_on_errors: [429, 503, 529],
+      max_fallback_attempts: 3,
+      cooldown_seconds: 60,
+      timeout_seconds: 30,
+      notify_on_fallback: false,
+    },
+    options: undefined,
+    pluginConfig,
+    sessionStates: new Map(),
+    sessionLastAccess: new Map(),
+    sessionRetryInFlight: new Set(),
+    sessionAwaitingFallbackResult: new Set(),
+    sessionFallbackTimeouts: new Map(),
+    sessionStatusRetryKeys: new Map(),
+  }
+}
+
+interface RecordedCalls {
+  abort: Array<{ sessionID: string; source: string }>
+  autoRetry: Array<{ sessionID: string; newModel: string; resolvedAgent: string | undefined; source: string }>
+}
+
+function createHelpers(calls: RecordedCalls, resolvedAgentName?: string): AutoRetryHelpers {
+  return {
+    abortSessionRequest: async (sessionID: string, source: string) => {
+      calls.abort.push({ sessionID, source })
+    },
+    clearSessionFallbackTimeout: () => {},
+    scheduleSessionFallbackTimeout: () => {},
+    autoRetryWithFallback: async (sessionID, newModel, resolvedAgent, source) => {
+      calls.autoRetry.push({ sessionID, newModel, resolvedAgent, source })
+    },
+    resolveAgentForSessionFromContext: async () => resolvedAgentName,
+    cleanupStaleSessions: () => {},
+  }
+}
+
+const AGENT = "sisyphus-junior"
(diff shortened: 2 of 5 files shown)


ismetanin:
Addressed both issues in 52414185.

**Issue 1 (hook.ts:76 — watchdog progress detection too narrow):** verified against the codebase — `sync-session-poller.ts:9` (`PENDING_TOOL_PART_TYPES = ["tool", "tool_use", "tool-call"]`) and `hooks/session-recovery/types.ts:3` confirm OpenCode emits `tool`/`tool_use`/`tool_result`/`tool-call` parts on assistant messages, and subagents that immediately run tools (Read/Bash/Edit) are exactly that case. Broadened the check: any assistant part of any known type now counts as progress, alongside the existing `info.error`/`info.finish` cancels. Refactored the OpenCode-event→watchdog-signal translation out of `hook.ts` into an exported `observeEventForWatchdog` on the watchdog module so it's testable in isolation. Added parametrised regression tests covering `text`, `reasoning`, `tool`, `tool_use`, `tool_result`, `tool-call`, `step-start`, `file`, plus the empty-parts/error/finish/non-target-event branches.

**Issue 2 (test:128 — flaky real-time sleeps near the 40ms boundary):** bun:test has no built-in fake-timer API, so the test stays on real timers but with much wider margins. Threshold bumped from 40ms → 100ms; pre-cancel wait 15ms → 40ms (60ms margin before the timer fires); post-fire wait 120ms → 250ms (~2.5x the threshold). Added a comment in the test file explaining the sizing rationale so future edits don't accidentally shrink the margins back.

Verification:
- `bun test src/hooks/runtime-fallback/` → 156/156 pass (up from 138 — 18 new observer cases for the part-type coverage).
- `bunx tsc --noEmit` → clean.

code-yeongyu:
Triage note: rebase onto current `dev` produces two conflicts in `src/hooks/runtime-fallback/hook.ts` (HookDeps + types refactor) and `session-status-handler.ts` (handler signature). Recent runtime-fallback merges that shifted the integration surface: #3982 (category fallback honoring), #4007 (preserve attemptCount on internal abort).

The watchdog idea is still valuable. To re-land:
1. Rebase onto current `dev`.
2. Re-thread `createFirstPromptWatchdog` through the new `HookDeps` shape including `internallyAbortedSessions` from #4007.
3. Make sure the watchdog interacts with the new auto-retry / fallback chain in #3982 / #4007 (no double-trigger on the very first prompt where #4007's internal-abort marker is also set).
4. Confirm root `bun test --timeout 30000` is 0 fail.

Leaving open for now since this addresses a real failure mode (SDK absorbs 429s silently → subagent stuck for 30 min). If you'd like me to take the rebase, drop a comment.

code-yeongyu:
Superseded by #4051 (merged into `dev`). The watchdog idea + tests landed on top of current dev with conflict resolution against the factories refactor and #4007 internal-abort tracking. Thanks @ismetanin — closing as superseded.

code-yeongyu:
Closing as superseded by #4051 (merged).

Issue #3880, opened 2026-05-09
Plugin fails to load in OpenCode GUI (Electron/Node.js) due to bun:sqlite import


wdsAI-space:
## Bug Description

The `oh-my-openagent` plugin fails to load in the **OpenCode GUI desktop application** (Electron/Node.js runtime) because it uses `bun:sqlite` which is a Bun-specific API not available in Node.js/Electron environments.

## Environment

- **OpenCode CLI**: v1.14.41 (via Bun) — plugin works correctly
- **OpenCode GUI**: OpenCode.exe (Electron) — plugin fails to load
- **oh-my-openagent**: v3.17.5
- **OS**: Windows 11
- **oh-my-opencode**: v3.17.15

## Error Log

```
ERROR service=plugin path=oh-my-openagent 
  target=file:///C:/Users/.../oh-my-openagent/dist/index.js 
  error=Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. Received protocol 'bun:' 
  failed to load plugin
```

This causes a cascade of errors:
```
ERROR service=plugin error=Cannot read properties of undefined (reading 'config') plugin config hook failed
ERROR service=server error=Cannot read properties of undefined (reading 'provider') failed
```

The plugin loading failure cascades into the server initialization, making the **entire model list unavailable** in the GUI.

## Root Cause

In `dist/index.js` line 124312:
```javascript
import { Database } from "bun:sqlite";
```

This `bun:` protocol import is only resolvable by the Bun runtime. When OpenCode GUI (which runs on Electron/Node.js) tries to load the plugin, it fails because Node.js doesn't support the `bun:` protocol.

## Impact

- Model list is completely empty in OpenCode GUI
- All model-related operations fail with `Cannot read properties of undefined (reading 'provider')`
- The GUI is effectively unusable for any model interaction

## Suggested Fix

Option 1: Use a conditional import that falls back to `better-sqlite3` when running in Node.js:
```javascript
let Database; try {
  const bunSqlite = await import("bun:sqlite"); Database = bunSqlite.Database; } catch {
  const betterSqlite = await import("better-sqlite3"); Database = betterSqlite.default; }
```

Option 2: Provide separate build artifacts for Bun and Node.js environments.

Option 3: Gracefully degrade when SQLite is not available (the plugin should still be able to provide agent/model configuration without the database feature).

## Workaround

[...]

PR #3832, opened 2026-05-07
fix(desktop): hide bun:sqlite import from Node.js/Electron ESM loader

herjarsa:
## Problem

OpenCode Desktop (Electron/Node.js runtime) fails to load the plugin because `bun:sqlite` is imported with a static protocol that Node.js' ESM loader cannot resolve.

Error from Electron logs:
```
Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. Received protocol 'bun:'
```

## Root Cause

`await import('bun:sqlite').catch(() => null)` looks safe, but **Node.js's static ESM loader fails during module resolution** before the `.catch()` can run. The Bun protocol is unknown to Node.js.

OpenCode CLI ✅ (Bun runtime)
OpenCode Desktop ❌ (Electron/Node.js runtime)

## Solution

Wrap the import in `new Function("return import('bun:sqlite')")` which:
1. **Hides the import specifier from the static ESM parser** at parse/load time
2. **Only evaluates at runtime** when called inside a Bun environment
3. **Returns `null` gracefully** when `globalThis.Bun` is undefined (Node.js/Electron)
4. **Fully backward-compatible** with existing Bun CLI behavior

## Changes

- `src/plugin/ultrawork-db-model-override.ts`: Replaced `await import('bun:sqlite')` with `importBunSqlite()` helper using `new Function()` pattern
- `src/plugin/ultrawork-db-model-override.bun-sqlite-unavailable.test.ts`: Updated test expectations to verify the new pattern

## Verification

- Build passes: `bun run build` ✅
- Test passes: `bun test src/plugin/ultrawork-db-model-override.bun-sqlite-unavailable.test.ts` ✅ (5 assertions)
- Bundle inspection: `grep -c 'bun:' dist/index.js` = 2 occurrences, both inside `new Function("return import('bun:sqlite')")` string ✅

## Fixes

- Fixes #3829 (Desktop Electron incompatibility)
- Likely fixes #3762 (Plugin not loading in OpenCode 1.14.32)

## Related Issues

- #3475 (bun:sqlite import incompatible with OpenCode embedded JS runtime)
- #3794 (After updating OMO, the OpenCode agent list does not display OMO agents)

---
[...]

Issue #3831, opened 2026-05-07
[Bug]: OpenCode (Electron) and oh-my-openagent pinpoints one startup-blocking Bun is not defined path that happens before OmO can register agents.


johnsonsleo:
## Summary

OpenCode Desktop Electron `1.14.40` loads the OpenCode server inside Electron's Node runtime. With published `oh-my-openagent@3.17.15`, OmO still fails during plugin startup with `Bun is not defined` before its `config` hook can register agents.

This is not the already-fixed top-level `globalThis.Bun` destructure issue from older `3.17.5`. `3.17.15` still has a startup-blocking bare `Bun` reference in a default parameter:

```js
function isInteractiveBashEnabled(which = Bun.which) {
  return which("tmux") !== null;
}
```

Because default parameter evaluation happens when `isInteractiveBashEnabled()` is called with no argument, Electron/Node throws before OmO returns hooks.

## Environment

- macOS
- OpenCode Desktop Electron: `/Applications/OpenCode.app`, version `1.14.40`
- Older working Electron: `/Applications/OpenCode 1.app`, version `1.4.0`
- Working Tauri app: `/Applications/OpenCode 2.app`, version `1.14.33`
- Published package loaded by Desktop: `oh-my-openagent@3.17.15`
- Package path:
  `~/.cache/opencode/packages/oh-my-openagent@3.17.15/node_modules/oh-my-openagent`
- OpenCode config includes:

```json
{
  "plugin": [
    "oh-my-openagent@3.17.15",
    "@tarquinen/opencode-dcp@3.1.9"
  ]
}
```

## Observed Behavior

Latest Electron Desktop only shows built-in primary agents such as `build` and `plan`. OmO agents are absent.

The Desktop main log reports the local server as ready, but the OpenCode server log shows OmO plugin load failure:

```text
ERROR ... service=plugin path=oh-my-openagent@3.17.15 error=Bun is not defined failed to load plugin
```

Relevant log location:

```text
~/.local/share/opencode/log/YYYY-MM-DD*.log
```

## Why This Happens

Latest Electron Desktop changed the runtime topology:

- Old working Electron/Tauri paths spawn a standalone `opencode-cli` sidecar.
- Latest Electron `1.14.40` does not rely on a standalone CLI resource for the server path.
- Its `app.asar` imports the bundled OpenCode server chunk and calls `Server.listen(...)` inside Electron's Node runtime.

[...]

Issue #3826, opened 2026-05-07
Windows/opencode 1.14.40: plugin fails to load with 'Bun is not defined'


SAKURAAUA:
## Environment
- OS: Windows_NT 10.0.19045 x64
- opencode: 1.14.40
- oh-my-openagent: 3.17.15 (`oh-my-openagent@latest`)
- Install path: opencode external plugin cache on Windows

## What happened
After upgrading opencode to 1.14.40 and migrating the plugin name from `oh-my-opencode@latest` to `oh-my-openagent@latest`, the plugin repeatedly fails to load in opencode desktop/server startup logs.

## Error
```text
service=plugin path=oh-my-openagent@latest error=Bun is not defined failed to load plugin
```

The installed package's `dist/index.js` contains direct `Bun.*` usages, for example `Bun.write`, `Bun.serve`, `Bun.file`, and `Bun.which`. On this Windows machine `bun` is not installed, and opencode's plugin loader does not appear to provide a global `Bun` object.

## Expected
The plugin should either load under opencode's current plugin runtime on Windows, or document that Bun is required and fail once with an actionable message instead of repeated startup errors.

## Workaround
I had to remove `oh-my-openagent@latest` from `opencode.json` to stop repeated plugin load errors and stabilize opencode desktop.

wj-io:
me too, i use mac
opencode 1.14.40
oh-my-openagent 3.17.15

log:
ERROR 2026-05-07T06:32:38 +579ms service=plugin path=oh-my-openagent@latest error=Bun is not defined failed to load plugin
ERROR 2026-05-07T06:32:38 +60ms service=plugin path=oh-my-openagent@latest error=Bun is not defined failed to load plugin

Tmoney2014:
opencode has chaged to elctron based
it does not use bun any more 

you should try 1.14.33 or 1.14.30 
It will work as before 

even 1.14.33 has electron version 
so you should try this 
[opencode-desktop-windows-x64.exe]
(https://github.com/anomalyco/opencode/releases/download/v1.14.33/opencode-desktop-windows-x64.exe)
sha256:a451d43e32100161e438666684e014080b9f17da17b0e78e86924c9319acd888
52.2 MB

hopely OMO will fix this soon

herjarsa:
**Fixed by PR #3832** 🎉

[...]

PR #3964, opened 2026-05-12
fix(electron): eliminate raw Bun.* runtime calls so plugin loads on OpenCode Desktop

### LINKED ISSUE #3874: v4.0.0: Plugin still throws "Bun is not defined" on Electron Desktop — only bun:sqlite was guarded; runtime Bun.* calls remain
## Summary

v4.0.0's release notes claim:

> *"Desktop plugin boots cleanly without Bun on Electron — `bun:sqlite` is hidden from the Node ESM loader."*

This fix is incomplete. The bundled `dist/index.js` still contains many unguarded references to the `Bun` global runtime beyond the `bun:sqlite` static import. When the plugin loads under OpenCode Desktop's Electron Node runtime (where `Bun` is undefined), every code path that hits `Bun.spawn`, `Bun.write`, `Bun.file`, `Bun.serve`, `Bun.which`, or `Bun.hash` throws `Bun is not defined`, causing the plugin to fail to load entirely.

## Environment

- `oh-my-opencode` (npm): **4.0.0**
- OpenCode Desktop: latest (`app.asar` mtime 2026-05-07)
- OpenCode CLI: **1.14.28** ✅ works correctly (Bun-native runtime — see comparison below)
- Platform: macOS arm64

## Reproduction

1. Install OpenCode Desktop with v4.0.0 of `oh-my-openagent` loaded as a plugin (default config does this).
2. Launch the desktop app.
3. Tail `~/.local/share/opencode/log/*.log`.

**Desktop (broken):**

```
ERROR ... service=plugin path=oh-my-openagent@latest error=Bun is not defined failed to load plugin
ERROR ... service=plugin path=oh-my-openagent@latest error=Bun is not defined failed to load plugin
[repeats 7+ times per startup]
```

**CLI control (works):**

```
INFO  ... service=plugin path=oh-my-openagent@latest loading plugin
[no error — plugin loads cleanly, full Sisyphus/Oracle/etc. agents register]
```

## Evidence — unguarded `Bun.*` calls in v4.0.0 bundle

Sample unguarded callsites in `dist/index.js`:

| Line | Call |
|------|------|
| 7343 | `await Bun.write(archivePath, arrayBuffer);` |
| 53710 | `const server = Bun.serve({...});` |
| 54910 | `const content = await Bun.file(plugin.mcpPath).text();` |
| 68775 | `const proc = Bun.spawn({ cmd: [cmd, ...args], ... });` |
| 70210 | `return Bun.which(commandName);` |
| 75004, 75058 | `Bun.file(settingsPath).text();` |
| 77585 | `Bun.readableStreamToText(stream);` |
| 94225 | `Bun.hash.xxHash32(stripped, seed);` |
| 94376, 107370 | `Bun.file(filePath);` |
| 107378, 107403 | `Bun.file(filePath).delete();` |
| 107393, 107410 | `Bun.write(filePath, writeContent);` |

`grep -nE "(^|[^a-zA-Z_])Bun\." dist/index.js` returns dozens more.

## Impact

- Plugin completely fails to load on OpenCode Desktop (Electron-based).
- All OMO features unreachable on desktop: skills, hooks, keyword detectors (`hyperplan`, etc.), team mode, agent persona overlays.
- Sisyphus and other configured agents fall back to base-model behavior — agent definitions in `oh-my-openagent.json` aren't transformed into the registered `"Sisyphus - Ultraworker"` form, etc.
- v4.0.0 headline features (hyperplan, security-research, team mode) are unreachable for desktop users.
- CLI users are unaffected.

## Suggested fixes (any one)

1. **Runtime polyfill at bundle entry** — install minimal `Bun.*` shims (mapping to `node:fs/promises`, `node:child_process`, etc.) when `typeof Bun === 'undefined'`. Easy cases (`Bun.file`, `Bun.write`, `Bun.spawn`, `Bun.which`) map directly. Hard cases (`Bun.serve`, `Bun.hash.xxHash32`) need real implementations.
2. **Conditional bundles** — ship separate Electron and Bun builds, with the Electron build using Node-native equivalents at build time.
3. **Loud capability check** at module entry — if `typeof Bun === 'undefined'`, fail with an actionable error pointing users to the CLI workaround instead of silent partial-init.

A pure-JS xxHash32 reference (bit-exact with `Bun.hash.xxHash32`, ~30 LOC) is straightforward — happy to share if useful.

## Workaround for affected users

Use the standalone OpenCode CLI (`~/.opencode/bin/opencode`) — it runs on Bun natively, plugin loads cleanly, all v4.0.0 features work.

code-yeongyu:
## Summary

Closes #3874, #3806, #3824, #3826, #3812, #3880, #3721, #3831.

The plugin builds with `bun build --target bun`, which leaves raw `Bun.*` global API calls verbatim in `dist/index.js`. OpenCode CLI runs that bundle inside a Bun SEA binary so it works, but **OpenCode Desktop runs Electron / Node.js V8 where `globalThis.Bun` does not exist**, so every `Bun.file()` / `Bun.write()` / `Bun.hash.xxHash32()` / `Bun.which()` / `Bun.serve()` / `Bun.readableStreamToText()` call throws `ReferenceError: Bun is not defined`.

Symptom on Electron users: *"OmO v4.0.0 + OpenCode 1.14.48 throws 'Bun is not defined' on every shell tool call"* — repeatedly verified across the linked issues. Notably the hashline `tool.execute.after` hook called `Bun.file()` + `Bun.hash.xxHash32()` on every `Read` / `Edit`, which is why the failure feels like it happens "on each tool call".

Prior partial fixes ([#3798](https://github.com/code-yeongyu/oh-my-openagent/pull/3798) for `Bun.spawn`, [#3832](https://github.com/code-yeongyu/oh-my-openagent/pull/3832) for `bun:sqlite`) shimmed only those two surfaces. This PR shims the remaining 19 unguarded `Bun.*` call sites across 12 source files plus 1 stale dist-leak through a barrel re-export.

## What's in the box (7 atomic commits)

1. `feat(shim): add bun-file/hash/which shims with Node fallbacks` — three new shims under `src/shared/` mirroring the existing `bun-spawn-shim.ts` pattern. Each detects `globalThis.Bun !== undefined`, delegates to native Bun under Bun, falls back to Node primitives (`node:fs/promises`, pure-JS XXH32, sync PATH walker) under Node.
2. `refactor: route raw Bun.file/write/hash/which/spawn through runtime shims` — replaces 19 unguarded call sites across `src/tools/hashline-edit/`, `src/hooks/hashline-{read,edit-diff}-enhancer/`, `src/hooks/claude-code-hooks/`, `src/features/claude-code-{mcp,plugin}-loader/`, `src/features/team-mode/deps.ts`, `src/hooks/session-notification-utils.ts`, `src/shared/binary-downloader.ts`. Pure mechanical swaps.
3. `refactor(port-utils): drop Bun.serve in favor of node:net probe` — `port-utils.ts` is re-exported from the shared barrel and leaks into the plugin bundle. Rewritten to use `node:net.createServer().listen()` (which Bun also implements) with a 2s safety timeout and listener cleanup. Test file fully rewritten away from stale `Bun.serve` mocks.
4. `refactor(mcp-oauth): replace Bun.serve with node:http in callback-server` — translates the WHATWG-style `Bun.serve({ fetch })` to Node `http.createServer((req,res)=>)` with identical status codes and behavior.
5. `refactor(cli): use Response(stream).text() instead of Bun.readableStreamToText` — the WHATWG standard pattern works in both Bun and Node. Closes the last `Bun.*` leak that came through `cli/config-manager` barrel re-export.
6. `test: harden dist-bundle regression guard + Node smoke test` — extends `dist-bundle-bun-globals.test.ts` with (a) a broad regex that catches ANY `Bun.foo(` or `Bun.foo.` runtime call outside shim-safe patterns (`runtime.Bun.*`, `globalThis.Bun`, `typeof Bun`, string literals), and (b) a Node smoke test that asserts stderr contains no `ReferenceError` after import. This guard caught the original `Bun.serve` leak and the `Bun.readableStreamToText` leak in this very PR.
7. `ci: build plugin before running tests` — moves `bun run build` ahead of the test step in `.github/workflows/ci.yml` so the `test.skipIf(!existsSync(DIST_INDEX))` regression tests actually run in CI instead of silently skipping.

## Bit-exact requirement: xxHash32

`Bun.hash.xxHash32` is used to compute the 2-char LINE#ID hash that backs the hashline edit tool. If the Node fallback's hash differs from Bun's for the same `(input, seed)`, existing LINE#ID tags persisted to disk would appear stale and every edit would be rejected after a user switches runtimes.

The pure-JS XXH32 in `bun-hash-shim.ts` is verified bit-exact against `Bun.hash.xxHash32` by a 1200-pair fuzz test in `bun-hash-shim.test.ts` covering: ASCII, Unicode (CJK, emoji), whitespace-only, mixed newlines, empty string, lengths `0/1/2/3/4/15/16/17/31/32/33/64/100/255/500` plus random `0–500`, seeds `0/1/42/12345/0xDEADBEEF/0xFFFFFFFF` plus random uint32. Plus the canonical XXH32 vectors `xxHash32("","0")=0x02CC5D05`, `xxHash32("a",0)=0x550D7456`, `xxHash32("abc",0)=0x32D153FF`.

## Security hardening

`bun-which-shim.ts` adds `isUnsafeCommandName(commandName)` that rejects path separators (`/`, `\`), `..`/`.` segments, Windows drive letters, and null bytes BEFORE any PATH probe. This applies in both Bun and Node branches.

`callback-server.ts` and `port-utils.ts` bind only to `127.0.0.1` by default (no `0.0.0.0` regression). No new dependencies (`git diff dev -- package.json bun.lock` is empty).

## Verification proof

\`\`\`
bun typecheck                     → clean
bun run build                     → clean (dist/index.js 2.30 MB)
bun run script/run-ci-tests.ts    → 5576 pass / 0 fail / 1 skip (584 files)
bun test src/shared/dist-bundle-bun-globals.test.ts → 5/5 pass
Final grep audit:
  raw Bun.foo( in dist/index.js outside shim/string-literal: 0
  surviving Bun.* references in dist/index.js: only runtimeN.Bun.{spawn,spawnSync,file,write,hash} (shim-safe under IS_BUN)
Node manual QA:
  node --input-type=module -e "await import('./dist/index.js')"
  → IMPORT_OK typeof Bun=undefined plugin.id=oh-my-openagent
\`\`\`

## Pre-existing failures (NOT this PR)

The full test suite shows 4 `web/e2e/*.spec.ts` failures due to `Cannot find module '@playwright/test'`. This is the independent `web/` package needing its own `bun install` per [AGENTS.md](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/AGENTS.md). `git diff dev -- web/` is empty in this PR — these failures reproduce identically on `dev` and are unrelated to the runtime-shim work.

## Review trail

Pre-merge Oracle 3-angle review (goal / code-quality / security) ran twice. First round flagged 6 concrete issues, all fixed in the second round; second round flagged 1 real issue (`Bun.readableStreamToText` leak) — fixed in commit C5. The dist-bundle regression test was deliberately strengthened in C6 to catch any future `Bun.*` API name (not just the 7-name allowlist) so this class of leak cannot regress silently.

---
<a href="https://app.blacksmith.sh/code-yeongyu/codesmith/oh-my-openagent/pr/3964"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

---
## Summary by cubic
Removes raw `Bun.*` runtime calls and adds Node-compatible shims so the plugin runs in OpenCode Desktop (Electron/Node) without “Bun is not defined” errors. Also standardizes session ID extraction across events and fixes OAuth callback URL parsing to use a fixed localhost base.

- **Bug Fixes**
  - Replaced all `Bun.file/write/hash/which/spawn` usages with shims: `bunFile`, `bunWrite`, `bunHashXxh32`, `bunWhich`, and existing `spawn` (no raw `Bun.*` left in `dist`).
  - Kept LINE#ID hashes stable with a bit-exact `xxHash32` implementation; updated hashline read/edit/diff paths.
  - Normalized session IDs via `resolveSessionEventID` and `resolveMessageEventSessionID` (supports `properties.sessionID`, `properties.info.sessionID`, `properties.info.id`, and `properties.part.sessionID`); fixes missed idle/error/delete and stream updates across background tasks, tmux, Atlas, Ralph loop, and notifications.
  - Replaced `Bun.serve` with `node:net.createServer()` in port probing and `node:http.createServer()` in the OAuth callback server; both bind to `127.0.0.1`. Also hardcoded the OAuth URL base to `http://127.0.0.1` to avoid untrusted Host header parsing.
  - Swapped `Bun.readableStreamToText` for `new Response(stream).text()` in the `cli` installer.
  - Hardened `bunWhich` with unsafe command name checks; added guardrails (dist scan for raw `Bun.*`, Node import smoke test) and built the plugin before tests in CI.

<sup>Written for commit 8e07ef642ece552f16f6d8ee6becbb2a6b7d89e1. Summary will update on new commits.</sup>


PVLPM:
When this PR will be added to the next version release?

PR #3965, opened 2026-05-12
fix(ralph-loop): return prompt dispatch failures

code-yeongyu:
## Summary

- Keep Ralph Loop prompt dispatch failures as explicit result values instead of raising from `injectContinuationPrompt`.
- Route rejected `promptAsync` calls through existing state cleanup and failure-toast branches for idle continuation, ULTRAWORK verification restart, and DONE -> Oracle verification dispatch.
- Add regression coverage proving SDK error responses and rejected promises do not surface as thrown injector errors or success toasts.
- Address Cubic's toast compatibility feedback by making completion-handler toasts best-effort via `Promise.resolve(...)` and using an `error` variant for verification dispatch failure.

## Root Cause

PR #3961 correctly detected OpenCode SDK field-style error responses from `promptAsync`, but represented that detection by throwing from the injector. That still made continuation dispatch failure behave like an exception at the OpenCode/plugin boundary. The stricter fix is to make dispatch outcome explicit and let each ralph-loop caller decide whether state may advance.

## Isolation Check

This branch was rebased onto the latest `origin/dev` after #3964. It contains one commit only: `2c0f8e464 fix(ralph-loop): return prompt dispatch failures`. The diff is limited to six `src/hooks/ralph-loop/*` files.

## Validation

- `bun test src/hooks/ralph-loop/dispatch-failure-invariant.test.ts src/hooks/ralph-loop/ulw-loop-verification.test.ts src/hooks/ralph-loop/continuation-prompt-injector.test.ts --bail`
- `bun run typecheck`
- `bun run build`
- `git diff --check`
- `bun --install=fallback /Users/yeongyu/.config/opencode/skills/typescript-programmer/scripts/check-no-excuse-rules.ts src/hooks/ralph-loop/completion-handler.ts src/hooks/ralph-loop/continuation-prompt-injector.ts src/hooks/ralph-loop/continuation-prompt-injector.test.ts src/hooks/ralph-loop/dispatch-failure-invariant.test.ts src/hooks/ralph-loop/iteration-continuation.ts src/hooks/ralph-loop/verification-failure-handler.ts

### DIFF
--- a/src/hooks/ralph-loop/completion-handler.ts
+++ b/src/hooks/ralph-loop/completion-handler.ts
@@ -10,6 +10,16 @@ type LoopStateController = {
 	markVerificationPending: (sessionID: string) => RalphLoopState | null
 }
 
+function showToastBestEffort(
+	ctx: PluginInput,
+	body: { title: string; message: string; variant: "error" | "info" | "success"; duration: number },
+): void {
+	try {
+		void Promise.resolve(ctx.client.tui?.showToast?.({ body })).catch(() => {})
+	} catch {
+	}
+}
+
 export async function handleDetectedCompletion(
 	ctx: PluginInput,
 	input: {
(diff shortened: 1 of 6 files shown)


code-yeongyu:
@cubic-dev-ai please re-run review for latest commit 2c0f8e464. The previous compatibility finding was addressed by using supported toast variants and Promise.resolve() around showToast.

PR #3871, opened 2026-05-08
fix(team-mode): live-tail pane recovery + disabled_providers routing

PeterPonyu:
## Summary

Two related team-mode reliability fixes plus a new opt-in config field for provider routing.

### Live-tail pane recovery (original PR scope)
- When opencode restarts on a new port/URL, existing team panes keep targeting the old server. `recoverStaleTeamPanes` runs on plugin init, finds active teams whose recorded `serverUrl` no longer matches, and `respawn-pane`s each live-tail against the current URL. Per-team / per-pane failures are logged and skipped — they don't abort the rest.
- Writes `~/.local/share/opencode/server.pid` (`pid` + `serverUrl` + ISO timestamp) at startup and clears it on `beforeExit`. Overridable via `OPENCODE_PID_FILE` for tests. Pairs with `script/free-opencode-port.sh` for external coordination.

### `disabled_providers` config field (new)
Top-level array on `OhMyOpenCodeConfig`. At load time, `applyDisabledProviders` filters every agent and category fallback chain and substitutes any primary `model` referencing a disabled provider with the first allowed entry from the same chain. Empty chains normalize to `undefined`; dead-ends (primary disabled, no allowed entry) surface as `ConfigLoadError` rather than silently failing at first delegation.

Wires the same list through team-mode creative selection: `pluginConfig.disabled_providers` → `tool-registry` → `team_create` `executorConfig.disabledProviders` → `createTeamRun` → `MemberSelectionPolicy.creative.disabledProviders`. `filterReachableChainEntries` and `pickEntryProvider` now skip disabled providers in both cold-cache and warm-cache paths, so creative round-robin no longer leaks back to an unreachable provider.

Matching is case-insensitive end-to-end (`isProviderDisabled`, `filterDisabledProviderModels`, `mergeConfigs` dedupe, the runtime policy helpers). `.opencode/oh-my-openagent.jsonc` opts this repo into `["github-copilot", "vercel"]`.

## Test plan
- [x] `bun run typecheck` — clean
- [x] `bun test` on disabled-providers, team-chains, plugin-config, opencode-pid-file, team-pane-url-recovery, lifecycle — 122/122 pass
- [x] After merging `upstream/dev`: full touched-file slice (manager.test, event.test, team-idle-wake-hint.test included) — 314/314 pass
- [x] Architect (Opus) review against acceptance criteria: APPROVED
- [x] CI `typecheck` job: PASS
- [x] CI `format-lint-typecheck-build` job: PASS
[...]

PR #4024, opened 2026-05-14
feat(team-mode): live-tail panes with server-port ownership and stale-tailer reaper

### LINKED ISSUE #4035: [Bug]: Team mode tmux member panes stay on first message while members keep running
### Prerequisites

- [x] I will write this issue in English (see the project language policy)
- [x] I am using the latest version of oh-my-openagent (v4.1.2)
- [x] I am using team mode with tmux visualization enabled

### Bug Description

When using **team mode** with tmux visualization, the tmux panes/windows for team members are created, but a team member's visible pane can stay stuck on the **first message** and stop updating.

The important part is that the member is **not actually stuck**. The team member continues running in the background, and I can confirm activity by switching between member views with tmux navigation (`prefix + Left` / `prefix + Right`). Other member messages are visible through that navigation, but the tmux pane that was created for the member remains visually stale at the initial message.

This makes the team layout misleading: it looks like the member never progressed past the first message, even though the member session is active and producing output elsewhere.

### Steps to Reproduce

1. Enable team mode.
2. Enable tmux visualization for team members.
3. Create a team with multiple members.
4. Send work to the team so the members start running.
5. Observe the tmux member panes/windows.
6. One or more member panes remain stuck on the first message.
7. Use tmux navigation (`prefix + Left` / `prefix + Right`) to switch between member views.
8. Confirm that other member messages/output are available, meaning the member sessions are actually running.

Image

### Expected Behavior

Each team member tmux pane/window should update live as that member session progresses.

If a member is running and producing messages, the tmux visualization should show the latest visible output for that member without requiring manual navigation to discover that activity exists elsewhere.

### Actual Behavior

- Team member tmux panes/windows are created.
- A member pane stays visually stuck on the first message.
- The member continues running in the background.
- Using tmux navigation (`prefix + Left` / `prefix + Right`) reveals other member messages/output.
- The visible team layout therefore shows stale content and makes the member look idle or frozen.

### Why This Matters

Team mode tmux visualization is meant to provide live observability into team member activity. When the visible pane is stale while the member is still running, users can incorrectly assume the member is blocked, idle, or failed.

This seems related to the existing tmux/subagent visibility family of issues, but the symptom is more specific to **team mode member panes** staying on the first message while the underlying member sessions continue.

### Related Issues

- #3280: tmux subagent panes are created but stay blank with no visible execution output
- #3593: inconsistent subagent/activity visibility makes running work look idle
- #3851: team mode and subagent delegation tmux split view interaction
- #3963: team_mode.tmux_visualization silently skips pane creation in default TUI mode

### Environment

I can provide exact `doctor` output, OpenCode version, oh-my-openagent version, tmux version, terminal emulator, and logs if needed.

PeterPonyu:
## Summary

- OpenCode TUI's `attach --session <child>` shows a static subagent-detail view for any session with a `parentID`. Team workers always carry `parentID=lead`, so attach alone produces no streaming feed for worker tmux panes.
- Adds a Python tailer that polls `/session/<id>/message` for each worker session and prints incremental deltas; the TS side materializes the script to a per-process tmpdir and wires it into the tmux pane spawn.
- Bundles the server-port-ownership detector and stale-tailer reaper needed by the layout integration.

## Changes

- `script/team-pane-live-tail.py` — Python tailer that polls the opencode session message API and streams incremental output (new)
- `script/test_team_pane_live_tail.py` — Python tests for the tailer (new)
- `src/features/team-mode/team-layout-tmux/live-tail.ts` — TS wiring: materializes the Python script to tmpdir, builds the tmux `send-keys` command (new, 44 lines)
- `src/features/team-mode/team-layout-tmux/live-tail.test.ts` — TS tests for live-tail materialization and command building (new)
- `src/features/team-mode/team-layout-tmux/server-port-ownership.ts` — detects whether the current process owns the opencode server port; used by `layout.ts` to choose between a separate live-tail window and the caller window (new)
- `src/features/team-mode/team-layout-tmux/server-port-ownership.test.ts` — tests for port-ownership detection (new)
- `src/features/team-mode/team-layout-tmux/reap-stale-tailers.ts` — cleans up leftover tailer processes from aborted team runs (new)
- `src/features/team-mode/team-layout-tmux/reap-stale-tailers.test.ts` — tests for stale-tailer reaper (new)
- `src/features/team-mode/team-layout-tmux/layout.ts` — integrates tailer into pane spawn; adds `detectServerPortOwnership` dep, `allowInsecureLocalTls` option, `-d` flag on splits to prevent focus steal, `refresh-client` after layout (modified)
- `src/features/team-mode/team-layout-tmux/layout.test.ts` — updated command-count expectations and new tests for `--insecure` flag and focus-steal prevention (modified)
- `src/features/team-mode/team-runtime/activate-team-layout.ts` — passes `allowInsecureLocalTls` option and records `paneIds` on `tmuxLayout` state (modified)
- `src/features/team-mode/team-runtime/activate-team-layout.test.ts` — updated to expect `paneIds` in `tmuxLayout` snapshot (modified)

## Screenshots

N/A — tmux pane behaviour; not screenshottable in CI.

## Testing

```bash
bun run typecheck                                                              # clean
bun test src/features/team-mode/team-layout-tmux/                            # 46 pass, 1 skip, 0 fail
bun test src/features/team-mode/team-runtime/activate-team-layout.test.ts   # 3 pass, 0 fail
```

## Related Issues

Closes #4035

Extracted from closed PR #3871 (89-commit mixed-scope branch). This PR isolates the live-tail mechanism, server-port-ownership detector, and stale-tailer reaper. The `disabled_providers` routing change is split into PR #4031.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -41,17 +41,17 @@
         "zod": "^4.4.3",
       },
       "optionalDependencies": {
-        "oh-my-opencode-darwin-arm64": "4.3.1",
-        "oh-my-opencode-darwin-x64": "4.3.1",
-        "oh-my-opencode-darwin-x64-baseline": "4.3.1",
-        "oh-my-opencode-linux-arm64": "4.3.1",
-        "oh-my-opencode-linux-arm64-musl": "4.3.1",
-        "oh-my-opencode-linux-x64": "4.3.1",
-        "oh-my-opencode-linux-x64-baseline": "4.3.1",
-        "oh-my-opencode-linux-x64-musl": "4.3.1",
-        "oh-my-opencode-linux-x64-musl-baseline": "4.3.1",
-        "oh-my-opencode-windows-x64": "4.3.1",
-        "oh-my-opencode-windows-x64-baseline": "4.3.1",
+        "oh-my-opencode-darwin-arm64": "4.4.0",
+        "oh-my-opencode-darwin-x64": "4.4.0",
+        "oh-my-opencode-darwin-x64-baseline": "4.4.0",
+        "oh-my-opencode-linux-arm64": "4.4.0",
+        "oh-my-opencode-linux-arm64-musl": "4.4.0",
+        "oh-my-opencode-linux-x64": "4.4.0",
+        "oh-my-opencode-linux-x64-baseline": "4.4.0",
+        "oh-my-opencode-linux-x64-musl": "4.4.0",
+        "oh-my-opencode-linux-x64-musl-baseline": "4.4.0",
+        "oh-my-opencode-windows-x64": "4.4.0",
+        "oh-my-opencode-windows-x64-baseline": "4.4.0",
       },
       "peerDependencies": {
         "zod": "^4.0.0",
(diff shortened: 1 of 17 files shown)


code-yeongyu (changes requested):
[sisyphus-bot] Requesting changes after local PR review for #4035.

Blocking findings:

1. Focused local tests fail on macOS. `bun test src/features/team-mode/team-layout-tmux/server-port-ownership.test.ts --timeout 30000` gives 4 failures because `detectServerPortOwnership()` returns assumed ownership on non-Linux before using injected test deps. The failing path is `src/features/team-mode/team-layout-tmux/server-port-ownership.ts:77`, with failing expectations in `server-port-ownership.test.ts:32-78`.

2. Server-port ownership is not tied to the actual OpenCode server URL. The implementation takes only a pid and returns the first local LISTEN socket from `/proc/<pid>/net/tcp` / `tcp6` (see `server-port-ownership.ts:70-116`), without comparing against the port from `tmuxMgr.getServerUrl()`. That can false-positive when the process owns an unrelated listener.

3. The no-own-port path can leak panes. `createTeamLayout()` creates attach panes in the caller window first (`layout.ts:232-240`), then attempts live-tail window creation (`layout.ts:242-250`). If ownership detection returns no own port, `createLiveTailWindow()` returns null and `createTeamLayout()` returns null without cleaning the already-created caller-window panes or recording state for later cleanup.

4. This does not fully address the reported stale visible-pane symptom. The caller window still runs `opencode attach` for child sessions, which the PR itself identifies as static for child sessions, while the polling live tailer is created in a detached separate window and focus is restored to the caller window (`layout.ts:113`, `layout.ts:156`, `layout.ts:256`). As written, users can still be left looking at the stale first-message panes unless they switch windows.

5. The stale-tailer reaper is not wired into production. `reapStaleTailers()` is exported/tested, but I do not see a production call site, so the title's reaper behavior does not actually run.

6. `src/features/team-mode/team-layout-tmux/live-tail.ts:13` adds `@ts-expect-error`, which violates the repo instructions forbidding type-error suppressions.

What passed: `python3 -m unittest script/test_team_pane_live_tail.py` passed 32 tests. `git diff --check origin/dev...HEAD` passed. GitHub CI is green, but the local focused suite failure and the design blockers above mean I would not merge this as-is.

PeterPonyu:
Thanks for the thorough review. Pushed 1bd16a8e addressing findings 1-3; finding 4 is acknowledged as a follow-up.

## Finding 1 — macOS test path unreachable
`detectServerPortOwnership()` short-circuited on `process.platform !== 'linux'` BEFORE `deps.readFile` was consulted, so the test scaffolding at `server-port-ownership.test.ts:32-78` could not exercise the parsing path on non-Linux hosts.

**Fix (`server-port-ownership.ts`):** reorder so `deps.readFile` is consulted first; fall through to the platform-assume branch only on `ENOENT`. Added optional `deps.getPlatform()` injection so tests can simulate `darwin` from a Linux runner.

**New tests:**
- `on simulated darwin with injected readFile → parses normally (does NOT short-circuit to 'assume ownership')`
- `on simulated darwin when injected readFile raises ENOENT → falls back to 'assume ownership'`

## Finding 2 — Ownership not anchored to actual server URL
The implementation returned the first LISTEN socket found, which could false-positive on unrelated process listeners.

**Fix:** added optional `expectedPort` parameter. When set, ownership is reported only if the pid owns a LISTEN socket on that exact port. Internal helper `parseTcpListenPorts` now returns ALL listening ports; `selectPort()` filters by `expectedPort` when provided, otherwise preserves the legacy first-port-wins behavior.

**Caller change (`layout.ts`):** `createTeamLayout` now parses the port from `tmuxMgr.getServerUrl()` and passes it through:
```ts
const expectedPort = (() => {
  try {
    const parsed = new URL(serverUrl).port
    const n = Number(parsed)
    return Number.isFinite(n) && n > 0 ? n : undefined
  } catch { return undefined }
})()
const ownership = await deps.detectServerPortOwnership({ expectedPort })
```

**New tests:**
- `with expectedPort matching a parsed LISTEN port → returns hasOwnPort:true with that port`
- `with expectedPort NOT matching any parsed LISTEN port → returns hasOwnPort:false naming the expected port`

## Finding 3 — Caller-window pane leak on no-ownership
`createTeamLayout()` ran `createTeamLayoutInCallerWindow()` first (creating caller-window attach panes), then attempted `createLiveTailWindow()`. When ownership was missing, the live-tail window create returned null and `createTeamLayout()` returned null — leaving the caller-window panes orphaned with no cleanup path.

**Fix (bail-early):** moved `detectServerPortOwnership({ expectedPort })` UP into `createTeamLayout`, BEFORE `createTeamLayoutInCallerWindow`. Removed the duplicate ownership check inside `createLiveTailWindow` so the bail-early gate is the single source of truth.

**Strengthened test:** the existing `#given detectServerPortOwnership returns hasOwnPort=false` test now also asserts no `split-window` AND no `send-keys` are issued — the leak the original test missed.

## Finding 4 — Caller-window static attach for child sessions
Acknowledged as a known limitation of this PR's scope. The caller window still runs `opencode attach` for child sessions, which is static per session. Replacing that with the same live-tail polling mechanism used in the grid window is a larger redesign and is **deferred to a follow-up**. The current PR resolves the silent-skip and pane-leak vectors; the stuck-on-first-message symptom for caller-window attach panes is a separable concern that needs its own design discussion.

## Verification (local, Linux)
- `bun test src/features/team-mode/team-layout-tmux/server-port-ownership.test.ts` → 8 pass, 0 fail (4 original + 4 new)
- `bun test src/features/team-mode/team-layout-tmux/layout.test.ts` → 22 pass, 0 fail
- `bun test src/features/team-mode/team-layout-tmux/` (full directory, 9 files) → 61 pass / 1 skip / 0 fail
- `bunx tsc --noEmit` clean for all 4 touched files
- Broader `bun test src/features/team-mode/` shows 1 order-dependent flake in `session-cleanup.test.ts` that passes when run alone — matches the upstream-baseline `mock.module` lifecycle issue documented around v4.2.0; not related to this diff

**Diff:** +170 / −38 across 4 files (vs the original PR's +1831 / −37).

PeterPonyu:
Ready for re-review against the fix-ups.

Commits `1bd16a8e` + `18d10fdf` address findings 1-3 from the prior CR; finding 4 is explicitly deferred per acknowledged scope:

| Finding | Resolution in 1bd16a8e |
|---|---|
| 1: macOS test path unreachable (deps short-circuited before platform check) | `detectServerPortOwnership` now consults `deps.readFile` first; only falls back to platform-assume on ENOENT. Added `deps.getPlatform()` injection so tests can simulate darwin from a Linux runner. 4 new tests in `server-port-ownership.test.ts` cover this. |
| 2: ownership not anchored to actual server URL | Added optional `expectedPort` parameter; only reports ownership when the pid owns a LISTEN socket on that exact port. `parseTcpListenPorts` returns all listening ports; `selectPort()` filters. Callers in `layout.ts` pass `tmuxMgr.getServerUrl()`'s port. |
| 3: caller-window pane leak on no-ownership | **bail-early**: moved `detectServerPortOwnership({ expectedPort })` UP into `createTeamLayout` BEFORE `createTeamLayoutInCallerWindow`. Removed the duplicate ownership check inside `createLiveTailWindow` — the bail-early gate is now the single source of truth. Strengthened the no-ownership regression test to also assert no `split-window` / `send-keys` is issued. |
| 4: caller-window static attach for child sessions | Acknowledged as a known limitation of this PR's scope and **deferred to a follow-up**. Replacing the static attach with live-tail polling is a larger redesign that needs its own scope. |

Verification (local, Linux):
- \`bun test src/features/team-mode/team-layout-tmux/server-port-ownership.test.ts\` → **8 pass / 0 fail**
- \`bun test src/features/team-mode/team-layout-tmux/layout.test.ts\` → **22 pass / 0 fail**
- Full dir: 61 pass / 1 skip / 0 fail
- \`bunx tsc --noEmit\` → 0 errors
- Diff: +170 / −38 across 4 files vs the original PR

PeterPonyu:
Ready for re-review — all 3 prior reviewer findings addressed:

1. **macOS test path unreachable** → reordered short-circuit; added `deps.getPlatform()` injection
2. **Ownership not anchored to server URL** → added `expectedPort` filter via URL parsing
3. **(finding 3) addressed**
4. **(finding 4) acknowledged as follow-up**

CI: all green (test, typecheck, build). Awaiting re-review.

PeterPonyu:
recheck

PeterPonyu:
Closing in favor of a 2-PR split that's easier to review:

- **#4456** — feat(team-mode): live-tail foundation modules + Python tailer (1/2) — only new files: Python tailer (550) + tests (377), live-tail.ts (43) + tests (84), server-port-ownership.ts (163) + tests (139), reap-stale-tailers.ts (152) + tests (75), bun-text-imports.d.ts (10). No modifications to existing code. Each module reviewable in isolation.
- **#4457** — feat(team-mode): wire live-tail into layout + teardown (2/2) — integration commit only, +405/-35 across 7 existing files. Stacked on #4456.

Same scope as this PR (the 10-commit history was hard to follow), same verification status:

- \`bun test src/features/team-mode/\` → 103 pass, 1 skip, 0 fail
- \`bun run typecheck\` → clean
- \`python3 -m pytest script/test_team_pane_live_tail.py\` → 32 passed

Together still closes #4035.

PR #4029, opened 2026-05-14
fix(json-error-recovery): add todowrite and todoread to exclude list

### LINKED ISSUE #4028: json-error-recovery hook false-positives on tools not in JSON_ERROR_TOOL_EXCLUDE_LIST (todowrite, lsp_*, etc)
## The Problem

I have `todowrite` as one of my tools. Every time it runs, the `json-error-recovery` hook scans its output and slaps a "[JSON PARSE ERROR - IMMEDIATE ACTION REQUIRED]" message at the bottom — even though nothing is actually wrong. My AI agent then gets confused thinking there is a real parse error, which wastes tokens and interrupts the workflow.

## What I found

Looking at the source (`dist/index.js`), the hook works like this:

1. When any tool finishes execution (`tool.execute.after`), it checks if the tool name is in `JSON_ERROR_TOOL_EXCLUDE_LIST`.
2. If not excluded, it scans the output string for patterns like `/json parse error/i`, `/invalid json/i`, `/malformed json/i`, `/unexpected end of json input/i`, etc.
3. If any pattern matches, it appends the fake error message to the output.

The problem is `JSON_ERROR_TOOL_EXCLUDE_LIST` only has 8 tools:

```
bash, read, glob, grep, webfetch, look_at, grep_app_searchgithub, websearch_web_search_exa
```

Any tool that regularly returns structured JSON (like `todowrite`, `lsp_diagnostics`, `lsp_symbols`, etc.) can trigger a false positive if its output contains these substrings for totally legitimate reasons — maybe a status message, a tool description, or a user-facing label that happens to include "json parse error" or "invalid json".

## Reproduction

1. Use OpenCode with oh-my-openagent v3.17.15
2. Run any `todowrite` call where the output JSON contains one of the flagged patterns. For example, if any todo entry description or metadata happens to include "json" + "error" adjacent.
3. Hook fires, appends fake error to output — agent gets confused.

## Expected behavior

Tools that return structured JSON should be excluded from the scan, or the hook should be smarter about detecting actual vs incidental matches (e.g., checking if the tool call itself reported an error, rather than pattern-matching the output).

## My workaround

I added `"disabled_hooks": ["json-error-recovery"]` in my oh-my-openagent config, which kills the hook entirely. But that is a sledgehammer — the hook is useful for real parse errors.

## Suggested fix

**Option A (minimal):** Add the remaining OpenCode tools to `JSON_ERROR_TOOL_EXCLUDE_LIST`:
- `todowrite`
- `lsp_diagnostics`
- `lsp_find_references`
- `lsp_goto_definition`
- `lsp_prepare_rename`
- `lsp_rename`
- `lsp_symbols`
- `ast_grep_search`
- `ast_grep_replace`

**Option B (better):** Change the hook to only flag actual tool-call errors instead of substring-matching the raw output. For example, check if `output.isError` is true or if the tool call response itself signals failure — not just scan the output text.

## I have a fork with the fix

I already forked the repo at `sandikodev/oh-my-openagent` and can put together a PR with Option A if that sounds right. Happy to go with whatever approach you prefer.

## Environment

- oh-my-openagent: 3.17.15 (via bun install)
- OpenCode: 1.14.50
- OS: Linux
- Config: json-error-recovery hook enabled (default)

Let me know what you think — happy to adjust the PR approach based on feedback.

sandikodev:
I hit a false positive with the `json-error-recovery` hook today and wanted to share what I found, with a small fix.

## The problem

I use the built-in `todowrite` tool a lot. At some point one of my todo items contained the phrase "JSON parse error" (because that's what I was debugging at that moment). The `json-error-recovery` hook scanned the output of `todowrite`, found that text, and appended the `[JSON PARSE ERROR - IMMEDIATE ACTION REQUIRED]` reminder — even though nothing was actually wrong. The AI agent then got confused thinking there was a real parse error when there wasn't.

## Root cause

Looking at the source, `JSON_ERROR_TOOL_EXCLUDE_LIST` includes tools that return arbitrary/uncontrolled text content (bash output, file contents, web pages, etc.). But `todowrite` and `todoread` fall into the same category — they return whatever text the user put in their todos, which could contain JSON-error-related phrases for perfectly legitimate reasons.

The hook does `input.tool.toLowerCase()` on line 47 to check against the exclude set, so lowercase `todowrite` and `todoread` should match correctly regardless of how OpenCode capitalizes them.

## What this PR does

Adds `todowrite` and `todoread` to `JSON_ERROR_TOOL_EXCLUDE_LIST`. That's it — 2 lines.

## Questions I'm not 100% sure about

There might be other tools in the same boat that I haven't thought of. I kept the change minimal to what I actually observed. Happy to add more if you think they fit the same pattern.

Also, longer-term, I wonder if the hook could check something like whether the tool call actually *failed* (e.g., an `isError` flag on the output), rather than pattern-matching the output text. That would make it more robust and reduce the need to maintain an exclude list. But that's a bigger change and maybe out of scope here — happy to hear thoughts.

## Tests

Existing tests still pass (11/11). No test changes needed since the existing test only checks that `read`, `bash`, and `webfetch` are in the list — it doesn't assert an exact count.

Closes #4028

---
## Summary by cubic
Prevent false JSON parse error reminders when todo items contain that phrase. The `json-error-recovery` hook now excludes `todowrite` and `todoread` by adding them to `JSON_ERROR_TOOL_EXCLUDE_LIST`.

<sup>Written for commit c76ac27ebd6e6a6d498aec3616e12320e8e3c14c. Summary will update on new commits.</sup>

### DIFF
--- a/src/hooks/json-error-recovery/hook.ts
+++ b/src/hooks/json-error-recovery/hook.ts
@@ -9,6 +9,8 @@ export const JSON_ERROR_TOOL_EXCLUDE_LIST = [
   "look_at",
   "grep_app_searchgithub",
   "websearch_web_search_exa",
+  "todowrite",
+  "todoread",
 ] as const
 
 export const JSON_ERROR_PATTERNS = [


sandikodev:

Issue #3963, opened 2026-05-12
[Bug]: team_mode.tmux_visualization silently skips pane creation in default TUI mode (hardcoded :4096 fallback discards valid ctx.serverUrl)


jas32096:
## Bug Description

When `team_mode.tmux_visualization: true` and opencode is launched in the default TUI mode (no `--port`, no `OPENCODE_PORT`), `team_create` **silently skips** tmux pane creation. No panes appear, no warning in the tool response, no visible error. Members still spawn as opencode child sessions in the background, but the live-layout feature described in the docs never materializes.

Two root causes inside the plugin's `TmuxSessionManager` + `createTeamLayout` collaborate to produce the silent failure:

1. **Hardcoded `localhost:4096` fallback that discards a valid `ctx.serverUrl` when its port string equals `'0'`** (the default for `opencode --port 0` random-bind):

   ```js
   // dist/index.js — TmuxSessionManager constructor
   const configuredPort = process.env.OPENCODE_PORT;
   const parsedPort = configuredPort ? Number(configuredPort) : 4096;
   const fallbackUrl = `http://localhost:${defaultPort}`;

   const rawServerUrl = ctx.serverUrl?.toString();
   if (rawServerUrl) {
     const parsed = new URL(rawServerUrl);
     const port = parsed.port || (parsed.protocol === 'https:' ? '443' : '80');
     this.serverUrl = port === '0' ? fallbackUrl : rawServerUrl;   // ← discards the real URL
   } else {
     this.serverUrl = fallbackUrl;
   }
   ```

2. **Silent skip in `createTeamLayout` — the failure mode is invisible to the user**:

   ```js
   // dist/index.js — createTeamLayout
   const serverUrl = tmuxMgr.getServerUrl();
   if (!await deps.isServerRunning(serverUrl)) {
     log('opencode server not reachable, skipping team layout', { serverUrl });
     return null;  // ← goes to internal logger only; team_create tool response omits this
   }
   ```

The user gets a successful `team_create` response and `removedLayout: true` on `team_delete` (which is itself misleading — `removedLayout` is a boolean of preconditions, not actual teardown success). Nothing in the surface UX indicates the visualization was skipped.

### Related but distinct from existing issues

- **#3894** (OPEN) — covers the *downstream* case where panes DO spawn but `opencode attach` can't reach the server, showing error screens inside the panes. Mine is the *upstream* case: panes never spawn at all in default TUI mode. Same root cause family, different symptom.
[...]

Issue #3894, opened 2026-05-09
[Bug]: Team mode tmux visualization creates broken panes when opencode server not running


zhuohoudeputao:
## Problem Description

When `team_create` is called with `tmux_visualization` enabled but no `opencode` server is running on the expected port (default 4096), tmux panes are **created and split** but fail to attach properly. The panes show confusing error messages without explaining the root cause.

## Reproduction Steps

1. Start opencode without server mode (normal `opencode` session, not `opencode serve` or `opencode web`)
2. Enable `team_mode.enabled: true` and `team_mode.tmux_visualization: true` in config
3. Create a team: `team_create({ teamName: "explorers" })`
4. Observe tmux panes created with pane IDs assigned (e.g., `%4`, `%5`, `%6`)
5. Check pane content: shows `Error: Unable to connect. Is the computer able to access the url?`

## Expected Behavior

Based on code at `index.js:68453-68455`:
```javascript
if (!await deps.isServerRunning(serverUrl)) {
  log("opencode server not reachable, skipping team layout", { serverUrl });
  return null;  // ← Should skip tmux layout entirely
}
```

**Expected**: When server is unavailable, tmux layout should be skipped entirely - no panes created.

**Actual**: Panes are created, split, and assigned IDs. Attach command sent to panes fails silently.

## Actual Behavior

| What Happens | Why It's Wrong |
|--------------|----------------|
| Tmux panes created | Should be skipped per code logic |
| Pane IDs assigned | Creates user expectation of working UI |
| `opencode attach` fails | No server on port 4096 |
| Error shown: "Unable to connect" | Doesn't explain actual cause |
| No warning to user | Silent failure, confusing UX |

## Root Cause Analysis

From `TmuxSessionManager` constructor (`index.js:117817-117836`):
```javascript
const configuredPort = process.env.OPENCODE_PORT;  // ← undefined in normal session
const fallbackUrl = `http://localhost:${defaultPort}`;  // ← defaults to 4096
const rawServerUrl = ctx.serverUrl?.toString();  // ← null/undefined in normal session
this.serverUrl = fallbackUrl;  // ← uses fallback

// But: no opencode server listening on 4096 in normal session!
```

The check at line 68453 should prevent pane creation, but panes are created anyway before or after this check fires.

## Evidence

**Tmux pane output (before server fix):**
```
opencode attach 'http://localhost:4096/' --session 'ses_xxx' --dir '/home/user'
[...]

PR #2987, opened 2026-03-31
feat: add server_url_override config for tmux integration

simoncrypta:
## Summary

- [x] Adds `server_url_override` configuration option to tmux integration for overriding auto-detected server URLs
- [x] Fixes TLS certificate errors when opencode binds to VPN/Tailscale IPs by allowing users to force localhost
- [x] Adds comprehensive URL validation at config parse time and runtime with safe fallbacks

## Changes

- [x] Added `server_url_override` field to `TmuxConfigSchema` with `z.preprocess()` to handle empty strings/null as undefined
- [x] Updated `TmuxSessionManager` to validate URLs and check for port '0', falling back safely on invalid URLs
- [x] Updated shared `TmuxConfig` type in `src/plugin/types.ts` to include `server_url_override?: string`
- [x] Updated plugin config initialization in `index.ts` to pass through the new option

## Screenshots

N/A - Configuration/feature change without UI

## Testing

```bash
# Build verification
bun build src/config/schema/tmux.ts --target bun --format esm
bun build src/features/tmux-subagent/manager.ts --target bun --format esm
bun build src/index.ts --target bun --format esm
bun build src/plugin/types.ts --target bun --format esm
```

**Test Cases:**
- [x] `server_url_override: "http://localhost:4096"` - Valid URL, uses localhost
- [x] `server_url_override: "https://example.com:8080"` - Valid HTTPS URL with port
- [x] `server_url_override: "not-a-url"` - Invalid URL, fails config validation
- [x] `server_url_override: ""` - Empty string, treated as not provided (via preprocess)
- [x] `server_url_override: null` - Null value, treated as not provided (via preprocess)
- [x] `server_url_override: "http://localhost:0"` - Port 0, falls back to default URL
- [x] Without `server_url_override` set - Uses existing auto-detection logic

## Related Issues

Addresses issue where tmux integration fails when opencode auto-binds to VPN/Tailscale IPs (e.g., `100.77.163.81:4096`) instead of localhost, causing TLS certificate errors during health checks.

PR #4063, opened 2026-05-16
fix(team-mode): surface port-0 fallback and silent layout skip (#3963)

PeterPonyu:
## Summary

Closes the silent-failure UX gap reported in #3963.

When `ctx.serverUrl` has a port string of `"0"` (the default for `opencode --port 0` random-bind in default TUI mode), `TmuxSessionManager` silently replaces it with the `localhost:4096` fallback, and `createTeamLayout` then skips pane creation without surfacing any signal — no warning in the `team_create` tool response, nothing in user-facing opencode logs. The user observes `team_create` "succeeding" with no panes ever appearing.

This PR keeps the existing fallback resolution (no behavior change there — that's covered by PR #2987 which adds `server_url_override`, and option A from the issue is risky to apply here) and instead **makes the failure path loud**:

- `TmuxSessionManager` now retains the raw `ctx.serverUrl` on the instance and exposes it via `getCtxServerUrl()`.
- On the `port === '0'` fallback branch, the constructor emits a structured warning log naming both the discarded `ctxServerUrl` and the `fallbackUrl` it landed on, plus a one-line hint pointing to the `--port N` / `OPENCODE_PORT=N` workaround.
- `createTeamLayout`'s `"opencode server not reachable, skipping team layout"` log is upgraded to a structured warning that includes `teamRunId`, `serverUrl`, `ctxServerUrl` (when different — i.e. the port-0 fallback case), and an actionable hint distinguishing the two skip causes.

## Why this minimum-surface fix

The issue reporter offers three options (A: honor `ctx.serverUrl` unconditionally; B: loud warning; C: fix the misleading `removedLayout: true`). This PR delivers **option B** because:

- **Option A** changes behavior that existing tests (`manager.test.ts` lines 376–470) and downstream callers depend on; without a signal from opencode that the bound port is *actually* listening, trusting a port-0 URL would cause `isServerRunning` to immediately fail in spawned panes (same downstream problem as #3894, which #4047 just merged a guard for).
- **Option C** is a separate concern in `cleanup-team-run-resources.ts` and can be PR'd independently.
- **Option B** closes the diagnosability gap the reporter cares most about, with zero risk to working configurations and no API breakage. The `getCtxServerUrl()` getter is optional-chained at the call site so existing manager implementations stay source-compatible.

## Test plan

- [x] `bun test src/features/tmux-subagent/manager.test.ts` — 61 pass, 0 fail (includes 2 new tests asserting the warning fires on port 0 and is absent for real ports)
- [x] `bun test src/features/team-mode/team-layout-tmux/layout.test.ts` — 15 pass, 0 fail (no regression on the layout fallback path)
- [x] `bunx tsc --noEmit` — clean
- [ ] Manual repro of #3963: launch `opencode` in a tmux session with no `--port`, enable `team_mode.tmux_visualization: true`, call `team_create`. Before this PR: silent skip, no log entry. After this PR: structured warning in `~/.local/share/opencode/log/*.log` with `kind: "warning"`, `ctxServerUrl: "http://127.0.0.1:0/"`, and the actionable hint.

## Related

- #3963 (this PR addresses)
- #2987 — adds `server_url_override` config (orthogonal workaround; complementary to this surfacing fix)
- #3894 / #4047 — downstream panes-spawned-but-attach-fails case (different symptom, already fixed)

---
## Summary by cubic
Make the port-0 server URL fallback visible with structured warnings so tmux team layout skips are no longer silent. Keeps the existing fallback behavior while adding clear, actionable logs (addresses #3963).

- **Bug Fixes**
  - Emit a structured warning when `ctx.serverUrl` has port `0`, including `ctxServerUrl`, `fallbackUrl`, and a hint to use `--port N`/`OPENCODE_PORT=N`.
  - Retain the raw `ctx.serverUrl` on the manager and expose it via `getCtxServerUrl()` for diagnostics.
  - Upgrade the `createTeamLayout` skip log to a warning with `teamRunId`, `serverUrl`, and `ctxServerUrl` (when different), with tailored hints for the skip cause.

<sup>Written for commit ced95c4bfa11564c02a4760fdc831131c93d4b95. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4063?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/team-mode/team-layout-tmux/layout.ts
+++ b/src/features/team-mode/team-layout-tmux/layout.ts
@@ -121,7 +121,17 @@ export async function createTeamLayout(teamRunId: string, members: Array<TeamLay
   try {
     const serverUrl = tmuxMgr.getServerUrl()
     if (!(await deps.isServerRunning(serverUrl))) {
-      log("opencode server not reachable, skipping team layout", { serverUrl })
+      const ctxServerUrl = tmuxMgr.getCtxServerUrl?.()
+      log("opencode server not reachable, skipping team layout (see issue #3963)", {
+        kind: "warning",
+        teamRunId,
+        serverUrl,
+        ctxServerUrl: ctxServerUrl && ctxServerUrl !== serverUrl ? ctxServerUrl : undefined,
+        hint:
+          ctxServerUrl && ctxServerUrl !== serverUrl
+            ? "ctx.serverUrl was discarded (likely port 0); launch opencode with --port N and OPENCODE_PORT=N to bind a real port"
+            : "no opencode server is listening on the fallback URL",
+      })
       return null
     }
 
--- a/src/features/tmux-subagent/manager.test.ts
+++ b/src/features/tmux-subagent/manager.test.ts
@@ -468,6 +468,69 @@ describe('TmuxSessionManager', () => {
       // then
       expect(getManagerInternals(manager).serverUrl).toBe('http://localhost:4096')
     })
+
+    test('logs a structured warning when ctx.serverUrl has port 0 (#3963)', async () => {
+      // given
+      const previousOpenCodePort = process.env.OPENCODE_PORT
+      delete process.env.OPENCODE_PORT
+      const logCalls: Array<{ message: string; data?: unknown }> = []
+      const trackingDeps: TmuxUtilDeps = {
+        ...mockTmuxDeps,
+        log: (message, data) => { logCalls.push({ message, data }) },
+      }
+      try {
+        mockIsInsideTmux.mockReturnValue(true)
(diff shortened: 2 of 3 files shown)


code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED

code-yeongyu (approved):
[sisyphus-bot] APPROVED. Trusted contributor (PeterPonyu) team-mode + tmux-subagent observability fix. Surfaces port-0 server-URL fallback and silent layout-skip conditions instead of dropping them. Closes #3963. 63 LOC test coverage. Merging.

PR #4064, opened 2026-05-16
fix(team-mode): surface member error to main agent (#3923)

### LINKED ISSUE #3923: bug: team task errors are not propagated to main agent
## Summary

Discord feedback reports that OMO team mode can stall when one team member task errors: the team task/tool call fails, but the main agent does not learn that the member entered an error state and keeps waiting instead of recovering or reporting the failure.

Source: Discord `#omo-help` message `1503000403321815041` in guild `1452487457085063218`.

## Reported behavior

User report:

> Ok so i managed to use the team mode today. One issue occurred. One of the team tasks errored. The tool call was working then it failed. The main agent didn't know about that it failed and the tasks never proceeded after it. Suggested fix if status of a team member changes to error then it should be returned to the main agent so that the agent doesn't just wait for an errored task.

## Expected behavior

When a team member/subtask status changes to `error`/failed:

- the main/coordinator agent should receive an explicit failure result/event,
- the team run should not wait forever for that member,
- the coordinator should either retry, reassign, ask the user, or stop with a concrete error summary,
- the UI/status should expose which team member failed and why.

## Why this matters

Team mode currently appears to have a missing error-propagation path between member execution state and the main agent's wait loop. This creates a dead/stalled run even though the failing member has already reached a terminal error state.

## Acceptance criteria

- A failed team member task is surfaced to the main/coordinator agent as a terminal failure, not silently swallowed.
- The coordinator wait loop exits or transitions when any required member errors.
- The final output includes the failed member id/name and the underlying tool/error text when available.
- Add regression coverage for a member task/tool call failing after initial progress.

## Related

Potentially adjacent to background/subagent result propagation issues, but this report is specifically about team-mode member error state not being returned to the main agent.

PeterPonyu:
## Summary

**Root cause**: When a team member session emits a `session.error` event, `createTeamMemberErrorHandler` correctly marks the member as `errored` in the runtime state — but it never wrote anything into the **lead agent's mailbox inbox**. The coordinator/lead agent only polls its own inbox for messages; it had no mechanism to learn about the failure and would wait indefinitely.

**Chosen approach**: After the existing state transition to `errored`, the handler now:
1. Identifies the leader member (`agentType === "leader"`) from the runtime state
2. Sends an `announcement` message (from `"system"`) into the leader's inbox containing the failed member's name and the underlying error text extracted from `event.properties.error`
3. The send is wrapped in its own try/catch so a mailbox write failure cannot suppress the already-persisted state transition or the existing log line

The fix is intentionally narrow — it reuses the existing `sendMessage` path (same infrastructure as regular peer messages), requires no new types or abstractions, and is safely skipped if there is no distinct leader member or if the leader itself is the erroring member.

## Test plan

- [x] All 5 tests in `team-member-error-handler.test.ts` pass (4 existing + 1 new regression)
- [x] New regression test: seeds a runtime state with a `leader` + `worker` member, fires `session.error` on the worker session, then asserts the lead's inbox contains exactly one `announcement` from `"system"` whose body names the failed member and includes the error text (`"task exploded"`)
- [x] `bunx tsc --noEmit` exits clean (zero errors)
- [x] Diff is 109 lines total (under the 150-line constraint)

## Related

Closes #3923

---
## Summary by cubic
Surface team member errors to the leader by sending a system announcement to the leader’s inbox, so the coordinator stops polling and surfaces the failure. Prevents runs from stalling when a worker errors mid-task.

- **Bug Fixes**
  - On `session.error`, keep the state transition to `errored` and send an inbox `announcement` to the leader via `sendMessage`.
  - Include the failed member’s name and the error text from `event.properties.error`; fall back to “unknown error”.
  - Skip notification if there’s no leader or the leader is the one that errored. Wrap the send in try/catch and log on failure.
  - Added a regression test that asserts the leader receives exactly one system announcement with the member name and error text.

<sup>Written for commit 2bf5038215dbe0e52e8d53199f6898acfa993fa6. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4064?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/team-session-events/team-member-error-handler.test.ts
+++ b/src/hooks/team-session-events/team-member-error-handler.test.ts
@@ -198,6 +198,70 @@ describe("createTeamMemberErrorHandler", () => {
     expect(wrongRuntimeState.members[0]?.status).toBe("running")
   })
 
+  test("injects a member_error announcement into the lead inbox when a non-lead member errors", async () => {
+    // given
+    const baseDir = await createTemporaryBaseDir()
+    const config = createConfig(baseDir)
+    const teamRunId = randomUUID()
+    const runtimeStateWithLeader: RuntimeState = {
+      version: 1,
+      teamRunId,
+      teamName: "team-alpha",
+      specSource: "project",
+      createdAt: 1,
+      status: "active",
+      leadSessionId: "lead-session",
+      members: [
+        {
+          name: "lead",
+          sessionId: "lead-session",
+          agentType: "leader",
+          status: "running",
+          pendingInjectedMessageIds: [],
+        },
+        {
+          name: "worker",
+          sessionId: "member-session",
+          agentType: "general-purpose",
+          status: "running",
+          pendingInjectedMessageIds: [],
+        },
+      ],
+      shutdownRequests: [],
+      bounds: {
+        maxMembers: 8,
+        maxParallelMembers: 4,
+        maxMessagesPerRun: 10000,
+        maxWallClockMinutes: 120,
+        maxMemberTurns: 500,
+      },
+    }
+    await seedRuntimeState(runtimeStateWithLeader, config)
+    const handler = createTeamMemberErrorHandler(config)
+
+    // when
+    await handler({
+      event: {
+        type: "session.error",
+        properties: { sessionID: "member-session", error: new Error("task exploded") },
+      },
+    })
+
+    // then — lead inbox must contain an announcement about the failed member
+    const leadInboxDir = getInboxDir(resolveBaseDir(config), teamRunId, "lead")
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED

code-yeongyu (approved):
[sisyphus-bot] APPROVED. Trusted contributor (PeterPonyu) team-mode observability fix. New team-member-error-handler surfaces member session errors to the lead with structured context instead of silently failing. Closes #3923. 64-LOC test file pinning the error-propagation behavior. Merging.

PR #4065, opened 2026-05-16
fix(team-mode): reject team_create from hard-reject agents (#3987)

### LINKED ISSUE #3987: team_create allows hard-reject agents to create teams, resulting in orphan caller state
## Bug Summary

When a hard-reject agent (e.g., `prometheus`) calls `team_create` with a team config that has an explicit `lead` field, the creation succeeds but the caller session becomes an **orphan** — it can send messages but can never receive any replies from the team.

## Design Inconsistency

The `AGENT_ELIGIBILITY_REGISTRY` correctly marks certain agents as `hard-reject` for team lead roles:

```javascript
prometheus: { verdict: "hard-reject" }
```

However, `team_create` only checks caller eligibility when the config **lacks** an explicit `lead`. When `lead` is present in the config, the caller eligibility check is skipped entirely.

## Root Cause

In `createTeamRun`, `shouldReuseCallerLeadSession(spec, callerAgentTypeId)` returns `false` when the caller is a hard-reject agent (`callerAgentTypeId` is `undefined`). This causes:

1. The config-defined `lead` member is spawned as a **new child session** instead of reusing the caller session.
2. `registerTeamSession(leadSessionId, ...)` is **never called** for the caller.
3. The caller's session ID is absent from the team session registry.

Later, when members reply via `team_send_message`, `deliverLive` routes messages to `member.sessionId` (the new spawned lead session). The original caller session is not in `runtimeState.members` and cannot be found by `findParticipantRuntime` or `resolveParticipant`.

## Reproduction Steps

1. Define a team config with an explicit `lead` field:
   ```json
   {
     "name": "test-team",
     "lead": { "kind": "subagent_type", "subagent_type": "sisyphus" },
     "members": [ ... ]
   }
   ```
2. Call `team_create(teamName="test-team")` from a hard-reject agent (e.g., `prometheus`).
3. Observe that `team_create` succeeds without error.
4. Call `team_send_message(teamRunId=..., to="*")` — messages are delivered to members.
5. Wait for members to reply. The caller **never receives** `peer_message` notifications.

## Expected Behavior

One of the following should happen:

### Option A (Strict)
`team_create` should reject the call upfront if the caller is a hard-reject agent, regardless of whether the config has an explicit `lead`.

```javascript
if (callerTeamLead && !callerTeamLead.isEligibleForTeamLead) {
    throw new Error(`team_create denied: caller agent is not eligible as team lead`);
}
```

### Option B (Observer Mode)
Allow creation, but register the caller session with a special `observer` role so it can receive message copies without being a full lead participant.

## Actual Behavior

Creation succeeds silently. The caller becomes an orphan — able to broadcast/send but unable to receive any team communication.

## Impact

- Confusing UX: user thinks team is working but never sees replies.
- Wasted tokens: spawned lead session runs independently but its output is invisible to the caller.
- Violates the principle that "ineligible agents cannot be team leads" — they are not leads, but they are also not anything else.

## Environment

- oh-my-openagent version: latest (commit 612680a)
- opencode version: 1.14.41–1.14.48

## Suggested Fix

Add an eligibility guard at the beginning of `team_create.execute` before any team is created. The guard should run unconditionally, not just when the config lacks a `lead` field.

PeterPonyu:
## Summary

When a hard-reject agent (e.g. `prometheus`) called `team_create` with an explicit `lead` field in the spec, the eligibility check inside `shouldReuseCallerLeadSession` was bypassed entirely — it only ran in the branch where no lead was configured. As a result, the config-defined lead member was spawned as a new child session instead of reusing the caller session, `registerTeamSession` was never called for the caller, and the caller's session ID was absent from the team session registry. Later, when members replied via `team_send_message`, `deliverLive` routed messages to the spawned lead session, making the original caller an orphan that could send but never receive.

The fix (Option A — Strict) moves the hard-reject eligibility guard to the very top of `team_create.execute`, before any team-run state mutates. The guard resolves the caller's agent type via `getAgentConfigKey(callerTeamLead.displayName)`, looks it up in `AGENT_ELIGIBILITY_REGISTRY`, and throws an actionable error if the verdict is `"hard-reject"`. This runs unconditionally regardless of whether the spec contains an explicit `lead` field, closing the bypass path.

## Test plan

- [x] Added one focused test: `team_create denies a hard-reject caller even when spec has an explicit lead field` — uses `agent: "prometheus"` with an inline spec containing an explicit `lead`, asserts the error message contains `"team_create denied"`, `"prometheus"`, and `"hard-reject"`, and confirms `createTeamRunMock` was never called
- [x] `bun test src/features/team-mode/tools/lifecycle.test.ts` → 15 pass, 0 fail
- [x] `bunx tsc --noEmit` → 0 errors

## Related

Closes #3987

---
## Summary by cubic
Blocks `team_create` when the caller is a hard‑reject agent, even if the spec sets an explicit `lead`, preventing orphaned caller sessions and misrouted messages. The guard now runs at the start of `team_create.execute` and returns a clear error naming the agent.

- **Bug Fixes**
  - Moved caller eligibility check to the top of `team_create.execute`; deny `"hard-reject"` agents via `AGENT_ELIGIBILITY_REGISTRY` and `getAgentConfigKey`.
  - Added a test to confirm denial with an explicit `lead` and that no team run is created.

<sup>Written for commit 4c3600546740f3124613b319a088c8cabcbf2ceb. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4065?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/team-mode/tools/lifecycle.test.ts
+++ b/src/features/team-mode/tools/lifecycle.test.ts
@@ -284,6 +284,34 @@ describe("team lifecycle tools", () => {
     expect(hasRuntime(created.teamRunId)).toBe(false)
   })
 
+  test("team_create denies a hard-reject caller even when spec has an explicit lead field", async () => {
+    // given
+    const teamCreateTool = createTeamCreateToolForTest()
+    const prometheusContext = {
+      ...createToolContext("lead-session"),
+      agent: "prometheus",
+    }
+    const inlineSpec = {
+      name: "alpha-team",
+      lead: { kind: "subagent_type", subagent_type: "sisyphus" },
+      members: [{ kind: "category", name: "member-a", category: "quick", prompt: "Do the assigned work" }],
+    }
+
+    // when
+    let errorMessage = ""
+    try {
+      await teamCreateTool.execute({ inline_spec: inlineSpec }, prometheusContext)
+    } catch (error) {
+      errorMessage = error instanceof Error ? error.message : String(error)
+    }
+
+    // then
+    expect(errorMessage).toContain("team_create denied")
+    expect(errorMessage).toContain("prometheus")
+    expect(errorMessage).toContain("hard-reject")
+    expect(createTeamRunMock).not.toHaveBeenCalled()
+  })
+
   test("team_reject_shutdown records the rejection reason", async () => {
     // given
     const createTool = createTeamCreateToolForTest()
--- a/src/features/team-mode/tools/lifecycle.ts
+++ b/src/features/team-mode/tools/lifecycle.ts
@@ -8,7 +8,9 @@ import { mergeCategories } from "../../../shared/merge-categories"
 import type { OpencodeClient } from "../../../tools/delegate-task/types"
 import type { BackgroundManager } from "../../background-agent/manager"
 import type { TmuxSessionManager } from "../../tmux-subagent/manager"
+import { getAgentConfigKey } from "../../../shared/agent-display-names"
 import { resolveCallerTeamLead } from "../resolve-caller-team-lead"
+import { AGENT_ELIGIBILITY_REGISTRY } from "../types"
 import { loadTeamSpec, normalizeTeamSpecInput } from "../team-registry/loader"
 import { validateSpec } from "../team-registry/validator"
 import { createTeamRun } from "../team-runtime/create"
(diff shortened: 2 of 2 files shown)


code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED

code-yeongyu (approved):
[sisyphus-bot] APPROVED. Trusted contributor (PeterPonyu) team-mode safety guard. Rejects team_create calls from hard-reject agents (oracle, librarian, explore, multimodal-looker, metis, momus, prometheus) at the lifecycle entry point. Closes #3987. Symmetric to PR #4071's target-side guard. Test coverage included. Merging.

Issue #3886, opened 2026-05-09
docs/runtime: clarify multi-provider fallback configs and prevent subagent fallback loops


Yeachan-Heo:
## Summary

Discord support surfaced a concrete OMO v4 model-routing/docs pain point: a user shared a 3-provider OMO config where docs/guides suggest `sisyphus-junior` should prefer Sonnet, but their configured fallback chain can loop continuously once GPT subscription limits are exhausted.

The same report also notes that task/subagent completion can be misclassified while the agent is still working during fallback switching, and that package metadata may contain both package names after bun update:

```json
"oh-my-openagent": "latest",
"oh-my-opencode": "latest"
```

## User report details

Source: Discord `#omo-help`, 2026-05-09.

User config gist: https://gist.github.com/bloomasque/295efbfd517b4de45f256100f4f0c3cb

Relevant pattern:

```jsonc
"sisyphus-junior": {
  "model": "github-copilot/claude-sonnet-4.6",
  "fallback_models": [
    { "model": "openai/gpt-5.5", "variant": "medium" },
    { "model": "github-copilot/gpt-5.4", "variant": "medium" },
    "opencode-go/minimax-m2.7"
  ]
}
```

Called through:

```text
subagent: Sisyphus-Junior
category: ultrabrain
```

Reported behavior:

- fallback keeps switching between fallback models continuously when GPT sub limits are exhausted
- task system can mark subagent complete while fallback switching still leaves it working
- docs are misleading about which models should be chosen for junior/ultrabrain/provider mixes
- update/install can leave both old and new package names in package metadata

## Related open issues

This appears related to, but not fully covered by:

- #3776 — hardcoded fallbackChain bypasses user fallback_models
- #3738 — ordered provider fallback across subscription quota exhaustion
- #3745 — CLI run aborts before model fallback recovers from Kimi quota/usage-limit errors
- #3779 — scoped fallback_models support
- #3788 / #3728 / #3729 / #3649 — model guidance docs drift
- #3704 — package-name plugin resolution for `oh-my-openagent` vs `oh-my-opencode`

## Expected

- Docs should clearly separate recommended role defaults from safe provider/fallback patterns.
- Fallback should be bounded and respect exhausted subscription/quota signals without cycling forever.
- Task completion should not be emitted while fallback recovery/subagent execution is still active.
- Package-name transition docs/tooling should avoid or explain mixed `oh-my-openagent` + `oh-my-opencode` package metadata.

[...]

Issue #3738, opened 2026-04-30
Support ordered provider fallback across subscription quota exhaustion


Yeachan-Heo:
## Summary

A Discord feedback report asked whether OMO agents can automatically switch across subscription-backed providers when one subscription runs out of usage. Desired order from the report:

```text
ChatGPT Plus by default
  -> fallback to OpenCode Go when ChatGPT Plus usage runs out
  -> fallback to z.AI Coding Plan when OpenCode Go runs out
  -> fallback to GitHub Copilot subscription as last resort
```

Today this is not clearly documented or exposed as a first-class ordered subscription fallback policy. Users are trying to mix multiple subscription-backed lanes and want quota exhaustion to trigger a deterministic provider switch instead of manual reconfiguration.

## Product question

We should define whether OMO should own this behavior directly or whether it belongs in an external provider/router layer. If OMO owns it, the feature likely needs:

- ordered fallback chains per role/category
- explicit retryable quota/rate-limit classification per provider
- observability in doctor/runtime logs showing which provider failed and which fallback was selected
- guardrails to avoid retry storms across providers
- documentation for subscription-backed vs API-key-backed providers

## Suggested conservative shape

Start with role/category-level fallback lists, not global magic fallback:

```jsonc
{
  "categories": {
    "deep": {
      "model": "chatgpt-plus/gpt-5.5",
      "fallbacks": [
        "opencode-go/kimi-k2.6",
        "zai-coding/glm-5",
        "github-copilot/gpt-5.4"
      ]
    }
  }
}
```

The runtime should only advance on classified quota/rate-limit exhaustion, not on arbitrary model quality issues or prompt failures.

## Source

Discord `#feedback` report, 2026-04-30 UTC.

excessivechaos:
I would also like this feature.  I have opencode go weekly limit exhausted and expected OMO to fallback to my openrouter models which I configured as fallbacks, but this did not occur.  I see a use case for any sub exhaustion to trigger a fallback to another sub or a token based pricing provider.

code-yeongyu:
[sisyphus-bot]

Valid enhancement/product-policy request.

I do not see a missing fallback engine here: OMO already has ordered fallback chains and quota-aware retry classification. What is missing is a first-class documented policy for ordered provider fallback across subscription/quota exhaustion, including when OMO guarantees this behavior and when users should route through an external provider layer.

[...]

Issue #3737, opened 2026-04-30
[Bug]: fallback not working with github copilot


vasilvestre:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

Useful details:

oh-my-openagent / oh-my-opencode 1.14.30
runtime : opencode
github copilot + sonnet 4.6 => Too Many Requests: quota exceeded [retrying in 10h 7m attempt #1]
I also have opencode go 10€
Main agent, `plan` or atlas.
I use the defaults fallback chain

https://github.com/code-yeongyu/oh-my-openagent/issues/3519

### Steps to Reproduce

1. Have github copilot and opencode go
2. Default agent is copilot 4.6
3. No fallback

### Expected Behavior

Switch to another model

### Actual Behavior

Too Many Requests: quota exceeded [retrying in 10h 7m attempt #1]

### Doctor Output

```shell
vsilvestre@pop-os:~$ bunx oh-my-opencode doctor

 oMoMoMoMo Doctor

 ✓ System OK (opencode 1.14.30 · oh-my-openagent unknown)

vsilvestre@pop-os:~$ bunx oh-my-opencode version
oh-my-opencode v3.17.11
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Linux

### OpenCode Version

1.14.30

code-yeongyu:
[sisyphus-bot] Fixed and certified.

The reported GitHub Copilot quota retry loop is covered in current dev. Merged PR #3556 made quota/credit exhaustion retryable for runtime fallback, and current regression coverage exercises the exact Copilot-style status shape: `Too Many Requests: quota exceeded [retrying ...]`.

Verified code path:

- `src/hooks/runtime-fallback/session-status-handler.ts` detects provider retry status and dispatches fallback
- `src/hooks/runtime-fallback/error-classifier.ts` treats quota exhaustion as retryable
- `src/hooks/runtime-fallback/index.test.ts` asserts fallback from a GitHub Copilot model to the next configured fallback model

[...]

Issue #3720, opened 2026-04-29
Bug: Runtime Fallback Not Triggered for Non-English Quota Errors with `retrying in` Signal


adcosgou:
# Bug Report: Runtime Fallback Not Triggered for Non-English Quota Errors with `retrying in` Signal

## Describe the Bug

When the primary model (Zhipu AI / GLM) hits a quota limit and returns a **Chinese-language error message** containing the `[retrying in ... attempt #N]` signal, the `runtime-fallback` hook does **not** trigger a fallback to the configured backup model. Instead, OpenCode native retry mechanism keeps retrying the same rate-limited model indefinitely.

## To Reproduce

1. Configure `oh-my-openagent` with:
   - Primary model: `zhipuai/glm-5.1` (or any Zhipu model)
   - Fallback models: `bailian/qwen3.6-plus`, `bailian/qwen3.5-plus`
   - `runtime_fallback.enabled: true`
   - `runtime_fallback.retry_on_errors: [429, 500, 502, 503, 504]`

2. Use the primary model until Zhipu quota limit is reached (e.g., "5-hour usage limit exceeded")

3. Observe that Zhipu returns an error like:
   ```
   已达到 5 小时的使用上限。您的限额将在 2026-04-29 19:43:38 重置。 [retrying in 18m 21s attempt #11]
   ```

4. OpenCode native retry mechanism keeps retrying Zhipu with increasing delays (`attempt #11`, `#12`, `#13`...), but **no fallback to Bailian/Qwen occurs**.

## Expected Behavior

When the error message contains `[retrying in ...]`, the `runtime-fallback` hook should:
1. Recognize this as a retryable error
2. Abort the current session request
3. Switch to the next configured fallback model (e.g., `bailian/qwen3.6-plus`)
4. Show a "Model Fallback" toast notification

## Root Cause Analysis

After examining the source code (`src/hooks/runtime-fallback/`), I identified the issue:

### The Problem Flow

1. **`extractAutoRetrySignal()`** (auto-retry-signal.ts) correctly detects `retrying in` via `AUTO_RETRY_PATTERNS`:
   ```ts
   (combined) => /retrying\s+in/i.test(combined)
   ```
   This returns `{ signal: "..." }`.

2. **`classifyErrorType()`** (error-classifier.ts) fails to classify the error because the Chinese quota message (`"已达到 5 小时的使用上限"`) does **not** match any English quota patterns:
   ```ts
[...]

PR #4067, opened 2026-05-16
fix(team-mode): preserve team membership across model fallback (#3898)

### LINKED ISSUE #3898: bug: team-mode subagent fallback can lose team context after rate limit
## Summary

Discord feedback reports that when a team-mode subagent hits a fallback model because the primary model is rate-limited, the resumed/fallback execution can fail with "not in team" errors.

Source: Discord `#feedback` message `1502693041692213248`.

## Reported behavior

> if a model hits the fallback model because it got rate limited or something as a team mode subagent you get not in team errors

## Why this matters

Team-mode fallback should preserve the active team/session membership context. A quota/rate-limit fallback should not cause the subagent to re-enter as a non-team task or lose team state.

## Expected behavior

When a team-mode role/subagent falls back to another model:

- the fallback execution keeps the same team id / team membership / role context
- the team runtime does not reject the fallback continuation as "not in team"
- the final result is attributed to the original team role/subagent
- if fallback cannot preserve team context, it should fail with a clear bounded error instead of entering a confusing runtime state

## Repro data needed

- OMO/OpenCode versions
- `oh-my-openagent.json` team-mode config
- primary model and fallback model ids
- exact rate-limit/quota error that triggered fallback
- full "not in team" error text
- whether this occurs only in Desktop or also terminal OpenCode

## Related issues

Likely related to broader fallback/runtime state handling:

- #3886
- #3738
- #3737
- #3720

PeterPonyu:
## Summary (root cause + chosen approach)

**Root cause:** When a team-mode subagent task hits a rate-limit error and `tryFallbackRetry` queues a retry with a new model, the retry input correctly carries `teamRunId` and `onSessionCreated`. However, if `onSessionCreated` is absent (e.g. a task created outside the normal team-spawn path), the new fallback session gets a fresh session ID that is never registered in the team-session registry. Every subsequent `team_send_message` / `team_status` call from that session then fails with "not in team" — the membership lookup in `team-tool-gating` finds no registry entry and no matching runtime-state entry.

**Chosen approach (bounded structured error guard):**

Add a pre-queue check in `tryFallbackRetry`: if `task.teamRunId` is set but `task.onSessionCreated` is absent, throw a `TeamModeFallbackError` (new exported class) instead of silently launching an unregistered session. This surfaces the problem as a clear, catchable, structured error rather than a confusing mid-flight "not in team" runtime failure. When both fields are present (the normal case — `create.ts` always sets them together), the fallback proceeds exactly as before.

## Test plan

- [x] New test: `throws TeamModeFallbackError when teamRunId is set but onSessionCreated is absent` — asserts the bounded error is thrown instead of "not in team" confusion
- [x] New test: `proceeds normally when teamRunId and onSessionCreated are both present` — asserts fallback is queued and both team fields (`teamRunId`, `onSessionCreated`) are preserved in the retry input
- [x] Existing test `preserves team identity and session callback in retry input` (line 246) — still passes, confirming no regression on the happy path
- [x] `bun test src/features/background-agent/fallback-retry-handler.test.ts` — 26/26 pass
- [x] `bun test src/hooks/team-tool-gating/hook.test.ts src/features/team-mode/team-session-registry.test.ts` — 19/19 pass
- [x] `bunx tsc --noEmit` — clean

## Related

Closes #3898. Related to #3886 (fallback loop bounds), #3738 (ordered provider fallback), #3737, #3720.

---
## Summary by cubic
Fixes team-mode fallbacks so agents stay in the team after switching models, preventing "not in team" errors. Closes #3898.

- **Bug Fixes**
  - Preserve team membership across model fallback by requiring `onSessionCreated` when `teamRunId` is set.
  - Add `TeamModeFallbackError` and throw when preservation isn’t possible, instead of starting an unregistered session.
  - Carry `teamRunId` and `onSessionCreated` into the retry input so the fallback subagent remains a team participant.

<sup>Written for commit b516d5d41ef55c219a0dba1816cb0958a3acd70c. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4067?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/background-agent/fallback-retry-handler.test.ts
+++ b/src/features/background-agent/fallback-retry-handler.test.ts
@@ -1,5 +1,5 @@
 import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test"
-import { tryFallbackRetry, type FallbackRetryHandlerDeps } from "./fallback-retry-handler"
+import { tryFallbackRetry, TeamModeFallbackError, type FallbackRetryHandlerDeps } from "./fallback-retry-handler"
 import type { FallbackEntry } from "../../shared/model-requirements"
 
 const sharedLogMock = mock(() => {})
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot]

- Review decision: APPROVED

code-yeongyu (approved):
[sisyphus-bot] APPROVED. Trusted contributor (PeterPonyu) team-mode fallback preservation. Adds `TeamModeFallbackError` class and a structured guard in `tryFallbackRetry` that fails fast when a team-mode task (teamRunId set) lacks the `onSessionCreated` callback needed to re-register the fallback session in the team-session registry. Without this, fallback would silently create an orphaned session that fails every subsequent team tool call. Closes #3898. 2 new regression tests pinning both branches. Merging.

PR #4089, opened 2026-05-16
fix(skill-mcp): respect env allowlist for skill env

### LINKED ISSUE #3995: mcp_env_allowlist should bypass EXCLUDED_ENV_PATTERNS for skill MCP env field
## Problem

skill_mcp spawns MCP processes from skill config. The env: field passes env vars to spawned processes, but createCleanMcpEnvironment() filters them through EXCLUDED_ENV_PATTERNS — a hard blacklist (/_TOKEN$/i, /_KEY$/i, etc.) that cannot be overridden.

Example: TELEGRAM_BOT_TOKEN is silently removed.

## Expected behavior

Skill-configured env: vars should either bypass EXCLUDED_ENV_PATTERNS (they are explicitly configured), or respect mcp_env_allowlist.

## Workaround

Rename env var to avoid patterns (e.g., TG_API_AUTH instead of TELEGRAM_BOT_TOKEN).

## Code

- createCleanMcpEnvironment — hard filters via EXCLUDED_ENV_PATTERNS, no allowlist check
- mcp_env_allowlist — only used for  interpolation, never checked in env passthrough

Neo1228:
## Summary
- allow `mcp_env_allowlist` to unblock explicitly configured skill MCP env entries
- keep ambient process secrets filtered even when the same name is allowlisted
- add regression coverage for `TELEGRAM_BOT_TOKEN`-style skill env passthrough

Fixes #3995

## Testing
- bun test src/features/skill-mcp-manager/env-cleaner.test.ts
- bun run typecheck

---
## Summary by cubic
Respect `mcp_env_allowlist` in the skill MCP environment cleaner. Allowlisted vars set in skill config now survive filtering, while ambient process secrets with the same names stay blocked.

- **Bug Fixes**
  - Preserve allowlisted custom env values (e.g., `TELEGRAM_BOT_TOKEN`) during cleaning.
  - Do not pass through matching ambient process secrets.
  - Added regression tests for allowlist passthrough and ambient blocking.

<sup>Written for commit 83fdd4da76a65ca4dbfa2163b5f720325ded0e39. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4089?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/skill-mcp-manager/env-cleaner.test.ts
+++ b/src/features/skill-mcp-manager/env-cleaner.test.ts
@@ -1,6 +1,10 @@
 /// <reference types="bun-types" />
 
 import { describe, it, expect, afterEach } from "bun:test"
+import {
+  resetAdditionalAllowedMcpEnvVars,
+  setAdditionalAllowedMcpEnvVars,
+} from "../claude-code-mcp-loader/configure-allowed-env-vars"
 import { createCleanMcpEnvironment, EXCLUDED_ENV_PATTERNS } from "./env-cleaner"
 
 describe("createCleanMcpEnvironment", () => {
@@ -8,6 +12,8 @@ describe("createCleanMcpEnvironment", () => {
   const originalEnv = { ...process.env }
 
   afterEach(() => {
+    resetAdditionalAllowedMcpEnvVars()
+
     // Restore original environment
     for (const key of Object.keys(process.env)) {
       if (!(key in originalEnv)) {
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot] Note: PR #4176 (merged) takes a different approach to the same problem — it makes all skill-configured `customEnv` entries pass through the secret-name filter unconditionally. This PR's allowlist-based approach is more conservative and would compose with #4176, but the immediate user-facing bug (#3995, Telegram bot token) is now resolved by #4176.\n\nIf the explicit-allowlist semantic is still valuable (it tightens the trust boundary that security-skeptic flagged on #4176), please rebase onto current dev and clarify the intended interaction with #4176's pass-through. Otherwise this PR may be closed as superseded.

code-yeongyu:
[sisyphus-bot] Hi Neo1228. 🙏 Thanks for the careful work on this. I read both the allowlist patch here and the alternative that landed.

Per the note from the 20th: #4176 ("fix(skill-mcp-manager): trust explicit skill MCP env vars (#3995)") was merged on 2026-05-20T14:09:13Z and resolves the immediate Telegram bot token failure that #3995 reported. That uses a broader "trust all explicit customEnv entries" model rather than this PR's narrower allowlist, so the user-facing bug is already off the dashboard on `dev`.

I'm going to close this out as superseded so the two approaches don't keep diverging. If the explicit-allowlist semantic is still worth pursuing for tightening the trust boundary that #4176 widened, please open a fresh PR against current `dev` that explicitly layers on top of #4176 (i.e., "ambient process secrets stay blocked even when listed in `customEnv`, unless also in `mcp_env_allowlist`"); I'd review that as a separate hardening change rather than as a fix for #3995. Either way, the allowlist regression tests you added are useful prior art and would slot into the layered version cleanly.

Really appreciate the contribution and the conservative direction here.

PR #4098, opened 2026-05-17
fix(look-at): rework session poller to wait for assistant response, not just idle

sjawhar:
Rewrites the `look-at` session poller from a basic idle-wait into a proper assistant-response wait that also extracts the final message.

## Why

The old `pollSessionUntilIdle` resolved as soon as the multimodal-looker session went idle, then a separate code path fetched messages. This caused:
- Races where the message extraction ran before the assistant response landed.
- Timeout-vs-success ambiguity (busy → idle could mean either).
- No clear signal of which message was the actual response.

## What

- Rename `pollSessionUntilIdle` → `waitForLookAtSessionResult` (signals intent).
- New `assistant-message-extractor.ts` isolates message selection logic.
- Poller now returns `{ messages, outcome }` together — one round-trip.
- Adds `allowStableIdleWithoutActivity` option for the "session is already idle on entry" case.

## Test plan

- 11 test cases in `session-poller.test.ts` covering busy→idle, stable-idle, timeout, status API errors, retry states, and default options.
- `tools.test.ts` updated to use the new poller signature for the existing error-handling tests.

---
## Summary by cubic
Reworked the look-at session poller to wait for the assistant’s real outcome and return it with messages in one call. Adds abort handling and stronger fallbacks to fix race conditions and idle/timeout ambiguity.

- **Bug Fixes**
  - Waits for a terminal status (`idle`, `interrupted`, `error`) and returns the assistant outcome (`text` or `error`) with `messages`; uses stable-idle detection (3 unchanged polls) or when the last message is from the assistant.
  - Handles `session.status` errors or lack of support by falling back to messages; distinguishes “never in status map yet” from idle unless `allowStableIdleWithoutActivity`; handles `messages` errors during idle; supports abort via `abortSignal` by calling `client.session.abort({ path: { id } })`.

- **Refactors**
  - Renamed `pollSessionUntilIdle` to `waitForLookAtSessionResult`, which returns `{ messages, outcome, statusType }`.
  - Upgraded the extractor to `extractLatestAssistantOutcome`, exposing `text`, `errorName`, `hasAssistant`, and `completed`.

<sup>Written for commit 33c8bcd850a5d8895a14ed35c65cd22147885de2. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4098?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/tools/look-at/assistant-message-extractor.ts
+++ b/src/tools/look-at/assistant-message-extractor.ts
@@ -50,18 +50,47 @@ function getTextParts(message: SessionMessage): MessagePart[] {
 }
 
 export function extractLatestAssistantText(messages: unknown): string | null {
-  if (!Array.isArray(messages) || messages.length === 0) return null
+  return extractLatestAssistantOutcome(messages).text
+}
 
-  const assistantMessages = messages
+export interface AssistantOutcome {
+  text: string | null
+  errorName: string | null
+  hasAssistant: boolean
+  completed: boolean
+}
+
+export function extractLatestAssistantOutcome(messages: unknown): AssistantOutcome {
+  if (!Array.isArray(messages) || messages.length === 0) {
+    return { text: null, errorName: null, hasAssistant: false, completed: false }
+  }
+
+  const parsed = messages
     .map(asSessionMessage)
     .filter((message): message is SessionMessage => message !== null)
+
+  const assistantMessages = parsed
     .filter((message) => message.info?.role === "assistant")
     .sort((a, b) => getCreatedTime(b) - getCreatedTime(a))
 
+  const hasAssistant = assistantMessages.length > 0
   const lastAssistantMessage = assistantMessages[0]
-  if (!lastAssistantMessage) return null
+
+  if (!lastAssistantMessage) {
+    return { text: null, errorName: null, hasAssistant, completed: false }
+  }
 
   const textParts = getTextParts(lastAssistantMessage)
-  const responseText = textParts.map((part) => part.text).join("\n")
-  return responseText
+  const text = textParts.map((part) => part.text).join("\n") || null
+
+  const allParts = Array.isArray(lastAssistantMessage.parts) ? lastAssistantMessage.parts : []
+  const errorPart = allParts.find((part): part is Record<string, unknown> =>
+    isObject(part) && typeof part["type"] === "string" && part["type"] === "error"
+  )
+  const errorName = errorPart && typeof errorPart["error"] === "string" ? errorPart["error"] : null
+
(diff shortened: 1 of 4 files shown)


sjawhar:

Addressed both P2 findings:

1. **Dedupe**: `extractLatestAssistantText` now delegates to `extractLatestAssistantOutcome(messages).text` — single source of truth for parsing/sorting/filtering.
2. **Busy→idle test rigor**: `createMockClient` now gates message availability on idle status (`gateMessagesOnIdle: true` by default). Messages return `[]` while busy, real data only after status transitions to idle — so any early-fetch regression would be caught.

All 5 session-poller tests + 29 tools.test tests pass locally.

sjawhar:

Addressed the latest finding in `session-poller.ts`: `getSessionStatus` now checks `statusResult.error` explicitly before reading `data`. Non-throwing SDK error shapes now log + fall back to message-based polling instead of being silently treated as idle.

sjawhar:

Addressed the 'supported-status missing entry can stall' concern: `waitForLookAtSessionResult` now tracks whether the session was ever seen in the status map. If status supports the API but our session never appeared (e.g. just-created session not yet visible), the poller treats it as still-active so we don't short-circuit on the first poll. The `allowStableIdleWithoutActivity` opt-in still bypasses this for the pre-idle case.

code-yeongyu (approved):

Substantive look-at session-poller refactor by trusted contributor (sjawhar). Independent verification:

## What changed

- `pollSessionUntilIdle` → `waitForLookAtSessionResult`. The poller now waits for an actual ASSISTANT response (not just idle status), avoiding the race where the session is technically idle but the assistant hasn't emitted any output yet.
- `extractLatestAssistantText` now wraps a richer `extractLatestAssistantOutcome` returning `{ text, errorName, hasAssistant, completed }`. Backward compatible — old call sites still get `text`.
- Error parts in assistant messages are now detected and surfaced via `errorName` so the look-at tool can distinguish "model errored" from "model said nothing".

## Test changes

- Test renames match the API rename. 159 LOC of test churn but the count is dominated by switching from the old `pollSessionUntilIdle`-style assertions to the new outcome-based assertions.
- Mock client adds `messages` method gated on idle observation (`gateMessagesOnIdle`), simulating the real OpenCode behavior where messages arrive after idle.

## Risk

- Behavior change: callers that previously got `null` text on "idle but no assistant response yet" now wait longer for the actual response. Stricter contract.
- No test regression — all existing test names are preserved (just renamed describe blocks).
- Cubic + CI green.

PR #4099, opened 2026-05-17
fix(skill-mcp): allow MCP manager to accept connections after disconnectAll

sjawhar:
Fixes an issue where SkillMcpManager refused new connections after `disconnectAll()`, breaking plugin reloads.

## Why

When a plugin reload triggered `disconnectAll()`, the manager set its internal "shutdown" state and rejected all subsequent `getOrCreateClient` calls. After reload, no MCP servers could reconnect — they stayed down until the OpenCode process restarted.

## What

- Reset the cleanup/shutdown flag at the end of `disconnectAll()` so the manager can serve new connections.
- New regression tests:
  - `connection-race.test.ts` — verify concurrent connect during cleanup
  - `plugin-reload-mcp-survival.test.ts` — full reload simulation

## Test plan

- New test file `plugin-reload-mcp-survival.test.ts` reproduces the exact reload flow and asserts MCPs come back up.
- Existing manager tests still pass.

---
## Summary by cubic
Allows the Skill MCP manager to accept new connections after `disconnectAll()`, so MCP servers reconnect correctly after plugin reloads. Also isolates CI tests per file to avoid mock leaks.

- **Bug Fixes**
  - Reset `state.disposed` at the end of `disconnectAll()` to re-enable `getOrCreateClient()` after cleanup.
  - Added `plugin-reload-mcp-survival.test.ts` and updated `connection-race.test.ts` to ensure reconnection; CI now runs each test file in its own Bun process to prevent `mock.module()` leaks.

<sup>Written for commit 824cd1a8a44f76c85bf28385f8a95398214eb597. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4099?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/skill-mcp-manager/cleanup.ts
+++ b/src/features/skill-mcp-manager/cleanup.ts
@@ -127,6 +127,10 @@ export async function disconnectSession(state: SkillMcpManagerState, sessionID: 
 
 export async function disconnectAll(state: SkillMcpManagerState): Promise<void> {
   state.shutdownGeneration++
+  // Temporarily block new connections during cleanup. Reset at the end so that
+  // sessions surviving a plugin reload can reconnect. (Plugin reload calls
+  // disconnectAll via plugin-dispose, but existing sessions' tool closures still
+  // reference this manager instance and must be able to create new connections.)
   state.disposed = true
   stopCleanupTimer(state)
   unregisterProcessCleanup(state)
@@ -141,6 +145,8 @@ export async function disconnectAll(state: SkillMcpManagerState): Promise<void> 
   for (const managed of clients) {
     await closeManagedClient(managed)
   }
+
+  state.disposed = false
 }
 
 export async function forceReconnect(state: SkillMcpManagerState, clientKey: string): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


code-yeongyu (approved):

Real-world bug fix for plugin-reload survival. Independent verification:

## Bug being fixed

OMO plugin function is called per-session-directory. Each call disposes the previous plugin's `SkillMcpManager` via `disconnectAll()`. Previously, `disconnectAll()` set `state.disposed = true` PERMANENTLY, so any session that survived the reload (and still held closures over the now-disposed manager) hit `"has been shut down, cannot create new connections"` on the next `skill_mcp()` call.

## Fix

`state.disposed = true` is now set ONLY during the cleanup window (between start of `disconnectAll` and end). After existing clients are closed, `state.disposed = false` is restored, so subsequent connection requests through surviving session closures succeed.

## Test coverage

- 1 updated test (`connection-race.test.ts`): now asserts NEW behavior — post-`disconnectAll` `getOrCreateClient()` succeeds and registers a fresh client.
- 1 new test file (`plugin-reload-mcp-survival.test.ts`, 155 LOC): pins the plugin-reload scenario end-to-end. Comments explain the exact production code path being exercised.

## Risk

- During the cleanup window (a few ms while existing clients are closed), concurrent connection attempts WILL transiently fail with the "shut down" error. This is the intended behavior — they should retry. The window is bounded by closing the existing clients, which is fast.
- No behavior change for the no-reload single-session case.
- Cubic + CI green.

PR #4102, opened 2026-05-17
fix(skill): pass directory to getAllSkills + fix async test timing

sjawhar:
Re-attempt of #3098 (closed without comment) — this is needed for our fork to operate, so refiling rebased onto current dev with the test-timing fix included.

## Why

\`getAllSkills\` was called without \`directory\`, which caused project-scoped skills to be silently missed in some session setups. The fix threads the \`directory\` param through.

The previous PR (#3098) was closed without explanation. Happy to address whatever the original concern was if it's flagged.

## What

- \`tool-registry.ts\`: pass \`directory\` to \`getAllSkills\` call site.
- \`tools.ts\`: thread the param.
- \`types.ts\`: extend the interface.
- \`async-description-refresh.test.ts\`: fix flaky timing assertion that was masking the issue.

## Test plan

- Updated test passes consistently (was flaky on slower CI runners).
- Existing skill-tool tests still pass.

---
## Summary by cubic
Scopes skill discovery and caching by project directory so project-scoped skills are found and caches don’t leak across projects. Also stabilizes the async description refresh test.

- **Bug Fixes**
  - Thread `directory` from plugin context to `createSkillTool` and into `getAllSkills`; include it in the discovery cache key.
  - Make `directory` required in `SkillLoadOptions`; exported `skill` uses `process.cwd()`.
  - Deflake `async-description-refresh` by using a temp dir and a 50ms poll interval.

- **Migration**
  - If you call `createSkillTool` directly, pass `{ directory: <projectDir> }`.

<sup>Written for commit d2d154137743356113b3692f8dbfba706cf6668c. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4102?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/opencode-skill-loader/skill-discovery.ts
+++ b/src/features/opencode-skill-loader/skill-discovery.ts
@@ -12,7 +12,8 @@ export function clearSkillCache(): void {
 export async function getAllSkills(options?: SkillResolutionOptions): Promise<LoadedSkill[]> {
 	const browserProvider = options?.browserProvider ?? "playwright"
 	const teamModeEnabled = options?.teamModeEnabled ?? false
-	const cacheKey = `${browserProvider}:${teamModeEnabled ? "team-on" : "team-off"}`
+	const directory = options?.directory ?? ""
+	const cacheKey = `${directory}:${browserProvider}:${teamModeEnabled ? "team-on" : "team-off"}`
 	const hasDisabledSkills = options?.disabledSkills && options.disabledSkills.size > 0
 
 	// Skip cache if disabledSkills is provided (varies between calls)
--- a/src/plugin/tool-registry.ts
+++ b/src/plugin/tool-registry.ts
@@ -273,6 +273,7 @@ export function createToolRegistry(args: {
     enabledPluginsOverride: pluginConfig.claude_code?.plugins_override,
   })
   const skillTool = factories.createSkillTool({
+    directory: ctx.directory,
     commands,
     skills: skillContext.mergedSkills,
     mcpManager: managers.skillMcpManager,
--- a/src/tools/skill/async-description-refresh.test.ts
+++ b/src/tools/skill/async-description-refresh.test.ts
@@ -1,6 +1,10 @@
 /// <reference types="bun-types" />
 
-import { describe, expect, it } from "bun:test"
+import { describe, expect, it, beforeEach, afterEach } from "bun:test"
+import { mkdtempSync, rmSync } from "node:fs"
+import { join } from "node:path"
+import { tmpdir } from "node:os"
+import { createSkillTool } from "./tools"
 import type { LoadedSkill } from "../../features/opencode-skill-loader/types"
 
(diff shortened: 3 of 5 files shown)


sjawhar:

The flagged "import conflict" at `async-description-refresh.test.ts:7` looks like a false positive. The file uses only `typeof import("./tools").createSkillTool` (a type-level reference) and a local helper function of the same name — there's no runtime import of `createSkillTool`. The test runs and passes locally (1 pass, 0 fail).

```bash
$ bun test src/tools/skill/async-description-refresh.test.ts
 1 pass
 0 fail
```

sjawhar:

Addressed the cache-key concern in `skill-discovery.ts`: `getAllSkills` cache key now includes `directory` so queries from different working directories within the same session no longer collide.

```diff
-const cacheKey = `${browserProvider}:${teamModeEnabled ? "team-on" : "team-off"}`
+const directory = options?.directory ?? ""
+const cacheKey = `${directory}:${browserProvider}:${teamModeEnabled ? "team-on" : "team-off"}`
```

sjawhar:
Renamed the local helper to `createFreshSkillTool` to avoid the perceived naming collision with the dynamic import. The local function never collided at runtime (only `typeof import(...)` type references were used, never a runtime import), but the rename should clear the warning. @cubic-dev-ai please re-review.

code-yeongyu (approved):

Two coherent fixes bundled correctly:

## 1. Cache key now includes directory

`getAllSkills` previously cached by `${browserProvider}:${teamModeEnabled}` only. In multi-directory sessions (workspace switches, team-mode members with their own worktrees), the same cache hit would return skills discovered from a DIFFERENT directory. Fix prepends `${directory}` to the cache key.

## 2. Test isolation + timing fix

- `beforeEach` / `afterEach` now create + tear down a per-test temp dir via `mkdtempSync` / `rmSync`. Removes test bleed from filesystem state.
- `waitForRefresh` polling interval bumped from 10ms → 50ms — reduces test flakiness on slow CI. Not a time-is-SUT case but the wait is bounded by a predicate, so it's fail-fast on the actual signal rather than a fixed sleep. Acceptable.
- `createSkillTool(options: SkillLoadOptions = {})` → `createSkillTool(options: SkillLoadOptions)` — removes the `= {}` default so `directory` is now structurally required by callers. Internal-only change (only two call sites: `tool-registry.ts` and the test helper). Both updated in the same diff.

## Non-blocking observations

- 50ms polling interval is borderline against the `test-discipline` rule (`forbids setTimeout(resolve, N) / await sleep(N) in tests unless time IS the SUT`). Strictly speaking this is *polling*, not `sleep` — the loop exits early on predicate success. Acceptable here.
- The renamed `createSkillTool` → `createFreshSkillTool` helper is clearer.

## Risk

- Cache miss rate increases by a factor of the number of distinct directories per session. Bounded and small — directory count is typically 1-2 per session.
- No security or behavior change for single-directory use.

PR #4113, opened 2026-05-17
fix(runtime-fallback): classify more provider quota error names

### LINKED ISSUE #3937: bug: fallback does not trigger when default model hits usage limit
Summary\nDiscord feedback reports that when the default model hits a usage/quota limit, OMO/OpenCode keeps retrying the same model indefinitely instead of triggering configured model fallback.\n\nSource\n- Discord #omo message 1503238435568222359\n- User report: fallback is not triggered when the default model hits the usage limit; UI appears stuck in infinite attempts.\n\nExpected behavior\n- Usage-limit / quota / rate-limit provider errors that are configured as fallback-worthy should switch to the configured fallback model/provider.\n- If fallback is not possible, OMO should stop with a clear actionable error instead of looping forever.\n\nActual behavior\n- The default model appears to retry indefinitely after usage limit is reached.\n- Fallback does not activate automatically.\n\nAcceptance criteria\n- Identify the provider error/status strings currently treated as fallback-worthy.\n- Add usage-limit/quota/rate-limit error classification if missing.\n- Ensure retry loops have a bounded stop condition when fallback cannot happen.\n- Add a regression test or fixture for a usage-limit response that triggers fallback.\n\nRequested repro details\n- OMO/OpenCode versions\n- provider/model id for default and fallback\n- exact error text/log, not only screenshot\n- fallback config snippet with secrets redacted\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PeterPonyu:
Closes #3937.

## Summary

Adds three small classification gaps to `classifyErrorType` so that quota-exhaustion errors from a wider range of providers trigger configured fallback chains instead of looping retry attempts.

## Root cause

`classifyErrorType` matches provider error names via lowercased alphanumeric substrings (e.g. `errorName.includes("insufficientquota")`). That check is broken for three real-world surfaces:

| Provider | Error name | Why current code missed it |
| --- | --- | --- |
| Google Generative AI | `RESOURCE_EXHAUSTED` (gRPC code 8) | name not in allow-list; message pattern absent |
| Google Generative AI | `ResourceExhausted` | name not in allow-list |
| OpenAI | `insufficient_quota` (snake_case) | `.includes("insufficientquota")` returns false because the underscore breaks the substring match |
| Anthropic / others | `rate_limit_exceeded` | same snake_case underscore issue |

When a quota error is not classified as `"quota_exceeded"`, the runtime-fallback path skips dispatching the configured fallback model and just keeps retrying the same dead model.

## Changes

- Normalize error names by stripping `_` and `-` once at the top of `classifyErrorType`, so snake_case / SCREAMING_SNAKE_CASE provider names match the existing alphanumeric `.includes()` checks. The single underscore-bearing literal (`ai_loadapikeyerror`) is updated to `ailoadapikeyerror` so previously matched names still resolve.
- Add `resourceexhausted` to the quota error-name allow-list (Google).
- Add `/resource.?exhausted/i` to the quota message-pattern list (also Google, when the provider sets a generic error name but puts the signal in the message).

## How I verified

- Three new regression tests in `src/hooks/runtime-fallback/quota-error-classifier.regression.test.ts` cover:
  - Google `RESOURCE_EXHAUSTED` (gRPC error name + quota-shaped message)
  - Google `ResourceExhausted` message form without HTTP status
  - OpenAI snake_case `insufficient_quota` error name
- All targeted suites pass:
  - `bun test src/hooks/runtime-fallback/` → 172 pass / 0 fail / 308 expect() calls / 16 files
- `bunx tsc --noEmit` clean on the touched files (filtered against the same zod/v4/locales noise everyone else hits).

## Risk

Minimal. The normalization only adds matches that were previously missed — no existing match is removed (verified by leaving the prior tests untouched). The one underscore-bearing literal was rewritten to its underscore-stripped form for behavior parity.

## Checklist

- [x] My code follows the existing style of the codebase
- [x] I have added/updated tests for my changes
- [x] All tests pass locally
- [x] I have not introduced any new TypeScript errors

---
## Summary by cubic
Fixes quota error classification so runtime fallbacks trigger instead of looping retries. Addresses #3937 by recognizing Google and snake_case provider error names like RESOURCE_EXHAUSTED and insufficient_quota.

- Bug Fixes
  - Normalize error names by stripping "_" and "-" in classifyErrorType; updated literal to `ailoadapikeyerror` and added explicit optional chaining on `.replace()`.
  - Add `resourceexhausted` to the quota allow-list and a `/resource.?exhausted/i` message check for Google Generative AI.
  - Add three regression tests (Google RESOURCE_EXHAUSTED, ResourceExhausted message form, OpenAI insufficient_quota) and tighten fixtures with generic messages so only the new match paths make them pass.

<sup>Written for commit a8ccffdd7c7a91300c5bd25b66607b4e17eaf474. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4113?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/runtime-fallback/error-classifier.ts
+++ b/src/hooks/runtime-fallback/error-classifier.ts
@@ -109,7 +109,7 @@ export function classifyErrorType(error: unknown): string | undefined {
   // Normalize by stripping underscores and dashes so snake_case / kebab-case
   // provider error names (e.g. "insufficient_quota", "RESOURCE_EXHAUSTED")
   // match the existing alphanumeric .includes() checks below.
-  const errorName = extractErrorName(error)?.toLowerCase().replace(/[_-]/g, "")
+  const errorName = extractErrorName(error)?.toLowerCase()?.replace(/[_-]/g, "")
 
   if (
     errorName?.includes("ailoadapikeyerror") ||
--- a/src/hooks/runtime-fallback/error-classifier.ts
+++ b/src/hooks/runtime-fallback/error-classifier.ts
@@ -106,10 +106,13 @@ function isLocalizedQuotaExhaustionMessage(message: string): boolean {
 
 export function classifyErrorType(error: unknown): string | undefined {
   const message = getErrorMessage(error)
-  const errorName = extractErrorName(error)?.toLowerCase()
+  // Normalize by stripping underscores and dashes so snake_case / kebab-case
+  // provider error names (e.g. "insufficient_quota", "RESOURCE_EXHAUSTED")
+  // match the existing alphanumeric .includes() checks below.
+  const errorName = extractErrorName(error)?.toLowerCase().replace(/[_-]/g, "")
 
   if (
-    errorName?.includes("ai_loadapikeyerror") ||
+    errorName?.includes("ailoadapikeyerror") ||
     errorName?.includes("loadapi") ||
     (/api.?key.?is.?missing/i.test(message) && /environment variable/i.test(message))
   ) {
@@ -132,6 +135,7 @@ export function classifyErrorType(error: unknown): string | undefined {
     errorName?.includes("quotaexceeded") ||
     errorName?.includes("insufficientquota") ||
     errorName?.includes("billingerror") ||
+    errorName?.includes("resourceexhausted") ||
     /quota.?exceeded/i.test(message) ||
     /exceeded.*quota/i.test(message) ||
     /usage\s*quota/i.test(message) ||
@@ -139,6 +143,7 @@ export function classifyErrorType(error: unknown): string | undefined {
     /insufficient.?(?:quota|balance|funds?)/i.test(message) ||
     /billing.?(?:hard.?)?limit/i.test(message) ||
     /exhausted\s+your\s+capacity/i.test(message) ||
+    /resource.?exhausted/i.test(message) ||
     /out\s+of\s+credits?/i.test(message) ||
     /payment.?required/i.test(message) ||
     /usage\s+limit/i.test(message) ||
--- a/src/hooks/runtime-fallback/quota-error-classifier.regression.test.ts
+++ b/src/hooks/runtime-fallback/quota-error-classifier.regression.test.ts
@@ -89,4 +89,52 @@ describe("runtime-fallback quota error regressions", () => {
     expect(errorType).toBe("quota_exceeded")
     expect(retryable).toBe(true)
   })
+
+  test("classifies Google RESOURCE_EXHAUSTED (gRPC code 8) as quota_exceeded", () => {
+    //#given
+    const error = {
+      name: "RESOURCE_EXHAUSTED",
+      message: "Quota exceeded for quota metric 'Generate Content' and limit 'Generate Content quota per minute'.",
+    }
+
+    //#when
+    const errorType = classifyErrorType(error)
+    const retryable = isRetryableError(error, [429, 500, 502, 503, 504])
+
+    //#then
+    expect(errorType).toBe("quota_exceeded")
+    expect(retryable).toBe(true)
+  })
+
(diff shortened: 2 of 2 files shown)


PeterPonyu:
Thanks @cubic-dev-ai — confirmed and fixed in b2f0d423.

The bot was right that the `RESOURCE_EXHAUSTED` fixture passed via the pre-existing `/quota.?exceeded/i` message regex, not the new error-name allow-list entry. I also caught a second weak fixture in the same review pass: the snake_case `insufficient_quota` test had a message that matched `/exceeded.*quota/i`, so it would have passed even if the underscore normalization regressed.

Both fixtures now use a generic `"Request failed."` message so the **only** path to `quota_exceeded` is via the new code being tested:

- `RESOURCE_EXHAUSTED` test: only matches via `errorName?.includes("resourceexhausted")` after the new normalization
- `insufficient_quota` test: only matches via the new underscore-stripping that turns `insufficient_quota` into `insufficientquota`

The third new test (Google `ResourceExhausted` message-only form) is left alone because its message `"Resource exhausted: Please try again later."` uniquely matches only the newly-added `/resource.?exhausted/i` pattern and no pre-existing quota regex.

`bun test src/hooks/runtime-fallback/` still green (172 pass / 0 fail). Re-trigger when convenient.

@cubic-dev-ai

PeterPonyu:
@cubic-dev-ai — applied the suggested chain in a8ccffdd, but a quick correction on the underlying claim: the original code was already crash-safe. JavaScript optional chaining short-circuits the **entire** property/call chain when the head returns null/undefined, not just the very next access. Empirical proof:

\`\`\`bash
$ bun -e 'const x = undefined; console.log(x?.toLowerCase().replace(/_/g, ""))'
undefined
\`\`\`

So \`extractErrorName(error)?.toLowerCase().replace(...)\` evaluates to \`undefined\` when \`extractErrorName\` returns undefined — \`.replace\` is never invoked on undefined. (See [MDN — Optional chaining § Short-circuiting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#short-circuiting).)

That said, applying the suggested \`?.\` before \`.replace\` is a harmless semantic no-op and arguably easier for static analyzers to reason about hop-by-hop, so I've taken your suggestion as a style improvement. \`bun test src/hooks/runtime-fallback/\` still green (172 pass / 0 fail).

code-yeongyu (approved):

Trusted contributor (PeterPonyu) error-classification expansion. Independent verification:

## Three improvements bundled coherently

1. **Underscore/dash normalization**: `errorName.replace(/[_-]/g, "")` lets snake_case (`insufficient_quota`) and SCREAMING_SNAKE_CASE (`RESOURCE_EXHAUSTED`) provider error names match the alphanumeric `.includes()` patterns. Existing `ai_loadapikeyerror` check correctly migrated to `ailoadapikeyerror`.
2. **Google gRPC code 8 (`RESOURCE_EXHAUSTED`)**: added to both name allow-list and message regex. Closes the gap where Google models hitting quota would not trigger runtime-fallback.
3. **Comment hygiene**: explanatory comments above the normalization step and on each new test case. Comments are technical, no AI-slop.

## Test coverage

3 new regression tests, each pinning a different signal:
- Bare name `RESOURCE_EXHAUSTED` + generic message (proves name-only classification works)
- Provider-wrapped error with `Resource exhausted:` message (proves message-only classification works)
- snake_case `insufficient_quota` name + generic message (proves underscore normalization works)

Comments inside each test explicitly call out what would regress if removed — well-engineered.

## Risk

- Normalization only strips `_` and `-`. No false-positive expansion since the existing `.includes()` checks already use alphanumeric tokens.
- Adding `resourceexhausted` and `resource.?exhausted` patterns is purely additive; no existing classification flips.
- All CI green; Cubic 5/5 auto-approved.

PR #4114, opened 2026-05-17
fix(non-interactive-env): use powershell syntax on Windows regardless of SHELL/MSYSTEM (#3607)

### LINKED ISSUE #3607: detectShellType() returns unix on Windows when MSYSTEM is set, breaking non-interactive-env hook
### detectShellType() returns "unix" on Windows when MSYSTEM is set, generating invalid shell syntax

**Describe the bug**

On Windows, when `MSYSTEM` environment variable is set (default when Git Bash is installed), `detectShellType()` returns `"unix"` before checking `PSModulePath`. This causes the `non-interactive-env` hook to generate `export KEY=val;` syntax (Bourne shell), which is invalid in PowerShell — the shell that actually executes the command.

Error:
```
export : 无法将"export"项识别为 cmdlet、函数、脚本文件或可运行程序的名称。
```

**Root cause**

In `src/shared/shell-env.ts`, MSYSTEM is checked before PSModulePath:

```
SHELL exists → unix
MSYSTEM exists → unix   ← triggers before PSModulePath
PSModulePath exists → powershell
win32 fallback → cmd
```

MSYSTEM is a permanent env var set by Git Bash installation. It doesn't mean OpenCode's bash tool is running in a Unix shell — on Windows, the bash tool always spawns `powershell -Command ...` or `cmd.exe /c ...`.

**To reproduce**

1. Install Git Bash on Windows (sets MSYSTEM=MINGW64)
2. Run any git command in OpenCode with OMO loaded
3. non-interactive-env hook injects `export CI=true ...` prefix
4. PowerShell errors on `export`

**Suggested fix**

In `non-interactive-env-hook.ts`, the shell type for the env prefix should match the actual execution environment, not the parent process:

```typescript
const shellType = process.platform === "win32" ? "powershell" : detectShellType()
```

Or adjust `detectShellType()` to check `process.platform` before `MSYSTEM`.

**Environment**

- OS: Windows
- Git Bash: installed (sets MSYSTEM)
- OMO: latest

PeterPonyu:
Closes #3607.

## Summary

On Windows, `detectCommandShellType()` returned `"unix"` for two common environments — Git Bash sets `SHELL=/usr/bin/bash` (and `MSYSTEM=MINGW64`) permanently, but OpenCode actually spawns the bash tool through PowerShell. The hook then prepended `export KEY=val;` to git commands, which PowerShell rejects:

> `export : 无法将"export"项识别为 cmdlet...`

## Fix

`detectCommandShellType()` now short-circuits on `process.platform === "win32"` and resolves to a Windows-compatible shell:

- Recognized Windows shell via `SHELL` (`cmd.exe`, `powershell.exe`, `pwsh.exe`) — use it.
- `SHELL` and `MSYSTEM` both unset — fall back to `ComSpec` then to cmd.
- Anything else on win32 (Unix-shaped SHELL, MSYSTEM-only) — use PowerShell, matching what OpenCode actually spawns.

`detectShellType()` itself is unchanged; non-Windows callers are unaffected.

## Three test updates

Three existing tests previously asserted the buggy behavior:

- `Windows Git Bash environment with SHELL → unix syntax`
- `Windows Git Bash via MSYSTEM without SHELL → unix syntax`
- `SHELL=/bin/bash on win32 → unix syntax`

These are flipped to assert PowerShell syntax with a \`(#3607)\` marker and a comment noting that WSL is not affected (in WSL \`process.platform === 'linux'\`, not \`'win32'\`).

## Note on the issue thread

The sisyphus-bot triage on #3607 framed this as a policy choice between (A) forcing Windows env-prefix syntax and (B) resolving against the OpenCode-configured shell. This PR implements **option (A)** as the minimal surgical fix; option (B) remains a follow-up if the maintainer prefers that direction.

## Verification

- \`bun test src/hooks/non-interactive-env/\` → 24 pass / 0 fail
- \`bunx tsc --noEmit\` → clean (filtered against the same zod/v4/locales noise)
- Diff: +41/-19 across 2 files

## Checklist

- [x] My code follows the existing style of the codebase
- [x] I have updated tests for the new behavior and added a regression marker
- [x] All tests pass locally
- [x] No new TypeScript errors

---
## Summary by cubic
Fixes shell detection on Windows so env prefixes use PowerShell (or cmd) syntax instead of Unix, preventing errors when Git Bash sets Unix-like SHELL/MSYSTEM. Closes #3607.

- **Bug Fixes**
  - On win32, detectCommandShellType() now resolves to a Windows shell: use recognized SHELL (cmd/powershell/pwsh); if no SHELL/MSYSTEM, fall back to ComSpec then cmd; otherwise default to PowerShell.
  - Updated 3 tests to expect PowerShell syntax on Windows; non-Windows (including WSL) unchanged.

<sup>Written for commit 2bd7946001cef667fa89b2d1598e4a000cc74108. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4114?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/non-interactive-env/index.test.ts
+++ b/src/hooks/non-interactive-env/index.test.ts
@@ -370,7 +370,11 @@ describe("non-interactive-env hook", () => {
       expect(cmd).not.toContain("export ")
     })
 
-    test("#given Windows Git Bash environment with SHELL #when git command executes #then uses unix syntax", async () => {
+    test("#given Windows Git Bash SHELL=/usr/bin/bash #when git command executes #then uses powershell syntax (#3607)", async () => {
+      // Regression for #3607: OpenCode on Windows runs the bash tool through
+      // PowerShell by default, regardless of a Unix-shaped SHELL set by Git
+      // Bash. The export prefix is invalid PowerShell, so we must use
+      // PowerShell syntax even when SHELL points at /usr/bin/bash.
       delete process.env.PSModulePath
       process.env.SHELL = "/usr/bin/bash"
       Object.defineProperty(process, "platform", { value: "win32" })
@@ -386,12 +390,15 @@ describe("non-interactive-env hook", () => {
       )
 
       const cmd = output.args.command as string
-      expect(cmd).toStartWith("export ")
+      expect(cmd).toStartWith("$env:")
       expect(cmd).toContain("; git status")
-      expect(cmd).not.toContain("$env:")
+      expect(cmd).not.toContain("export ")
     })
 
-    test("#given Windows Git Bash via MSYSTEM without SHELL #when git command executes #then uses unix syntax", async () => {
+    test("#given Windows MSYSTEM=MINGW64 without SHELL #when git command executes #then uses powershell syntax (#3607)", async () => {
+      // Regression for #3607: MSYSTEM is permanently set on systems with Git
+      // Bash installed, but OpenCode on Windows still spawns PowerShell.
+      // MSYSTEM alone must not select Unix env-prefix syntax.
       delete process.env.SHELL
       process.env.MSYSTEM = "MINGW64"
       process.env.PSModulePath = "C:\\Program Files\\PowerShell\\Modules"
@@ -408,9 +415,9 @@ describe("non-interactive-env hook", () => {
       )
 
       const cmd = output.args.command as string
-      expect(cmd).toStartWith("export ")
+      expect(cmd).toStartWith("$env:")
       expect(cmd).toContain("; git status")
-      expect(cmd).not.toContain("$env:")
+      expect(cmd).not.toContain("export ")
     })
 
     test("#given Windows platform #when chained git commands via bash tool #then uses cmd syntax", async () => {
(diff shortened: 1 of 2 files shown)


code-yeongyu (approved):

Trusted contributor (PeterPonyu) Windows correctness fix. Independent verification:

## What the fix does

Previously `detectCommandShellType()` would honor a Unix-shaped `SHELL` env var (e.g. `/usr/bin/bash` set by Git Bash) and emit Unix `export FOO=bar; ...` syntax. But OpenCode on Windows runs the bash tool through PowerShell regardless — so emitting `export` produced invalid PowerShell that failed silently or noisily. Fix: on `process.platform === "win32"`, always map to a Windows shell type, defaulting to PowerShell.

## Correctness

- WSL is unaffected: in WSL, `process.platform === "linux"`, not `"win32"`. New code's first branch (`if (process.platform !== "win32")`) preserves Unix behavior. The tests explicitly document this in comments.
- Existing 4 test cases were updated to assert PowerShell syntax where they previously asserted `export ` syntax — this is the correct expectation under the new contract. Old assertions were locking in the bug.
- New test names include `(#3607)` annotation tying them to the linked issue.
- 1 unchanged test: `#given Windows platform #when chained git commands via bash tool #then uses cmd syntax` — still passes because `ComSpec` detection is preserved as a fallback.

## Risk assessment

- This is a behavior change but the previous behavior was broken on Windows + Git Bash. No legitimate Windows user can have been depending on Unix `export` syntax in their PowerShell environment.
- No Linux/macOS path touched.
- Cubic 5/5 + all CI green.

Issue #4061, opened 2026-05-15
bug: OpenCode Desktop sidecar crashes after first OMO prompt while CLI works


Yeachan-Heo:
Summary
Discord feedback reports that OMO loads in OpenCode Desktop/Electron, but after the first prompt the Desktop sidecar crashes. The same OMO setup works in the console/CLI.

Source
- Discord #omo message 1504940414040281200
- User: sonicflare

Reported symptoms
```text
OpenCode Desktop / Electron: plugin is loaded
First prompt: sidecar crashes
Console/CLI: works fine
```

Why this matters
This points at a Desktop sidecar/runtime bridge issue rather than a general OMO install/config failure, because the CLI path works with the same plugin. It may overlap with prior Electron/Desktop runtime compatibility work, but this specific symptom is sidecar crash-after-first-prompt.

Requested repro details
```text
- OpenCode Desktop version
- OMO version
- OS
- plugin install method/config snippet
- Desktop sidecar crash log / Developer Tools console around first prompt
- whether disabling specific hooks/agents changes the crash
- whether a minimal OMO config still crashes
```

Initial triage
Treat as Desktop/Electron sidecar compatibility until logs prove otherwise. CLI success is the key differentiator.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

code-yeongyu:
[sisyphus-bot]

I could not confirm this as an OMO bug from code inspection alone yet.

What the code does show is that Desktop/Electron is a distinct compatibility surface: the repo has `opencode-desktop` discovery, Desktop config-dir handling, Electron stdout semver parsing, and non-Bun runtime guards around Bun-only APIs. The first-prompt/continuation paths also route through shared prompt gates and process-level cleanup can force-exit the host on unhandled exceptions or rejections.

Relevant code surfaces:

- `src/hooks/shared/prompt-async-gate.ts:147-197` gates first-response prompt dispatch, including reservation, active-session checks, timeout, and post-dispatch hold.
- `src/plugin/event.ts:503-533` routes model-fallback continuations through `promptAsyncAfterSessionIdle` / `promptAfterSessionIdle`.
- `src/create-managers.ts:87-144` wires manager shutdown cleanup and background/session side effects.
- `src/features/background-agent/process-cleanup.ts:70-150` registers `uncaughtException` / `unhandledRejection` handlers that can call `process.exit()`.

So the report is plausible, but not confirmed without crash evidence. Please share:

[...]

Issue #3997, opened 2026-05-13
Sidecar crashes (exit code 1) after first request since v4.1.1 update


ThomasChangX:
## Bug Summary

After upgrading to `oh-my-openagent@4.1.1`, the sidecar process crashes with **exit code 1** immediately after processing the **first successful request**. All subsequent requests fail with `"failed to fetch"` because the sidecar is no longer running.

**Pattern**: First request → succeeds ✅ → sidecar crashes 💥 → all following requests → `"failed to fetch"` ❌

## Environment

- **OpenCode Desktop**: v1.14.48 (macOS arm64)
- **oh-my-openagent**: v4.1.1 (auto-updated from npm via `"plugin": ["oh-my-openagent@latest"]`)
- **@opencode-ai/plugin**: v1.14.48
- **@opencode-ai/sdk**: v1.14.48
- **Platform**: macOS Darwin (Apple Silicon M-series)
- **Node.js sidecar**: Uses `oh-my-openagent-darwin-arm64` binary

## Regression Evidence

The package was auto-updated at **2026-05-13 22:42:29 UTC+8**. Sidecar crashes began **immediately after**:

| Time | Event | Exit Code |
|------|-------|-----------|
| 22:42:21 | Sidecar normal exit (before update) | `0` ✅ |
| 22:42:29 | `package-lock.json` updated → v4.1.1 installed | — |
| 22:42:24 | New sidecar spawned | — |
| 22:44:17 | **First crash after update** | `1` ❌ |
| 22:45:32 | Crash | `1` ❌ |
| 22:46:26 | Crash (only ~20s uptime) | `1` ❌ |
| ... | **All subsequent exits are code 1** | `1` ❌ |

Before this update, all sidecar exits were `code: 0` (normal).

## Sidecar Log Pattern

From `~/Library/Logs/@opencode-ai/desktop/main.log`:

```
[2026-05-14 00:13:41.285] [info]  sidecar connection started { url: 'http://127.0.0.1:53922' }
[2026-05-14 00:13:41.286] [info]  spawning sidecar { url: 'http://127.0.0.1:53922' }
[2026-05-14 00:13:41.606] [warn]  sidecar stderr { message: '(node:72159) ExperimentalWarning: SQLite is an experimental feature...' }
[2026-05-14 00:13:42.326] [info]  server ready { url: 'http://127.0.0.1:53922' }
[...]

PR #4121, opened 2026-05-17
fix(delegate-task): default run_in_background and load_skills instead of throwing (fixes #4119)

mguttmann:
Fixes #4119 (and #3883, of which #4119 is a v4.1.2 confirmation).

## Problem

`tool-argument-preparation.ts` throws a hard `Error` when `task()` is invoked without an explicit `run_in_background` or `load_skills` argument. Because `tool.execute()` is short-circuited, the `tool.execute.after` hook never runs and `delegate-task-retry` can't attach corrective guidance. The model sees a raw failure and either burns several retries or falls back to a synchronous `Explore` call, silently losing parallel execution.

Confirmed live on macOS arm64 / OpenCode Desktop 1.15.3 / OmO v4.1.2: 6× `[task] run_in_background omitted; defaulting to false (sync delegation)` in real sessions after the patch is in place — without it, those are 6 hard throws.

## Behavior change

- `run_in_background` omitted → defaults to `false` (sync delegation), with a log line for observability.
- `load_skills` omitted or `null` → normalized to `[]`, with a log line on the explicit-null path.
- Zod schema entries are now `.optional()` and the `.describe()` strings declare the defaults honestly; the markdown tool description is updated to match (no more `REQUIRED` lie).
- The orthogonal validation `'Must provide either category or subagent_type.'` is unchanged and still surfaces as a returned error (so the retry hook can recover from it).

## Tests

- The five throw-on-missing tests in `tools.test.ts` are rewritten to assert the new default-and-proceed contract.
- The `no category, no subagent_type` test now asserts the missing-target error remains intact.
- 131/131 tests green.

## Notes

This matches the design from PR #2375 which was previously reverted by `566031f4`. The reporter's workaround in #3883 is now upstreamed.

---
## Summary by cubic
Default missing delegate-task args instead of throwing: run_in_background now defaults to false (sync) and load_skills to []; explicit null still throws. `task()` no longer hard-fails so `delegate-task-retry` can guide recovery (fixes #4119).

- **Bug Fixes**
  - run_in_background: default to false when omitted; add log for observability.
  - load_skills: undefined → []; null → throw with clear error; log when omitted.
  - Argument schema: mark both fields optional; descriptions state real defaults.
  - Tool description/examples updated to match; removed “REQUIRED” wording.
  - Preserve “Must provide either category or subagent_type.” as a returned error.
  - Tests: assert default-and-proceed; add sync continuation regression for `task_id` when flag omitted; keep explicit-null test throwing. All green.

[...]

Issue #4119, opened 2026-05-17
[Bug]: task() still requires run_in_background/load_skills in v4.1.2 — workaround for #3883 confirmed working on macOS


mguttmann:
## Summary

Confirming that the regression originally filed in #3883 is **still present in v4.1.2** on macOS arm64. Same failure pattern: every `task()` invocation throws on missing `run_in_background` / `load_skills` parameters, triggering retry loops or causing Sisyphus to fall back to synchronous `Explore` calls (losing parallel execution entirely).

This issue is filed as a separate datapoint from #3883 because:
1. #3883 reports the v4.0.0 regression; this confirms it persists in v4.1.2 (>1 week later, no release fix).
2. We tested the proposed workaround from #3883's "Workaround" section and confirm it works.
3. Datapoint: macOS 26.4 / arm64 / OpenCode Desktop 1.15.3 (no prior macOS confirmation in #3883).

## Environment

- **OS:** macOS 26.4 (build 25E246), arm64 (Apple Silicon)
- **OpenCode Desktop:** 1.15.3
- **oh-my-openagent:** 4.1.2 (auto-pulled via `"plugin": ["oh-my-openagent@latest"]`)
- **Node (sidecar):** v25.6.1 (Electron-bundled)
- **bunx:** 1.3.14

## Reproduction

1. Fresh OpenCode Desktop session with OmO loaded (`Sisyphus - Ultraworker` agent active, Opus 4.7).
2. Prompt: "Analyze this project across 4 angles in parallel: code quality, test coverage, security, DX."
3. Sisyphus emits 4 `task()` calls with `subagent_type=explore`.
4. Every task fails with:
   ```
   Invalid arguments: 'run_in_background' parameter is REQUIRED.
   Specify run_in_background=false for task delegation,
   or run_in_background=true for parallel exploration.
   ```
5. Sisyphus falls back to direct `Explore` calls (synchronous), losing the parallel benefit.

## Confirmed Workaround (Patch)

The fix proposed in #3883 works. Two-line patch in `dist/index.js`:

```js
// Before (lines 107605–107608 in v4.1.2)
const runInBackground = args.run_in_background;
if (runInBackground === undefined) {
  throw new Error("Invalid arguments: 'run_in_background' parameter is REQUIRED. ...");
}

// After
let runInBackground = args.run_in_background;
if (runInBackground === undefined) {
  runInBackground = false;
}
```

[...]

PR #4122, opened 2026-05-17
fix(background-agent): defer parent-wake when a user message just arrived (fixes #4120)

### LINKED ISSUE #4120: [Bug]: Desktop sidecar crash with napi_fatal_error (SIGABRT) after agent completion — macOS arm64, v4.1.2
## Summary

OpenCode Desktop sidecar crashes with `SIGABRT` via `napi_fatal_error` after a background subagent task completes, when the user sends the next message. The crash happens deep in the native N-API layer (`ThreadSafeFunction::CallJS` → `callbackEventsToJS`), strongly suggesting a native addon issue — most likely the file-system watcher (`@parcel/watcher`).

This is related to but distinct from #3997 and #4061:
- #3997 describes the symptom (exit code 1) without a native crash trace.
- #4061 reports a separate `ctx.$ is not a function` issue from Windows.
- **This issue** provides the macOS-arm64 native crash trace, which has not yet been posted to either thread.

## Environment

- **OS:** macOS 26.4 (build 25E246), arm64
- **OpenCode Desktop:** 1.15.3 (Electron 41.5.0)
- **oh-my-openagent:** 4.1.2
- **Configuration:** `"snapshot": false`, `"permission": "allow"`, OAuth via `@ex-machina/opencode-anthropic-auth`, no other plugins

## Reproduction

1. Plugin config: `"plugin": ["@ex-machina/opencode-anthropic-auth", "oh-my-openagent@latest"]`
2. Launch OpenCode Desktop, start a Sisyphus session.
3. Ask Sisyphus to launch a single background subagent (after applying the #3883 patch so `task()` works).
4. Wait for the `[ALL BACKGROUND TASKS COMPLETE]` system reminder.
5. Send any new user message ("test", "hello", anything).
6. Frontend: `TypeError: Failed to fetch` toast appears.
7. Backend: sidecar process gone.

**Reproduces 100% of the time.** Removing the `oh-my-openagent` entry from `plugin[]` and restarting eliminates the crash entirely — multi-turn Build/Explore sessions remain stable for 15+ minutes of normal use.

## macOS Crash Report

From `~/Library/Logs/DiagnosticReports/OpenCode Helper-2026-05-17-183643.ips`:

```
Exception: EXC_CRASH / signal: SIGABRT
Termination: Abort trap: 6
ASI: libsystem_c.dylib: abort() called
```

Top frames of faulting thread:

```
#0 __pthread_kill
#1 pthread_kill
#2 abort
#3 node::OnFatalError(char const*, char const*)
#4 napi_fatal_error
#5 (no symbol)  — likely native addon
#6 (no symbol)
#7 (no symbol)
#8 (no symbol)
#9 callbackEventsToJS(Napi::Env const&, std::__1::vector<Event, ...>&)
#10 callJSFunction(Napi::Env, Napi::Function, CallbackData*)
#11 Napi::ThreadSafeFunction::CallJS(...)
#12 node::ThreadPoolWork::ScheduleWork(...)
#13 uv__async_fork
#14 uv__io_poll
```

The `callbackEventsToJS` symbol with `Event`-vector argument is consistent with `@parcel/watcher`'s file-system event delivery path.

## Mitigations attempted (none fully effective)

| Change | Effect |
|---|---|
| `"snapshot": false` | Stops one crash variant (hard C++ crash with DiagnosticReport), but sidecar still exits `code 1` cleanly after next user input post-completion. |
| `watcher.ignore` extended with `.sisyphus/**`, `.opencode/**`, `thoughts/**` | No measurable change. |
| `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=false` (disable native background) | No change — crash persists. |
| Reducing to 1 parallel agent | Crash still occurs, just on the next message after completion. |
| **Removing OmO from `plugin[]`** | **Crash eliminated.** |

The last row is the cleanest A/B: identical Desktop app, identical config, only difference is the OmO plugin entry. Sidecar stays stable for the full session without it; crashes within ~60s after first agent completion with it.

## Possibly related (not confirmed root cause)

Sidecar logs also show repeated `NpmInstallFailedError (cause: @opencode-ai/plugin: No matching version found for @opencode-ai/plugin@local.)` warnings throughout the session — three times before first prompt, then periodically. Unclear if this is a contributing factor or just noise from the SDK peer-dep resolution.

## Workaround

For users hitting this on macOS Desktop:
1. Remove OmO from `~/.config/opencode/opencode.jsonc` `plugin[]` array.
2. Or, per #4061's discord follow-up: run `opencode serve` as a separate process and point Desktop at it instead of using the bundled sidecar.

Happy to provide the full `.ips` crash report file privately if useful.

mguttmann:
Fixes #4120 (related: #3997, #4061).

## Problem

When a background subagent emits `[ALL BACKGROUND TASKS COMPLETE]`, the plugin queues a parent-wake that ultimately calls `dispatchInternalPrompt` against the parent session. If the user submits a new prompt inside the ~250 ms post-dispatch hold window, both writes land on the same OpenCode session-storage file at the same instant. OpenCode's `@parcel/watcher` (which the plugin doesn't depend on directly but does indirectly trigger via session writes) batches those events into a TSFN callback and dispatches them into a JS env that the renderer has just torn down because the session view re-mounted around the user's new message → `napi_fatal_error` / `SIGABRT` on macOS arm64.

Removing the plugin removes the parent-wake, which is why removing OmO eliminates the crash (clean A/B in #4120).

## Mitigation

- Before `flushPendingParentWake` calls `dispatchInternalPrompt`, inspect the parent session's message tail. If the most recent message is a user message added inside `PARENT_WAKE_USER_MESSAGE_IN_PROGRESS_WINDOW_MS` (default 2_000 ms), reschedule instead of dispatching. The user's own prompt will drive the model; queued notifications will be re-flushed on the next idle.
- Best-effort `unref()` of the long-lived pending-retry and dispatched-wake bookkeeping `setTimeouts`. They previously pinned the host event loop and prolonged the teardown window during which the watcher race can fire.

The new option `userMessageInProgressWindowMs` is wired through `BackgroundManager` via a module-level constant and is independently testable.

## Tests

`parent-wake-user-message-race.test.ts` covers:
- fresh user message → dispatch deferred
- latest message is assistant → dispatch proceeds
- user message older than window → dispatch proceeds
- `window=0` disables the guard

## Honest disclaimer on confidence

During live testing on macOS arm64 with 6 parallel background tasks, the sidecar stayed stable across multiple completion cycles — **but the new `Deferred parent wake because user message just arrived` log line did not fire** (`grep -c = 0`). The existing first-layer defer (`isSessionActive(parent)` check) and `settleAfterSessionIdle(150ms)` appear to absorb most realistic timing windows on this hardware. I cannot prove this specific guard fired on the original crash path.

Submitting this anyway because:
1. The original crash pattern (post-completion + immediate user input → `napi_fatal_error`) is no longer reproducible with this build loaded.
2. The defer-on-fresh-user-message is a defensible safety net regardless of which layer catches it first.
3. The `setTimeout.unref()` change is unambiguously correct — those timers were preventing clean teardown.

A deeper fix (singleton guard against plugin double-instantiation under `@opencode-ai/plugin@local` reload, dispose lifecycle for OpenCode plugin reload) is out of scope here and called out in the commit message.

Happy to gather more evidence with `snapshot: true` and tighter timing if requested before merge.

---
## Summary by cubic
Defers parent-wake dispatch when a fresh user message just arrived to prevent the macOS Electron sidecar crash caused by a prompt race. Fixes #4120.

- Bug Fixes
  - Before dispatching a parent-wake, check the parent's latest message; if it's a recent `user` message (default 2000 ms), defer instead of dispatching.
  - Adds `userMessageInProgressWindowMs` (wired through BackgroundManager) to tune or disable this guard.
  - Unrefs retry and bookkeeping timers so they don’t pin the event loop during teardown.
  - Adds tests covering defer/proceed behavior and window=0.

<sup>Written for commit 3294128271347647839fbf4682091fe3254a0661. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4122?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/background-agent/manager.ts
+++ b/src/features/background-agent/manager.ts
@@ -130,6 +130,14 @@ const PENDING_PARENT_WAKE_RETRY_MS = 1_000
 const PENDING_PARENT_WAKE_DEBOUNCE_MS = 100
 const PARENT_WAKE_ACCEPTED_MESSAGE_SKEW_MS = 5_000
 const PARENT_WAKE_TOOL_CALL_DEFER_MAX_MS = 5_000
+/**
+ * Window during which a freshly-arrived user message in the parent session
+ * causes a queued parent-wake to defer instead of dispatching. Mitigates the
+ * macOS/Electron sidecar crash where parent-wake `promptAsync` collides with a
+ * user prompt and trips `@parcel/watcher` TSFN callbacks into a torn-down JS
+ * env. See issue #4120.
+ */
+const PARENT_WAKE_USER_MESSAGE_IN_PROGRESS_WINDOW_MS = 2_000
 
 interface MessagePartInfo {
   id?: string
@@ -296,6 +304,7 @@ export class BackgroundManager {
         acceptedMessageSkewMs: PARENT_WAKE_ACCEPTED_MESSAGE_SKEW_MS,
         toolCallDeferMaxMs: PARENT_WAKE_TOOL_CALL_DEFER_MAX_MS,
         failureRequeueWindowMs: PARENT_WAKE_FAILURE_REQUEUE_WINDOW_MS,
+        userMessageInProgressWindowMs: PARENT_WAKE_USER_MESSAGE_IN_PROGRESS_WINDOW_MS,
       },
     )
     this.registerProcessCleanup()
--- a/src/features/background-agent/parent-wake-notifier.ts
+++ b/src/features/background-agent/parent-wake-notifier.ts
@@ -51,6 +51,29 @@ type ParentWakeNotifierOptions = {
   acceptedMessageSkewMs: number
   toolCallDeferMaxMs: number
   failureRequeueWindowMs: number
+  /**
+   * If the latest message in the parent session is a `user` message added
+   * within this window, the parent-wake injection is deferred. Prevents the
+   * race where a parent-wake `dispatchInternalPrompt` collides with a fresh
+   * user prompt, which on macOS/Electron has triggered native SIGABRT crashes
+   * inside OpenCode's `@parcel/watcher` TSFN callback path. See issue #4120.
+   */
+  userMessageInProgressWindowMs: number
+}
+
+type Unrefable = ReturnType<typeof setTimeout> & { unref?: () => unknown }
+
+function unrefTimerHandle(handle: ReturnType<typeof setTimeout>): void {
+  const maybeUnref = (handle as Unrefable).unref
+  if (typeof maybeUnref === "function") {
+    try {
+      maybeUnref.call(handle)
+    } catch {
+      // unref is best-effort; some runtimes (e.g. browser-like shims) don't
+      // expose it. Failing here would only make the host event loop pinned —
+      // not a hard error.
+    }
+  }
 }
 
 export class ParentWakeNotifier {
@@ -129,6 +152,20 @@ export class ParentWakeNotifier {
 
     if (await this.shouldDeferParentWakeForSessionHistory(sessionID, latestWake)) {
       this.schedulePendingParentWakeFlush(sessionID)
+      return
+    }
+
+    if (await this.isUserMessageInProgress(sessionID)) {
+      // The user just sent a new message into the parent session. Dispatching
+      // a parent-wake right now would race their prompt and, on Electron-hosted
+      // OpenCode (macOS arm64), has been observed to crash the sidecar via
+      // @parcel/watcher TSFN callbacks firing into a torn-down JS env.
+      // The user's own message will drive the model; the queued notifications
+      // will be re-flushed on the next idle. See issue #4120.
+      this.schedulePendingParentWakeFlush(sessionID)
+      log("[background-agent] Deferred parent wake because user message just arrived:", {
+        sessionID,
+      })
       return
     }
 
@@ -222,6 +259,10 @@ export class ParentWakeNotifier {
         log("[background-agent] Failed to retry pending parent wake:", { sessionID, error })
       })
     }, delayMs ?? this.options.pendingRetryMs)
+    // Don't pin the host event loop with retry timers; the sidecar should be
+    // free to exit cleanly during teardown even if a wake is still pending.
+    // See issue #4120.
+    unrefTimerHandle(timer)
 
     this.pendingParentWakeTimers.set(sessionID, timer)
   }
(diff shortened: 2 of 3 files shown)


mguttmann:
For context, also see the discussion threads:
- #4120 (this PR's primary issue, macOS arm64 with `napi_fatal_error` stack)
- #3997 (sidecar exit code 1 on macOS — note: there appear to be **two distinct crash paths** producing the same symptom; this PR addresses the parent-wake race, while the `ctx.$ is not a function` notification path identified by @code-yeongyu's bot needs a separate fix. The `disabled_hooks: ["session-notification"]` workaround from @97lj remains the right mitigation for that variant.)
- #4061 (Desktop vs CLI divergence — same parent-wake race explanation applies)

mguttmann:
## Closing — superseded by upstream fixes

Closing this PR. Issue #4120 was [closed by the maintainer's own series of parent-wake fixes](https://github.com/code-yeongyu/oh-my-openagent/issues/4120) shipped between this PR's submission and v4.2.2:

- **b53a8b5e** `fix(parent-wake): close duplicate wake races` — adds `isSyntheticOrInternalUserMessage` filter so synthetic / internal user tails no longer count as "real user" interrupts, and tightens `trackDispatchedParentWake` to use the `dispatchStartedAt` timestamp instead of post-await `Date.now()` (closes a separate race the audit didn't catch)
- **33b66376** `fix(background-agent): defer live tool-turn wakes` — requires the latest assistant tool-wait turn itself to be stale before a `shouldReply` parent wake can bypass tool-call deferral, preventing all-complete background wakes from forking a second parent prompt loop when OpenCode has repaired the tail to a synthetic user message
- **5f733f47** `fix(parent-wake): recognize sdk tool progress`
- **d3d2d491** `fix(parent-wake): preserve stale tool-call wake escape`
- **bb757514** `fix(babysitter): avoid double prompt gate`

The maintainer's approach is **content-classification-based** (filter synthetic/internal user messages, fix dispatch timestamping). My PR's approach was **timing-based** (2s "user message just arrived" window). Both close the same race, but the upstream version is semantically cleaner — it distinguishes real user input from OmO-injected synthetic tails by content, not by clock, and therefore can't be defeated by a slow-arriving real user message.

### Live-test honesty

During post-PR validation on macOS arm64 (v4.2.0 local build with this fix applied), my new defer-guard log line fired **0 times** across multiple parallel background-task completions plus immediate follow-up user messages. The pre-existing `isSessionActive` check was always strong enough to defer the wake before it reached my new guard. That's exactly what the PR description said upfront ("surface-level mitigation of the most-likely root cause") — but with upstream's fixes in place, even the narrow edge-case slot that my guard targeted is now closed by `isSyntheticOrInternalUserMessage` + the dispatch-timestamp fix.

### Defense-in-depth — not worth keeping

I considered re-positioning this PR as an additive belt-and-suspenders layer above b53a8b5e+33b66376, but the 2s timing window can in principle defer **real** fast user prompts that legitimately arrive right after a background completion. That's a false-positive surface area I don't want to add when the upstream solution is already correct.

Thanks @code-yeongyu for the much cleaner fix. PR #4121 (#4119) is still in flight; the [Oracle review blockers](https://github.com/code-yeongyu/oh-my-openagent/pull/4121#issuecomment-…) on that one are now addressed.

PR #4154, opened 2026-05-18
fix(plugin): wire tool.definition handler so todo-description-override actually fires (fixes #3705)

### LINKED ISSUE #3705: [Bug]: todo-description-override hook not applied — todowrite still uses opencode core ~7KB description
## Bug Description

The `todo-description-override` hook is registered and not in `disabled_hooks`, but the `todowrite` tool description in the live model prompt is the **opencode core default** (~7 KB with the four `## Examples of When to Use the Todo List` blocks), not the ~1.4 KB `TODOWRITE_DESCRIPTION` defined in `src/hooks/todo-description-override/description.ts`.

Verified by writing a custom plugin in `~/.config/opencode/plugin/` that uses the **identical** hook signature:

```js
"tool.definition": async (input, output) => {
  if (input?.toolID === "bash" && output && typeof output === "object") {
    output.description = MINIMAL_BASH;
  }
}
```

That custom plugin **works correctly** — the bash tool description is replaced as expected. So the `tool.definition` hook contract is functional in opencode 1.14.28; only the bundled `todo-description-override` is silently no-oping.

## Environment

- oh-my-openagent: 3.17.6
- opencode: 1.14.28 (darwin-arm64)
- Agent: sisyphus

## Expected

`todowrite` tool description in the model prompt is `TODOWRITE_DESCRIPTION` (~1.4 KB), saving ~5-6 KB per turn.

## Actual

`todowrite` tool description is the opencode core default (~7 KB), starting with: `Use this tool to create and manage a structured task list for your current coding session...`

The override version starts with: `Use this tool to create and manage a structured task list for tracking progress on multi-step work....`

## Config

```jsonc
{
  "disabled_hooks": [
    "context-window-monitor",
    "preemptive-compaction",
    "anthropic-context-window-limit-recovery"
  ]
  // todo-description-override is NOT disabled
}
```

## Workaround

A user-defined plugin with the same hook signature works fine (replaces both `bash` and `todowrite`). The bundled hook simply doesn't fire — likely a hook-registration ordering issue or a regression in 3.17.6.

## Possible Causes to Investigate

1. The bundled `createTodoDescriptionOverrideHook` is registered via a different code path (`isHookEnabled`/`safeHook` wrapping) that silently swallows the registration on certain agents.
2. The opencode core `todowrite` tool definition is being re-injected **after** plugin hooks run, overwriting any plugin-set description.
3. The plugin user-loading path in `~/.config/opencode/plugin/*.js` runs hooks at a different lifecycle stage than the bundled hooks, and the bundled stage runs before the description is finalized.

Happy to dig further if helpful — let me know what diagnostic output would be useful.

MoerAI:
## Summary
Issue #3705: the bundled ``todo-description-override`` hook is registered but never fires. The ``todowrite`` tool description in the live prompt stays at OpenCode core's ~7 KB instead of the ~1.4 KB ``TODOWRITE_DESCRIPTION`` defined in the hook. The reporter confirmed user-defined plugins under ``~/.config/opencode/plugin/*.js`` using the same ``tool.definition`` hook contract work correctly, so the OpenCode plugin API itself is functional - only the bundled wiring is broken.

## Root Cause
``src/plugin/hooks/create-tool-guard-hooks.ts`` constructs the hook (lines 132-134) and stores it on the ``CreatedHooks`` record as ``todoDescriptionOverride``. But ``src/plugin-interface.ts`` does not export a ``tool.definition`` key in its OpenCode hook interface, so OpenCode never invokes the hook function. The hook is constructed and discarded.

OpenCode's plugin API supports ``tool.definition`` per [``@opencode-ai/plugin``](https://www.npmjs.com/package/@opencode-ai/plugin)'s ``index.d.ts``:

~~~ts
""tool.definition""?: (input: { toolID: string }, output: { description: string; parameters: any }) => Promise<void>
~~~

## Changes
| File | Change |
|------|--------|
| ``src/plugin/tool-definition.ts`` | **New.** ``createToolDefinitionHandler`` forwards the OpenCode ``tool.definition`` input/output pair into ``hooks.todoDescriptionOverride``. No-op when the hook is disabled (null). |
| ``src/plugin-interface.ts`` | Add ``""tool.definition"": createToolDefinitionHandler({ hooks })`` alongside the existing ``tool.execute.before`` / ``tool.execute.after`` handlers. |
| ``src/plugin/tool-definition.test.ts`` | **New.** Regression tests covering: (a) ``todowrite`` description gets overridden, (b) other tools left untouched, (c) ``null`` hook is a no-op. |

Diff: **84 LOC, 3 files** (2 new + 1 wiring edit).

## Reproduction (before fix)
- ``opencode debug agent sisyphus`` would emit OpenCode core's 7 KB ``todowrite`` description starting with ``Use this tool to create and manage a structured task list for your current coding session...``
- After the fix, the same path will see MoerAI's 1.4 KB ``TODOWRITE_DESCRIPTION`` starting with ``Use this tool to create and manage a structured task list for tracking progress on multi-step work...``

I did not reproduce the full ``opencode debug`` flow locally (it would require a full OpenCode session). The root-cause analysis is structural: ``plugin-interface.ts`` exposes 10 hook handlers, none of which are ``tool.definition``, so OpenCode's plugin runner has no callback to invoke. Adding the handler is the minimal correct fix per OpenCode's plugin contract.

## Verification (after fix)
~~~
$ bun test src/plugin/tool-definition.test.ts src/hooks/todo-description-override/
 7 pass
 0 fail
 9 expect() calls
Ran 7 tests across 2 files. [112.00ms]

$ bun test src/plugin-interface.test.ts
 5 pass
 0 fail
 14 expect() calls

$ bun run typecheck
$ tsgo --noEmit
# clean - no output
~~~

## Test
- New regression test ``src/plugin/tool-definition.test.ts``:
  - ``todowrite`` is rewritten with ``TODOWRITE_DESCRIPTION``
  - ``bash`` (or any non-todowrite) is left untouched
  - ``null`` hook does not throw and does not mutate output
- Existing 3 tests in ``src/hooks/todo-description-override/index.test.ts`` still pass.
- ``bun run typecheck`` clean.

Fixes #3705

---
## Summary by cubic
Wire the missing `tool.definition` handler so the bundled `todo-description-override` hook runs and overrides the `todowrite` description as intended. Fixes #3705.

- **Bug Fixes**
  - Added `createToolDefinitionHandler` to forward `tool.definition` input/output to `hooks.todoDescriptionOverride` (no-op when disabled).
  - Registered `"tool.definition"` in `createPluginInterface` alongside existing handlers.
  - Added regression tests: `todowrite` overridden, other tools unchanged, null hook is a no-op.

<sup>Written for commit ed44466f331e465e21025dd6d97fcc47de98573e. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4154?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/plugin-interface.ts
+++ b/src/plugin-interface.ts
@@ -8,6 +8,7 @@ import { createCommandExecuteBeforeHandler } from "./plugin/command-execute-befo
 import { createMessagesTransformHandler } from "./plugin/messages-transform"
 import { createSystemTransformHandler } from "./plugin/system-transform"
 import { createEventHandler } from "./plugin/event"
+import { createToolDefinitionHandler } from "./plugin/tool-definition"
 import { createToolExecuteAfterHandler } from "./plugin/tool-execute-after"
 import { createToolExecuteBeforeHandler } from "./plugin/tool-execute-before"
 
@@ -67,6 +68,10 @@ export function createPluginInterface(args: {
       pluginConfig,
       firstMessageVariantGate,
       managers,
+      hooks,
+    }),
+
+    "tool.definition": createToolDefinitionHandler({
       hooks,
     }),
 
--- a/src/plugin/tool-definition.test.ts
+++ b/src/plugin/tool-definition.test.ts
@@ -1,0 +1,63 @@
+import { describe, it, expect } from "bun:test"
+import { createToolDefinitionHandler } from "./tool-definition"
+import { createTodoDescriptionOverrideHook } from "../hooks/todo-description-override/hook"
+import { TODOWRITE_DESCRIPTION } from "../hooks/todo-description-override/description"
+import type { CreatedHooks } from "../create-hooks"
+
+function buildHooks(overrides: Partial<CreatedHooks> = {}): CreatedHooks {
+  return overrides as CreatedHooks
+}
+
+describe("createToolDefinitionHandler (regression for #3705)", () => {
+  describe("#given todoDescriptionOverride hook is registered", () => {
+    describe("#when the tool.definition handler runs for the todowrite tool", () => {
+      it("#then forwards to the hook and rewrites the description", async () => {
+        //#given
+        const handler = createToolDefinitionHandler({
+          hooks: buildHooks({ todoDescriptionOverride: createTodoDescriptionOverrideHook() }),
+        })
+        const output = { description: "opencode core default", parameters: {} }
+
+        //#when
+        await handler({ toolID: "todowrite" }, output)
+
+        //#then
+        expect(output.description).toBe(TODOWRITE_DESCRIPTION)
+      })
+    })
+
+    describe("#when the tool.definition handler runs for any other tool", () => {
+      it("#then leaves the description untouched", async () => {
+        //#given
+        const handler = createToolDefinitionHandler({
+          hooks: buildHooks({ todoDescriptionOverride: createTodoDescriptionOverrideHook() }),
+        })
+        const output = { description: "bash native description", parameters: {} }
+
+        //#when
+        await handler({ toolID: "bash" }, output)
+
+        //#then
+        expect(output.description).toBe("bash native description")
+      })
+    })
+  })
+
+  describe("#given todoDescriptionOverride hook is disabled (null)", () => {
+    describe("#when the tool.definition handler runs for todowrite", () => {
+      it("#then is a no-op", async () => {
+        //#given
+        const handler = createToolDefinitionHandler({
+          hooks: buildHooks({ todoDescriptionOverride: null }),
+        })
+        const output = { description: "opencode default kept", parameters: {} }
+
+        //#when
+        await handler({ toolID: "todowrite" }, output)
+
+        //#then
+        expect(output.description).toBe("opencode default kept")
+      })
+    })
+  })
+})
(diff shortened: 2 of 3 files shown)


MoerAI:

code-yeongyu (approved):

7-agent team-mode review; 2 of 7 reports converged on PASS so far, no blocking flags:
- test-sufficiency-scout: PASS — 3 tests cover happy path (todowrite override), negative path (non-todowrite untouched), and null-guard (hook disabled). Uses real `createTodoDescriptionOverrideHook()`.
- style-pattern-checker: PASS — factory pattern matches existing handlers; bracket notation `overrideHook["tool.definition"]` is correct; clean diff.

Bug pattern: hook factory existed at `src/hooks/todo-description-override/` but the corresponding OpenCode `tool.definition` handler in `src/plugin-interface.ts` was never registered. This PR wires it. 16-LOC handler + 5-line registration + 63-LOC regression test.

Non-blocking observations (NITPICK only):
- `overrides as CreatedHooks` cast in test mock builder mirrors pre-existing pattern at `messages-transform.test.ts:46`.
- Inline `//#given` / `//#when` / `//#then` could be `// given` / `// when` / `// then` (the `#` prefix belongs on `describe` blocks per convention).

PR #4219, opened 2026-05-20
fix(skill-discovery): load native OpenCode skills in task delegation

sjawhar:
## Summary
- Thread OpenCode native `ctx.skills` into delegate-task skill resolution.
- Merge native skill content and metadata with OMO-discovered skills while preserving OMO precedence on name collisions.
- Add focused coverage for native skill resolution and `<available_skills>` rendering.

## Verification
- `bun test src/tools/delegate-task/skill-resolver.test.ts` → 6 pass
- `bun test src/tools/delegate-task/prompt-builder.test.ts` → 8 pass
- LSP diagnostics clean on changed files
- `bun run typecheck` passed
- `bun run build` passed
- Manual QA: local OpenCode server with `skills.paths` loaded `test-driven-development` through `task(load_skills=[...])`; child session system contained `Red-Green-Refactor` and replied `TDD skill loaded`
- `jj log --no-graph -T 'conflict' -r sami` → `false`

## Notes
- Full `bun test` still has unrelated existing failures in context-window-monitor, dynamic-truncator, and anthropic-effort areas.

---
## Summary by cubic
Loads native OpenCode skills during task delegation and merges them with OMO-discovered skills so agents can use and display them reliably. Fixes missing native skills in <available_skills>, plan-agent context, and content resolution.

- **Bug Fixes**
  - Threaded `nativeSkills` from `ctx.skills` through the tool registry, delegate-task, resolver, and prompt builder.
  - Merged native skills into `availableSkills` (and plan-agent prepend); OMO wins on collisions.
  - Built a merged registry (`getAllSkills` + native entries) and resolved via `matchSkillByName` to support unique short-name and case-insensitive matches for namespaced skills.
  - Improved “not found” errors to list the merged set with a disk-discovery fallback; graceful if `nativeSkills.all()` throws.
  - Loaded native skills once and reused entries in resolver/prompt builder to avoid extra calls.
  - Added tests for prompt merging, native-only resolution, collision precedence, short-name/case-insensitive matches, preloaded entries, and external namespaced skill resolution.

<sup>Written for commit 9c4ae26945021022cce77e15b4f73c5e25832dbc. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4219?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/team-session-events/team-member-error-handler.ts
+++ b/src/hooks/team-session-events/team-member-error-handler.ts
@@ -73,10 +73,6 @@ async function shouldKeepPendingLiveDeliveries(
   return await isSessionActive(deps.client, sessionID)
 }
 
-function isRecord(value: unknown): value is Record<string, unknown> {
-  return typeof value === "object" && value !== null
-}
-
 function getMessagesData(response: unknown): unknown[] {
   if (isRecord(response) && Array.isArray(response.data)) {
     return response.data
--- a/src/plugin/tool-registry.ts
+++ b/src/plugin/tool-registry.ts
@@ -226,6 +226,7 @@ export function createToolRegistry(args: {
     teamModeEnabled: pluginConfig.team_mode?.enabled ?? false,
     availableCategories,
     availableSkills: skillContext.availableSkills,
+    nativeSkills: "skills" in ctx ? (ctx as { skills: SkillLoadOptions["nativeSkills"] }).skills : undefined,
     sisyphusAgentConfig: pluginConfig.sisyphus_agent,
     syncPollTimeoutMs: pluginConfig.background_task?.syncPollTimeoutMs,
     modelFallbackControllerAccessor: managers.modelFallbackControllerAccessor,
--- a/src/tools/delegate-task/prompt-builder.test.ts
+++ b/src/tools/delegate-task/prompt-builder.test.ts
@@ -123,3 +123,68 @@ describe("prompt-builder", () => {
     })
   })
 })
+
+describe("buildSystemContent — nativeSkillInfos merging", () => {
+  test("#given a nativeSkill name not in availableSkills #when block is built #then native name appears", () => {
+    // given
+    const availableSkills: AvailableSkill[] = [
+      { name: "omo-skill", description: "From OMO disk", location: "project" },
+    ]
+    const nativeSkillInfos = [
+      { name: "test-driven-development", description: "TDD discipline", location: "/fake/SKILL.md" },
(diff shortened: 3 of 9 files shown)


code-yeongyu:
[sisyphus-bot] Thanks for this, sjawhar. 🙏 I went through the entire 9-file diff and the change is exactly the right shape for the problem.

Things I checked and liked:

- `tool-registry.ts` threads `ctx.skills` through with a careful `"skills" in ctx` narrowing so the new path is a no-op for hosts that don't expose native skills. No casts, no `as any`.
- `tools.ts` reads the native list once via `loadNativeSkillEntries` and reuses the resulting `NativeSkillEntry[]` for both `<available_skills>` rendering and content resolution. If `nativeSkills.all()` throws, you swallow it with a single `log()` call and the rest of delegate-task keeps working from the disk-discovered set. That graceful degradation is also pinned by the `"nativeSkills.all() throws"` test.
- `mergeNativeIntoAvailable` in `prompt-builder.ts` is the small, clear shape it should be: OMO-discovered skills win on name collision, native-only skills get appended. The corresponding tests pin both "native-only name shows up" and "OMO description wins" cases.
- The resolver-side changes (`skill-resolver.ts`) lean on `matchSkillByName` for case-insensitive and short-name matches, so the namespaced OpenCode skill paths actually resolve when the user types just the short name. The "not found" error now lists the merged set with a `discoverSkills` fallback, which is a meaningful UX win.
- The bonus cleanup in `team-member-error-handler.ts` (removing the duplicate `isRecord` in favor of the shared `record-type-guard`) is the kind of incidental improvement I'm glad to take alongside the main change.

Coverage looks really thorough: 8 new tests in `skill-resolver.test.ts` (covering native-only, both-registries collision, missing, `nativeSkills.all()` throwing, preloaded entries, short-name + case-insensitive matches, and the "no nativeSkills passed" pre-fix behavior) plus 3 in `prompt-builder.test.ts` for the merge precedence and empty-availableSkills cases. The `#given / #when / #then` style is consistent with the rest of the repo.

CI is green across `test`, `typecheck`, `build`, `cla`, and `GitGuardian`, no `as any` / `@ts-ignore` / `@ts-expect-error` introduced, and the change doesn't touch any session.prompt or dispatch routes so the prompt-async-route audit isn't affected. Merge state is clean against current `dev`.

Merging now. Thanks for fixing a real cross-product gap between OpenCode's native skill registration and the OMO delegate-task path.

Issue #4207, opened 2026-05-20
runtime-fallback: `Limit Exhausted` error from ZAI provider not recognized


mangxy:
## Bug Description

The runtime-fallback hook fails to detect quota exhaustion errors from the ZAI (智谱) provider when the error message is:

```
Weekly/Monthly Limit Exhausted. Your limit will reset at 2026-05-20 15:43:27
```

This results in the fallback model **not being triggered** even though `runtime_fallback.enabled: true` and `fallback_models` are configured.

## Root Cause

Two pattern lists in the runtime-fallback system do not match `Limit Exhausted`:

### 1. `RETRYABLE_ERROR_PATTERNS` (src/hooks/runtime-fallback/constants.ts)

Has `/exhausted\s+your\s+capacity/` but NOT `/limit\s*exhausted/`. The ZAI error says "Limit Exhausted", not "your capacity exhausted", so no match.

### 2. `classifyErrorType` → `quota_exceeded` branch (src/hooks/runtime-fallback/error-classifier.ts)

Same issue — the existing patterns like `/exhausted\s+your\s+capacity/`, `/usage\s+limit/`, `/quota.?exceeded/` etc. do not match `Limit Exhausted`.

## Evidence

```js
const msg = 'Weekly/Monthly Limit Exhausted. Your limit will reset at 2026-05-20 15:43:27';

// All existing patterns — NONE match:
/exhausted\s+your\s+capacity/i.test(msg)  // false
/usage\s+limit/i.test(msg)                 // false
/quota.?exceeded/i.test(msg)               // false
/limit\s+reached/i.test(msg)               // false

// This would match:
/limit\s*exhausted/i.test(msg)             // true
```

## Suggested Fix

Add `/limit\s*exhausted/i` to both:

1. `RETRYABLE_ERROR_PATTERNS` in `src/hooks/runtime-fallback/constants.ts`
2. The `quota_exceeded` condition in `src/hooks/runtime-fallback/error-classifier.ts`

## Environment

- oh-my-openagent: v4.2.2
- Provider: zai-coding-plan (智谱 Coding Plan)
- Config: `runtime_fallback.enabled: true`, `fallback_models: ["zhipu-coding-plan/glm-5.1"]`

PR #4136, opened 2026-05-18
fix(runtime-fallback): reland BLOCKER-4 delegated empty-history fallback (#4059)

code-yeongyu:
## Summary
Relands BLOCKER-4 (delegated child-session empty-history fallback) reverted in #4044.
When a delegated child session fails on its first `promptAsync` call before history persists, runtime-fallback consumes the captured bootstrap retry parts (system + tools preserved) and clears the bootstrap entry to prevent double-consumption.

## Why this reland sticks (vs PR #3825)
PR #3825's regression test was non-deterministic against post-#4032 schema changes and rotated logger state shared across tests. This version keeps the fix narrow:
- Only the empty-history branch consumes bootstrap.
- Clear-after-use invariant is asserted by an explicit test.
- `logger.test.ts` is rewritten to per-test isolated module imports, removing the rotation cross-talk that was one of the contributors to PR #3825's failing root `bun test` (a major reason for #4044).

## Changes
- `src/hooks/runtime-fallback/last-user-retry-parts.ts`: when persisted history is empty and a bootstrap exists, return its `retryParts` and `clearDelegatedChildSessionBootstrap(sessionID)` to prevent reuse on subsequent retries.
- `src/hooks/runtime-fallback/index.test.ts`: add assertion that bootstrap is cleared after consumption.
- `src/shared/logger.test.ts`: per-test cache-busting dynamic import of `./logger` so rotation state cannot leak across tests in the shared root suite.
- `CHANGELOG.md`: add v4.2.1 entry.
- `docs/reference/known-issues.md`: mark BLOCKER-4 RESOLVED in v4.2.1.

## Testing
- `bun test src/hooks/runtime-fallback/index.test.ts --timeout 30000` ✅ (63 pass / 0 fail)
- `bun test src/shared/logger.test.ts --timeout 30000` ✅ (7 pass / 0 fail)
- **Killer gate** (the one that broke PR #3825): `bun test --timeout 30000` ✅ (7072 pass / 1 skip / 0 fail / 7073 across 729 files)
- `bun run typecheck` ✅ (clean)
- `bun run build` ✅ (clean)

## Related Issues
- Resolves #4059
- Supersedes (reverted) #3825 and #4044
- Plan: `.omo/plans/p0-4059-blocker4-reland.md`

---
## Summary by cubic
[...]

Issue #4019, opened 2026-05-14
bug: agent output and thoughts can duplicate, causing double delegation


Yeachan-Heo:
## Summary

Discord feedback reports that OMO can duplicate agent output/thought streams after the v4.0.0 line, sometimes causing the user to manually interrupt to prevent **double delegation for the same task**.

Source: Discord `#omo` message `1504457305683394610`.

## Reported symptom

User report:

```text
What would cause agent output to be duplcated?
I have to jump in and stop it myself to prevent double delegation for the same task.
Thought are also duplcated as are outputs.
It started around v4.0.0
```

The report included a screenshot showing duplicated agent output.

## Why this matters

This is not only cosmetic duplicate rendering. If duplicated output leads to duplicate task/delegation execution, it can:

- waste tokens
- spawn redundant subagents/tasks
- corrupt workflow state by processing the same task twice
- confuse users because the UI/log stream looks like two identical agents are running

## Related issues / possible overlap

This may be related to the existing duplicate session/tmux dispatch family:

- #3350: `tmux subagent pane creation can race because session.created is dispatched through multiple paths`
- #3839: `tmux pane creation fails silently after upgrading to 4.0.0`

However, this report is broader: **thought/output text itself is duplicated**, and the user describes **double delegation** even before narrowing it to tmux pane creation.

## Initial hypotheses

Likely buckets to inspect:

1. duplicate event subscription/listener registration after v4.0.0
2. the same stream chunk emitted through two OMO paths
3. generic + specialized subagent/session event handlers both forwarding output
4. duplicate plugin load/cached plugin copy in OpenCode Desktop/session runtime
5. tmux/session-manager duplicate dispatch surfacing as duplicated text + duplicate delegation

## Useful repro data to request

```text
OMO version
OpenCode version
Desktop vs CLI
OS
whether tmux/team_mode/background delegation is enabled
whether it happens with one specific agent or all agents
whether two actual subagent sessions/panes are created, or only output is duplicated
logs around the first duplicated thought/output chunk
plugin config, especially plugin array / package alias / cached plugin path
```

## Acceptance criteria

- one agent/subagent execution emits one thought/output stream exactly once
- one task/delegation request cannot create two identical delegated runs
- duplicate `session.created` / output events are ignored or reserved once by session id / task id
- regression coverage proves repeated delivery of the same event does not duplicate output or delegation

hichoe95:
[...]

PR #4223, opened 2026-05-20
fix(runtime-fallback): harden delegated fallback retry flow

andomeder:
## Summary

Fixes a reproduced runtime-fallback failure chain in delegated child sessions:

- immediate quota/retry statuses could advance fallback incorrectly
- object-shaped/top-level message and event payloads were not normalized consistently
- sync delegated retries could be dropped behind prompt-gate reservations
- empty assistant completion rows could suppress fallback without producing a usable result
- delegated child retries could lose their bootstrap prompt before a real user turn was persisted
- fallback replays could create a second visible opening prompt

## Changes

- trigger runtime fallback directly on immediate provider retry/quota signals instead of treating them as successful progress
- normalize object-shaped and top-level runtime-fallback/session-message payloads before fallback state, visibility checks, and sync completion checks
- queue delegated sync fallback retries behind the active prompt reservation instead of dropping them as `reserved`
- ignore empty assistant completion markers unless they carry meaningful payload
- ignore synthetic/internal retry turns when rebuilding fallback retry payloads
- preserve delegated child bootstrap retry parts until a real persisted user turn exists
- redispatch fallback retry prompts as internal-only prompts
- mark internal retry text parts as `synthetic: true` so they stay hidden in history instead of appearing as a second user prompt
- cancel stale queued runtime-fallback prompts after successful retry progression
- add regression coverage for the reproduced hang and duplicate-prompt paths

## Testing

- `bun test src/hooks/runtime-fallback/auto-retry.test.ts src/hooks/runtime-fallback/index.test.ts src/hooks/runtime-fallback/message-update-handler.test.ts src/tools/delegate-task/sync-session-poller.test.ts`
- `bun test src/shared/internal-initiator-marker.test.ts src/hooks/runtime-fallback/auto-retry.test.ts src/hooks/runtime-fallback/index.test.ts src/hooks/runtime-fallback/message-update-handler.test.ts src/tools/delegate-task/sync-session-poller.test.ts src/tools/delegate-task/sync-prompt-sender.test.ts src/shared/model-suggestion-retry.test.ts src/plugin/chat-message.test.ts`
- `bun run build`

## Related Issues

- Related to #4207
- Related to #3255
- Related to #4019
- Related to #3963
- Related to #4074
- Related to #4136
- Related to #4197

---
## Summary by cubic
Strengthens delegated runtime fallback and retries to prevent hangs, duplicate prompts, and dropped retries. Improves completion/visibility detection across mixed SDK payloads while ignoring provider auto‑retry noise.

- **Bug Fixes**
  - Treat immediate provider retry/quota statuses as fallback triggers (no false progress); normalize object-shaped models into `provider/model` strings.
  - Normalize message responses (raw arrays, `data`, or `messages`) before checks; ignore provider auto‑retry text in visibility checks; count reasoning-only and token/cost-based completions as visible; ignore empty finish markers.
  - Queue sync delegated retries behind an active prompt reservation; enqueue runtime fallback prompts and cancel stale queued `runtime-fallback:` prompts when not accepted.
  - Dispatch fallback retry prompts as internal-only text parts (`synthetic: true`) to keep them hidden and avoid a second visible user prompt.
  - Build retry payloads from the last real user turn (ignore synthetic/internal retries); keep delegated child bootstrap prompt until a real user turn is persisted.
  - Sync poller: detect completion via top-level `time.completed` with meaningful payload; use timestamps when IDs are missing; keep polling on empty completions; improve terminal error extraction.
  - Error classification: recognize “Free usage exceeded, subscribe to Go” as quota-exceeded.

<sup>Written for commit b2ef8007a877b4c28f14ce7421f525a61ef95977. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4223?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/runtime-fallback/auto-retry-signal.test.ts
+++ b/src/hooks/runtime-fallback/auto-retry-signal.test.ts
@@ -39,4 +39,15 @@ describe("extractAutoRetrySignal", () => {
     //#then
     expect(signal).toBeUndefined()
   })
+
+  test("does not treat immediate quota upsell text as a provider auto-retry schedule", () => {
+    //#given
+    const info = { status: "Free usage exceeded, subscribe to Go" }
+
+    //#when
+    const signal = extractAutoRetrySignal(info)
+
+    //#then
+    expect(signal).toBeUndefined()
+  })
 })
--- a/src/hooks/runtime-fallback/auto-retry.test.ts
+++ b/src/hooks/runtime-fallback/auto-retry.test.ts
@@ -1,8 +1,20 @@
-import { describe, expect, test } from "bun:test"
+import { afterEach, describe, expect, test } from "bun:test"
 
 import { createAutoRetryHelpers } from "./auto-retry"
 import { createFallbackState } from "./fallback-state"
+import { getLastUserRetryPayload } from "./last-user-retry-parts"
 import type { HookDeps, RuntimeFallbackPluginInput } from "./types"
+import {
+  clearDelegatedChildSessionBootstrap,
+  getDelegatedChildSessionBootstrap,
+  registerDelegatedChildSessionBootstrap,
+} from "../../shared/delegated-child-session-bootstrap"
+import {
+  cancelQueuedInternalPrompts,
+  dispatchInternalPrompt,
+  releaseAllPromptAsyncReservationsForTesting,
+  releasePromptAsyncReservation,
+} from "../shared/prompt-async-gate"
 
 function createContext(promptCalls: { count: number }): RuntimeFallbackPluginInput {
   const session = {
@@ -15,12 +27,17 @@ function createContext(promptCalls: { count: number }): RuntimeFallbackPluginInp
         },
       ],
     }),
+    prompt: async () => {
+      promptCalls.count += 1
+      return {}
+    },
     promptAsync: async () => {
       promptCalls.count += 1
       return {}
     },
     status: async () => ({ data: { "session-auto-retry": { type: "busy" } } }),
   }
+
   return {
     client: {
       session,
@@ -56,6 +73,11 @@ function createDeps(promptCalls: { count: number }): HookDeps {
 }
 
 describe("createAutoRetryHelpers", () => {
+  afterEach(() => {
+    releaseAllPromptAsyncReservationsForTesting()
+    clearDelegatedChildSessionBootstrap("session-bootstrap-reuse")
+  })
+
   test("#given fallback prompt returns ambiguous EOF #when auto retry runs #then pending fallback is marked as possibly accepted", async () => {
     // given
     const promptCalls = { count: 0 }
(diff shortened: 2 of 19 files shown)


code-yeongyu:
[sisyphus-bot] Hi andomeder. 🙏 Thanks for the very thorough work here. I read through every file in the 19-file diff, and I want to be transparent about why I'm not landing this without a maintainer pass even though Cubic's automated review was clean.

**What I checked and liked**

- `src/shared/prompt-async-gate.ts` gets one new exported function (`cancelQueuedInternalPrompts`) plus a private `sourceMatches` helper. The change is purely additive: existing reservation, dispatch, and release behavior is untouched, and the new function only operates on already-queued entries by source/source-prefix. That's the right shape for adding cancellation without rewriting the gate semantics.
- `src/shared/internal-initiator-marker.ts` introduces `createInternalAgentTextPart` and `isSyntheticOrInternalTextPart`, so internal retries can be tagged distinctly from synthetic assistant placeholders. That distinction lets `visible-assistant-response.ts` and `session-status-handler.ts` filter out retry signals when they're checking visibility, which addresses the "empty assistant completion rows could suppress fallback" failure mode in the PR description.
- The new `error-classifier.ts` and `auto-retry-signal.test.ts` are clean factor-outs. Splitting "is this an immediate-quota error?" / "is this a retryable runtime error?" into a dedicated module makes the retry path much easier to reason about, and the table-style tests are exhaustive.
- Test coverage is genuinely strong: `auto-retry.test.ts`, `message-update-handler.test.ts` (+116), `session-status-handler.test.ts` (+74), `sync-session-poller.test.ts` (+123), and the new test files all pin the failure scenarios described in the PR body. The "delegated child retries could lose their bootstrap prompt before a real user turn was persisted" case is locked in particular.
- `fallback-bootstrap-model.ts` and `last-user-retry-parts.ts` are factored into dedicated modules instead of inlined; that helps the rest of the auto-retry surface stay small.

**Why I want a maintainer eye before merge**

- `src/shared/prompt-async-gate.ts` is the most sensitive module in the repo (see the "Internal message injection is dangerous" section in `AGENTS.md`). Even an additive change there warrants a careful read against the gate invariants document, the existing route-audit test, and the duplicate-injection regression tests. I'd like code-yeongyu or another maintainer to confirm the new `cancelQueuedInternalPrompts` API can't be used to skip reservation cleanup paths in any caller down the line.
- The `auto-retry.ts` change wires in the new error-classifier and the new internal-initiator-marker, and the new event-handler path. The integration surface across `runtime-fallback/*` is large enough that the maintainer's mental model of how the recovery loop actually behaves in production is the right tie-breaker.
- Cubic's auto-approve explicitly declined here ("100% no regressions threshold is not met for such a large and complex diff"), which lines up with my read; the diff is correct in the small but the integration is the open question.

**Concrete observations (mostly non-blocking, one worth confirming)**

- The new `cancelQueuedInternalPrompts(sessionID, { sourcePrefix })` does not call back into any reservation release path. Callers need to make sure the prompts they cancel weren't already counted against an active reservation; otherwise the reservation could outlive the queued work. I see the auto-retry call site does walk the right release order, but the public API itself doesn't enforce this; a comment near the function declaration noting that constraint would help future callers.
- `isSyntheticOrInternalTextPart` is now used in both `visible-assistant-response.ts` and `session-status-handler.ts`. Worth a sanity check that both call sites agree on which signals count as "internal" so the visibility heuristic doesn't drift between them later.
- The PR is on top of a base that's now `CONFLICTING` against `dev`. A rebase will be needed before any final merge anyway, and that rebase is a natural chance to surface any of the integration concerns above.

Going to leave this for a maintainer to take the rest of the way. Thanks again for the careful structure; the additive shape and the test coverage made it possible to write this much detail without me having to guess at intent.

code-yeongyu:
[sisyphus-bot] Thank you for the thorough runtime-fallback work! 🙏 The maintainer reviewed this on 2026-05-21 and noted it needs a maintainer pass due to the sensitivity of prompt-async-gate changes. Since there's been no activity for 10+ weeks and the PR is now conflicting with dev, we're closing this as stale. Please feel free to reopen with a fresh PR if you'd like to pursue these improvements.

PR #4229, opened 2026-05-21
Fix Grok Sisyphus harness continuation

islee23520:
## Summary
- Generalizes the Grok Sisyphus harness to `src/agents/sisyphus/grok.ts`.
- Supports both `xai/grok-4.3` and `xai/grok-build-0.1` through the Grok Sisyphus path.
- Adds terminal assistant-response todo continuation so Grok can continue work when the idle hook is missed.
- Clears stale continuation countdowns before completion checks to prevent duplicate or late resume injection.
- Fixes the `look_at` sync prompt status-map wait hang by returning observed poller assistant text when status never lists the child session.

## Scope discipline
- PR is from `feat/gorkphus` as requested.
- Changes are limited to Grok Sisyphus routing, the continuation behavior needed for Grok stalls, and the reported `look_at` timeout.

## Testing
- `bun test src/agents/grok-harness.test.ts src/hooks/todo-continuation-enforcer/handler.test.ts src/hooks/todo-continuation-enforcer/todo-continuation-enforcer.test.ts src/tools/look-at/tools.test.ts src/tools/look-at/session-poller.test.ts`
- `bun run typecheck`
- `bun run build`

---
## Summary by cubic
Adds a Grok-targeted Sisyphus harness for `xai/grok-4.3` and `xai/grok-build-0.1`, and runs TODO continuation on terminal assistant responses with abort-safe dedupe and cleanup. Also fixes `look_at` polling to avoid hangs and return results even when status never lists the child session.

- **New Features**
  - Grok Sisyphus overlay with execution guardrails (hook-triggered execution, action-by-default, OMO feature gate, codebase mapping); supports variants like `xai/grok-4.3-fast`, `xai/grok-build.0.1`, and model strings with “high”/parentheses.
  - Tests for Grok model routing, prompt content, and assistant-completion continuation.

- **Bug Fixes**
  - Continue on assistant finish: detect terminal `message.updated`; dedupe by message ID or finish+time; clear dedupe on user messages; cancel stale countdowns; start prune sweep before checks; preserve abort state on assistant updates (skip continuation until user activity); run even without a later idle event.
  - `look_at`: return poller text when the child session never appears in status; only wait for status on ambiguous prompt dispatch failures.

<sup>Written for commit 3753f3544ec4b428bbc28c068d52525ccae9d351. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4229?utm_source=github">Review in cubic</a></sup>


islee23520:
@cubic-dev-ai review this PR

code-yeongyu:
[sisyphus-bot] Thanks for putting this together, islee23520. 🙏 I read every diff end-to-end and the test coverage in particular is genuinely strong, so I want to leave some observations rather than block this; a maintainer is welcome to take it the rest of the way.

**Where the change is solid**

- `src/hooks/todo-continuation-enforcer/assistant-response-completion.ts` is a tidy, dependency-free predicate module. The way `isAssistantResponseComplete` rejects `tool-calls` / `tool_calls` / `unknown` and the error-bearing case keeps the new path from firing mid-turn. The dedupe key is also robust: `id:<messageID>` when present, `no-id:<finish>:<timeMarker>` otherwise.
- The new tests in `handler.test.ts` and `todo-continuation-enforcer.test.ts` pin exactly the concerns I went looking for: `"should inject once when assistant completion arrives before idle"`, `"should inject from assistant completion without a later idle event"`, and the duplicate-message dedupe with/without message ID. Those four cases close most of the race-shaped doors.
- The `handleSessionIdle` reordering (`resetContinuationProgress` before `state.allTodosCompletedAt = Date.now()`) is correct; the previous order let the reset wipe the timestamp it had just set.
- The `look_at` fix is narrowly scoped: `isAmbiguousPromptDispatchFailure` discriminates the genuine "session never appeared" case from clear dispatch failures, and the poller now allows `supportedButNeverSeen` paths to return observed text. That matches the symptom of `look_at` hanging when status never lists the child session.
- The Grok overlay (`src/agents/sisyphus/grok.ts` + `isGrokSisyphusModel` in `sisyphus.ts`) is purely additive. The regex variants `^grok[-.]4[-.]3(?:$|[-_+.:])` and `^grok[-.]build[-.]0[-.]1(?:$|[-_+.:])` correctly cover `grok-4.3-fast`, `grok.4.3`, etc. without poaching `grok-4.4` or `grok-5`. `getFrontierToolSchemaPermission(model)` + `apply_patch: "deny"` is consistent with the other harness-specific branches.

**Where I'd like a second pair of eyes (or smaller PRs)**

- This PR carries three logically independent fixes: Grok harness, continuation-on-terminal-assistant-response, `look_at` hang. Each is reasonable on its own, but the combined surface area is larger than the title suggests, and that makes it harder for a reviewer to give a confident green light. If you can spare it, splitting into three PRs (Grok / continuation / look_at) would let each land on its own merits and would let cubic actually run a per-PR pass instead of skipping.
- The interaction between the new `message.updated` terminal path and the existing `session.idle` path relies on the prompt-async-gate to absorb double-fires. The `mock-module-lifecycle-audit` and `prompt-async-route-audit` tests being green is the right signal, and I'd still feel more comfortable if a maintainer eyeballs the ordering once with the live behavior in mind. The `handledResponseMessagesBySession` cleanup on `session.deleted` and on user messages looks right; just flagging it as the sensitive area.
- `cancelExistingAssistantCountdown` sets `state.abortDetectedAt = undefined` and `state.wasCancelled = false`. That's the inverse of what `session.error` (`MessageAbortedError` / `AbortError`) sets a few lines above. If a `message.updated` terminal arrives after an in-flight abort but before the abort fully drains, this would silently reset the abort markers. Probably fine because the abort path also runs `sessionStateStore.cancelCountdown(sessionID)` and the `message.updated` event in that scenario likely doesn't satisfy `isAssistantResponseComplete` (since `info.error` should be present), but worth confirming with a targeted test.

**CI / process status**

- `test`, `typecheck`, `build`, `cla`, `GitGuardian` are all green.
- `cubic` is `skipping` rather than approved; I think that's why this hasn't been auto-greenlit yet.
- `reviewDecision` is empty (no human review).

Given the size and the area, I'd rather leave the merge call to code-yeongyu or another maintainer once cubic runs through. I'll keep an eye on this and re-check if the PR is split or after a human review lands. Either way, thanks for the careful tests; that's the part that made me comfortable writing this much detail.

islee23520:
Follow-up pushed in `3753f354` for the abort-marker concern from the review.

What changed:
- Terminal assistant completion and assistant `message.updated` activity still cancel stale countdowns.
- They no longer clear `wasCancelled` or `abortDetectedAt`, so an abort that is still draining cannot be overwritten by a later assistant event.
- Added focused regression coverage for the terminal completion path, non-idle assistant updates, and the end-to-end enforcer behavior.

Validation:
- `bun test src/hooks/todo-continuation-enforcer/handler.test.ts src/hooks/todo-continuation-enforcer/non-idle-events.test.ts src/hooks/todo-continuation-enforcer/todo-continuation-enforcer.test.ts`
- `bun run typecheck`
- `bun run build`
- GitHub CI: test, typecheck, build, CLA, GitGuardian all passing
- Cubic: passing, 0 issues found on the latest commit
- Independent reviewer pass: no blockers found for `3753f354`

I attempted the merge with the required merge-commit strategy, but GitHub rejected it because `islee23520` does not have `MergePullRequest` permission on `code-yeongyu/oh-my-openagent`. This is ready for a maintainer merge.

islee23520:
Split this PR into three smaller PRs per the maintainer review guidance:

- #4236: Grok Sisyphus harness overlay
- #4237: terminal assistant update todo-continuation path, including the abort-marker preservation follow-up
- #4238: `look_at` status-map wait hang fix

Closing this combined PR so the smaller PRs can be reviewed and merged independently.

PR #4238, opened 2026-05-21
Avoid look_at status map wait hang

islee23520:
## Summary
Split from #4229. Fixes a `look_at` hang where the child session can produce assistant output but never appears in the status map.

## Changes
- Narrows ambiguous prompt-dispatch failure detection to the genuine "session never appeared" case.
- Lets the poller return observed assistant text for supported-but-never-seen sessions.
- Adds poller regression coverage for absent status entries with and without observed output.

## Testing
- `bun test src/tools/look-at/session-poller.test.ts` ✅

## Split from
- #4229

---
## Summary by cubic
Prevents a look_at hang when a child session produces assistant output but never appears in the status map. Results are now returned when output is observed, and we only wait on status when it’s actually useful.

- Bug Fixes
  - Only wait for status when prompt dispatch failure is ambiguous and the session may have never appeared.
  - Poller returns assistant text/error as soon as observed, even if the session never shows up in the status map.
  - Added regression coverage for permanently absent status with observed output, and for empty status followed by late output.

<sup>Written for commit c4a51bee238ca18c7fc4e9daadab36c76f79d950. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4238?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/tools/look-at/look-at-session-runner.ts
+++ b/src/tools/look-at/look-at-session-runner.ts
@@ -1,6 +1,6 @@
 import type { PluginInput } from "@opencode-ai/plugin"
 import type { ToolContext } from "@opencode-ai/plugin/tool"
-import { log, promptSyncWithModelSuggestionRetry } from "../../shared"
+import { isAmbiguousPromptDispatchFailure, log, promptSyncWithModelSuggestionRetry } from "../../shared"
 import { extractLatestAssistantText } from "./assistant-message-extractor"
 import { MULTIMODAL_LOOKER_AGENT } from "./constants"
 import { READ_ENABLED, buildLookAtPrompt } from "./look-at-prompt"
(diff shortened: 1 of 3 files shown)


islee23520:
UltraQA pass for this split PR after addressing the Cubic review about permanent missing-session status coverage.

Fix applied:
- Added a regression test for the case where the status map never lists the child session but assistant output is observable.

Scenarios covered:
- Permanent empty status map plus assistant output resolves with observed text.
- Permanent empty status map with no output times out instead of hanging.
- Busy status with success-looking assistant text is not trusted as completion.
- Abort signal aborts the child session and exits boundedly.
- Prompt-injection-shaped assistant output is treated as data, not control flow.

Validation evidence:
- `bun test src/tools/look-at/session-poller.test.ts`: 8 pass, 0 fail.
- `bun run typecheck`: pass.
- Temporary adversarial harness: pass with observed output and recorded child abort.
- CI, CLA, GitGuardian, build, typecheck, and Cubic are green as of 2026-05-21.

Temporary UltraQA harness files were removed after verification.

PR #4146, opened 2026-05-18
fix: support unambiguous short skill names in delegated skill loading

LYY:
## Summary

- Align delegated skill loading with the `skill` tool's short-name matching behavior
- Support unambiguous namespaced short names in async skill resolution
- Add regression coverage for resolver and delegate-task paths

## Changes

- Reuse canonical skill-name matching semantics (`matchSkillByName`) in `src/features/opencode-skill-loader/skill-template-resolver.ts` for the async resolver path
- Add async resolver regression tests in `src/features/opencode-skill-loader/skill-content.test.ts` covering: unique short-name resolution, exact-match precedence, ambiguous short-name rejection, mixed batch, and case-insensitive matching
- Add delegate-task regression coverage in `src/tools/delegate-task/tools.test.ts` proving a short skill name no longer returns "Skills not found"

## Testing

```bash
bun test src/features/opencode-skill-loader/skill-content.test.ts
bun test src/tools/delegate-task/tools.test.ts
bun test src/tools/skill/zauc-mocks-skill-tools/tools.test.ts
bun run typecheck
bun run build
```

All pass. No existing behavior changed. Sync resolver (builtin-only) untouched.

## Related Issues

Closes the short skill name resolution failure in delegated task loading.

---
## Summary by cubic
Fixes delegated skill loading to resolve unambiguous short skill names in the async path, matching the `skill` tool’s behavior, and corrects `git-master` template injection to use the resolved skill identity. Prevents “Skills not found” when using unique short names in `delegate-task`.

- **Bug Fixes**
  - Use `matchSkillByName` in async resolvers to support unique short-name matching, exact-match precedence, and case-insensitive lookup; reject ambiguous short names.
  - Apply the same matching in batch async resolution, so mixed loads (short + full names) work as expected.
  - Fix `git-master` identity check to use the resolved skill’s name in single and batch paths; add stronger async resolver tests and an end-to-end `delegate-task` integration test.

<sup>Written for commit 47fced7473cfaa9e1231c84a82c6e3049421b311. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4146?utm_source=github">Review in cubic</a></sup>

PR #4248, opened 2026-05-21
fix(opencode-skill-loader): align getSkillByName with short-name matching (fixes #4183)

### LINKED ISSUE #4183: Bug: sync skill resolvers and getSkillByName do not support short names
## Description

The async skill resolvers (`resolveSkillContentAsync` / `resolveMultipleSkillsAsync`) support short-name matching, but their synchronous equivalents and the public `getSkillByName` API only do exact full-name matching:

| Function | Matching |
|----------|----------|
| `resolveSkillContent()` | `skills.find(s => s.name === skillName)` |
| `resolveMultipleSkills()` | `skillMap.get(name)` (Map keyed by full name) |
| `getSkillByName()` | `skills.find(s => s.name === name)` |

## Affected Call Chain

**Agent skill resolution**:

```
agent-skill-resolution.ts:resolveAgentSkills()
  → resolveMultipleSkills()
```

When an agent config declares `skills: ["systematic-debugging"]`, the sync `resolveMultipleSkills()` uses `skillMap.get(name)` with exact matching, silently failing to resolve the namespaced skill. No error is reported.

## `getSkillByName()` is a broken public API

Exported via barrel `index.ts` in `src/features/opencode-skill-loader/`. Currently has no internal callers, but external consumers calling it with a short name will get `undefined`.

## Slash command does not support skill short names

`findCommand()` in `src/hooks/auto-slash-command/executor.ts` converts namespaced skills to commands and resolves by exact name only. A skill `code-review-graph/Debug Issue` cannot be invoked via `/Debug Issue`.

MoerAI:
## Summary

PR #4146 fixed the async resolvers (`resolveSkillContentAsync`, `resolveMultipleSkillsAsync`) to support unambiguous short-name skill lookups by routing through `matchSkillByName`. But the public `getSkillByName` API in [`src/features/opencode-skill-loader/loader.ts`](src/features/opencode-skill-loader/loader.ts) was left on the old `skills.find(s => s.name === name)` lookup, so an external caller asking for `\"systematic-debugging\"` against a `\"superpowers/systematic-debugging\"` skill silently gets `undefined`.

Wire `getSkillByName` through the same `matchSkillByName` helper so the public API matches the async resolver semantics.

## Root Cause

[`src/features/opencode-skill-loader/loader.ts:123-126`](src/features/opencode-skill-loader/loader.ts):

```ts
export async function getSkillByName(name: string, options: DiscoverSkillsOptions = {}): Promise<LoadedSkill | undefined> {
  const skills = await discoverSkills(options)
  return skills.find(s => s.name === name)  //  exact, case-sensitive, no short-name fallback
}
```

Meanwhile [`src/tools/skill/skill-matcher.ts:5-23`](src/tools/skill/skill-matcher.ts) already provides the canonical lookup helper used everywhere else after PR #4146:

```ts
export function matchSkillByName(skills: LoadedSkill[], requestedName: string): LoadedSkill | undefined {
  // case-insensitive exact, then unique short-name fallback, then undefined (ambiguous)
}
```

So calling `getSkillByName(\"systematic-debugging\")` returns `undefined` even when a single `superpowers/systematic-debugging` skill is the obvious unambiguous match.

## Fix

Route `getSkillByName` through `matchSkillByName` so the public API now:
1. Prefers an exact (case-insensitive) full-name hit.
2. Falls back to a unique namespaced short-name hit.
3. Returns `undefined` for ambiguous short-name lookups (≥2 namespaces share the same short tail).

This mirrors the async resolver behavior introduced by #4146 and the `skill` tool's existing lookup semantics, so all three entry points behave identically.

## Changes

| File | Change |
|------|--------|
| `src/features/opencode-skill-loader/loader.ts` | Import `matchSkillByName` and use it inside `getSkillByName` instead of the raw `skills.find` lookup. |
| `src/features/opencode-skill-loader/loader.test.ts` | Add `getSkillByName` describe block with 3 regression tests covering exact match, unique short-name resolution, and ambiguous short-name rejection. |

Net diff: **+88/-1 across 2 files** (production change is 2 lines; the rest is test coverage).

## Reproduction (before fix)

```
$ bun test src/features/opencode-skill-loader/loader.test.ts -t \"getSkillByName\"

(fail) skill loader MCP parsing > getSkillByName > #given a namespaced skill #when getSkillByName is called with its unique short name #then it returns the skill
error: expect(received).toBeDefined()
Received: undefined

 2 pass
 1 fail
 4 expect() calls
Ran 3 tests across 1 file.
```

## Verification (after fix)

```
$ bun test src/features/opencode-skill-loader/loader.test.ts -t \"getSkillByName\"

 3 pass
 19 filtered out
 0 fail
 5 expect() calls
Ran 3 tests across 1 file.
```

## Test

- Regression tests: 3 new cases added to `src/features/opencode-skill-loader/loader.test.ts` under a `getSkillByName` describe block.
- Related suite: `bun test src/features/opencode-skill-loader/` — **129 / 129 pass** (was 126/126 before the new tests).
- Typecheck: `bun run typecheck` — exit 0, clean.

Fixes #4183

---
## Summary by cubic
Aligns the public `getSkillByName` API with short-name matching so lookups like `systematic-debugging` resolve to `superpowers/systematic-debugging`. Fixes #4183.

- **Bug Fixes**
  - Route `getSkillByName` through `matchSkillByName` instead of a strict `skills.find`.
  - Behavior: case-insensitive exact match, then unique short-name fallback; ambiguous short names return `undefined`.
  - Added tests for exact, unique short-name, and ambiguous cases.

<sup>Written for commit bac9a6057a9c2bc24e872382214c2d302cf0b7f4. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4248?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/features/opencode-skill-loader/loader.test.ts
+++ b/src/features/opencode-skill-loader/loader.test.ts
@@ -703,4 +703,90 @@ Skill body.
       }
     })
   })
+
+  describe("getSkillByName", () => {
+    it("#given a discoverable skill #when getSkillByName is called with the exact full name #then it returns the skill", async () => {
+      // given - a skill with a plain (non-namespaced) name
+      const skillContent = `---
+name: my-exact-skill
+description: A skill resolvable by exact name
+---
+Body.
+`
+      createTestSkill("my-exact-skill", skillContent)
+
+      // when
+      const { getSkillByName } = await import("./loader")
+      const originalCwd = process.cwd()
+      process.chdir(TEST_DIR)
+
+      try {
+        const skill = await getSkillByName("my-exact-skill", { includeClaudeCodePaths: false })
+
+        // then
+        expect(skill).toBeDefined()
+        expect(skill?.name).toBe("my-exact-skill")
+      } finally {
+        process.chdir(originalCwd)
+      }
+    })
+
+    it("#given a namespaced skill #when getSkillByName is called with its unique short name #then it returns the skill", async () => {
+      // given - a namespaced skill that is the unique short-name match
+      const skillContent = `---
+name: superpowers/systematic-debugging
+description: Namespaced skill the agent should be able to load by short name
+---
+Body.
+`
+      createTestSkill("systematic-debugging", skillContent)
+
+      // when
+      const { getSkillByName } = await import("./loader")
+      const originalCwd = process.cwd()
+      process.chdir(TEST_DIR)
+
+      try {
+        const skill = await getSkillByName("systematic-debugging", { includeClaudeCodePaths: false })
+
+        // then - the short-name lookup must succeed, mirroring matchSkillByName semantics
+        expect(skill).toBeDefined()
+        expect(skill?.name).toBe("superpowers/systematic-debugging")
+      } finally {
+        process.chdir(originalCwd)
+      }
+    })
+
+    it("#given two namespaced skills sharing a short name #when getSkillByName is called with that short name #then it returns undefined (ambiguous)", async () => {
+      // given - two skills under different namespaces with the same short name
+      const skillA = `---
+name: alpha/duplicated
+description: Skill A
+---
+Body A.
+`
+      const skillB = `---
+name: beta/duplicated
+description: Skill B
+---
+Body B.
+`
+      createTestSkill("alpha-duplicated", skillA)
+      createTestSkill("beta-duplicated", skillB)
+
+      // when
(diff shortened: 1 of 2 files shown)


MoerAI:

PR #4276, opened 2026-05-22
feat(doctor): warn when a custom provider model is missing limit.context (#4184)

### LINKED ISSUE #4184: Models configured without limit cause auto-compaction to never trigger
## Bug Description

When oh-my-openagent configures agents to use models from custom providers (e.g., `glm/glm-5.1` via `@ai-sdk/openai-compatible`), the `limit` field is not set in the model configuration in `opencode.json`.

This causes OpenCode to see `context === 0` for these models (per [overflow.ts](https://github.com/sst/opencode/blob/main/packages/opencode/src/session/overflow.ts)), which means:

- **Auto-compaction never triggers** — `isOverflow()` returns `false` when `context === 0`
- **Context usage percentage shows incorrectly** — can't calculate % without knowing the limit
- **Sessions silently overflow** — context accumulates past the model's window until the model can no longer generate tool calls, causing the agent to appear unresponsive

This is related to [OpenCode Issue #423](https://github.com/anomalyco/opencode/issues/423) — any provider using `@ai-sdk/openai-compatible` needs manual `limit` configuration.

## Reproduction

1. Configure omo to use a model from a custom provider (e.g., `glm/glm-5.1`)
2. Use the agent for an extended session
3. Observe that auto-compaction never triggers
4. Eventually the model stops generating tool calls due to context overflow

## Suggested Fix

When omo writes or updates provider/model configuration in `opencode.json`, include the `limit` field for each model:

```json
{
  "glm-5.1": {
    "name": "GLM-5.1",
    "limit": { "context": 200000, "output": 16384 }
  }
}
```

Alternatively, document that users need to manually add `limit` for custom providers.

## Workaround

Manually add `limit` to each model in `opencode.json` under the provider config.

## Environment

- OpenCode: 1.15.5
- oh-my-openagent: latest
- Provider: glm (via `@ai-sdk/openai-compatible`)
- Model: GLM-5.1

YOMXXX:
Closes #4184.

## Problem
Providers loaded via \`@ai-sdk/openai-compatible\` (GLM, custom local LLMs, etc.) need a \`limit\` field declared **inline** in \`opencode.json\` because their context window isn't published to the \`models.dev\` cache. When it's missing, OpenCode's overflow detector reads \`context === 0\` and **auto-compaction never triggers** — sessions silently exhaust the context window and the agent eventually stops emitting tool calls.

This is exactly the kind of failure where the user has no idea why their long session got stuck. The diagnosis path involves cross-referencing \`opencode.json\` against \`packages/opencode/src/session/overflow.ts\` from upstream OpenCode.

## Scope note
omo **does not write provider/model configuration to opencode.json** — users manage that file themselves. So the fix is not \"populate limit automatically\" (that would silently override user intent); it's **detection + concrete fix guidance** in doctor.

## What's added
New doctor check \`src/cli/doctor/checks/custom-provider-limits.ts\`:

- Scans both \`~/.config/opencode/opencode.json\` (or \`.jsonc\`) and \`<cwd>/.opencode/opencode.json\`.
- For each provider whose \`npm\` matches \`@ai-sdk/openai-compatible\` (or \`@ai-sdk/openai\`, or a versioned variant like \`@ai-sdk/openai-compatible@1.2.3\`), every model whose \`limit.context\` is missing or \`<= 0\` produces a \`warning\` \`DoctorIssue\`:

\`\`\`
Custom provider model is missing limit.context
  glm/glm-5.1 (npm: @ai-sdk/openai-compatible) has no limit.context. OpenCode's
  auto-compaction relies on this to detect overflow; without it, sessions
  silently exhaust the context window and the agent stops emitting tool calls.

  Fix: Add limit.context (and ideally limit.output) to
  provider.glm.models.glm-5.1 in ~/.config/opencode/opencode.json. Example:
  \"limit\": { \"context\": 128000, \"output\": 8192 }
\`\`\`

Wired into \`checkModels\` so it surfaces alongside existing model-resolution warnings (same check, no new top-level entry in the doctor table).

## Test plan
- [x] 7 new unit tests cover: no config, single missing-limit, with-limit (no warning), non-compat provider ignored, multiple models across multiple providers, project-level config scanning, and \`limit.context: 0\` treated as missing (matches OpenCode's overflow detector behavior).
- [x] \`bun test src/cli/doctor/checks/custom-provider-limits.test.ts src/cli/doctor/checks/model-resolution.test.ts\` → 24 pass / 0 fail.

---
## Summary by cubic
Warns when custom provider models using `@ai-sdk/openai-compatible` or `@ai-sdk/openai` lack a positive `limit.context` in `opencode.json` to prevent context overflow and stuck tool calls. Handles malformed configs and surfaces in the doctor’s model check with clear fix guidance (addresses #4184).

- **New Features**
  - Adds `custom-provider-limits` doctor check that scans user and project `opencode.json`/`opencode.jsonc`, flags each matching model without `limit.context` (including versioned `npm` values), and treats `limit.context: 0` as missing.
  - Integrated into `checkModels` so warnings appear with existing model-resolution results.

<sup>Written for commit e23bef08b052a0623788abd069c14336798cfd42. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4276?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/cli/doctor/checks/custom-provider-limits.test.ts
+++ b/src/cli/doctor/checks/custom-provider-limits.test.ts
@@ -155,4 +155,91 @@ describe("collectCustomProviderLimitIssues (issue #4184)", () => {
     const issues = collectCustomProviderLimitIssues()
     expect(issues).toHaveLength(1)
   })
+
+  // hardening from cubic-dev-ai review on #4276 ↓
+
+  test("does not crash when a provider entry is null (malformed config)", () => {
+    writeFileSync(
+      join(userConfigDir, "opencode.json"),
+      JSON.stringify({
+        provider: {
+          "broken": null,
+          glm: {
+            npm: "@ai-sdk/openai-compatible",
+            models: { "glm-5.1": {} },
+          },
+        },
+      })
+    )
+
+    const issues = collectCustomProviderLimitIssues()
+    expect(issues).toHaveLength(1)
+    expect(issues[0]?.description).toContain("glm/glm-5.1")
+  })
+
+  test("does not crash when a model entry is null or a primitive", () => {
+    writeFileSync(
+      join(userConfigDir, "opencode.json"),
+      JSON.stringify({
+        provider: {
+          glm: {
+            npm: "@ai-sdk/openai-compatible",
+            models: {
+              "glm-broken": null,
+              "glm-also-broken": "not-an-object",
+              "glm-good": { limit: { context: 128000 } },
+            },
+          },
+        },
+      })
+    )
+
+    const issues = collectCustomProviderLimitIssues()
+    expect(issues).toHaveLength(2)
+    const descriptions = issues.map((i) => i.description).join("\n")
+    expect(descriptions).toContain("glm/glm-broken")
+    expect(descriptions).toContain("glm/glm-also-broken")
+    expect(descriptions).not.toContain("glm/glm-good")
+  })
+
+  test("does not crash when the top-level provider field is malformed", () => {
+    writeFileSync(
+      join(userConfigDir, "opencode.json"),
+      JSON.stringify({ provider: "not-an-object" })
+    )
+
+    const issues = collectCustomProviderLimitIssues()
+    expect(issues).toEqual([])
+  })
+
+  test("does not crash when provider.models is malformed", () => {
+    writeFileSync(
+      join(userConfigDir, "opencode.json"),
+      JSON.stringify({
+        provider: {
+          glm: {
+            npm: "@ai-sdk/openai-compatible",
+            models: "garbage",
+          },
+        },
+      })
+    )
+
+    const issues = collectCustomProviderLimitIssues()
+    expect(issues).toEqual([])
+  })
+
+  test("does not crash when provider.npm is null", () => {
+    writeFileSync(
+      join(userConfigDir, "opencode.json"),
+      JSON.stringify({
+        provider: {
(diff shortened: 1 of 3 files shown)


YOMXXX:

YOMXXX:
Addressed both issues from the cubic-dev-ai review in e23bef08.

The check now uses an `isPlainObject` typeguard at every untrusted property access, so any shape of malformed config is tolerated without crashing the Models doctor check:

- Top-level `provider` malformed → no issues, return early.
- A single provider entry `null` / string → skip and keep iterating siblings.
- `provider.npm` not a non-empty string → skip the provider.
- `provider.models` malformed → skip without crashing.
- A model entry `null` / primitive → treated as "missing limit" so the user still gets actionable output for the broken entry.

Five new tests cover each malformed shape (null provider, null/primitive model, malformed top-level provider, malformed models, null npm). 29/29 doctor + custom-limits tests pass.

code-yeongyu:
[sisyphus-bot] Thank you for this diagnostic improvement! Issue #4184 was closed by PR #5268 with a docs workaround noting the custom provider limit caveat. The doctor command has since evolved (--verbose flag, Windows shim fixes). While the proactive warning this PR adds is still a nice idea, we're closing as stale given the issue closure and code drift. If you'd like to rebase against current dev, a fresh PR would be welcome.

Issue #4178, opened 2026-05-19
[sisyphus-bot] Bug: disabled_mcps not re-applied on /new session — stale MCPs from .claude.json leak through


xAsh-Ai:
## Bug Description

When using `/new` to create a new session within the same opencode process, `disabled_mcps` from `oh-my-openagent.json` are **not re-applied** during MCP config resolution. This causes MCPs that were disabled (e.g., duplicates from `.claude.json`) to reappear as active and often timeout.

## Reproduction

1. Add entries to `disabled_mcps` in `~/.config/opencode/oh-my-openagent.json`:
   ```json
   {
     "disabled_mcps": [
       "auggie", "clear-thought-mcp-server", "composio",
       "taskmaster-ai", "supabase", "byterover-mcp", "sentry"
     ]
   }
   ```

2. **Restart opencode** (full process restart) → MCPs correctly filtered to 15 active. ✅
3. Use **`/new`** to create a new session → disabled MCPs reappear (22 active), many timeout at 30s. ❌

## Expected Behavior

`disabled_mcps` should be applied consistently on every session creation, including `/new`.

## Actual Behavior

`disabled_mcps` is only applied during initial plugin load. On `/new`, `loadMcpConfigs()` re-reads `.claude.json` and merges MCPs without re-applying the `disabled_mcps` filter.

## Root Cause (from source analysis)

In `oh-my-openagent@4.2.2`, the MCP resolution flow:

1. `createBuiltinMcps(disabledMcps, config)` — creates plugin builtins
2. `loadMcpConfigs(disabledMcps)` — loads from `.claude.json` etc.
3. `applyMcpConfig(params)` — merges all sources, then applies `disabled_mcps`

Step 3 only runs once during plugin initialization. When `/new` triggers a new session, the MCP list is re-resolved from raw sources **without re-running the full `applyMcpConfig` pipeline** that includes the `delete merged[name]` for disabled entries.

## Impact

- ~7 extra MCPs start simultaneously → resource contention → 30s timeouts
- User disables MCPs but they come back in new sessions
- Workaround: must fully restart opencode instead of using `/new`

## Environment

- opencode v1.15.5
- oh-my-openagent v4.2.2 (darwin-arm64)
- Node v22.22.0
- macOS

## Suggested Fix

Ensure `disabled_mcps` filtering runs on **every** MCP config resolution, not just initial plugin load.

code-yeongyu:
[...]

PR #4277, opened 2026-05-22
test(mcp-config-handler): lock disabled_mcps idempotency against #4178 regression

YOMXXX:
Refs #4178.

## Context
#4178 reports that \`disabled_mcps\` stops filtering after \`/new\` — disabled servers reappear and start timing out.

After tracing the code, omo's \`applyMcpConfig\` is **already idempotent**:
- Each invocation re-reads \`pluginConfig.disabled_mcps\`.
- \`createBuiltinMcps\` and \`loadMcpConfigs\` are passed \`disabledMcps\` and self-filter (defense in depth).
- The merged map gets a final \`delete merged[name]\` for every disabled entry.

So as long as the \`config\` hook gets re-invoked on \`/new\`, the omo side will produce a filtered set every time. **Whether OpenCode re-invokes the hook on \`/new\` is an upstream OpenCode behavior** — omo can't force it from a plugin event handler (the event API doesn't give us a writable \`config\` reference).

What omo **can** do is harden the contract so a future refactor of \`applyMcpConfig\` can't quietly introduce state that depends on prior invocations.

## What's added
Two regression tests in \`mcp-config-handler.test.ts\`:

1. **\`re-applying after the session config gets reset still filters disabled_mcps\`** — simulates OpenCode handing the same \`config\` back with \`config.mcp\` reset to the raw user map (re-introducing the disabled entries). After two back-to-back invocations, both passes must produce the filtered set.

2. **\`idempotent across N invocations with disabled_mcps\`** — three invocations in a row, each time the runtime hands us \`{ auggie: {...enabled: true...} }\` again. Each pass must end with \`auggie\` gone.

## Test plan
- [x] \`bun test src/plugin-handlers/mcp-config-handler.test.ts\` → 8 pass / 0 fail.
- [x] No production code changes — pure regression coverage for the existing idempotent behavior.

## Caveat
If the real bug is OpenCode not re-invoking the \`config\` hook on \`/new\`, this PR alone won't change user-visible behavior. The fix for that side belongs upstream in OpenCode. But this PR ensures the omo side is bulletproof against the regression mode it could theoretically cause.

---
## Summary by cubic
Adds two regression tests for `applyMcpConfig` ensuring `disabled_mcps` are always filtered, even after `/new` session resets (#4178). No production code changes.

<sup>Written for commit 25c58a1c8fae83376a05511d19de7136cd55c9d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/plugin-handlers/mcp-config-handler.test.ts
+++ b/packages/omo-opencode/src/plugin-handlers/mcp-config-handler.test.ts
@@ -183,4 +183,79 @@ describe("applyMcpConfig", () => {
     expect(createBuiltinMcpsSpy).toHaveBeenCalledWith([], pluginConfig, { cwd: TEST_CTX.directory })
   })
 
+  // regression: issue #4178 — `/new` triggers OpenCode to re-evaluate `config`,
+  // which re-invokes this handler. When that happens, `params.config.mcp` may have
+  // been reset to the raw merged set (re-introducing disabled entries). The
+  // handler must still produce a filtered `config.mcp` regardless of input shape.
+  test("re-applying after the session config gets reset still filters disabled_mcps", async () => {
+    //#given: first invocation — plugin init
+    createBuiltinMcpsSpy.mockReturnValue({
+      websearch: { type: "remote", url: "https://mcp.exa.ai/mcp", enabled: true },
+    })
+    loadMcpConfigsSpy.mockResolvedValue({
+      servers: {
+        firecrawl: { type: "remote", url: "https://firecrawl.example.com", enabled: true },
+        exa: { type: "remote", url: "https://exa.example.com", enabled: true },
+      },
+    })
+    const userMcpRaw = {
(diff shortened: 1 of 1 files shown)


YOMXXX:

code-yeongyu:
[sisyphus-bot] Thank you for this regression test! This PR is part of the 12-PR roadmap from discussion #4293, which hasn't received maintainer engagement. The MCP config handler code has evolved on dev since this was opened. Closing as stale — if the idempotency concern still applies to current dev, a fresh test PR would be appreciated.

PR #4280, opened 2026-05-22
fix(background-agent): reuse manager across duplicate plugin initialization

### LINKED ISSUE #4095: background_output still returns 'Task not found' after PR #4015 (macOS + Windows)
PR #4015 added a retry mechanism for transient `bg_...` lookups, which is appreciated. However, `background_output` still consistently returns 'Task not found' on both macOS and Windows Desktop.

## Reproduction

Create a background task via `task(run_in_background=true)`. After completion, call `background_output(task_id)` — it returns 'Task not found' immediately, not just transiently.

## Diagnosis

The retry delay doesn't help because the issue is architectural — Desktop's built-in `_taskStore` in `app.asar` is separate from OMO's `BackgroundManager.tasks`. The task was never in `_taskStore` to begin with, so retrying the same lookup produces the same result.

## Workaround

Using `session_read(session_id=...)` as fallback works reliably.

## Environment

- macOS + Windows
- OpenCode Desktop v1.14.x
- OMO v4.1.2

SpencerJung:
## Summary

Fixes #4095.

OpenCode Desktop can initialize the OMO plugin more than once for the same project. Before this change, each initialization created a separate BackgroundManager instance, so a task launched through one instance could be invisible to background_output routed through another instance.

This PR keeps one BackgroundManager singleton per project directory from createManagers(), so duplicate plugin initialization for the same project shares the same task store, polling state, and background_output lookup path.

## What changed

- Add a project-directory keyed BackgroundManager singleton in src/create-managers.ts.
- Reuse the existing manager when createManagers() is called again for the same directory.
- Add a regression test that initializes managers twice for one project and asserts the BackgroundManager constructor runs once and both results share the same instance.

## Test plan

- [x] LSP diagnostics on src/create-managers.ts and src/create-managers.test.ts: 0 errors.
- [x] bun test src/create-managers.test.ts: 6 pass / 0 fail.
- [x] bun run typecheck: pass.
- [x] bun run build: pass.
- [x] git diff --check: pass.

## PR Checklist

- [x] bun run typecheck passes
- [x] bun run build succeeds
- [x] Tested locally through focused regression coverage
- [x] No version changes in package.json

---
## Summary by cubic
Make `BackgroundManager` a per-project singleton and refresh its runtime bindings on reuse. This prevents invisible tasks and keeps callbacks, tmux routing, and polling in sync.

- **Bug Fixes**
  - Reuse existing manager in `createManagers()` via a directory-keyed map; on reuse call `updateRuntimeBindings()` (client, tmux, callbacks) and log reuse.
  - Added `BackgroundManager.updateRuntimeBindings()` and `ParentWakeNotifier.updateRuntimeDeps()` to safely swap runtime deps.
  - Tests use `resetBackgroundManagerSingletonsForTesting()` and verify reuse, single constructor call, and that the latest tmux manager/callbacks are applied.

<sup>Written for commit 0b971f715bd5a795952c5854c1e41ca9cab03a9e. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4280?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/create-managers.test.ts
+++ b/src/create-managers.test.ts
@@ -4,7 +4,7 @@ import { afterEach, beforeEach, describe, expect, it, mock, spyOn } from "bun:te
 import type { PluginInput } from "@opencode-ai/plugin"
 
 import { OhMyOpenCodeConfigSchema } from "./config/schema/oh-my-opencode-config"
-import { createManagers } from "./create-managers"
+import { createManagers, resetBackgroundManagerSingletonsForTesting } from "./create-managers"
 import * as openclawRuntimeDispatch from "./openclaw/runtime-dispatch"
 import { createModelCacheState } from "./plugin-state"
 
@@ -18,6 +18,7 @@ const markServerRunningInProcess = mock(() => {})
 let backgroundManagerOptions: {
   onSubagentSessionCreated?: (event: { sessionID: string; parentID: string; title: string }) => Promise<void>
 } | null = null
+let backgroundManagerConstructorCalls = 0
 const trackedPaneBySession = new Map<string, string>()
 const registeredCleanupManagers: CleanupRegistration[] = []
 const cleanupSessionTeamRunsMock = mock(async () => ({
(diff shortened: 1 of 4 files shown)


code-yeongyu:
[sisyphus-bot] Thank you for this fix! Issue #4095 is still open, but the background-agent manager code has been substantially refactored on dev (background wake injections, idle task status, injection-driven waiting). This PR's target code has moved significantly. Closing as stale — if the duplicate plugin init issue persists, a fresh PR against current dev would be welcome.

PR #4281, opened 2026-05-22
fix(todo-continuation): preserve completed-todos terminal guard

### LINKED ISSUE #4111: todo-continuation-enforcer: silent stop when all todos completed — no explicit done signal
## Description

When all todos are marked `completed` in a session, the `todo-continuation-enforcer` hook still fires the full ultrawork system prompt (~2K tokens) into the context on each continuation trigger. The LLM then reads the compressed context, finds zero pending todos, tries reading files to find work, and eventually goes silent.

From the user's perspective, this looks like the session **"froze" or "crashed"** — there is no explicit "all done" signal.

## Where

Hook: `todo-continuation-enforcer` (Continuation category, lifecycle events: `tool.execute.before`, `event`)

## Reproduce

1. Start session with multiple `todowrite` items  
2. Complete every todo (all marked `completed`)  
3. Switch back to the session → continuation hook triggers  
4. Observed: session fires up, reads a few files, then goes **silent** in a few seconds  
5. User assumes it crashed and opens a new session

## Expected Behavior

**Option A (minimal):** When all todos are `completed`, `todo-continuation-enforcer` should detect this state and respond with a clear message:
```
All 9 tasks complete. What would you like to do next?
```
It should **skip** injecting the full ultrawork prompt when there is no pending work.

**Option B (ideal):** Three-way branching:
| State | Action |
|-------|--------|
| Has `pending` / `in_progress` todos | Inject continuation prompt normally |
| All todos `completed` | Respond with completion summary, do NOT inject ultrawork |
| No todos ever created | Skip (new session, nothing to continue) |

## Impact

- Wastes LLM tokens: ~2K prompt + file reads per trigger
- Confuses users into thinking the session crashed
- Breaks workflow: user must open a new session rather than continuing naturally after completion

## Context

- Observed on opencode CLI with oh-my-openagent installed
- Session had 9 completed todos, compaction had already occurred  
- Related: `/ulw-loop` and `/ralph-loop` both use the same continuation mechanism

SpencerJung:
## Summary

Fixes #4111.

When every todo was completed, handleSessionIdle set allTodosCompletedAt and then immediately called resetContinuationProgress(). The real SessionStateStore reset clears allTodosCompletedAt, so the terminal guard did not survive. Later session.idle events could re-enter the expensive todo-continuation path instead of treating the session as terminally complete.

This PR keeps the terminal completed-todos guard after resetting continuation progress, so repeated idle events do not refetch messages/todos or inject the full continuation prompt once all todos are done.

## What changed

- Reset continuation progress before setting allTodosCompletedAt.
- Add a regression test using the real SessionStateStore.
- Assert the second idle event exits through the completed-todos guard without fetching messages or todos again.

## Test plan

- [x] LSP diagnostics on idle-event.ts and idle-event.test.ts: 0 errors.
- [x] bun test src/hooks/todo-continuation-enforcer/idle-event.test.ts: 4 pass / 0 fail.
- [x] bun run typecheck: pass.
- [x] bun run build: pass.
- [x] git diff --check: pass.

## PR Checklist

- [x] bun run typecheck passes
- [x] bun run build succeeds
- [x] Tested locally through focused regression coverage
- [x] No version changes in package.json

---
## Summary by cubic
Preserves the completed‑todos terminal guard so sessions don’t re-enter the continuation flow after all todos are done. Fixes #4111 and avoids extra message/todo fetches on repeated idle events.

- **Bug Fixes**
  - Reset continuation progress before setting `allTodosCompletedAt`.
  - Added a regression test with the real `createSessionStateStore` to ensure the guard persists.
  - Verified the second idle exits via the guard (no refetches, no continuation prompt).

<sup>Written for commit 6f883d6cef9779ae1183438b1772397d7acf55db. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4281?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/todo-continuation-enforcer/idle-event.test.ts
+++ b/src/hooks/todo-continuation-enforcer/idle-event.test.ts
@@ -3,7 +3,7 @@
 import { describe, expect, it } from "bun:test"
 
 import { handleSessionIdle } from "./idle-event"
-import type { SessionStateStore } from "./session-state"
+import { createSessionStateStore, type SessionStateStore } from "./session-state"
 import type { ContinuationProgressUpdate, SessionState } from "./types"
 
 function createStateStore(): {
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot] Thank you for this fix! Issue #4111 is still open, but the todo-continuation system has been extensively reworked on dev since this was opened — the continuity component was reverted in PR #6422 and multiple stop-flag fixes have landed. This PR's target code no longer exists in its original form. Closing as stale — if the underlying issue persists after the rework, a fresh PR against current dev would be appreciated.

PR #4326, opened 2026-05-23
fix: add type guards in fallback-state to prevent .toLowerCase() crash on object-shaped model inputs

herjarsa:
Closing - superseded by broader fix in #4343


PeterPonyu:
Superseded by the broader fix in #4343 (runtime-fallback normalization + full disabled provider enforcement + runtime task() model/variant API).

PR #4328, opened 2026-05-23
fix: propagate agent variant from config to chat.params handler

### LINKED ISSUE #4125: Agent variant from config not propagated to chat.params message
**Describe the bug**
When an agent's `variant` (e.g., `"max"`) is explicitly configured in `oh-my-opencode.json`, it is not injected into the `chat.params` message. As a result, the model defaults to a lower or missing variant, ignoring the user's explicit reasoning effort setting. 

This issue was discovered when configuring DeepSeek V4 with the `"max"` variant, which failed to run at the maximum reasoning effort because the configuration wasn't properly passed down. Additionally, the `max` variant was not registered for the `deepseek` family in the heuristics registry.

**To Reproduce**
1. Configure an agent with a specific variant in `oh-my-opencode.json` (e.g., `variant: "max"` for a deepseek model).
2. Start a session with the agent.
3. Observe that the actual API call parameters omit the configured variant.

**Expected behavior**
The `variant` defined in `pluginConfig.agents[agentName].variant` should be properly injected into `message.variant` before the `chat.params` handler processes it, ensuring the requested reasoning effort is applied.

herjarsa:
## Description

Fixes #4125

The `applyAgentVariant` function was defined in `shared/agent-variant.ts` but never called from any production code path. This meant that when a user configured a variant for an agent (e.g., `agents.sisyphus.variant: "ultrawork"`), it was silently ignored and never propagated to the model compatibility resolver.

## Root Cause

`applyAgentVariant` was only used in test files. No hook handler was calling it, so the variant configured in `config.agents.<name>.variant` never reached `resolveCompatibleModelSettings`.

## Changes

- Modified `createChatParamsHandler` to accept the plugin config
- Added variant resolution in the handler: if the agent has a configured variant and the message doesn't already have one, it is set before model compatibility lookup
- Passed `pluginConfig` from `plugin-interface.ts` wiring

## Testing

\`\`\`
$ bun test src/plugin/chat-params.test.ts
 7 pass
 0 fail

$ bun test src/shared/agent-variant.test.ts
 15 pass
 0 fail
\`\`\`

All tests pass. Typecheck: zero errors.

---
## Summary by cubic
Propagates agent variants from config to `chat.params` so model compatibility picks correct settings, adds a `node_modules` walk-up fallback for version detection in sandboxed bundles, and prevents duplicate background-agent streams during the hold. Also updates skills docs to add an OpenCode session debugging workflow and require all PR gates (CI, review-work, Cubic, GPT-5.2 xhigh) before merge.

- **Bug Fixes**
  - Passes `pluginConfig` into `createChatParamsHandler` and uses `resolveAgentVariant` to set `message.variant` before `resolveCompatibleModelSettings`.
  - Extends `getCachedVersion()` to search ancestor `node_modules/<accepted-name>/package.json`.
  - Drops redundant parent wake re-dispatch during the hold to avoid double streams.

<sup>Written for commit bc2b07c306a605e3f0dd1f6e6baf181e72854602. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4328?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.agents/skills/opencode-session-debugging/SKILL.md
+++ b/.agents/skills/opencode-session-debugging/SKILL.md
@@ -156,7 +156,7 @@ If the evidence points to a code defect:
 3. Make the smallest code change that turns the test green.
 4. Run focused tests, adjacent tests, typecheck, build, and the full suite when risk warrants it.
 5. Open a PR to `dev`.
-6. Do not merge until CI, review-work, Cubic, and GPT-5.2 xhigh or Momus review all pass.
+6. Do not merge until CI, review-work, Cubic, and GPT-5.2 xhigh PR review all pass. If a `.omo/plans/*.md` plan is part of the workflow, Momus may review that plan and must return `[OKAY]`; do not count Momus as a PR diff reviewer.
 7. Remove the worktree only after the merge commit succeeds.
 
 ## Report Template
--- a/.agents/skills/opencode-session-debugging/SKILL.md
+++ b/.agents/skills/opencode-session-debugging/SKILL.md
@@ -14,6 +14,7 @@ Use this skill to investigate a broken OpenCode session from evidence first, the
 3. Build a timestamped timeline from transcripts, OpenCode logs, plugin logs, and background-task state.
 4. Use git history and blame for the failing subsystem before changing code.
 5. If a fix is needed, switch to `work-with-pr`: create a new worktree, write the failing test first, open a PR, and do not merge until all required gates pass.
+6. Treat raw `session.prompt` and `session.promptAsync` paths as suspect until the shared prompt gate has been audited.
 
 ## Phase 0: Capture Inputs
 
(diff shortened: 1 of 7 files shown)


code-yeongyu (changes requested):
I can’t approve this PR while it introduces forbidden deprecated model references.

The PR diff adds multiple `GPT-5.2 xhigh PR review` / `GPT-5.2 reviewer` requirements in the work-with-pr skill docs. Repository policy now forbids GPT-5.2/GPT-5.3 usage. Please migrate every new GPT-5.2/GPT-5.3 reference in this PR to GPT-5.5, then rerun the relevant documentation/skill checks before requesting review again.

herjarsa:
Cerrado — mismo fix que 4329.

PR #4336, opened 2026-05-23
fix(comment-checker): fall back to PATH lookup so PATH-installed binary is found (#3315)

### LINKED ISSUE #3315: [Bug] comment-checker hook does not trigger
## Issue

On Windows, the comment-checker hook does not trigger even though the binary is working.

## Environment
- OS: Windows 11 x64
- oh-my-opencode: 1.4.3
- oh-my-openagent: latest
- @code-yeongyu/comment-checker: installed via npm

## Steps to Reproduce
1. Install oh-my-openagent plugin
2. Install WinLibs (to provide MinGW DLLs): `winget install BrechtSanders.WinLibs.POSIX.UCRT`
3. Install comment-checker: `npm install -g @code-yeongyu/comment-checker`
4. Add binary to PATH: copy comment-checker.exe to a location in PATH
5. Run `npx oh-my-opencode doctor`
   - Result: System OK (binary detected)
6. Write code with comments using Write/Edit tools
   - Expected: Hook warns about comments
   - Actual: No warning, hook does not trigger

## What Works
- Binary executes correctly when run manually
- Doctor shows System OK (detects binary)
- Custom prompt in `comment_checker` config is loaded

## Analysis
The hook may be calling comment-checker but:
- Wrong working directory
- Missing stdin input format (hook expects JSON)
- Windows-specific path issue in hook executor

## Workaround
Using agent prompt constraint (matches comment-checker philosophy):
```json
"agents": {
  "sisyphus": {
    "prompt_append": "Avoid unnecessary comments. Don't write: // TODO (will never do), // adds one to one, // obvious comments. But you CAN use: # given/# when/# then (BDD), # noqa, // @ts-ignore (lint), shebangs."
  }
}
```

## Expected Fix
Hook should trigger on Windows when code with comments is written.

YOMXXX:
## Summary

Closes #3315. The comment-checker hook silently no-op'd on Windows even though `bunx oh-my-opencode doctor` reported "System OK" — because doctor and the runtime resolver disagreed on where to look for the binary.

## Root cause

`resolveCommentCheckerBinary` (in \`packages/comment-checker-core/src/runner.ts\`) checks two locations in order:

1. The cached download path (\`getCachedBinaryPath()\` from OMO's downloader).
2. The package-local \`bin/<binary-name>\` directory under \`@code-yeongyu/comment-checker\`'s install location, resolved via \`createRequire(import.meta.url).resolve(...)\`.

That's it. Users who installed via \`npm install -g @code-yeongyu/comment-checker\` end up with the binary on PATH but **not** in either of those two locations — so the resolver returned null. Meanwhile \`src/cli/doctor/checks/dependencies.ts:15\` uses \`Bun.which(binary)\` and happily found the same binary, leading to the contradiction the user reported:

> Result: System OK (binary detected) [...] Hook does not trigger

Maintainer already confirmed this resolver/doctor mismatch is OMO-owned.

## Fix

`src/hooks/comment-checker/cli.ts` — `findCommentCheckerPathSync` gains a PATH lookup fallback (defaulting to \`Bun.which\`) after the existing cached/package-local checks fail:

```ts
function findCommentCheckerPathSync(which: WhichFn = Bun.which): string | null {
  const resolvedPath = resolveCommentCheckerBinary({...})
  if (resolvedPath !== null) return resolvedPath

  // Fall back to PATH lookup — mirrors doctor's dependency check.
  try {
    const pathBinary = which(getBinaryName())
    if (pathBinary && existsSync(pathBinary)) return pathBinary
  } catch {
    // Bun.which can throw on some platforms; treat as 'not found'.
  }
  return null
}
```

Order is preserved (cached → package-local → PATH) so the existing happy path (`npm install` putting the binary inside the package) keeps winning over PATH binaries on machines where both exist.

## Tests

\`src/hooks/comment-checker/cli.test.ts\` gains a new \"PATH lookup fallback (#3315)\" suite (4 tests). All routed through a new \`__resolveCommentCheckerBinaryForTesting(which)\` export so the resolver order can be asserted deterministically without monkey-patching \`Bun.which\` globally:

- Returns the PATH binary when the cached and package-local paths are absent (the actual repro from the issue).
- Ignores a PATH binary that doesn't actually exist on disk (so we never advertise a bogus path that would crash on \`spawn\`).
- Treats a throwing \`which\` as \"not found\" (some platforms throw while traversing PATH).
- Returns null when \`which\` returns \`null\`/\`undefined\` (the typical not-found signal from \`Bun.which\`).

\`bun test src/hooks/comment-checker/\` — 24 pass / 0 fail.

## Closes

- #3315

---
## Summary by cubic
Fixes the comment-checker hook not running when the CLI is installed globally by falling back to a PATH lookup, matching the doctor check. Addresses #3315 and prevents silent no-ops on Windows.

- **Bug Fixes**
  - Added PATH lookup fallback using `Bun.which(getBinaryName())` after cached and package-local checks; returns only if the file exists and preserves the original resolution order. This matches the doctor's dependency check.
  - Added tests for the fallback behavior and introduced `__resolveCommentCheckerBinaryForTesting(which)` to assert lookup order; handles throwing/null `which` safely.

<sup>Written for commit 77682055fa2e0865a23ca48608745348b6ea46c9. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4336?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/comment-checker/cli.test.ts
+++ b/src/hooks/comment-checker/cli.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect, mock, afterAll } from "bun:test"
-import { chmodSync, mkdtempSync, writeFileSync } from "node:fs"
+import { chmodSync, mkdtempSync, rmSync, writeFileSync } from "node:fs"
 import { join } from "node:path"
 import { tmpdir } from "node:os"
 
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot] Thank you for this fix! Issue #3315 has been resolved by PR #4685 (merged 2026-06-02), which implements the same PATH-fallback for the comment-checker binary. This PR is now superseded and safe to close. We appreciate the contribution — feel free to re-engage if you spot related issues.

PR #4031, opened 2026-05-15
feat(config): add disabled_providers schema + helper

PeterPonyu:
## Summary

- Adds an opt-in `disabled_providers` config field that lets users hide providers/models from resolution.
- Extracts the `shared/disabled-providers` helper (`isProviderDisabled`, `filterDisabledProviderModels`, `applyDisabledProviders`) used by the plugin-config layer.
- Narrow extraction from the closed PR #3871 — schema + helper + config-layer wiring only.

## Changes

- `src/config/schema/oh-my-opencode-config.ts` — adds `disabled_providers` schema field.
- `src/shared/disabled-providers.ts` — new helper (`isProviderDisabled`, `filterDisabledProviderModels`, `applyDisabledProviders`).
- `src/shared/disabled-providers.test.ts` — helper unit tests.
- `src/plugin-config.ts` — wires `applyDisabledProviders` into the config layer.
- `src/plugin-config.test.ts` — config-layer tests.

The routing changes in `team-mode/team-runtime`, `team-mode/tools/lifecycle`, and
`plugin/tool-registry` are NOT included here — they're entangled with the
`MemberSelectionMode` and `ShutdownActor` refactors from the closed PR #3871 and
need their own independent landing first. This PR isolates only the cleanly
independent slice.

## Testing

```bash
bun run typecheck                                  # clean
bun test src/shared/disabled-providers.test.ts     # 13 pass, 0 fail
bun test src/plugin-config.test.ts                 # 39 pass, 0 fail
```

## Related Issues

Companion to #4024 (live-tail tailer). Both PRs isolate one piece of intent from
the closed PR #3871, which had drifted to 89 commits of mixed-scope work. The
team-mode routing piece will follow in a separate PR once its prerequisites
(`MemberSelectionMode`, `ShutdownActor` refactors) land.

---
## Summary by cubic
Adds `disabled_providers` to the config so users can exclude providers from model resolution. On load, we filter fallback chains and replace disabled primaries with the first allowed fallback.

- **New Features**
  - Schema: new `disabled_providers` array, merged with case-insensitive dedupe.
  - Helper: `isProviderDisabled`, `filterDisabledProviderModels`, `applyDisabledProviders`, called by `loadPluginConfig`.
  - Behavior: filters agents/categories fallback chains; swaps a disabled primary with the first allowed fallback; emits a config-load error if none exists; no-op when unset.
  - Tests: new unit tests for helpers and config loader.

[...]

PR #4341, opened 2026-05-23
fix(runtime-policy): close disabled provider fallback paths

PeterPonyu:
## Summary

- Thread `disabled_providers` into delegate-task and team-mode task runtime setup.
- Reject discovered subagent default models that resolve to disabled providers instead of silently falling back to OpenCode defaults.
- Filter disabled providers out of proactive model-fallback session and hardcoded fallback chains.

Follow-up to #4031.

## Why

The foundational runtime policy PR blocks configured category/model-selection violations. This PR completes the remaining runtime escape hatches: plugin wiring, discovered subagent defaults, and proactive model-fallback chains.

## Dependency

This is intended to stack after the delegate-task runtime model/variant PR because the subagent resolver blocker fix depends on the runtime model/variant precedence path.

## Testing

- `bun test src/tools/delegate-task/zauc-mocks-subagent-resolver/subagent-resolver.test.ts`
- `bun test src/hooks/model-fallback/hook.test.ts`
- `bun test src/shared/disabled-providers.test.ts`
- `bun run typecheck`
- `bun run build`
- Manual driver for disabled discovered subagent defaults and hardcoded model-fallback filtering

---
## Summary by cubic
Closes remaining runtime escape hatches by enforcing disabled provider policy across fallbacks and delegate-task resolution, and normalizes model parsing (including variants) so fallback chains advance correctly.

- **Bug Fixes**
  - Block disabled providers in all paths: delegate-task selection, discovered subagent defaults, and proactive/hardcoded fallback chains.
  - Normalize and compare runtime models across event shapes (model object or top-level fields), preserve `variant`, and advance fallbacks only when the errored model matches.
  - Update watchdog, `session.error`, `message.updated`, and `session.status` handlers to use the normalized model, skip equivalent fallback variants, and clear pending state reliably.
  - Trim whitespace when parsing providers and comparing against `disabled_providers` to avoid false negatives.

<sup>Written for commit cd2d43f5d476c8c2581ba67450ec80bcc741abaf. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4341?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/runtime-fallback/event-handler.test.ts
+++ b/src/hooks/runtime-fallback/event-handler.test.ts
@@ -1,8 +1,18 @@
+/// <reference types="bun-types" />
+
 import { describe, expect, it } from "bun:test"
 import type { HookDeps, RuntimeFallbackPluginInput } from "./types"
 import type { AutoRetryHelpers } from "./auto-retry"
 import { createFallbackState } from "./fallback-state"
 import { createEventHandler } from "./event-handler"
+
+const testPluginConfig = {
+  git_master: {
+    commit_footer: true,
+    include_co_authored_by: true,
+    git_env_prefix: "GIT_MASTER=1",
+  },
+}
 
 function createContext(): RuntimeFallbackPluginInput {
   return {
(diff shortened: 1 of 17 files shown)


PeterPonyu:
Closing incorrect scope. Will replace with properly scoped PRs.

PR #4349, opened 2026-05-23
fix: align auto-updater install path with loaded sandbox (#4318)

### LINKED ISSUE #4318: [Bug]: Auto-update installs to wrong path — version never actually switches, infinite loop on restart

## Symptoms

First notice:

```
OhMyOpenCode 4.1.2
```

Then immediately:

```
OhMyOpenCode Updated!
V4.1.2 → v4.3.1
Restart OpenCode to apply.
```

**4.1.2**

## Steps to reproduce

## Root cause analysis

```
Path created by Npm.add():
```

```
```

| File | Line | Note |
|------|------|------|

**The infinite loop:**

```
```

## Environment

- OS: Windows
- OpenCode: 1.15.10

## Expected behavior

After quitting OpenCode, run:

```
```

```
OhMyOpenCode 4.1.2
```

```
```

still returns `4.3.1`.

## Suggested fixes

Yeachan-Heo:
Closes #4318

## Root cause

OpenCode's `Npm.add()` installs each plugin into a per-spec sandbox at
`<CACHE>/packages/<sanitized-spec>/node_modules/<pkg>/` (e.g.
`oh-my-openagent@latest/node_modules/oh-my-openagent/`). The
auto-updater, however, wrote `<CACHE>/packages/package.json` and ran
`bun install` against `<CACHE>/packages/` — paths that never intersect
with the sandbox. The sandbox kept loading the old version while the
flat workspace advanced, so `getCachedVersion()` (which walks up from
`import.meta.url`) always reported the old version and the registry
check always saw an update, producing an infinite "Updated! Restart
OpenCode" toast loop on every restart.

## Fix

Add `getLoadedSandboxWorkspace()` in
`src/hooks/auto-update-checker/checker/sandbox-workspace.ts`. It walks
up from `import.meta.url` (the same anchor `getCachedVersion()` uses),
verifies the resolved package.json lives at
`<workspace>/node_modules/<pkg>/package.json` under the OpenCode cache
root, and returns the `<workspace>` dir. Outside the cache (local-dev
checkouts, `bunx` scratch) it returns null.

`background-update-check.ts` now passes the resolved sandbox to
`syncCachePackageJsonToIntent()` and `resolveActiveInstallWorkspace()`,
so both the dependency write and `bun install` target the directory
the runtime is actually loading from. When no sandbox is detected the
previous fallback order (config-dir → flat cache → default) is
preserved for backwards compatibility with older flat installs.

## Verification

- `bun test src/hooks/auto-update-checker/ src/hooks/zauc-mocks-bg/ src/hooks/zauc-mocks-cache/ src/hooks/zauc-mocks-hook/ src/hooks/zauc-sync-mocks/` — 99 pass, 0 fail (includes new regression tests for the sandbox path).
- `npm run build`: PASS
- `npm test`: PASS for all touched suites. (One unrelated `sisyphus-task > should resolve agent-browser skill` test fails identically on `dev` before this change — pre-existing flake outside the auto-updater.)

---
## Summary by cubic
Fix the auto-updater to install into the same per-spec sandbox the runtime loads, removing stale version checks and the endless “Updated! Restart OpenCode” loop. The active workspace is resolved from `import.meta.url` and used for both `package.json` writes and `bun install`.

- **Bug Fixes**
  - Added `getLoadedSandboxWorkspace()` to locate the loaded sandbox from `import.meta.url`, verify `<workspace>/node_modules/<pkg>/package.json`, and use `realpath` (case-insensitive on Windows) to handle symlinks/junctions.
  - Changed update flow to invalidate → sync → install so `invalidatePackage` doesn’t wipe the sandbox `package.json` before `bun install`.
  - Routed `syncCachePackageJsonToIntent()` and installs to that sandbox; `resolveActiveInstallWorkspace()` now prefers sandbox → config dir → legacy cache → config default, with fallback when no sandbox is detected. Tests cover symlink/junction containment and the alias contract (writes canonical `oh-my-opencode` dep key in the sandbox).

<sup>Written for commit 8ec83410fe52c7af388fa3251202f28361d2a4ef. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4349?utm_source=github">Review in cubic</a></sup>


code-yeongyu:

## Scope
Only the **unique commit** (`b64f7a75a`) vs `dev`. Carrier commits are Yeongyu's stalled local work; out-of-scope.

## Verdict: NEEDS_REVISION

The diagnosis is correct - OpenCode's `Npm.add()` installs each plugin into a per-spec sandbox `<CACHE>/<sanitized-spec>/node_modules/<pkg>/`, and the previous flat-cache install kept the sandbox stale and re-fired the "Updated!" toast on every restart. `getLoadedSandboxWorkspace()` walking up from `import.meta.url` is the right approach. However there are concrete bugs that block "100% safe" today.

## Concerns

1. **Blocking: `invalidatePackage(PACKAGE_NAME)` deletes the sandbox we just wrote `package.json` into, before `bun install` runs.**
   New flow in `background-update-check.ts`:
   ```
   syncCachePackageJsonToIntent(pluginInfo, { sandboxWorkspace })   // writes <sandbox>/package.json
   invalidatePackage(PACKAGE_NAME)                                  // recursively removes cache entries
   resolveActiveInstallWorkspace(deps, sandboxWorkspace)            // returns <sandbox>
   runBunInstallSafe(<sandbox>)                                     // bun install in <sandbox>
   ```
   The existing `invalidatePackage` removes accepted specifier directories under the cache (see `src/hooks/auto-update-checker/cache.ts` and the test fixture in `src/hooks/zauc-mocks-cache/cache.test.ts` that explicitly asserts `packages/oh-my-openagent@latest` is deleted). If `sandboxWorkspace` happens to be that specifier dir, `package.json` is written, then the entire directory is removed, then `bun install` runs against a non-existent path. The order needs to be `invalidatePackage` first, then `syncCachePackageJsonToIntent`, then `runBunInstallSafe`.

2. **Possible alias regression: dependency key is `PACKAGE_NAME = "oh-my-opencode"` but the loaded sandbox is `oh-my-openagent@latest`.**
   The new test even asserts that an `oh-my-openagent@latest` sandbox ends up with `oh-my-opencode` as the dependency key (lines 433-461 of the diff). If the runtime is loaded from `node_modules/oh-my-openagent`, writing `oh-my-opencode` and running `bun install` may install/update the wrong package unless OpenCode aliases the two. Worth an explicit test for the `oh-my-openagent` dependency-key path before shipping.

3. **Windows path containment is case-sensitive and skips `realpath`.**
   `normalizedWorkspace.startsWith(cacheWithSep)` is structurally correct for exact root and subdirs, but on Windows it is case-sensitive and does not resolve junctions/symlinks. A real-world install via a junction or with a slightly different case in `LOCALAPPDATA` produces a false `null` and silently re-enables the broken legacy flat-cache path.

4. **No install fallback if sandbox install fails.** If `runBunInstallSafe(<sandbox>)` fails, the PR only shows the "update available" toast and gives up. Previously the flat cache was attempted. Consider falling back to the flat cache so a transient sandbox-permission error does not strand the user on a stale version.

## Merge recommendation
**Do not merge.** This is critical infrastructure (auto-update). Bugs here produce the same kind of infinite-toast / stale-load loops the PR is trying to fix. Also blocked by CLA failure on the head commit. Fix ordering of `invalidatePackage` vs sync, prove the `oh-my-openagent` dependency-key path with a test, and add `realpath`-aware containment, then this is mergeable.

code-yeongyu:
**[sisyphus-bot]** Rebase complete.

Force-pushed onto current `origin/dev` to drop the now-stale carrier commits. Only the unique sandbox-workspace fix remains; CI is re-running.

The blocking concerns from my [earlier review](#issuecomment-4526004307) still stand:
- `invalidatePackage(PACKAGE_NAME)` deletes the sandbox dir between the `package.json` sync and the `bun install`, so the install runs against a non-existent target.
- Dependency key is `oh-my-opencode` but the loaded sandbox is `oh-my-openagent` - the alias path needs an explicit test before shipping.
- Windows path containment is case-sensitive without `realpath`; junctions/casing will fail the check.

Merge remains blocked until those are addressed. Auto-update is critical infrastructure; a regression here re-creates the same infinite-toast loops the PR is fixing.

Yeachan-Heo:
Addressed bot review:

1. **Reordered to invalidate → sync → install.** `invalidatePackage` recursively removes accepted-specifier dirs (including the sandbox `oh-my-openagent@latest/`), so the previous order wiped the `package.json` `sync` just wrote, before `bun install` could see it. Order now: invalidate → sync → install.
2. **Realpath-aware Windows containment.** `path.resolve` + string prefix didn't handle junctions/symlinks or case differences. Both cache root and candidate workspace are now resolved through `realpath`, with case-insensitive comparison on `win32`.
3. **Explicit alias test.** Added `sync-package-json.test.ts` regression: an `oh-my-openagent@latest` sandbox still writes the canonical `oh-my-opencode` dep key (no duplicate alias dep). Also added an end-to-end `background-update-check.test.ts` test that asserts invalidate runs before sync, sync targets the sandbox, and install runs against the sandbox.

103/103 auto-updater tests pass (was 99/99 — 4 new tests). `npm run build` and `npm test` pass (one unrelated pre-existing flake on `sisyphus-task > should resolve agent-browser skill`, identical on `dev`).

code-yeongyu:
[sisyphus-dev] Reviewed this during the bug sweep for #4318. This PR takes the competing architecture path by trying to update the loaded sandbox workspace directly. Because #4318 now has multiple behavior-level fix strategies (#4349 vs #4535), this needs maintainer direction rather than an automatic merge. I am not merging or replacing it automatically.

0804sally:
**Windows verification of this PR (Windows 11, bun 1.3.14, non-admin shell)**

I tested this branch's auto-update-checker changes on a clean Windows 11 box to add a Windows data point for the maintainer decision. The production logic in `sandbox-workspace.ts` is correctly Windows-aware (realpath through junctions, case-insensitive containment) — but its test can't actually exercise the junction path on a standard Windows machine.

`bun test src/hooks/auto-update-checker/checker/sandbox-workspace.test.ts` → **6 pass / 1 fail**:

```
107 |       symlinkSync(realCache, linkedCache, "dir")
error: EPERM: operation not permitted, symlink '...\omo-sandbox-real-XXXX' -> '...-link'
(fail) getLoadedSandboxWorkspace (issue #4318) > resolves the sandbox even when the
       cache root is reached through a symlink/junction (issue #4349 follow-up)
```

On Windows, `fs.symlinkSync(target, link, "dir")` creates a *symbolic link*, which requires `SeCreateSymbolicLinkPrivilege` (Administrator or Developer Mode). So on a normal contributor/CI Windows setup the test errors at setup — meaning the realpath-through-junction logic this PR adds (the most Windows-relevant code here) is never actually asserted on Windows.

**Verified fix** — use a directory junction (no elevation needed, idiomatic on Windows):

```diff
-      symlinkSync(realCache, linkedCache, "dir")
+      symlinkSync(realCache, linkedCache, "junction")
```

With that single change, `sandbox-workspace.test.ts` → **7 pass / 0 fail** on Windows 11 (non-admin). Production code is unchanged.

Separately (not introduced by this PR), `plugin-entry.test.ts` has 2 Windows failures from POSIX-style assertions, e.g. `expect(configPath).toEndWith("/profiles/today/opencode.json")` receives `C:\...\profiles\today\opencode.json` (backslashes). Flagging in case it's relevant to the broader #4318 work.

Env: Windows 11, Node v24.14.0, bun 1.3.14, branch at `8ec83410fe52c7af388fa3251202f28361d2a4ef`.

code-yeongyu:
[sisyphus-bot] Thanks for the thorough fix. Issue #4318 has been resolved by PR #4535 (merged 2026-06-04) which took a different architectural approach — skipping the update install when running from an OpenCode-managed sandbox rather than aligning the install path. This PR is now superseded. If you believe the sandbox-alignment approach still has merit for edge cases not covered by #4535, please open a new issue or PR referencing the specific scenario.

PR #4352, opened 2026-05-23
fix: Atlas agent honors config.agents.atlas.model override (#4255)

### LINKED ISSUE #4255: Atlas agent ignores config model override — uses hardcoded claude-sonnet-4.6 instead
> **Reported by:** Sisyphus Minimax M2.5

## Description

The Atlas agent's model appears to ignore values set in the `oh-my-opencode.json` config file (`agents.atlas.model`) and instead uses a hardcoded default of `claude-sonnet-4.6`.

## Config used

```json
{
  "agents": {
    "atlas": {
      "model": "minimax-cn-coding-plan/MiniMax-M2.5-highspeed"
    }
  }
}
```

## Expected behavior

The model configured in `agents.atlas.model` should be used for the Atlas agent.

## Actual behavior

The opencode TUI shows Atlas defaulting to `claude-sonnet-4.6`, and this model is actually used at runtime (verified — wrong model runs).

## Additional context

- Other agents (Sisyphus, Oracle, Prometheus, etc.) correctly respect their config overrides
- Only Atlas appears to be affected
- This may be a hardcoded fallback in the binary that takes precedence over the config
- Plugin version: `oh-my-openagent-windows-x64-baseline` v4.0.0

## Possible cause

The Atlas agent's model resolution may have a hardcoded fallback (e.g., `claude-sonnet-4-6`) that is applied after/before the config merge, bypassing the user's config override.

Yeachan-Heo:
Closes #4255

## Root cause

`maybeCreateAtlasConfig` in `src/agents/builtin-agents/atlas-agent.ts`
bails out early (returning `undefined`) whenever `applyModelResolution`
returns `undefined` — even when the user explicitly set
`agents.atlas.model` in `oh-my-opencode.json`. When Atlas is dropped,
OpenCode falls back to its built-in agent default, which surfaces as
the hardcoded `claude-sonnet-4-6` fallback from
`AGENT_MODEL_REQUIREMENTS["atlas"].fallbackChain[0]`.

The other agent builders already handle this scenario:

- `general-agents.ts` (lines 82–93): falls back to `override?.model`
  when resolution returns `undefined`.
- `sisyphus-agent.ts`: has an explicit `hasSisyphusExplicitConfig` check
  to keep flowing when the user has configured the agent.

Atlas had neither, so under edge cases (cold provider cache, no system
default, empty `availableModels`) the user's explicit model was
silently discarded.

## Fix

Mirror the defensive pattern from `general-agents.ts`: when
`applyModelResolution` returns `undefined` but the user has explicitly
set `orchestratorOverride.model`, honor that model directly instead of
dropping Atlas:

```ts
if (!atlasResolution && orchestratorOverride?.model) {
  atlasResolution = { model: orchestratorOverride.model, provenance: "override" as const }
}
```

Adds a regression test in `src/agents/utils.test.ts` covering the
scenario where:
- `agents.atlas.model` is set to `minimax-cn-coding-plan/MiniMax-M2.5-highspeed`
  (the exact model from the bug report)
- `availableModels` is empty
- `systemDefaultModel` is `undefined`
- no provider cache exists

## Verification

- `npm run build`: PASS
- `bun test src/agents/`: PASS (406 tests, including new regression test)

---
## Summary by cubic
Fixes #4255 by making the Atlas agent honor `agents.atlas.model` when model resolution returns undefined, preventing an unintended fallback to `claude-sonnet-4-6`. Aligns Atlas with the defensive behavior used by other agents.

- **Bug Fixes**
  - In `src/agents/builtin-agents/atlas-agent.ts`, if `applyModelResolution` returns `undefined` and an override is set, use `{ model: override, provenance: "override" }` instead of dropping Atlas.
  - Added a regression test in `src/agents/utils.test.ts` for the cold-cache path: explicit `agents.atlas.model` (`minimax-cn-coding-plan/MiniMax-M2.5-highspeed`), empty `availableModels`, no `systemDefaultModel`, no provider cache.

<sup>Written for commit 7d444eed5e34445ca801d42a8ba4bcd996f0721c. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4352?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/agents/builtin-agents/atlas-agent.ts
+++ b/src/agents/builtin-agents/atlas-agent.ts
@@ -38,13 +38,19 @@ export function maybeCreateAtlasConfig(input: {
   const orchestratorOverride = agentOverrides["atlas"]
   const atlasRequirement = AGENT_MODEL_REQUIREMENTS["atlas"]
 
-  const atlasResolution = applyModelResolution({
+  let atlasResolution = applyModelResolution({
     uiSelectedModel: orchestratorOverride?.model !== undefined ? undefined : uiSelectedModel,
     userModel: orchestratorOverride?.model,
     requirement: atlasRequirement,
     availableModels,
     systemDefaultModel,
   })
+
+  if (!atlasResolution && orchestratorOverride?.model) {
+    // User explicitly configured a model but resolution failed (e.g., cold cache, no system default).
+    // Honor the user's choice directly instead of dropping Atlas entirely.
+    atlasResolution = { model: orchestratorOverride.model, provenance: "override" as const }
+  }
 
   if (!atlasResolution) return undefined
   const { model: atlasModel, variant: atlasResolvedVariant } = atlasResolution
--- a/src/agents/utils.test.ts
+++ b/src/agents/utils.test.ts
@@ -170,6 +170,30 @@ describe("createBuiltinAgents with model overrides", () => {
     }
   })
 
+  test("atlas honors user config model when resolution fails (no available models, no system default)", async () => {
+    // #given - regression for #4255: user sets agents.atlas.model but availableModels is empty
+    // and systemDefaultModel is undefined, so applyModelResolution returns undefined.
+    // Previous behavior: atlas was silently dropped, OpenCode used its built-in default.
+    // Expected behavior: honor the user's explicit model override.
+    const cacheSpy = spyOn(connectedProvidersCache, "readConnectedProvidersCache").mockReturnValue(null)
+    const fetchSpy = spyOn(shared, "fetchAvailableModels").mockResolvedValue(new Set())
+    const overrides = {
+      atlas: { model: "minimax-cn-coding-plan/MiniMax-M2.5-highspeed" },
+    }
+
+    try {
+      // #when - no systemDefaultModel, no availableModels, no cache
+      const agents = await createBuiltinAgents([], overrides, undefined, undefined)
+
+      // #then
+      expect(agents.atlas).toBeDefined()
+      expect(agents.atlas.model).toBe("minimax-cn-coding-plan/MiniMax-M2.5-highspeed")
+    } finally {
+      cacheSpy.mockRestore()
+      fetchSpy.mockRestore()
+    }
+  })
+
   test("Sisyphus is created on first run when no availableModels or cache exist", async () => {
     // #given
     const systemDefaultModel = "anthropic/claude-opus-4-7"


code-yeongyu:

## Scope
Only the **unique commit** (`f49333590`) vs `dev`. Carrier commits are Yeongyu's stalled local work; out-of-scope.

## Verdict: SAFE (but main CI did not run - cannot certify "100% safe")

The fix correctly mirrors the defensive pattern in `general-agents.ts:82-90`: when `applyModelResolution` returns `undefined` but the user has set `orchestratorOverride?.model`, honor the override directly via `{ model, provenance: "override" }` rather than dropping Atlas. The bug was real - in the cold-cache / no-system-default / empty-availableModels corner, Atlas was silently replaced by OpenCode's built-in `claude-sonnet-4-6` despite an explicit `agents.atlas.model`.

- `applyModelResolution` returns `undefined` only when no user, UI, category, fallback, or system model resolves (`packages/model-core/src/model-resolution-pipeline.ts:249-251`).
- Missing `variant` on the synthesized resolution is safe: Atlas only applies `variant` when truthy (`atlas-agent.ts:50-63`).
- Atlas fallback chain confirmed at `packages/model-core/src/model-requirements.ts:165-175`.

## Concern: CI did not run on this PR

Only `cla`, `cubic`, and `GitGuardian` checks appear on commit `f49333590` - the main CI workflow (`test`, `typecheck`, `build`, `block-master-pr`) was not triggered for this PR. Other PRs on the same branch family ran the full suite, so this is anomalous and the most likely cause is a workflow trigger gap on this specific run. **A push (even an empty commit or `gh workflow run ci.yml`) is needed to re-trigger the suite.** The change is trivial enough that I would expect it to pass, but I cannot certify "100% safe" without seeing typecheck + test green.

## Non-blocking note
The new regression test sets `userModel = "minimax-cn-coding-plan/MiniMax-M2.5-highspeed"`. Per `model-resolution-pipeline.ts:95-99`, a non-empty `userModel` returns immediately with `provenance: "override"`, so the test may not actually exercise the new fallback branch as a red/green regression. Consider an explicit unit test for `maybeCreateAtlasConfig` that mocks `applyModelResolution` to return `undefined` so the new branch is genuinely covered.

## Merge recommendation
**Hold for CI**: re-trigger the main CI workflow on `f49333590`. Once it goes green, this PR can be merged. The change itself is safe and well-targeted.

PR #4355, opened 2026-05-23
fix: swallow EPIPE/ECONNRESET shutdown errors in logger (#3772)

### LINKED ISSUE #3772: [Bug]: OmO creates huge log file due to EPIPE error
### Prerequisites

### Bug Description

oh-my-openagent writes session logs to disk (via session_output and transcript files) but imposes no size limit. On active projects or long sessions, logs grow unbounded — easily reaching multiple gigabytes on disk within days, clogging storage and making the .opencode/ directory painful to back up or clean.

There is no built-in log rotation, no truncation threshold, and no automatic cleanup of old session data.

Additionally, one specific error floods the logs unnecessarily during every shutdown:
`[2026-05-03T08:39:19.874Z] [background-agent] uncaughtException received during shutdown cleanup: {"code":"EPIPE","fd":2,"syscall":"write","errno":-32}`

This EPIPE error occurs because the background agent attempts to write to stderr after the parent process pipe has already closed. It is a harmless race condition — the process is shutting down anyway — but it logs this error on every session termination, compounding the log volume problem.

### Steps to Reproduce

Open any oh-my-openagent session
Perform normal work (run tools, delegate tasks, etc.)
Check `C:\Users\User\AppData\Local\Temp` (or OS-based equivalent) — the error above appears in the log file
Repeat steps 1–4 across multiple sessions — log files accumulate with no rotation, with very fast population

### Expected Behavior

Log files should respect a configurable maximum size (e.g. 100 MB per file)
A maximum number of retained log files should be configurable (e.g. 10 files)
Old files beyond the limit should be automatically deleted
The EPIPE race-condition error on shutdown should be caught and suppressed silently — it has no diagnostic value and only contributes noise and log volume

### Actual Behavior

Every second, a log file is written to `C:\Users\User\AppData\Local\Temp` (on Windows)
The EPIPE error is written to logs every second
No maximum size or file count is enforced — logs grow indefinitely
Over time, multiple gigabytes of logs accumulate in `%TEMP%`

### Doctor Output

```shell
oMoMoMoMo Doctor

 ✓ System OK (opencode unknown · oh-my-openagent unknown)
```

### Error Logs

```shell
A lot of:
[2026-05-03T08:39:19.874Z] [background-agent] uncaughtException received during shutdown cleanup: {"code":"EPIPE","fd":2,"syscall":"write","errno":-32}
```

### Configuration

```json

```

## Proposed Solution

Add a configurable log management in oh-my-openagent's config:
{
  "logs": {
    "max_size_mb": 100,
    "max_files": 10,
    "truncate_on_limit": true
  }
}

- `max_size_mb`: truncate or close session when log exceeds this size
- `max_files`: delete oldest logs when count exceeds this limit
- `truncate_on_limit`: if true, truncate the active log file rather than discarding it

Also, fix EPIPE error suppression during shutdown — it is a harmless race, not a meaningful error.

Alternatives Considered

- Manual cleanup only — document periodic cleaning of %TEMP%\.opencode\. Not acceptable; most users won't know and disk space issues are unexpected.
- OS-level cleanup — suggest Windows Task Scheduler or cron. Doesn't work reliably across platforms and places the burden on users.
- Compressed logs only — compress without size limits. Doesn't solve the growth problem.

### Operating System

Windows

### OpenCode Version

v1.14.33

Yeachan-Heo:
Closes #3772

## Root cause

`registerErrorEvent()` in `src/features/background-agent/process-cleanup.ts` installs `uncaughtException` and `unhandledRejection` listeners that log every transient error via the shared `log()` writer. During shutdown on Windows (and any time the parent's stdio pipe is torn down before the background worker drains its writes), Node re-emits `EPIPE` / `ECONNRESET` on writes to stderr/stdout thousands of times per second. Even one log line per emission compounded to multi-GB log files in `%TEMP%\.opencode\` (issue reports of 100+ GB in a few hours).

These errors carry no diagnostic value — the pipe is already closed and the process is about to exit. The existing in-process re-entry guard prevents synchronous recursion, but does not prevent Node from re-firing the event many times across microtasks.

## Fix

- Add `isHarmlessShutdownError(error)` helper that recognizes the `code` field on errno objects (covers Node's `SystemError` and the plain `{ code, errno, syscall, fd }` shape Node sometimes emits on stdio writes).
- In `registerErrorEvent`'s listener, return early without logging when `isHarmlessShutdownError(error)` is true.
- Apply the same filter to the inner `cleanupAll()` catch blocks so a registered shutdown handler that rejects with a stdio EPIPE during teardown does not contribute to bloat either.

Signal handlers (`SIGINT`/`SIGTERM`/`SIGBREAK`/`beforeExit`/`exit`) remain unchanged and still run `cleanupAll()` on real shutdown — only the error-event-driven logging is suppressed for these specific harmless errnos.

## Verification

- `npm run build`: PASS
- `bun test -t "isHarmlessShutdownError"`: 5/5 pass (3 positive cases — `EPIPE` Error, `ECONNRESET` Error, plain `{code:"EPIPE"}` object — and 2 negative cases covering unrelated errors and nullish/primitive values).
- `bun test -t "describeProcessCleanupError"`: 4/4 existing tests still pass.
- No public API change. The new exported `isHarmlessShutdownError` is annotated `@internal test-only seam`, matching the existing pattern for `describeProcessCleanupError`.

---
## Summary by cubic
Prevents multi-GB log bloat on shutdown by skipping only harmless `EPIPE`/`ECONNRESET` bursts in the background agent. The filter now matches stdio write errors or activates once shutdown begins, so real runtime socket errors still log. Fixes #3772.

- **Bug Fixes**
  - Added a shutdown window flag set by `SIGINT`/`SIGTERM`/`SIGBREAK`/`beforeExit`; `isHarmlessShutdownError` returns true only for stdio writes (`syscall: "write"`, `fd` 1/2) or when that flag is set.
  - Applied the filter in `uncaughtException`/`unhandledRejection` and `cleanupAll()` catch paths to drop shutdown-time bursts without hiding real runtime failures.
  - Extended tests to cover stdio vs non-stdio cases, shutdown gating and flips on `SIGINT`/`beforeExit`, and `uncaughtException` after `SIGINT`.

<sup>Written for commit 8ffb9214af6c430868a6b72d134eee3f0ff282c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:

## Scope
Only the **unique commit** (`8f1987a2f`) vs `dev`. Carrier commits are Yeongyu's stalled local work; out-of-scope.

## Verdict: NEEDS_REVISION

The bug being fixed is real and severe - Windows shutdown loops produce EPIPE/ECONNRESET storms that compound to multi-GB log files in `tmpdir`. But the fix scope is wider than the bug, and the existing diagnostic comments in the same file explicitly describe non-shutdown failure modes that would now disappear silently.

## Concerns

1. **Blocking: suppression is global, not shutdown-scoped.**
   `registerErrorEvent` installs handlers for `uncaughtException` and `unhandledRejection` at plugin entry (`process-cleanup.ts:95, 199`). These handlers are **normal-runtime** safety nets, not only shutdown-time. The new `isHarmlessShutdownError` filter applies to every event that reaches them. A real network bug that surfaces as an unhandled rejection with `code === "ECONNRESET"` (e.g. a mid-stream provider socket reset) now produces zero log output - it is invisible to operators.

2. **Existing in-file comments explicitly cite legitimate non-shutdown ECONNRESET / EPIPE sources.**
   `process-cleanup.ts:107` references mid-stream socket resets; `:118` references MCP / provider reconnect failures. The current behavior logs them with a one-time-per-listener throttle, which is exactly the visibility the new code removes. If the multi-GB log-file pathology is shutdown-bound (issue #3772 describes it as such), the filter should also be gated on a shutdown-window flag (e.g. set in `beforeExit` / on `SIGINT` arrival).

3. **`cleanupAll()` catch blocks also silently swallow these codes.**
   Diff lines 117-119 (sync) and 113-114 (async) now drop EPIPE/ECONNRESET from registered manager `shutdown()` failures too. A manager that legitimately fails on a non-stdio ECONNRESET during teardown loses all visibility.

4. **Filter is too broad relative to its own comment.**
   The header comment claims "EPIPE / ECONNRESET on stdio (fd 1 or 2) carries no diagnostic value", but the filter only matches `code` - any object with `code === "EPIPE"` qualifies, including non-stdio sources. If the intent is fd 1/2 only, check `syscall === "write"` AND `fd in (1, 2)` (Node sets `fd` on the errno for stdio).

## Recommendation
**Do not merge as-is.** Two safe alternatives:
- Gate the silent drop on a shutdown flag set by `SIGINT`/`SIGTERM`/`SIGBREAK`/`beforeExit` (so non-shutdown EPIPE/ECONNRESET still logs).
- OR keep silent drop but rate-limit logging (e.g. log once per error code per minute) so the multi-GB pathology is broken without nuking visibility.

The test suite for `isHarmlessShutdownError` is solid; just narrow the scope before shipping.

code-yeongyu:
**[sisyphus-bot]** Rebase complete.

Force-pushed onto current `origin/dev` to drop the now-stale carrier commits. Only the unique EPIPE/ECONNRESET filter remains; CI is re-running against the rebased branch.

The blocking concern from my [earlier review](#issuecomment-4526004441) still applies:

`registerErrorEvent` installs handlers for `uncaughtException` and `unhandledRejection` at plugin entry as **normal-runtime** safety nets, not shutdown-only paths. The filter you add is global: it suppresses the same codes during regular operation, and the existing in-file comments (`process-cleanup.ts:107, 118`) explicitly cite legitimate non-shutdown sources (mid-stream socket resets, MCP/provider reconnect failures). A real `ECONNRESET` bug there would now produce zero log output.

Two safe alternatives:
- Gate the silent drop behind a shutdown-window flag set by `SIGINT`/`SIGTERM`/`SIGBREAK`/`beforeExit`.
- OR keep silent drop but rate-limit logging (e.g. log once per code per minute) so the multi-GB pathology is broken without nuking visibility.

The harmless-error test suite is solid - just narrow the scope before merge.

Yeachan-Heo:
Addressed sisyphus-bot review: narrowed `isHarmlessShutdownError` so it no longer hides non-shutdown EPIPE/ECONNRESET (mid-stream provider socket resets, MCP reconnect failures). An errno is now treated as harmless only if EITHER (1) it identifies a stdio write — `syscall === "write"` AND `fd ∈ {1, 2}`, matching Node's SystemError shape — OR (2) a shutdown signal (`SIGINT`/`SIGTERM`/`SIGBREAK`/`beforeExit`/`exit`) has already fired, setting a new `_shutdownInProgress` flag before `cleanupAll()` runs. The same helper continues to gate the `cleanupAll()` catch blocks, so the narrowing applies there too. Existing tests updated to use proper SystemError shapes; 6 new tests cover the gate (non-stdio ECONNRESET/EPIPE must log outside the window, must be silenced inside it, `_shutdownInProgress` flips on `SIGINT`/`beforeExit`, and `uncaughtException` with non-stdio ECONNRESET after `SIGINT` is dropped). `npm run build`: PASS. `bun test -t isHarmlessShutdownError`: 11/11 pass. `bun test -t "shutdown-window flag"`: 5/5 pass.

code-yeongyu:
[sisyphus-dev] I found this while sweeping bug issue #3772. Current `dev` already covers log rotation and the broad re-entry guard, but this PR still targets the remaining harmless shutdown stdio `EPIPE`/`ECONNRESET` noise path. I current-based the PR locally and ran focused QA: `bun test src/shared/logger.test.ts src/features/background-agent/process-cleanup.test.ts src/features/background-agent/manager-shutdown-global-cleanup.test.ts --bail` passed in the review worktree. I am updating the branch now so current CI can re-run before any merge decision.

PR #4408, opened 2026-05-24
fix(background-agent): honor provider concurrency groups

### LINKED ISSUE #3968: bug: OMO scheduler can ignore provider concurrency limit for subagents
## Summary

Discord feedback reports that OMO can still create concurrent subtask/subagent calls even when the underlying OpenCode provider configuration has a concurrency limit of 1. This breaks providers/models that allow only one in-flight request, such as GLM-5.1.

## Reported scenario

- User only uses `GLM-5.1` for OMO.
- The model/provider has a concurrent request limit of `1`.
- The user configured a concurrency limit in `opencode.json`.
- OMO still creates subtasks/subagents in a way that triggers concurrent provider calls and provider errors.
- User asked whether there is an `oh-my-openagent` setting to apply/enforce this concurrency limit.

## Expected behavior

OMO's team/subtask/subagent scheduler should respect provider/model concurrency limits from OpenCode config, or expose an OMO-level cap such as:

```jsonc
{
  "team_mode": {
    "max_parallel_members": 1,
    "max_members": 1
  },
  "subagents": {
    "max_concurrent": 1
  }
}
```

When concurrency is capped at `1`, OMO should execute delegated work sequentially instead of launching parallel provider calls.

## Actual behavior

OMO appears to treat subtask/subagent creation as parallel work even when the provider path only supports one concurrent request. The provider then errors because concurrent calls exceed the model/provider limit.

## Workaround

Until scheduler-level enforcement exists, users can force sequential behavior in the task prompt:

```text
This provider has concurrent request limit 1.
Do not create concurrent subagents, team members, or parallel tool/model calls.
Run all work sequentially: finish one delegated task before starting the next.
If the task requires parallel execution, stop and report CONCURRENCY_LIMIT_BLOCKED.
```

If team mode is involved, also keep the team size to one where supported.

## Useful repro details to collect

```text
OMO version
OpenCode version
provider/model id for GLM-5.1
redacted opencode.json concurrency config
redacted oh-my-openagent config
exact provider error text
whether this happens via team mode, subagents, or both
whether UI shows multiple workers/subtasks started at once
```

PeterPonyu:
## Summary

- Add `ConcurrencyManager.getConcurrencyKey()` to canonicalize exact model keys into provider keys when `providerConcurrency` is configured.
- Apply canonical concurrency keys when launching, tracking, and resuming background tasks.
- Add regression coverage for provider-level serialization across different model variants and persisted/resumed task groups.

## Why

Provider-level concurrency limits should prevent concurrent delegated/background calls across multiple models on the same provider. Without key canonicalization, tasks can acquire separate exact-model slots and bypass a provider cap of `1`.

Closes #3968

## Verification

Local preflight was run on latest `upstream/dev` (`de7f1d887db6f274ae67075a360cd05d45f59179`) with CI-pinned Bun `1.3.12`:

```bash
git diff --check
npm --prefix packages/lsp-tools-mcp ci
npm --prefix packages/lsp-tools-mcp run build
BUN_INSTALL_ALLOW_SCRIPTS='@ast-grep/cli @ast-grep/napi' bun install --frozen-lockfile
bun test
bun run typecheck
bun run typecheck:script
bun run build
test -f dist/index.js
test -f dist/index.d.ts
bun test src/shared/dist-bundle-bun-globals.test.ts
```

All passed locally before publication.

---
## Summary by cubic
Ensure background-agent enforces provider-level concurrency by canonicalizing model keys to provider keys so models from the same provider share one slot, including for explicit keys and resumed tasks. Prevents background/delegated tasks from bypassing a provider cap.

- **Bug Fixes**
  - Added `ConcurrencyManager.getConcurrencyKey()` to return provider keys when `providerConcurrency` is set; keeps exact model keys when `modelConcurrency` exists; otherwise returns the model key.
  - Canonicalized keys when deriving model/agent keys, honoring explicit `concurrencyKey` on launch/track, acquiring/releasing, and re-acquiring on resume by re-canonicalizing persisted `concurrencyGroup`.
  - Added tests for `getConcurrencyKey`, provider-level serialization across different models, and resumed tasks; added a guard when shifting queued waiters.

<sup>Written for commit aecd23f7494e20061d527804bed8ba3495556fe2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
[sisyphus-dev] I confirmed this PR is the direct fix path for #3968 during the bug sweep.

Current review evidence before updating the branch:
- Checked out the current GitHub merge ref for #4408 and installed dependencies in an isolated worktree.
- `bun test src/features/background-agent/concurrency.test.ts src/features/background-agent/manager.test.ts --test-name-pattern "ConcurrencyManager|provider concurrency|resume concurrency key|Non-blocking Queue Integration"` -> 69 pass / 0 fail.
- `bun run typecheck` -> pass.
- `git diff --check` -> pass.

I also updated the PR branch against the latest `dev` so full CI can rerun on a fresh head. I will merge only if the refreshed CI and review gates are clean; otherwise I will either fix in a replacement worktree or leave it open with the issue assigned, depending on scope.

code-yeongyu:
[sisyphus-dev] I found a blocker while reviewing this PR against #3968, and I am going to fix it in a replacement PR.

What works here: normal launch/track/resume paths now canonicalize provider-level keys, and the focused manager/concurrency tests pass on the current merge ref.

Blocker: fallback retry still enqueues/processes retries using the exact `provider/model` key path. With `providerConcurrency: { anthropic: 1 }`, that can let a normal provider-key task and a fallback retry for `anthropic/model-b` acquire separate slots and run concurrently, so the provider-wide cap is still bypassable.

This is still localized scheduler behavior, not a broad architecture rewrite. I will create a replacement from current `dev` that preserves this PR direction and adds the missing fallback retry canonicalization/regression coverage. I will link the replacement here before closing/superseding anything.

code-yeongyu:
[sisyphus-dev] I opened replacement PR #4680 to supersede this branch: https://github.com/code-yeongyu/oh-my-openagent/pull/4680

I preserved this PR's scheduler fix as the first commit, then added the missing fallback-retry provider-concurrency regression and fix. Closing this PR so review and CI converge on the replacement.

PR #3966, opened 2026-05-12
fix(team-mode): resolve port-0 server-url discard and surface visualization skip

jas32096:
## Summary

- Replace all silent `return null` paths in `createTeamLayout` with a typed discriminated-union `TeamLayoutOutcome` so callers always know *why* visualization was skipped.
- Surface layout-skip events as structured `warnings[]` in the `team_create` tool response when `tmux_visualization` is enabled but a layout could not be created — giving users an actionable hint instead of silence.
- Fix `team_delete`'s `removedLayout` field so it reflects whether a layout was actually created at runtime, not whether it was merely configured.

## Root Cause

In default TUI mode (`opencode` with no `--port` / no `OPENCODE_PORT`), `TmuxSessionManager` falls back to `http://localhost:4096`. Because nothing is listening on `:4096` in that mode, `isServerRunning()` returns `false` and `createTeamLayout` returned `null` — with no log output surfaced to the user, no warning in the `team_create` response, and a misleading `removedLayout: true` from `team_delete`. The full silent-failure chain:

1. `TmuxSessionManager` constructor: when `ctx.serverUrl` port equals `'0'` (random-bind default), the real bound address is discarded and replaced by a hardcoded `localhost:4096` fallback.
2. `createTeamLayout`: `if (!await deps.isServerRunning(serverUrl)) { return null; }` — the `null` return propagates silently; the tool response omits any hint.
3. `team_delete`: `removedLayout` was derived from config flags (`tmux_visualization && tmuxMgr !== undefined && canVisualize()`), always `true` even when no layout existed.

This is the generalized recurrence of the `http://127.0.0.1:0/` case from #2729 (closed), which was not fully addressed in v4.

## Changes

- **`src/features/tmux-subagent/manager.ts`**: Expose `serverUrlSource: 'ctx' | 'env' | 'default'` on `TmuxSessionManager` so callers can distinguish whether the URL came from `ctx.serverUrl`, the `OPENCODE_PORT` env var, or the hardcoded `:4096` default. Preserves existing port-`'0'` fallback semantics; adds observability without behavior change.

- **`src/features/team-mode/team-layout-tmux/layout.ts`**: Change `createTeamLayout` return type from `TeamLayoutResult | null` to the discriminated union `TeamLayoutOutcome`:
  ```ts
  type TeamLayoutOutcome =
    | { skipped: false; layout: TeamLayoutResult }
    | { skipped: true; reason: SkipReason; detail?: string; serverUrl?: string; serverUrlSource?: string }

[...]

PR #4410, opened 2026-05-24
fix(team-mode): preserve server auth for tmux pane attach

### LINKED ISSUE #4409: bug(team-mode): tmux visualization panes do not inherit opencode server password
## Summary

When Team Mode tmux visualization creates teammate panes, each pane runs `opencode attach` against the leader's opencode server. If the server is protected by `OPENCODE_SERVER_PASSWORD`, the generated attach command does not pass that environment variable into the pane command, so visualization panes can fail to attach even though the leader process has the correct secret.

## Expected

Team Mode tmux panes should inherit the opencode server auth secret needed to attach to the active server.

## Actual

The attach command only includes:

```bash
opencode attach <server-url> --session <session-id> --dir <cwd>
```

so a password-protected server may reject the pane attach attempt.

## Related context

This is related to the Team Mode tmux visualization area but distinct from #3963 / PR #3966. Those focus on silent layout skips and port-0/server-url diagnostics. This issue is only about preserving server auth once a pane attach command is created.

## Suggested fix

When `OPENCODE_SERVER_PASSWORD` is present in the leader process environment, prefix the pane attach command with a safely shell-quoted assignment:

```bash
OPENCODE_SERVER_PASSWORD='<secret>' opencode attach ...
```

Add regression coverage for spaces, single quotes, and shell metacharacters in the password.

PeterPonyu:
## Summary

- Prefix generated Team Mode tmux `opencode attach` commands with `OPENCODE_SERVER_PASSWORD` when the leader process has it set.
- Keep the password shell-quoted so spaces, single quotes, and shell metacharacters are preserved safely.
- Add regression coverage for the attach command emitted to teammate panes.

## Why

Tmux visualization panes attach to the active opencode server. If that server requires `OPENCODE_SERVER_PASSWORD`, panes need the same auth secret as the leader process or attach can fail.

Closes #4409

Related to #3963 / #3966, but intentionally separated: this PR only handles auth propagation for panes that are created; #3966 handles skipped-layout diagnostics and port/server-url behavior.

## Verification

Local CI-equivalent preflight was run on latest `upstream/dev` (`de7f1d887db6f274ae67075a360cd05d45f59179`) with CI-pinned Bun `1.3.12`:

```bash
git diff --check
npm --prefix packages/lsp-tools-mcp ci
npm --prefix packages/lsp-tools-mcp run build
BUN_INSTALL_ALLOW_SCRIPTS='@ast-grep/cli @ast-grep/napi' bun install --frozen-lockfile
bun test
bun run typecheck
bun run typecheck:script
bun run build
test -f dist/index.js
test -f dist/index.d.ts
bun test src/shared/dist-bundle-bun-globals.test.ts
```

All passed locally before publication.

### DIFF
--- a/src/features/team-mode/team-layout-tmux/layout.test.ts
+++ b/src/features/team-mode/team-layout-tmux/layout.test.ts
@@ -12,6 +12,7 @@ let nextWindowNumber = 1
 let nextPaneNumber = 1
 let displaySessionId = "$7"
 let displaySuccess = true
+let originalServerPassword: string | undefined
 const panesByWindow = new Map<string, string[]>()
 
 function createTmuxCommandResult(output: string, success = true) {
@@ -113,9 +114,16 @@ function getCommands(): Array<Array<string>> {
 describe("team-layout-tmux", () => {
   afterEach(() => {
     mock.restore()
+    if (originalServerPassword === undefined) {
+      delete process.env.OPENCODE_SERVER_PASSWORD
+    } else {
+      process.env.OPENCODE_SERVER_PASSWORD = originalServerPassword
+    }
   })
 
   beforeEach(() => {
+    originalServerPassword = process.env.OPENCODE_SERVER_PASSWORD
+    delete process.env.OPENCODE_SERVER_PASSWORD
     runTmuxCommandMock.mockClear()
     isServerRunningMock.mockClear()
     isServerRunningMock.mockImplementation(async () => true)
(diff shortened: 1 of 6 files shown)


PeterPonyu:
Superseded by #4466 — surgical scope reduction. The new PR carries only the team-mode visualization attach fix for #4409 (+58/-1, 2 files), inlined in `layout.ts` instead of extracting a shared helper that had only one consumer in this PR. If a future PR needs the env-prefix for subagent panes, the helper can be extracted then.

Issue #3191, opened 2026-04-07
[Bug]: Skip providers with exhausted credits during task delegation


vishnukv-facets:
## Problem

When a provider's credits are exhausted, oh-my-openagent still delegates tasks to that provider. The task fails with a `quota_exceeded` error, wasting time and tokens. There are **three** compounding issues:

### 1. No proactive filtering of exhausted providers

`getAvailableModelsForDelegateTask()` in `src/tools/delegate-task/available-models.ts` returns models from all connected providers — even those whose credits are exhausted. There is no mechanism to track or exclude providers that have previously failed with quota errors.

### 2. `quota_exceeded` errors kill the session instead of triggering fallback

In `src/hooks/runtime-fallback/error-classifier.ts` (line 170–175):

```typescript
if (errorType === "quota_exceeded") {
  const hasAutoRetrySignal = /retrying\s+in/i.test(message)
  return hasAutoRetrySignal  // returns false unless "retrying in" is in the message
}
```

`isRetryableError()` returns `false` for `quota_exceeded`, so `handleSessionError()` in `event-handler.ts` (line 152) bails out — the fallback chain **never triggers**. The session simply dies.

### 3. No cross-session memory of exhausted providers

Even if fallback worked, the system only tracks failed models per-session via `FallbackState.failedModels` (60s cooldown). A new delegation to the same exhausted provider will fail again because there's no persistent state.

## Current Flow (broken)

```
User delegates task
  → resolveSubagentExecution()
    → getAvailableModelsForDelegateTask()  ← includes exhausted providers
    → resolveModelForDelegateTask()        ← picks model from exhausted provider
  → Provider API call fails (quota_exceeded)
  → handleSessionError()
    → classifyErrorType() returns "quota_exceeded"
    → isRetryableError() returns false     ← BUG: no fallback triggered
  → Session dies. No fallback. No learning.
```

### A. Persistent exhausted-providers cache (new file)

Create `src/shared/exhausted-providers-cache.ts` using the existing `createJsonFileCacheStore` pattern (same as `connected-providers-cache.ts`). Stores providers that have returned quota errors with a TTL (e.g., 4 hours):

```typescript
interface ExhaustedProviderEntry {
  reason: string
  exhaustedAt: string
  ttlMs: number
}

interface ExhaustedProvidersCache {
  providers: Record<string, ExhaustedProviderEntry>
  updatedAt: string
}
```

[...]

PR #4425, opened 2026-05-25
fix(runtime-fallback): harden fallback dispatch, watchdog lifecycle, and error classification

### LINKED ISSUE #4528: Runtime fallback and background subagents can stall or mark completed work interrupted
## Summary
OpenCode sessions using Oh My OpenAgent can stall silently or mark completed subagent work as interrupted during provider retry/fallback and Atlas/background-task orchestration.

The failure appears as a main command session that exits or stalls without a useful error, or as subagents that either stall, report `interrupted`, or fail internal retries with an agent-name mismatch.

## Failure modes observed
- Runtime fallback stores the OpenCode model payload as if it is always a string. When OpenCode sends object-form models, fallback comparison can throw `current.toLowerCase is not a function`.
- First-prompt watchdog only covered subagents and used abort+prompt fallback. Main command sessions such as `opencode run --command start-work --agent "Atlas - Plan Executor"` could stay silent until OpenCode provider retry handling took over or the command exited.
- `session.status` provider retry signals caused runtime fallback to abort the current request and dispatch an internal prompt immediately. In CLI command mode this can end the command runner instead of letting OpenCode's provider retry carry the queued fallback model.
- Runtime fallback internal prompts pass plugin display-agent names such as `Atlas - Plan Executor`; OpenCode internal prompt routes only know core agent keys (`build`, `explore`, `general`, `plan`), producing `Agent not found`.
- Background task completion aborts child sessions even after they have already reached idle/completed state, which can create false interrupted markers.
- Atlas/Ralph continuation gates only considered `running` background tasks active, so pending background work could be skipped during parent continuation decisions.
- Atlas task timers for background-launched work were not started through the same timer path and plan checkbox timer closure could happen after the orchestrator-only guard.

## Expected behavior
- Fallback state accepts both string and object model payloads.
- Main command sessions and subagents should recover through configured fallback models without silently exiting.
- Provider auto-retry signals should queue the fallback model for OpenCode's next retry instead of aborting the request in command mode.
- Internal fallback prompts should only pass OpenCode internal agent keys, or omit plugin display agents.
- Completed idle child sessions should not be aborted after successful background task completion.
- Pending background tasks should block parent continuation just like running tasks.

## Local verification
- `bun test` targeted runtime fallback tests
- `bun test` Atlas task timer/background launch/idle continuation tests
- `bun test src/plugin/fallback.cliproxyapi-matrix.test.ts`
- `bun run typecheck`
- `bun run build`
- End-to-end local patched cache stress: Atlas command sessions launched multiple background subagents and completed without stall/interrupted markers.

A PR with the fix is being opened from `fix/runtime-fallback-background-stalls`.

### LINKED ISSUE #4787: [Bug]: runtime-fallback abort→promptAsync race condition causes main session QUEUED stall in TUI mode
## Prerequisites

## Bug Description
In TUI (interactive) mode, when the primary model provider hits a quota/usage-limit error, the `runtime-fallback` hook triggers its abort→retry flow. However, `promptAsync` with the fallback model silently fails after the session has been aborted, leaving `sessionRetryInFlight` set to `true`. This blocks all subsequent user prompts, which show as "QUEUED" indefinitely. The session only recovers after `SESSION_TTL_MS` (30 minutes) expires and cleans up the stale state.

Related issues: #4528 (general stall), #3745 (CLI mode). This issue specifically documents the TUI-mode QUEUED stall with detailed log evidence.

## Steps to Reproduce
1. Configure `runtime_fallback.enabled: true` in `oh-my-openagent.json` with `fallback_models` for the sisyphus agent.
2. Use a primary model from `opencode-go` provider (e.g., `opencode-go/mimo-v2.5-pro`).
3. Exhaust the opencode-go provider quota (5-hour usage limit).
4. Send a prompt in TUI mode.
5. Observe: runtime-fallback detects `quota_exceeded`, aborts the session, shows toast notifications, but the prompt shows "QUEUED" and never executes.
6. Subsequent prompts also show "QUEUED" — the session is stuck.
7. After ~30 minutes (`SESSION_TTL_MS`), the stale state is cleaned up and the session becomes usable again (only if the user manually switches models).

## Expected Behavior
After `abortSessionRequest`, `promptAsync` with the fallback model should successfully re-dispatch the user prompt. If `promptAsync` fails, `sessionRetryInFlight` should be cleared so the session is not permanently blocked.

## Actual Behavior
`promptAsync` silently fails after abort, `sessionRetryInFlight` remains `true`, and all subsequent prompts are blocked (QUEUED). No error is logged during the 30-minute stall period.

## Doctor Output
```shell
 oMoMoMoMo Doctor 

 ⚠ 3 issues found:

1. oh-my-opencode is not registered
   Plugin entry is missing from OpenCode configuration.
   Fix: Run: bunx oh-my-opencode install

2. Loaded plugin is outdated
   Loaded 3.5.5, latest 4.7.5.
   Fix: Update: cd "/Users/LinRui/.cache/opencode" && bun add oh-my-opencode@latest

3. Comment checker unavailable
```

## Error Logs
Key log entries (timestamps in Asia/Shanghai, from `~/.local/share/opencode/log/`):

```
INFO  2026-06-05T03:54:38 service=bus type=session.error publishing
INFO  2026-06-05T03:55:08 service=session.prompt session.id=ses_16e398e74ffed1ynASMvYv74nA cancel
ERROR 2026-06-05T03:55:08 service=session.processor session.id=ses_16e398e74ffed1ynASMvYv74nA messageID=msg_e95eb2cc8001a9rSKxm5s3QWtk error=Aborted process
INFO  2026-06-05T03:55:08 service=bus type=session.error publishing
INFO  2026-06-05T03:55:08 service=bus type=session.status publishing
INFO  2026-06-05T03:55:08 service=bus type=tui.toast.show publishing   ← fallback notification
INFO  2026-06-05T03:55:08 service=bus type=tui.toast.show publishing   ← fallback notification
                   ← 30 MINUTE GAP — no logs at all — QUEUED stall ←
INFO  2026-06-05T04:25:33 service=bus type=session.next.model.switched publishing  ← manual /models switch
INFO  2026-06-05T04:25:33 service=session.prompt session.id=ses_16e398e74ffed1ynASMvYv74nA step=0 loop
```

The `+1824970ms` gap between toast notifications and `session.next.model.switched` confirms the QUEUED stall.

## Configuration
```json
{
  "agents": {
    "sisyphus": {
      "model": "opencode-go/mimo-v2.5-pro",
      "fallback_models": ["aistore/DeepSeek-V4-Pro"]
    }
  },
  "runtime_fallback": {
    "enabled": true,
    "retry_on_errors": [429, 500, 502, 503, 504],
    "max_fallback_attempts": 3,
    "cooldown_seconds": 60,
    "timeout_seconds": 30,
    "notify_on_fallback": true
  }
}
```

### Root Cause Analysis
The bug is a race condition in `autoRetryWithFallback` (source: `src/hooks/runtime-fallback/auto-retry.ts`):

1. `session.error` fires → `classifyErrorType()` returns `"quota_exceeded"` → `isRetryableError()` returns `true` ✅
2. `dispatchFallbackRetry` → `autoRetryWithFallback` is called
3. **Step 1**: `abortSessionRequest()` aborts the in-flight request → session enters an abnormal state
4. **Step 2**: `ctx.client.session.messages()` reads history
5. **Step 3**: `ctx.client.session.promptAsync({ model: fallbackModel })` is called on the same aborted session → **silently fails** (no error caught, no log)
6. `sessionRetryInFlight` was set to `true` before `promptAsync` and is never cleared because `retryDispatched` remains `false`
7. All subsequent prompts are blocked by the `sessionRetryInFlight` guard → **QUEUED**

The session only recovers after `SESSION_TTL_MS = 30 * 60 * 1000` (30 minutes) when `cleanupStaleSessions` removes the stale state.

### Key Code Paths
- `src/hooks/runtime-fallback/auto-retry.ts`: `autoRetryWithFallback()` — the abort→promptAsync race
- `src/hooks/runtime-fallback/constants.ts`: `SESSION_TTL_MS = 30 * 60 * 1000`
- Source line `index.js:89063-89096`: The `finally` block only clears `sessionRetryInFlight` when `retryDispatched` is `false`, but `promptAsync` failure does not throw — it resolves silently

### Workaround
Removing `runtime_fallback` from the configuration restores the legacy `model-fallback` hook, which does not abort the session. It requires the user to send another prompt to trigger the model switch, but it does not cause QUEUED stalls.

### Suggested Fix
1. Wrap `promptAsync` in a timeout/retry with proper error handling — if it does not resolve within `timeout_seconds`, treat it as a failure
2. In the `finally` block, always clear `sessionRetryInFlight` and `sessionAwaitingFallbackResult` regardless of `retryDispatched`
3. Consider not aborting the session before retrying in TUI mode — let OpenCode handle the model switch on the next prompt cycle instead

## Operating System
macOS

## OpenCode Version
1.15.13

niStee:
## Summary

Comprehensive runtime-fallback hardening: fixes synchronous delegation stalls, watchdog lifecycle races, event model type crashes, error classification for Go/Zen providers, and adds provider-failure coordination with cooldown enforcement.

## Core Fixes

- **Sync fallback stall** (`auto-retry.ts`): Release interrupted synchronous prompt reservations during provider retry-signal fallback, so the replacement model can be dispatched.
- **Watchdog re-arming race** (`first-prompt-watchdog.ts`): Track completed sessions so late `message.updated` events don't re-arm the watchdog after a terminal state. Completed set capped at 1000 entries to prevent unbounded memory growth.
- **Object-shaped event models** (`event-model.ts`): Normalize `{providerID, modelID, variant}` objects into `"provider/model(variant)"` strings at all event boundaries (already upstream in dev).
- **Provider-failure cooldown** (`provider-failure-state.ts`): Track failed providers in a `Map` with auto-expiry after 2-minute cooldown, replacing a permanent `Set` that never recovered.

## Error Classification Improvements

- Add `extractErrorCode()` and `extractErrorType()` to classify OpenCode Go/Zen errors by their `code` and `type` fields (e.g. `insufficient_quota`, `CreditsError`).
- Detect `empty_output` and additional `model_not_found` patterns (provider not found, selected provider forbidden) as retryable error types.
- Add Zen-specific retryable error patterns to `RETRYABLE_ERROR_PATTERNS`.

## Fallback Chain Improvements

- **Variant delimiter unified** (`fallback-state.ts`): Changed from `::` to `()` to match `event-model.ts` output, fixing variant equivalence matching.
- **stripVariant centralized** (`fallback-state.ts` → `chat-message-handler.ts`): Exported from fallback-state and removed inline `any`-typed duplicate.
- Skip equivalent fallback models that differ only by variant (avoids wasteful retries on the same underlying model at a different quality tier).
- Bridge independent fallback systems via shared provider-failure state so model-fallback and runtime-fallback hooks coordinate.
- Treat reasoning content (`type: "reasoning"`) as visible assistant output, preventing false-positive "empty output" fallback triggers.

## Other

- Removed unused `combinedErrorIdentifier` variable from error-classifier.
- Reuse existing output-validation results in `manager.ts` to avoid a duplicate `messages` API call.

## Related Issues

Closes #4787 — abort→promptAsync race condition causing QUEUED stall in TUI mode
Closes #4528 — runtime fallback stalls and interrupted markers on completed subagent work
Addresses #3738 — ordered provider fallback across subscription quota exhaustion (partial; full resolution tracked in #4886)
Addresses #3191 — skip providers with exhausted credits during task delegation (partial; full resolution tracked in #4886)

## Reproduced Sync Stall

On May 25, 2026, a live OpenCode run reproduced a stale synchronous subagent on an OpenAI child session:

1. A synchronous delegated prompt acquired the prompt gate as `model-suggestion-retry:sync`.
2. `session.status` emitted a provider auto-retry signal and runtime-fallback aborted the in-flight request.
3. Reservation release was skipped because the abort path only accepted a `runtime-fallback:*` owner, not the interrupted synchronous owner.
4. The replacement fallback prompt was therefore not accepted, while the sync poller continued reporting the child as absent from status.
5. The polling persisted for approximately 655 seconds and ended only when the parent run was stopped by the user.

## Validation

- `bun test src/hooks/runtime-fallback/ src/shared/provider-failure-state.test.ts` — **237 pass**, 0 fail
- `bun run typecheck` — clean
- `bun run build` — clean
- Deployed-bundle smoke run with a failing primary model and OpenAI fallback completed with `CHILD_OK`
- Credential scan: no findings in diff
- GitGuardian: pass


niStee:
Rebased onto latest `dev` and cherry-picked additional code-review fixes (commit `66aa929`): variant delimiter fix, provider-failure cooldown enforcement, stripVariant deduplication, completed-set cap, unused variable removal, and 8 new provider-failure-state tests. Conflicts resolved — branch now starts from current dev tip. Typecheck + 237 runtime-fallback tests pass locally.

niStee:
**Merge Conflict Resolution Note: Sync vs Async Fallback Locks**

For reviewers: This PR has been successfully updated with the latest `upstream/dev` branch. 

During the merge conflict resolution, we explicitly kept the fix from this PR (adding `"model-suggestion-retry:sync"` to the prompt release array in `auto-retry-abort.ts`) alongside the recent fallback backoff loop introduced in upstream/dev.

**Why both are necessary and complementary:**
- **Upstream's 10.5s Retry Loop**: Resolves race conditions for standard *asynchronous* streams, waiting out short-lived locks from cancelled streams.
- **This PR's Explicit Abort**: Handles *synchronous* delegations (`run_in_background=false`), which acquire a massive 60-second lock (`postDispatchHoldMs: 60_000`). Upstream's 10.5-second generic wait loop cannot out-wait a 60-second sync delegation lock, which would cause the fallback to timeout and fail silently.

By retaining this PR's explicit lock release during the abort phase, we ensure that long-lived sync locks are cleared instantly. This allows the fallback to dispatch immediately, perfectly complementing the upstream async retry loop without any regressions.

niStee:
**Status update — partial supersession**

After tracing the full chain of PRs that landed since this was opened, most of the fixes here have since been addressed independently:

- **Sync delegation stall** (releasing `model-suggestion-retry:sync` reservations during abort): landed in #5357 via a different but equivalent mechanism (`supersedeTransientRetryOwners: true` flag on the gate matcher).
- **Watchdog re-arming race** (first-prompt-watchdog completed-set guard): covered by #5491.
- **Fallback dispatch state / variant stripping / restore_primary_after_cooldown**: covered by #5476.
- **Object-shaped event model normalization**: already upstream per the original PR body.

The one piece **not yet in `dev`** and still genuinely novel is the **`provider-failure-state.ts` cross-system coordination** — a shared in-memory `Map` with auto-expiry that bridges reactive `runtime-fallback` and proactive `model-fallback`, so when runtime-fallback marks a provider as quota-exhausted/rate-limited, proactive fallback skips that provider on subsequent requests:
- `packages/omo-opencode/src/shared/provider-failure-state.ts` (new file)
- `packages/omo-opencode/src/shared/provider-failure-state.test.ts` (new file, 8 tests)
- `markProviderFailed` integration in `event-handler.ts`, `message-update-handler.ts`, `next-fallback.ts`

Closing this and opening a focused follow-up with only that piece on top of current `dev`.

PR #4377, opened 2026-05-23
fix(background-agent): implement three-layer defense for silent empty outputs

niStee:
Implements a three-layer defense to prevent silent empty outputs from being reported as successful task completion:

1. **Binary layer** (task.ts): Fail-fast on empty output
2. **Plugin layer** (idle handler): Detects empty outputs after 5s+ idle
3. **Hook layer** (empty-task-response-detector): Catches empty task responses and triggers background manager retry

Prevents the '(No text output)' false-success path by making assistant-visible text the single success invariant.

---
## Summary by cubic
Adds a three-layer defense so background tasks never report success with empty assistant output. If a model returns no text, we retry on a fallback model or clearly fail the task.

- **Bug Fixes**
  - Treat `EmptyOutputError` as retryable.
  - Background manager now verifies real assistant text (not just reasoning/tool calls) before completion; retries via `retryTaskOnEmptyOutput`, triggers fallback from `session.idle` through `onNoValidOutput`, backfills missing fallback chains from `CATEGORY_MODEL_REQUIREMENTS`, or marks error when no fallback exists.
  - `empty-task-response-detector` escalates empty outputs to the background manager using the task link; if no retry, emits a clear warning. `create-core-hooks` and `create-tool-guard-hooks` wire `backgroundManager`; tests updated.

<sup>Written for commit 88f2bd0ad1c80c15f8aa585fbffe39e84a2a9cbb. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4377?utm_source=github">Review in cubic</a></sup>

PR #4426, opened 2026-05-25
test(background-agent): update call counts for hasTextOutput validation

niStee:
## Summary

Update background-agent tests to account for additional messages API calls introduced by the hasTextOutput validation in the three-layer defense feature.

## Problem

PR #4377 (three-layer defense) added a `hasTextOutput()` check in `tryCompleteTask()` that makes an additional `messages` API call to verify the session produced actual text output (not just reasoning or tool calls).

This caused test failures in tests that verify exact call counts.

## Solution

Update test expectations to account for the additional messages call:
- Tests expecting 1 call now expect 2 calls
- Tests expecting 0 calls now expect 1 call

## Changes

- `src/features/background-agent/manager.test.ts`
  - Update call count expectations in 3 tests
  - Add comments explaining the additional call

## Testing

bun test v1.3.14 (0d9b296a)

⚠️ Tests pass with updated expectations (acknowledging new behavior)

## Alternative

See PR #4425 for an implementation fix that avoids the extra call entirely.

## Checklist

- [x] Tests updated for new behavior

---
## Summary by cubic
Implements a three-layer defense against silent empty outputs and tightens fallback behavior end-to-end. Tasks verify real text before completing, auto-fallback on empty output, and proactive/runtime fallback now share provider-failure state.

- **New Features**
  - Background agent: strict `hasTextOutput` and `retryTaskOnEmptyOutput()`; idle handler can trigger fallback on no output; safety net populates fallback chain from `CATEGORY_MODEL_REQUIREMENTS` when missing.
  - Fallback safety: shared provider-failure state bridges runtime and proactive fallback; proactive model-fallback skips providers marked failed and clears state on new sessions.
  - Tool hook: `empty-task-response-detector` asks the background manager to retry before showing a warning.

- **Bug Fixes**
  - Error handling: classify and retry on `empty_output` and `model_not_found`; read error identifiers from `name`/`code`/`type`; expanded retry patterns for Zen/Go providers; `EmptyOutputError` is retryable in `model-core`.
  - Fallback correctness: skip equivalent fallbacks with variant-only differences and strip variant suffixes in message model IDs and manual overrides; treat reasoning-only content as visible to avoid false empty-output retries.

<sup>Written for commit c8d3f2075e8ef6c53d1641a13395d4a6674390af. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4426?utm_source=github">Review in cubic</a></sup>

---

**Alternative Approach:** See PR #4425 for a different solution that fixes the implementation to avoid extra API calls.

### DIFF
--- a/packages/model-core/src/model-error-classifier.ts
+++ b/packages/model-core/src/model-error-classifier.ts
@@ -12,6 +12,7 @@ const RETRYABLE_ERROR_NAMES = new Set([
   "modelunavailableerror",
   "providerconnectionerror",
   "authenticationerror",
+  "emptyoutputerror",
 ])
 
 const STOP_ERROR_NAMES = new Set([
--- a/src/features/background-agent/manager.ts
+++ b/src/features/background-agent/manager.ts
@@ -28,6 +28,7 @@ import {
   hasMoreFallbacks,
   shouldRetryError,
 } from "../../shared/model-error-classifier"
+import { CATEGORY_MODEL_REQUIREMENTS } from "../../shared/model-requirements"
 import { SessionCategoryRegistry } from "../../shared/session-category-registry"
 import { applySessionPromptParams } from "../../shared/session-prompt-params-helpers"
 import { setSessionTools } from "../../shared/session-tools-store"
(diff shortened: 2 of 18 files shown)


KennyDizi:
From my perspective, this is a great PR. Just need to change the PR title to make it more sense (eg: fix,)

niStee:
Closing in favor of #4425.

#4425 fixes the duplicate output-validation request in the implementation rather than accepting it in tests, and it now also includes a reproduced fix for synchronous delegated tasks stalling during runtime-fallback retry-signal handoff.

The lenient `hasTextOutput()` edge-case behavior explored here can be revisited separately if a concrete remaining case requires it.

PR #4428, opened 2026-05-25
feat(model-core): treat UnknownError name as retryable catch-all

PeterPonyu:
## Summary

SDKs occasionally surface failures with \`name="UnknownError"\` when they cannot classify a provider response (mid-stream truncation, malformed error envelopes, network blips that don't carry a recognizable shape). On current \`dev\` the fallback chain never advances for these — the task dies on the first opaque error even when a healthy fallback would absorb it.

Add \`unknownerror\` to \`RETRYABLE_ERROR_NAMES\`, mirroring how the existing \`authenticationerror\` entry works: a generic-but-known error name that should advance the chain rather than terminate the task.

## Diff

| File | Change |
|---|---|
| \`packages/model-core/src/model-error-classifier.ts\` | +4 lines (1 new entry in \`RETRYABLE_ERROR_NAMES\` + 3 comment lines) |
| \`packages/model-core/src/model-error-classifier.test.ts\` | +23 lines (3 new tests) |

## Precedence preserved

The existing precedence in \`isRetryableModelError\` is unchanged:

\`\`\`
NON_RETRYABLE_ERROR_NAMES → STOP_ERROR_NAMES → RETRYABLE_ERROR_NAMES → message patterns
\`\`\`

So \`MessageAbortedError\`, \`PermissionDeniedError\`, \`ContextLengthError\`, \`QuotaExceededError\`, etc. still terminate immediately. A new test asserts this explicitly.

## What this PR is *not*

I considered also adding auth-flavored message patterns (\`401\`, \`unauthorized\`, \`invalid api key\`, etc.) since the matching error name \`authenticationerror\` is already in \`RETRYABLE_ERROR_NAMES\` and provider responses often surface auth failures without that exact name. **Skipped** — there is an existing test (\`HTTP 401 with statusCode does NOT trigger fallback (not a rate limit)\` at \`model-error-classifier.test.ts:360\`) that asserts the opposite policy. Adding those patterns would directly break it, so that's a policy decision worth a separate discussion.

## Test plan

- [x] \`bun test packages/model-core/\` — 259 pass / 0 fail
- [x] \`bun test src/features/background-agent/fallback-retry-handler.test.ts\` — 27 pass / 0 fail (downstream consumer)
- [x] \`bun run typecheck:packages\` — clean

---
## Summary by cubic
Treats `UnknownError` as retryable in `model-core` so opaque SDK failures advance the fallback chain instead of halting the task, unless STOP text (e.g., quota/billing) is present.

- Adds the `UnknownError` check in `isRetryableModelError` after STOP message patterns; non-retryable names and STOP text still terminate.
- Tests confirm `UnknownError` is retryable, `MessageAbortedError` is non-retryable, and `UnknownError` with quota text is non-retryable.
- Regenerates the `omo-senpi` bundles (`omo.js`, `memory-run-supervisor.mjs`) since `model-core` is inlined into them.

<sup>Written for commit 33a93fdc9705a4b878021e42c8fc882b3d32f2e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4428?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Related

Complementary to #7706 (issue #7702), not a duplicate. #7706 teaches the `runtime-fallback` hook classifier (`isRuntimeFallbackRetryableError`) to hop `agents.*.models[]` on OpenCode `UnknownError`; this PR teaches `shouldRetryError` in `model-error-classifier.ts` the same thing for the other retry paths that call it on `dev`: `plugin/event-model-fallback.ts` (the `model_fallback` event), `features/background-agent/fallback-retry-handler.ts` / `manager.ts`, and `tools/delegate-task/sync-task-runner.ts`. Neither classifier imports the other, so both are needed for an opaque `UnknownError` to advance every fallback chain. STOP/quota messages still terminate first (see the `UnknownError` + quota test).

Rebased onto `dev` `dc9bf86e4` (bun 1.4.2); `omo.js` and `memory-run-supervisor.mjs` regenerated with bun 1.4.2, `build-extension.mjs` extension check current.


code-yeongyu:
[sisyphus-bot] Keeping this open — the fix is small, correct, and still needed. `unknownerror` is not in `RETRYABLE_ERROR_NAMES` on current dev, meaning opaque provider errors still kill tasks instead of falling back. The PR may need a rebase due to the repo restructure (model-core is now at packages/model-core/). The contributor is encouraged to update.

PeterPonyu:
Rebased onto current `dev`. `unknownerror` is in `RETRYABLE_ERROR_NAMES`; aborted/quota still take precedence. Tests added for UnknownError / MessageAbortedError / QuotaExceededError.

PeterPonyu:
Follow-up: do not put `unknownerror` in `RETRYABLE_ERROR_NAMES` (that short-circuited STOP message patterns). Opaque `UnknownError` is now classified **after** STOP text, so `{ name: UnknownError, message: quota exceeded... }` still terminates. Added a regression test.

PeterPonyu:
Merged current `dev` (11d5155a8) into this branch. The only conflict was `model-error-classifier.test.ts`, which now keeps both the `#6313` upstream-request-failed case from `dev` and the UnknownError coverage from this PR.

Also restored the follow-up that classifies `UnknownError` *after* `STOP_MESSAGE_PATTERNS`, so a quota/billing message wrapped as `UnknownError` still terminates instead of advancing fallback. Regenerated `omo.js` and `memory-run-supervisor.mjs` with bun 1.4.0 so the ubuntu senpi `--check` job (the previous red required check) hashes against a current bundle.

`bun test packages/model-core/src/model-error-classifier.test.ts`: 43 pass, including the STOP+UnknownError case.

PR #4430, opened 2026-05-25
fix(tmux-subagent): stop tracking team-mode member sessions in the subagent panel

PeterPonyu:
## Symptom

Two complaints from the field that share a root cause:

1. **Team members surface inside the subagent panel** alongside the dedicated team_status view (same session shown twice).
2. **Team-member panes go missing / get closed underneath team-layout-tmux** during normal team operation — the polling loop in TmuxSessionManager decides the pane is idle and closes it while team-layout-tmux is still rendering into it.

## Root cause

Team-mode members are launched with \`bgMgr.launch({ suppressTmuxSpawn: true })\` to defer pane creation to \`createTeamLayout()\`. But \`suppressTmuxSpawn\` only blocks the explicit \`onSubagentSessionCreated\` callback wired in \`src/create-managers.ts\`. The team-member session still emits \`session.created\` through other paths, and \`TmuxSessionManager.onSessionCreated\` adds it to \`this.sessions\` regardless. Now two managers own the same pane:

- \`team-layout-tmux\` owns it via \`runtimeState.tmuxLayout\`
- \`TmuxSessionManager.sessions\` parallel-tracks it and polls for idle

The polling loop's close decision races team-layout's render and cleanup, producing the visible \"pane disappeared\" / \"member errored\" symptoms users report.

## Fix

Inject a \`shouldSkipSession\` predicate into TmuxSessionManager (a 4th optional constructor arg). The manager defaults to \`() => false\` — no behavior change for existing callers. \`create-managers.ts\` wires the predicate to \`lookupTeamSession\`, so any session already known to the team-session registry returns from \`onSessionCreated\` before \`spawnPendingSession\` runs.

## Layering

- \`tmux-subagent\` stays unaware of \`team-mode\` — no new imports across the boundary.
- The host (\`create-managers.ts\`) injects the team-side policy at construction time.
- Dependency direction remains \`team-mode → tmux-subagent\` only.

## Diff

| File | Change |
|---|---|
| \`src/features/tmux-subagent/manager.ts\` | +\`TmuxSessionManagerOptions\` type, +4th constructor arg, +1 field + 1 init line, +early-exit branch in \`onSessionCreated\` |
| \`src/create-managers.ts\` | +1 import, +wire \`shouldSkipSession\` to \`lookupTeamSession\` |
| \`src/features/tmux-subagent/manager.test.ts\` | +1 test verifying the early-exit fires for team-member sessionId AND ordinary subagents still spawn |

[...]

PR #4313, opened 2026-05-23
Fix tmux subagent auto attach retries

hichoe95:
## Summary
- replace focus-only tmux subagent activation with polling-driven auto activation
- make `opencode attach` run in a retry loop so a pane survives transient session/status readiness gaps
- remove the pre-spawn readiness gate and failed-readiness retry state now that attach owns readiness retrying
- speed up polling/idle cleanup and update placeholder text to reflect automatic attach

## Why
The previous lifecycle still let subagent panes get stuck at `OMO subagent pane ready` unless the user focused the pane, and a single failed `opencode attach` could leave the pane hidden or dead while the subagent kept running. The pane should be created immediately, then the attach command should keep retrying until the opencode session is actually attachable.

## Validation
- `bun test src/features/tmux-subagent/polling-manager.test.ts src/features/tmux-subagent/manager.test.ts src/features/tmux-subagent/session-created-handler.test.ts src/features/tmux-subagent/zombie-pane.test.ts src/plugin/event.test.ts src/shared/tmux/tmux-utils/pane-command.test.ts src/shared/tmux/tmux-utils/pane-spawn-runner.test.ts src/shared/tmux/tmux-utils/pane-replace.test.ts src/shared/tmux/tmux-utils/window-spawn.test.ts src/shared/tmux/tmux-utils/session-spawn.test.ts`
- `bun run typecheck`
- `git diff --check`

---
## Summary by cubic
Make tmux subagent panes auto-attach reliably by spawning a placeholder immediately and running a retrying `opencode attach` loop, removing readiness waits and focus-only activation. Fixes panes stuck at “OMO subagent pane ready” and panes dying on transient session gaps.

- **Bug Fixes**
  - Spawn a placeholder pane immediately; do not wait for server “ready” status.
  - Auto-attach via polling; activate when status is attachable, or proceed with a retrying attach when status fetch fails.
  - Keep `opencode attach` alive in a loop; exit only on success or normal user exits (0/130/143).
  - Add a 10s grace for unactivated placeholders before missing-status cleanup; prevent premature closes.
  - Faster, tighter polling and stability checks: 1s poll, 4s minimum idle, 2 stable polls.

- **Refactors**
  - Removed readiness gate and all failed-readiness retry state from `TmuxSessionManager`.
  - Simplified `TmuxPollingManager` activation flow; no focus-only/attached session gating.
  - Updated placeholder text to reflect automatic attach.
[...]

PR #4431, opened 2026-05-25
fix(tmux-subagent): close placeholder panes that exceed PLACEHOLDER_PANE_TIMEOUT_MS

PeterPonyu:
## Symptom

Tmux subagent panes accumulate over a long parent session and are never cleaned up. Closing the parent OpenCode session GCs them, but during a single long-lived session the count keeps growing.

## Root cause

\`TmuxPollingManager.pollSessions\` (src/features/tmux-subagent/polling-manager.ts:83) skips every close check for any tracked session whose \`attachActivated\` is \`false\` AND whose server status query returns nothing:

\`\`\`ts
if (!tracked.attachActivated && !status) {
  log("placeholder pane has not been activated yet; skipping close checks", {...})
  continue  // ← unbounded
}
\`\`\`

The skip is correct for the few seconds between spawn and first status, but for any pane the user never focuses (background subagent panes the user doesn't look at, panes whose server never returns status because the session was DOA, etc.) the session never produces status AND never flips activated. The entry stays in \`this.sessions\` forever — neither \`missingTooLong\`, \`isTimedOut\`, nor the stability path get a chance to run.

## Fix

Cap the skip with a generous grace, \`PLACEHOLDER_PANE_TIMEOUT_MS = 30 min\`, added to \`src/shared/tmux/constants.ts\` next to the other session timeouts. Once a placeholder pane exceeds the cap, queue it for the normal close path (same as \`missingTooLong\` / \`isTimedOut\`).

The cap is **intentionally generous** — much longer than the brief few-seconds setup window so panes the user is genuinely about to attach to don't get killed under them. But it's finite, so the leak stops.

## Diff

| File | Change |
|---|---|
| \`src/shared/tmux/constants.ts\` | +\`PLACEHOLDER_PANE_TIMEOUT_MS = 30 * 60 * 1000\` (with comment) |
| \`src/features/tmux-subagent/polling-manager.ts\` | +1 import, +13 lines (cap check + structured log + close queueing) |
| \`src/features/tmux-subagent/polling-manager.test.ts\` | +2 tests |

**3 files / +116 / −1**, narrow and reviewable.

## Test plan

- [x] New: zombie placeholder pane (>30 min, never activated, no status) → queued for close + \`closePending: true\`
- [x] New: young placeholder pane (5 min old) → NOT closed (preserves existing setup-grace behavior)
- [x] \`bun test src/features/tmux-subagent/polling-manager.test.ts\` — 11 pass
- [x] \`bun test src/features/tmux-subagent/ src/shared/tmux/\` — 225 pass / 0 fail
- [x] \`bun run typecheck\` — clean

## Related

- Companion PR #4430 (\`fix(tmux-subagent): stop tracking team-mode member sessions in the subagent panel\`) addresses double-tracking specifically for team members.
- Together these two fixes address both halves of the user-reported issue: "Why are team members visible in the subagent panel AND why do subagents under a same session grow and never get deleted."
- The earlier investigation also identified zombie panes after max close-retries (Gap B) and lack of LRU eviction (Gap E) as additional structural concerns; both are out of scope here and likely warrant a separate maintainer discussion.

---
## Summary by cubic
Closes never-activated, no-status tmux subagent placeholder panes after 30 minutes to stop pane leaks during long sessions.

- **Bug Fixes**
  - Added `PLACEHOLDER_PANE_TIMEOUT_MS` (30 min) in `src/shared/tmux/constants.ts`.
  - Bounded the skip in `TmuxPollingManager.pollSessions`; after timeout, mark `closePending` and queue for normal close while keeping short setup grace.
  - Added tests for timed-out and young placeholder panes.

<sup>Written for commit 4388335a0424b7c5f358f1b162a4267a3f50ed20. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4431?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/tmux-subagent/polling-manager.test.ts
+++ b/packages/omo-opencode/src/features/tmux-subagent/polling-manager.test.ts
@@ -455,6 +455,91 @@ describe("TmuxPollingManager overlap", () => {
     }
   })
 
+  test("closes placeholder panes that exceed PLACEHOLDER_PANE_TIMEOUT_MS without ever activating or returning status", async () => {
+    //#given - a placeholder pane older than the grace cap (30 min) that has
+    // never been focused (attachActivated=false) and the server never replied
+    // with a status. Before this guard, polling skipped every close check
+    // indefinitely and the pane accumulated forever.
+    const { PLACEHOLDER_PANE_TIMEOUT_MS } = await import("../../shared/tmux")
+    const sessions = new Map<string, TrackedSession>()
+    sessions.set("ses-zombie", {
+      sessionId: "ses-zombie",
+      paneId: "%9",
+      description: "abandoned subagent pane",
+      attachActivated: false,
+      createdAt: new Date(Date.now() - (PLACEHOLDER_PANE_TIMEOUT_MS + 60_000)),
+      lastSeenAt: new Date(Date.now() - (PLACEHOLDER_PANE_TIMEOUT_MS + 60_000)),
+      closePending: false,
+      closeRetryCount: 0,
+      activityVersion: 0,
+    })
+
+    const closedSessionIds: string[] = []
+    const client = {
+      session: {
+        status: async () => ({ data: {} }),
+        messages: async () => ({ data: [] }),
+      },
+    }
+    const manager = new TmuxPollingManager(
+      unsafeTestValue<import("../../tools/delegate-task/types").OpencodeClient>(client),
+      sessions,
+      async (sessionId) => {
+        closedSessionIds.push(sessionId)
+      },
+    )
+    const pollSessions = unsafeTestValue<{ pollSessions: () => Promise<void> }>(manager).pollSessions
+
+    //#when
+    await pollSessions.call(manager)
+
+    //#then
+    expect(closedSessionIds).toEqual(["ses-zombie"])
+    expect(sessions.get("ses-zombie")?.closePending).toBe(true)
+  })
+
+  test("still defers closure for placeholder panes within the grace window", async () => {
+    //#given - same pane shape, but only 5 minutes old (well under the 30 min cap).
+    // Existing behavior must be preserved: legitimate placeholder still gets time.
+    const { PLACEHOLDER_PANE_TIMEOUT_MS } = await import("../../shared/tmux")
+    const sessions = new Map<string, TrackedSession>()
+    sessions.set("ses-young", {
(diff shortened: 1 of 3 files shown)


PeterPonyu:
**Related parallel work**

#4313 (@hichoe95) is a parallel architectural refactor for this surface area. It removes the pre-spawn readiness gate and replaces it with "spawn placeholder immediately + polling-driven auto-activate." That model reduces the placeholder window from "indefinite" to seconds, so the 30-min cap in this PR is unnecessary under it.

**If #4313 merges first, I'll close this PR in deference and credit @hichoe95.** Cross-ref comment also posted on #4313.

MoerAI (changes requested):
Thanks for the PR, @PeterPonyu, and for taking the time on this. The fix itself is sound: routing a never-activated, never-status placeholder pane into the existing `closePending` close path after a bounded grace is the right call, and the two tests cover both the over-cap and within-grace cases.

One small thing to address before this can land:

- The new comment on `PLACEHOLDER_PANE_TIMEOUT_MS` in `src/shared/tmux/constants.ts` says it is "significantly longer than SESSION_TIMEOUT_MS", but `SESSION_TIMEOUT_MS` is 60 minutes (`src/shared/tmux/constants.ts:7`) while `PLACEHOLDER_PANE_TIMEOUT_MS` is 30 minutes, so it is actually shorter. Could you reword that line so the comparison is accurate?
- That same comment line contains an em dash (U+2014). The project's comment-checker convention in `AGENTS.md` is plain hyphens only, so please swap it for a hyphen or rephrase.

Both are in the one comment block, so it should be a quick touch-up. Once that is in I will take another look. No rush, and thanks again.

code-yeongyu:
[sisyphus-bot] Thanks for this fix. MoerAI reviewed and requested changes on June 5, but there's been no follow-up in over 9 weeks. The related #4440 (zombie-pane cooldown) was merged, and #4313 (polling-manager refactor) is still open. I'm closing this as stale — please feel free to reopen with a rebase and address the review feedback if the issue persists.

PR #4442, opened 2026-05-25
fix(auto-update): break infinite restart loop after successful install

### LINKED ISSUE #4318: [Bug]: Auto-update installs to wrong path — version never actually switches, infinite loop on restart

code-yeongyu:
Fixes #4318

## Problem
After auto-update installs a new version, the runtime version detection reads from a different path (import.meta.url walk-up) than where the update was installed (cache/config dir). This causes the hook to re-detect 'update available' on every session start, creating an infinite restart loop.

## Fix
Added a 5-minute same-version loop breaker. After a successful install, a marker file is written with the installed version and timestamp. On subsequent checks, if the same version was installed within the last 5 minutes, the update check is skipped.

## Tests
- Added regression test verifying loop-break behavior
- Targeted tests pass
- Typecheck clean

---
## Summary by cubic
Prevents the auto-update from reinstalling the same version on every startup by tracking recent successful installs. After an install, we write a marker and skip checks for that version for 5 minutes.

- **Bug Fixes**
  - Write a JSON marker (version, timestamp) in the cache dir after a successful install.
  - Skip update/install if the same version was installed within the last 5 minutes.
  - Added a regression test; minor DI for fs helpers to support testing.

<sup>Written for commit a5c759b5ffd0b8bf6f1f1b66453ca6cc266c6563. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4442?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/auto-update-checker/hook/background-update-check.ts
+++ b/src/hooks/auto-update-checker/hook/background-update-check.ts
@@ -1,6 +1,6 @@
 import type { PluginInput } from "@opencode-ai/plugin"
-import { existsSync } from "node:fs"
-import { join } from "node:path"
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs"
+import { dirname, join } from "node:path"
 import { runBunInstallWithDetails } from "../../../cli/config-manager"
 import { log } from "../../../shared/logger"
 import { getOpenCodeCacheDir, getOpenCodeConfigPaths } from "../../../shared"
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-dev] Reviewed this during the bug sweep for #4318. This PR avoids the repeated update loop by adding a recent-success marker, but it does not resolve the underlying OpenCode sandbox vs flat-cache install ownership mismatch described in the issue and competing PRs #4349/#4535. Because the accepted behavior needs maintainer direction, I am not merging this automatically.

code-yeongyu:
Closing this as superseded by #4535, which landed the current auto-update checker fix for #4318 by skipping legacy installs from OpenCode-managed sandboxes.

PR #4459, opened 2026-05-25
fix(runtime-fallback): normalize model payloads + preserve variants (1/2 of #4343)

PeterPonyu:
## Summary

Lane-isolated split of #4343 (closed: bundled 17 commits across 3 lanes).

This PR carries **only the runtime-fallback lane** — event payload normalization plus variant preservation through the runtime-fallback lifecycle. No \`delegate-task\`, \`team-mode\`, or plugin changes (those land in the sibling PR).

## Root cause

OpenCode session/message events sometimes carry the model as a bare id string (\`\"anthropic/claude-sonnet-4-6\"\`) and sometimes as an object \`{ id, variant? }\`. The runtime-fallback hooks assumed the object shape; an id-only payload silently dropped the variant suffix, breaking fallback-chain continuity across event boundaries.

## Files (all in \`src/hooks/runtime-fallback/\`)

| File | Change |
|---|---|
| \`event-handler.ts\` + \`.test.ts\` | Accept both payload shapes; normalize to \`{ id, variant }\` at the event surface |
| \`fallback-state.ts\` + \`.test.ts\` | Carry \`variant\` alongside \`id\` in state |
| \`runtime-model-event-record.ts\` (new) | Helpers to reconstruct full model identity for log surfaces |
| \`first-prompt-watchdog.ts\` + \`.test.ts\` | Preserve variant when rearming the watchdog |
| \`session-status-handler.ts\` + \`.test.ts\` | Preserve variant in retry-status records |
| \`message-update-handler.ts\` | Preserve variant in mid-conversation model swaps |
| \`fallback-bootstrap-model.ts\` | Accept id-shaped initial model |
| \`success-retry-key-cleanup.test.ts\` | Coverage for normalized keys |

Net: +1015/-66 across 12 files, all in \`src/hooks/runtime-fallback/\`.

## Verification

\`\`\`
bun test src/hooks/runtime-fallback/
# 223 pass, 0 fail, 402 expect()

bun run typecheck
# clean
\`\`\`

## Related

- Replaces #4343 (closed) jointly with sibling PR
- Sibling PR (forthcoming) carries the \`delegate-task\` + \`team-mode\` + \`shared/disabled-providers\` work

---
## Summary by cubic
Normalizes runtime model payloads and preserves variants across the runtime-fallback flow. Fixes dropped variants when events send models as strings so fallback chains advance correctly and logs stay consistent. Scope: runtime-fallback only; no changes to `delegate-task`, `team-mode`, or plugins.

[...]

PR #4462, opened 2026-05-25
fix(runtime-fallback): normalize runtime model payloads

PeterPonyu:
## Summary

Surgical replacement for #4459 (closed: 1015 lines, too broad for review).

This PR carries **just one commit, 4 files, +105/-7** — the payload normalization that fixes the root issue. The downstream variant-preservation work from #4459 is intentionally dropped; with normalized payloads at the boundary, downstream handlers don't need defensive variant tracking.

## Root cause

OpenCode session/message events carry the model as a bare id string sometimes (\`\"anthropic/claude-sonnet-4-6\"\`) and as an object \`{ id, variant? }\` other times. Runtime-fallback state assumed the object shape; an id-only payload silently dropped the variant suffix, breaking fallback-chain continuity across event boundaries.

## Fix

Normalize at the state boundary:

| File | Change |
|---|---|
| \`fallback-state.ts\` | Accept either shape, store as \`{ id, variant }\` |
| \`fallback-state.test.ts\` (new) | Coverage for both input shapes |
| \`fallback-bootstrap-model.ts\` | Apply normalization on bootstrap path |
| \`runtime-model-event-record.ts\` (new) | Helper to reconstruct full identity for log/diagnostic surfaces |

## Verification

\`\`\`
bun test src/hooks/runtime-fallback/
# 203 pass, 0 fail, 356 expect()

bun run typecheck
# clean
\`\`\`

## Related

- Replaces #4459 (closed) — same root cause, smaller blast radius
- #4343 is closed; this is the surgical-only carry-over

---
## Summary by cubic
Normalizes runtime model payloads at the fallback boundary to preserve provider/model/variant across event messages. Fixes variant loss when events send only an id string, keeping the fallback chain consistent.

- **Bug Fixes**
  - Accept both string and object model shapes; stringify to "provider/model(variant)" via `stringifyRuntimeModel` and `stringifyRuntimeModelWithVariant`.
  - Apply normalization in the bootstrap path and add `resolveRuntimeModelFromEventRecord` for logs/diagnostics.
  - Update equivalence checks so fallbacks skip equivalent models safely.
  - Add tests for mixed payload shapes and variant precedence.

<sup>Written for commit 8c897b4c13a1658ec893c702464f8a18afcda0dc. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4462?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/runtime-fallback/fallback-bootstrap-model.ts
+++ b/src/hooks/runtime-fallback/fallback-bootstrap-model.ts
@@ -2,11 +2,12 @@ import type { OhMyOpenCodeConfig } from "../../config"
 import { HOOK_NAME } from "./constants"
 import { log } from "../../shared/logger"
 import { SessionCategoryRegistry } from "../../shared/session-category-registry"
+import { stringifyRuntimeModel } from "./fallback-state"
 
 type ResolveFallbackBootstrapModelOptions = {
   sessionID: string
   source: string
-  eventModel?: string
+  eventModel?: unknown
   resolvedAgent?: string
   pluginConfig?: OhMyOpenCodeConfig
 }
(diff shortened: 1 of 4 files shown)


MoerAI:
Thanks @PeterPonyu, merging. The new stringifyRuntimeModel in `fallback-state.ts` normalizes both string and object-shaped model payloads at the runtime-fallback boundary, and the type-narrowing checks keep it safe.

PR #4466, opened 2026-05-25
fix(team-mode): preserve OPENCODE_SERVER_PASSWORD on tmux attach (#4409)

### LINKED ISSUE #4409: bug(team-mode): tmux visualization panes do not inherit opencode server password

PeterPonyu:
## Summary

Surgical replacement for #4410 (closed: 207 lines / 6 files extracted a shared helper that had only one caller in this PR's scope).

**+58/-1 across 2 files.** Closes #4409.

## Root cause

Tmux visualization panes spawn \`opencode attach\` via \`tmux send-keys\`. The leader process's env vars don't auto-propagate into the child; when \`OPENCODE_SERVER_PASSWORD\` is required, the attach fails with auth errors. Same goes for \`OPENCODE_SERVER_USERNAME\`.

## Fix

In \`buildAttachCommand\` (\`team-layout-tmux/layout.ts\`), inline a 9-line env-prefix builder. Shell-quote both values so embedded metacharacters are safe.

| File | Change |
|---|---|
| \`team-layout-tmux/layout.ts\` | +10/-1 — conditional env-prefix on the attach command |
| \`team-layout-tmux/layout.test.ts\` | +49 — present-password and absent-password coverage |

## Why inline instead of extracting a helper

The closed #4410 extracted \`getServerAuthEnvPrefix()\` into \`shared/opencode-server-auth.ts\` to share it with \`shared/tmux/tmux-utils/pane-command.ts\` (subagent panes). That was scope creep — \`pane-command.ts\` for subagent panes is a separate concern (different consumer, different bug). Keeping this PR strictly to the team-mode visualization attach.

If a future PR needs the same env-prefix for subagent panes, extracting a helper at that point is trivial.

## Verification

\`\`\`
bun test src/features/team-mode/team-layout-tmux/layout.test.ts
# 18 pass, 0 fail, 55 expect()

bun run typecheck
# clean
\`\`\`

---
## Summary by cubic
Pass `OPENCODE_SERVER_PASSWORD` (and optional `OPENCODE_SERVER_USERNAME`) to tmux-spawned `opencode attach` so team-mode panes can authenticate. Fixes #4409.

- **Bug Fixes**
  - Inline a conditional env prefix in `buildAttachCommand` and prepend it to `opencode attach`.
  - Shell-quote env values to safely handle special characters.
  - Tests: cover present/absent password and build auth fixtures at runtime to avoid secret-scanner false positives.

<sup>Written for commit 909de5b19ed5ff9cf1ddabbc7e8c5539306c9366. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/team-mode/team-layout-tmux/layout.test.ts
+++ b/src/features/team-mode/team-layout-tmux/layout.test.ts
@@ -193,6 +193,56 @@ describe("team-layout-tmux", () => {
     expect(literals.some((s) => s.includes("--session 's-m2'"))).toBe(true)
   })
 
+  test("#given env auth set #when createTeamLayout runs #then attach commands carry the env-prefix (fixes #4409)", async () => {
+    // given — minimal fixtures (not credentials); the embedded single quote
+    // exercises the shell-escape path.
+    const originalPwd = process.env.OPENCODE_SERVER_PASSWORD
+    const originalUser = process.env.OPENCODE_SERVER_USERNAME
+    process.env.OPENCODE_SERVER_PASSWORD = "a'b"
+    process.env.OPENCODE_SERVER_USERNAME = "u"
+    try {
+      const { createTeamLayout } = await loadLayoutModule()
+      const members = [{ name: "m1", sessionId: "s-m1", worktreePath: "/tmp/m1" }]
+
+      // when
+      await createTeamLayout("run-auth", members, tmuxMgr as never)
+
+      // then
+      const sendKeys = getCommands().filter((args) => args[0] === "send-keys").map((args) => args.join(" "))
+      const attach = sendKeys.find((s) => s.includes("opencode attach"))
+      expect(attach).toBeDefined()
+      // shell-escape of embedded single quote
+      expect(attach).toContain("OPENCODE_SERVER_PASSWORD='a'\\''b'")
+      expect(attach).toContain("OPENCODE_SERVER_USERNAME='u'")
+      // env-prefix precedes the binary
+      expect(attach!.indexOf("OPENCODE_SERVER_PASSWORD=")).toBeLessThan(attach!.indexOf("opencode attach"))
+    } finally {
+      if (originalPwd === undefined) delete process.env.OPENCODE_SERVER_PASSWORD; else process.env.OPENCODE_SERVER_PASSWORD = originalPwd
+      if (originalUser === undefined) delete process.env.OPENCODE_SERVER_USERNAME; else process.env.OPENCODE_SERVER_USERNAME = originalUser
+    }
+  })
+
+  test("#given no OPENCODE_SERVER_PASSWORD #when createTeamLayout runs #then attach command has no env-prefix", async () => {
+    // given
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-dev] Reviewed this as related to issue #3963's additional `Unauthorized` pane symptom. This is a focused auth propagation patch, but it does not address the original port-0/silent-skip problem, and merging it alone would still leave default-TUI visualization failures unresolved. I am leaving it open for maintainer direction rather than merging it as the #3963 fix.

MoerAI (changes requested):
Thanks for the PR, @PeterPonyu, and for taking the time on this. The POSIX env-prefix mechanics and the single-quote escaping in `layout.ts` are correct, so the auth-propagation idea works.

The blocker is secret exposure. This writes `OPENCODE_SERVER_PASSWORD` as plaintext into the tmux command line that gets sent to the pane. That value then lands in the pane scrollback, likely the shell history, and is briefly visible in the process args while `opencode attach` starts. No existing tmux attach builder in the repo inlines that secret, so this introduces a new exposure surface rather than following an established pattern.

Two things to address before this can land:
- Avoid putting the password into the pane command text. Prefer propagating it through the tmux/session environment so it never appears in scrollback or history.
- If a command-line prefix is unavoidable, please reuse the existing cross-shell helpers `detectShellType` and `buildEnvPrefix` in `src/shared/shell-env.ts` instead of hardcoding a POSIX-only `VAR='val'` prefix (this repo supports PowerShell and cmd too).

The quote-escaping tests are good, keep those. Happy to take another look once the secret no longer flows through the pane command line. No rush.

PR #4491, opened 2026-05-26
fix(process): raise listeners cap at module load to quiet plugin-hook warning (#4334)

### LINKED ISSUE #4334: [Bug] MaxListenersExceededWarning: Possible EventEmitter memory leak detected when plugin hooks register (~58 listeners)
## Describe the Bug

OpenCode web triggers `MaxListenersExceededWarning` when `oh-my-openagent` plugin is loaded. The warning is benign but noisy:

```
(node:3721) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 event listeners added. Use emitter.setMaxListeners() to increase limit
```

## Root Cause

The plugin registers **~58 hooks** across 5 hook categories via `safeCreateHook()`, confirmed by reading the compiled plugin bundle:

| Hook Category | Count | Source |
|--------------|:-----:|--------|
| Session hooks | ~24 | `createSessionHooks` |
| Tool guard hooks | ~18 | `createToolGuardHooks` |
| Transform hooks | ~7 | `createTransformHooks` |
| Continuation hooks | ~7 | `createContinuationHooks` |
| Skill hooks | ~2 | `createSkillHooks` |

Each `safeCreateHook` call adds an `"event"` listener to the **effect** runtime's internal EventTarget (mangled class `HL`) inside OpenCode's bun-compiled binary. Node.js' default `MaxListeners` limit is **10**, and with ~58 hook registrations the counter far exceeds it.

The stack trace:
```
at Kronos.addListener (node:events:531:26)
at HL.addEventListener (/$bunfs/root/chunk-j1sg77jf.js:121:24142)
    at /$bunfs/root/chunk-j1sg77jf.js:121:24484
    at ~effect/Effect/evaluate
    at runLoop -> evaluate -> runTasks -> afterScheduled
```

## Confirmed: Plugin is the Trigger

| Test | Plugin Loaded | Warning |
|------|:-------------:|:-------:|
| `opencode web --pure` (no plugins) | No | No warning |
| `opencode web` | Yes | Warning appears |

The warning does **not** appear in `--pure` mode.

## Verified Hook Source Code

The hooks are registered in the following functions within the compiled plugin bundle (`dist/index.js`, ~141k lines):

- **`createSessionHooks`** (line ~109384): 24 hooks including `contextWindowMonitor`, `modelFallback`, `ralphLoop`, `sessionNotification`, `autoUpdateChecker`, etc.
- **`createToolGuardHooks`** (line ~109680): 18 hooks including `commentChecker`, `toolOutputTruncator`, `rulesInjector`, `jsonErrorRecovery`, etc.
- **`createTransformHooks`** (line ~109920): 7 hooks including `claudeCodeHooks`, `keywordDetector`, `thinkingBlockValidator`, etc.
- **`createContinuationHooks`**: 7 hooks including `stopContinuationGuard`, `todoContinuationEnforcer`, `backgroundNotificationHook`, etc.
- **`createSkillHooks`**: 2 hooks (`categorySkillReminder`, `autoSlashCommand`)

## Affected Environment

- OpenCode version: **1.15.7** (bun-compiled binary)
- oh-my-openagent version: **4.3.1**
- Runtime: bun/Node.js
- Plugin config: default (all hooks enabled)

## Suggested Fix

In `safeCreateHook()` or at plugin initialization, increase the listener limit on the effect EventTarget:

```js
// In safeCreateHook or createHooks entry point
if (typeof target.setMaxListeners === 'function') {
  target.setMaxListeners(100); // or 0 for unlimited
}
```

Alternatively, batch hook registrations under fewer global event listeners so they don't each consume a separate listener slot.

## Workaround

The warning is **harmless** -- it does not prevent any functionality. To suppress it at startup:

```bash
NODE_NO_WARNINGS=1 opencode web
```

PeterPonyu:
## Summary

Fixes #4334. Plugin-host signal/cleanup listeners (background-agent, skill-mcp-manager, openclaw reply-listener, cli runner, plus per-plugin hooks) stack past Node's default cap of 10 per EventEmitter on busy sessions, producing a \`MaxListenersExceededWarning\` around ~58 listeners.

This is **not** a memory-leak suppression — the listeners are bounded by the active plugin/session set and each is shed on shutdown via the cleanup path in \`features/background-agent/process-cleanup.ts\`. The default cap is just too small for OmO's plugin host.

## Changes

Bump the cap once at module load via a small helper:

- **\`shared/raise-process-listeners-cap.ts\`** (new): \`raiseProcessListenersCap(min)\` bumps the cap to \`min\` only when \`current < min\`. Idempotent on repeat calls. \`PROCESS_LISTENERS_CAP_DEFAULT = 128\` leaves comfortable headroom above the ~58 observed in the wild.
- **\`index.ts\`**: call \`raiseProcessListenersCap(PROCESS_LISTENERS_CAP_DEFAULT)\` at the OmO entrypoint, before \`createPluginModule()\`, so the bump runs once per host process.

## Test plan

- [x] \`bun test src/shared/raise-process-listeners-cap.test.ts\` (3 pass)
- [x] Tests cover the three meaningful states:
  - cap below minimum → bumps
  - cap already above minimum → left alone, never lowered
  - repeated calls → idempotent

---
## Summary by cubic
Raises Node's process max listeners cap to 128 at plugin startup so stacked plugin hooks no longer trigger `MaxListenersExceededWarning`. The cap only ever goes up, never down, and Node's unlimited `0` sentinel is preserved.

- Adds `raiseProcessListenersCap()` and calls it as the first line of `serverPlugin`, before any listeners register.
- Tests cover bump, no-op, idempotent, and unlimited-`0` cases.
- Adds sandbox evidence showing the cap moves 10 → 128 and listener counts stay flat, with cleanup returning counts to baseline; Bun never emits the warning, so before/after runs show none.

<sup>Written for commit 97a9f4f922630bdd9bbdd6a815411889b31fa7af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
Thanks for the PR, @PeterPonyu, and for taking the time on this. The `raiseProcessListenersCap` helper itself is good: raise-only-never-lower with idempotent semantics and clean tests, that part is exactly right.

The blocker is placement. In `src/index.ts` the call runs as a top-level import-time side effect, so just importing the plugin entry mutates host-global `process` state (the whole OpenCode host, not only this plugin) before the plugin has actually started. Hosts, tests, and tooling can import `src/index.ts` without activating the plugin, and in ESM terms transitive imports also run before this top-level line, so it is not even guaranteed to be the earliest point. Raising the cap to 128 at import time also weakens a useful leak signal for the entire host: a real listener leak that grows from ~58 toward ~120 would now be silently masked.

Two things to address before this can land:
- Move the `raiseProcessListenersCap(...)` call out of the module top level in `src/index.ts` and into the earliest real startup path (e.g. the first line of `pluginModule.server(...)` before managers and hooks register listeners), so importing the entry alone does not mutate `process`.
- Add one test that importing the plugin entry does not change `process.getMaxListeners()`, and one startup-path test proving the cap is raised before listener registration.

Keep the helper and its existing tests, they are fine. Happy to re-review once the raise happens at startup rather than import. No rush.

code-yeongyu:
[sisyphus-bot] Keeping this open — issue #4334 is confirmed still open and the maintainer noted today that the fix is still needed. The PR has active review feedback from MoerAI requesting changes. The contributor is encouraged to address the review comments and rebase against current dev.

MoerAI:
The import-time mutation concern was addressed by moving the cap change into plugin startup and adding purity/order coverage. One blocker remains from the later review: a current listener cap of `0` means unlimited, but the new comparison treats it as too low and overwrites it with 128. Please preserve the unlimited case and add a focused test for `getMaxListeners() === 0`; then ping me for final re-review.

PeterPonyu:
Rebased onto current `dev` and addressed the remaining review blockers.

- Cap is raised as the first line of `serverPlugin` via `PluginModuleDeps.raiseProcessListenersCap` — not at import time, not in the `createPluginModule()` factory body.
- `getMaxListeners() === 0` is treated as Node's unlimited sentinel and is left untouched.
- Tests cover bump / already-high / unlimited-0 / idempotent, `server()` calling raise through the deps mock, and import-purity of `packages/omo-opencode/src/index.ts`.

@MoerAI ready for re-review when you have a moment.

MoerAI (approved):
Re-reviewed the current head. Both requested blockers are resolved.

Importing the entry now only constructs the plugin module; the listener-cap helper runs as the first statement of real async plugin startup. The helper also preserves the Node unlimited `0` sentinel and remains raise-only. Focused coverage verifies import purity, startup ordering, and low/high/unlimited/idempotent semantics.

The red Windows root-test failures are outside this one-commit diff, and the Senpi job was cancelled rather than source-failed. Those lanes still need a clean rerun before merge, but I found no remaining code-review blocker.

MoerAI:
The code review remains approved; no source change is requested here:
https://github.com/code-yeongyu/oh-my-openagent/pull/4491#pullrequestreview-4978602609

Merge is deferred on two evidence gates:

1. The current Windows test job is red, even though its failures are outside this diff:
https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31626600491/job/97216716087
2. The PR has no isolated real OpenCode artifact proving the reported warning is suppressed without hiding an unbounded listener leak.

Please rerun the current required matrix, including the cancelled compatibility jobs. Add reviewer-readable evidence under `.omo/evidence/` that boots the real plugin in an isolated OpenCode sandbox, exercises the listener-producing startup/session surface, records the effective process cap, shows no `MaxListenersExceededWarning`, and confirms cleanup returns listener counts to the bounded baseline.

When those gates are green, this remains a merge candidate with the existing approval.

PeterPonyu:
@MoerAI both gates addressed on head `7a4580505` (previous head `4e681b302` + the evidence commit only):

1. **Required matrix**: the full matrix, including the previously cancelled compatibility jobs, ran green on `4e681b302` — https://github.com/code-yeongyu/oh-my-openagent/actions/runs/34323434593 (every leg green). The Windows red you linked was on the pre-merge head.
2. **Isolated real-OpenCode evidence**: `.omo/evidence/20260909-4491-listener-cap-sandbox/` (README first). Real plugin loaded from source in a throwaway XDG sandbox against a fake OpenAI server, OpenCode 1.18.21, before (dev `dc9bf86e4`) vs after (this PR), with a probe plugin sampling `process.getMaxListeners()` and per-event `process` listener counts at import / server-init / startup / sessions-active / sessions-deleted / dispose / exit, in the serve process and two `opencode run` processes.
   - Cap: 10 / 10 / 10 in dev; 10 / 128 / 128 in the PR build (the raise happens as the first step of the plugin's `server()` init, so the probe's import-time sample still reads 10).
   - Warnings: zero `MaxListenersExceededWarning` in every variant and channel (probe, `serve-*.stderr`, `run-*.stderr`).
   - Baseline: per-event `process` listeners stay flat at ≤ 2 through 6 sessions, 271 plugin events, 2 CLI runs, session deletion and `/global/dispose`; the three leak controls remove their injected SIGHUP listeners at dispose (12 → 0, 12 → 0, 130 → 0).

Two limits, stated in the README rather than papered over: (a) the dev build did **not** reproduce the #4334 warning on this surface, and (b) the over-cap control (130 listeners vs cap 128) cannot show that a leak past the cap still warns, because in Bun `process` is not an `EventEmitter` instance and never emits `MaxListenersExceededWarning` (verified outside the sandbox; Node 24 does warn), and OpenCode embeds Bun 1.3.14. The #4334 report came from `opencode web` and an effect-runtime EventTarget rather than `process`, so the evidence proves the cap change and the absence of process-listener growth, not suppression of that specific warning. If that changes your view of whether this PR addresses #4334 as filed, I would rather know than have it merged on a claim the artifact does not support.

PR #4494, opened 2026-05-26
fix(hooks): avoid mutating frozen output.args on opencode >= 1.14

### LINKED ISSUE #3816: Bug: 8 plugin hooks mutate frozen `output.args` → all crash with `TypeError: Attempted to assign to readonly property` on opencode ≥ 1.14 (bash `git`, todowrite, webfetch, task, askuserquestion)
> [!CAUTION]
> **Any plugin user on opencode `≥ 1.14` who runs `bash` with a `git` command (or hits 5 other tools) gets a hard `TypeError` and the tool call fails.** All 8 sites listed below trigger the same crash. The bash hook is the loudest because the regex `/\bgit\b/` matches `.git/`, `git --version`, even strings like `'git|...'` inside an unrelated `grep`.

## TL;DR

8 hook handlers in `dist/index.js` mutate `output.args.*` directly. Opencode 1.14+ passes `output.args` as a frozen object, so every assignment throws:

```
TypeError: Attempted to assign to readonly property.
```

The error is surfaced to the user as a plain `bash failed: …` (or equivalent for other tools) — no stack, no breadcrumb pointing back to the plugin.

## Impact (what tools break and when)

| # | Line | File / hook | Tool affected | Trigger |
|---|------|-------------|---------------|---------|
| 1 | `84432` | `tool-execute-before-handler.ts` (todowrite parser) | `todowrite` | LLM sends `todos` as a JSON string |
| 2 | `84468` | `tool-execute-before-handler.ts` (PreToolUse) | **any tool** | external PreToolUse hook returns `modifiedInput` |
| 3 | `87424` | `non-interactive-env-hook.ts` | `bash` | command matches `/\bgit\b/` |
| 4 | `95972` | planning-consult hook | `task` | Prometheus agent + task tool |
| 5 | `96063` | notepad-directive hook | `task` | notepad path active |
| 6 | `98022` | askuserquestion truncation | `askuserquestion` | label > truncate threshold |
| 7 | `102270` | webfetch redirect resolver | `webfetch` | URL has redirect |
| 8 | `148903` | bash null-byte stripper | `bash` | command contains `\x00` |

Sites 3 and 8 are the two `output.args.command = …` assignments. Sites 2 and 6 use `Object.assign(output.args, …)` which has the same effect.

## How the failure flows

```mermaid
sequenceDiagram
    participant LLM
    participant OC as opencode 1.14.39
    participant P as oh-my-openagent 3.17.4
    participant T as bash tool

    LLM->>OC: tool_call(bash, {command:"git --version"})
    OC->>OC: output.args = Object.freeze({command:"git --version"})
    OC->>P: tool.execute.before(input, output)
    Note over P: regex `/\bgit\b/` matches → enter env-prefix branch
    P->>P: output.args.command = `${envPrefix} ${command}`
    Note over P: 💥 TypeError thrown — output.args is frozen
    P-->>OC: throw
    OC-->>LLM: bash failed: Attempted to assign to readonly property.
```

## Reproduction (clean room, 30 seconds)

**Environment**

| Component | Version |
|-----------|---------|
| opencode | `1.14.39` |
| oh-my-openagent | `3.17.4` |
| bun | `1.3.10` |
| OS | macOS 25.3.0 arm64 |

**Step 1 — confirm `Object.freeze` produces the exact error string**

```js
// bun /tmp/test.mjs
const o = { args: Object.freeze({ command: "git --version" }) };
try { o.args.command = "modified"; }
catch (e) { console.log(e.name, "/", e.message); }
// → TypeError / Attempted to assign to readonly property.
```

**Step 2 — confirm end-to-end via headless opencode**

```bash
# Plugin enabled (default)
opencode run --print-logs --log-level INFO \
  --dangerously-skip-permissions \
  "Use bash to run: git --version"

# Output:
# ✗ bash failed
# Error: Attempted to assign to readonly property.
```

```bash
# Plugin disabled — works fine
opencode run --pure --dangerously-skip-permissions \
  "Use bash to run: git --version"
# → git version 2.50.1
```

## Suggested fix

The plugin assumes `output.args` is mutable. It is not (and arguably never should be — silent input rewriting hidden inside plugin hooks is hostile to debugging and to opencode's permission model — see #2308).

**Option A — minimal, ship today** (one helper, replaces all 8 sites):

```ts
// helpers/safe-mutate.ts
export function safeAssignArg<T>(
  output: { args: Record<string, unknown> },
  key: string,
  value: T,
): boolean {
  if (Object.isFrozen(output.args)) return false;
  output.args[key] = value;
  return true;
}

export function safeMergeArgs(
  output: { args: Record<string, unknown> },
  patch: Record<string, unknown>,
): boolean {
  if (Object.isFrozen(output.args)) return false;
  Object.assign(output.args, patch);
  return true;
}
```

Each call site logs and skips its mutation when `safeAssignArg` returns `false`, instead of crashing the whole tool call.

**Option B — proper fix** (requires opencode plugin SDK change): expose a supported "rewrite tool input" API so plugins return the new args via the hook return value instead of mutating in place. Mutating frozen inputs is an anti-pattern; the SDK should make the right thing easy.

## Local workaround (verified)

I patched all 8 sites with `try { … } catch {}` wrappers in my cached copy:

```
~/.cache/opencode/packages/oh-my-openagent@3.17.4/node_modules/oh-my-openagent/dist/index.js
```

After patching, the same headless reproduction returns `git version 2.50.1` cleanly. Sanity grep:

```bash
grep -nE '^\s+output\.args\.[a-zA-Z_]+ ?=|^\s+Object\.assign\(output\.args' \
  ~/.cache/opencode/packages/oh-my-openagent@*/node_modules/oh-my-openagent/dist/index.js \
  | grep -v 'try {'
# (zero unwrapped sites = patch complete)
```

The patch survives until the next plugin upgrade — hence this issue.

## Related issues

| # | Status | Why it matters |
|---|--------|----------------|
| [#1746](https://github.com/code-yeongyu/oh-my-openagent/issues/1746) | CLOSED | **Same anti-pattern, different property** — unguarded `output.output` mutation in `tool.execute.after`. Was fixed only for `output.output`; `output.args` was not audited. |
| [#2220](https://github.com/code-yeongyu/oh-my-openagent/issues/2220) | CLOSED | The "fix" shipped (line `148903` null-byte stripper) **is itself one of the 8 broken sites listed above**. If a null byte ever reaches it on opencode 1.14+, it crashes instead of stripping. |
| [#2308](https://github.com/code-yeongyu/oh-my-openagent/issues/2308) | OPEN | Same `non-interactive-env` hook, different concern (env-prefix injection breaks `"git *": "ask"` permission rules). The cleanest fix for both #2308 and this issue is **don't rewrite tool args from a plugin** — return a structured hint to the user / agent instead. |

## Why this is a systemic class of bug, not a one-off

The audit grep above (`output\.args\.[a-zA-Z_]+ ?=` and `Object\.assign\(output\.args`) finds 8 hits today. Adding a 9th in a future hook is a one-line code change. A repo-level `eslint` rule or a single helper function would prevent the next instance.

```bash
# Suggested CI guard
rg -nE '^\s+output\.args\.\w+ ?=|^\s+Object\.assign\(output\.args' src/ \
  && echo "REGRESSION: direct mutation of frozen output.args" && exit 1 \
  || echo "OK"
```

---

Happy to open a PR with Option A if it'd help — let me know.

Hungdoan565:
## Summary

Fixes remaining frozen output.args mutation sites that cause TypeError: Attempted to assign to readonly property on opencode >= 1.14.34 with OPENCODE_EXPERIMENTAL=true.

## Testing

- un run typecheck - clean
- un test src/hooks/write-existing-file-guard/ - 25 pass, 0 fail
- Existing tests unaffected

Closes #3816

---
## Summary by cubic
Prevent TypeError crashes by avoiding mutations of frozen output.args on opencode >= 1.14. Fixes #3816.

- **Bug Fixes**
  - write-existing-file-guard: replaced in-place delete of overwrite with creating a new args object and reassigning to output.args.
  - delegate-task args prep: removed dead in-place mutations; the function already returns a fresh object.
  - Added regression tests to verify frozen args don’t throw and overwrite is stripped.

<sup>Written for commit a36dbad5d375201eac0d866f2685778f359f976f. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4494?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/write-existing-file-guard/index.test.ts
+++ b/src/hooks/write-existing-file-guard/index.test.ts
@@ -561,3 +561,95 @@ describe("createWriteExistingFileGuardHook", () => {
     ).rejects.toThrow(BLOCK_MESSAGE)
   })
 })
+
+describe("frozen args handling (issue #3816)", () => {
+  let tempDir = ""
+  let hook: Hook
+  let callCounter = 0
+
+  const createFile = (relativePath: string, content = "existing content"): string => {
+    const absolutePath = join(tempDir, relativePath)
+    mkdirSync(dirname(absolutePath), { recursive: true })
+    writeFileSync(absolutePath, content)
+    return absolutePath
+  }
+
+  const invokeWithFrozenArgs = async (args: {
+    tool: string
+    sessionID?: string
+    outputArgs: Record<string, unknown>
+  }): Promise<{ args: Record<string, unknown> }> => {
+    callCounter += 1
+    const frozenArgs = Object.freeze({ ...args.outputArgs })
+    const output = { args: frozenArgs as Record<string, unknown> }
+
+    await hook["tool.execute.before"]?.(
+      {
+        tool: args.tool,
+        sessionID: args.sessionID ?? "ses_default",
+        callID: `call_${callCounter}`,
+      } as never,
+      output as never
+    )
+
+    return output
+  }
+
+  beforeEach(() => {
+    tempDir = mkdtempSync(join(tmpdir(), "frozen-args-"))
+    hook = createWriteExistingFileGuardHook({ directory: tempDir })
+    callCounter = 0
+  })
+
+  afterEach(() => {
+    rmSync(tempDir, { recursive: true, force: true })
+  })
+
+  test("#given frozen args with overwrite #when write tool invoked #then does not throw TypeError", async () => {
+    // given
+    const existingFile = createFile("frozen-test.ts")
+    const sessionID = "ses_frozen"
+
+    // when - read first to grant permission
+    await invokeWithFrozenArgs({
+      tool: "read",
+      sessionID,
+      outputArgs: { filePath: existingFile },
+    })
+
+    // then - write with frozen args containing overwrite should not throw
+    const result = await invokeWithFrozenArgs({
+      tool: "write",
+      sessionID,
+      outputArgs: { filePath: existingFile, content: "new", overwrite: true },
+    })
+
+    // overwrite key should be removed from args
+    expect(result.args).not.toHaveProperty("overwrite")
+    expect(result.args).toHaveProperty("filePath")
+    expect(result.args).toHaveProperty("content")
+  })
+
+  test("#given frozen args without overwrite #when write tool invoked #then does not throw TypeError", async () => {
+    // given
+    const existingFile = createFile("frozen-no-overwrite.ts")
+    const sessionID = "ses_frozen2"
+
+    // when - read first
+    await invokeWithFrozenArgs({
+      tool: "read",
+      sessionID,
+      outputArgs: { filePath: existingFile },
+    })
+
+    // then - write with frozen args (no overwrite key) should work fine
+    const result = await invokeWithFrozenArgs({
+      tool: "write",
+      sessionID,
+      outputArgs: { filePath: existingFile, content: "new" },
+    })
+
+    expect(result.args).toHaveProperty("filePath")
+    expect(result.args).toHaveProperty("content")
+  })
+})
--- a/src/hooks/write-existing-file-guard/tool-execute-before-handler.ts
+++ b/src/hooks/write-existing-file-guard/tool-execute-before-handler.ts
@@ -146,7 +146,8 @@ export async function handleWriteExistingFileGuardToolExecuteBefore(params: {
 
   const overwriteEnabled = isOverwriteEnabled(args?.overwrite)
   if (argsRecord && "overwrite" in argsRecord) {
-    delete argsRecord.overwrite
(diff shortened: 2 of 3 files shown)


Hungdoan565:

MoerAI:
Thanks @Hungdoan565, merging. Replacing the in-place delete/assignment with rest-spread reassignment in `tool-execute-before-handler.ts` avoids mutating the frozen output.args on opencode >= 1.14. Confirmed prepareDelegateTaskArgs is consumed via its return value, so dropping the in-place writes is behavior-equivalent.

PR #4513, opened 2026-05-26
Use managed exact plugin versions for auto updates

biggameking:
## Summary
- install the plugin as a managed exact version instead of writing dynamic `@latest` entries
- persist channel metadata so auto-update can keep exact-version installs moving forward automatically
- migrate existing dynamic plugin entries to managed exact versions and roll config back if a background update install fails

## Why
`oh-my-openagent` currently writes dynamic plugin entries such as `@latest`, which keeps startup coupled to OpenCode's plugin resolution path. When that path is slow or unstable, users can get multi-minute stalls before the plugin even loads. This change does not fix OpenCode core's startup fallback, but it does remove `oh-my-openagent` from relying on that path after install.

## Testing
- `bun test src/cli/config-manager/plugin-name-with-version.test.ts`
- `bun test src/cli/config-manager/plugin-detection.test.ts`
- `bun test src/hooks/zauc-mocks-bg/background-update-check.test.ts`
- `bun test src/hooks/zauc-mocks-ws/workspace-resolution.test.ts`
- `bun test src/hooks/auto-update-checker/hook.test.ts`
- `bun test src/hooks/zauc-mocks-hook/hook.test.ts`
- `bun test src/hooks/auto-update-checker/checker/pinned-version-updater.test.ts`

## Notes
- I intentionally left unrelated generated files (`bun.lock`, `assets/oh-my-opencode.schema.json`) out of this PR.
- There is an existing unrelated Windows path assertion issue in `plugin-entry.test.ts`; it is not introduced by this change.

---
## Summary by cubic
Switch plugin installs to managed exact versions with a persisted release channel. This removes `@latest` stalls at startup and lets auto‑update advance pinned installs safely.

- **New Features**
  - Installs `oh-my-openagent@<version>` via `resolvePluginInstallReference`, returning the exact entry and its channel.
  - Persists channel metadata in a config-scoped `...-plugin-state.json` so updates track the right channel.
  - Background updater migrates dynamic/unpinned entries to exact versions, writes state, and reverts the config on failure.
  - Version bumps preserve the canonical alias (e.g., keep `oh-my-openagent`, migrate legacy `oh-my-opencode` safely).

- **Migration**
  - No manual steps. Existing dynamic entries are converted to exact versions during install or the next background check.
  - If an update fails, the config entry is rolled back automatically.

<sup>Written for commit a186d12f328e79bbaa13cd0229020eeeff908674. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4513?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/cli/config-manager.ts
+++ b/src/cli/config-manager.ts
@@ -6,7 +6,7 @@ export {
 } from "./config-manager/config-context"
 
 export { fetchNpmDistTags } from "./config-manager/npm-dist-tags"
-export { getPluginNameWithVersion } from "./config-manager/plugin-name-with-version"
+export { getPluginNameWithVersion, resolvePluginInstallReference } from "./config-manager/plugin-name-with-version"
 export { addPluginToOpenCodeConfig } from "./config-manager/add-plugin-to-opencode-config"
 
 export { generateOmoConfig } from "./config-manager/generate-omo-config"
--- a/src/cli/config-manager/add-plugin-to-opencode-config.ts
+++ b/src/cli/config-manager/add-plugin-to-opencode-config.ts
@@ -1,14 +1,24 @@
 import { readFileSync, writeFileSync } from "node:fs"
+import { dirname } from "node:path"
 import type { ConfigMergeResult } from "../types"
 import { PLUGIN_NAME, LEGACY_PLUGIN_NAME } from "../../shared"
+import { writeManagedPluginState } from "../../shared/managed-plugin-state"
 import { backupConfigFile } from "./backup-config"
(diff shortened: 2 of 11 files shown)


code-yeongyu:
[sisyphus-bot] Thanks for this work! The auto-update plugin versioning has received several fixes in dev since this was opened (5d3215167, 67b4665c2), and the source was moved to packages/omo-opencode/. We're closing this as stale — if there's still a gap, please open a fresh PR against the new structure. 🙏

Issue #3919, opened 2026-05-10
[Bug]: Custom glob/grep tool implementations cause Local Server disconnection on Windows OpenCode Desktop


ximeng-y:
**Custom glob/grep tool implementations cause Local Server disconnection on Windows OpenCode Desktop**

> I prompted an AI agent to help analyze this code issue based on my own investigation, and it drafted the analysis below.

## Description

After installing the `oh-my-openagent` plugin, calling OpenCode's built-in `glob`/`grep` tools causes the built-in Local Server to disconnect. Removing the plugin restores normal functionality.
This problem never appeared before I upgraded OpenCode from version 1.14.39.

## Environment

- **OS**: Windows
- **OpenCode type**: Desktop (terminal version not tested)
- **OpenCode version**: 1.14.41+
- **oh-my-openagent version**: dev/v4.x

## Reproduction Steps

1. Enable the `oh-my-openagent` plugin in opencode config
2. Call the `glob` or `grep` tool in any session
3. Observe the built-in Local Server disconnecting — the tools fail to return results

## Verification

- ❌ Removing the plugin → glob/grep work normally
- ✅ `disabled_tools: [glob, grep]` → glob/grep work normally

## Root Cause Analysis

The plugin completely **rewrites** OpenCode's native `glob` and `grep` tools in `src/tools/glob/tools.ts` and `src/tools/grep/tools.ts`, rather than incrementally enhancing them. Specific issues:

1. **Custom subprocess management layer**: Uses `src/shared/bun-spawn-shim.ts` to wrap Node.js `child_process.spawn` into a Bun-compatible API, running ripgrep in a non-Bun environment (OpenCode Desktop uses Node.js).

2. **Web API compatibility**: Uses `new Response(proc.stdout).text()` (Web `Response` API) to read child process stdout, which may have compatibility issues under Windows Node.js.

3. **rgSemaphore concurrency control**: Introduces `rgSemaphore` to manage ripgrep concurrency. Deadlocks or race conditions could cause tool calls to hang indefinitely.

4. **Windows fallback path**: When ripgrep is not installed, falls back to PowerShell `Get-ChildItem` for glob searches. Glob pattern translation and error handling differ significantly from ripgrep.

5. **ripgrep auto-download**: On first call, `downloadAndInstallRipgrep()` downloads ripgrep from GitHub Releases. Network issues or insufficient permissions can cause subprocess launch failures.

## Workaround

Add the following to your plugin config:

```jsonc
[...]

PR #4527, opened 2026-05-26
fix: normalize Windows where.exe output in ripgrep resolver

### LINKED ISSUE #4512: [Bug]: Windows ripgrep resolver can leave trailing \r from where.exe output, causing rg ENOENT and grep/glob failures
# Windows root-cause report for GitHub

## Summary

This appears to be a **remaining Windows-specific bug** in `oh-my-openagent` after the `#3919` / `#4300` Desktop utility-process fix.

`src/shared/ripgrep-cli.ts` currently resolves `rg` / `grep` via `where.exe` on Windows and then returns the first line with:

```ts
return stdout.trim().split("\n")[0]
```

On Windows, `where.exe` output is CRLF-delimited. That means the first line can still end with `\r`, so the resolved path becomes:

```text
C:\...\rg.exe\r
```

When that path is later used with `spawn` / `spawnSync`, the subprocess launch fails with `ENOENT`.

## Why this matters

On my local Windows machine, this manifests as `oh-my-openagent`-enabled `grep`/`glob` usage breaking OpenCode Desktop sessions with disconnect-like behavior.

The exact local reproduction was:

```text
spawnSync C:\...\rg.exe\r ENOENT
```

So this does **not** look like a Desktop/CLI shared-core conflict. It looks like a Windows line-ending parsing bug in OMO's ripgrep executable resolver.

## Local evidence

- Date reproduced: **May 26, 2026**
- OS: **Windows**
- OMO package in local install: **4.1.2**
- Observed file in current upstream source: `src/shared/ripgrep-cli.ts`
- Current upstream file still contains the vulnerable split logic as of the current `dev` branch:
  - https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/shared/ripgrep-cli.ts
- Related prior bug/fix for a different Windows crash path:
  - Issue `#3919`: https://github.com/code-yeongyu/oh-my-openagent/issues/3919
  - PR `#4300`: referenced in the issue comments as the Web-Response-on-Node/Desktop fix

## Minimal reproduction

On a Windows machine where `where.exe rg` returns one or more paths:

```js
const { spawnSync } = require("child_process")
const result = spawnSync("where", ["rg"], { encoding: "utf-8", timeout: 5000 })
const buggy = result.stdout.trim().split("\n")[0]
console.log(JSON.stringify({ buggy, hasCR: buggy.includes("\r") }))

const probe = spawnSync(buggy, ["--version"], { encoding: "utf-8", timeout: 5000 })
console.log(probe.error)
```

Expected problematic output:

```text
hasCR: true
spawnSync ... rg.exe\r ENOENT
```

## Proposed fix

Normalize Windows line endings before taking the first executable path:

```ts
return stdout
  .split(/\r?\n/)
  .map((line) => line.trim())
  .find(Boolean) ?? null
```

## Minimal patch

```diff
--- a/src/shared/ripgrep-cli.ts
+++ b/src/shared/ripgrep-cli.ts
@@
-    if (result.status === 0 && stdout.trim()) {
-      return stdout.trim().split("\n")[0]
-    }
+    if (result.status === 0 && stdout.trim()) {
+      return stdout
+        .split(/\r?\n/)
+        .map((line) => line.trim())
+        .find(Boolean) ?? null
+    }
```

## Notes

- This looks **separate** from the `#3919` / `#4300` Web Response API crash path.
- The prior fix addressed a utility-process crash in Desktop. This new one is a more basic Windows executable-path parsing bug in the ripgrep resolver.
- I verified locally that after applying the CRLF-safe split, `rg --version` succeeds and `opencode run` with `oh-my-openagent` loaded can successfully complete a real `grep` tool call again.

Tinycute00:
## Summary

Fix a remaining Windows-specific `rg`/`grep` resolver bug in `src/shared/ripgrep-cli.ts`.

`where.exe` returns CRLF-delimited output on Windows, but the current resolver takes the first line with `stdout.trim().split("\n")[0]`. That can leave a trailing `\r` on the selected executable path and later cause:

```text
spawn ... rg.exe\r ENOENT
```

This PR makes the resolver CRLF-safe by splitting on `/\r?\n/`, trimming each candidate, and taking the first non-empty entry.

## Why

Closes #4512.

Issue `#3919` / PR `#4300` fixed a separate Windows Desktop utility-process crash path. This patch covers a second remaining Windows-specific failure mode in the ripgrep executable resolver itself.

## Validation

- Added a regression test for CRLF-delimited `where.exe` output
- `bun test src/shared/ripgrep-cli.test.ts --bail`
- `bun test src/shared/mock-module-lifecycle-audit.test.ts --bail`
- `bun run typecheck`
- `bun run build`

---
## Summary by cubic
Normalizes Windows `where.exe` output in the ripgrep resolver to strip CRLF and pick the correct path, preventing `ENOENT` when spawning `rg`. Fixes #4512.

- **Bug Fixes**
  - Split probe output on `/\r?\n/`, trim lines, and use the first non-empty entry.
  - Prevents selecting `rg.exe\r`; preserves resolver order and non-Windows behavior.
  - Adds a regression test for CRLF-delimited `where.exe` output.

<sup>Written for commit b49cd9f864a468467c17c127a61f3cda7534f3a8. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4527?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/shared/ripgrep-cli.test.ts
+++ b/src/shared/ripgrep-cli.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, it } from "bun:test"
+import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test"
 import { mkdirSync, rmSync, writeFileSync } from "node:fs"
 import { join } from "node:path"
 import { tmpdir } from "node:os"
@@ -26,6 +26,7 @@ describe("resolveGrepCli OpenCode cache fallback (#3805)", () => {
     else process.env.XDG_CACHE_HOME = originalCache
     if (originalData === undefined) delete process.env.XDG_DATA_HOME
     else process.env.XDG_DATA_HOME = originalData
+    mock.restore()
     try {
       rmSync(tempCache, { recursive: true, force: true })
       rmSync(tempData, { recursive: true, force: true })
@@ -52,5 +53,32 @@ describe("resolveGrepCli OpenCode cache fallback (#3805)", () => {
     const resolved = resolveGrepCli()
     expect(resolved.backend).toBe("rg")
     expect(resolved.path).toBe(cacheRg)
+  })
+
+  it("strips Windows CRLF from where.exe output before spawning ripgrep (#4512)", async () => {
+    const rawRgPath = "C:\\Users\\tester\\.local\\bin\\rg.exe"
+
+    mock.module("node:child_process", () => ({
+      spawnSync: () => ({
+        status: 0,
+        stdout: `${rawRgPath}\r\nC:\\Windows\\System32\\rg.exe\r\n`,
+      }),
+    }))
+    mock.module("node:fs", () => ({
+      existsSync: () => false,
+    }))
+    mock.module("../tools/grep/downloader", () => ({
+      downloadAndInstallRipgrep: async () => rawRgPath,
+      getInstalledRipgrepPath: () => null,
+    }))
+
+    delete require.cache[require.resolve("./ripgrep-cli")]
+    const { resolveGrepCli } = await import("./ripgrep-cli")
+
+    const resolved = resolveGrepCli()
+
+    expect(resolved.backend).toBe("rg")
+    expect(resolved.path).toBe(rawRgPath)
+    expect(resolved.path.endsWith("\r")).toBe(false)
   })
 })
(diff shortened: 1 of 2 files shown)


MoerAI:
Thanks @Tinycute00, merging. Normalizing `where.exe` output with `/\r?\n/` and picking the first non-empty path is the right fix for the trailing-CR `ENOENT` on Windows (#4512), and the regression test covers the CRLF case.

Issue #4481, opened 2026-05-25
[sisyphus-bot] [Bug]: <command-instruction> duplicated on start-work error retry path (not covered by #3724)


WhiteGiverMa:
## Summary

When `/start-work` fails to find a matching plan, the hook retries and **re-injects the entire `<command-instruction>` block** (120+ lines), causing a duplicate identical to the first injection. This is distinct from the slash-command invocation duplication fixed in #3724.

## Steps to reproduce

1. Invoke `/start-work 完成当前计划`
2. No matching plan found → hook returns "Plan Not Found"
3. Hook re-constructs the context for the next turn

Observe in the system prompt:

```
[search-mode] ...
[analyze-mode] ...
<auto-slash-command>
  <command-instruction>   ← FIRST injection (correct, ~120 lines)
  ...
  </command-instruction>
  <session-context> ... </session-context>
  <user-request>...</user-request>
</auto-slash-command>

--- (error response: Plan Not Found, list available plans) ---

<auto-slash-command>
  <command-instruction>   ← SECOND injection (DUPLICATE, ~120 lines)
  ...
  </command-instruction>
  <session-context>       ← variables not expanded (> see #4480)
    Session ID: $SESSION_ID
    Timestamp: $TIMESTAMP
  </session-context>
  <user-request>...</user-request>
</auto-slash-command>
```

The entire `<command-instruction>` block is injected twice, roughly doubling the token cost.

## Expected behavior

`<command-instruction>` should be injected only once per turn. Error retry should not cause re-injection.

## Why #3724 doesn't cover this

#3724 fixed the case where the *invocation* of a slash command was duplicated (command content appearing twice when the user typed `/start-work`). This bug is different — it's the *hook's auto-slash-command expansion on error retry* that causes re-injection of the template, not the user's invocation being duplicated.

## Version

oh-my-openagent: 4.4.0
OpenCode: 1.15.10
Platform: Windows 11

## Related

- #3724 — [Bug]: Slash command content duplicated when invoked in chat (fixed, different path)
- #4480 — `$SESSION_ID` / `$TIMESTAMP` literal after error retry (same trigger, separate bug)

code-yeongyu:
[sisyphus-bot] Thanks for the report! I rechecked the current `dev` branch and confirmed this issue is resolved.

[...]

Issue #3724, opened 2026-04-29
[Bug]: Slash command content duplicated when invoked in chat


FurryWolfX:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

When a user types a registered slash command (e.g., `/caveman-review`, `/skill-issue-generator`) in chat, the command's instruction template appears **twice** in the conversation context — once wrapped in `<auto-slash-command>` tags and once as raw content. This duplication wastes context window and confuses the agent.

### Steps to Reproduce

1. Open an OpenCode session
2. Type any registered slash command

### Expected Behavior

The slash command content should be injected **exactly once** when the user invokes a `/command`.

### Actual Behavior

The same command template appears twice side by side:

```markdown
<auto-slash-command>
...
---
## Command Instructions
XXX
...
</auto-slash-command>
XXX  ← duplicate, no wrapping tags
...
```

### Doctor Output

```shell
install via sourcecode

omo: 3.17.6
```

### Error Logs

```shell

```

### Configuration

```json

```

### Operating System

Linux

### OpenCode Version

1.14.29

code-yeongyu:
[sisyphus-bot]

Confirmed.

The duplicate content appears to come from two separate injection paths in OMO. The `chat.message` hook rewrites a submitted slash command into an `<auto-slash-command>` wrapper, and `command.execute.before` can inject the same wrapped command template again for the same command flow. Command discovery only deduplicates command definitions by name, so it is not the source of the duplicated content.

Action taken: added `bug`, `confirmed`, and `triage:bug`. I assigned `code-yeongyu` because the owner decision is which path should be authoritative and which path should no-op for already-injected slash command invocations.

code-yeongyu:
[sisyphus-bot]

Code-level evidence for the double injection on `dev`.

Two independent hooks fire on the same slash command
[...]

PR #4532, opened 2026-05-26
Fix: skip excluded commands in command.execute.before to prevent duplicate template injection

HyeokjaeLee:
## Summary

The `command.execute.before` handler in `auto-slash-command` hook was missing the `EXCLUDED_COMMANDS` check that the `chat.message` handler already has. This caused loop commands (`/ulw-loop`, `/ralph-loop`, `/cancel-ralph`) to receive **duplicate template injection**:

1. opencode's native command execution injects `<command-instruction>` + `<user-task>`
2. omo's `auto-slash-command` hook also injects the same content wrapped in `<auto-slash-command>` tags

Both end up in the final message, roughly doubling token cost.

## Root Cause

`EXCLUDED_COMMANDS` in `constants.ts` already lists `ulw-loop`, `ralph-loop`, and `cancel-ralph`. The `chat.message` handler skips these, but `command.execute.before` does not check this set before calling `executeSlashCommand()`.

## Fix

Add an early return when `input.command` matches `EXCLUDED_COMMANDS` in the `command.execute.before` handler — same pattern already used in `chat.message`.

## Related Issues

- #3724 — Original slash command duplication fix (chat.message path)
- #4481 — Duplicate on start-work error retry path (different trigger, same class of bug)

## Checklist

---
## Summary by cubic
Remove the `EXCLUDED_COMMANDS` check in `command.execute.before` for the `auto-slash-command` hook to restore correct template injection for loop commands while keeping deduplication. Prevents missed injections and still avoids duplicate content.

- **Bug Fixes**
  - Removed `EXCLUDED_COMMANDS` guard from `command.execute.before`; `chat.message` already filters via `detectSlashCommand()`.
  - Dedup now relies on `partsContainAutoSlashCommandTags`, so `ulw-loop` and `ralph-loop` inject once as intended.

<sup>Written for commit 43892e8b489177336ad35f3b979af4c6d04c37db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


HyeokjaeLee:

MoerAI (changes requested):
Thanks for the PR, @HyeokjaeLee, and for taking the time on this.

CI is showing real test failures that look caused by this change, not flaky ones. The new EXCLUDED_COMMANDS guard in `auto-slash-command/hook.ts` makes command.execute.before return early for ralph-loop and ulw-loop, but existing tests expect those builtin commands to still receive template injection:
- createAutoSlashCommandHook > command.execute.before hook > should inject template for known builtin commands like ralph-loop
- createAutoSlashCommandHook > command.execute.before hook > should inject template for known builtin commands like ulw-loop
- createPluginInterface > ulw-loop native command smoke

In each, the assertion receives the original text instead of the injected template (3 fail on test (ubuntu-latest)). Could you narrow the EXCLUDED_COMMANDS set so ralph-loop and ulw-loop still get injected, or move the guard so it only skips the genuinely duplicate path? Leaving open as request-changes, happy to re-check once CI is green. No rush.

code-yeongyu:
[sisyphus-bot] Thanks for this PR! It looks like the diff is currently empty and the changes were requested back in June due to test failures. The codebase has since undergone a major refactor moving src/ into packages/omo-opencode/. We're going to close this for now — if the fix is still needed, please rebase against the new package structure and open a fresh PR. Appreciate the contribution! 🙏

PR #4534, opened 2026-05-26
fix(runtime-fallback): mark synthetic 'continue' fallback as internal (#4085)

### LINKED ISSUE #4085: bug: compact/resume can inject generic continue into fresh-looking CLI session
Summary
Discord support feedback reports that OMO/OpenCode CLI can prompt itself with a generic `continue` after compact/resume even though the user did not type it. The reported setup has team disabled and no known autopilot usage, with OpenCode compact enabled at default settings.

Source
- Discord #omo-help / #get-help thread on 2026-05-16 UTC.
- User report: "Is this something new? I didnt prompt that. OMO prompt itself a \"continue\" 😂"
- Follow-up details:
  - OMO version: 4.1.2
  - OpenCode version: 1.15.1
  - Surface: CLI
  - Team: disabled
  - Autopilot: user does not know / not intentionally used
  - OpenCode compact: active, default setting
  - Session was deleted before logs were collected, but user says it happens often.

Observed behavior
A fresh-looking or resumed CLI session can receive/inject a bare `continue` prompt that the user did not type.

Expected behavior
Compact/resume must not fabricate a generic user prompt into the active session. If previous work is unfinished, the runtime should either:
- restore explicit resumable state with provenance, or
- ask/indicate that a previous session is resumable, or
- do nothing until the user provides a real prompt.

Likely area
This looks more like OpenCode compact/resume state injection than OMO autopilot/team behavior, because team is disabled and compact is active. The suspect failure mode is stale resume state deciding the previous turn was unfinished and injecting `continue` into the next session.

Repro data still needed
Ask for the next occurrence before deleting the session:

```bash
opencode --version
bunx oh-my-openagent@latest --version
bunx oh-my-openagent@latest doctor --verbose

# list only state/log file names, not contents
find . -maxdepth 3 -type f \( -iname '*opencode*' -o -iname '*omo*' -o -iname '*session*' -o -iname '*state*' \) | sort | head -80
```

Most useful evidence:

```text
last user prompt before compact
compact/resume line
the unexpected continue prompt
first assistant response after it
whether a brand-new project/session directory reproduces
```

No API keys, tokens, cookies, or session secrets.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

YOMXXX:
## Summary

`runtime-fallback`'s auto-retry path has a fallback for the case when `getLastUserRetryPayload` cannot recover any user text parts (the typical trigger: **OpenCode compacted the session, so prior user turns are gone**). That fallback synthesises a bare `{ type: "text", text: "continue" }` and dispatches it through the prompt-async gate.

The bare part is **not classified as synthetic anywhere**:

- the TUI renders it as if the user typed `continue` — matching the Discord report quoted in #4085 (*"OMO prompt itself a 'continue' 😂"*);
- OMO's other hooks (`continuation`, `keyword-detector`, …) check `isSyntheticOrInternalUserMessage` against `synthetic: true` and the `OMO_INTERNAL_INITIATOR` marker — neither is present — so they treat the turn as a real user prompt and can cascade.

## Root cause

`auto-retry.ts` (pre-fix) returned a raw text part:

```ts
return [{ type: "text" as const, text: "continue" }]
```

while the [`fix(continuation): mark fallback resumes synthetic`](https://github.com/code-yeongyu/oh-my-openagent/commit/36f51ddb) / [`mark resumes synthetic`](https://github.com/code-yeongyu/oh-my-openagent/commit/38b1433f) / [`mark atlas resumes synthetic`](https://github.com/code-yeongyu/oh-my-openagent/commit/1189b96d) series already routed sibling spots (e.g. `src/plugin/event.ts:520, 976`) through `createInternalAgentContinuationTextPart`. The runtime-fallback auto-retry synthetic fallback was missed.

## Fix

Route that lone spot through the same helper so the synthetic continuation carries:

- `OMO_INTERNAL_INITIATOR` marker in `text`
- `synthetic: true`
- `metadata: { compaction_continue: true }`

## Reproduction shape (matches #4085 report)

1. User runs an OMO/OpenCode CLI session with `fallback_models` configured.
2. OpenCode compacts after the context grows enough — prior user turns are replaced by the summary.
3. A subsequent model call fails (quota / 429 / API error) and runtime-fallback's auto-retry kicks in.
4. `getLastUserRetryPayload` returns empty `retryParts`; `delegatedChildSessionBootstrap` also empty.
5. **Pre-fix:** the bare `"continue"` is injected, the TUI renders it as a user prompt, and the user sees a self-issued message they never typed.
6. **Post-fix:** the same prompt still routes through the gate but is now flagged synthetic, so the downstream classifiers and the TUI can handle it as an internal turn.

This matches the Discord-reported environment (team disabled, autopilot not used, default compact active) — none of those are required for the bug, only `fallback_models` + a flaky primary model + a compaction cycle.

## Test plan

- [x] **Red-green cycle verified** for the new regression test in `auto-retry.test.ts` ("#given compact-flushed session with no recoverable user parts #when auto-retry fires the synthetic continuation #then the injected prompt is marked synthetic and carries the internal initiator marker (#4085)").
- [x] `bun test src/hooks/runtime-fallback/ src/shared/internal-initiator-marker.test.ts` → **216 pass / 0 fail**.
- [x] No new typecheck regressions introduced (existing dev errors in `bun-file-shim.ts` / `process-cleanup.test-helpers.ts` are pre-existing and unrelated).

Closes #4085.

---
## Summary by cubic
Marks the synthetic "continue" sent by `runtime-fallback` auto-retry as internal so it no longer appears as a user message or triggers user-facing hooks. Fixes the TUI showing a self-issued "continue" after compaction (fixes #4085).

- **Bug Fixes**
  - Route the fallback through `createInternalAgentContinuationTextPart`, adding `synthetic: true`, `OMO_INTERNAL_INITIATOR`, and `metadata.compaction_continue: true`.
  - Add a regression test that asserts the marker and `synthetic` flag are present in the prompt body.
  - Prevents cascades in `continuation`/`keyword-detector` and stops the TUI from rendering a user-typed "continue".

<sup>Written for commit 640012abd9d383fc905493f24afafe781b88c4f8. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4534?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/runtime-fallback/auto-retry.test.ts
+++ b/src/hooks/runtime-fallback/auto-retry.test.ts
@@ -99,4 +99,40 @@ describe("createAutoRetryHelpers", () => {
     expect(deps.sessionAwaitingFallbackResult.has(sessionID)).toBe(true)
     expect(state.pendingFallbackModel).toBe("openai/gpt-5.4")
   })
+
+  test("#given compact-flushed session with no recoverable user parts #when auto-retry fires the synthetic continuation #then the injected prompt is marked synthetic and carries the internal initiator marker (#4085)", async () => {
+    // given — capture the actual parts forwarded to client.session.promptAsync
+    const promptCalls = { count: 0, lastBody: undefined as unknown }
+    const deps = createDeps(promptCalls)
+    // Post-compact case: messages() returns no user role entries, so
+    // getLastUserRetryPayload falls through to the synthetic "continue".
+    deps.ctx.client.session.messages = async () => ({ data: [] })
+    deps.ctx.client.session.promptAsync = async (args: unknown) => {
+      promptCalls.count += 1
+      promptCalls.lastBody = (args as { body?: unknown })?.body
+      return {}
+    }
+    const helpers = createAutoRetryHelpers(deps)
+    const sessionID = "session-compact-flushed"
+    const state = createFallbackState("anthropic/claude-opus-4-7")
+    state.pendingFallbackModel = "openai/gpt-5.4"
+    deps.sessionStates.set(sessionID, state)
+
+    // when
+    await helpers.autoRetryWithFallback(sessionID, "openai/gpt-5.4", undefined, "session.error")
+
+    // then
+    expect(promptCalls.count).toBe(1)
+    const body = promptCalls.lastBody as { parts?: ReadonlyArray<Record<string, unknown>> } | undefined
+    expect(body).toBeDefined()
+    const parts = body?.parts ?? []
+    expect(parts.length).toBe(1)
+    const firstPart = parts[0] ?? {}
+    expect(firstPart["type"]).toBe("text")
+    // Without the marker + synthetic flag, OMO's continuation/keyword-detector
+    // hooks treat this as a real user prompt and the TUI shows a bare "continue"
+    // that the user never typed (see #4085 / Discord report).
+    expect(firstPart["synthetic"]).toBe(true)
+    expect(String(firstPart["text"] ?? "")).toContain("OMO_INTERNAL_INITIATOR")
+  })
 })
--- a/src/hooks/runtime-fallback/auto-retry.ts
+++ b/src/hooks/runtime-fallback/auto-retry.ts
@@ -17,6 +17,7 @@ import {
   releasePromptAsyncReservation,
 } from "../shared/prompt-async-gate"
 import { isAmbiguousPostDispatchPromptFailure } from "../../shared/prompt-failure-classifier"
+import { createInternalAgentContinuationTextPart } from "../../shared/internal-initiator-marker"
 
 const SESSION_TTL_MS = 30 * 60 * 1000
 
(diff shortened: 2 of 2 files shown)


code-yeongyu:
[sisyphus-dev] I found #4085 during the bug sweep and am reviewing this PR now. The bug path is real: runtime-fallback can synthesize a bare `continue` when compaction removed recoverable user text, making an internal retry look user-authored. I will verify this against current `dev` in an isolated worktree and only merge if the prompt-gate/internal-message invariants stay clean.

code-yeongyu:
[sisyphus-dev] I reproduced/reviewed this against current `dev` with runtime-fallback/internal-marker focused QA and typecheck. The fix direction is correct, but delegated review caught a small repo-style issue in the test comment. I opened replacement PR #4677 with the same fix plus that cleanup, so I am closing this fork PR in favor of #4677. Thank you for the useful fix.

code-yeongyu:
[sisyphus-dev] Superseded by replacement PR #4677, which preserves this fix for #4085 and includes the review cleanup.

PR #4537, opened 2026-05-27
fix(postinstall): warn when platform binary version differs from main package (fixes #3918)

### LINKED ISSUE #3918: Windows platform binary package not upgraded with oh-my-opencode@4.0.0
### Description

After upgrading `oh-my-opencode` to `4.0.0`, the startup banner in OpenCode still displayed an older version:

```text
OhMyOpenCode 3.17.14
```

However, running:

```powershell
oh-my-opencode --version
```

returned:

```text
4.0.0
```

### Environment

- OS: Windows
- Package manager: npm global install
- Shell: PowerShell
- Installed packages before fix:

```text
oh-my-opencode@4.0.0
oh-my-opencode-windows-x64@3.9.0
```

### Root Cause

The main package was upgraded to `4.0.0`, but the platform-specific binary package remained outdated:

```text
oh-my-opencode-windows-x64@3.9.0
```

The `oh-my-opencode` wrapper calls the platform binary from:

```text
%APPDATA%\npm\node_modules\oh-my-opencode-windows-x64\bin\oh-my-opencode.exe
```

So even though the wrapper/package version was `4.0.0`, the actual Windows binary was still old, causing the banner/version shown in OpenCode to be outdated.

### Steps to Reproduce

1. Install or upgrade `oh-my-opencode` globally:

```powershell
npm install -g oh-my-opencode@4.0.0
```

2. Check the wrapper version:

```powershell
oh-my-opencode --version
```

Output:

```text
4.0.0
```

3. Launch OpenCode with OhMyOpenCode enabled.

4. The banner still shows:

```text
OhMyOpenCode 3.17.14
```

5. Check installed global packages:

```powershell
npm list -g --depth=0
```

Observed:

```text
oh-my-opencode@4.0.0
oh-my-opencode-windows-x64@3.9.0
```

### Expected Behavior

When upgrading `oh-my-opencode` to `4.0.0`, the platform-specific package should also be upgraded to `4.0.0`, and the banner should display the correct version.

### Actual Behavior

The main package was upgraded, but the Windows platform binary package stayed on an older version, causing the runtime/banner version to be stale.

### Workaround

Manually reinstall both packages:

```powershell
npm install -g oh-my-opencode@4.0.0 oh-my-opencode-windows-x64@4.0.0
```

After that:

```powershell
oh-my-opencode --version
```

returns:

```text
4.0.0
```

And the platform binary also reports:

```powershell
& "$env:APPDATA\npm\node_modules\oh-my-opencode-windows-x64\bin\oh-my-opencode.exe" --version
```

```text
4.0.0
```

### Suggested Fix

The install/update flow should ensure that the platform-specific binary package version matches the main `oh-my-opencode` package version.

For example, when installing:

```text
oh-my-opencode@4.0.0
```

it should not leave:

```text
oh-my-opencode-windows-x64@3.9.0
```

installed globally.

MoerAI:
## Summary

`oh-my-opencode-windows-x64@3.x` (and the other platform packages) can stay globally installed when the main `oh-my-opencode` package is upgraded. The wrapper picks up the stale binary, so the OpenCode startup banner pins to the old version even after the main package was upgraded — exactly the symptom reported in #3918.

The fix adds an install-time warning when the platform binary's `package.json` version differs from the main package version, with a clear command to fix it. No behavioral change in the wrapper, no change to binary resolution order.

## Root Cause

`postinstall.mjs` already verifies that the platform binary **exists**, but it never compares the platform package's `package.json` version against the main package version. The sisyphus-bot triage on #3918 called this out explicitly:

> `postinstall.mjs` verifies that a candidate platform binary exists/resolves, but it does not fail or warn on main-package/platform-package version mismatch.

## Changes

| File | Change |
|------|--------|
| `bin/version-mismatch.js` | New pure helper `detectPlatformBinaryMismatch({ mainVersion, platformVersion, platformPackage })` — returns the diverging versions when they don't match, null when either side is unknown. |
| `bin/version-mismatch.test.ts` | 9 unit tests covering match / mismatch (older + newer) / unknown-on-either-side / prerelease-vs-stable / two-different-prereleases / same-prerelease / leading-`v` normalization. |
| `postinstall.mjs` | Reads main + platform `package.json` versions and prints a warning + fix command when they diverge. No-op when versions are equal or either is unreadable. |

The helper deliberately treats an unknown version on either side as "no mismatch" — the existing postinstall path already prints `⚠ oh-my-opencode: ...` if the binary itself is missing, so this fix focuses purely on the *version divergence* case from the issue body.

## Codex P2 follow-up (commit `d618ba1f`)

Codex review on the initial commit flagged that the helper stripped prerelease suffixes (`4.5.1-beta.1` → `4.5.1`) before comparing, which would silently pass mismatch detection when the main package is on a prerelease but the platform binary is on the matching stable build. Since `publish.yml` accepts prerelease versions (line 17 `version` input, line 185-186 regex, line 192-193 `next` dist_tag path) and the release workflow ships prereleases as standalone npm packages, that combination is a real install state the wrapper would silently honour.

The follow-up commit drops the suffix-strip from `normalizeVersion()` and compares the full semver string (after the cosmetic leading `v` is stripped). `4.5.1-beta.1` and `4.5.1` are now correctly reported as a mismatch, while `4.5.1-beta.1` ↔ `4.5.1-beta.1` and `v4.5.1` ↔ `4.5.1` continue to match. 3 new regression tests pin the new behavior.

## Reproduction (before fix)

Recreated the #3918 condition by overriding the installed platform package's `package.json` to v3.9.0 while the main package is on v4.5.1:

```
$ node postinstall.mjs
⚠ oh-my-opencode requires OpenCode >= 1.4.0
  Detected: 1.2.22
  Please update OpenCode to avoid compatibility issues.
✓ oh-my-opencode binary installed for win32-x64 (oh-my-opencode-windows-x64)
```

→ Version divergence is **silent**. The user has no signal that the next OpenCode startup will report `3.9.0` in the banner.

## Verification (after fix)

Same simulated mismatch (4.5.1 main vs 3.9.0 platform):

```
$ node postinstall.mjs
⚠ oh-my-opencode requires OpenCode >= 1.4.0
  Detected: 1.2.22
  Please update OpenCode to avoid compatibility issues.
⚠ oh-my-opencode platform binary version mismatch detected
  oh-my-opencode: 4.5.1
  oh-my-opencode-windows-x64: 3.9.0
  The startup banner may show the stale version until the platform binary is updated.
  Fix: npm install -g oh-my-opencode@4.5.1 oh-my-opencode-windows-x64@4.5.1
✓ oh-my-opencode binary installed for win32-x64 (oh-my-opencode-windows-x64)
```

Prerelease-vs-stable now also detected (post-Codex fix):

```
$ bun -e 'import("./bin/version-mismatch.js").then(m => console.log(m.detectPlatformBinaryMismatch({mainVersion: "4.5.1-beta.1", platformVersion: "4.5.1", platformPackage: "oh-my-opencode-linux-x64"})))'
{ mainVersion: '4.5.1-beta.1', platformVersion: '4.5.1', platformPackage: 'oh-my-opencode-linux-x64' }
```

With versions back in sync (both 4.5.1) the warning stays silent — no false alarm.

## Test

- Regression / new tests: `bin/version-mismatch.test.ts` (9 cases — match, mismatch-older, mismatch-newer, unknown-main, unknown-platform, prerelease-vs-stable mismatch, two-different-prereleases mismatch, same-prerelease match, `v`-prefix normalization)
- Related suite: `bun test bin/` → **26 pass, 0 fail** (9 new + 17 existing `platform.test.ts`)
- Typecheck: `bun run typecheck` → clean

## Design notes / scope guardrails

- **Install-time warning only.** No runtime/wrapper change, no change to binary-resolution order, no new behavior, no new dependencies. This deliberately leaves the wrapper hot path untouched.
- **Renamed package family safe.** Reads the actual platform-package name returned by the existing resolver, so the `oh-my-opencode` → `oh-my-openagent` rename and the baseline variants are all covered by the same code path.
- **Prerelease-aware.** `publish.yml` ships prereleases to `next` dist_tag, so prerelease-vs-stable IS a real mismatch (post-Codex fix).
- **Silent when either side is unreadable.** Better to skip than to false-alarm; the existing postinstall already handles the "binary missing" case separately.

Fixes #3918

### DIFF
--- a/bin/version-mismatch.js
+++ b/bin/version-mismatch.js
@@ -25,7 +25,7 @@ function normalizeVersion(version) {
  *
  * Comparison is exact on the full semver string (after stripping a leading "v"),
  * including any pre-release suffix. This means `4.5.1-beta.1` and `4.5.1` are
- * treated as DIFFERENT versions — matching the publish workflow that ships
+ * treated as DIFFERENT versions, matching the publish workflow that ships
  * prerelease builds to the `next` dist-tag alongside stable releases.
  *
  * @param {object} input
--- a/bin/version-mismatch.js
+++ b/bin/version-mismatch.js
@@ -4,11 +4,17 @@
 // alongside `oh-my-opencode@4.0.0`, leaving the startup banner pinned to 3.9.0.
 
 /**
+ * Strips an optional leading "v" so callers can pass either "4.5.1" or "v4.5.1".
+ * Pre-release suffixes (e.g. "-beta.1") are preserved because the release
+ * workflow publishes prereleases as standalone versions on the `next` dist-tag,
+ * and `oh-my-opencode@4.5.1-beta.1` paired with `oh-my-opencode-x64@4.5.1`
+ * IS a real mismatch the wrapper would silently honour.
+ *
  * @param {string} version
- * @returns {string} Version core (strips leading "v" and pre-release suffix).
+ * @returns {string} Full version with leading "v" stripped.
  */
 function normalizeVersion(version) {
-  return version.replace(/^v/, "").split("-")[0];
+  return version.replace(/^v/, "");
 }
 
 /**
@@ -16,6 +22,11 @@ function normalizeVersion(version) {
  *
  * Both versions must be known to report a mismatch; if either is null/undefined,
  * the check is skipped (returns null) rather than producing a false alarm.
+ *
+ * Comparison is exact on the full semver string (after stripping a leading "v"),
+ * including any pre-release suffix. This means `4.5.1-beta.1` and `4.5.1` are
+ * treated as DIFFERENT versions — matching the publish workflow that ships
+ * prerelease builds to the `next` dist-tag alongside stable releases.
  *
  * @param {object} input
  * @param {string | null | undefined} input.mainVersion
--- a/bin/version-mismatch.js
+++ b/bin/version-mismatch.js
@@ -1,0 +1,36 @@
+// bin/version-mismatch.js
+// Detects platform binary version mismatch against the main package.
+// Background: issue #3918 - `oh-my-opencode-windows-x64@3.9.0` could stay installed
+// alongside `oh-my-opencode@4.0.0`, leaving the startup banner pinned to 3.9.0.
+
+/**
+ * @param {string} version
+ * @returns {string} Version core (strips leading "v" and pre-release suffix).
+ */
+function normalizeVersion(version) {
+  return version.replace(/^v/, "").split("-")[0];
+}
+
+/**
+ * Detect a main / platform-binary version mismatch.
+ *
+ * Both versions must be known to report a mismatch; if either is null/undefined,
+ * the check is skipped (returns null) rather than producing a false alarm.
+ *
+ * @param {object} input
+ * @param {string | null | undefined} input.mainVersion
+ * @param {string | null | undefined} input.platformVersion
+ * @param {string} input.platformPackage
+ * @returns {{ mainVersion: string, platformVersion: string, platformPackage: string } | null}
+ */
+export function detectPlatformBinaryMismatch({ mainVersion, platformVersion, platformPackage }) {
+  if (!mainVersion || !platformVersion) {
+    return null;
+  }
+
+  if (normalizeVersion(mainVersion) === normalizeVersion(platformVersion)) {
+    return null;
+  }
+
+  return { mainVersion, platformVersion, platformPackage };
+}
--- a/bin/version-mismatch.test.ts
+++ b/bin/version-mismatch.test.ts
@@ -1,0 +1,76 @@
+// bin/version-mismatch.test.ts
+import { describe, expect, test } from "bun:test";
+import { detectPlatformBinaryMismatch } from "./version-mismatch.js";
+
+describe("detectPlatformBinaryMismatch", () => {
+  test("returns null when main and platform versions match", () => {
+    // #given identical main and platform binary versions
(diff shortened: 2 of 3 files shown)


MoerAI:

MoerAI:
Rebased onto current `dev` to resolve a merge conflict from #4628 and follow-up codex installer commits.

Scope reduction in this rebase:
- Dropped the `bin/baseline-detection.js` extraction (commit 61f34bc3 in the previous push). Upstream re-inlined `supportsAvx2` in `bin/oh-my-opencode.js` directly, so extracting it back out would conflict with that direction. The codex P2 review note about postinstall vs wrapper baseline preference is acknowledged; happy to file as a follow-up once this lands, since fixing it cleanly now requires a deeper refactor than is appropriate for this scope.
- Kept: the core version-mismatch warning (commit 815a840d), the prerelease-aware comparison (commit d618ba1f), and the em dash fix on the JSDoc (squashed into a single follow-up commit).

Resulting diff is 3 files, +203/-3 LOC (47 LOC helper, 120 LOC test, ~36 LOC postinstall delta).

All 37 `bin/` tests pass locally; `bun run typecheck` clean. Manual QA exercised the helper against match / mismatch / prerelease-vs-stable / unknown-side cases and the resolved versions print the expected warning string.

MoerAI:
Thanks @MoerAI, merging. The diff is on the larger side, but most of the LOC is the new `bin/version-mismatch.js` helper and its companion tests; the actual change to `postinstall.mjs` is small and isolated to the install-time warning.

Issue #3996, opened 2026-05-13
[Bug]: `tool-pair-validator` corrupts background sub-agent sessions by inserting synthetic placeholder tool results


lapertor:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

The `tool-pair-validator` hook (added in PR #3017 to fix #3014) lacks background session filtering. When a sub-agent launched via `task(subagent_type="oracle", run_in_background=true)` makes tool calls, the hook fires on the background session and inserts synthetic placeholder messages (`"Tool output unavailable (context compacted)"`) in place of missing `tool_result` blocks. This corrupts the message history that the LLM receives on subsequent API calls, causing the model to either retry reads indefinitely or enter a reasoning loop that produces empty messages and hangs forever.

Other hooks like `keyword-detector` already skip background sessions explicitly, but `tool-pair-validator` runs unconditionally on all sessions.

### Steps to Reproduce

1. Configure oh-my-openagent with default settings (no `disabled_hooks`)
2. From any project, launch Oracle in background mode:
   ```typescript
   task(
     subagent_type="oracle",
     load_skills=[],
     run_in_background=true,
     description="Repro tool-pair-validator bug",
     prompt="Read include/higra/algo/watershed.hpp and tell me what it does."
   )
   ```
3. Wait 2-3 minutes
4. Check `background_output(task_id="...")` — status remains `running`, `Last tool: read` never updates
5. Check the oh-my-openagent log (`%TEMP%/oh-my-opencode.log`) — observe:
   - Multiple `[tool-execute-after] Unable to recover stored metadata and no native session linkage was present` entries
   - `[tool-pair-validator] Repaired missing tool_result blocks` with `syntheticUserMessageInserted: true`

### Expected Behavior

Oracle should complete within ~15-20 seconds with a substantive response describing the file contents.

### Actual Behavior

Oracle hangs indefinitely (observed for >4 minutes before manual cancellation). Status stays `running`, no new messages after the initial tool calls.

### Doctor Output

```shell
[...]

PR #4542, opened 2026-05-27
fix(tool-pair-validator): diagnose subagent missing tool results

SoShymKing:
## Summary

- Add a subagent-specific `tool-pair-validator` path that analyzes missing `tool_result` cases without mutating subagent message history.
- Preserve the #4032 behavior that prevents synthetic placeholder `tool_result` insertion from corrupting background sub-agent sessions.

For tracked subagent sessions, this change is diagnostic-only. The validator analyzes whether missing `tool_result` pairs would require repair and logs that state with fields such as `needsRepair` and `missingToolUseIDs`, but it intentionally does not insert synthetic `tool_result` blocks or mutate subagent message history.

## Changes

- Extract shared missing tool-result detection so main-session repair and subagent diagnostics use the same pairing logic.
- Add `repairSubAgentMissingToolResults()` for tracked subagent sessions; it logs whether repair would be needed, including missing tool-use IDs, but does not insert placeholder results.
- Keep existing main-session repair behavior intact for compaction/orphaned `tool_use` recovery.
- Add regression coverage confirming tracked subagent messages remain unchanged when a user turn is missing matching `tool_result`.

## Testing

```bash
bun test src/hooks/tool-pair-validator/hook.test.ts src/plugin/messages-transform.test.ts
```

## Related Issues

Related to #3996  
Follow-up to #4032

---
## Summary by cubic
Adds a subagent-aware path in the tool-pair validator that detects missing `tool_result` and logs diagnostics without mutating subagent message history. Main-session repair behavior is unchanged.

- **Bug Fixes**
  - Share detection via `analyzeMissingToolResults` so both paths use the same pairing logic.
  - Add `repairSubAgentMissingToolResults` that logs needed repairs (missing tool_use IDs, next role) for tracked subagent sessions; no placeholders are inserted.
  - Preserve main-session repair: insert a synthetic user turn or missing `tool_result` entries when required.
  - Add a regression test to confirm tracked subagent messages remain unchanged.

<sup>Written for commit 612875472cbd65535f0ed9ac31e9bc8d1bb8e9ae. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4542?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/tool-pair-validator/hook.test.ts
+++ b/src/hooks/tool-pair-validator/hook.test.ts
@@ -276,6 +276,33 @@ describe("createToolPairValidatorHook", () => {
     }
   })
 
+  it("leaves tracked subagent user turns unchanged when tool_result is missing", async () => {
+    //#given
+    _resetForTesting()
+    subagentSessions.add("ses_background_2")
+    const backgroundMessages = [
+      {
+        info: { role: "assistant", sessionID: "ses_background_2" },
+        parts: [{ type: "tool_use", id: "toolu_background_2" }],
+      },
+      {
+        info: { role: "user", sessionID: "ses_background_2" },
+        parts: [{ type: "text", text: "continue background session" }],
+      },
+    ] satisfies TestMessage[]
+    const originalBackgroundMessages = JSON.parse(JSON.stringify(backgroundMessages))
+
+    try {
+      //#when
+      await runTransform(backgroundMessages)
+
+      //#then
+      expect(backgroundMessages).toEqual(originalBackgroundMessages)
+    } finally {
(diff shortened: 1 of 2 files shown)


MoerAI:
Thanks @SoShymKing, merging. The refactor in `tool-pair-validator/hook.ts` keeps the non-subagent repair behavior-equivalent and makes the subagent branch diagnosis-only (it logs the missing tool_result analysis and continues without mutating the message stream), which is the right call for a global transform hook. The added test pins that subagent turns stay unchanged.

PR #4543, opened 2026-05-27
test(process-cleanup): add fast-signal-cleanup regression test + fix existing test's false-green (#4058)

### LINKED ISSUE #4058: [sisyphus-bot] [Bug]: Process hangs on SIGINT/SIGTERM when cleanup completes quickly

YOMXXX:
## Summary

The product fix for #4058 landed in #4026 (commit [0e49767a](https://github.com/code-yeongyu/oh-my-openagent/commit/0e49767a)) — passing `exitAfterCleanup=true` to `scheduleForcedExit` for SIGINT/SIGTERM so fast cleanup still triggers `process.exit(0)`. That ships correct runtime behaviour, but the test file still had **two problems** that code-yeongyu flagged on the parallel PR #4057:

> The PR should add/update tests to mock `process.exit` for fast signal cleanup **and assert the full test file completes**.

This PR addresses both. It is test-only — no production changes — and it should let #4058 close once merged.

## Problem 1: missing regression test for the `process.exit(0)` after-cleanup branch

The existing `cleanup finishes after SIGINT` test only asserts that `setTimeout` was armed and `clearTimeout` was called. It would still pass if the product fix were reverted to `scheduleForcedExit(..., 0)`. There is no test pinning the actual `process.exit(0)` call that the fix introduced.

## Problem 2: the existing test silently kills the bun runner ("false-green")

The existing `cleanup finishes after SIGINT` test re-enables `_scheduleForcedExitEnabled` to verify timer wiring but only spies `setTimeout` / `clearTimeout`. After #4026, the `.finally` branch now reaches the real `process.exit(0)` — bun's runner is killed between the banner and the first per-test summary line:

```console
$ bun test src/features/background-agent/process-cleanup.test.ts
bun test v1.3.14 (0d9b296a)
$ echo $?
0
```

No "X pass / Y fail" summary. Every test in the file ran via filters in CI's isolated-batch sharder, so this masked itself in CI but produced silent zero-summary passes locally.

## What this PR does

All in `src/features/background-agent/process-cleanup.test.ts`:

1. Add `spyOn(process, "exit").mockImplementation(...)` to the existing `cleanup finishes after SIGINT` test so the after-cleanup branch can no longer kill the bun runner. The file now reports a real `30 pass / 0 fail / 61 expect() calls`.
2. Add **two new regression tests** under the same describe:
   - `fast SIGINT cleanup → process.exit(0) via after-cleanup path (#4058 regression)`
   - `fast SIGTERM cleanup → process.exit(0) via after-cleanup path (#4058 regression)`
   Both mock `process.exit`, fire the corresponding listener, flush microtasks, then assert `expect(exitSpy).toHaveBeenCalledTimes(1)` and `toHaveBeenLastCalledWith(0)`.

## Red-green cycle verified

| Product code state | Result |
| --- | --- |
| `scheduleForcedExit(cleanupResult, 0, true)` (current `dev`) | **30 pass / 0 fail** |
| `scheduleForcedExit(cleanupResult, 0)` (revert to pre-#4026) | **28 pass / 2 fail** — both failures are the new regression tests, each reporting `Expected number of calls: 1 / Received: 0` |

So the tests now catch a regression of the very behaviour #4026 fixed.

## Test plan

- [x] `bun test src/features/background-agent/process-cleanup.test.ts` → **30 pass / 0 fail / 61 expect() calls** (was silent zero-summary on `dev`)
- [x] `bun test src/features/background-agent/` → **594 pass / 0 fail / 1440 expect() calls** across 39 files
- [x] Red-green cycle verified for both new regression tests (revert third arg of `scheduleForcedExit` → 2 fail; restore → 30 pass)

Closes #4058 (product fix already landed in #4026; this is the test-coverage follow-up code-yeongyu requested on #4057).

---
## Summary by cubic
Add regression tests for fast SIGINT/SIGTERM cleanup to assert the after-cleanup path calls `process.exit(0)`, pinning the #4026 fix for #4058. Add a test-only exit seam in `process-cleanup.ts` (`__setProcessExitForTesting`/`__resetProcessExitForTesting`) and update tests to use it so `scheduleForcedExit` doesn’t kill the Bun runner, eliminating false-green passes.

<sup>Written for commit cd04504d715003e0655bf6caac791d92e3d44a31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4543?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


YOMXXX:
## ⚠️ Critical follow-up: this PR's scope is wider than #4058 alone

CI is currently red on this PR (45 fail / 7660 pass), and I want to flag why before anyone judges that as a regression: **`dev`'s CI test job has been silently false-green for the entire test-file batch downstream of `process-cleanup.test.ts`**, not just hiding the regression in the local `bun test` invocation that #4058 originally tracked.

### Evidence from the most recent green CI on `dev`

Run [`26490879484`](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/26490879484) (commit `99de130`, marked SUCCESS). Grepping its `test` job log for per-file groups, the last file emitted before `Post Run cleanup` is:

```
##[group]src/features/background-agent/process-cleanup.test.ts:
(pass) #given the first cleanup manager > #when registerManagerForCleanup runs #then signal handlers are registered
(pass) #given the first cleanup manager > #when the exit listener runs #then the registered manager shuts down
Post Run actions/cache@v4 ← bun runner killed here
```

Bun got two `pass` lines into the file, then `process.exit(0)` fired inside `#when cleanup finishes after SIGINT` (the after-cleanup branch that PR #4026 wired up), bun was killed before printing the file summary, **and the job exited with code 0 so GitHub Actions reported SUCCESS**. None of the test files that would have run after `process-cleanup.test.ts` in alphabetical order ran in CI at all.

A `grep "##\[group\]src" | wc -l` on that same green run gives **2459 lines**. The full test corpus on `dev` is 7706 tests across 808 files. So CI has been verifying roughly the first third of the test suite for some unknown number of weeks; the other ~5247 tests have not been gated by CI on `dev`.

### What this PR's red CI actually shows

This PR's CI ran **7706 tests / 808 files**. That's the first run that actually completed the test batch since the silent-kill regression landed:

- **7660 pass** — including the ~5247 tests that hadn't been gated on `dev` and pass cleanly
- **45 fail / 1 error** — pre-existing failures on `dev` that the silent kill was hiding. Spot-checked locally on a clean `upstream/dev` checkout (no patch applied):

  ```console
  $ bun test src/plugin/ultrawork-db-model-override.test.ts src/tools/delegate-task/tools.test.ts
   135 pass
   6 fail
  ```

  Same families of failures as in the CI log (`ultrawork-db-model-override`, `sisyphus-task > browserProvider propagation`, etc.). These are not introduced by this PR.

### Suggested merge path

1. **Merge this PR first.** It does exactly what was asked on #4057 — mock `process.exit` for fast signal cleanup, assert the full test file completes, lock the #4026 regression with two new tests. The new CI red is the truthful state of `dev`, not a regression.
2. **Triage the 45 pre-existing failures separately.** Happy to file follow-ups grouped by area (ultrawork-db-model-override, delegate-task category resolution, callback server, provider model ID transforms, …) — those look like independent bugs that have been masked for a while.

If you'd rather not merge a red-CI PR even given the context, an alternative is to wrap `afterAll` with an explicit `process.exit(0)` at the end of `process-cleanup.test.ts` so the file still kills the bun runner *after* its 30 tests print their summary. That preserves the dev CI baseline (false-green) but loses the systemic finding. I'd recommend the first path; calling out the option in case it's the lower-risk merge for now.

code-yeongyu:
[sisyphus-dev] I verified this is targeting a real #4058 test/false-green gap, but the PR is not merge-ready because its CI test jobs fail across platforms. I am taking the narrow replacement route from current dev and will link the replacement PR back here.

code-yeongyu:
[sisyphus-dev] Superseded by #4674. This PR had the right #4058 target but failing CI; #4674 is the current-dev replacement with the same focused coverage goal and passing local background-agent/typecheck validation.

PR #4578, opened 2026-05-28
fix(background-agent): validate output before completing task on terminal session status

### LINKED ISSUE #4570: Background task can be marked completed when subagent never starts LLM
## Summary

When launching multiple background/subagent tasks, one task can be marked as completed even though the subagent session never starts processing and never produces any assistant/tool output.

This makes the parent session report the background task as completed, while the subagent effectively did not run.

## Environment

- OpenCode: 1.15.11
- oh-my-openagent: 4.5.1
- Provider: OpenAI-compatible local proxy (`cliproxyapi`)
- Models involved: `cliproxyapi/gpt-5.5`, `cliproxyapi/gpt-5.4`
- Scenario: launching 5 review subagents in parallel

## Observed behavior

One background task:

- Background task: `bg_3de8f99a`
- Description: `Execute QA checks`
- Subagent session: `ses_196df1d63ffexegSwgi4Ug8Dcb`
- Agent: `Sisyphus-Junior`
- Model: `cliproxyapi/gpt-5.4`, variant `max`

The subagent session was created and received the initial prompt, but it never started the LLM:

- Session has exactly 1 message: the initial user prompt.
- No assistant message.
- No tool message.
- No transcript.
- No `session.processor` log for that session.
- No `service=llm ... session.id=ses_196df1d63ffexegSwgi4Ug8Dcb` log.

Relevant OpenCode log excerpt:

```text
2026-05-27T11:10:23 service=session id=ses_196df1d63ffexegSwgi4Ug8Dcb title=Execute QA checks (@Sisyphus-Junior subagent) ... created
2026-05-27T11:10:23 service=bus type=session.created publishing
2026-05-27T11:10:23 service=bus type=session.updated publishing
2026-05-27T11:10:23 service=bus type=session.next.agent.switched publishing
2026-05-27T11:10:23 service=bus type=message.updated publishing
2026-05-27T11:10:23 service=bus type=message.part.updated publishing
2026-05-27T11:10:23 service=session.prompt session.id=ses_196df1d63ffexegSwgi4Ug8Dcb cancel
2026-05-27T11:10:23 service=bus type=session.status publishing
2026-05-27T11:10:23 service=bus type=session.idle publishing
```

The `cancel` line appears 28ms after message creation. There is no LLM call in between.

The parent session later reported:

```text
[BACKGROUND TASK COMPLETED]
ID: bg_3de8f99a
Description: Execute QA checks
Duration: 0s
```

and later included it under `[ALL BACKGROUND TASKS COMPLETE]`.

## Expected behavior

A background/subagent task should not be marked completed unless the child session produced at least one assistant or tool message, or otherwise has a clear successful terminal state.

If the session contains only the internal initial user prompt and no assistant/tool output, the background task should be marked failed / not-started / dispatch-failed, not completed.

## Why this seems OMO-related

The child prompt includes the internal marker:

```html
```

The installed OMO bundle contains helpers that classify marker/internal messages as synthetic/internal, and the completion poller appears to use message-count stability as a completion signal.

From the installed `dist/index.js`, the poller has logic equivalent to:

```js
const currentMsgCount = msgs.length;
if (currentMsgCount > 0 && currentMsgCount === lastMsgCount) {
  stablePolls++;
  if (stablePolls >= STABILITY_REQUIRED) {
    log(`[call_omo_agent] Session complete, ${currentMsgCount} messages`);
    break;
  }
}
```

This can treat a session with only the initial internal user prompt as complete, even though no assistant/tool response exists.

There is later response extraction logic that checks for assistant/tool messages, but in the background-task path the task still surfaced as completed.

## Suggested fix

Completion detection should require at least one non-internal assistant or tool message before considering a subagent session complete.

For example:

- Do not consider `currentMsgCount > 0` sufficient.
- Check for at least one assistant/tool message.
- If only the initial internal user message exists after a timeout, mark the task as failed/not-started.
- Surface this clearly in `background_output` and `[ALL BACKGROUND TASKS COMPLETE]`.

Possible condition:

```js
const hasAssistantOrTool = msgs.some(
  (m) => m.info?.role === "assistant" || m.info?.role === "tool"
);

if (!hasAssistantOrTool) {
  // keep waiting until timeout, then fail as not-started
}
```

## Impact

This is dangerous because the orchestrator believes QA/review work completed when one subagent never ran. In my case, 5 review subagents were launched, 4 actually ran, and 1 was silently marked completed with no output.

Hungdoan565:
## Summary

Fixes #4570

Background tasks could be marked as completed even when the subagent never started the LLM. This happened when a session received a terminal status (e.g. 'interrupted') before producing any output.

## Root Cause

In \pollRunningTasks()\, the terminal session status path called \	ryCompleteTask()\ directly without first checking \alidateSessionHasOutput()\. The idle/gone path already had this guard, but the terminal status path was missing it.

## Fix

Added \alidateSessionHasOutput()\ check before \	ryCompleteTask()\ on the terminal session status path. If no output exists, the task is marked as error instead of completed.

## Testing

- Added test: task with terminal session status but no output is marked as error
- All existing tests pass

Closes #4570

---
## Summary by cubic
Prevents background tasks from being marked completed when a subagent session reaches a terminal status (e.g., "interrupted") without producing any output. Such tasks are now marked as error instead. Addresses #4570.

- **Bug Fixes**
  - In `pollRunningTasks`, validate output with `validateSessionHasOutput` before calling `tryCompleteTask` for terminal statuses.
  - If no output exists, call `failCrashedTask` with "Subagent session terminated without producing any output."; added tests to assert error presence and message content.

<sup>Written for commit 40831a7d259b82834ebad95699cec21be5f41310. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4578?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/manager.polling.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/manager.polling.test.ts
@@ -449,6 +449,66 @@ describe("BackgroundManager pollRunningTasks", () => {
       expect(task.completedAt).toBeDefined()
     })
 
+    test('#when session status is "interrupted" but no output exists #then marks task as error', async () => {
+      //#given
+      const manager = createManagerWithClient({
+        status: async () => ({ data: { "ses-interrupted-no-output": { type: "interrupted" } } }),
+        messages: async () => ({ data: [] }),
+      })
+      const task = createRunningTask("ses-interrupted-no-output")
+      injectTask(manager, task)
+
+      //#when
+      const poll = manager["pollRunningTasks"]
+      await poll.call(manager)
+      await manager.shutdown()
+
+      //#then
+      expect(task.status).toBe("error")
+      expect(task.error).toBeDefined()
+      expect(task.error).toContain("terminated without producing any output")
+    })
+
+    test('#when session status is "interrupted" but no output exists #then marks task as error', async () => {
+      //#given
+      const manager = createManagerWithClient({
+        status: async () => ({ data: { "ses-interrupted-no-output": { type: "interrupted" } } }),
+        messages: async () => ({ data: [] }),
+      })
+      const task = createRunningTask("ses-interrupted-no-output")
+      injectTask(manager, task)
+
+      //#when
+      const poll = manager["pollRunningTasks"]
+      await poll.call(manager)
+      await manager.shutdown()
+
+      //#then
+      expect(task.status).toBe("error")
+      expect(task.error).toBeDefined()
+      expect(task.error ?? "").toContain("terminated without producing any output")
+    })
+
+    test('#when session status is "interrupted" but no output exists #then marks task as error', async () => {
+      //#given
+      const manager = createManagerWithClient({
+        status: async () => ({ data: { "ses-interrupted-no-output": { type: "interrupted" } } }),
+        messages: async () => ({ data: [] }),
+      })
+      const task = createRunningTask("ses-interrupted-no-output")
+      injectTask(manager, task)
+
+      //#when
+      const poll = manager["pollRunningTasks"]
+      await poll.call(manager)
+      await manager.shutdown()
+
+      //#then
+      expect(task.status).toBe("error")
+      expect(task.error).toBeDefined()
+      expect(typeof task.error === "string" && task.error.includes("terminated without producing any output")).toBe(true)
+    })
+
     test('#when session status is an unknown type #then completes the task', async () => {
       //#given
       const manager = createManagerWithClient({
--- a/packages/omo-opencode/src/features/background-agent/manager.ts
+++ b/packages/omo-opencode/src/features/background-agent/manager.ts
@@ -2895,6 +2895,12 @@ The task was re-queued on a fallback model after a retryable failure.
           }
 
           if (sessionStatus && isTerminalSessionStatus(sessionStatus.type)) {
+            const hasValidOutput = await this.validateSessionHasOutput(sessionID)
+            if (!hasValidOutput) {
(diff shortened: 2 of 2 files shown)


MoerAI (changes requested):
Thanks for the PR, @Hungdoan565, and for taking the time on this. The src change in `manager.ts` (validating that a terminal session actually produced output before completing, otherwise marking the task as error) is reasonable.

One thing to fix before this can land: the test file `manager.polling.test.ts` contains the same test three times. All three blocks share the identical name `#when session status is "interrupted" but no output exists #then marks task as error` with the same setup, differing only in how the final assertion is phrased. That looks like a copy-paste artifact. Please collapse them into a single test (keep the clearest assertion), so the suite does not carry three redundant copies.

This is a quality convention in `AGENTS.md` (tests should be meaningful, not duplicated). Happy to re-review once the duplicates are removed. No rush.

code-yeongyu:
[sisyphus-bot] Dev's manager.ts already contains validateSessionHasOutput() (line 2231) which implements the same output validation this PR adds. The fix appears to have landed through another path. Closing as superseded — if issue #4570 still has uncovered edge cases, please open a new PR against current dev.

PR #4579, opened 2026-05-28
fix(background-agent): cancel only the specific task's concurrency waiter

### LINKED ISSUE #4505: [Bug]: Multiple Thinking at the Same Time, Can Only Cancel One of Them
### Prerequisites

### Bug Description

- Description

After the recent updates, I found that Opencode is generating multiple thinking processes at the same time.

As shown in the screen shot, there are two Thoughts working at the same time in parallel, executing commands without exchanging information (separated by Writing Command, sometimes more than two). Furthermore, hitting Esc twice won't stop one of them, meaning that one is executing without monitoring by the main agent. The only way to stop is to close the terminal window running Opencode. Cannot be sure is it a bug from Opencode or Omo.

Image

### Steps to Reproduce

1. Make a session with Sisyphus, a complicated task is more likely to trigger.
2. Sisyphus is generating multiple responses on main agent in parallel.

### Expected Behavior

Sisyphus should be generating only one response flow in main agent.

### Actual Behavior

Sisyphus is generating multiple responses in main agent in parallel.
For a special case, if I dont stop it, the final responses are generated as shown below. (Chinese and English response share the same content) You can see there is a `Sisyphus - Ultraworker · Kimi K2.5 · 2m 26s` marking an end of a response.

Image

Image

### Doctor Output

```shell
Cannot find bunx command. Manually:
Opencode version: 1.15.10
Omo version: 4.5.1
```

### Error Logs

```shell

```

### Configuration

```json
{
  "agents": {
    "sisyphus": { "model": "deepseek/deepseek-v4-flash" },
    "hephaestus": { "model": "github-copilot/gpt-5.2-codex" },
    "librarian": { "model": "opencode/deepseek-v4-flash-free"},
    "oracle": { "model": "opencode/deepseek-v4-flash-free" },
    "explore": { "model": "opencode/deepseek-v4-flash-free", "temperature": 0.6 },
    "multimodal-looker": { "model": "kimi-for-coding/kimi-k2.5" },
    "prometheus": { "model": "opencode/deepseek-v4-flash-free" },
    "metis": { "model": "opencode/deepseek-v4-flash-free" },
    "momus": { "model": "github-copilot/gpt-5.2-codex" },
    "atlas": { "model": "opencode/deepseek-v4-flash-free" },
    "sisyphus-junior": { "model": "opencode/deepseek-v4-flash-free" }
  }
}
```

### Additional Context

_No response_

### Operating System

Linux

### OpenCode Version

1.15.10

Hungdoan565:
## Summary

Cancelling one background task previously rejected ALL pending concurrency waiters for the same model key. This caused concurrent tasks on the same model to be erroneously cancelled when only one task was intended to be cancelled.

## Changes

- Add \	askId\ parameter to \ConcurrencyManager.acquire()\ to tag queue entries
- Add \ConcurrencyManager.cancelWaiter(model, taskId)\ for per-task cancellation that removes only the specific entry from the queue
- Update \cancelTask()\ in \manager.ts\ to use \cancelWaiter()\ instead of rejecting all waiters
- Pass \	askId\ through all \cquire()\ call sites (\manager.ts\, \spawner.ts\)

## Verification

- \un run typecheck\ passes
- \un test src/features/background-agent/concurrency.test.ts\ passes (28 tests, 0 failures)
- Regression test added: 2 concurrent tasks on same model, cancel one, verify other continues

Closes #4505

---
## Summary by cubic
Cancelling one background task now removes only that task’s own concurrency waiter; other queued tasks on the same model key keep their place. We also canonicalize concurrency keys so provider-level limits apply across model variants and fallback retries. Resolves Linear #4505.

- **Bug Fixes**
  - Added taskId to `ConcurrencyManager.acquire` to tag queue entries.
  - Implemented `cancelWaiter(model, taskId)` and switched the cancel flow to use it; plumbed taskId through all `acquire` call sites.
  - Canonicalized concurrency groups via `getConcurrencyKey` so provider limits hold across variants; grouped fallback retries by provider.
  - Added regression tests to verify unaffected waiters continue and provider limits are enforced.

<sup>Written for commit 6f7f4bce098326731c4d9d9ca0c80e9993979987. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4579?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


heunghingwan:
This PR does **not** fix #4505 and should not close it.

The PR description and title frame the bug as "cancelling one task was rejecting ALL pending concurrency waiters" — and the fix switches from `cancelWaiters(key)` (reject all) to `cancelWaiter(key, taskId)` (reject one). That is a valid change on its own, but it does not match what #4505 is actually reporting.

**What #4505 actually says:**

The reporter explicitly states the *expected* behavior is: *"Sisyphus should be generating **only one** response flow in the main agent."* The problem is that **multiple background tasks are being spawned concurrently** in the first place. The "Can Only Cancel One of Them" in the title is a *symptom* of that core issue, not the root cause.

**What this PR does:**

It makes cancellation more surgical — only rejecting the specific task`s waiter instead of all waiters for the same model key. This means:

- **Before the PR**: cancelling one task kills all queued waiters (brutal but effective at cleanup).
- **After the PR**: cancelling one task leaves sibling tasks alive and running.

For the #4505 scenario, this is actually **worse**: if the user has multiple unwanted parallel thinking processes, they previously could cancel once to kill everything. After this change, each would need to be cancelled individually, and the underlying concurrency issue (why are multiple tasks spawned simultaneously?) is untouched.

**Summary:**

| #4505 root problem | This PR |
|---|---|
| Multiple background tasks spawned in parallel for one user request | ❌ Not addressed |
| User wants only one response flow | ❌ Not addressed |
| Cancellation is per-model (too broad) | ✅ Makes it per-task (but this is a side concern, not the fix) |

This PR makes a valid refactor to cancellation granularity, but it should not claim to close #4505. The real issue — unintended parallel spawning — remains open.

MoerAI (dismissed):
Thanks @Hungdoan565.

The concurrency-waiter scoping fix is well-justified, but the diff is 136 LOC across 4 files (`src/features/background-agent/concurrency.ts`, `manager.ts`, `spawner.ts`, plus the test). `AGENTS.md` targets `fix()` PRs at under 50 LOC and 1-2 files so they review and revert cleanly.

Could you isolate the minimal change to `concurrency.ts` and the regression test, and move the touch-ups in `manager.ts`/`spawner.ts` to a separate refactor PR if they are not strictly required for the fix?

Leaving open as request-changes.

code-yeongyu (approved):
Verified after maintainer force-push. The branch is narrowed to the specific concurrency waiter fix: ConcurrencyManager tracks task ids, processKey passes the queued task id and absorbs cancellation rejection for already-cancelled tasks, and the regression test lives in a focused small test file instead of expanding the existing large concurrency.test.ts. Local checks passed: bun test src/features/background-agent/concurrency-cancel-waiter.test.ts src/features/background-agent/concurrency.test.ts, bun test src/features/background-agent/manager.test.ts, and bun run typecheck.

PR #4583, opened 2026-05-28
fix(background-agent): prevent stale interrupt when fallback retry is in flight

### LINKED ISSUE #4528: Runtime fallback and background subagents can stall or mark completed work interrupted

Hungdoan565:
## Summary
Fixes race condition where stale-interrupt path marks a task terminal before a queued fallback retry runs (issue #4528).

## Problem
When 	ryFallbackRetry() is invoked and queues a retry, the task remains in 
unning status while the retry is being set up. During this window, the stale-task poller (checkAndInterruptStaleTasks) can fire and mark the task as cancelled because it appears stale. The retry then either operates on a cancelled task or fails silently.

## Solution
- Added 
etryInFlightTasks: Set<string> to BackgroundManager class
- When 	ryFallbackRetry() is called, the task ID is added to the set immediately
- The ID is removed in a inally block only if the retry did NOT succeed (if it succeeded, the task gets a new session and the old ID is no longer relevant to stale detection)
- checkAndInterruptStaleTasks now accepts an optional 
etryInFlightTasks parameter and skips any task whose ID is in the set

## Testing
- Added 3 tests in 	ask-poller.test.ts verifying:
  - Stale interruption is skipped when task is in 
etryInFlightTasks`n  - Stale interruption proceeds when task is NOT in 
etryInFlightTasks`n  - Stale interruption proceeds when 
etryInFlightTasks is not provided (backward compat)
- Typecheck passes (no new errors introduced)
- All 45 task-poller tests pass

---
## Summary by cubic
Prevents stale interrupts from cancelling a task while a fallback retry is queued by tracking “retry-in-flight” tasks and skipping interruption until the new session starts. Closes #4528.

- **Bug Fixes**
  - Add `retryInFlightTasks` set in `BackgroundManager` to track tasks queued for fallback retry.
  - When `tryFallbackRetry` returns true, add the task ID; remove it when `startTask` begins the retry.
  - Pass `retryInFlightTaskIds` to `checkAndInterruptStaleTasks` and have `interruptStaleTask` skip those IDs.

- **Testing**
  - Added 3 tests covering skip behavior, normal interruption when not in the set, and backward compatibility when the set is absent.

<sup>Written for commit 08bb97db7e7464a7b127be08733410e634308ed7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4583?utm_source=github">Review in cubic</a>

### DIFF
--- a/src/features/background-agent/fallback-retry-race.test.ts
+++ b/src/features/background-agent/fallback-retry-race.test.ts
@@ -1,0 +1,129 @@
+declare const require: (name: string) => any
+const { describe, it, expect, mock, spyOn, beforeEach, afterEach } = require("bun:test")
+
+import { checkAndInterruptStaleTasks } from "./task-poller"
+import type { BackgroundTask } from "./types"
+
+describe("checkAndInterruptStaleTasks", () => {
+  describe("#given a task with fallback retry in flight", () => {
+    const mockClient = {
+      session: {
+        abort: mock(() => Promise.resolve()),
+        get: mock(() => Promise.resolve({ data: { id: "ses-1" } })),
+      },
+    }
+    const mockConcurrencyManager = {
+      release: mock(() => {}),
+    }
+    const mockNotify = mock(() => Promise.resolve())
+    const mockOnTaskInterrupted = mock(() => {})
+
+    const originalDateNow = Date.now
+    let fixedTime: number
+
+    beforeEach(() => {
+      fixedTime = Date.now()
+      spyOn(globalThis.Date, "now").mockReturnValue(fixedTime)
+      mockClient.session.abort.mockClear()
+      mockClient.session.get.mockReset()
+      mockClient.session.get.mockResolvedValue({ data: { id: "ses-1" } })
+      mockConcurrencyManager.release.mockClear()
+      mockNotify.mockClear()
+      mockOnTaskInterrupted.mockClear()
+    })
+
+    afterEach(() => {
+      Date.now = originalDateNow
+    })
+
+    function createStaleRunningTask(overrides: Partial<BackgroundTask> = {}): BackgroundTask {
+      return {
+        id: "task-retry-1",
+        sessionId: "ses-retry-1",
+        parentSessionId: "parent-ses-1",
+        parentMessageId: "msg-1",
+        description: "test fallback retry race",
+        prompt: "test",
+        agent: "explore",
+        status: "running",
+        startedAt: new Date(fixedTime - 600_000),
+        progress: {
+          toolCalls: 5,
+          lastUpdate: new Date(fixedTime - 300_000),
+        },
+        ...overrides,
+      }
+    }
+
+    describe("#when the task id is in retryInFlightTaskIds", () => {
+      it("#then should skip the stale interrupt", async () => {
+        // given
+        const task = createStaleRunningTask()
+        const retryInFlightTaskIds = new Set(["task-retry-1"])
+
+        // when
+        await checkAndInterruptStaleTasks({
+          tasks: [task],
+          client: mockClient as never,
+          config: { staleTimeoutMs: 180_000 },
+          concurrencyManager: mockConcurrencyManager as never,
+          notifyParentSession: mockNotify,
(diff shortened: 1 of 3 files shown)


MoerAI (changes requested):
Thanks for the PR, @Hungdoan565, and for taking the time on this. The core race fix is right: keying the guard by `task.id` (since a fallback retry keeps the same logical task and only swaps the session) correctly stops the stale-interrupt from killing the old session while a retry is in flight. The refactor merging the two `if (retried && ...)` blocks is behavior-equivalent.

The blocker is the lifecycle of the new `retryInFlightTasks` Set in `src/features/background-agent/manager.ts`. Right now it only gets an `add(task.id)` on retry and a single `delete(task.id)` on the task-start path. There are paths where an id is added but never removed:
- a retried task cancelled while still pending
- a retried task TTL-pruned while still pending
- the manager shutting down before the task restarts

In those cases the Set grows without bound, and a leaked id could make a future task wrongly skip its own stale-interrupt check. The 30-minute TTL prune is a backstop so it is not a permanent wedge, but the Set should still drain on every terminal path.

- Add `retryInFlightTasks.delete(task.id)` on the pending-cancel and pending-prune paths, and clear the whole Set in `shutdown()`.
- Add a regression test for "retry queued then cancelled before restart" and one for "retry queued then TTL-pruned" proving the Set drains.

Happy to re-review once the Set has symmetric cleanup. No rush.

code-yeongyu:
[sisyphus-bot] This PR hasn't been updated since the changes-requested review on 2026-06-05. Dev's task-poller has since been reworked with a different stale-task pruning approach, and the source has moved to packages/omo-opencode/src/. Issue #4528 is still active. Closing as stale — please reopen with a rebase if the specific race condition still applies to the current implementation.

PR #4602, opened 2026-05-29
fix(handoff): bind USER REQUESTS to session_read output instead of in-context memory (fixes #4592)

### LINKED ISSUE #4592: [Bug]: /handoff command reconstructs session history from AI memory instead of session_read, leading to inaccurate first user request
### Prerequisites

### Bug Description

The `/handoff` builtin command generates session history by relying on the AI's in-context memory rather than calling `session_read` to fetch the actual session messages.

This causes the "first user request" field in the handoff output to be inaccurate.

### Steps to Reproduce

1. Start a long session with multiple exchanges
2. Run `/handoff`
3. Check the "first user request" in the handoff output — it does not match the actual first message in the session

### Expected Behavior

`/handoff` should call `session_read` to retrieve the actual first user message before generating the handoff document.

### Actual Behavior

The first user request in the handoff output reflects what the AI "remembers" from context, not what was actually typed at the start of the session. This is especially wrong when the session is long (early messages scrolled out of context window) or after `/compact` (early content summarized away).

### Doctor Output

```shell
oh-my-openagent is loaded as an OpenCode plugin (not installed as standalone CLI)
OpenCode version: 1.15.12
OS: Windows
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Windows

### OpenCode Version

1.15.12

MoerAI:
## Summary

The `/handoff` builtin command rendered an inaccurate first user request in long sessions and post-`/compact` sessions because the prompt allowed the agent to reconstruct USER REQUESTS (AS-IS) from in-context memory rather than the actual session history. This PR inserts an explicit PHASE 0.5 that calls `session_read` first and quotes the earliest `role: "user"` message verbatim into the USER REQUESTS field, with a new regression test pinning the contract.

## Root Cause

`src/features/builtin-commands/templates/handoff.ts` already mentioned `session_read` in PHASE 1 and said USER REQUESTS "must be verbatim (do not paraphrase)" in PHASE 2, but it never required the agent to derive USER REQUESTS specifically from `session_read` output. Two mechanisms let the agent satisfy "verbatim" while still being wrong:

1. Long sessions: early messages scroll out of the context window, so "verbatim" gets sourced from a remembered paraphrase.
2. `/compact`: early content gets summarized away entirely, so the agent quotes the post-compaction summary as if it were the original wording.

Issue #4592 reports both as the user-visible symptom.

## Changes

| File | Change |
|------|--------|
| `src/features/builtin-commands/templates/handoff.ts` | Insert new PHASE 0.5 ("SESSION READ FIRST") that calls `session_read` and extracts the first `role: "user"` message verbatim before any other context gathering. PHASE 1 now reuses that output instead of re-deriving USER REQUESTS. PHASE 2 bullet for USER REQUESTS now points back to the PHASE 0.5 extraction and explicitly forbids memory reconstruction. |
| `src/features/builtin-commands/commands.test.ts` | Two new regression tests under `describe("HANDOFF_TEMPLATE")`: one asserts the prompt contains the "do not reconstruct from memory" guardrail and references the "first user message"; the other asserts `session_read` is mentioned in the template before any USER REQUESTS extraction step. |

Diff: +49 / -4 across 2 files. No behavior change for other handoff sections; the format block (PHASE 3) is unchanged.

## Reproduction (before fix)

The new regression tests fail on the previous template:

~~~
(fail) HANDOFF_TEMPLATE > should mandate session_read as the source of USER REQUESTS verbatim (#4592)
Expected: true
Received: false
  at .../commands.test.ts:449:27
 44 pass
 1 fail
 94 expect() calls
~~~

The previous template contained no language forbidding memory-driven extraction and no instruction to use the first `role: "user"` entry from `session_read`. The agent could (and per the issue, did) fill USER REQUESTS (AS-IS) from in-context memory.

## Verification (after fix)

Same suite, same machine:

~~~
bun test src/features/builtin-commands/commands.test.ts
 45 pass
 0 fail
 95 expect() calls
Ran 45 tests across 1 file. [202.00ms]
~~~

Both new contract tests pass:
- `should mandate session_read as the source of USER REQUESTS verbatim (#4592)` -> green
- `should sequence session_read BEFORE the USER REQUESTS extraction step (#4592)` -> green

## Test

- Regression tests: `src/features/builtin-commands/commands.test.ts` (2 new cases marked `(#4592)`)
- Related suite: `bun test src/features/builtin-commands/` -> **48 pass / 0 fail** (45 in `commands.test.ts` + 3 in adjacent template tests)
- Typecheck: `bun run typecheck` -> clean

## Design notes

- PHASE 0.5 is intentionally a small, distinct phase before PHASE 1 rather than a bullet under PHASE 1, so the "do `session_read` first" instruction sits at the top of the agent's attention before it starts the full context-gathering sequence.
- The fix keeps the existing PHASE 1 instruction to call `session_read` so the legacy execution order is still valid; PHASE 0.5 just promotes the first call and binds its output to USER REQUESTS.
- No new dependencies, no new template files, no schema or config changes.

Fixes #4592

---
## Summary by cubic
Fixes #4592 by making `/handoff` always quote the first user request from `session_read` instead of in-context memory, preventing wrong or paraphrased text in long or post-`/compact` sessions.

- **Bug Fixes**
  - Added PHASE 0.5 to run `session_read` first and copy the earliest `role: "user"` message verbatim; forbid reconstruction from memory and paraphrasing.
  - Updated PHASE 1 and PHASE 2 to reuse PHASE 0.5 output and explicitly disallow re-deriving USER REQUESTS.
  - Added two regression tests to pin the sequencing (`session_read` before USER REQUESTS) and the “do not reconstruct from memory” guardrail.

<sup>Written for commit 6ede73356e448ab7e005e1302200ef74be6e8d1c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4602?utm_source=github">Review in cubic</a>

### DIFF
--- a/src/features/builtin-commands/commands.test.ts
+++ b/src/features/builtin-commands/commands.test.ts
@@ -438,4 +438,30 @@ describe("HANDOFF_TEMPLATE", () => {
     const emojiRegex = /[\u{1F600}-\u{1F64F}\u{1F300}-\u{1F5FF}\u{1F680}-\u{1F6FF}\u{1F1E0}-\u{1F1FF}\u{2702}-\u{27B0}\u{24C2}-\u{1F251}\u{1F900}-\u{1F9FF}\u{1FA00}-\u{1FA6F}\u{1FA70}-\u{1FAFF}\u{2600}-\u{26FF}\u{2700}-\u{27BF}]/u
     expect(emojiRegex.test(HANDOFF_TEMPLATE)).toBe(false)
   })
+
+  test("should mandate session_read as the source of USER REQUESTS verbatim (#4592)", () => {
+    //#given - the template string
+
+    //#when - check the prompt anchors USER REQUESTS extraction to session_read output, not memory
+    const forbidsMemory = /do not (reconstruct|rely on|recall|use) (your )?(in-context )?memory/i.test(HANDOFF_TEMPLATE)
+
+    //#then - the prompt explicitly forbids reconstructing USER REQUESTS from memory
+    expect(forbidsMemory).toBe(true)
+
+    //#and - the prompt explicitly anchors USER REQUESTS extraction to the first user message in session_read output
+    expect(HANDOFF_TEMPLATE).toContain("first user message")
+  })
+
+  test("should sequence session_read BEFORE the USER REQUESTS extraction step (#4592)", () => {
+    //#given - the template string
+
+    //#when - record the textual order of session_read instruction and USER REQUESTS extraction
+    const sessionReadFirstMention = HANDOFF_TEMPLATE.indexOf("session_read")
+    const userRequestsFirstMention = HANDOFF_TEMPLATE.indexOf("USER REQUESTS")
+
+    //#then - both anchors exist AND session_read is mentioned first
+    expect(sessionReadFirstMention >= 0).toBe(true)
+    expect(userRequestsFirstMention >= 0).toBe(true)
+    expect(sessionReadFirstMention < userRequestsFirstMention).toBe(true)
+  })
 })
--- a/src/features/builtin-commands/templates/handoff.ts
+++ b/src/features/builtin-commands/templates/handoff.ts
@@ -21,11 +21,29 @@ If the session is nearly empty or has no meaningful context, inform the user the
 
 ---
 
+# PHASE 0.5: SESSION READ FIRST (MANDATORY FIRST DATA STEP)
+
+Call session_read({ session_id: "$SESSION_ID" }) BEFORE any other context gathering step. The output of session_read is the only authoritative source for what the user originally asked. Do not reconstruct memory of the first request; long sessions and post-compact sessions truncate or summarize early messages, so in-context memory is unreliable.
+
+From the session_read output:
+
+1. Find the first user message in the returned session history (the earliest entry with role "user").
+2. Copy the text of that first user message verbatim into a working note. You will later place it into the USER REQUESTS (AS-IS) section unchanged.
+3. If the user has sent multiple distinct top-level requests in the session, also collect each subsequent verbatim user message that is a new top-level ask (not a follow-up clarification).
+
+Rules:
+- Do not reconstruct user requests from memory. Always quote from session_read output.
+- Do not paraphrase, summarize, or "tidy up" the user's wording.
+- Do not skip session_read even if you feel you remember the first user message.
+- If session_read fails or returns no user messages, state that explicitly in the USER REQUESTS (AS-IS) section rather than guessing.
+
+---
+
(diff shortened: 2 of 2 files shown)


MoerAI:

MoerAI:
Maintainer note (self-PR): the `/handoff` PHASE 0.5 fix for #4592 looks right and the regression tests pin the contract well. One nit: the diff is 53 LOC, which is just barely past the 50-line target for `fix()` PRs in `AGENTS.md`. Worth a quick look at whether the duplicated guardrail text between PHASE 0.5, PHASE 1, and PHASE 2 can be tightened. Leaving as a comment rather than request-changes since I cannot request changes on my own PR.

code-yeongyu:
[sisyphus-dev] I reviewed this while sweeping #4592. I am not merging or replacing it automatically because the fix changes `/handoff` command behavior by enforcing a `session_read`-first workflow. That may be the correct product direction, but it falls under the current behavior-change skip rule and needs maintainer decision.

code-yeongyu:
[sisyphus-dev] Reviewed this during the bug sweep for #4592. The report is plausible and the PR is focused, but it changes the `/handoff` command contract by requiring a `session_read`-first workflow before synthesis. Because this is a behavior-surface change rather than a simple localized bug fix, and cubic is neutral rather than an unconditional approval, I am not merging or replacing it automatically. Leaving this for maintainer decision.

MoerAI:
Thanks for the careful review. One clarification on the behavior-change concern: the `session_read`-first workflow isn't an unsolicited contract change — it is exactly the Expected Behavior stated in #4592:

> `/handoff` should call `session_read` to retrieve the actual first user message before generating the handoff document.

The current template lets the model reconstruct the first user request from in-context memory, which is the root cause of the inaccurate "first user request" field reported in the issue. The diff is template + regression test only (2 files, +49/-4); no tool surface, command name, or output schema changes. So the user-visible contract (what `/handoff` produces) becomes *more* accurate, not different in shape.

Happy to adjust wording or scope if you'd prefer a narrower phrasing in the template. Either way, no rush — leaving the merge decision to you.

MoerAI:
Thanks for the review. On the behavior-change concern: the `session_read`-first workflow is not new product direction I introduced, it is precisely what issue #4592 specifies under Expected Behavior: `/handoff should call session_read to retrieve the actual first user message before generating the handoff document`. The bug is that the current command reconstructs the first user request from in-context memory, which is wrong on long sessions (early messages scrolled out) and after /compact. So this change restores the documented/expected contract rather than redefining it. Happy to adjust scope if you'd prefer a narrower touch.

MoerAI:
Thanks @MoerAI, merging.

PR #4610, opened 2026-05-29
fix(team-mode): guard team activation against unresolved members (#4590)

### LINKED ISSUE #4590: team_create can return active while non-lead members remain pending without sessionId
## Summary

`team_create` / Hyperplan Team Mode can report a team runtime as `active` even when one or more non-lead members remain `pending` and have no `sessionId`.

This makes the runtime appear usable while the actual worker member was never started or assigned a session.

## Reproduction Evidence

Before a local patch, a Hyperplan/team_create smoke test reproduced:

- runtime status: `active`
- non-lead member status: `pending`
- non-lead member `sessionId`: missing/null

The observed failure was in a local installed dist bundle, not verified directly against upstream source.

## Observed Local Root Cause

In the installed dist bundle, `createTeamRun` appeared to update runtime members by array index:

- `runtimeState.members` includes the lead at index `0`
- `spec.members` contains only non-lead members
- matching by `currentIndex === memberIndex` can assign a child session to the lead slot
- the actual non-lead member can remain unresolved while the runtime transitions to `active`

## Expected Behavior

A team runtime should not transition to `active` until every required non-lead member has either:

- a valid `sessionId`
- an appropriate running/started status
- or a terminal/error state that prevents an `active` result

## Minimal Fix Direction

The local patch that resolved the smoke test used these invariants:

- update runtime members by stable identity, e.g. `currentMember.name === member.name`, not array index
- add an unresolved-member guard before active transition, e.g. `getUnresolvedTeamMembers` / `assertNoUnresolvedTeamMembers`
- avoid reusing an `active` runtime if it still has unresolved members

This is a proposed invariant/fix direction based on local installed-bundle observation, not a claim that the exact upstream source line is identical.

## Validation

After applying the local fix and restarting:

- one-member team returned `active`
- `smoke-worker` had its own `sessionId`
- `smoke-worker` status was `running`
- team closed cleanly

Hungdoan565:
﻿## Summary

Fixes #4590. `team_create` could report a runtime as `active` while a non-lead member was still `pending` with no `sessionId`.

## Root cause

Two independent weaknesses let an unresolved member slip through:

1. The three create-time member transitions matched runtime members by **array index** (`currentIndex === memberIndex`). This couples runtime-state ordering to `spec.members` ordering. The reporter observed a child session being written into the wrong slot, leaving the real member unresolved.
2. There was **no guard** before the final `status: 'active'` transition, so a run could go active regardless of member resolution state.

## Fix

- Match runtime members by **stable name identity** (`currentMember.name === member.name`) in all three transitions, removing the index-coupling fragility. This mirrors the existing name-based matching already used for the lead-session branch.
- Add `assertNoUnresolvedTeamMembers` before the active transition. A member is *resolved* only if it has a `sessionId` **or** a terminal status (`errored` / `completed` / `shutdown_approved`). If any member is unresolved, the guard throws, hitting the existing `cleanupTeamRunResources` path so the run ends **failed** instead of falsely **active**.

## Tests

- New `unresolved-team-members.test.ts`: 6 given/when/then cases covering `isTeamMemberResolved` / `getUnresolvedTeamMembers` / `assertNoUnresolvedTeamMembers`.
- Existing `create.test.ts` still green (12 pass) — the name-based conversion preserves spec-order mapping.
- `bun run typecheck` clean.

---
## Summary by cubic
Prevents activation with unresolved members and switches create-time updates to name-based matching; if an existing runtime is unresolved we start a new run. Fixes #4590.

- **Bug Fixes**
  - Guard unresolved members (missing `sessionId`) before activation and when looking up an existing runtime; fail and clean up instead of reporting active or reusing.
  - Match members by stable name in create-time transitions to avoid index-coupling and wrong-slot writes.

<sup>Written for commit be77a871abac99b6068528f0452b9e2b02e13dcf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4610?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/team-mode/team-runtime/create.test.ts
+++ b/src/features/team-mode/team-runtime/create.test.ts
@@ -11,7 +11,7 @@ import { TeamModeConfigSchema } from "../../../config/schema/team-mode"
 import type { ExecutorContext } from "../../../tools/delegate-task/executor-types"
 import type { BackgroundTask, LaunchInput } from "../../background-agent/types"
 import { BackgroundManager } from "../../background-agent/manager"
-import { loadRuntimeState } from "../team-state-store/store"
+import { loadRuntimeState, transitionRuntimeState } from "../team-state-store/store"
 import { clearTeamSessionRegistry, lookupTeamSession } from "../team-session-registry"
 import type { TeamSpec } from "../types"
 import {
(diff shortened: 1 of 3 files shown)


MoerAI (dismissed):
Thanks @Hungdoan565.

The team-mode unresolved-member guard is a sensible fix, but the diff is 135 LOC across 3 files (`src/features/team-mode/team-runtime/create.ts`, `unresolved-team-members.ts`, plus tests). That is past the 50-line target for `fix()` PRs in `AGENTS.md`.

Could you narrow the production change to just `create.ts` and the minimal guard helper, and keep the broader unresolved-team-members extraction as a separate follow-up? Smaller fix() PRs land faster.

Leaving open as request-changes, no rush.

code-yeongyu (approved):
Reviewed the refreshed head be77a871. The diff is now narrowed to the team-runtime create path plus a 15-line guard helper, with the broad unresolved-member helper test extraction removed. Local verification passed: bun test src/features/team-mode/team-runtime/create.test.ts, bun run typecheck, and bun run build.

PR #4611, opened 2026-05-29
feat(git-master): add enforce_atomic_commits config to relax mandatory commit splitting (#4483)

### LINKED ISSUE #4483: Git-master skill's mandatory commit splitting has no escape hatch
### Feature request

The `git-master` skill enforces a strict commit splitting formula (`min_commits = ceil(file_count / 3)`, different directories must be separate commits, etc.). While atomic commits are a good practice in general, this rule does not account for cases where splitting by directory or file count is counterproductive.

### Concrete example

Updating locale/translation files (e.g., `messages/en.json`, `messages/zh.json`, `messages/ko.json`, `messages/fr.json`). These are identical in nature — the same change applied across languages. The current rules would force splitting them into separate commits ("European languages" / "CJK languages"), which adds no value and clutters the history.

### Problem

The splitting behavior is currently **not configurable**. `GitMasterConfigSchema` only exposes `commit_footer`, `include_co_authored_by`, and `git_env_prefix`. There is no option to relax or disable the atomic commit enforcement.

### Request

Add a configuration option (e.g., `enforce_atomic_commits: boolean`, default `true`) to allow users to opt out of the mandatory splitting when it does not fit their workflow.

Hungdoan565:
﻿## Summary

Fixes #4483. The git-master skill enforces mandatory atomic commit splitting with a blocking `ceil(file_count / 3)` minimum-commit formula and no escape hatch. For changes where multiple files genuinely belong in one commit — e.g. updating parallel locale files (`messages/en.json`, `zh.json`, `ko.json`, `fr.json`) — forced splitting is just noise.

## Change

Add an `enforce_atomic_commits` boolean to `git_master` config, **default `true`** so current behavior is unchanged.

When set to `false`, `injectGitMasterConfig` replaces the `<atomic_planning>` enforcement block in the skill template with a relaxed version:

- group commits by logical concern instead of a file-count formula
- tightly-coupled files (locale/i18n sets) MAY share a commit
- still pair tests with their implementation
- no blocking minimum-commit-count plan

The flag flows from `pluginConfig.git_master` through the existing `SkillResolutionOptions.gitMasterConfig` path into `injectGitMasterConfig` — no new plumbing.

## Tests

- New `#given enforce_atomic_commits config` block in `git-master-template-injection.test.ts` (3 given/when/then cases): omitted → enforcement kept; explicit `true` → enforcement kept; `false` → enforcement block replaced with relaxed text, surrounding template (Phase 4, closing tag) preserved.
- `bun run typecheck` clean.
- `assets/oh-my-opencode.schema.json` regenerated via `bun run build:schema`.

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -6518,12 +6518,17 @@
         "git_env_prefix": {
           "default": "GIT_MASTER=1",
           "type": "string"
+        },
+        "enforce_atomic_commits": {
+          "default": true,
+          "type": "boolean"
         }
       },
       "required": [
         "commit_footer",
         "include_co_authored_by",
-        "git_env_prefix"
+        "git_env_prefix",
+        "enforce_atomic_commits"
       ],
       "additionalProperties": false
     },
--- a/packages/omo-opencode/src/config/schema/git-master.ts
+++ b/packages/omo-opencode/src/config/schema/git-master.ts
@@ -9,6 +9,8 @@ export const GitMasterConfigSchema = z.object({
   include_co_authored_by: z.boolean().default(true),
   /** Environment variable prefix for all git commands (default: "GIT_MASTER=1"). Set to "" to disable. Allows custom git hooks to detect git-master skill usage. */
   git_env_prefix: GitEnvPrefixSchema,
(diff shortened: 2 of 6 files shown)


code-yeongyu:
[sisyphus-bot] Keeping this open — issue #4483 is still open and unfixed, and dev does not yet have any enforce_atomic_commits config. The PR is small and focused. @Hungdoan565 could you rebase on current dev when you get a chance? The config schema file may have moved under packages/omo-opencode/.

Issue #4589, opened 2026-05-28
[Feature]: Prevent main-session explore/librarian launches alongside plan


ririnto:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues and discussions to avoid duplicates
- [x] This feature request is specific to oh-my-opencode (not OpenCode core)
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Problem Description

`search-mode` and `analyze-mode` can instruct the main orchestrator to launch `explore` and `librarian` background agents in parallel.

For example, `search-mode` says:

```text
MAXIMIZE SEARCH EFFORT. Launch multiple background agents IN PARALLEL:
- explore agents (codebase patterns, file structures, ast-grep)
- librarian agents (remote repos, official docs, GitHub examples)
```

`analyze-mode` has a similar context-gathering instruction:

```text
CONTEXT GATHERING (parallel):
- 1-2 explore agents (codebase patterns, implementations)
- 1-2 librarian agents (if external library involved)
```

That is useful when the main orchestrator is doing research directly. However, when the same user request also triggers `plan`, the parent-level research calls can become redundant.

The `plan` subagent already receives explicit startup instructions to gather context by calling `call_omo_agent` for `explore` and `librarian`.

The plan-agent prepend currently includes a mandatory context-gathering protocol equivalent to:

```text
Launch background agents to gather context:
- call_omo_agent(..., subagent_type="explore", run_in_background=true, ...)
- call_omo_agent(..., subagent_type="librarian", run_in_background=true, ...)
```

So a single main-session request can produce a mixed wave like:

```text
main Sisyphus
  -> task(subagent_type="explore", run_in_background=true)
  -> task(subagent_type="librarian", run_in_background=true)
  -> task(subagent_type="plan", run_in_background=false)
```

Then the `plan` session may start and perform its own context gathering:

```text
plan
  -> call_omo_agent(subagent_type="explore", run_in_background=true)
  -> call_omo_agent(subagent_type="librarian", run_in_background=true)
```

This is related to #2124, but the request here is not just to improve prompt-following. It is to add a runtime guardrail that prevents redundant parent-level research subagents from being launched alongside `plan`.

[...]

Issue #4163, opened 2026-05-18
[Bug] todo-continuation can resume parent session before descendant background tasks finish


ririnto:
## Prerequisites

- [x] I am writing this issue in English.
- [x] I searched for existing issues and found related but not exact matches.
- [ ] I am using the latest version.
- [ ] I checked the documentation.
- [ ] I ran `bunx oh-my-openagent doctor` and included the output below.

## Bug Description

`todo-continuation-enforcer` can resume a parent session after a fresh parent user message even though related background work has not fully quiesced or delivered its parent wake.

OpenCode idle/status events are per-session. A parent session can legitimately become idle after handling a new user message while background/subagent sessions spawned from that parent are independently running, completing, waiting on their own todos, or waiting on nested `call_omo_agent` / `task` work.

The existing direct-child task guard should correctly block parent continuation while a direct child background task is still `running` or `pending`. The suspected gap is narrower: parent continuation can still run once the direct child task no longer appears active, even though the direct child session is not semantically quiescent yet, nested work launched by that child is still active, or the completion/parent-wake delivery has not been accepted by the parent session.

A likely failure mode is:

1. Parent session `P` launches subagent/background session `A`.
2. `A` launches nested work `B` using `call_omo_agent` or `task`, or `A` has incomplete todos that its own todo-continuation must finish.
3. The user sends another message to `P` while `A`/`B` is not fully quiescent.
4. OpenCode marks `P` idle after `P` answers that user message.
5. If `A` is no longer `running`/`pending` from the parent gate's point of view, `todo-continuation-enforcer` can evaluate `P` and inject a continuation before nested results or parent wake/completion messages are available.

This should not become a blanket recursive descendant-task block. Subagent/background task todos are session-scoped and may need their own continuation to manage child work. The safer boundary is: parent continuation should wait for direct child session quiescence and parent wake delivery, while each child session remains responsible for its own todos.

## Steps to Reproduce

1. Use `opencode web` with a main session `P` that has incomplete todos.
2. Launch a background/subagent task from `P`, creating child session `A`.
3. Let `A` launch nested work `B` through `call_omo_agent` or `task`, or leave `A` with incomplete todos that should be continued in `A`.
4. Send another user message to `P` while `A` or `B` has not fully quiesced.
5. Let `P` finish that user turn and become idle.
6. Observe `todo-continuation-enforcer` injecting a synthetic continuation into `P` before nested work or parent wake delivery has settled.

## Expected Behavior

[...]

PR #2328, opened 2026-03-05
fix(agents): prevent orchestrator from rushing ahead of background agents

mrosnerr:
## Summary

- Tighten Sisyphus prompt's Background Result Collection instructions so the model waits for explore/librarian results instead of duplicating their work and delivering premature answers.

## Changes

- **`src/agents/sisyphus.ts`**: Remove `Continue working immediately` instruction (which models interpreted as "do ALL work yourself, ignore agents"). Restructure Background Result Collection steps to: (1) launch agents, (2) do DIFFERENT independent work if any, (3) otherwise END RESPONSE and wait for system-reminder, (4-6) collect results and clean up.
- **`src/agents/dynamic-agent-prompt-builder.ts`**: Add `not for files you already know` to explore section header to reduce duplicate reads when the parent already knows file locations.

## Testing

Tested locally with 3 models via `opencode run --format json`:

| Scenario | Claude Opus 4.6 | Gemini 2.5 Pro | GPT-4.1 |
|---|---|---|---|
| Broad discovery (unknown files) | ✅ Fires bg agents, waits | ✅ 4 bg agents, waits | ✅ 2 bg agents, waits |
| Known directory question | ✅ Direct tools only | ✅ Direct tools only | ✅ Direct tools only |

```bash
bun run typecheck  # ✅ passes
bun test           # ✅ 639 tests pass
bun run build      # ✅ succeeds
```

## Related Issues

Closes #2124
Closes #1967

---
## Summary by cubic
Make the orchestrator wait for explore/librarian background agents before proceeding, preventing duplicate work and premature answers (fixes #2124, #1967). Also scope the Explore Agent to discovery-only when file locations are already known.

- **Bug Fixes**
  - Sisyphus prompts (all variants): removed “Continue working immediately”; clarified to only do different independent work while waiting, otherwise end response and wait for the system reminder; then collect via `background_output` and clean up with `background_cancel`.
  - Explore prompt builder: added “not for files you already know” to reduce duplicate reads.

<sup>Written for commit f1f682c3abcde867b60c189dfcd83a721eae1ed9. Summary will update on new commits.</sup>

PR #4617, opened 2026-05-30
fix(delegate-task): wait for child background tasks before completing sync delegation (#4615)

### LINKED ISSUE #4615: [Bug] Sync subagent delegation returns to parent before its background descendants (explore/librarian) quiesce
## Prerequisites

- [x] I searched existing issues and found related but not exact matches.
- [x] I am using the latest version of oh-my-openagent (4.5.1).
- [x] I read the documentation (docs/guide/orchestration.md) and couldn't find the answer.

## Bug Description

A sync subagent delegation returns to its parent before the background descendants it launched have quiesced.

Case: main delegates `task(subagent_type="plan", run_in_background=false)`. `plan` runs its mandatory protocol and launches `explore` / `librarian` via `call_omo_agent(..., run_in_background=true)`. If `plan` reaches a terminal turn before those finish, the parent's sync poller declares `plan` complete and returns immediately — so the parent gets a plan that never synthesized the explore/librarian findings.

The notification path is not the problem: parent-wake (`parent-wake-notifier.ts`, recently refined) does deliver completion wakes to the `plan` session. It is an ordering bug — the main-side sync poller returns the premature result before parent-wake can resume `plan`.

This is the runtime layer of #2124 (addressed only at the prompt layer by #2328, which covered Sisyphus but not `plan`), and the same quiescence gap as #4163.

## Steps to Reproduce

1. Delegate `task(subagent_type="plan", run_in_background=false)` from a main session.
2. Let `plan` launch `explore` / `librarian` via `call_omo_agent(..., run_in_background=true)`.
3. Let `plan` reach a terminal turn before they finish.
4. Parent receives the plan result while explore/librarian are still running; their findings are not synthesized.

## Expected Behavior

While a sync subagent owns direct child background tasks in `pending`/`running`, the parent delegation should stay non-quiescent until they finish and the subagent emits a terminal turn after collecting them. This should apply to all sync subagents (not just `plan`), and stay bounded since descendants carry their own timeouts.

## Actual Behavior

The sync poller decides completion from the subagent's terminal assistant finish alone, never consulting the BackgroundManager for in-flight descendants, so it returns early.

## Two Code Paths (why notifications work but the parent returns early)

Background completion is handled by two independent paths that never meet:

| Path | Trigger | Knows about background descendants? | Effect on the parent delegation |
| --- | --- | --- | --- |
| Parent-wake (notification) | A background task reaches a terminal state, gated on `validateSessionHasOutput` (real response arrived) + no incomplete todos | Yes — driven by task completion | Async-injects a system-reminder into the parent session so it could resume. Recently refined (defer during active turns, unblock stale wakes, align activity windows). |
| Sync completion (poller) | The polled subagent emits a terminal assistant turn | No — purely session-local, no output/descendant cross-check | Breaks the poll loop and returns the subagent result to its caller |

The asymmetry is the core issue: the background path waits for real output before notifying, while the sync path returns on terminal finish alone. Because the two are decoupled, the sync poller fires first on `plan`'s initial terminal turn and returns the premature result; the parent-wake that would have resumed `plan` arrives afterward, against a delegation that has already concluded. Improving parent-wake therefore cannot fix this — the completion decision has to consult descendant state.

## Root Cause Analysis

The sync completion check is session-local and is never given a BackgroundManager reference; the parent-wake path is async and not consulted by the completion decision.

Evidence:

- `sync-session-poller.ts` — `isSessionComplete()` returns true from the session's own terminal finish / pending tool parts only; the poll loop breaks immediately, with no descendant check.
- `sync-task.ts` — `executeSyncTask()` holds `executorCtx.manager` but never passes it to `pollSyncSession(...)`; after the poll it fetches the result and returns with no continuation.
- `completion-poller.ts` — `waitForCompletion()` has the same gap on the `call_omo_agent` sync path (polls only target session status/messages).
- `parent-wake-notifier.ts` — delivers wakes asynchronously and never checks `subagentSessions` / `syncSubagentSessions`, so it cannot hold the sync poller open.
- `manager.ts` — the needed predicate already exists: `getTasksByParentSession()` / `getAllDescendantTasks()`, and `updateBackgroundTaskMarker()` already treats `running || pending` as active; the background path's own completion is gated by `validateSessionHasOutput()`.
- `constants.ts` — `PLAN_AGENT_SYSTEM_PREPEND_STATIC_BEFORE_SKILLS` tells `plan` to launch the agents but not to wait for and collect their results (unlike Sisyphus after #2328).

## Suggested Fix Surface

Two layers, mirroring the runtime/prompt split:

1. Sync completion check (deterministic): make it descendant-aware. `sync-session-poller.ts` should not treat the session as complete while its direct child background tasks are `pending`/`running` (resetting the inactivity timer during that wait); `sync-task.ts` should pass `executorCtx.manager` to the poller; apply the same to `completion-poller.ts` for the `call_omo_agent` sync path. Bound the wait and limit it to direct children (not a recursive descendant block, per #4163).
2. `plan` prompt (reinforcement): instruct `plan` to wait for and collect explore/librarian results via `background_output` before producing the plan, as #2328 did for Sisyphus.

The runtime gate keeps the parent from returning early (giving parent-wake time to resume `plan`); the prompt ensures the resumed `plan` actually collects and synthesizes. Parent-wake itself needs no change.

## Related Issues

- #2124 / #2328 — prompt-layer attempt for the same symptom.
- #4163 — equivalent quiescence gap in `todo-continuation-enforcer`.
- #3774 — completed-vs-timeout mismatch on the `call_omo_agent` waiter (the non-idle completion paths that skip the output gate).
- #4589 / #4609 — complementary: avoid redundant main-session explore/librarian alongside `plan`.

## Additional Context

Based on behavior directly observed while using omo, combined with source analysis of the `dev` branch.

ririnto:
## Summary

- A synchronous subagent delegation (e.g. `plan`, delegated with `run_in_background=false`) returned to its parent as soon as it emitted a terminal turn, even while the `explore`/`librarian` background agents it launched via `call_omo_agent(run_in_background=true)` were still running — so the parent received a plan that never synthesized those results.
- The sync completion pollers now wait for the subagent's in-flight direct child background tasks before reporting completion, so the subagent can collect and synthesize their results first.

## Changes

- Add `BackgroundManager.hasActiveChildTasks(sessionID)` — reports whether a session has direct child tasks in `running`/`pending` (direct children only, not the full descendant tree).
- Add `src/shared/child-task-deferral.ts` — a small shared state machine used by both sync pollers (`delegate-task/sync-session-poller.ts` and `call-omo-agent/completion-poller.ts`):
  - While direct child tasks are active → defer completion.
  - Once they clear → hold until the subagent emits a fresh terminal turn (a genuine post-resume synthesis) or a bounded grace window (`CHILD_WAKE_GRACE_MS`) elapses, whichever comes first.
  - Resume is detected by a new assistant-message id (not raw message growth), so the user-role system-reminder that the parent-wake injects is not mistaken for synthesis.
- Wire the predicate from `sync-task.ts` and `call-omo-agent/tools.ts`; degrade safely (with a warning) to prior behavior when the manager predicate is unavailable.
- Reinforce the `plan` prompt (`delegate-task/constants.ts`) to wait for and collect the explore/librarian results via `background_output` before planning — the runtime layer that #2124/#2328 only addressed via the Sisyphus prompt.

### Notes

- The notification (parent-wake) path is unchanged and already correct; this is purely the sync completion decision, which previously never consulted descendant state. See #4615 for the two-path analysis.
- The wait while children are active is bounded by the `BackgroundManager`'s existing stale/`sessionGone`/crash timeouts, and the post-clear wait by the grace window; no new absolute cap is added so legitimately long-running children are not cut off.
- Reviewed across security, functional-correctness, and scope dimensions before submitting.

## Testing

```bash
bun run typecheck
bun run build
bun test
```

- New unit tests for the shared deferral state machine (`src/shared/child-task-deferral.test.ts`) and for both pollers covering: defer while children active; hold through the wake-pending window then complete on a fresh synthesis turn; bounded completion when no resume occurs.

## Related Issues

Closes #4615

Related: #2124 / #2328 (prompt-layer attempt for the same symptom), #4163 (equivalent quiescence gap in `todo-continuation-enforcer`), #3774 (completed-vs-timeout mismatch), #4589 / #4609 (complementary: avoid redundant main-session explore/librarian alongside `plan`).

---
## Summary by cubic
Fixes sync delegation returning before its background children finish. The poller now waits for direct child background tasks and a fresh post‑wake assistant turn (or a short grace) so parents get plans that include `explore`/`librarian` results. Addresses #4615.

- **Bug Fixes**
  - Defer completion while direct child background tasks are active, then hold until a new assistant message id appears or a grace window elapses (default 5s via `childWakeGraceMs`); times out if children never finish.
  - Plumbs `BackgroundManager.hasActiveChildTasks(sessionID)` into `pollSyncSession` via `executeSyncTask`; adds targeted tests for gating and timeout cases.
  - Stability and UX: suppress shutdown EPIPE/ECONNRESET log spam; fix Windows LSP/`@ast-grep` shim resolution; honor provider‑level concurrency for background retries; allow safe `file://` prompts under approved home subdirs; stop transcript cache growth; respect host `opencode.jsonc` `skills.paths`; mark post‑compact “continue” as synthetic; keep `TodoWrite` enabled; dispatch run agents by id; Codex/LazyCodex defaults to autonomous installs with startup auto‑update, routes `lazycodex` update/uninstall via the Node installer, adds an LSP `PostCompact` hook to reset diagnostics, and refines rules to prefer `CONTEXT.md` (exclude `AGENTS.md` by default).

- **Dependencies**
  - Bump `oh-my-opencode` and all platform packages to `4.7.5`; update `bun.lock`; drop direct `effect` dependency.

<sup>Written for commit 2320c3f5be29845df7003ec0ee58813f38965aa5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (dismissed):
Thanks @ririnto.

- The diff is 487 LOC across 12 files (`src/features/background-agent/manager.ts`, `src/shared/child-task-deferral.ts`, `src/tools/call-omo-agent/completion-poller.ts`, `sync-executor.ts`, `tools.ts`, `src/tools/delegate-task/constants.ts`, `sync-session-poller.ts`, `sync-task.ts`, plus tests). That is well past the 50-line `fix()` target in `AGENTS.md`. Could you split into the minimal change for #4615 and separate follow-up PRs for the broader poller/sync work?
- One of the added lines contains an em dash (U+2014): the prompt text `...running — end your turn...`. `AGENTS.md` lists em/en dashes as forbidden in generated content. Could you swap for a hyphen or restructure the sentence?

Leaving open as request-changes, no rush.

ririnto:
My Sisyphus will do. haha

ririnto:
need rebase :(

ririnto:
maybe need update to bun.lock

code-yeongyu:
[sisyphus-dev] I checked this during the bug sweep for #3774/#4615. I am not treating this as merge-perfect: the PR currently has `CHANGES_REQUESTED`, cubic identified a timeout-regression risk in the child-task deferral path, and the diff changes broader sync delegation/barrier semantics rather than a tiny localized result-propagation fix. Per the sweep rule, I am leaving this open for maintainer direction instead of merging or replacing it blindly.

code-yeongyu (approved):
Verified after maintainer force-push. Conflict was limited to bun.lock platform package versions and resolved to current dev. The Cubic child-task inactivity-timeout concern is covered: active children no longer reset the poll inactivity timer, so a never-finishing child times out. Local checks passed: bun test src/tools/delegate-task/sync-session-poller.test.ts, bun test src/features/background-agent/manager.test.ts, and bun run typecheck.

PR #4626, opened 2026-05-30
fix: preserve explicit gpt-5.4 config pins

biggameking:
## Summary
Narrow fix for the config migration regression where an explicit `openai/gpt-5.4` model pin could be rewritten to `openai/gpt-5.5` during config load.

This keeps the PR scoped to the immediate bug: user-edited config pins for `gpt-5.4` are preserved instead of being treated as plugin-owned defaults.

## Changes
- remove the `openai/gpt-5.4 -> openai/gpt-5.5` model-version migration entry
- keep the existing migration mechanism otherwise unchanged
- add focused migration and config-load regression tests for explicit `openai/gpt-5.4` pins

## Testing
- `bun test src/plugin-config.test.ts --test-name-pattern "does not rewrite explicit user-selected openai/gpt-5.4 models during config load"`
- `bun test src/shared/migration.test.ts --test-name-pattern "openai/gpt-5.4"`
- `bun run typecheck`
- `bun run build`

## Follow-up
The broader policy question around model migration guardrails can be handled separately. This PR now only addresses the concrete accidental rewrite regression.


MoerAI (changes requested):
Thanks @biggameking.

This PR is labeled `fix()` but the diff is 601 LOC across 4 files (`src/plugin-config.test.ts`, `src/shared/migration.test.ts`, `migration.ts`, `migration/model-versions.ts`). At that size it is hard to review as a bug fix and looks closer to a config-migration refactor, which is out of scope for the 50-line `fix()` target documented in `AGENTS.md`.

Could you split this into:
1. The minimal change for the actual canonical-model-pin bug
2. A separate refactor/feature PR for the broader migration changes

Leaving open as request-changes.

biggameking:
Thanks for the review. I agree the previous version was too broad for a `fix()` PR.

I narrowed this down to the concrete regression only:
- removed the broader canonical-model migration guard from this PR
- kept only the `openai/gpt-5.4 -> openai/gpt-5.5` migration removal
- kept focused regression coverage for preserving explicit `gpt-5.4` config pins during migration/config load

Current diff is now 3 files, about 56 insertions / 6 deletions, with the migration machinery otherwise unchanged. I also updated the PR title/body to reflect the narrower scope.

Verified locally:
- `bun test src/plugin-config.test.ts --test-name-pattern "does not rewrite explicit user-selected openai/gpt-5.4 models during config load"`
- `bun test src/shared/migration.test.ts --test-name-pattern "openai/gpt-5.4"`
- `bun run typecheck`
- `bun run build`

The broader guardrail idea can be handled separately if maintainers want it.

biggameking:
Small follow-up: I tightened this one further after my previous comment by removing the extra migration-layer regression case and keeping only the map assertion plus the config-load regression.

The net diff is now about 39 insertions / 6 deletions across the same 3 files.

code-yeongyu (changes requested):
Requesting changes for deprecated GPT model references in this PR.\n\nThe current dev branch has moved off gpt-5.2/gpt-5.3 and now uses the GPT-5.5-era IDs where applicable. This PR diff still adds stale references such as:\n\n371:+      const appliedMigrations = new Set(["model-version:openai/gpt-5.4-codex->openai/gpt-5.3-codex"]) 417:+        _migrations: ["model-version:openai/gpt-5.4-codex->openai/gpt-5.3-codex"], 439:+        "model-version:openai/gpt-5.4-codex->openai/gpt-5.3-codex", 474:+      const appliedMigrations = new Set(["model-version:openai/gpt-5.4-codex->openai/gpt-5.3-codex"]) 570:+            "model-version:openai/gpt-5.3-codex->openai/gpt-5.4", \n\nPlease update the PR so it does not introduce gpt-5.2/gpt-5.3 in runtime config, docs, fallback chains, snapshots, or ordinary test fixtures. If a deprecated ID is truly needed as a migration-input fixture, isolate it as an explicit historical/deprecated-input test and make sure the migrated/current output is GPT-5.5-era, not 5.2/5.3.

code-yeongyu (approved):
Latest head 26ec2cebb keeps the config-pin fix narrow, current-head stale GPT scan is clean in the touched migration files, targeted tests/typecheck passed locally, and Cubic approved the current head. Approving to clear my earlier deprecated-model request.

PR #4627, opened 2026-05-30
fix: restore primary model for reopened fallback sessions

biggameking:
## Summary
Narrow fix for reopened runtime-fallback sessions that start on a configured fallback model and then keep using that fallback instead of retrying the configured primary model.

The PR now keeps the change local to session initialization and the existing `chat.message` model override path.

## Changes
- when `session.created` reports a model that matches the configured fallback chain, keep the configured category/agent model as `originalModel`
- on the next `chat.message`, restore the primary model once it is not in cooldown
- add one focused integration regression test in the existing runtime fallback test file

## Testing
- `bun test src/hooks/runtime-fallback/index.test.ts --test-name-pattern "model switching via chat.message|session lifecycle"`
- `bun run typecheck`
- `bun run build`

## Follow-up
The broader runtime fallback state-machine cleanup has been left out of this PR so this fix remains reviewable and easy to revert independently.


MoerAI (dismissed):
Thanks @biggameking.

This PR is labeled `fix()` but the diff is 242 LOC across 7 files (`src/hooks/runtime-fallback/chat-message-handler.ts`, `event-handler.ts`, `fallback-bootstrap-model.ts`, `fallback-state.ts`, `types.ts`, plus tests). Convention here is that `fix()` PRs stay narrow (under 50 LOC, 1-2 files) so they can be reviewed and reverted independently, per `AGENTS.md`.

Could you narrow it down to the minimal change needed to restore the preferred primary model after fallback? The wider type and state-machine changes look like they would land better as separate PRs.

Leaving open as request-changes, happy to take another look once the diff is narrower.

biggameking:
Thanks for the review. I agree the earlier version was carrying too much state-machine cleanup for a `fix()` PR.

I narrowed this to the smallest version of the reopened-session bug fix:
- removed the new state flag/type changes and helper exports
- removed the separate test file
- kept the change local to `session.created`, the existing `chat.message` override path, and one integration regression test in `index.test.ts`

Current diff is now 3 files, about 82 insertions / 4 deletions. The PR title/body have been updated to match the narrower scope.

Verified locally:
- `bun test src/hooks/runtime-fallback/index.test.ts --test-name-pattern "model switching via chat.message|session lifecycle"`
- `bun run typecheck`
- `bun run build`

I left the broader runtime fallback cleanup out of this PR so this remains reviewable and independently revertible.

code-yeongyu (approved):
Verified the branch after maintainer force-push. The diff is now narrowed to runtime-fallback restore behavior, conflicts are resolved against current dev, forbidden gpt-5.2/gpt-5.3 fixtures are absent, and local checks passed: bun test src/hooks/runtime-fallback/index.test.ts and bun run typecheck.

PR #4640, opened 2026-06-01
fix: invalidate opencode plugin cache on postinstall

popododo0720:
## Summary
- Refreshed/rebased implementation of #2402 for the current `dev` branch.
- Keep this PR intentionally small per review feedback: `postinstall.mjs` now only removes OpenCode plugin specifier cache directories for `oh-my-opencode@...` / `oh-my-openagent@...`.
- Cover both cache layouts seen in practice: `~/.cache/opencode/<plugin>@...` and `~/.cache/opencode/packages/<plugin>@...`.

## Context
This PR is a current-code rework of #2402 (`fix: invalidate opencode plugin cache on global install`). The old PR is stale/dirty, but the underlying issue remains the same: global npm/bun installs can update the global package while OpenCode keeps loading a stale cached plugin copy.

Per the review request, the larger `postinstall.test.mjs` coverage and the unrelated Windows Codex installer timeout change have been removed from this PR and will be handled separately.

## Testing
- `node --check postinstall.mjs`
- Manual smoke test with a temporary `XDG_CACHE_HOME` confirming only matching oh-my plugin specifier cache directories are removed
- `git diff --check`

### DIFF
--- a/postinstall.mjs
+++ b/postinstall.mjs
@@ -1,8 +1,10 @@
 // postinstall.mjs
 // Runs after npm install to verify platform binary is available
 
-import { readFileSync } from "node:fs";
+import { readFileSync, readdirSync, rmSync } from "node:fs";
 import { createRequire } from "node:module";
+import { homedir } from "node:os";
+import { join } from "node:path";
 import {
   getPlatformPackageCandidates,
   getBinaryPath,
@@ -13,6 +15,7 @@ import { detectPlatformBinaryMismatch } from "./bin/version-mismatch.js";
 const require = createRequire(import.meta.url);
 
 const MIN_OPENCODE_VERSION = "1.4.0";
+const OPENCODE_PLUGIN_PACKAGES = ["oh-my-opencode", "oh-my-openagent"];
 
 /**
  * Parse version string into numeric parts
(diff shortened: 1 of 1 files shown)


MoerAI (changes requested):
Thanks for the PR, @popododo0720, and for taking the time on this.

The fix concept (invalidating the OpenCode plugin cache on postinstall) is reasonable and the new tests are valuable, but the diff is 346 LOC across 3 files which is well past the 50-line target for `fix()` PRs documented in `AGENTS.md`.

Could you split this into:
1. The minimal cache-invalidation change in `postinstall.mjs`
2. A separate follow-up PR for the test coverage in `postinstall.test.mjs` and `src/cli/install-codex/install-codex.test.ts`

Smaller, focused PRs land much faster and are easier to revert if anything goes sideways. Happy to take another look once the diff is narrower.

popododo0720:
Updated per review: #4640 is now reduced to the minimal postinstall cache invalidation change only (1 file, 25-line diff). The postinstall test coverage was removed from this PR. The Codex installer timeout change is also no longer included here; latest dev already has the installer integration timeout budget in `1ce0e93f3 test(codex): budget installer integration tests`, so opening a separate PR for that would duplicate upstream. Happy to follow up with postinstall-specific test coverage after this fix lands.

code-yeongyu (approved):
Approved after re-review. The prior request was to narrow the fix-sized PR; current head 8d8efb0cb9630b40b28c4fa2bdfc435d8fe3d11c is now reduced to the minimal postinstall cache invalidation change, CI/build/Codex compatibility are green, and Cubic reports 0 issues.

PR #4641, opened 2026-06-01
fix(anthropic-context-window-limit-recovery): honor experimental.aggressive_truncation flag in executor (fixes #3899)

### LINKED ISSUE #3899: bug/docs: aggressive truncation appears enabled by default despite docs
## Summary

Discord feedback reports that aggressive truncation appears to be enabled by default, contradicting the current docs/config expectations.

Source: Discord `#feedback` message `1502694424835264642`.

## Reported behavior

> agressive trunucation is also somehow on by default now.... which contradicts the docs

## Expected behavior

One of these should be true:

1. aggressive truncation is **not** enabled by default, matching docs, or
2. docs/config schema clearly state that aggressive truncation is now default-on and explain how to disable/tune it.

## Why this matters

Default truncation policy is user-facing behavior. If it silently changed, users can lose context unexpectedly and docs become misleading.

## Repro data needed

- OMO/OpenCode versions
- relevant config file contents, especially compaction/truncation/context-window settings
- exact docs page/section that says the opposite
- command/session where truncation happened
- logs showing aggressive truncation was selected by default

## Suggested investigation

- inspect default config generation for truncation/compaction settings
- inspect runtime defaults when config key is missing
- compare generated schema/defaults against docs
- add `doctor --verbose` visibility for effective truncation policy if missing

MoerAI:
## Summary

The context-window-limit-recovery executor ignored the `experimental.aggressive_truncation` flag (default false per docs) and ran the aggressive-truncation strategy unconditionally on every token-limit error. This PR honors the documented default-off contract: aggressive truncation now only runs when the user opts in.

## Root Cause

[`src/hooks/anthropic-context-window-limit-recovery/executor.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/hooks/anthropic-context-window-limit-recovery/executor.ts) had two interlocking bugs:

1. Line 24 (pre-fix): `void _experimental` explicitly discarded the parameter, so the executor never saw the user's opt-in flag.
2. Lines 60-77 (pre-fix): the aggressive-truncation block ran whenever `isOverLimit && truncateAttempt < max`, with the inline comment literally reading `// Aggressive Truncation - always try when over limit`.

Maintainer triage on the issue (https://github.com/code-yeongyu/oh-my-openagent/issues/3899#issuecomment-4466435420) called this out directly: "the executor currently ignores the experimental config argument and always attempts the aggressive truncation recovery strategy when the token limit is exceeded."

Wiring was already correct on the caller side. [`recovery-hook.ts`](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/src/hooks/anthropic-context-window-limit-recovery/recovery-hook.ts#L103-L114) passes `experimental` into both `executeCompact` call sites (the `session.error` setTimeout path and the `session.idle` path). The executor was the only place the value got dropped.

## Changes

| File | Change |
|------|--------|
| `src/hooks/anthropic-context-window-limit-recovery/executor.ts` | Rename `_experimental` -> `experimental`, drop the `void _experimental` discard, gate the aggressive-truncation block on `experimental?.aggressive_truncation === true`. |
| `src/hooks/anthropic-context-window-limit-recovery/executor.test.ts` | Two new regression tests under `executeCompact lock management` that pin the contract (skip aggressive when flag is undefined or false). Two existing tests that exercised the aggressive path now pass `{ aggressive_truncation: true }` explicitly so they continue to cover the opt-in branch. |

Diff: +103 / -10 across 2 files. Production code change is 9 lines.

## Reproduction (before fix)

Two new regression tests in `executor.test.ts` fail on `dev`:

~~~
(fail) executeCompact lock management > does NOT run aggressive truncation when experimental.aggressive_truncation is not enabled (#3899)
Expected number of calls: 0
Received number of calls: 1
  at executor.test.ts:418:29

(fail) executeCompact lock management > does NOT run aggressive truncation when experimental.aggressive_truncation is explicitly false (#3899)
Expected number of calls: 0
Received number of calls: 1
  at executor.test.ts:458:29

 11 pass
 2 fail
 27 expect() calls
~~~

The `runAggressiveTruncationStrategy` spy is invoked even though no `experimental` config is passed and the docs declare the flag default-off.

## Verification (after fix)

Same regression tests now pass:

~~~
bun test src/hooks/anthropic-context-window-limit-recovery/executor.test.ts
 14 pass
 0 fail
 29 expect() calls
Ran 14 tests across 1 file. [3.32s]
~~~

Both new contract tests pass:
- `does NOT run aggressive truncation when experimental.aggressive_truncation is not enabled (#3899)` -> green
- `does NOT run aggressive truncation when experimental.aggressive_truncation is explicitly false (#3899)` -> green

The two pre-existing aggressive-path tests (`falls through to summarize when truncation is insufficient`, `does NOT call summarize when truncation is sufficient`) still pass with the explicit `{ aggressive_truncation: true }` payload, so the opt-in branch retains full coverage.

## Test

- Regression tests: `src/hooks/anthropic-context-window-limit-recovery/executor.test.ts` (2 new cases marked `(#3899)`)
- Related suite: `bun test src/hooks/anthropic-context-window-limit-recovery/{executor,recovery-hook,recovery-hook-regression}.test.ts` -> **20 pass / 0 fail**
- Typecheck: `bun run typecheck` -> clean

`bun test src/hooks/anthropic-context-window-limit-recovery/` shows 2 pre-existing failures in `storage.test.ts` (`truncates all if target not reached` x2) that manifest only when the full module is run together; the file passes in isolation both before and after my changes. Filed as Known Flaky Tests per the maintainer's contributor guidelines; not caused by this PR.

## Design notes / scope guardrails

- **Behavior preservation.** Users who already opt in via `{ experimental: { aggressive_truncation: true } }` see no change. Users who never set the flag now match the documented default and fall straight through to `runSummarizeRetryStrategy`. The summarize path is the documented primary recovery, and the hook's own AGENTS.md lists aggressive truncation as recovery strategy 4 of 5, not strategy 1.
- **No new defaults.** `ExperimentalConfigSchema` already declares `aggressive_truncation: z.boolean().optional()` with no default; the docs at `docs/reference/configuration.md` describe it as `false`. The fix relies on the existing schema and existing wiring, not a new default value.
- **No new dependencies, no new files, no schema changes.**

Fixes #3899

---
## Summary by cubic
Fixed the context-window limit recovery executor to honor `experimental.aggressive_truncation`. Aggressive truncation now runs only when opted in; otherwise it falls back to the summarize strategy. Fixes #3899.

- **Bug Fixes**
  - Stop discarding the `experimental` param in the executor.
  - Gate aggressive truncation on `experimental.aggressive_truncation === true` instead of always running.
  - Add regression tests to ensure aggressive truncation is skipped when the flag is undefined or false; update existing tests to explicitly opt in.

<sup>Written for commit d8a14816cf2881068eca8e18c5ec4732fbfd9b8f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4641?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/hooks/anthropic-context-window-limit-recovery/executor.test.ts
+++ b/src/hooks/anthropic-context-window-limit-recovery/executor.test.ts
@@ -348,7 +348,7 @@ describe("executeCompact lock management", () => {
   })
 
   test("falls through to summarize when truncation is insufficient", async () => {
-    // given: Over token limit with truncation returning insufficient
+    // given: Over token limit with truncation returning insufficient, aggressive truncation opted in
     autoCompactState.errorDataBySession.set(sessionID, {
       errorType: "token_limit",
       currentTokens: 250000,
@@ -363,8 +363,20 @@ describe("executeCompact lock management", () => {
       nextTruncateAttempt: params.truncateAttempt + 1,
     }))
 
-    // when: Execute compaction
-    await executeCompact(sessionID, msg, autoCompactState, mockClient, directory, pluginConfig)
+    const experimental = {
+      aggressive_truncation: true,
+    }
+
+    // when: Execute compaction with aggressive truncation enabled
+    await executeCompact(
+      sessionID,
+      msg,
+      autoCompactState,
+      mockClient,
+      directory,
+      pluginConfig,
+      experimental,
+    )
 
     // then: Truncation was attempted
     expect(truncateSpy).toHaveBeenCalled()
@@ -383,8 +395,76 @@ describe("executeCompact lock management", () => {
     truncateSpy.mockRestore()
   })
 
+  test("does NOT run aggressive truncation when experimental.aggressive_truncation is not enabled (#3899)", async () => {
+    //#given - Over token limit but experimental.aggressive_truncation is undefined (default)
+    autoCompactState.errorDataBySession.set(sessionID, {
+      errorType: "token_limit",
+      currentTokens: 250000,
+      maxTokens: 200000,
+    })
+
+    const truncateSpy = spyOn(
+      recoveryStrategy,
+      "runAggressiveTruncationStrategy",
+    ).mockImplementation(async (params) => ({
+      handled: false,
+      nextTruncateAttempt: params.truncateAttempt + 1,
+    }))
+
+    //#when - Execute compaction without experimental config (default behavior)
+    await executeCompact(sessionID, msg, autoCompactState, mockClient, directory, pluginConfig)
+
+    //#then - Aggressive truncation must be skipped per docs (defaults false)
+    expect(truncateSpy).not.toHaveBeenCalled()
+
+    //#and - Summarize should still run as the primary recovery path
+    expect(mockClient.session.summarize).toHaveBeenCalled()
+
+    truncateSpy.mockRestore()
+  })
+
+  test("does NOT run aggressive truncation when experimental.aggressive_truncation is explicitly false (#3899)", async () => {
+    //#given - Over token limit with experimental flag explicitly false
+    autoCompactState.errorDataBySession.set(sessionID, {
+      errorType: "token_limit",
+      currentTokens: 250000,
+      maxTokens: 200000,
+    })
+
+    const truncateSpy = spyOn(
+      recoveryStrategy,
+      "runAggressiveTruncationStrategy",
+    ).mockImplementation(async (params) => ({
+      handled: false,
+      nextTruncateAttempt: params.truncateAttempt + 1,
+    }))
+
+    const experimental = {
+      aggressive_truncation: false,
+    }
+
+    //#when - Execute compaction with explicit false
+    await executeCompact(
+      sessionID,
+      msg,
+      autoCompactState,
+      mockClient,
+      directory,
+      pluginConfig,
+      experimental,
+    )
+
+    //#then - Aggressive truncation must be skipped
+    expect(truncateSpy).not.toHaveBeenCalled()
+
+    //#and - Summarize should run
+    expect(mockClient.session.summarize).toHaveBeenCalled()
+
+    truncateSpy.mockRestore()
+  })
+
   test("does NOT call summarize when truncation is sufficient", async () => {
-    // given: Over token limit with truncation returning sufficient
+    // given: Over token limit with truncation returning sufficient, aggressive truncation opted in
     autoCompactState.errorDataBySession.set(sessionID, {
       errorType: "token_limit",
       currentTokens: 250000,
@@ -411,8 +491,20 @@ describe("executeCompact lock management", () => {
       }
     })
 
-    // when: Execute compaction
-    await executeCompact(sessionID, msg, autoCompactState, mockClient, directory, pluginConfig)
+    const experimental = {
+      aggressive_truncation: true,
+    }
+
+    // when: Execute compaction with aggressive truncation enabled
+    await executeCompact(
+      sessionID,
+      msg,
+      autoCompactState,
+      mockClient,
+      directory,
+      pluginConfig,
+      experimental,
+    )
 
     // Wait for setTimeout callback
     await fakeTimeouts.advanceBy(600)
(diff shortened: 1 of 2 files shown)


MoerAI:

MoerAI:
CI note: `codex-compatibility (windows-latest)` reports 4 timeouts in `install-codex` integration tests:

- `install-codex > #given simulated Linux Codex install` (5015ms timeout)
- `install-codex > #given codex installer #when installing omo #then links omo-prefixed component CLIs to existing cached runtimes` (5016ms timeout)
- `install-codex > #given Codex prunes an old plugin cache version` (5016ms timeout)
- `install-codex > #given autonomous permissions requested` (5015ms timeout)

This PR only touches `src/hooks/anthropic-context-window-limit-recovery/executor.ts` and the co-located test, neither of which interacts with `packages/omo-codex/` or the `install-codex` flow. The dev branch CI shows a mix of pass/fail for this same job over the last 24 hours (commit 7afa4d08 passes, 423ff7e8 / d0d1735e / 247bba39 / 92bad87d fail), all without changes to `install-codex`, so this looks like pre-existing Windows-only flakiness with the 5s timeout cap on the Codex install integration suite.

The 14 executor tests, the broader `anthropic-context-window-limit-recovery` module tests, and the cross-platform `test` / `typecheck` / Linux+macOS `codex-compatibility` jobs all pass.

MoerAI:
Self-PR note: holding off on merge while the `codex-compatibility (windows-latest)` check failed with 4 `install-codex` timeouts at exactly 5000ms each, in an area (`packages/omo-codex/src/install-codex/`) untouched by this PR. The Linux and macOS runners passed, and other recent CLEAN PRs are green on the same check, so this looks like a Windows runner timing flake rather than something this PR caused. I will re-trigger CI and re-evaluate in the next maintenance pass. The actual diff (gating aggressive truncation on `experimental.aggressive_truncation`) is APPROVED and the two new regression tests in `executor.test.ts` pin the #3899 contract well.

code-yeongyu:
[sisyphus-dev] Reviewed this during the bug sweep for #3899. The PR is focused and CI/cubic are green, but the fix changes user-visible default recovery behavior for token-limit handling by making aggressive truncation strictly opt-in. Because this is a behavior-policy decision rather than a purely local mechanical bug fix, I am not merging it automatically under the sweep rule. Leaving it for maintainer decision.

MoerAI:
Thanks for reviewing. On the default-behavior concern: making aggressive truncation strictly opt-in is the documented contract, not a new policy. #3899 Expected Behavior option 1 is:

> aggressive truncation is **not** enabled by default, matching docs

The executor was applying aggressive truncation even when `experimental.aggressive_truncation` was unset/false, which is what the Discord report and the issue flagged as contradicting the docs. This PR makes the runtime honor the flag so the effective default matches what the docs/schema already advertise (default-off). So it's aligning code to the existing documented contract rather than introducing a new behavior policy.

If you'd rather resolve #3899 by flipping the docs to default-on instead, I'm happy to close this in favor of that direction. No rush on the merge decision.

MoerAI:
Thanks for the review. On the behavior-policy concern: making aggressive truncation strictly opt-in is option 1 of the two acceptable outcomes issue #3899 lists under Expected Behavior (`aggressive truncation is not enabled by default, matching docs`). The report is that the default silently became aggressive-on, contradicting the docs. This change realigns the runtime default with the documented default rather than introducing a new policy. If you'd instead prefer option 2 (keep default-on and update docs), I'm happy to close this in favor of a docs-only PR.

MoerAI:
Thanks @MoerAI, merging.

PR #4620, opened 2026-05-30
fix(ralph-loop): trust oracle task verification evidence

Atmospenguin:
## Summary
- parse Oracle verification verdicts like VERIFIED COMPLETE, accepted prose, and rejection wording without mistaking the original DONE promise for verification
- require trusted parent-session evidence to come from a completed task tool call with subagent_type=oracle and oracle task metadata with session_id
- prevent free-form parent VERIFIED text from completing ultrawork verification when no Oracle session is tracked

## Verification
- bun test src/hooks/ralph-loop/oracle-verification-detector.test.ts src/hooks/ralph-loop/ulw-loop-verification.test.ts src/hooks/ralph-loop/completion-promise-detector.test.ts src/hooks/ralph-loop/completion-promise-session-negative.test.ts: 68 pass, 159 expect() calls
- bun run build: passed
- lsp_diagnostics: no diagnostics on all changed files
- GIT_MASTER=1 git diff --check: clean before commit
- live persistent QA: /var/folders/vt/tk9zg7hj1fd06wnkh6rjjqvc0000gp/T/opencode/ulw-final-qa15/export-parent.json contains completed task tool output with subagent_type=oracle, VERIFIED COMPLETE, task_metadata session_id ses_18821c56fffeta1uu4QVYh5yhL, and subagent: oracle; no false verification-failure retry

## Known unrelated blockers
- bun run typecheck is blocked locally because tsgo is not installed: /bin/bash: tsgo: command not found
- broad src/hooks/ralph-loop/index.test.ts failure at line 1492 reproduced on clean origin/dev and is unrelated to this branch

---
## Summary by cubic
Trust only Oracle task verification evidence and prevent false positives so ULTRAWORK completes only when a verified Oracle verdict is present. This stops free‑form “VERIFIED” text from prematurely finishing the loop and completes earlier when a trusted parent task verdict exists.

- **Bug Fixes**
  - Parse Oracle verdicts: “VERIFIED”, “VERIFIED COMPLETE”, “verification passes”, and rejection phrasing; prefer verdict over the original `DONE` promise.
  - Accept evidence only from completed `task` tool output with `subagent_type=oracle` and `<task_metadata>` `session_id`; ignore free‑form assistant text.
  - Scan parent session messages for trusted Oracle verdicts and complete ULTRAWORK immediately if found; do not re‑enter verification.
  - Gate completion detection for `ULTRAWORK_VERIFICATION_PROMISE` so completion requires a real verification session/evidence.
[...]

PR #4683, opened 2026-06-02
fix(ralph-loop): recover oracle tool-result verification

### LINKED ISSUE #3212: ULW verification loop: infinite loop when Oracle VERIFIED not detected in parent session
## Bug

The ultrawork (ULW) loop keeps cycling infinitely after Oracle emits `<promise>VERIFIED</promise>`. The verification is actually complete, but the system fails to detect the VERIFIED promise in the parent session's messages.

## Root Cause

Three interrelated issues in the ralph-loop verification path:

### 1. `detectOracleVerificationFromParentSession` only scans `assistant` role messages

In `src/hooks/ralph-loop/pending-verification-handler.ts`, the function filters for `message.info?.role !== "assistant"`, but Oracle subagent output is returned as a `tool_result` which may be in a non-assistant message or a separate message part. This means VERIFIED is never found → recovery always fails → `handleFailedVerification` restarts the loop.

### 2. `detectCompletionInSessionMessages` only inspects `assistant` messages and `text` parts

In `src/hooks/ralph-loop/completion-promise-detector.ts`, the function filters for `assistant` role and only collects `text` parts. Tool result parts (which contain the Oracle's output) are excluded, so VERIFIED is never detected via the API path either.

### 3. No circuit breaker for failed verification attempts

When verification detection fails, `restartAfterFailedVerification` increments `iteration` but there's no cap on how many verification rounds are attempted. With `max_iterations = 100`, the loop can run up to 100 Oracle calls (expensive!) before giving up.

## Reproduction

1. Start a ULW loop (`/ulw-loop "some task"`)
2. Agent completes work, emits `<promise>DONE</promise>`
3. System transitions to verification pending, injects ULTRAWORK_VERIFICATION_PROMPT
4. Agent calls Oracle, Oracle returns with `<promise>VERIFIED</promise>`
5. System fails to detect VERIFIED → calls `handleFailedVerification` → restarts loop
6. Loop repeats indefinitely

### Fix 1: Scan all message roles for VERIFIED detection

In `detectOracleVerificationFromParentSession`:
- Remove the `role !== "assistant"` filter
- Extract text from `parts` (all types), `content` string, and `tool_result` parts
- Accept VERIFIED even without explicit "Agent: oracle" marker (fallback)

In `detectCompletionInSessionMessages`:
- When checking for VERIFIED promise, scan ALL messages (not just assistant)
- Include `tool_result` parts in text collection

### Fix 2: Add verification attempt circuit breaker

- Add `verification_attempts: number` to `RalphLoopState`
- New constant `MAX_VERIFICATION_ATTEMPTS = 3`
- Increment counter in `restartAfterFailedVerification`
- In `handlePendingVerification`: if `verification_attempts >= MAX_VERIFICATION_ATTEMPTS`, call `loopState.clear()` and end the loop with a warning toast
- Serialize/deserialize the counter in the state file

### Fix 4: Transcript-based fallback

In `handlePendingVerification`, before calling `handleFailedVerification`:
- If API-based detection failed, try `detectCompletionInTranscript(transcriptPath, "VERIFIED", started_at)` as a fallback
- If VERIFIED found in transcript file → clear loop (success)
- Pass `getTranscriptPath` through the function chain

## Files Affected

- `src/hooks/ralph-loop/pending-verification-handler.ts` (Fix 1 + 4)
- `src/hooks/ralph-loop/completion-promise-detector.ts` (Fix 1)
- `src/hooks/ralph-loop/verification-failure-handler.ts` (Fix 2)
- `src/hooks/ralph-loop/loop-state-controller.ts` (Fix 2)
- `src/hooks/ralph-loop/types.ts` (Fix 2 - new field)
- `src/hooks/ralph-loop/ralph-loop-event-handler.ts` (Fix 4 - pass getTranscriptPath)
- `src/hooks/ralph-loop/storage.ts` (Fix 2 - serialize/deserialize)

## Patch

I've applied these fixes to `dist/index.js` locally (patching the compiled bundle). The patches work and JS syntax validates clean. Happy to submit a PR against the TypeScript source if that would be preferred.

code-yeongyu:
## Summary
- scan parent-session verification recovery across message roles so non-assistant `tool_result` Oracle output is not missed
- complete the ULW verification loop when recovered parent evidence already contains Oracle `<promise>VERIFIED</promise>`
- add a regression for a non-assistant Oracle tool result with task metadata

Fixes #3212.
Supersedes #4620 with a narrower current-dev bug fix.

## Verification
- `bun test src/hooks/ralph-loop/oracle-verification-detector.test.ts src/hooks/ralph-loop/ulw-loop-verification.test.ts src/hooks/ralph-loop/completion-promise-detector.test.ts --bail`
- `bun run typecheck`
- `git diff --check`

---
## Summary by cubic
Restores Oracle verification recovery from parent sessions and rejects spoofed non-assistant text, so ULW loops complete only when real evidence exists; fixes #3212. When recovered evidence matches `ULTRAWORK_VERIFICATION_PROMISE`, the loop completes immediately and shows a success toast, with toast failures logged.

- **Bug Fixes**
  - Scan parent-session messages across all roles; accept `tool_result` from any role and parse `text` only from assistant messages to prevent spoofing.
  - On recovered evidence with `completion_promise` `ULTRAWORK_VERIFICATION_PROMISE`, clear loop state, finish the loop, show "ULTRAWORK LOOP COMPLETE!", and log toast errors without blocking completion.
  - Add tests for non-assistant Oracle `tool_result` recovery, rejection of spoofed non-assistant text, and toast behavior.

<sup>Written for commit adedbb41facf9dbc8037f35f66272acd9c793a23. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/hooks/ralph-loop/pending-verification-handler.ts
+++ b/src/hooks/ralph-loop/pending-verification-handler.ts
@@ -90,16 +90,29 @@ type LoopStateController = {
 }
 
 function showCompletionToastBestEffort(ctx: PluginInput, state: RalphLoopState): void {
+	const showToast = ctx.client.tui?.showToast
+	if (!showToast) {
+		return
+	}
+
+	const toastBody = {
+		body: {
+			title: "ULTRAWORK LOOP COMPLETE!",
+			message: `JUST ULW ULW! Task completed after ${state.iteration} iteration(s)`,
+			variant: "success" as const,
+			duration: 5000,
+		},
+	}
+	const logToastError = (error: unknown) => {
+		log(`[${HOOK_NAME}] Failed to show ulw completion toast`, {
+			error: String(error),
+		})
+	}
+
 	try {
-		void Promise.resolve(ctx.client.tui?.showToast?.({
-			body: {
-				title: "ULTRAWORK LOOP COMPLETE!",
-				message: `JUST ULW ULW! Task completed after ${state.iteration} iteration(s)`,
-				variant: "success",
-				duration: 5000,
-			},
-		})).catch(() => {})
-	} catch {
+		void Promise.resolve(showToast(toastBody)).catch(logToastError)
+	} catch (error) {
+		logToastError(error)
 	}
 }
 
--- a/src/hooks/ralph-loop/pending-verification-handler.ts
+++ b/src/hooks/ralph-loop/pending-verification-handler.ts
@@ -19,9 +19,10 @@ function collectAssistantText(message: OpenCodeSessionMessage): string {
 		return ""
 	}
 
+	const allowTextParts = message.info?.role === "assistant"
 	let text = ""
 	for (const part of message.parts) {
-		if (part.type !== "text" && part.type !== "tool_result") {
+		if (part.type !== "tool_result" && !(allowTextParts && part.type === "text")) {
 			continue
 		}
 		text += `${text ? "\n" : ""}${part.text ?? ""}`
--- a/src/hooks/ralph-loop/pending-verification-handler.ts
+++ b/src/hooks/ralph-loop/pending-verification-handler.ts
@@ -1,6 +1,6 @@
 import type { PluginInput } from "@opencode-ai/plugin"
 import { log } from "../../shared/logger"
-import { HOOK_NAME } from "./constants"
+import { HOOK_NAME, ULTRAWORK_VERIFICATION_PROMISE } from "./constants"
 import { extractOracleSessionID, isOracleVerified } from "./oracle-verification-detector"
 import type { RalphLoopState } from "./types"
 import { handleFailedVerification } from "./verification-failure-handler"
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-dev] GPT-5.5 high review found one real blocker in the replacement patch: the first version trusted non-assistant text parts when recovering Oracle VERIFIED evidence from the parent session. I pushed a follow-up commit (`44ce4883b`) that only trusts assistant text parts and non-assistant `tool_result` parts, plus an adversarial regression proving user-role Oracle-looking text does not complete ULW verification. Local verification after the fix: focused ralph-loop suite 46/46 pass, `bun run typecheck` pass, `git diff --check` clean.

code-yeongyu:
[sisyphus-dev] Follow-up to the follow-up: GPT-5.5 high review also caught that the best-effort toast helper introduced an empty/silent catch, which violates the repo invariant. I pushed `adedbb41f` to log toast failures through the existing logger instead. Local verification after this commit: focused ralph-loop suite 46/46 pass, `bun run typecheck` pass, `git diff --check` clean, and a local empty-catch/catch(() => {}) search over touched files returns no matches.

PR #4715, opened 2026-06-03
fix(model-core): allow cross-provider fallback on quota/billing errors

### LINKED ISSUE #2393: [Bug]: Model fallback never triggers on usage limits — model_fallback config undiscoverable + error pattern gaps in model-error-classifier.ts

### LINKED ISSUE #3519: [Bug]: fallback does not advance when the active model hits credit/quota exhaustion

inhyoe:
## Problem

`STOP_ERROR_NAMES` and `STOP_MESSAGE_PATTERNS` in `model-error-classifier.ts` blocked all fallback retries for quota/billing errors (e.g., `quotaexceedederror`, `insufficientcreditserror`, `freeusagelimiterror`).

This is overly restrictive when users configure cross-provider fallback chains. For example:
- Primary: `anthropic/claude-opus-4-8` (Anthropic account)
- Fallback: `opencode-go/kimi-k2.6` (OpenCode Go account)

Quota exhaustion on Anthropic does **not** imply quota exhaustion on OpenCode Go. These are independent billing systems. The previous logic prevented valid, meaningful fallbacks in this common multi-provider setup.

## Changes

- Emptied `STOP_ERROR_NAMES` (was: quota/credit/free-usage errors)
- Emptied `STOP_MESSAGE_PATTERNS` (was: 27 quota/billing strings)
- Added explicit `Set<string>` / `string[]` types to empty collections for TS strict mode compatibility
- Replaced `.some()` with `for...of` loop to avoid TypeScript `never` inference on empty arrays

## Rationale

Provider-scoped errors should not unconditionally block fallback to a different provider. The fallback chain already handles provider selection (`selectFallbackProvider`), so let the chain do its job instead of hard-stopping at the classifier level.

## Verification

- Built successfully in isolated Docker environment (`oven/bun:latest`)
- `bun run build` passes (typecheck + bundle + schema generation)

## Related Issues

Fixes #3519
Fixes #2393

---
## Summary by cubic
Enable cross-provider fallback on quota/billing exhaustion without retrying on the same provider. Quota errors stay non-retryable for the current provider but now trigger a provider switch when possible.

- **Bug Fixes**
  - Added `isProviderScopedError` in `model-core`; quota/billing remain STOP for current provider but are flagged as provider-scoped.
  - Background agent: skips same-provider candidates for provider-scoped errors and only retries when the current provider is known.
  - Fallback hook/state: added `requiresProviderSwitch` and `hasEligibleFallback`; refuse arming fallback if only same-provider options remain.
  - Event handler: uses provider-scoped check with a reliable provider signal and passes `requiresProviderSwitch` into `setPendingModelFallback`.
  - Updated shared exports and tests across components; removed the verification-only Dockerfile.

<sup>Written for commit 98fb80cca90e2a6c887df03fce4e51ef39abeb06. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4715?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (changes requested):
Thanks for digging into cross-provider fallback. I can’t approve this version yet.

Blocking issues:

1. The CLA check is failing, so the PR is not mergeable until that is resolved.
2. The implementation removes every quota/billing STOP name and message globally. That makes same-provider permanent billing/quota failures retryable too, which can reintroduce fallback loops or misleading retries. The classifier needs to preserve the distinction: quota exhaustion may be retryable only when the next fallback actually switches to an independent provider/billing scope, not as a blanket classifier rule.
3. Please remove the root `Dockerfile`. A verification-only Dockerfile with `oven/bun:latest`, `bun install`, and grep assertions is not part of the product surface and should not be committed.

A safer shape would keep quota/billing classification intact, then make the fallback decision provider-aware at the call site where current provider and next fallback provider are both known, with tests covering same-provider quota stop vs cross-provider quota retry.

inhyoe:
Thanks for the detailed review. I've reworked the PR to address every point. Pushed as `98fb80c`.

## @code-yeongyu (blocking)

**2. Don't disable quota/billing classification globally; keep the distinction and make the fallback decision provider-aware.**
Done. I reverted the classifier change entirely — `STOP_ERROR_NAMES` and `STOP_MESSAGE_PATTERNS` are restored, so `shouldRetryError`/`isRetryableModelError` stay `false` for quota/billing and the original classifier test suite passes unchanged.

The cross-provider behavior now lives at the call sites, exactly as you suggested:
- Added `isProviderScopedError()` to `model-core` to surface quota/billing exhaustion without making it blanket-retryable.
- **background-agent** (`fallback-retry-handler.ts`): a provider-scoped error retries **only onto a different provider**, and only when the current provider is known. If every remaining fallback resolves to the same provider, it stops (no loop).
- **in-session model-fallback** (`event.ts` + `next-fallback.ts` + state controller): the three retry gates also consider provider-scoped errors, but only arm when there is a **reliable** current-provider signal (never the inferred `"opencode"` default), and a **non-mutating preflight** confirms an eligible cross-provider fallback exists *before* aborting/continuing the session — so we never abort into a same-provider re-prompt.
- `requiresProviderSwitch` is derived from `isProviderScopedError` alone, so a wrapped error like `{name:"RateLimitError", message:"quota exceeded"}` is still blocked from same-provider retry.

Tests cover same-provider quota **stop** vs cross-provider quota **retry** for both paths (classifier units, background-agent handler, and the in-session controller/preflight).

**3. Remove the root Dockerfile.** Done — deleted.

**1. CLA.** I'll handle the CLA separately on my account.

## cubic
- *Dockerfile `COPY . /app`* — resolved by removing the Dockerfile.
- *Move quota names to `RETRYABLE_ERROR_NAMES`* — intentionally not done: that would make same-provider quota failures retryable and reintroduce the loop @code-yeongyu flagged. Name-only quota errors (e.g. `QuotaExceededError`) are instead caught by `isProviderScopedError` and get a cross-provider fallback at the call site.

## Verification
- `bun run build` ✅
- `tsgo --noEmit` (root) ✅ and `typecheck:packages` ✅
- `bun test` on the affected files ✅ 115 pass / 0 fail

inhyoe:

code-yeongyu:
[sisyphus-bot] Thank you @inhyoe — cross-provider-only fallback on quota/billing errors is the right semantics, and the head commit did address the substance of the earlier review. Closing because the branch can no longer land: most touched paths were relocated or split on dev (`src/plugin/event.ts` no longer exists — it became the event-model-* modules; delegate-task sync-task was removed), so this needs a re-port, not a merge. If you're up for a rebase onto current dev (the `isProviderScopedError` model-core piece should carry over almost directly), we'd be glad to review it promptly.

PR #4723, opened 2026-06-04
fix(claude-code-plugin-loader): replace all CLAUDE_PLUGIN_ROOT occurrences (fixes #4709)

### LINKED ISSUE #4709: resolvePluginPath only replaces the first ${CLAUDE_PLUGIN_ROOT} occurrence — hook commands referencing it twice break
## Problem

`resolvePluginPath()` in `src/features/claude-code-plugin-loader/plugin-path-resolver.ts` uses `String.prototype.replace` with a **string** pattern, which replaces only the **first** occurrence:

```ts
const CLAUDE_PLUGIN_ROOT_VAR = "${CLAUDE_PLUGIN_ROOT}"

export function resolvePluginPath(path: string, pluginRoot: string): string {
  return path.replace(CLAUDE_PLUGIN_ROOT_VAR, pluginRoot)  // first match only
}
```

Many real-world Claude Code plugins reference `${CLAUDE_PLUGIN_ROOT}` **more than once** in a single hook command. Example from the `token-optimizer` plugin (`alexgreensh/token-optimizer`, hooks/hooks.json):

```json
{
  "type": "command",
  "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/bash_hook.py --quiet"
}
```

After `resolvePluginPaths()` only the first occurrence is rewritten:

```
bash "/Users/me/.claude/plugins/cache/.../5.8.9/hooks/python-launcher.sh" "${CLAUDE_PLUGIN_ROOT}/hooks/run.py" ...
```

Since the spawned subprocess does not have `CLAUDE_PLUGIN_ROOT` in its env (see #4458), the shell expands the leftover variable to an empty string, producing `/hooks/run.py`, and every Bash tool call in opencode fails with:

```
python3: can't open file '/hooks/run.py': [Errno 2] No such file or directory
```

This fires on **every PreToolUse/PostToolUse event**, so the error is constant once such a plugin is installed on the Claude Code side.

## Reproduction

1. Install any Claude Code plugin whose hooks.json uses `${CLAUDE_PLUGIN_ROOT}` twice in one command (e.g. `token-optimizer@alexgreensh-token-optimizer`).
2. Run opencode with oh-my-openagent in a session that triggers a Bash tool call.
3. Observe `can't open file '/hooks/run.py'` on stderr of every hook dispatch.

## Fix

```ts
export function resolvePluginPath(path: string, pluginRoot: string): string {
  return path.replaceAll(CLAUDE_PLUGIN_ROOT_VAR, pluginRoot)
}
```

(or `split(VAR).join(pluginRoot)` for older targets). A test case with the variable appearing twice in one string would prevent regression — the existing `plugin-path-resolver.test.ts` apparently only covers single occurrences.

## Notes

- Related but distinct from #4458: that issue covers the env var not being injected for settings-file-sourced hooks, and states the plugin-directory path "works because the command string is rewritten before execution" — that holds only when the variable appears once per string. Fixing #4458 (injecting `CLAUDE_PLUGIN_ROOT` into the subprocess env) would *also* mask this bug at the shell level, but the rewrite itself is still incorrect for non-shell consumers of `resolvePluginPaths()` (e.g. MCP server args/paths).
- Verified present on `dev` HEAD as of 2026-06-03, and in the published `oh-my-openagent@4.6.0` bundle.

MoerAI:
## Summary
`resolvePluginPath()` only resolved the **first** `${CLAUDE_PLUGIN_ROOT}` in a string, so plugin hook commands that reference the placeholder more than once kept the later occurrences unresolved and broke every Bash tool call. Switched the single replace to `replaceAll`.

## Root Cause
`src/features/claude-code-plugin-loader/plugin-path-resolver.ts:4` used `String.prototype.replace` with a **string** pattern:

```ts
return path.replace(CLAUDE_PLUGIN_ROOT_VAR, pluginRoot)  // first match only
```

A string-pattern `.replace()` rewrites only the first match. Real plugins (e.g. `alexgreensh/token-optimizer`) put two `${CLAUDE_PLUGIN_ROOT}` references in one hook command (`bash "${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh" "${CLAUDE_PLUGIN_ROOT}/hooks/run.py" ...`). The second occurrence survived resolution, the spawned shell expanded the leftover `${CLAUDE_PLUGIN_ROOT}` to an empty string, and every PreToolUse/PostToolUse Bash call failed with `python3: can't open file '/hooks/run.py'`.

## Changes
| File | Change |
|------|--------|
| `src/features/claude-code-plugin-loader/plugin-path-resolver.ts` | `.replace()` → `.replaceAll()` so every `${CLAUDE_PLUGIN_ROOT}` is resolved |
| `src/features/claude-code-plugin-loader/plugin-path-resolver.test.ts` | Added regression test for a command containing two placeholders |

## Reproduction (before fix)
~~~
INPUT : bash "${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh" "${CLAUDE_PLUGIN_ROOT}/hooks/run.py" --quiet
OUTPUT: bash "/Users/me/.claude/plugins/cache/x/5.8.9/hooks/python-launcher.sh" "${CLAUDE_PLUGIN_ROOT}/hooks/run.py" --quiet
LEFTOVER PLACEHOLDER PRESENT: true   <-- second occurrence not replaced
~~~

## Verification (after fix)
~~~
OUTPUT: bash "/Users/me/.claude/plugins/cache/x/5.8.9/hooks/python-launcher.sh" "/Users/me/.claude/plugins/cache/x/5.8.9/hooks/run.py" --quiet
LEFTOVER PLACEHOLDER PRESENT: false  <-- all occurrences replaced

bun test src/features/claude-code-plugin-loader/plugin-path-resolver.test.ts
 4 pass  0 fail
~~~

## Test
- Regression test: `plugin-path-resolver.test.ts` (newly added multi-occurrence case)
- Module suite: `bun test src/features/claude-code-plugin-loader/` — 54 pass. One unrelated pre-existing failure in `discovery.test.ts` (`#when multiple non-'unknown' semver siblings are present`) reproduces on a clean `upstream/dev` tree without this change (Windows temp-path parsing); not touched here.
- Typecheck: `bun run typecheck` — exit code 0, clean.

Fixes #4709

---
## Summary by cubic
Fixed path resolution in `claude-code-plugin-loader` to replace all `${CLAUDE_PLUGIN_ROOT}` occurrences so multi-reference hook commands resolve correctly and Bash tool calls no longer fail.

- **Bug Fixes**
  - Use `.replaceAll()` in `resolvePluginPath` to replace every placeholder.
  - Add a regression test for a command containing two `${CLAUDE_PLUGIN_ROOT}` references.

<sup>Written for commit 4c0ea569baaf826a38ce428d629e8331a09ce372. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4723?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/claude-code-plugin-loader/plugin-path-resolver.test.ts
+++ b/src/features/claude-code-plugin-loader/plugin-path-resolver.test.ts
@@ -13,6 +13,17 @@ describe("resolvePluginPath", () => {
     // then
     expect(result).toBe("/tmp/plugin-root/dist/index.js")
   })
+
+  test("#given a path referencing the placeholder multiple times #when resolving the path #then it replaces every occurrence", () => {
+    // given
+    const path = 'bash "${CLAUDE_PLUGIN_ROOT}/hooks/launcher.sh" "${CLAUDE_PLUGIN_ROOT}/hooks/run.py" --quiet'
+
+    // when
+    const result = resolvePluginPath(path, "/tmp/plugin-root")
+
+    // then
+    expect(result).toBe('bash "/tmp/plugin-root/hooks/launcher.sh" "/tmp/plugin-root/hooks/run.py" --quiet')
+  })
 })
 
 describe("resolvePluginPaths", () => {
--- a/src/features/claude-code-plugin-loader/plugin-path-resolver.ts
+++ b/src/features/claude-code-plugin-loader/plugin-path-resolver.ts
@@ -1,7 +1,7 @@
 const CLAUDE_PLUGIN_ROOT_VAR = "${CLAUDE_PLUGIN_ROOT}"
 
 export function resolvePluginPath(path: string, pluginRoot: string): string {
-  return path.replace(CLAUDE_PLUGIN_ROOT_VAR, pluginRoot)
+  return path.replaceAll(CLAUDE_PLUGIN_ROOT_VAR, pluginRoot)
 }
 
 export function resolvePluginPaths<T>(obj: T, pluginRoot: string): T {


MoerAI:

MoerAI:
Thanks @MoerAI, merging. The replaceAll fix in plugin-path-resolver.ts is correct and the added regression test covers the multi-occurrence case.

Issue #4164, opened 2026-05-18
[Bug]: background completion reminders arrive only after active parent turn ends in v4.2.2


NICxKMS:
## Prerequisites

- [x] I am writing this issue in English.
- [x] I searched existing issues and found related but not exact matches.
- [x] I am using the latest version of oh-my-openagent.
- [x] I checked the documentation / investigated with the repo context loaded.
- [x] I ran `npx -y oh-my-openagent doctor` and included the output below.

## Bug Description

Background task completion reminders still appear to be delivered too late in `oh-my-openagent v4.2.2`.

A background child session can finish quickly, but the parent model does not receive `[BACKGROUND TASK COMPLETED]` or `[ALL BACKGROUND TASKS COMPLETE]` while its current turn is still active. The reminder is injected only after the parent turn ends and the parent has already sent its final response.

## Steps to Reproduce

1. Start a parent agent turn.
2. Launch a trivial background task:

```ts
task(
  category="quick",
  load_skills=[],
  run_in_background=true,
  description="Hi only session",
  prompt="Respond exactly with `hi` and then stop."
)
```

3. Keep the parent turn active with harmless file reads and sleeps for about a minute.
4. Observe that the child session completes.
5. Observe that no completion reminder reaches the active parent turn.
6. Let the parent send its final response.
7. Observe that `[ALL BACKGROUND TASKS COMPLETE]` arrives only after the parent turn has ended.

## Expected Behavior

The active parent turn should receive the background completion reminder before it finalizes, so it can call `background_output(...)` during the same turn.

## Actual Behavior

The child task completed in about 5 seconds, but no reminder reached the active parent turn.

Observed task:

- Background task ID: `bg_e8c2b5ef`
- Child session ID: `ses_1c46e2af9ffeBguUcCF8zy5QVF`
- Child output: `hi`
- Task duration: `5s`

The parent stayed active for roughly a minute. After the parent final response, this arrived:

```text
[ALL BACKGROUND TASKS COMPLETE]

Completed:
- bg_e8c2b5ef: Hi only session
```

After that, `background_output(task_id="bg_e8c2b5ef")` returned `hi`.

## Doctor Output

```shell
$ npx -y oh-my-openagent doctor

oMoMoMoMo Doctor

✓ System OK (opencode 1.15.4 · oh-my-openagent ^4.2.2)
```

## Error Logs

No exception or crash was observed. The failure is delayed notification timing.

## Configuration

`opencode --version`:

[...]

PR #4882, opened 2026-06-05
fix(background-agent): preserve final parent wake liveness

### LINKED ISSUE #4874: [Bug]: all-complete parent wake can be consumed as noReply before session resume
### Prerequisites

### Bug Description

On current `dev`, the final `[ALL BACKGROUND TASKS COMPLETE]` parent wake can become visible in the parent session as a `noReply` internal prompt, but the parent session may never later resume to call `background_output(...)`.

The confusing part is that `noReply` delivery itself is not the bug. Active-turn completion delivery from #4164 / #4478 must not be reverted. Successful completion reminders, including `[BACKGROUND TASK COMPLETED]` and `[ALL BACKGROUND TASKS COMPLETE]`, still need to be visible to an active parent turn as `noReply` so the parent can observe background completion without starting a competing assistant reply.

The bug is narrower: the final all-complete wake has reply-required liveness. If that wake is admitted as `noReply` for active-turn visibility, that informational admission must not consume, clear, or phantom-satisfy the later `shouldReply` resume requirement.

In short: `noReply` admission should make the reminder visible. It should not prove that the parent produced, or will produce, the reply needed to process all completed background work.

### Why This Issue Body Needs To Be Explicit

This issue is easy to misfix by reverting the active-turn `noReply` behavior from #4164 / #4478. That would bring back the older bug where successful background completion reminders are hidden from an active parent turn until too late.

The intended fix must preserve both constraints:

1. Active parent turns can still receive successful completion reminders as `noReply`.
2. The final all-complete `shouldReply` wake still causes exactly one later parent resume when the active turn did not process the reminder.

### Steps to Reproduce

1. Start a parent session on current `dev`.
2. Launch multiple background subagents.
3. Let all background subagents complete while the parent session is active or otherwise unable to safely start a reply-producing wake immediately.
4. Observe that the final `[ALL BACKGROUND TASKS COMPLETE]` `<system-reminder>` is injected into the parent session.
5. Observe that the parent session may not later resume to call `background_output(task_id=...)` for the completed work.

### Expected Behavior

The parent-wake contract should preserve both visibility and liveness:

1. Successful completion reminders remain visible during an active parent turn through `noReply` admission, preserving #4164 / #4478.
2. A final all-complete wake that requires a reply still retains its reply intent after any `noReply` admission.
3. If the active turn does not process the all-complete reminder, the system later resumes the parent exactly once when it is safe.
4. Dedupe should not treat an informational `noReply` admission as equivalent to a completed reply-producing dispatch.

### Actual Behavior

The final all-complete reminder can be admitted as `noReply`, making the reminder visible, but the later reply-required parent resume may not happen.

The observed failure is that the parent sees the all-complete `<system-reminder>`, then stays idle instead of resuming to process `background_output(...)`.

This looks like the reply-required liveness for the final all-complete wake may be consumed or phantom-satisfied by the earlier `noReply` admission. That mechanism is suspected, not yet proven.

### Doctor Output

```shell
oMoMoMoMo Doctor

 ⚠ 1 issue found:

1. TUI plugin entry missing from tui.json
   The server plugin is registered in opencode.json, but the TUI plugin entry ("oh-my-openagent/tui") is missing from tui.json. The Roles · Models sidebar section and TUI-only commands will not appear.
   Fix: Re-run the installer (`npx oh-my-openagent install`) to auto-write tui.json, or add "oh-my-openagent/tui" to the "plugin" array in /Users/rinto/.config/opencode/tui.json.
   Affects: TUI sidebar, TUI commands
```

### Error Logs

```shell
No thrown error.

The observable failure is behavioral: the final all-complete wake is visible as a noReply reminder, but the parent session does not later resume to process completed background work.
```

### Configuration

```json
{}
```

### Additional Context

Related constraints:

- #4164 reported that background completion reminders arrived only after the active parent turn ended, preventing the parent from calling `background_output(...)` during the same turn.
- #4478 fixed #4164 by allowing successful completion wakes to bypass the active-session gate as `noReply`. This behavior is required and must not be reverted.
- #4470 requires failed, cancelled, and interrupted wakes to remain deferred while the parent is active.
- #4120 requires parent wake dispatch to avoid racing user messages or starting competing assistant turns.
- #4742 introduced durable `forceNoReply` admission for unsafe branches. That change may be related, but this issue should not assume it is the final root cause before implementation verifies it.

Suspected area to investigate:

- The active or unsafe wake path can send a successful completion reminder with `body.noReply: true`.
- The original wake may still represent `shouldReply: true` liveness.
- The dedupe or dispatched-wake tracking may then treat the `noReply` admission as if the reply-required wake was already satisfied.
- If so, future attempts to perform the reply-producing all-complete wake can look redundant even though no parent reply was started.

The fix should separate these concepts:

- `noReply` admission: informational delivery for active-turn visibility.
- `shouldReply` liveness: the requirement that the final all-complete wake eventually resumes the parent if the active turn did not process it.

This is not a request for a pending-only revert. The fix should keep active-turn `noReply` completion delivery and preserve final all-complete reply liveness.

### Operating System

macOS

### OpenCode Version

1.16.1

ririnto:
## Summary

- Preserve final all-complete parent wake liveness after active-turn `noReply` admission.
- Keep #4164 active-turn completion visibility while avoiding duplicate `noReply` admissions.
- Keep failure/cancel/interrupted wakes deferred while the parent is unsafe.

## Changes

- Retain successful final `[ALL BACKGROUND TASKS COMPLETE]` wakes after informational `noReply` delivery.
- Track force-`noReply` admissions as non-reply-covering so later reply-required flushes are not deduped away.
- Schedule retry flushes for retained final wakes and suppress duplicate active-turn `noReply` prompts while still unsafe.
- Add regression coverage for TDD red/green behavior: active `noReply` admission, retained `shouldReply`, retry timer, duplicate active flush suppression, and later idle `noReply: false` resume.

## Testing

```bash
bun test src/features/background-agent/parent-wake-active-turn-event.test.ts
bun test src/features/background-agent/parent-wake-active-turn-event.test.ts src/features/background-agent/parent-wake-user-message-race.test.ts
bun test src/features/background-agent/parent-wake-*.test.ts
bun run typecheck
bun run build
git diff --check
```

TDD note: before the scheduling fix, `bun test src/features/background-agent/parent-wake-active-turn-event.test.ts` failed because the retained final wake had no retry timer after `noReply` admission. The test passed after adding retained-admission scheduling.

Foreground review:

- Goal/constraint review: PASS
- Code quality review: PASS
- Context/missed-requirements review: PASS
- QA execution: PASS
- Security review: PASS

## Related Issues

Closes #4874

---
## Summary by cubic
Preserves liveness for final “[ALL BACKGROUND TASKS COMPLETE]” wakes after an active-turn `noReply`, so the session resumes with a real reply once idle, while avoiding duplicate prompts and deferring unsafe failure wakes.

- **Bug Fixes**
  - Retain final wakes after `noReply`, track `noReplyAdmittedAt`, and schedule a retry flush; resume with a reply when the parent goes idle.
  - Defer failure/cancel/interrupted wakes and retained finals while the parent is active or otherwise unsafe; schedule a later flush instead of prompting.
  - Treat forced `noReply` admissions as non-reply-covering so later reply-required flushes aren’t deduped; track dispatched wakes without consuming reply liveness.
  - Coalesce duplicates to prevent repeated `noReply` during active turns; add timer tracking and update tests for idle resume, tool deferral window, retries, and user-message races.

<sup>Written for commit 3d0eac3992ae1a42505ab7021e0c24f3d3150844. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/background-agent/parent-wake-active-turn-event.test.ts
+++ b/src/features/background-agent/parent-wake-active-turn-event.test.ts
@@ -3,7 +3,7 @@ import { afterEach, describe, expect, test } from "bun:test"
 import type { PluginInput } from "@opencode-ai/plugin"
 import { BackgroundManager } from "./manager"
 import type { BackgroundTask } from "./types"
-import { releaseAllPromptAsyncReservationsForTesting } from "../../hooks/shared/prompt-async-gate"
+import { releaseAllPromptAsyncReservationsForTesting, releasePromptAsyncReservation } from "../../hooks/shared/prompt-async-gate"
 
 type PromptAsyncCall = {
   path: { id: string }
@@ -97,6 +97,13 @@ function getPendingParentWakes(manager: BackgroundManager): Map<string, PendingP
   return parentWakeNotifier.getPendingParentWakes()
 }
 
+function getPendingParentWakeTimers(manager: BackgroundManager): Map<string, ReturnType<typeof setTimeout>> {
+  const parentWakeNotifier = Reflect.get(manager, "parentWakeNotifier") as {
+    getPendingParentWakeTimers: () => Map<string, ReturnType<typeof setTimeout>>
+  }
+  return parentWakeNotifier.getPendingParentWakeTimers()
+}
+
 async function notifyParentSessionForTest(manager: BackgroundManager, task: BackgroundTask): Promise<void> {
   const notifyParentSession = Reflect.get(manager, "notifyParentSession") as (task: BackgroundTask) => Promise<void>
   return notifyParentSession.call(manager, task)
@@ -108,7 +115,7 @@ async function flushPendingParentWakeForTest(manager: BackgroundManager, session
 }
 
 describe("BackgroundManager parent wake active turn events", () => {
-  test("#when background task completes during active parent turn #then parent wake is recorded without forking a reply", async () => {
+  test("#when background task completes during active parent turn #then noReply admission is followed by idle resume", async () => {
     // given
     const sessionStatuses: Record<string, { type: string }> = {
       "parent-1": { type: "busy" },
@@ -132,10 +139,27 @@ describe("BackgroundManager parent wake active turn events", () => {
     // then
     expect(promptAsyncCalls).toHaveLength(1)
     expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
+    expect(getPendingParentWakes(manager).get("parent-1")?.shouldReply).toBe(true)
+    expect(getPendingParentWakeTimers(manager).has("parent-1")).toBe(true)
+
+    // when
+    await flushPendingParentWakeForTest(manager, "parent-1")
+
+    // then
+    expect(promptAsyncCalls).toHaveLength(1)
+    releasePromptAsyncReservation("parent-1", "background-agent-parent-wake")
+
+    // when
+    sessionStatuses["parent-1"] = { type: "idle" }
+    await flushPendingParentWakeForTest(manager, "parent-1")
+
+    // then
+    expect(promptAsyncCalls).toHaveLength(2)
+    expect(promptAsyncCalls[1]?.body.noReply).toBe(false)
     expect(getPendingParentWakes(manager).has("parent-1")).toBe(false)
   })
 
-  test("#when duplicate background completions overlap an active parent turn #then one coalesced admit-only wake is recorded", async () => {
+  test("#when duplicate background completions overlap an active parent turn #then coalesced noReply admission is followed by idle resume", async () => {
     // given
     const sessionStatuses: Record<string, { type: string }> = {
       "parent-1": { type: "busy" },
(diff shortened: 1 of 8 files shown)


shgew:
Stumbled into the same issue and an agent has produced a very similar fix. Would be great to merge this.

ririnto:
Solved via #5114

PR #4894, opened 2026-06-05
fix(background-agent): bypass session-active gate for child-session prompt dispatch

postrednik:
## Summary

Background task launches (`task(run_in_background=true)`) can fail to start with:

```
Task failed to start (status: interrupt)
... promptAsync skipped by gate: active
```

When this happens, sync delegation keeps working but every background subagent is rejected before it runs.

## Root cause

`BackgroundManager.launch()` and its resume/fallback paths dispatch the child-session prompt through `promptWithRetryInDirectory` (`src/shared/session-route.ts`). That wrapper forwarded only `{ queueBehavior: "defer" }` to the prompt-async gate, leaving the session-active check (`checkStatus`) enabled.

The dispatch target is a freshly created, manager-owned child session, which can momentarily report an active status (`busy`/`running`). When it does, `dispatchAfterSessionIdle` returns `{ status: "active" }`, the launch promise rejects with `promptAsync skipped by gate: active`, and `interruptTaskFromAsyncPromptFailure` marks the task `interrupt`.

The synchronous delegate path (`sendSyncPrompt`) already avoids this by passing `checkStatus: false` / `checkToolState: false`, precisely because it targets a manager-owned child session where the anti-duplicate gate does not apply.

## Fix

Pass `checkStatus: false` and `checkToolState: false` from `promptWithRetryInDirectory`, mirroring `sendSyncPrompt`. All six call sites of this wrapper are background-agent launch/resume dispatches into manager-owned child sessions, so none of them want the session-active gate.

## Tests

Added two self-validating cases in `src/shared/session-route.test.ts`:

- the gated `promptAsync` route still defers when the session reports an active status (proves the active status is detected);
- the retry wrapper dispatches without consulting session status (proves the bypass).

```
bun test src/shared/session-route.test.ts
6 pass / 0 fail
```

## Checklist

- [x] `bun run build` succeeds (ESM bundle regenerated)
- [x] Tested locally with OpenCode via the `file://` plugin method — background subagents now launch instead of interrupting
- [x] `bun run typecheck` is clean for the changed files (the change adds no type errors)

---
## Summary by cubic
Fixes background agent launch failures by bypassing the session-active gate for child-session prompts in `promptWithRetryInDirectory`. Background tasks now start reliably instead of being interrupted with "promptAsync skipped by gate: active".

- Bug Fixes
  - In `promptWithRetryInDirectory`, bypass session-active/tool-state checks with `{ queueBehavior: "defer", checkStatus: false, checkToolState: false }` (mirrors `sendSyncPrompt`) for manager-owned child sessions during background-agent launch/resume.
  - Tests in `src/shared/session-route.test.ts` confirm the gated route still defers on active, and the retry wrapper dispatches without consulting `status`.

<sup>Written for commit fa6227671b3b5dcb7f028b7cc5085f1dd820313b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4894?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/shared/session-route.test.ts
+++ b/packages/omo-opencode/src/shared/session-route.test.ts
@@ -90,6 +90,32 @@ describe("promptAsyncInDirectory", () => {
     expect(result).toBeUndefined()
     expect(promptAsync).toHaveBeenCalledTimes(1)
   })
+
+  test("#given the session reports an active status #when routing a gated promptAsync request #then the active status defers the prompt", async () => {
+    // given
+    const promptAsync = mock(async () => ({ data: "sent" }))
+    const status = mock(async () => ({ data: { ses_route_active: { type: "running" } } }))
+    const client = {
+      session: {
+        promptAsync,
+        status,
+      },
+    }
+    const args = {
+      path: { id: "ses_route_active" },
+      body: { parts: [{ type: "text", text: "continue" }] },
+    }
+
+    // when, then
+    await expect(
+      promptAsyncInDirectory(
+        unsafeTestValue(client),
+        unsafeTestValue(args),
+        "/workspace/project",
+      ),
+    ).rejects.toThrow("promptAsync skipped by gate: active")
+    expect(promptAsync).toHaveBeenCalledTimes(0)
+  })
 })
 
 describe("promptWithRetryInDirectory", () => {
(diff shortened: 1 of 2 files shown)


MoerAI:
Thanks for the PR, @postrednik. GitHub shows this as blocked from merging, and the CLA does not appear to be signed yet for your account, which is a required check. Could you post this exact comment on the PR:

Once the CLA bot picks it up the block should clear. This also changes the session-active gate path for child-session prompt dispatch, which is a sensitive area, so it will get a careful review pass after the CLA is in. Thanks for the contribution.

code-yeongyu:
Reviewed against current dev (`ae3ca8332`) with the prompt-async-gate invariants in mind. Rejecting this approach — the core design disables the exact protections the shared gate exists to provide.

The change sets `checkStatus: false, checkToolState: false` on the shared async child wrapper used by six launch/retry/resume call sites. A child session reporting active during a resume can genuinely have an in-flight model/tool turn; dispatching another internal user message without status/tool-state checks creates the duplicate/concurrent-turn condition that produced real incidents before the gate hardening (see AGENTS.md "Internal message injection is dangerous"). The added tests pin the unsafe behavior (asserting status is never consulted) rather than proving concurrent triggers collapse to one dispatch.

A viable replacement would:
1. Keep `checkStatus` and `checkToolState` enabled.
2. Use the gate's queue semantics (`queueBehavior: "enqueue"`) so an active-child prompt is accepted into the shared queue and dispatched at a safe boundary — `queued` is already an accepted outcome for the retry wrapper.
3. Separate fresh-launch vs resume semantics if they need different handling.
4. Add duplicate-injection regressions: two concurrent launches, resume-while-running, pending tool state, late `session.error` — each proving exactly one resulting prompt.

Happy to review that version. Closing this one.

PR #5023, opened 2026-06-06
fix(background-agent): requeue accepted-but-unprocessed parent wakes instead of refreshing the recovery timer forever

sjawhar:
## Problem
When a background task completes, the parent wake is dispatched via `dispatchInternalPrompt` (async). "Accepted" only means `status: "dispatched" | "queued"` — not that the parent processed it. If the recovery window elapses with no assistant/tool output, `handleDispatchedParentWakeWindowElapsed()` calls `refreshWakeTimer()` and reschedules the same timer **indefinitely** — it never requeues for redispatch. So a completion notification that was accepted-but-not-processed leaves the parent **stuck idle forever**.

## Fix
Bounded requeue: when the window elapses with no output, clone the wake (carrying `dispatchedAt` + a new `dispatchRetryCount`), clear the dispatched entry, requeue, and schedule a redispatch after a delay **>** the 15s semantic-dedupe hold (so the retry isn't coalesced into a no-op). Cap at 2 redispatch attempts, then fall back to the prior keep-tracked-for-late-output behavior.

To avoid duplicate notifications if the parent was merely slow, `flushPendingParentWake()` drops the requeued wake when `hasAssistantOrToolOutputAfterDispatchedWake()` shows the parent woke after the original dispatch.

## Test
`parent-wake-late-error-recovery.test.ts` updated: after the window elapses with no output, the wake is proactively requeued and redispatches — fails on old code (refresh-forever), passes on new.

---
## Summary by cubic
Requeues accepted-but-unprocessed parent wakes instead of refreshing the timer forever, so completion notifications reliably wake the parent. Adds a bounded retry with a late-output guard to avoid duplicates.

- **Bug Fixes**
  - Added `dispatchRetryCount` to `PendingParentWake` and clone logic.
  - On window elapsed with no output: clear dispatched entry, requeue cloned wake, and schedule redispatch after 16s to avoid dedupe collisions.
  - Capped redispatch to 2 attempts; then fall back to tracking for late output.
  - Flush runner now drops requeued wakes if assistant/tool output arrived after the original dispatch.
  - Updated `parent-wake-late-error-recovery.test.ts` to assert proactive requeue behavior.

<sup>Written for commit 385818e72ba49b56239bdf51a5c503f0374f322e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5023?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/background-agent/parent-wake-dedupe.ts
+++ b/src/features/background-agent/parent-wake-dedupe.ts
@@ -14,6 +14,7 @@ export type PendingParentWake = {
   dispatchedAt?: number
   toolCallDeferralStartedAt?: number
   allowEmptyAssistantTurnRetry?: boolean
+  dispatchRetryCount?: number
 }
 
 export function resolveParentWakePromptContext(promptContext: ParentWakePromptContext): ParentWakePromptContext {
@@ -38,6 +39,9 @@ export function cloneParentWake(wake: PendingParentWake): PendingParentWake {
       : {}),
     ...(wake.allowEmptyAssistantTurnRetry !== undefined
       ? { allowEmptyAssistantTurnRetry: wake.allowEmptyAssistantTurnRetry }
+      : {}),
+    ...(wake.dispatchRetryCount !== undefined
+      ? { dispatchRetryCount: wake.dispatchRetryCount }
       : {}),
   }
 }
--- a/src/features/background-agent/parent-wake-flush-runner.ts
+++ b/src/features/background-agent/parent-wake-flush-runner.ts
@@ -46,6 +46,17 @@ export class ParentWakeFlushRunner {
     if (!latestWake) {
       return
     }
+    if (
+      latestWake.dispatchedAt !== undefined
+      && await this.deps.sessionInspector.hasAssistantOrToolOutputAfterDispatchedWake(sessionID, latestWake)
+    ) {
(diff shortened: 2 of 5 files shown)


sjawhar:
Superseding this with https://github.com/code-yeongyu/oh-my-openagent/pull/5038.

While verifying this fix against current `dev`, the real root cause turned out to be in the history-deferral *detection* layer, not the recovery layer: a completed-but-unmarked assistant turn (`finish: "unknown"` + substantive output, no `info.time.completed`) is classified as blocking by `latestAssistantTurnBlocksInternalPrompt()`, and the staleness escape only covered tool-call blocks — so the wake deferred forever. This PR's requeue-in-window-recovery approach re-dispatched the same wake into the same block (then exhausted), which produced churn/log-spam without fixing the underlying classification. The replacement PR fixes it at the right layer (a staleness escape for stale unknown-finish turns in `parent-wake-session-history.ts`). Closing in favor of that.

Issue #4732, opened 2026-06-04
[Bug] Plugin fails to load on OpenCode Desktop due to Bun.serve requirement in createRuntimeSkillSourceServer


yale-Liyao:
## Description

The oh-my-openagent plugin completely fails to load on **OpenCode Desktop** (the Electron-based desktop app) because `createRuntimeSkillSourceServer` throws a fatal error when `Bun.serve` is unavailable.

## Environment

- **OS**: Windows 11
- **OpenCode Desktop**: `@opencode-aidesktop` (installed at `C:\Users\<user>\AppData\Local\Programs\@opencode-aidesktop\OpenCode.exe`)
- **OpenCode CLI**: v1.15.13 (works fine)
- **oh-my-openagent**: v4.7.5 (`@latest` on npm)
- **Bun**: v1.3.14 (installed, in user PATH)

## Error Log

```
ERROR service=plugin path=oh-my-openagent@latest error=Runtime skill source server requires Bun.serve failed to load plugin
```

This error appears 3 times at startup and **the entire plugin fails to load** — no agents, no hooks, no skills.

## Root Cause

In `dist/index.js` (~line 128640):

```typescript
function createRuntimeSkillSourceServer(options) {
  // ...
  const bun = globalThis.Bun;
  if (!bun) {
    throw new Error("Runtime skill source server requires Bun.serve");
  }
  // ...
}
```

This function is called during plugin initialization when `security-research` or `security-review` skills are not disabled:

```typescript
const runtimeSecuritySkills = selectRuntimeSecuritySkills(pluginConfig);
const runtimeSkillSource = runtimeSecuritySkills.length > 0
  ? deps.createRuntimeSkillSourceServer({ skills: runtimeSecuritySkills })
  : undefined;
```

The OpenCode Desktop app (`@opencode-aidesktop`) is an Electron app that does NOT include Bun runtime, so `globalThis.Bun` is `undefined`.

## Impact

- All versions since at least 4.5.12 contain this code (confirmed by checking cached packages)
- The same code exists in `@next` (4.5.12), but the error does not manifest because `security-research`/`security-review` skills are only registered when the plugin is loaded fresh (after cache clear + version upgrade)
- **Complete plugin failure** — no agent configuration loads, no hooks fire

## Workaround

Disable the security skills in `oh-my-openagent.json`:

```json
{
  "disabled_skills": ["security-research", "security-review"]
}
```

[...]

PR #4720, opened 2026-06-03
fix(desktop): fix boot crash on OpenCode Desktop by adding Node.js http fallback

Chewji9875:
# CRITICAL: Fix Boot Crash on OpenCode Desktop (Node.js Fallback for Runtime Skill Server & ConfigContext Dependency Decoupling)

## 🚨 Impact & Severity
**Critical / Blocker**: This bug entirely blocks the `oh-my-openagent` plugin from loading when executed inside the **OpenCode Desktop** client (Electron/Node.js environment). It triggers an unhandled `ReferenceError: Bun is not defined` exception during the synchronous boot phase, leading to a silent plugin load failure.

---

### A. Electron/Node.js vs. Bun Runtime Incompatibility
The dynamic runtime security skills registration server (`createRuntimeSkillSourceServer`) introduced in recent versions assumed a Bun execution environment. It directly accessed `globalThis.Bun` and invoked `Bun.serve`. In OpenCode Desktop, which operates in a standard Node.js/Electron host:
* `Bun` is undefined.
* The plugin startup wrapper synchronously executed the server instantiation, causing the plugin registration hook to abort immediately on startup.

### B. Circular Dependency on CLI Config Context
Feature modules in `src/features/` that query project/user configuration layout required access to the active directory state. However, the `ConfigContext` state manager was situated in the CLI module (`src/cli/config-manager/config-context.ts`). Importing CLI modules into internal features caused a circular dependency, leading to compiler issues and initialization race conditions.

---

### A. Non-Blocking Node.js `http` Server Fallback
We introduced an environment-agnostic abstraction in `src/features/opencode-runtime-skills/source-server.ts`:
1. **Feature Detection**: The server checks for `globalThis.Bun`. If missing, it dynamically falls back to Node's native `node:http` module.
2. **Asynchronous Handshake**: Because Node’s `server.listen()` is asynchronous, `createRuntimeSkillSourceServer` was refactored to return a `Promise<RuntimeSkillSourceServer>`. We wrapped the socket binding in a Promise resolving on the `"listening"` event.
3. **Dynamic Port Allocation**: The Node fallback binds to port `0` (system-allocated free port) on loopback `127.0.0.1`. The resolved port is dynamically queried via `server.address()` to safely formulate the source URL.

### B. Clean Separation of Config Context State
1. **State Relocation**: Moved the active `ConfigContext` and its state getters/setters (`initConfigContext`, `getConfigContext`, `resetConfigContext`) to `src/shared/opencode-config-dir.ts`.
[...]

Issue #4662, opened 2026-06-01
[Bug]: Windows OpenCode Desktop fails to load oh-my-openagent@4.6.0 with "Runtime skill source server requires Bun.serve"


Yongzhaooo:
### Prerequisites

- [x] I searched existing issues and found related Electron/Bun-runtime issues, but not this exact `4.6.0` regression.
- [x] I tested a version-pin workaround.

### Bug Description

On Windows OpenCode Desktop, `oh-my-openagent@latest` currently resolves to `oh-my-openagent@4.6.0` and the plugin repeatedly fails to load with:

```text
service=plugin path=oh-my-openagent@latest error=Runtime skill source server requires Bun.serve failed to load plugin
```

When the same OpenCode config is pinned from:

```jsonc
"plugin": ["oh-my-openagent@latest"]
```

to:

```jsonc
"plugin": ["oh-my-openagent@4.5.1"]
```

and OpenCode Desktop is restarted, the OMO agents appear again. The first pinned startup is slow, presumably because the package/runtime cache is being rebuilt, but the agents are visible again.

This looks like a `4.6.0` publish/runtime compatibility regression for OpenCode Desktop, not an agent config issue.

### Environment

- OS: Windows 11 x64
- Client: OpenCode Desktop
- OpenCode version observed in plugin logs: `1.15.10`
- Broken package: `oh-my-openagent@latest` resolving to `4.6.0`
- Working workaround: pin plugin to `oh-my-openagent@4.5.1`
- Config path: `%USERPROFILE%\.config\opencode\opencode.jsonc`
- Package cache path: `%USERPROFILE%\.cache\opencode\packages\oh-my-openagent@latest`

### Steps to Reproduce

1. On Windows OpenCode Desktop, configure the plugin as:

   ```jsonc
   {
     "plugin": ["oh-my-openagent@latest"]
   }
   ```

2. Start or restart OpenCode Desktop.
3. Check `%USERPROFILE%\.local\share\opencode\log\*.log`.
4. OMO agents do not appear / plugin load fails.
5. Pin the plugin to `oh-my-openagent@4.5.1` and restart OpenCode Desktop.
6. OMO agents appear again.

### Expected Behavior

`oh-my-openagent@latest` / `4.6.0` should load successfully under OpenCode Desktop on Windows, or fail once with a clear actionable message if Desktop requires a different runtime/package variant.

### Actual Behavior

[...]

PR #5032, opened 2026-06-07
fix(runtime-skills): support Node HTTP source server

### LINKED ISSUE #5027: v4.7.5 fails to load on OpenCode Desktop (Electron/Node) - Bun.serve in runtime security skill source server
# Bug: v4.7.5 fails to load on OpenCode Desktop (Electron/Node) — `Bun.serve` in runtime security skill source server

## Environment

| Item | Value |
|------|-------|
| OS | Windows 11 (x64) |
| OpenCode Desktop | v1.16.2 (Electron, Node.js in-process) |
| oh-my-openagent | 4.7.5 |
| Plugin config path | `~/.config/opencode/oh-my-openagent.jsonc` |
| Plugin cache path | `~/.cache/opencode/packages/oh-my-openagent@latest/` |

## Problem

oh-my-openagent fails to load on OpenCode Desktop. The Agents list only shows built-in agents (Build, Plan); all oh-my-openagent agents (Sisyphus, Hephaestus, Prometheus, Atlas, etc.) are missing.

CLI works fine — only Desktop (Electron/Node) is affected.

## Root Cause

The plugin entry `createPluginModule()` unconditionally attempts to create a runtime skill source server for security skills (`security-research`, `security-review`) via `Bun.serve()`. Desktop runs in Electron/Node where `globalThis.Bun` is `undefined`, causing the plugin to throw and abort before agents are registered.

### Error Log

```
ERROR service=plugin path=oh-my-openagent@latest
  error=Runtime skill source server requires Bun.serve
  failed to load plugin
```

Repeated across multiple restart sessions (`2026-06-06T135336.log`, `2026-06-07T054540.log`).

### Code Path

**`dist/index.js:128630-128660`** — `createRuntimeSkillSourceServer()`:
```js
function createRuntimeSkillSourceServer(options) {
  // ...
  const bun = runtime5.Bun;           // globalThis.Bun
  if (!bun) {
    throw new Error("Runtime skill source server requires Bun.serve");  // ← crashes here on Node
  }
  const server = bun.serve({
    hostname: "127.0.0.1",
    port: 0,
    fetch(request) { /* ... */ }
  });
  return { url: server.url.toString(), stop: () => server.stop(true) };
}
```

**`dist/index.js:143860-143903`** — plugin entry, called **before** agents/managers/tools initialization:
```js
const pluginConfig = deps.loadPluginConfig(input.directory, input);
const runtimeSecuritySkills = selectRuntimeSecuritySkills(pluginConfig);
// ↓ Throws on Node/Electron — Bun is undefined
const runtimeSkillSource = runtimeSecuritySkills.length > 0
  ? deps.createRuntimeSkillSourceServer({ skills: runtimeSecuritySkills })
  : undefined;
// agents, managers, tools initialization happens AFTER this point — never reached on Desktop
```

**`dist/index.js:128587-128601`** — `selectRuntimeSecuritySkills()` enables both skills by default:
```js
function selectRuntimeSecuritySkills(pluginConfig = {}) {
  const disabledSkills = new Set(pluginConfig.disabled_skills ?? []);
  const includeResearch = !disabledSkills.has("security-research");
  const includeReview = !disabledSkills.has("security-review");
  if (!includeResearch && !includeReview) return [];
  // ...
}
```

## Historical Context

This is the same class of Desktop compatibility issue reported in #3797, #3794, #3762, #3721. Previous instances involved `bun:sqlite` static imports and `globalThis.Bun` destructuring. Those were fixed in v3.17.14–v3.17.15. This is a new instance: v4.7.5 introduced `Bun.serve()` usage in the runtime security skill source server with no Node fallback.

| Version | Fatal Point | Error |
|---------|-------------|-------|
| v3.17.13 | `import "bun:sqlite"` | `Received protocol 'bun:'` |
| v3.17.14 | `globalThis.Bun` destructuring (25 occurrences) | `Cannot destructure property 'spawn'` |
| v3.17.15 | Direct `Bun.*` calls (20+ occurrences) | `Bun is not defined` |
| **v4.7.5** | **`Bun.serve()` in `createRuntimeSkillSourceServer()`** | **`Runtime skill source server requires Bun.serve`** |

## Workaround

Add to `~/.config/opencode/oh-my-openagent.jsonc`:

```jsonc
{
  // ...existing config...
  "disabled_skills": [
    "security-research",
    "security-review"
  ]
}
```

Restart Desktop after saving. This disables the runtime skill source server and allows the rest of the plugin to initialize normally.

**Verified**: Agents list (Sisyphus, Hephaestus, Prometheus, Atlas, etc.) restored on Desktop after applying this workaround.

## Suggested Fix

**Option A** (recommended): Replace `Bun.serve()` with a Node-compatible fallback using `http.createServer()`:

```js
function createRuntimeSkillSourceServer(options) {
  // ...
  const bun = runtime5.Bun;
  if (bun) {
    // existing Bun.serve() path
    const server = bun.serve({ /* ... */ });
    return { url: server.url.toString(), stop: () => server.stop(true) };
  }
  // Node fallback
  const http = require("node:http");
  const server = http.createServer((req, res) => {
    // same fetch logic, adapted for Node request/response
    // ...
  });
  server.listen(0, "127.0.0.1");
  const port = server.address().port;
  return {
    url: `http://127.0.0.1:${port}/`,
    stop: () => server.close()
  };
}
```

**Option B** (minimal): In the `!bun` branch, return a no-op instead of throwing:

```js
if (!bun) {
  console.warn("[oh-my-openagent] Runtime skill source server skipped: Bun.serve not available (Node/Electron environment)");
  return { url: undefined, stop: () => {} };
}
```

This allows the plugin to load and register all agents, while runtime security skills simply won't have a source server (acceptable degradation for non-Bun environments).

Nigmat-future:
## Summary

- keep the existing Bun.serve path for Bun runtimes
- add a Node.js `node:http` fallback for OpenCode Desktop / Electron where `globalThis.Bun` is unavailable
- await runtime skill source startup so the dynamically assigned loopback port is known before `config.skills.urls` is populated
- add coverage for the no-Bun fallback path

## Why

OpenCode Desktop loads the plugin in a Node/Electron sidecar. In published builds such as v4.7.5, `createRuntimeSkillSourceServer()` hard-requires `Bun.serve`; when `Bun` is absent, plugin startup can fail before OMO agents are registered, leaving the GUI with only native Build/Plan modes.

The current dev branch already degrades if source creation fails, but that still hides bundled runtime security skills. This change fixes the root runtime mismatch by serving the same source over Node's built-in HTTP server when Bun is not available.

This is a narrower alternative to #4720: it only touches the runtime skill source path and the direct async caller, avoiding config-context and submodule changes.

Fixes #5027.
Relates to #4732, #4662, and #4720.

## Verification

- `bun test src/features/opencode-runtime-skills/source-server.test.ts src/testing/create-plugin-module.test.ts`
- `bun run typecheck`
- `bun run build`
- Node smoke: compiled `source-server.ts` with `--target node`, started fallback server, and fetched `/index.json` via `node:http` with HTTP 200

---
## Summary by cubic
Adds a Node `node:http` fallback for the runtime skill source server when `Bun` is unavailable, so Desktop/Electron can serve bundled security skills. Also awaits server startup to get the loopback port before populating `config.skills.urls`. Fixes #5027.

- **Bug Fixes**
  - Keep `Bun.serve` for Bun; start a loopback `node:http` server otherwise, waiting for a bound port.
  - Add tests for the no-Bun path and 404 handling.
  - Allow injecting a runtime env for tests via an optional param.

- **Migration**
  - `createRuntimeSkillSourceServer` is now async; update call sites to `await` its result.

<sup>Written for commit 976be43ebf16945fd6428a444c1665347a05180d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5032?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/opencode-runtime-skills/source-server.test.ts
+++ b/packages/omo-opencode/src/features/opencode-runtime-skills/source-server.test.ts
@@ -12,7 +12,7 @@ afterEach(() => {
 describe("runtime security skill source server", () => {
   test("serves an OpenCode skill index and markdown files with matching frontmatter names", async () => {
     // given
-    const source = createRuntimeSkillSourceServer({
+    const source = await createRuntimeSkillSourceServer({
       skills: selectRuntimeSecuritySkills(),
     })
     cleanupServer = source
@@ -43,7 +43,7 @@ describe("runtime security skill source server", () => {
 
   test("returns 404 for unknown paths", async () => {
     // given
-    const source = createRuntimeSkillSourceServer({
+    const source = await createRuntimeSkillSourceServer({
       skills: selectRuntimeSecuritySkills(),
     })
     cleanupServer = source
@@ -53,5 +53,30 @@ describe("runtime security skill source server", () => {
 
     // then
     expect(response.status).toBe(404)
+  })
+
+  test("falls back to a Node HTTP server when Bun.serve is unavailable", async () => {
+    // given
+    const source = await createRuntimeSkillSourceServer(
+      {
+        skills: selectRuntimeSecuritySkills(),
+      },
+      {},
+    )
+    cleanupServer = source
+
+    // when
+    const indexResponse = await source.fetch(new Request(new URL("index.json", source.url)))
+    const index = await indexResponse.json()
+
+    // then
+    expect(source.url).toStartWith("http://127.0.0.1:")
+    expect(indexResponse.status).toBe(200)
+    expect(index).toEqual({
+      skills: [
+        { name: "security-research", files: ["SKILL.md"] },
+        { name: "security-review", files: ["SKILL.md"] },
+      ],
+    })
   })
 })
--- a/packages/omo-opencode/src/features/opencode-runtime-skills/source-server.ts
+++ b/packages/omo-opencode/src/features/opencode-runtime-skills/source-server.ts
@@ -1,3 +1,4 @@
+import { createServer, type ServerResponse } from "node:http"
 import type { RuntimeSkillSourceEntry } from "./runtime-skill-config"
 
 export type RuntimeSkillSourceServer = {
@@ -19,7 +20,9 @@ type BunServeRuntime = {
   }): BunServeServer
 }
 
-const runtime = globalThis as typeof globalThis & { Bun?: BunServeRuntime }
+type RuntimeSkillSourceRuntime = typeof globalThis & { Bun?: BunServeRuntime }
+
+const runtime = globalThis as RuntimeSkillSourceRuntime
 
 function jsonResponse(body: unknown): Response {
   return Response.json(body, {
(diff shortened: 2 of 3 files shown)


Nigmat-future:

PR #5038, opened 2026-06-07
fix(background-agent): wake parent after stale unknown-finish assistant turn instead of deferring forever

### LINKED ISSUE #5401: [Bug]: Parent wake defers forever after a stale unknown-finish assistant turn (reminder lands, no turn fires)
## Summary

When a background task completes, the parent session is supposed to be woken: the `<system-reminder>` is injected **and** a model turn is triggered so the agent acts on it. In practice the reminder lands but **no turn fires** — the user has to send a follow-up message to get the agent going.

## Root cause

`latestAssistantTurnBlocksInternalPrompt()` conservatively treats an assistant turn with `finish === undefined`/`"unknown"` as still-blocking (correct for genuinely streaming turns). When OpenCode finishes a turn but doesn't stamp `info.time.completed`, that done-but-unmarked turn is misclassified as blocking. The staleness escape in `getParentWakeSessionHistoryDeferralDecision()` only rescues **tool-call** deferrals, so an unknown-finish turn with no tool block never escapes — it defers forever and the wake never reaches the real-reply path.

Observed in production logs: `"Deferred parent wake because latest assistant turn blocks internal prompts"` repeating indefinitely while the wake is never delivered.

## Distinct from prior reports

Same *symptom* as #5147 / #5189 (closed), but a different root cause: those addressed `forceNoReply`/noReply consumption paths; this is the stale unknown-finish history-deferral never aging out.

## Fix

Tracked in PR #5038 — extends the staleness escape to cover stale unknown-finish substantive-output turns with no fresh activity, fixing the wake while preserving the streaming-turn guard.

sjawhar:
Fixes #5401

## Problem

When a background task completes, the parent session is supposed to be woken: the `<system-reminder>` is injected **and** a model turn is triggered so the agent acts on it. In practice the reminder lands but **no turn fires** — the user must send a follow-up message to get the agent going.

Root cause is in the parent-wake history-deferral logic. `latestAssistantTurnBlocksInternalPrompt()` conservatively treats an assistant turn with `finish === undefined`/`"unknown"` as still-blocking (correct for genuinely streaming turns). When OpenCode finishes a turn but doesn't stamp `info.time.completed`, that done-but-unmarked turn is classified as blocking. The staleness escape in `getParentWakeSessionHistoryDeferralDecision()` only rescues **tool-call** deferrals (`latestAssistantTurnHasToolBlock()`), so an unknown-finish turn with no tool block never escapes — it defers forever and the wake never reaches the real-reply path.

Observed in production logs: `"Deferred parent wake because latest assistant turn blocks internal prompts"` repeating indefinitely while the wake is never delivered.

## Fix

Extend the staleness escape in `parent-wake-session-history.ts` to also cover stale **unknown-finish, substantive-output** turns that have no fresh activity:

- New helper `latestAssistantTurnHasUnknownFinishSubstantiveOutput()` in `parent-wake-history-state.ts` (finish is undefined/`"unknown"`, not `messageCompleted`, has substantive output).
- New branch after the existing tool-call staleness escape: if the deferral has aged past `toolCallDeferMaxMs`, the latest assistant turn is unknown-finish with substantive output, **and** there is no fresh activity (`latestAssistantTurnHasFreshToolActivity` — message `created` / part `time`/`state.time` within the window), clear `toolCallDeferralStartedAt` and return `{ defer: false, skipPromptGateToolStateCheck: true }` so the wake fires a real reply.

### Why this layer (and not loosening the shared gate)

`latestAssistantTurnBlocksInternalPrompt()` is a shared prompt-safety gate; it cannot distinguish "done but missing completion metadata" from "still streaming" using only `finish: "unknown"` + text, so it stays conservative. `parent-wake-session-history.ts` has the missing context (accumulated deferral age, staleness threshold) and is the right place to decide the turn has been blocking too long.

### Safety / preserved invariants

- **Freshness-gated**: a turn with recent activity stays deferred (admit-only) — genuinely streaming turns are not interrupted.
- **Empty / step-only turns** have no substantive output, so the new branch doesn't fire (their existing empty-turn retry path is unchanged).
- **Completed turns** short-circuit earlier in `latestAssistantTurnBlocksInternalPrompt()` and never reach this branch.
- User-message race / issue-#4120 macOS guard runs in the flush runner after this decision and is unaffected.

## Tests

`parent-wake-history-deferral.test.ts`:
- Updated the case that pinned "stale unknown-finish text keeps deferring" → now asserts it stops deferring (`defer: false`).
- Added a flush-level test asserting a stale unknown-finish turn with no fresh activity fires a real reply (`body.noReply === false`).
- Fresh-activity and message-inspection-failure cases remain deferred (unchanged).

Full parent-wake + task-completion suite: 81 pass / 0 fail. Project typecheck clean.

---
## Summary by cubic
Stops indefinite deferrals on stale unknown-finish assistant turns so background reminders wake the parent by triggering a real reply. If the turn is stale with substantive text and no fresh activity, we clear the deferral and proceed.

- **Bug Fixes**
  - After `toolCallDeferMaxMs`, if the latest assistant turn is finish: `"unknown"` with substantive output and no fresh activity, clear `toolCallDeferralStartedAt` and return `{ defer: false, skipPromptGateToolStateCheck: true }` to fire a real reply.
  - History rules: treat any `user` message as a boundary for stale-unknown and empty/no-progress checks; keep deferring on fresh activity or a pending internal-initiator user tail; completed turns (`finish: true`) do not defer. Tests cover real-reply on stale unknown, continued deferral for fresh text and internal-tail, non-forking flush on fresh text, and no deferral on `finish: true`.

<sup>Written for commit 1fa2100623a091e58866c918a9a836969d8c04e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-history-deferral.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-history-deferral.test.ts
@@ -8,6 +8,207 @@ import { ParentWakeNotifier } from "./parent-wake-notifier"
 type ParentWakeClient = ConstructorParameters<typeof ParentWakeNotifier>[0]["client"]
 
 describe("ParentWakeNotifier — assistant history deferral", () => {
+  test("#given stale unknown-finish assistant text with no fresh activity #when checking parent wake history #then parent wake stops deferring so a real reply can fire", async () => {
+     // given
+     const originalDateNow = Date.now
+     Date.now = () => 100_000
+     const client = unsafeTestValue<ParentWakeClient>({
+       session: {
+         messages: async () => ({
+           data: [
+             {
+               info: {
+                 role: "assistant",
+                 finish: "unknown",
+                 time: { created: 90_000 },
+               },
+               parts: [{ type: "text", text: "still streaming" }],
+             },
+           ],
+         }),
+         status: async () => ({ data: { "parent-stale-text": { type: "idle" } } }),
+         promptAsync: async () => {
+           return { data: {} }
+         },
+       },
+     })
+     const notifier = new ParentWakeNotifier(
+       {
+         client,
+         directory: "/tmp/test-omo",
+         enqueueNotificationForParent: async (_sessionID, operation) => {
+           await operation()
+         },
+       },
+       {
+         pendingRetryMs: 1_000,
+         acceptedMessageSkewMs: 5_000,
+         toolCallDeferMaxMs: 5_000,
+         failureRequeueWindowMs: 5_000,
+         userMessageInProgressWindowMs: 2_000,
+       },
+     )
+     notifier.queuePendingParentWake(
+       "parent-stale-text",
+       "task complete",
+       { agent: "sisyphus" },
+       true,
+     )
+     const pendingWake = notifier.getPendingParentWakes().get("parent-stale-text")
+     expect(pendingWake).toBeDefined()
+     if (!pendingWake) {
+       throw new Error("Missing pending parent wake")
+     }
+     pendingWake.toolCallDeferralStartedAt = 90_000
+
+     try {
+       // when
+       const decision = await notifier["shouldDeferParentWakeForSessionHistory"]("parent-stale-text", pendingWake)
+
+       // then
+      expect(decision).toEqual({ defer: false, skipPromptGateToolStateCheck: true })
+     } finally {
+       Date.now = originalDateNow
+       notifier.shutdown()
+       releaseAllPromptAsyncReservationsForTesting()
+     }
+   })
+ 
+   test("#given fresh unfinished assistant text has no pending tool call #when checking parent wake history #then parent wake continues deferring", async () => {
+     // given
+     const originalDateNow = Date.now
+     Date.now = () => 100_000
+     const client = unsafeTestValue<ParentWakeClient>({
+       session: {
+         messages: async () => ({
+           data: [
+             {
+               info: {
+                 role: "assistant",
+                 finish: "unknown",
+                 time: { created: 99_000 },
+               },
+               parts: [{ type: "text", text: "still streaming" }],
+             },
+           ],
+         }),
+         status: async () => ({ data: { "parent-fresh-text": { type: "idle" } } }),
+         promptAsync: async () => {
+           return { data: {} }
+         },
+       },
+     })
+     const notifier = new ParentWakeNotifier(
+       {
+         client,
+         directory: "/tmp/test-omo",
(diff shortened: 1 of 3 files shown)


MoerAI:
Thanks for the PR, @sjawhar, this is a careful and well-documented fix with a regression test. Two notes before it can land. GitHub currently shows it as blocked from merging (a required review or check is still outstanding), and it changes the parent-wake and prompt-gate path in `parent-wake-session-history.ts`, which is one of the most sensitive areas in the codebase per `AGENTS.md`. Could you add a linked issue (Fixes #N) so the bug is tracked? Given the area it touches it will also need a careful maintainer review pass before merge. Thanks for the thorough writeup.

code-yeongyu:
Thanks for the thorough diagnosis — the bug you targeted (#5401, stale unknown-finish assistant turn deferring the parent wake forever) was fixed by #5412 (merge `1607f1421`, implementation `2ed4af0f3`), and #5401 is closed.

Current dev proof:

- Stale unknown/missing-finish detection: `packages/omo-opencode/src/features/background-agent/parent-wake-history-state.ts:71-90`
- Bounded release after `toolCallDeferMaxMs`: `packages/omo-opencode/src/features/background-agent/parent-wake-session-history.ts:117-125`
- Regression test: `packages/omo-opencode/src/features/background-agent/parent-wake-assistant-text-deferral.test.ts`

The residual delta in this branch (treating **any** user message — including internal/synthetic tails — as a scan boundary) now conflicts with later liveness work that deliberately looks through terminal noReply tails and orphaned internal wakes to prevent wake deadlocks (#5171 / `1f6d0e4bb`, #5601 / `11a35ad5f`; see `packages/utils/src/prompt-async-gate/pending-tool-turn.ts:99-110` and its tests). Applying it would re-introduce the "reminder lands, no turn fires" failure for orphaned internal wakes.

If duplicate dispatch behind a genuinely live internal turn is still reproducible for you on a current release, please file a fresh issue with a repro — that needs a state-aware fix distinguishing live internal turns from orphaned tails, not a blanket boundary change.

Closing as superseded.

Issue #4718, opened 2026-06-03
[Bug]: heuristic supportsThinking overrides snapshot/runtime reasoning metadata


Ruifeng-Zhang:
## Summary

The `supportsThinking` field in `getModelCapabilities()` evaluates heuristic family metadata **before** runtime and snapshot metadata. This means a heuristic with `supportsThinking: false` can override a snapshot entry with `reasoning: true`, causing thinking-capable models to be incorrectly classified as non-thinking.

## Current logic (broken)

```typescript
supportsThinking: override?.supportsThinking
  ?? heuristicFamily?.supportsThinking    // <-- heuristic BEFORE snapshot
  ?? runtimeThinking
  ?? snapshotEntry?.reasoning             // <-- snapshot is the LAST fallback
```

## Problem

This inverts the intended priority. Heuristic family data is a **fallback** when no authoritative metadata exists. But the current chain evaluates it second, after only explicit overrides. This means:

- A model like `kimi-k2.5` with `reasoning: true` in the models.dev snapshot can still get `supportsThinking: false` because the generic `kimi` heuristic has `supportsThinking: false`
- The issue affects any model family where the heuristic is more conservative than the actual model capability data

Contrast with how `family` is resolved — snapshot takes priority correctly:
```typescript
family: snapshotEntry?.family ?? heuristicFamily?.family  // correct: snapshot first
```

## Affected models

- `kimi-k2.5` and other Kimi models with `reasoning: true` in the snapshot
- Potentially `minimax` family models (also `supportsThinking: false` in heuristic)
- Any future models where the snapshot has more authoritative data than the heuristic

## Proposed fix

```typescript
supportsThinking: override?.supportsThinking
  ?? runtimeThinking                       // runtime metadata
  ?? snapshotEntry?.reasoning              // snapshot metadata  
  ?? heuristicFamily?.supportsThinking     // heuristic LAST (true fallback)
```

Priority: override > runtime > snapshot > heuristic

The `supportsThinkingSource` diagnostics must be updated in sync to avoid lying about the source.

## Related

- #3945 — original thinking display bug
- #4541 — first fix attempt
- #4681 — k2p6 heuristic exclusion (merged)
- #4707 — k2p6 capability unknown investigation
- #4716 — k2p heuristic fallback fix (current PR)
- #4717 — snapshot auto-refresh broken

Issue #4717, opened 2026-06-03
[Bug]: refresh-model-capabilities workflow has been failing for 10+ weeks — PR creation blocked by repo setting


Ruifeng-Zhang:
## Summary

The weekly `refresh-model-capabilities` workflow has been failing for at least 10 consecutive weeks (since March 30, 2026). The snapshot build succeeds, but the final step that creates a PR to merge the updated data always fails.

## Root Cause

The workflow calls `peter-evans/create-pull-request@v7`, which requires the repository setting **"Allow GitHub Actions to create and approve pull requests"** to be enabled. This setting is currently disabled.

Error from every run:
```
GitHub Actions is not permitted to create or approve pull requests.
```

## Impact

The bundled model capabilities snapshot (`src/generated/model-capabilities.generated.json`) is stuck at **2026-04-17**. Any model added to models.dev after that date is missing — including **kimi-for-coding/k2p6** (added April 2026). This causes capability detection to fall through to heuristics for affected models, producing `resolutionMode: "unknown"`.

This directly causes #4707 and is related to #3945.

## Fix

Go to **Settings → Actions → General** and enable:

☑️ **Allow GitHub Actions to create and approve pull requests**

The next weekly run (or a manual dispatch) will then successfully create a PR with the updated snapshot including k2p6.

## Failed Runs

| Date | Run | Step 5 (snapshot build) | Step 7 (create PR) |
|------|-----|------------------------|---------------------|
| 2026-06-01 | [26737943799](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/26737943799) | ✅ success | ❌ PR creation blocked |
| 2026-05-25 | [26385561549](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/26385561549) | ✅ success | ❌ PR creation blocked |
| 2026-05-18 | [26015796432](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/26015796432) | ✅ success | ❌ PR creation blocked |
| ... and 7 more going back to March 30 | — | — | — |

## Workaround for Users

Run locally to refresh the capabilities cache:
```bash
bunx oh-my-openagent refresh-model-capabilities
```

code-yeongyu:
Confirmed during today’s model/deprecated-model triage. This one is not a code PR candidate: the workflow is blocked by the repository Actions setting that allows GitHub Actions to create/approve PRs.

[...]

Issue #3590, opened 2026-04-23
[Bug]: MiniMax M2.7 and Kimi K2.6 fail with 'A parameter specified in the request is not valid' error


keetyHuang:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

MiniMax M2.7 and Kimi K2.6 models fail with the error 'A parameter specified in the request is not valid' when using the oh-my-openagent plugin in OpenCode TUI.

Error message:
A parameter specified in the request is not valid: %s

This issue affects only MiniMax M2.7 and Kimi K2.6 models. Other models like GLM-5.1, doubao-seed-2.0-pro, and doubao-seed-2.0-code work correctly without any errors.

The error occurs consistently when these models are used with the plugin enabled, but works fine when the plugin is disabled.

### Steps to Reproduce

1. Enable the oh-my-openagent plugin in OpenCode TUI
2. Configure a model provider (volcengine) to use MiniMax M2.7 or Kimi K2.6
3. Attempt to send any request or chat message using these models
4. Observe the error: 'A parameter specified in the request is not valid: %s'

### Expected Behavior

The models should work correctly without any errors, similar to other models like GLM-5.1, doubao-seed-2.0-pro, and doubao-seed-2.0-code.

### Actual Behavior

The request fails with error: 'A parameter specified in the request is not valid: %s'. The error is caused by the oh-my-openagent plugin sending unsupported parameters (likely related to reasoning/thinking) to the volcengine API which MiniMax M2.7 and Kimi K2.6 do not support.

### Doctor Output

```shell
N/A - Error occurs when using MiniMax M2.7 or Kimi K2.6 with oh-my-openagent plugin
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

Root Cause Analysis:

The issue is likely related to how the oh-my-openagent plugin handles reasoning/thinking parameters for different models:

1. The plugin sends thinking/reasoning parameters to the volcengine API
[...]

Issue #3559, opened 2026-04-21
[Feature] Add kimi-k2.6 to the hard-coded model catalog


XIYBHK:
Moonshot has released **kimi-k2.6**, a multimodal reasoning model. It's live on the official Moonshot endpoint and on most common OpenAI-compatible aggregators. Verified with direct `curl` against a standard OpenAI-compatible endpoint — the model accepts base64 image input and returns `reasoning_content`.

The model is not in oh-my-openagent's hard-coded catalog yet, so users who wire kimi-k2.6 through a custom OpenAI-compatible provider have to declare every capability field manually in `opencode.json`.

### Model specs (verified via endpoint)

- Context: 262,144
- Output: 16,384
- Input modalities: `text`, `image`, `video`
- Output modalities: `text`
- Reasoning: yes (returns `reasoning_content`)
- Tool call: yes
- Temperature: supported

### Suggested catalog entry

Mirroring the shape of the existing `kimi-k2.5` entry (from `dist/cli/index.js`):

```js
"kimi-k2.6": {
  id: "kimi-k2.6",
  family: "kimi",
  reasoning: true,
  temperature: true,
  toolCall: true,
  modalities: {
    input: ["text", "image", "video"],
    output: ["text"]
  }
}
```

Also a `moonshotai/kimi-k2.6` alias with the same fields, matching the existing `kimi-k2.5` / `moonshotai/kimi-k2.5` pair.

Happy to send a PR if that's the preferred workflow.

PR #5044, opened 2026-06-07
fix(model-core): mark kimi-k2.6 as not supporting arbitrary temperature

### LINKED ISSUE #5043: [Bug]: kimi-k2.6 in supplemental-entries.ts has temperature: true; Moonshot API rejects non-1 temperature for reasoning models
### Prerequisites

- [x] I am using the latest version of oh-my-openagent (4.7.5)

### Bug Description

`kimi-k2.6` is a **reasoning model** but its entry in `packages/model-core/src/model-capabilities/supplemental-entries.ts` is incorrectly marked `temperature: true`. Moonshot's OpenAI-compatible API rejects any non-`1` temperature for reasoning models with the OpenAI-style error:

```
invalid temperature: only 1 is allowed for this model
```

The capability resolver in this repo already encodes the constraint via `temperature: false` and strips the field from the outbound request when set (see [`packages/model-core/src/model-capability-resolution.ts` lines around 5720 in dist](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/packages/model-core/src/model-capability-resolution.ts), where `supportsTemperature === false` → `temperature` is removed). Because the supplemental entry says `true`, the strip never runs, opencode's default temperature passes through, and Moonshot rejects.

This is the **direct root cause** of the symptom reported (and prematurely closed) in #3590 — `Kimi K2.6 fail with 'A parameter specified in the request is not valid' error`.

### Steps to Reproduce

1. Configure `oh-my-openagent.json` so any agent uses `moonshotai/kimi-k2.6` (or `kimi-for-coding/k2p6`):
   ```json
   { "agents": { "momus": { "model": "moonshotai/kimi-k2.6", "reasoning_effort": "xhigh" } } }
   ```
2. From inside opencode, invoke that agent (e.g. `task(subagent_type="Momus - Plan Critic", ...)`).
3. Observe:
   ```
   Session error: invalid temperature: only 1 is allowed for this model
   ```
   The session terminates in 1–3 seconds before any user-visible response.

### Expected Behavior

The request succeeds. The model receives no `temperature` field (the supported pattern for reasoning models in this same file — see `gpt-5.5` and `gpt-5.4-mini-fast` below).

### Actual Behavior

Same as the symptom in #3590. Reproducible against the user's configured Moonshot endpoint (no OpenAI account required — Moonshot's API copies OpenAI's error format).

### Root Cause

Existing convention in the **same file** is correct for other reasoning models:

```ts
"gpt-5.5":          { ..., reasoning: true, temperature: false, ... }
"gpt-5.4-mini-fast":{ ..., reasoning: true, temperature: false, ... }
```

`kimi-k2.6` is the outlier and is the only kimi reasoning entry in `supplemental-entries.ts`:

```ts
"kimi-k2.6": {
  id: "kimi-k2.6",
  family: "kimi",
  reasoning: true,
  temperature: true,   // ← should be false
  toolCall: true,
  ...
}
```

The bug originated in #3559 where the entry was added; that issue body claimed "Temperature: supported", confusing **"the OpenAI-compatible API accepts the `temperature` field at all"** with **"the model accepts arbitrary values for it"**. The field is accepted; values other than `1` are not.

### Proposed Fix

One-character change: `temperature: true` → `temperature: false` for `kimi-k2.6`. PR follows.

### Related Issues / Context

- #3590 — same symptom, closed "completed" but the bug is still present in v4.7.5.
- #3559 — the supplemental entry was added here with the incorrect `temperature: true`.
- #4717 — the meta-bug that explains *why* this supplemental override exists: the auto-refresh of `model-capabilities.generated.json` has been broken for 10+ weeks, so the bundled snapshot (2026-04-17) is missing `kimi-k2.6`, forcing the manual entry.
- #4718 — adjacent fix to the resolver's `supportsThinking` priority; this PR is in the same spirit but for `supportsTemperature`.

### Note on the broader catalog

In `models.dev`'s api.json (the source of `src/generated/model-capabilities.generated.json`), the official `moonshotai` provider currently lists `kimi-k2.6`, `kimi-k2-thinking`, and `kimi-k2-thinking-turbo` with `reasoning: true, temperature: true`, while `moonshotai/kimi-k2.5` is correctly `temperature: false`. The same class of fix is needed upstream in [`anomalyco/models.dev`](https://github.com/anomalyco/models.dev) for those entries (out of scope of this PR but worth tracking).

### Doctor Output

```shell
not relevant — the failure path is independent of installer/environment
and reproduces against any Moonshot OpenAI-compatible provider endpoint.
The error originates from the remote API, not the plugin's resolver.
```

gfnord:
## Summary

- Fix `kimi-k2.6` in `supplemental-entries.ts` to declare `temperature: false`, matching its `reasoning: true` capability and the existing convention this file already uses for other reasoning models.
- Unblocks any agent configured to `moonshotai/kimi-k2.6` / `kimi-for-coding/k2p6` — the session no longer dies with `invalid temperature: only 1 is allowed for this model` from Moonshot's API.

## Changes

```diff
-		temperature: true,
+		temperature: false,
```

Inside the `"kimi-k2.6"` block in [`packages/model-core/src/model-capabilities/supplemental-entries.ts`](packages/model-core/src/model-capabilities/supplemental-entries.ts). One-character change.

## Why this is the right fix

The repo already encodes "this reasoning model does not accept arbitrary `temperature`" via `temperature: false`. The capability-resolution pipeline then strips the field from the outbound request when `supportsTemperature === false`, and the OpenAI-compatible API defaults `temperature` to `1` server-side — which reasoning models accept.

The same convention is already correctly applied in this same file for `gpt-5.5` and `gpt-5.4-mini-fast` (both `reasoning: true, temperature: false`). `kimi-k2.6` was the lone outlier and the direct cause of:

- The user-visible session error `invalid temperature: only 1 is allowed for this model` reported in #3590 (closed "completed" but still reproducing in v4.7.5).
- The bug was introduced in #3559 when this supplemental entry was first added; the original issue body claimed "Temperature: supported", conflating "the API accepts the field" with "the model accepts arbitrary values".

## Testing

Hand-verified against a Moonshot OpenAI-compatible endpoint with a Momus subagent call:

```
agents.momus.model = "moonshotai/kimi-k2.6"
agents.momus.reasoning_effort = "xhigh"
```

- **Before:** `Session error: invalid temperature: only 1 is allowed for this model` (kills the session in 1–3 s, no model output).
- **After (hot-patched node_modules with the same one-char change):** request goes out with no `temperature` field, Moonshot accepts, subagent produces a normal verdict.

I'm happy to add a unit test if you want — but the existing pattern around `gpt-5.5`/`gpt-5.4-mini-fast` already exercises the same code path, so this is purely a data correction.

```bash
bun run typecheck
bun test
```

## Out of scope (separate issue)

The bundled `src/generated/model-capabilities.generated.json` snapshot (frozen at 2026-04-17 per #4717) also marks `kimi-k2-thinking`, `kimi-k2-thinking-turbo`, `kimi-k2.5-free`, and `kimi-k2-5` as `reasoning: true, temperature: true` — same class of bug. Those entries live upstream in [`anomalyco/models.dev`](https://github.com/anomalyco/models.dev) and need a separate fix there; opencode users currently get the bad values bundled because of #4717 (snapshot refresh has been blocked for 10+ weeks). I'm deliberately keeping this PR surgical — happy to follow up with supplemental overrides for the other four if you'd prefer them masked here too while #4717 is unresolved.

## Related Issues

Closes #5043

Related: #3590, #3559, #4717, #4718

---
## Summary by cubic
Mark `kimi-k2.6` as not supporting arbitrary `temperature` to stop Moonshot API rejections. Unblocks agents using `moonshotai/kimi-k2.6` or `kimi-for-coding/k2p6`.

- **Bug Fixes**
  - Set `temperature: false` for `kimi-k2.6` in `packages/model-core/src/model-capabilities/supplemental-entries.ts`, so the resolver strips `temperature` and the API uses the default `1` required by reasoning models.

<sup>Written for commit 756f3923ba8345feae137700c9aa00c2fc3afb83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5044?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/model-core/src/model-capabilities/supplemental-entries.ts
+++ b/packages/model-core/src/model-capabilities/supplemental-entries.ts
@@ -5,7 +5,7 @@ export const SUPPLEMENTAL_MODEL_CAPABILITIES: Record<string, ModelCapabilitiesSn
 		id: "kimi-k2.6",
 		family: "kimi",
 		reasoning: true,
-		temperature: true,
+		temperature: false,
 		toolCall: true,
 		modalities: {
 			input: ["text", "image", "video"],


gfnord:

MoerAI (changes requested):
Thanks for the PR, @gfnord, and for taking the time on this.

This is linked to issue #5043, which is already CLOSED (marked completed). I checked `packages/model-core/src/model-capabilities/supplemental-entries.ts` on current `dev` and `kimi-k2.6` still has `temperature: true`, so the fix does appear to still be needed. The closed issue plus the overlapping #5045 make the status unclear, though.

Could you confirm whether this should land independently or is superseded by #5045? A small regression test for the capability change would also help. Once that is clarified I will take another look.

gfnord:
Hi,

It is superseded by #5045
<https://github.com/code-yeongyu/oh-my-openagent/pull/5045>, I will work on
it shorty to add a regression test, thanks for pointing it out.

MoerAI:
Thanks for the PR, @gfnord.

This PR is linked to issue #5043, which is already closed as completed, and `dev` already carries a `kimi-k2.6` override entry in `supplemental-entries.ts`. There is also an outstanding change request on the PR. Since the underlying issue is resolved, I cannot merge this as a fresh fix.

If kimi-k2.6 still sends a rejected temperature against current `dev`, please reopen #5043 (or open a new issue) with the reproduction and I will revisit. Closing for now per the linked-issue rule, happy to reopen if it still reproduces.

PR #5064, opened 2026-06-08
fix(session-manager): retry transient SDK failures before file fallback (fixes #4622)

### LINKED ISSUE #4622: [session-manager] session_info and session_read return inconsistent results due to transient SDK API failures
## Bug Description

`session_info` and `session_read` tools return inconsistent results for the same session ID when the OpenCode SDK HTTP API experiences transient failures. Specifically:

1. `session_info` succeeds and returns session metadata
2. `session_read` immediately fails with "Session not found" for the **same session ID**

## Root Cause Analysis

After deep investigation, the issue stems from **dual storage architecture** + **transient SDK API failures**:

### Storage Architecture
- **SDK Storage (SQLite DB)**: `~/.local/share/opencode/opencode.db` - contains full session data
- **File Storage**: `~/.local/share/opencode-data/opencode/storage/message/` - fallback storage

For sessions created purely in SQLite backend (`isSqliteBackend() === true`), **no file storage exists** as fallback.

### Failure Chain
When SDK API transiently fails:

1. `session_info` �� `getSessionInfo()` �� SDK fails �� fallback to `getFileSessionInfo()` �� **file doesn't exist** �� returns `null` �� "Session not found"
2. `session_read` �� `sessionExists()` �� SDK's `session.list()` fails �� fallback to `fileSessionExists()` �� **file doesn't exist** �� returns `false` �� "Session not found"

Both paths fail because:
- SDK API is temporarily unavailable (timeout, network hiccup, race condition)
- File storage has no data for pure-SQLite sessions

### Key Code Locations

In `dist/index.js`:
- Line 107196-107208: `sessionExists2()` - existence check with SDK + file fallback
- Line 107210-107223: `readSessionMessages2()` - message read with fallback
- Line 107102-107132: `getSdkSessionMessages()` - SDK message fetch (no retry)
- Line 107496-107521: `session_read` tool definition

### Evidence

```bash
# Session exists in SQLite with full data
sqlite3 ~/.local/share/opencode/opencode.db "SELECT COUNT(*) FROM message WHERE session_id='ses_187699fafffe5TjZ0wrO97d21b'"
# Result: 157 messages, 485 parts

# But file storage is empty for this session
ls ~/.local/share/opencode-data/opencode/storage/message/ses_187699fafffe5TjZ0wrO97d21b
# Result: directory does not exist
```

## Proposed Solution

Add **retry logic** for SDK API calls in session tools, since transient failures are expected in networked environments:

```typescript
// In session_read tool (line ~107504)
execute: async (args, _context) => {
  const maxRetries = 2;
  const retryDelay = 500; // ms
  
  for (let attempt = 0; attempt < maxRetries; attempt++) {
    if (await resolvedDeps.sessionExists(args.session_id)) {
      break;
    }
    if (attempt < maxRetries - 1) {
      await new Promise(r => setTimeout(r, retryDelay));
    }
  }
  
  if (!await resolvedDeps.sessionExists(args.session_id)) {
    return `Session not found: ${args.session_id}`;
  }
  // ... rest unchanged
}
```

Alternatively, add retry logic directly in `getSdkSessionMessages()` and `getSdkSessionInfo()` functions.

## Environment

- OpenCode version: 1.15.12
- oh-my-openagent: latest (installed via npm)
- OS: Windows 11
- Storage backend: SQLite (`isSqliteBackend() === true`)

## Workaround

Wait a few seconds and retry the `session_read` call - the SDK API typically recovers quickly.

## Related

This issue particularly affects sessions that were created in SQLite backend without corresponding file storage. The fallback mechanism works correctly, but for pure-SQLite sessions there's nothing to fall back TO when SDK fails.

MoerAI:
## Summary
`session_read` returns "Session not found" for a session that `session_info` reports fine, when the OpenCode SDK HTTP API has a transient failure. The SDK call had no retry, so a momentary network blip fell straight through to file storage, which does not exist for pure-sqlite sessions.

## Root Cause
`session_read` makes **two** SDK calls (`session.list()` for the `sessionExists` preflight, then `session.messages()`), while `session_info` makes only one (`session.messages()`). Each call site in `src/tools/session-manager/sdk-storage.ts` ran the SDK request once; on a transient throw the caller in `storage.ts` falls back to file storage. For pure-sqlite sessions there is no file fallback, so the extra `session.list()` call in `session_read` turns a transient blip into a false `Session not found` while `session_info` happens to succeed.

## Changes
| File | Change |
|------|--------|
| `src/tools/session-manager/sdk-storage.ts` | Add `fetchSdkResponse()` retry wrapper (3 attempts) used by all session SDK calls; retries only transient/unavailable errors (`isSessionSdkUnavailableError`), rethrows semantic errors immediately |
| `src/tools/session-manager/storage-fallback.test.ts` | Regression tests: transient-then-recover for `sessionExists` + `readSessionMessages`, and a non-transient no-retry guard |

## Reproduction (before fix)
```
(fail) session-manager storage fallback > ...sessionExists...retries SDK and avoids false-negative file fallback
  expect(received).toBe(expected)  Expected: true   (got false)
(fail) session-manager storage fallback > ...readSessionMessages...retries SDK instead of returning empty file storage
  expect(received).toHaveLength(expected)  Expected length: 1
 13 pass, 2 fail
```

## Verification (after fix)
```
bun test src/tools/session-manager/   ->  63 pass, 0 fail
bun run typecheck                      ->  exit 0
```

## Test
- Regression tests: `src/tools/session-manager/storage-fallback.test.ts` (newly added: 2 transient-recovery + 1 no-retry guard)
- Existing fallback + semantic-rethrow tests still pass (transient errors that never recover still fall back to file; semantic errors still rethrow)
- Typecheck: `bun run typecheck` clean

Fixes #4622

---
## Summary by cubic
Add transient retry for OpenCode SDK calls in the session manager to prevent false "Session not found" on network blips, especially for pure-sqlite sessions. Fixes #4622.

- **Bug Fixes**
  - Added `fetchSdkResponse` (3 attempts) around `session.list`, `session.messages`, and `session.todo`.
  - Retries only transient/unavailable errors (`isSessionSdkUnavailableError`); semantic errors rethrow immediately.
  - Added regression tests for transient recovery and a no-retry guard.

<sup>Written for commit d984ccd5b9de36c4eb44bfa4bda903ae83be0988. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/tools/session-manager/sdk-storage.ts
+++ b/src/tools/session-manager/sdk-storage.ts
@@ -17,13 +17,33 @@ function throwOnNonFallbackableSdkError(response: unknown): void {
   throw error
 }
 
+const SDK_TRANSIENT_RETRY_ATTEMPTS = 3
+
+// session_read issues two SDK calls (session.list for existence + session.messages),
+// so a single transient HTTP failure on either call would fall back to file storage,
+// which does not exist for pure-sqlite sessions and surfaces a false "Session not found".
+// Retry only on transient/unavailable errors; semantic errors (e.g. "session not found")
+// still throw immediately so the caller can decide between fallback and rethrow.
+async function fetchSdkResponse(operation: () => Promise<unknown>): Promise<unknown> {
+  let lastError: unknown
+  for (let attempt = 1; attempt <= SDK_TRANSIENT_RETRY_ATTEMPTS; attempt++) {
+    try {
+      const response = await operation()
+      throwOnNonFallbackableSdkError(response)
+      return response
+    } catch (error) {
+      lastError = error
+      if (!isSessionSdkUnavailableError(error)) throw error
+    }
+  }
+  throw lastError
+}
+
 export async function getSdkMainSessions(
   client: PluginInput["client"],
   directory?: string,
 ): Promise<SessionMetadata[]> {
-  const response = await client.session.list()
-  const error = unwrapSdkResponseError(response)
-  if (error) throw error
+  const response = await fetchSdkResponse(() => client.session.list())
 
   const sessions = normalizeSDKResponse(response, [] as SessionMetadata[])
   const mainSessions = sessions.filter((session) => !session.parentID)
@@ -37,15 +57,13 @@ export async function getSdkMainSessions(
 }
 
 export async function getSdkAllSessions(client: PluginInput["client"]): Promise<string[]> {
-  const response = await client.session.list()
-  throwOnNonFallbackableSdkError(response)
+  const response = await fetchSdkResponse(() => client.session.list())
   const sessions = normalizeSDKResponse(response, [] as SessionMetadata[])
   return sessions.map((session) => session.id)
 }
 
 export async function sdkSessionExists(client: PluginInput["client"], sessionID: string): Promise<boolean> {
-  const response = await client.session.list()
-  throwOnNonFallbackableSdkError(response)
+  const response = await fetchSdkResponse(() => client.session.list())
   const sessions = normalizeSDKResponse(response, [] as Array<{ id?: string }>)
   return sessions.some((session) => session.id === sessionID)
 }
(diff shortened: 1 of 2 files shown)


MoerAI:

MoerAI:
Thanks @MoerAI, merging.

PR #5078, opened 2026-06-09
fix: reasoningEffort config ignored for main TUI session

MathiasDeWeerdt:
## Problem

`reasoningEffort` configured under `agents.<name>` in `oh-my-openagent.json` was silently ignored. OpenCode reads `agent.variant` to auto-apply model variants, but oh-my-openagent stored the value as `reasoningEffort` — the two naming conventions never connected.

Manually picking a variant in the TUI worked because it sets `variant` on the session, bypassing the agent config entirely.

## Fix

3 lines in `mergeAgentConfig`. When an agent override has `reasoningEffort` but no `variant`, copy it across so OpenCode picks it up natively in `prompt.ts`.

Also:

- `model-settings-compatibility`: passthrough `reasoningEffort` for unknown model families (DeepSeek, Qwen) instead of dropping it
- `sisyphus-agent`: removed unused `log` import


MathiasDeWeerdt:

PR #5110, opened 2026-06-10
fix(runtime-fallback): retry reserved session with backoff before giving up

### LINKED ISSUE #5109: fix(runtime-fallback): Fallback silently abandoned when session reservation is still active
## Bug Description

When the primary model hits a quota error (e.g. Zhipu GLM error 1308: "已达到5小时使用上限") and OMO correctly identifies a fallback model, the fallback dispatch is silently dropped if the session still has an active reservation from the cancelled stream. The user sees the toast "Model Fallback Switching to Kimi-K2.5 for next request" but the session enters a dead state with no further response.

## Root Cause

In `createAutoRetryDispatcher` (`src/hooks/runtime-fallback/auto-retry-dispatch.ts`), when `dispatchInternalPrompt` returns `{ status: "reserved" }`, the code logs "Auto-retry skipped by promptAsync gate" and returns immediately — **no retry, no backoff, no error propagation**.

This is a race condition: when the primary model stream fails, OpenCode cancels the session prompt, but the reservation cleanup is asynchronous. OMO dispatches the fallback prompt before the reservation is fully released.

```typescript
// Current behavior (auto-retry-dispatch.ts line 133-138):
if (!isInternalPromptDispatchAccepted(promptResult)) {
  log(`[${HOOK_NAME}] Auto-retry skipped by promptAsync gate (${source})`, {
    sessionID,
    status: promptResult.status,
  })
  return  // ← silently gives up, session dies
}
```

## Timeline (observed in production)

| Time (UTC+8) | Event |
|---|---|
| 18:00:44 | User sends "继续", glm-5.1 returns 1308, OpenCode cancels |
| 18:00:46–18:01:35 | OpenCode retries glm-5.1 five more times, all 1308 |
| 18:01:35 | Final attempt → 1308 → OMO shows toast "Switching to Kimi-K2.5" |
| 18:01:35+ | OMO dispatchInternalPrompt returns "reserved" → gives up |
| 18:01:35+ | **Session dead, no response** |
| 20:06:37 | User sends "继续" again, glm-5.1 quota recovered, works |

## Expected Behavior

When `dispatchInternalPrompt` returns "reserved", OMO should retry with backoff until the reservation is released (or a maximum number of attempts is exhausted).

## Proposed Fix

Add a linear backoff retry loop for the "reserved" status (up to 6 attempts, 500ms base delay = ~10.5s total worst case):

```typescript
if (promptResult.status === "reserved") {
  const MAX_RESERVED_RETRIES = 6
  const BASE_DELAY_MS = 500
  let reservedResult = promptResult
  for (let attempt = 0; attempt < MAX_RESERVED_RETRIES; attempt++) {
    const delay = BASE_DELAY_MS * (attempt + 1)
    log(...)
    await new Promise((r) => setTimeout(r, delay))
    reservedResult = await dispatchInternalPrompt({...})
    if (reservedResult.status !== "reserved") break
  }
  if (!isInternalPromptDispatchAccepted(reservedResult)) {
    log(...)
    return
  }
} else if (!isInternalPromptDispatchAccepted(promptResult)) {
  // existing behavior for other non-accepted statuses
  log(...)
  return
}
```

## Environment

- oh-my-openagent: v4.8.1
- OpenCode: latest
- Primary model: `zhipu-coding-plan/glm-5.1`
- Fallback chain: `modelscope/Kimi-K2.5 → ...`

wjiuxing:
## Summary

Fixes #5109

When the primary model fails with a quota error (e.g. 1308) and OMO dispatches a fallback prompt, `dispatchInternalPrompt` may return `{status: "reserved"}` because the cancelled stream's reservation has not yet been cleaned up. Previously, the auto-retry dispatcher logged `"skipped by promptAsync gate"` and **silently gave up**, leaving the session in a dead state with no response.

## Changes

**`src/hooks/runtime-fallback/auto-retry-dispatch.ts`**

Added a linear backoff retry loop for the `"reserved"` dispatch status:

- Up to 6 retry attempts
- 500ms base delay with linear increase (500ms → 1000ms → 1500ms → 2000ms → 2500ms → 3000ms)
- Total worst-case wait: ~10.5 seconds
- Breaks immediately when reservation is released
- Falls through to existing rejection handling if all retries exhausted

## Test Plan

1. Configure a primary model that will return a quota error (e.g. Zhipu GLM after 5-hour quota exhausted)
2. Configure a working fallback model in the chain
3. Send a message after quota is exhausted
4. **Before fix:** Toast shows "Switching to X" but session dies silently
5. **After fix:** Toast shows "Switching to X" and fallback model responds after a brief delay

## Reproduction

This was observed in production with:
- Primary: `zhipu-coding-plan/glm-5.1` (returned error 1308)
- Fallback: `modelscope/Kimi-K2.5`
- Result: Session dead for ~2 hours until primary quota recovered

See issue #5109 for full timeline and logs.

---
## Summary by cubic
Retries fallback dispatch when a session is temporarily “reserved” after a quota error, so the fallback model can respond instead of the session dying. Also fixes a TypeScript narrowing issue by annotating the retry result.

- **Bug Fixes**
  - Retries on `reserved` from `dispatchInternalPrompt` with linear backoff (6 attempts, 500ms → 3000ms; worst case ~10.5s), exiting early once the dispatch is accepted.
  - Avoids the immediate “skipped by promptAsync gate” when still reserved; after retries, if not accepted, logs and returns as before.
  - Adds explicit `InternalPromptDispatchResult` typing for `reservedResult` to prevent TS2322 from narrowing and allow reassignment in the retry loop.

<sup>Written for commit fb12f4ece9f61ba606fcebea08ac8a1c963672a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/runtime-fallback/auto-retry-dispatch.ts
+++ b/packages/omo-opencode/src/hooks/runtime-fallback/auto-retry-dispatch.ts
@@ -8,6 +8,7 @@ import { createInternalAgentContinuationTextPart } from "../../shared/internal-i
 import {
   dispatchInternalPrompt,
   isInternalPromptDispatchAccepted,
+  type InternalPromptDispatchResult,
 } from "../shared/prompt-async-gate"
 import { isAmbiguousPostDispatchPromptFailure } from "../../shared/prompt-failure-classifier"
 import { resolveOriginalUserRetryMetadata } from "./auto-retry-metadata"
@@ -135,7 +136,7 @@ export function createAutoRetryDispatcher(
         // Retry with linear backoff until the reservation is released.
         const MAX_RESERVED_RETRIES = 6
         const BASE_DELAY_MS = 500
-        let reservedResult = promptResult
+        let reservedResult: InternalPromptDispatchResult = promptResult
         for (let attempt = 0; attempt < MAX_RESERVED_RETRIES; attempt++) {
           const delay = BASE_DELAY_MS * (attempt + 1)
           log(`[${HOOK_NAME}] Session reserved, retrying fallback dispatch in ${delay}ms (${source})`, {
--- a/packages/omo-opencode/src/hooks/runtime-fallback/auto-retry-dispatch.ts
+++ b/packages/omo-opencode/src/hooks/runtime-fallback/auto-retry-dispatch.ts
@@ -130,7 +130,50 @@ export function createAutoRetryDispatcher(
         }
         throw promptResult.error
       }
-      if (!isInternalPromptDispatchAccepted(promptResult)) {
+      if (promptResult.status === "reserved") {
+        // Session still has an active reservation from the cancelled stream.
+        // Retry with linear backoff until the reservation is released.
+        const MAX_RESERVED_RETRIES = 6
+        const BASE_DELAY_MS = 500
+        let reservedResult = promptResult
+        for (let attempt = 0; attempt < MAX_RESERVED_RETRIES; attempt++) {
+          const delay = BASE_DELAY_MS * (attempt + 1)
+          log(`[${HOOK_NAME}] Session reserved, retrying fallback dispatch in ${delay}ms (${source})`, {
+            sessionID,
+            attempt: attempt + 1,
+            maxAttempts: MAX_RESERVED_RETRIES,
+          })
+          await new Promise((r) => setTimeout(r, delay))
+          reservedResult = await dispatchInternalPrompt({
+            mode: "async",
+            client: ctx.client,
+            sessionID,
+            source: `runtime-fallback:${source}:reserved-retry-${attempt + 1}`,
+            settleMs: 0,
+            queueBehavior: "defer",
+            input: {
+              path: { id: sessionID },
+              body: {
+                ...(launchAgent ? { agent: launchAgent } : {}),
+                ...retryModelPayload,
+                ...(retryPayload.system ? { system: retryPayload.system } : {}),
+                ...(retryPayload.tools ? { tools: retryPayload.tools } : {}),
+                ...(retryMessageID ? { messageID: retryMessageID } : {}),
+                parts: retryParts,
+              },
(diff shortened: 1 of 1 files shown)


MoerAI (changes requested):
Thanks for the PR, @wjiuxing, and for taking the time on this.

This adds a new internal prompt-dispatch route: the reserved-retry loop that re-calls `dispatchInternalPrompt` inside `auto-retry-dispatch.ts`. Per the prompt-async-gate rules documented in `AGENTS.md`, any new internal message route must ship with a duplicate-injection regression test proving that concurrent, live, idle, and error triggers collapse to a single dispatch (so the retry loop cannot double-inject into a live session).

This PR changes one file with no accompanying `*.test.ts`, so the new reserved-retry path is unguarded against that class of regression.

Could you add a regression test that pins the reserved-retry behavior and proves it does not double-dispatch? Once that is in I will take another look. No rush.

wjiuxing:
Thanks for the thorough review, @MoerAI! The duplicate-injection regression test concern is completely valid.

However, code-yeongyu has already cherry-picked this fix into #5130 with full test coverage, including a reserved-retry regression test and ambiguous failure path verification. I'd suggest reviewing #5130 instead — it should address the concern you raised here.

code-yeongyu:
[sisyphus-bot] Thank you @wjiuxing — landed! Both of your commits shipped to dev verbatim via #5130 (authorship preserved), extended with ambiguous-failure preservation on the reserved-dispatch retry path (#5109). Closing this copy.

PR #5119, opened 2026-06-10
fix(plugin): load native skills via raw fallback when PluginInput.skills absent

regulusleow:
## Summary

OMO currently retrieves native skills via `"skills" in ctx ? ctx.skills : undefined`, but the OpenCode plugin SDK `PluginInput` type does not include a `skills` property (confirmed absent from v1.15.13 through v1.17.0). This means `nativeSkills` is always `undefined`. Specific impact:

- `customize-opencode`: visible in the available skills list but fails to load (`location: "<built-in>"`)
- Skills registered via `"plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]` in `opencode.json`: completely invisible

This PR adds a `createNativeSkillAccessor` function: if `ctx.skills` exists it is passed through directly (host is responsible for filtering); otherwise falls back to `ctx.client._client.get("/skill")` to fetch the full native skill list from OpenCode's HTTP API, normalizes the response, and filters by `disabledSkills`.

## Screenshots

Before: `customize-opencode` fails to load, Superpowers skill not visible. After: both work correctly.

| Before | After |
|:---:|:---:|
| 1Capture_2026-06-10_16 51 53| 1Capture_2026-06-10_16 49 46 |

## Changes

- `src/plugin/tool-registry-core-tools.ts` (+53/-0): add `createNativeSkillAccessor` function, supporting types, replace two inline `"skills" in ctx ? ... : undefined` expressions
- `src/plugin/tool-registry.test.ts` (+269/-0): 6 test cases covering success path with disabled filter and shared accessor, direct array normalization, `{data:null}` empty fallback, missing `_client`, missing `_client.get`, and concurrent dedup

## Testing

```bash
bun test src/plugin/tool-registry.test.ts
```

15 pass / 0 fail (9 existing + 6 new)

### DIFF
--- a/packages/omo-opencode/src/plugin/tool-registry-core-tools.ts
+++ b/packages/omo-opencode/src/plugin/tool-registry-core-tools.ts
@@ -12,6 +12,54 @@ import * as openclawRuntimeDispatch from "../openclaw/runtime-dispatch"
 import { log } from "../shared"
 import { getSisyphusJuniorModelOverride } from "./tool-registry-team-tools"
 
+type NativeSkillAccessor = NonNullable<SkillLoadOptions["nativeSkills"]>
+type NativeSkillList = Awaited<ReturnType<NativeSkillAccessor["all"]>>
+type NativeSkillResponse = NativeSkillList | { data?: NativeSkillList | null }
+
+type RawSkillClient = {
+  _client?: {
+    get?: (input: { url: string; query: { directory: string } }) => Promise<NativeSkillResponse>
+  }
+}
+
+function createNativeSkillAccessor(ctx: PluginContext, disabledSkills?: Set<string>): NativeSkillAccessor | undefined {
+  if ("skills" in ctx) {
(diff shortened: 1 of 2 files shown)


regulusleow:

MoerAI:
Thanks for the PR, @regulusleow. GitHub currently shows this as blocked from merging, and the change is fairly large (around 322 lines) with no linked issue. Could you add a linked issue (Fixes #N) describing the bug, and confirm the scope is as tight as it can be? Once the block clears and it is tracked I will take another look. Thanks for the contribution.

code-yeongyu:
[sisyphus-bot] Thanks for this fix, @regulusleow. MoerAI left change requests on 2026-06-18 but we haven't heard back in several weeks. The branch is also blocked from merging. We're going to close this as stale - if you'd like to revisit, please rebase onto dev and address the review feedback, then open a fresh PR.

Issue #4721, opened 2026-06-04
[Bug]: `run --json` exits before background completion parent wake is processed


ManushGupta:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

`oh-my-openagent run --json` can return `success:true` and print `All tasks completed.` after a background subagent finishes, but before the parent session processes the all-background-tasks-complete wake that should produce the final answer.

This reproduces on `oh-my-openagent` `4.8.1` with OpenCode `1.16.12`. The parent run starts, launches a background subagent, waits while the background-task continuation marker is active, then exits as soon as the child session and marker become idle. The final JSON `summary` is still the parent session's interim launch/waiting text instead of the terminal parent response that retrieves `background_output(...)`.

This is not the `messageCount:0` startup race. In this failure mode the parent session starts, the child session starts, and the child task completes. The missing step is delivery/processing of the parent wake after background completion.

I checked nearby public issues. `code-yeongyu/oh-my-openagent#3452` is the closest match for premature `run` completion with background work. `#480`, `#2627`, `#2387`, `#1517`, and `#1115` look adjacent around background lifecycle, idle detection, or JSON/event-stream behavior, but I did not find an exact duplicate for this marker-idle-before-parent-wake ordering.

### Steps to Reproduce

Run a headless JSON command that launches a background subagent and asks the parent to report the child result.

Example command:

```bash
OPENCODE_CONFIG_DIR="$HOME/.config/opencode/profiles/omo" \
CI=true \
OMO_DISABLE_POSTHOG=1 \
OMO_SEND_ANONYMOUS_TELEMETRY=0 \
oh-my-openagent run --json --verbose \
  --directory "$REPRO_DIR" \
  'create a background subagent, which executes a script to count to 10, give the current time at the end'
```

Observed local run:

| Field | Value |
|---|---|
| Parent session | `ses_1702f8e47ffePjrUHT0WFriuqh` |
| Background task | `bg_72411dcf` |
[...]

PR #5148, opened 2026-06-10
fix(background-agent): break infinite deadlock in parent wake noReply admission deferral

vboufleur:
## Summary

Fixes the infinite deadlock where background task completion notifications never wake the parent agent. After an admit-only (`noReply: true`) wake is dispatched, `latestAssistantTurnBlocksInternalPrompt()` sees the resulting synthetic user message with no assistant response and permanently blocks further wake dispatches.

## Root Cause

1. Flush runner sends admit-only notification when parent has recent activity
2. The synthetic user message refreshes `recordParentSessionActivity()` (no filter for internal messages at `manager.ts:1584`)
3. Once `noReplyAdmittedAt` is set, `deferReplyWakeWhileUnsafe` perpetually reschedules
4. After activity window expires, `latestAssistantTurnBlocksInternalPrompt()` returns `true` (unanswered synthetic message)
5. The stale-tool escape requires `latestAssistantTurnHasToolBlock` which is `false` for normally-completed agent turns
6. Result: permanent infinite reschedule loop — the wake never dispatches with `noReply: false`

## Fix

Added a timeout-based escape hatch in `getParentWakeSessionHistoryDeferralDecision` that fires when:
- `toolCallDeferralStartedAt` has exceeded `toolCallDeferMaxMs` (5s)
- There is no tool block (`!latestAssistantTurnHasToolBlock`)
- The wake was previously admitted as noReply (`noReplyAdmittedAt !== undefined`)

Returns `{ defer: false, skipPromptGateToolStateCheck: true }` to bypass both the flush runner deferral and the prompt-async-gate tool state check (which would also be fooled by the same synthetic message).

Recovery time: ~10s (5s activity window + 5s deferral timeout) instead of infinite deadlock.

## Files Changed (2 files, +233 LOC)

- `packages/omo-opencode/src/features/background-agent/parent-wake-session-history.ts` — escape hatch logic
- `packages/omo-opencode/src/features/background-agent/parent-wake-history-deferral.test.ts` — 3 regression tests

## Testing

Three regression tests added:
- **Positive**: completed assistant + synthetic user msg + `noReplyAdmittedAt` + timeout → escape fires
- **Negative**: same scenario WITHOUT `noReplyAdmittedAt` → still defers (no false positives)
- **Negative**: tool block present → existing stale-tool escape handles it (no interference)

Verified locally: built the package, pointed OpenCode config to local build, main agent launched background agents and resumed automatically after completion.

---
## Summary by cubic
Fixes a deadlock that blocked parent wakes after admit-only (`noReply: true`) notifications by adding a timeout escape in the deferral logic. Parent agents now recover in ~10s instead of looping forever.

- Bug Fixes
  - Add timeout escape in `getParentWakeSessionHistoryDeferralDecision` when deferral exceeds `toolCallDeferMaxMs`, there’s no tool block, and `noReplyAdmittedAt` is set; returns `{ defer: false, skipPromptGateToolStateCheck: true }`.
  - Add 3 regression tests: escape fires on the admit-only case; stays deferred without `noReplyAdmittedAt`; tool-block path continues to use the stale-tool escape.

<sup>Written for commit 34c16193a2fc8e5138954aea9d886501f18635ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-history-deferral.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-history-deferral.test.ts
@@ -409,4 +409,228 @@ describe("ParentWakeNotifier — assistant history deferral", () => {
       releaseAllPromptAsyncReservationsForTesting()
     }
   })
+
+  test("#given completed assistant turn followed by synthetic internal user message with noReply admitted #when deferral has timed out #then escape fires and stops deferring", async () => {
+    // This test covers the deadlock where:
+    // 1. Agent finished normally (no pending tools)
+    // 2. Background tasks completed -> admit-only noReply notification was sent
+    // 3. latestAssistantTurnBlocksInternalPrompt returns true because
+    //    there's an unanswered synthetic user message with no assistant after it
+    // 4. Without the fix, this loops forever because there's no tool block to trigger stale-tool escape
+    const originalDateNow = Date.now
+    Date.now = () => 100_000
+    const client = unsafeTestValue<ParentWakeClient>({
+      session: {
+        messages: async () => ({
+          data: [
+            {
+              info: { role: "user", time: { created: 80_000 } },
+              parts: [{ type: "text", text: "start work" }],
+            },
+            {
+              info: { role: "assistant", finish: "stop", time: { created: 85_000 } },
+              parts: [{ type: "text", text: "I delegated to background agents" }],
+            },
+            {
+              // This is the admit-only noReply notification (synthetic/internal user message)
+              info: { role: "user", time: { created: 92_000 } },
+              parts: [{ type: "text", text: "<system-reminder>\n[ALL BACKGROUND TASKS COMPLETE]\n</system-reminder>\n<!-- OMO_INTERNAL_INITIATOR -->" }],
+            },
+          ],
+        }),
+        status: async () => ({ data: { "parent-noreply-deadlock": { type: "idle" } } }),
+        promptAsync: async () => {
+          return { data: {} }
+        },
+      },
+    })
+    const notifier = new ParentWakeNotifier(
+      {
+        client,
+        directory: "/tmp/test-omo",
+        enqueueNotificationForParent: async (_sessionID, operation) => {
+          await operation()
+        },
+      },
+      {
+        pendingRetryMs: 1_000,
+        acceptedMessageSkewMs: 5_000,
+        toolCallDeferMaxMs: 5_000,
+        failureRequeueWindowMs: 5_000,
+        userMessageInProgressWindowMs: 2_000,
+      },
+    )
+    notifier.queuePendingParentWake(
+      "parent-noreply-deadlock",
+      "task complete",
+      { agent: "sisyphus" },
+      true,
+    )
+    const pendingWake = notifier.getPendingParentWakes().get("parent-noreply-deadlock")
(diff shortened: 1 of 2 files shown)


vboufleur:

ManushGupta:
Does it also fix #4721 ?

vboufleur:
@ManushGupta I've asked Claude Opus if this PR could fix your issue and to include steps on how to verify it. Here's the answer from the agent:

> This PR addresses a **related but distinct** failure mode from your issue.

> **Your issue (#4721)**: The CLI `run --json` exits too early because the background-task marker flips `idle` before the parent wake is queued/dispatched (the `unref()`'d timer hasn't fired yet). The CLI sees idle markers → exits → misses the final response.

> **This PR (#5148)**: Fixes what happens *after* the wake fires — an infinite deadlock where the admit-only notification is dispatched but the follow-up reply-producing wake gets permanently stuck due to `latestAssistantTurnBlocksInternalPrompt()` treating the synthetic message as a blocker.

>**How they relate**: These are two sequential failure points in the same flow:
>1. CLI exits before wake fires (your issue — needs the marker-stays-active-while-wake-pending invariant you proposed)
>2. Even if the CLI waits, the wake gets stuck in an infinite deferral loop and never produces an assistant response (this PR's fix)

>**Will this PR help you?** Partially:
>- In **interactive sessions** — yes, this fully resolves background agents not resuming
>- In **`run --json`** — this is necessary but not sufficient. You'd still need the complementary fix to keep the continuation marker active while parent wakes are pending, otherwise the CLI can exit in the gap before the wake even attempts dispatch

>**To test** (interactive session, not `run --json`):
>```bash
>cd /your/project
>git clone git@github.com:vboufleur/oh-my-openagent.git /tmp/omo-fix
>cd /tmp/omo-fix && git checkout fix/parent-wake-noreply-admission-deadlock && bun install && bun run build
>```
>Then in `~/.config/opencode/opencode.json`, change the plugin line:
>```json
>"plugin": ["file:///tmp/omo-fix"]
>```
>Restart OpenCode and run a prompt that fires background agents. The parent should auto-resume ~10s after tasks >complete.

Kifah:
this is an annoying bug, and I did a little workaround plugin for it....would be nice if someones merges this.

acamq:
Thank you for the excellent investigation and contribution, @vboufleur. Your root cause analysis in the issue comments was spot-on and helped inform the final fix.

Closing this PR as superseded by #5171, which takes a different approach to the same deadlock. Your PR adds a timeout-based escape hatch in the deferral decision, while #5171 fixes the underlying gate function `latestAssistantTurnBlocksInternalPrompt()` to look through noReply tails. Both approaches are valid, but #5171 addresses why the gate blocks rather than adding a timed bypass, and it fixes both gate layers (wake-deferral and prompt gate) simultaneously.

Your detailed deadlock trace was invaluable for reviewing these PRs. Hope to see more contributions from you.

PR #5150, opened 2026-06-10
feat(preemptive-compaction): proactive DCP dedup at 40% threshold

EvangelosMoschou:
## Problem

Dynamic context pruning (DCP) deduplication currently only runs reactively inside `anthropic-context-window-limit-recovery` at 78% context usage — too late. By the time the LLM summarize fires, sessions have already accumulated dozens of duplicate tool outputs (reads, greps, globs) that bloat the context window and degrade retrieval accuracy across the session lifetime.

## Fix

Run the existing DCP dedup (`executeDeduplication` + `truncateToolOutputsByCallId`) proactively on every `tool.execute.after` when token ratio >= 40%, **before** the 78% LLM summarize. Gated behind `experimental.dynamic_context_pruning.enabled = true` (off by default — no behavior change for existing users).

- **Lossless**: dedup keeps the last occurrence of each duplicate tool call with identical args. Earlier copies are pruned — the same information survives in the last copy.
- **Fast**: file reads + JSON parse + set operations. No network. No LLM.
- **Non-blocking**: dedup runs early and falls through to the 78% summarize check. Does not prevent the LLM path from firing if context still grows.

## Why 40%?

Low enough to catch bloat early (duplicate reads/greps/globs accumulate fast), high enough that sessions with minimal tools never hit it. Practical balance of proactive vs. unnecessary I/O.

## TDD Evidence

- **RED**: added test asserting dedup fires at 40% — confirmed failure (dedup was never called below 78%)
- **GREEN**: after fix, test passes; dedup fires at 40% and falls through to 78% check
- 16 preemptive-compaction tests pass (unchanged — gated behind config defaulting to false)
- 51/55 anthropic-context-window-limit-recovery tests pass (4 pre-existing failures in `empty-content-recovery-sdk.test.ts` from test isolation bleed, unrelated)

## Files

- `packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts` (+56/-1)

## Usage

```jsonc
{
  "experimental": {
    "dynamic_context_pruning": {
      "enabled": true
    }
  }
}
```

Fixes nothing (feature addition, opt-in).

### DIFF
--- a/packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts
+++ b/packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts
@@ -10,9 +10,14 @@ import type {
   CachedCompactionState,
   PreemptiveCompactionContext,
 } from "./preemptive-compaction-types"
+import type { PruningState } from "./anthropic-context-window-limit-recovery/pruning-types"
+import type { DeduplicationConfig } from "./anthropic-context-window-limit-recovery/pruning-deduplication"
+import { executeDeduplication } from "./anthropic-context-window-limit-recovery/pruning-deduplication"
+import { truncateToolOutputsByCallId } from "./anthropic-context-window-limit-recovery/pruning-tool-output-truncation"
 
 const PREEMPTIVE_COMPACTION_TIMEOUT_MS = 60_000
 const PREEMPTIVE_COMPACTION_THRESHOLD = 0.78
+const PROACTIVE_DEDUP_THRESHOLD = 0.40
 const PREEMPTIVE_COMPACTION_COOLDOWN_MS = 60_000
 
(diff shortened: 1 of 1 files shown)


code-yeongyu:
[sisyphus-bot] Review-concerns: Windows CI is failing on both test and codex-compatibility. Please investigate whether the dedup/truncation logic has cross-platform issues (path separators, file system APIs, or timing). Also consider making the 40% threshold configurable via plugin config rather than hardcoded, and document the interaction between proactive dedup and the existing compaction cooldown.

EvangelosMoschou:
Closing to prune the queue. This PR predates the current dev line by ~2,000 commits and has been open since early August, which at this point makes review unlikely. The branch is preserved and the PR can be reopened with a rebase if the change is still wanted.

PR #5159, opened 2026-06-11
fix: remove incorrect markServerRunningInProcess short-circuit

### LINKED ISSUE #5158: resolveServerUrl should detect OpenCode fallback URL and warn users about missing HTTP server
## Background

When running `opencode` in default TUI mode (without `--port`), OpenCode uses internal RPC for client-server communication. No HTTP listener is started. However, the plugin system's `ctx.serverUrl` getter returns a hardcoded fallback:

```ts
// opencode/packages/opencode/src/plugin/index.ts
get serverUrl(): URL {
    return Server.url ?? new URL("http://localhost:4096")
}
```

This makes `ctx.serverUrl` always truthy — `resolveServerUrl` can't distinguish a real server URL from the fallback.

## Current behavior

In `resolveServerUrl` (`features/tmux-subagent/resolve-server-url.ts`), when `rawServerUrl` is `"http://localhost:4096/"` (the fallback), it is returned as-is. TmuxSessionManager then uses it for `opencode attach` commands in sub-agent panes — which fail because no HTTP server is listening.

## Proposal

In `resolveServerUrl`, detect when `rawServerUrl` is the fallback value and:

1. Log a user-visible warning: "tmux visualization requires a real HTTP server. Start opencode with `--port 4096` to enable sub-agent panes."
2. Optionally, return an empty/invalid URL so downstream code can gracefully degrade (skip spawns).

## Related

- PR #5159 removes the `markServerRunningInProcess` short-circuit that was masking this issue.

Eric-GoodBoy-Tech:
## Problem

OpenCode's plugin getter for `ctx.serverUrl` always returns a truthy URL:

```ts
// opencode/packages/opencode/src/plugin/index.ts
get serverUrl(): URL {
    return Server.url ?? new URL("http://localhost:4096")
}
```

Even when `Server.listen()` was never called (default TUI mode without `--port`), the fallback `http://localhost:4096` is still a truthy `URL` object.
The guard in `createManagers`:

```ts
if (tmuxConfig.enabled && ctx.serverUrl) {
    deps.markServerRunningInProcessFn()
}
```

...thus **always evaluated to true**, short-circuiting `isServerRunning()` even when no HTTP listener existed.

This caused tmux sub-agent panes to spawn with `opencode attach` commands that failed -- nothing was listening on port 4096.

## Fix

Remove the `markServerRunningInProcess()` call entirely. `isServerRunning()` now performs a real HTTP health check (`GET /global/health`) before each tmux spawn.

- `opencode --port 4096` or `omo run`: real HTTP server -> health check passes -> tmux panes work
- `opencode` (default TUI, no `--port`): no HTTP listener -> health check fails -> spawn gracefully skipped

## Changes

- `create-managers.ts`: removed `markServerRunningInProcess` import, deps entry, and call site
- `create-managers.test.ts`: updated tests to reflect new behavior

## Related

- Fixes #5158
- OpenCode feature request: anomalyco/opencode#31821 — add `ensureServer()` to PluginInput so plugins can ensure an HTTP server is running without requiring users to pass `--port`

### DIFF
--- a/packages/omo-opencode/src/create-managers.test.ts
+++ b/packages/omo-opencode/src/create-managers.test.ts
@@ -14,7 +14,6 @@ type CleanupRegistration = {
 
 type CleanupSessionTeamRunsFn = typeof import("./features/team-mode/team-runtime/session-cleanup").cleanupSessionTeamRuns
 
-const markServerRunningInProcess = mock(() => {})
 let backgroundManagerOptions: {
   onSubagentSessionCreated?: (event: { sessionID: string; parentID: string; title: string }) => Promise<void>
 } | null = null
@@ -76,7 +75,6 @@ function createDeps(): NonNullable<Parameters<typeof createManagers>[0]["deps"]>
     registerManagerForCleanupFn: registerManagerForCleanup,
     cleanupSessionTeamRunsFn: cleanupSessionTeamRunsMock as CleanupSessionTeamRunsFn,
     createConfigHandlerFn: createConfigHandler,
-    markServerRunningInProcessFn: markServerRunningInProcess,
   }
 }
 
@@ -133,7 +131,6 @@ describe("createManagers", () => {
 
   beforeEach(() => {
     dispatchOpenClawEvent = spyOn(openclawRuntimeDispatch, "dispatchOpenClawEvent")
-    markServerRunningInProcess.mockClear()
     dispatchOpenClawEvent.mockReset()
     backgroundManagerOptions = null
     trackedPaneBySession.clear()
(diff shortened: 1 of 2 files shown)


Eric-GoodBoy-Tech:

MoerAI:
Thanks for the PR, @Eric-GoodBoy-Tech. GitHub currently shows this as blocked from merging (a required check or review is outstanding), even though CI looks green. Could you tie the PR to a tracked issue with `Fixes #N` so the behavior is recorded? Once the block clears I will take another look. Thanks for the contribution.

code-yeongyu:
[sisyphus-bot] Keeping this open — issue #5158 is still open and the markServerRunningInProcess short-circuit is still present on dev. The fix is small and addresses a real health-check bypass. Would benefit from a rebase onto current dev.

MoerAI:
The issue link is fixed and the implementation still looks focused, but GitHub continues to report unresolved merge conflicts. Please rebase onto current `dev` and clear the conflicting state; once the block is gone I will complete the promised re-review.

MoerAI (changes requested):
Thanks for the focused health-check bypass fix. The tracked issue remains open, and current `createManagers` still marks a truthy `ctx.serverUrl` as in-process:
https://github.com/code-yeongyu/oh-my-openagent/issues/5158

Please rebase onto current `dev`; GitHub still reports this head as conflicting.

Reimplement the removal against the current tmux-core server-health path. Preserve its `/health` and `/global/health` probes, per-URL success cache, and caller behavior for team tmux/herdr layouts. Do not retain an unconditional in-process short circuit based only on the SDK fallback URL.

Add deterministic coverage for:

1. fallback `http://localhost:4096` with no listener returning unavailable;
2. a real healthy listener returning available;
3. a previously successful URL not making a different URL look healthy;
4. disabled tmux remaining side-effect free.

Then run current CI and isolated real OpenCode/tmux QA. In vanilla TUI mode, prove no broken `opencode attach` pane is spawned; with a real bound server, prove the attach pane starts successfully. Record evidence under `.omo/evidence/`.

This replaces the stale merge-blocked head while retaining the prior review direction:
https://github.com/code-yeongyu/oh-my-openagent/pull/5159#issuecomment-5226654818

PR #5160, opened 2026-06-11
feat: call ctx.ensureServer() on init when tmux is enabled

Eric-GoodBoy-Tech:
## Summary

When OpenCode merges `ensureServer()` into PluginInput (anomalyco/opencode#31821), OMO should call it during initialization to ensure an HTTP server is listening before tmux sub-agent panes are spawned.

This PR adds the call site now, as a **fire-and-forget** side effect, so it works the moment OpenCode ships the API — no additional OMO changes needed.

## How it works

```ts
// create-managers.ts
if (tmuxConfig.enabled) {
    const ensureServer = (ctx as any).ensureServer
    if (typeof ensureServer === "function") {
        void ensureServer.call(ctx).then(/* success */).catch(/* warn */)
    }
}
```

- **`ensureServer` available** (OpenCode merged the API): starts HTTP server, tmux panes work out of the box
- **`ensureServer` not available** (current OpenCode): fire-and-forget check fails silently, zero impact

The `(ctx as any)` runtime check is intentional — the type isn't in `@opencode-ai/plugin` yet. Once OpenCode publishes it, we'll update to use the typed API.

## Why fire-and-forget

`createManagers()` is synchronous and used in the plugin init pipeline. Making it async would change the signature and cascade through many callers. Fire-and-forget works because:

1. `ensureServer()` returns quickly if the server is already running
2. If the server needs to start, it's ready within milliseconds — well before the first sub-agent spawn
3. If it fails, we log a warning and let `isServerRunning()` handle it later

## Changes

- `create-managers.ts`: added fire-and-forget `ensureServer()` call when tmux is enabled
- `create-managers.test.ts`: added two test cases — one verifies `ensureServer` is called, one verifies backward compatibility

## Related

- Depends on #5159 (removes the `markServerRunningInProcess` short-circuit)
- Relates to #5158 (resolveServerUrl fallback URL detection)
- Relates to #5162 (sub-agent pane auth env vars)
- Upstream: anomalyco/opencode#31821 — Feature Request: add `ensureServer()` to PluginInput

### DIFF
--- a/packages/omo-opencode/src/create-managers.test.ts
+++ b/packages/omo-opencode/src/create-managers.test.ts
@@ -197,6 +197,45 @@ describe("createManagers", () => {
     expect(markServerRunningInProcess).not.toHaveBeenCalled()
   })
 
+  it("#given tmux is enabled and ctx.ensureServer is available #when managers are created #then ensureServer is called (fire-and-forget)", async () => {
+    // ctx.ensureServer() is an upcoming OpenCode plugin API (anomalyco/opencode#31821).
+    // When OpenCode merges it, OMO will call it as a fire-and-forget side effect
+    // during init to ensure an HTTP server is listening before tmux panes spawn.
+    const ensureServerMock = mock(() => Promise.resolve())
+    const ctx = createContext("/tmp")
+    const ctxWithEnsureServer = { ...ctx, ensureServer: ensureServerMock }
+    const args = {
+      ctx: ctxWithEnsureServer,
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot] Thanks for this proactive fix! The ctx.ensureServer() API doesn't appear to have landed in OpenCode yet, and there's no linked issue tracking the dependency. Closing as stale; please reopen with a linked issue once the upstream API is available.

Issue #5149, opened 2026-06-10
[LazyCodex Bug]: codex-lsp prebuild misses bundled lsp-tools runtime in installed cache


LilMGenius:
## Bug Description
On Windows, a LazyCodex / OMO installed Codex plugin cache can fail to rebuild or repair its hook components because `codex-lsp` only resolves `lsp-tools-mcp` from the source-repository sibling layout.

When the installed cache contains bundled `lsp-tools-mcp` runtime outputs under the plugin/cache-owned layout, but does not contain source package metadata, `components/lsp/scripts/build-lsp-tools.mjs` reports missing package metadata and exits non-zero. This can keep the installed cache in a state where Codex shows repeated hook failures.

This is a narrower root cause under the broader hook-code-1 class tracked in #4722, and it is distinct from the older opencode package layout issue in #4151.

## Steps to Reproduce
1. On Windows, install or update LazyCodex / OMO as a Codex plugin so the active plugin cache is under:

```text
%USERPROFILE%\.codex\plugins\cache\sisyphuslabs\omo\0.1.0
```

2. Start or continue a Codex session after the install/update.
3. Observe hook failures such as:

```text
SessionStart hook (failed)
  error: hook exited with code 1

UserPromptSubmit hook (failed)
  error: hook exited with code 1
```

4. Attempt to rebuild/repair the installed plugin cache components.
5. `components/lsp` prebuild fails while looking for a source-tree sibling package instead of the installed cache bundled runtime.

## Expected Behavior
`codex-lsp` prebuild should handle both supported layouts:

- source checkout layout, where `packages/lsp-tools-mcp/package.json` exists and can be built
- installed plugin cache layout, where source package metadata is absent but bundled runtime outputs are already present under the installed cache

In the installed cache case, `build-lsp-tools.mjs` should consume the bundled `lsp-tools-mcp` dist and exit successfully, allowing hook component rebuild/repair to continue.

## Actual Behavior
The installed cache can contain missing hook component CLI outputs, producing `MODULE_NOT_FOUND` for hook entrypoints such as:

```text
components/rules/dist/cli.js
components/telemetry/dist/cli.js
components/ultrawork/dist/cli.js
components/ulw-loop/dist/cli.js
```

When trying to rebuild/repair the cache, the LSP component can block with:

```text
[...]

PR #5048, opened 2026-06-08
fix(omo-codex): use bundled lsp runtime in installed cache

LilMGenius:
## Summary
- Fixes #5149 by making the `codex-lsp` prebuild resolver treat installed OMO/LazyCodex cache layouts as first-class runtime layouts, not as missing source checkouts.
- Preserve the existing source-repository behavior: when `packages/lsp-tools-mcp/package.json` is present and outputs are stale, the script still runs `npm ci` and `npm run build` in the source package.
- When package metadata is absent, resolve the bundled runtime outputs from the paths the installer/cache already owns: `components/lsp-tools-mcp/dist` and `mcp/lsp/dist`.
- Add a regression test for the installed cache layout and make the fake npm fixture Windows-aware by providing `npm.cmd` plus the platform PATH delimiter.

## Root cause
`components/lsp/scripts/build-lsp-tools.mjs` supported a packaged-runtime mode, but it only checked the source-repository sibling path that resolves to `packages/lsp-tools-mcp` in a checkout.

In an installed OMO/LazyCodex cache, the source package metadata is intentionally not present. The runtime dist can instead be present under the installed plugin cache itself. In that layout, the old script reported missing `lsp-tools-mcp` package metadata even though the installer had already provided the runtime outputs needed by `codex-lsp`.

This PR aligns the prebuild resolver with the cache layouts produced by the installer instead of requiring an unpackaged source-tree shape inside `~/.codex/plugins/cache`.

## Why this is standalone
- This is not a duplicate of `code-yeongyu/lazycodex#30` or `#32`. Those issues are subagent orchestration / TOML-backed routing guidance issues and are already closed via `code-yeongyu/oh-my-openagent#5031`. This PR does not touch agent routing, planner completion semantics, or issue-30/32 guidance surfaces.
- This is not the same fix as `code-yeongyu/oh-my-openagent#4763`. That PR hardened Windows cache activation, npm/npx shim resolution, and atomic install preservation. This PR addresses a separate installed-cache prebuild contract: `codex-lsp` must be able to consume the bundled `lsp-tools-mcp` runtime from the installed cache layout when source package metadata is absent.
- This is related to the still-open hook-code-1 class in `code-yeongyu/oh-my-openagent#4722`. The concrete failure here is that a Windows installed cache can have hook components needing rebuild/repair while `codex-lsp` blocks on a source-layout-only `lsp-tools-mcp` lookup.
[...]

PR #5164, opened 2026-06-11
test(omo-codex): make lsp prebuild npm fixture Windows-safe

LilMGenius:
## Summary
- Refs #5149 by preserving Windows test coverage for the LSP prebuild path after the installed-cache runtime fix landed.
- Make the `build-lsp-tools.test.mjs` fake npm fixture work on Windows by adding an `npm.cmd` shim next to the POSIX `npm` script.
- Use Node's platform `path.delimiter` when prepending the fake bin directory to `PATH`, so the fixture does not hard-code the POSIX `:` separator.
- Keep the change test-only and scoped to the existing LSP prebuild test fixture.

## Context
This is a follow-up to #5149 and the upstream runtime fix in `238ed70610ecf750ad7bdcb5f1e2af7b881531ad`.

That commit covers the installed-cache runtime resolver itself, including `lsp-tools-mcp` and `lsp-daemon`. After that landed on `dev`, the new `lsp-prebuild-layouts.test.mjs` passes on Windows, but the older `components/lsp/scripts/build-lsp-tools.test.mjs` still fails on Windows because its test fixture only creates a POSIX `npm` executable and hard-codes `:` as the `PATH` separator.

This PR keeps the Windows test environment aligned with the runtime behavior already expected by the project: npm subprocesses should be invocable through Windows `.cmd` shims when tests exercise npm-related paths.

## Why This Remains Useful After #5149
- #5149's runtime behavior has already been fixed upstream.
- This PR does not reopen, replace, or close that runtime issue.
- It preserves the Windows regression coverage around the LSP prebuild path by making the older test fixture executable on Windows too.
- The fix is intentionally smaller than #5048: one test file, no runtime code changes.

## Verification
- `node --test packages/omo-codex/plugin/components/lsp/scripts/build-lsp-tools.test.mjs`
- `node --test packages/omo-codex/plugin/test/lsp-prebuild-layouts.test.mjs`

### DIFF
--- a/packages/omo-codex/plugin/components/lsp/scripts/build-lsp-tools.test.mjs
+++ b/packages/omo-codex/plugin/components/lsp/scripts/build-lsp-tools.test.mjs
@@ -1,7 +1,7 @@
 import assert from "node:assert/strict";
 import { chmod, copyFile, mkdir, mkdtemp, readFile, rm, utimes, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { delimiter, join } from "node:path";
 import { spawnSync } from "node:child_process";
 import test from "node:test";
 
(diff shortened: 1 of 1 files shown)


LilMGenius:
@MoerAI Thanks again for the earlier review on #5048. After #5149 was fixed directly in `238ed70610ecf750ad7bdcb5f1e2af7b881531ad`, I rechecked the Windows path and found one remaining test-only issue: the older `build-lsp-tools.test.mjs` fixture still fails on Windows because it only creates a POSIX `npm` executable and uses `:` as the PATH separator.

I split that residual piece into this smaller PR. It does not touch the runtime resolver anymore; it only keeps the existing LSP prebuild test fixture runnable on Windows and verifies alongside the new `lsp-prebuild-layouts.test.mjs` coverage.

code-yeongyu:
[sisyphus-bot] Thanks @LilMGenius for the clean follow-up to #5149 — splitting the residual test-only Windows fix into this tiny, scoped PR made it easy to verify. The `npm.cmd` shim plus `path.delimiter` change is exactly right for the `execSync` path the build script uses, and it keeps the LSP prebuild fixture runnable on Windows without touching any runtime code. Merging — thank you for the contribution!

PR #5114, opened 2026-06-10
fix(background-agent): preserve reply liveness after noReply parent wake admission

code-yeongyu:
## Summary

Primary agents (e.g. Sisyphus) stayed asleep after all background tasks completed: the final `[ALL BACKGROUND TASKS COMPLETE]` wake was admitted into the parent session as a `noReply` internal prompt for active-turn visibility, but that informational admission **consumed the reply-required wake**, so no reply-producing resume ever happened.

Root cause: regression from `6e3565ebf` ("record active parent wakes without replies", part of the duplicated-responses fix train #4725/#4638). It converted the defer paths (recent activity / history deferral / user-message race) into admit-only `noReply` sends that delete the pending wake without retaining its `shouldReply` intent, and tracked those admissions with `shouldReply: true` so `isRedundantParentWake` phantom-satisfied any later reply-required flush. `c352b515f` later restored deferral for the two `sessionActive` branches only; the remaining three admit-only paths still dropped reply liveness on current `dev`.

## Changes

- `PendingParentWake.noReplyAdmittedAt`: reply-required wakes survive `noReply` admission — retained in the pending queue, stamped, and given a retry flush so the parent resumes **exactly once** when idle and safe (`noReply: false`).
- While the parent stays unsafe, an already-admitted retained wake is **not** re-admitted (no duplicate reminders); the marker resets only when genuinely new notification content merges in.
- Forced-`noReply` admissions are tracked as `shouldReply: false` in the dispatched tracker, so dedupe can no longer treat an informational admission as a completed reply-producing dispatch.
- Failure/cancel/interrupt wakes stay queued (never admit-only) until the parent session is safe, per #4874.

Both constraints from #4874 hold: active-turn `noReply` visibility (#4164/#4478) is preserved, and the duplicate-response guards (#4725/#4638/#4653) are untouched — reply wakes still only dispatch through the full prompt-gate safety checks, and all dedupe/race suites pass unchanged.

## Testing

- RED→GREEN: new `parent-wake-noreply-liveness.test.ts` (6 tests) — history-deferral/user-message/recent-activity admissions retain `shouldReply` + resume `noReply: false` exactly once; failure wakes defer with 0 prompts; dispatched tracker records admissions as non-reply-covering; pending-queue marker resets on new content. All 6 failed on `dev` before the fix.
[...]

PR #5169, opened 2026-06-11
fix(background-agent): prevent retained noReply parent wake stalls

### LINKED ISSUE #4874: [Bug]: all-complete parent wake can be consumed as noReply before session resume

### LINKED ISSUE #5086: [Bug]: primary agent is not really after all background agent is finished
### Prerequisites

### Bug Description

Yes, as the title says, primary agents like Sisyphus don't wake up after all background agents are finished.

### Steps to Reproduce

sisyphus delegates to background agents, all agents are finished, sisyphus remains asleep

### Expected Behavior

Sisyphus woke up after all the agents were finished

### Actual Behavior

Sisyphus keeps sleeping

### Doctor Output

```shell
nothing
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Linux

### OpenCode Version

1.16.2

SoShymKing:
## Summary

Builds on '9f426b823' by tightening retained 'noReply' parent-wake filtering with a timestamp guard.

This preserves retained 'noReply' liveness for all-complete background-task notifications while preventing older matching internal wake messages from being treated as the current retained admission.

## Changes

- Replace the inline retained 'noReply' deposit stripping with helper-based safety filtering.
- Only strip a matching retained admission when its message timestamp is missing or at/after 'noReplyAdmittedAt'.
- Keep the existing '9f426b823' behavior that allows the retained wake to resume once its own latest deposit is the trailing internal message.
- Add regression coverage for:
  - retained 'noReply' wake resuming after the prior assistant turn is safe
  - older matching internal wake messages remaining blocking

## Screenshots

N/A

## Testing

```bash
bun test packages/omo-opencode/src/features/background-agent/parent-wake-noreply-liveness.test.ts packages/omo-opencode/src/features/background-agent/parent-wake-history-deferral.test.ts --bail
bun test packages/omo-opencode/src/features/background-agent --bail
bun run typecheck
bun run build
git diff --check
```

## Related Issues

Closes #4874  
Closes #5086  
Refs #5114

---
## Summary by cubic
Fixes stalls caused by retained ‘noReply’ parent wakes. Reply-required wakes now resume as soon as the prior assistant turn is safe, even when the retained noReply deposit is the latest internal message (closes #4874, #5086; refs #5114).

- **Bug Fixes**
  - Use helper-based filtering to ignore the retained ‘noReply’ admission when checking parent-wake safety, including when it’s the latest internal message.
  - Guard by timestamp so only messages at/after `noReplyAdmittedAt` count; older lookalikes are ignored.
  - Add tests for resume behavior once safe and for stale internal wake handling.

<sup>Written for commit 4fdb8c9dae308f42c532fd98452377ccb6d0d7e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5169?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


SoShymKing:
CI failed on 'archive extraction preflight > rejects tar archives with traversal entries before extraction' due to a 5000ms timeout.
I reproduced the targeted test locally and it passed.
This looks like a flaky timeout under full-suite load.
Could someone with Actions permission rerun failed jobs?

code-yeongyu:
[sisyphus-bot] Thanks @SoShymKing — sharp diagnosis of a real race. Closing as superseded: #5171 (merged hours after this PR's head) solves the core scenario deterministically by tagging noReply admissions with OMO_INTERNAL_NOREPLY_MARKER at dispatch, and #5181 landed its own version of the Windows launcher test fixes bundled here. The residual timestamp-guard adds little over the marker approach. Appreciate the investigation — it was the right problem.

PR #5176, opened 2026-06-11
fix: eliminate duplicate background task completion notifications

### LINKED ISSUE #5177: fix: duplicate background task completion notifications
## Bug Description

When a background task completes, the `<system-reminder>` notification is dispatched **twice** to the parent session, resulting in duplicate visible notifications in the conversation.

## Steps to Reproduce

1. Start an OpenCode session
2. Launch a background task (e.g., `task(subagent_type="explore", run_in_background=true, ...)`)
3. Wait for the task to complete
4. Observe two identical `<system-reminder>` blocks in the conversation

## Expected Behavior

A single `<system-reminder>` notification should appear when a background task completes.

## Actual Behavior

Two identical `<system-reminder>` notifications appear. The first does not wake the session. The second wakes the session.

## Root Cause

In `parent-wake-flush-runner.ts`, the `flushPendingParentWake` method has 3 code paths that use the "admit-only" dispatch pattern (`sendParentWakePrompt` with `forceNoReply: true` and `retainPendingWake: true`):

1. `hasRecentParentSessionActivity` — parent session was recently active
2. `toolWaitDecision.defer` — tool state suggests deferral
3. `isUserMessageInProgress` — user message just arrived

Each admit-only dispatch injects the full `<system-reminder>` notification text into session history without waking the session. Later, when the retained wake is re-dispatched as a reply, the **same notification content is injected again**, producing duplicates.

## Environment

- oh-my-openagent: 4.9.1
- OpenCode: 1.17.3
- OS: macOS (Darwin 25.5.0, ARM64)

HyeokjaeLee:
## Problem

When a background task completes, the `<system-reminder>` notification is dispatched **twice** to the parent session, resulting in duplicate visible notifications.

## Root Cause

In `parent-wake-flush-runner.ts`, the `flushPendingParentWake` method has 3 code paths that call `sendParentWakePrompt` with `forceNoReply: true` and `retainPendingWake: true` (the "admit-only" pattern):

1. `hasRecentParentSessionActivity` — parent session was recently active
2. `toolWaitDecision.defer` — tool state suggests deferral  
3. `isUserMessageInProgress` — user message just arrived

Each admit-only dispatch injects the full `<system-reminder>` notification text into session history without waking the session. Later, when the retained wake is re-dispatched as a reply, the **same notification content is injected again**, producing duplicates.

## Fix

Replace the 3 `sendParentWakePrompt(forceNoReply: true, retainPendingWake: true)` calls with `schedulePendingParentWakeFlush(sessionID)`. This defers the flush without injecting anything into history. When the session becomes idle, a single full dispatch delivers the notification once.

All existing safety checks (`deferReplyWakeWhileUnsafe`) are preserved.

## Before / After

**Before:** Background task completion → two identical `<system-reminder>` blocks in conversation  
**After:** Background task completion → one `<system-reminder>` block, session wakes correctly

---
## Summary by cubic
Stops duplicate background task completion notifications by deferring the parent wake flush instead of admit-only dispatches. Users now see a single `<system-reminder>`, and the session wakes once.

- **Bug Fixes**
  - Replaced admit-only `sendParentWakePrompt(...forceNoReply: true, retainPendingWake: true)` in three cases (recent parent activity, deferred tool wait, in-progress user message) with `schedulePendingParentWakeFlush(sessionID)`; updated tests to expect 0 `promptAsyncCalls`, remove `noReply` checks, and revert normal-dispatch assertions (incl. `session.messages` rejection).
  - Preserves `deferReplyWakeWhileUnsafe`; no history injection; a single flush runs when idle.

<sup>Written for commit c302b876be118f26ec94c12c5001e9aeae15fdb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5176?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Fixes #5177

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-active-turn-event.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-active-turn-event.test.ts
@@ -231,8 +231,7 @@ describe("BackgroundManager parent wake active turn events", () => {
     await flushPendingParentWakeForTest(manager, "parent-1")
 
     // then
-    expect(promptAsyncCalls).toHaveLength(1)
-    expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
+    expect(promptAsyncCalls).toHaveLength(0)
     expect(getPendingParentWakes(manager).get("parent-1")?.shouldReply).toBe(true)
   })
 
@@ -267,8 +266,7 @@ describe("BackgroundManager parent wake active turn events", () => {
     await flushPendingParentWakeForTest(manager, "parent-1")
 
     // then
-    expect(promptAsyncCalls).toHaveLength(1)
-    expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
+    expect(promptAsyncCalls).toHaveLength(0)
     expect(getPendingParentWakes(manager).get("parent-1")?.shouldReply).toBe(true)
   })
 })
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts
@@ -57,13 +57,8 @@ export class ParentWakeFlushRunner {
       if (this.deferReplyWakeWhileUnsafe(sessionID, latestWake)) {
         return
       }
-      await this.sendParentWakePrompt(sessionID, latestWake, {
-        emptyAssistantTurnRetry: false,
-        toolWaitDecision: { defer: false, skipPromptGateToolStateCheck: true },
-        forceNoReply: true,
-        retainPendingWake: latestWake.shouldReply,
-      })
-      log("[background-agent] Recorded admit-only parent wake because parent session activity is still fresh:", {
+      this.schedulePendingParentWakeFlush(sessionID)
+      log("[background-agent] Deferred parent wake (no admit-only) because parent session activity is still fresh:", {
         sessionID,
       })
       return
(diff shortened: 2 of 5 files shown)


HyeokjaeLee:
After looking into this more closely, the issue behavior appears to have changed since v4.9.3 — background agents now silently terminate without delivering completion notifications to the parent session, rather than producing duplicate <system-reminder> blocks.

PR #5185 takes a broader approach: it fixes transport-disconnect-induced permanent running state, adds session existence checks, and importantly takes the opposite direction on parent-wake-flush-runner.ts — it admits completion wakes as noReply during active turns (instead of deferring all admits like this PR does). The two approaches conflict on the same file.

I think this PR (#5176) is likely superseded by #5185. Closing for now — happy to reopen if the duplicate-notification issue persists after #5185 lands.

MoerAI:
Thanks for the PR, @HyeokjaeLee.

This is linked to issue #5177, which is already closed as completed, and most of the change here is already present on current `dev` (the duplicate background-task notification fix landed). Merging a second copy would double up on a fix that already shipped.

Could you check whether duplicate completion notifications still happen against current `dev`? If they do, please reopen #5177 (or open a fresh issue) with the current behavior and I will take another look. If not, no further action needed. Closing for now per the linked-issue rule, happy to reopen if it still reproduces.

PR #5190, opened 2026-06-12
fix(background-agent): schedule re-flush for reply-required wake after activity window

### LINKED ISSUE #5189: Parent wake sent with forceNoReply=true, agent never auto-responds to background task completion
## Bug Description

When a background task (subagent) completes, the parent-wake pipeline correctly detects completion and dispatches a wake notification to the parent session (Sisyphus/Atlas). However, the wake is always sent with `forceNoReply: true` due to `hasRecentParentSessionActivity()` returning true within the 5-second activity window. The `retainPendingWake` flag is set, but no re-flush is ever scheduled, so the wake is effectively lost and the orchestrator agent never generates a follow-up response.

## Reproduction

1. Start opencode with oh-my-openagent plugin
2. Ask Sisyphus to launch a background task: `task(subagent_type="explore", run_in_background=true, ...)`
3. Wait for `<system-reminder>[BACKGROUND TASK COMPLETED]` to appear
4. Observe: Sisyphus does NOT respond automatically

## Root Cause

In `parent-wake-flush-runner.ts`, when `hasRecentParentSessionActivity()` returns `true`:

```typescript
if (this.hasRecentParentSessionActivity(sessionID)) {
    if (this.deferReplyWakeWhileUnsafe(sessionID, latestWake)) {
        return;  // re-scheduled only for failure wakes
    }
    await this.sendParentWakePrompt(sessionID, latestWake, {
        forceNoReply: true,
        retainPendingWake: latestWake.shouldReply  // saved but never re-flushed
    });
    return;  // ← BUG: no re-flush scheduled for shouldReply=true wakes
}
```

`deferReplyWakeWhileUnsafe()` only returns `true` for failure wakes or wakes that already have `noReplyAdmittedAt` set. For a successful single-task completion where `shouldReply=true` and it is the first noReply dispatch, the function returns `false`, and no re-flush is scheduled.

The `retainPendingWake: true` saves the wake in the pending queue, but without a scheduled flush, it remains there indefinitely.

## Environment

- OpenCode: 1.17.4
- oh-my-openagent: 4.9.2
- OS: macOS (darwin-arm64)
- Providers: zai-coding-plan/glm-5.1, kimi-code/kimi-for-coding (reproduces on all)
- Config: no `disabled_hooks`, all default hooks enabled

## Diagnosis

The `<system-reminder>` visible in the chat comes from OpenCode core (not oh-my-openagent). The `OMO_INTERNAL_NOREPLY` HTML comment marker confirms oh-my-openagent's parent-wake IS firing, but with `noReply: true`. This means the pipeline works up to the dispatch point, but the `forceNoReply` gate prevents the agent from responding.

## Suggested Fix

When `retainPendingWake: true` and `shouldReply: true`, schedule a re-flush after the activity window expires:

```typescript
if (this.hasRecentParentSessionActivity(sessionID)) {
    if (this.deferReplyWakeWhileUnsafe(sessionID, latestWake)) {
        return;
    }
    await this.sendParentWakePrompt(sessionID, latestWake, {
        forceNoReply: true,
        retainPendingWake: latestWake.shouldReply
    });
    // FIX: schedule re-flush so the retained wake gets retried
    if (latestWake.shouldReply) {
        this.schedulePendingParentWakeFlush(sessionID, PARENT_WAKE_SESSION_ACTIVITY_IN_PROGRESS_WINDOW_MS + 500);
    }
    return;
}
```

## Workaround

None found. The issue reproduces on all providers and all projects.

WhiteGiverMa:
## Summary

- Fix background agent wake deadlock: orchestrator agent never responds after task completes
- When `hasRecentParentSessionActivity()` fires and wake `shouldReply`, schedule a re-flush instead of leaving the retained wake stranded forever
- Targeted 3-line fix in `parent-wake-flush-runner.ts` — no behavioral change outside the activity-window path

## Changes

- **`packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts`** (+3 lines)
  - After forceNoReply admission in the `hasRecentParentSessionActivity()` block, call `this.schedulePendingParentWakeFlush(sessionID)` if `latestWake.shouldReply`
  - The retained wake previously stayed in the pending queue with no retry scheduled; now it retries after the activity window expires

## Testing

```bash
bun test packages/omo-opencode/src/features/background-agent/parent-wake-*.test.ts
```

Result: **78/78 pass, 0 fail** across 15 files.

Key coverage:

- parent-wake-activity-window.test.ts — verifies wake is admitted without forking a reply when parent has recent activity ✅
- parent-wake-noreply-liveness.test.ts — "recent-activity admission liveness" test verifies parent resumes after activity goes stale ✅
- All existing liveness, dedupe, deferral, and regression tests unaffected ✅

## Related Issues

Closes #5189 

---
## Summary by cubic
Schedules a re-flush for reply-required parent wakes retained during recent session activity, preventing stuck pending wakes and restoring orchestrator replies after the activity window. Closes #5189.

- **Bug Fixes**
  - When hasRecentParentSessionActivity() is true and a wake shouldReply, call schedulePendingParentWakeFlush(sessionID) so the retained wake retries after the activity window instead of remaining pending.
  - Assert the timer is armed in the recent-activity admission liveness test to catch regressions.

<sup>Written for commit c97528736c5e3f288d2b340ef9426ccf26d655aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5190?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts
@@ -66,6 +66,9 @@ export class ParentWakeFlushRunner {
       log("[background-agent] Recorded admit-only parent wake because parent session activity is still fresh:", {
         sessionID,
       })
+      if (latestWake.shouldReply) {
+        this.schedulePendingParentWakeFlush(sessionID)
+      }
       return
     }
 
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-noreply-liveness.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-noreply-liveness.test.ts
@@ -639,9 +639,12 @@ describe("BackgroundManager parent wake recent-activity admission liveness", () 
       expect(promptAsyncCalls).toHaveLength(1)
       expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
       expect(parentWakeNotifier.getPendingParentWakes().get("parent-1")?.shouldReply).toBe(true)
+      // regression: the reply-required wake must be scheduled for re-flush
+      expect(parentWakeNotifier.getPendingParentWakeTimers().has("parent-1")).toBe(true)
 
       // when: activity goes stale and the gate hold expires
       now = 110_000
+      parentWakeNotifier.clearPendingParentWakeTimer("parent-1")
       releaseParentWakeHold("parent-1")
       await flushPendingParentWake.call(manager, "parent-1")
 


WhiteGiverMa:
tests passed, real UX QA during. High intensity using on it. Can be reviewed if needed.

MoerAI:
Thanks for tracking this deadlock down, @WhiteGiverMa. The fix itself reads correctly: after the admit-only `return` in the `hasRecentParentSessionActivity()` branch of `parent-wake-flush-runner.ts`, a reply-required wake was left in the pending queue with no retry scheduled, and calling `schedulePendingParentWakeFlush(sessionID)` is the right way to arm the retry.

One gap before this can merge: the new scheduling behavior is not covered by a regression test. The closest test, `#given completion admitted during fresh parent activity #then the parent resumes with a reply once activity goes stale` in `parent-wake-noreply-liveness.test.ts`, drives the second resume by calling `flushPendingParentWake.call(manager, "parent-1")` by hand after advancing the clock. Because it triggers that second flush manually, it passes on `dev` with or without your three lines, so it would not catch a future change that dropped the `schedulePendingParentWakeFlush` call.

Could you add an assertion that the timer is actually armed on the admit-only path? Right after the first flush, something like:

```ts
expect(parentWakeNotifier.getPendingParentWakeTimers().has("parent-1")).toBe(true)
```

and ideally let the scheduled timer drive the resume instead of a manual `flushPendingParentWake`, so the test fails without the fix and passes with it. That locks the behavior in. Everything else looks good and CI is green across all platforms.

WhiteGiverMa:
regression test on the way. The ux testing revealed an extreme state issue: when the sub-agent results are about to be returned, the sisy query stops responding after checking the sub-agent status, triggering a request with an empty user message for a short period of time. This will be further verified in the actual build.

WhiteGiverMa:
Added the timer-armed assertion after the admit-only flush. The second resume is still driven by a manual flushPendingParentWake (consistent with the test pattern used throughout this file — every other test that checks timer state also uses manual re-flushing). Let me know if you'd like me to refactor to timer-driven resume as a follow-up. @MoerAI

MoerAI:
Thanks @WhiteGiverMa, merging. This schedules the re-flush for reply-required parent wakes in `parent-wake-flush-runner.ts` (#5189) and adds a regression test, with CI green on `dev`.

PR #5195, opened 2026-06-12
fix(doctor): warn on unresolved tui entry

### LINKED ISSUE #5173: [Bug]: doctor reports false pass for unresolvable TUI plugin (null !== false)
## Bug Description

`oh-my-openagent doctor` has a `null` vs `false` mismatch that causes it to falsely report a passing check when TUI plugin registration is broken.

When `tui.json` contains `"oh-my-openagent/tui"` but the package has no `./tui` subpath export (see #4643), the doctor should warn that the registration is **unresolvable**. Instead, it reports **0 warnings** and exits with a clean pass.

## Root Cause

In `src/cli/doctor/tui-plugin-config.ts`, function `packageJsonExportsTui()`:

1. It constructs a path to `~/.config/opencode/node_modules/oh-my-openagent/package.json`
2. Calls `existsSync()` on that path
3. If the file doesn't exist → returns `null` (not `false`)

But in the doctor check that consumes this value, the condition is:

```ts
if (packageJsonExportsTui() === false) {
  // warn "unresolvable"
}
```

Since `null !== false` (strict equality), the unresolvable branch is **never reached**. The doctor skips the warning entirely and reports pass.

### Why the file doesn't exist

When oh-my-openagent is installed via bun (the typical path), it lives in `~/.bun/install/global/node_modules/`, **not** `~/.config/opencode/node_modules/`. So `existsSync` always returns `false`, and `packageJsonExportsTui()` always returns `null` — meaning the unresolvable check is **dead code in practice**.

## Steps to Reproduce

1. Run `bunx oh-my-openagent doctor --verbose` — observe warning: "TUI plugin entry missing from tui.json"
2. Add `"oh-my-openagent/tui"` to `~/.config/opencode/tui.json`
3. Run `bunx oh-my-openagent doctor --verbose` again — observe **0 warnings** (false pass)
4. Start opencode — TUI plugin **never actually loads** (no sidebar, no TUI commands)

## Expected Behavior

Doctor should warn: "TUI plugin `oh-my-openagent/tui` is registered in tui.json but the package does not export a `./tui` subpath — registration is unresolvable."

## Actual Behavior

Doctor reports 0 warnings. User believes their config is correct, but TUI plugin never loads.

## Fix Suggestion

In `tui-plugin-config.ts`, change the `packageJsonExportsTui()` return type contract:

**Option A**: Return `false` instead of `null` when the file doesn't exist (the "not installed locally" case should also be treated as unresolvable).

**Option B**: Change the doctor check from `=== false` to `!packageJsonExportsTui()` to also catch `null`.

Additionally, the path construction should account for bun's global install location (`~/.bun/install/global/node_modules/`) rather than only `~/.config/opencode/node_modules/`.

## Impact

- Users following doctor's advice to add `"oh-my-openagent/tui"` to tui.json get a false sense of correct configuration
- The TUI plugin silently fails to load with no diagnostic feedback
- This masks bug #4643 (the underlying missing `./tui` export)

## Related Issues

- #4643 — The underlying problem: `./tui` subpath export doesn't exist in any released version
- #5072 — Blank TUI screen when plugin enabled (likely caused by the same root cause cascade)

## Verified Against

- `oh-my-openagent` versions v4.1.2 through v4.9.1 and current `dev` branch: **none** have `exports["./tui"]` in `package.json`
- Commit `08959d63` introduced the doctor check referencing `./tui` as if it existed (May 15)
- Commit `6764d991` added `packageExportsTui` check (June 2)
- Commit `8fe5af71` allowed `file:` entries — but still uses `=== false` check

JSap0914:
Fixes #5173.

`doctor` now warns when `tui.json` contains the canonical `oh-my-openagent/tui` entry but the package cannot be inspected from the OpenCode config `node_modules` path.

The false pass came from treating `packageExportsTui === null` as inconclusive even when a named canonical TUI entry was present. In Bun/global installs, the package is often outside `~/.config/opencode/node_modules`, so the check could pass while `oh-my-openagent/tui` was still unresolvable.

Changed:

- Added a regression test for a canonical named TUI entry with no resolvable installed package.
- Kept the legacy `oh-my-opencode/tui` compatibility pass.
- Warn for canonical named entries unless the registered package explicitly exports `./tui`.

Validation:

- RED: `bun test packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.test.ts --bail` failed with `Expected: "warn" Received: "pass"`.
- GREEN: `bun test packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.test.ts` passed 12 tests.
- Typecheck: `bun run typecheck:packages` passed.
- Build: `bun run build` passed.
- tmux QA: isolated `OPENCODE_CONFIG_DIR`, `opencode.json` with `oh-my-openagent`, `tui.json` with `oh-my-openagent/tui`, then `bun dist/cli/index.js doctor --json`; the TUI Plugin result was `warn` with `TUI plugin entry in tui.json is unresolvable`.

Plan: `.omo/plans/quick-fix-pr-waves-20260612.md`

### DIFF
--- a/packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.test.ts
+++ b/packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.test.ts
@@ -66,7 +66,9 @@ describe("tui-plugin-config check", () => {
   })
 
   it("passes when both server and TUI entries are registered", async () => {
-    //#given opencode.json has the server entry and tui.json has the TUI entry
+    //#given opencode.json has the server entry, tui.json has the TUI entry,
+    //#       and the installed package exports ./tui
+    writeInstalledPackage(PLUGIN_NAME, { ".": "./dist/index.js", "./tui": "./dist/tui.js" })
     writeOpenCodeConfig([PLUGIN_NAME])
     writeTuiConfig([`${PLUGIN_NAME}/tui`])
 
@@ -77,6 +79,21 @@ describe("tui-plugin-config check", () => {
     expect(result.status).toBe("pass")
     expect(result.issues).toHaveLength(0)
     expect(result.name).toBe("TUI Plugin")
+  })
+
+  it("warns when a named TUI entry points at a missing installed package", async () => {
+    //#given opencode.json has the server entry, tui.json has the TUI entry,
+    //#       but the installed package cannot be resolved from the config node_modules path
+    writeOpenCodeConfig([PLUGIN_NAME])
+    writeTuiConfig([`${PLUGIN_NAME}/tui`])
+
+    //#when running the check
+    const result = await checkTuiPluginConfig()
+
+    //#then doctor flags the named TUI entry as unresolvable instead of falsely passing
+    expect(result.status).toBe("warn")
+    expect(result.issues).toHaveLength(1)
+    expect(result.issues[0].title).toContain("unresolvable")
   })
 
   it("warns when server is registered but TUI entry is missing", async () => {
--- a/packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.ts
+++ b/packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.ts
@@ -34,6 +34,7 @@ interface TuiPluginInfo {
   configPath: string | null
   exists: boolean
   hasNamedTuiEntry: boolean
+  hasCanonicalNamedTuiEntry: boolean
 }
 
 function fileEntryPackageJsonPath(entry: string): string {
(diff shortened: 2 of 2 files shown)


cosmosjeon:

JSap0914:

JSap0914:
recheck

MoerAI:
Thanks for the PR, @JSap0914.

This is linked to issue #5173, which is already closed as completed, and almost all of this change is already present on current `dev` (the doctor TUI-entry false-pass fix landed). Merging it again would duplicate a fix that already shipped.

Could you confirm whether doctor still reports a false pass for an unresolvable TUI plugin against current `dev`? If it does, please reopen #5173 (or open a fresh issue) and I will take another look. If not, no further action needed. Closing for now per the linked-issue rule, happy to reopen if it still reproduces.

PR #5224, opened 2026-06-13
fix(delegate-task): deterministic deliverable selection + wake-gated sync completion

acamq:
## Problem

When a sync subagent (e.g. the plan agent invoked via `task(run_in_background=false)`) launches its own background tasks (explore/librarian via `run_in_background=true`), the caller received incomplete or wrong output:

1. **Wrong turn returned.** `fetchSyncResult` returns the newest assistant turn with text. After the subagent's background children complete, a notification-triggered follow-up turn ("results confirm the plan") can become the *newest* turn, so the caller gets a brief acknowledgement instead of the actual plan written in an earlier turn.

2. **Returned too early.** The sync poll loop only gave a fixed **5s** settle window after children cleared. Parent-wake notification dispatch (debounce + queue + promptAsync gate) routinely exceeds 5s, so the loop returned the *pre-results* turn before the continuation turn even landed.

## Fix

**Deliverable envelope (which turn).** The plan agent now wraps its final plan in a `<plan>…</plan>` envelope, and `fetchSyncResult` selects the newest turn carrying a *complete* envelope instead of guessing by recency. This keys extraction on the turn the model **marked** as its deliverable — something positional/recency rules fundamentally cannot disambiguate (the deliverable's position depends on whether the model wrote it before or after results arrived).

- Gated on `isPlanAgent()` via a new `getDeliverableTag()`, so extraction only expects a tag when the prompt asked for one.
- A missing or unclosed envelope falls back to the prior recency behavior — strictly safer than the old path, never a hang.

**Wake-gated completion (when to return).** 

- `hasActiveChildTasks` is now descendant-aware (`getAllDescendantTasks`), correct even when a child spawns its own background work.
- New `hasPendingParentWake` predicate lets the poll loop keep waiting while a parent-wake is queued, scheduled, or dispatched-but-not-yet-consumed — bridging the notification dispatch window the fixed 5s grace could not.
- The loop treats the task as done only when **no descendant work remains AND no wake is owed**. The dispatched wake clears on the continuation turn's first output, so the loop reliably returns *that* turn. The existing 30-min inactivity timeout remains the safety backstop (so a stuck notifier can't hang forever).

The notifier is **not** suppressed for subagent sessions — it remains the sole waker; the poll loop just stops declaring completion while a continuation is owed, so a correctly-waiting subagent is still woken and its real deliverable returned.

## Tests

- `sync-result-fetcher.test.ts`: the failure sequence (plan in an earlier turn + brief follow-up → returns the plan), newest-envelope-wins, no-envelope fallback, unclosed-envelope fallback.
- `sync-session-poller.test.ts`: a pending parent wake holds the loop past a tiny settle window even after children clear.
- Full `delegate-task` suite (449) and `background-agent/manager` suite (186) pass; touched files typecheck clean.

---
## Summary by cubic
Fixes wrong or early results from sync subagents by selecting the plan deliverable from a `<plan>` tag in final text and holding completion until the parent-wake continuation starts, including the teardown and in-flight dispatch windows.

- **Bug Fixes**
  - Deterministic deliverable: `fetchSyncResult` picks the newest complete `<plan>...</plan>` from final text only (non‑greedy, last complete block per message), falls back to newest text if missing/unclosed, gated by `getDeliverableTag()`. `strictAbortRecovery` now validates the latest assistant turn (error/empty) before any envelope extraction.
  - Wake‑gated completion: the poll loop waits until no direct child tasks (`manager.hasActiveChildTasks`) and no owed parent wake (`manager.hasPendingParentWake`). The owed‑wake check now covers notification‑preparation reservations (pre‑queue teardown gap), pending queue/timers, in‑flight dispatch, and dispatched wakes. The flush runner marks dispatch as in‑flight before deleting the pending entry; the manager reserves a parent wake before flipping a child to terminal and releases it after enqueue (also on async‑prompt failure), using a per‑parent counter to handle concurrent teardowns.

- **Tests**
  - Deliverable selection: newest‑tagged over newer untagged; newest of multiple envelopes; last complete envelope within a message; reasoning‑only envelopes ignored; no‑envelope and unclosed‑envelope fallbacks; abort guard beats older envelopes.
  - Wake gating: pending, in‑flight, and teardown‑gap reservations hold the loop past the settle window until the continuation turn starts; added regression tests for the dispatch and teardown windows.

<sup>Written for commit 31cd979900d911464a1e6f064b5d41376a8aaa5b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


acamq:
@codex please review, thanks

acamq:
@codex please review again

acamq:
@codex review again please

acamq:
@codex review

acamq:
@codex review

PR #5253, opened 2026-06-13
fix(background-task): remove background_output polling CTA from all launch paths (#5221)

### LINKED ISSUE #5221: [Bug]: Kimi K2 models immediately poll background_output after launching background tasks
## Summary

Kimi K2 models (notably `kimi-for-coding/k2p7`, and to a lesser extent `k2p6`) immediately poll `background_output` after launching background tasks instead of waiting for the `<system-reminder>` completion notification.

## Configuration

- OMO version: 4.9.2
- OpenCode: 1.17.4
- Models affected: `kimi-for-coding/k2p7`, `kimi-for-coding/k2p6`

```json
{
  "agents": {
    "sisyphus": {
      "model": "kimi-for-coding/k2p7"
    }
  }
}
```

## Observed behaviour

The orchestrator launches three background agents, says it will wait, then immediately calls `background_output` on all three. After seeing they are still running, it apologises and says it will stop polling — then polls again. This loop repeats even when the prompt explicitly forbids polling, indicating the tool-level CTA is overriding the prompt instruction.

## Root cause 1: `isKimiK2Model` excludes `k2p7`

In `packages/model-core/src/model-family-detectors.ts` (v4.9.2):

```ts
export function isKimiK2Model(model: string): boolean {
  const modelName = extractModelName(model).toLowerCase()
  if (modelName.includes("kimi")) return true
  if (/k2[-.]?p[56]/.test(modelName)) return true   // <-- only p5/p6
  return false
}
```

For `kimi-for-coding/k2p7`, `modelName` is `k2p7`, which does **not** match `/k2[-.]?p[56]/`. Therefore `isKimiK2Model("kimi-for-coding/k2p7")` returns `false`, and `src/agents/sisyphus.ts` falls through to the default prompt instead of `buildKimiK26SisyphusPrompt`.

The `HEURISTIC_MODEL_FAMILY_REGISTRY` in `model-capability-heuristics.ts` already correctly recognises `k2p7` as a thinking-capable Kimi model via `/k2(?:.*-(?:thinking|think)|[-.]?p\d)/`, so the capability detection and the prompt routing are inconsistent.

## Root cause 2: v4.9.2 still advertises `background_output` in background task responses

The v4.9.2 release still returns the older background-task launch text:

```
System notifies on completion. Use `background_output` with task_id="..." to check.

Do NOT call background_output now. Wait for <system-reminder> notification first.
```

The model sees the `background_output` CTA before the negation, and Kimi K2 is especially prone to follow the more recent/repeated instruction (recency bias). `BACKGROUND_TASK_DESCRIPTION` also still says `Use \`background_output\` to get results.`

Commit `912f0811a07439a47da0f93333aec55b263a9832` ("fix(prompt): remove background_output CTA from tool returns and agent prompts") removed this CTA, but GitHub reports that `912f081...v4.9.2` is `behind_by: 1`, meaning v4.9.2 does **not** contain that commit.

## Why prompt_append is not enough

Adding an anti-polling instruction via `prompt_append` reduces the behaviour but does not eliminate it. The model still polls immediately after launching tasks because the tool description and tool return value themselves contain the `background_output` CTA. Prompt-level instructions lose to tool-level recency bias for Kimi K2.

## Proposed fixes

1. Update `isKimiK2Model` to match any `k2[-.]?p\d+`, e.g.:
   ```ts
   if (/k2[-.]?p\d/.test(modelName)) return true
   ```
2. Cherry-pick/merge commit `912f081` into the release branch so the `background_output` CTA is removed from tool launch responses and tool descriptions.
3. Add `k2p7` (and ideally future `k2p\d+`) to the `isKimiK2Model` test cases in `packages/model-core/src/model-family-detectors.test.ts`.

## Related

- PR #3697 — introduced `isKimiK2Model()` and the Kimi K2.x Sisyphus prompt (matches only `k2p5`/`k2p6`).
- Issue #4304 — reports similar polling behaviour with deepseek-v4-pro, but the root causes here are Kimi-specific routing + lingering CTA in the shipped bundle.
- Commit `912f0811a07439a47da0f93333aec55b263a9832` — the anti-polling CTA removal that is missing from v4.9.2.

## Workaround (partial, for users)

Adding an explicit anti-polling `prompt_append` may reduce but will not stop the polling:

```json
"prompt_append": "CRITICAL: after launching any background task, you MUST wait for a <system-reminder> notification before calling background_output. Never poll. If the task is not complete, do something else or ask the user."
```

A complete fix requires the code changes listed above.

EvangelosMoschou:
## Problem

Kimi K2 (k2p5/k2p6/k2p7) is especially prone to follow the most recent/repeated instruction (recency bias). The launch return text and `BACKGROUND_TASK_DESCRIPTION` both advertised:

- `Use \`background_output\` to get results`
- `Use \`background_output\` with task_id="..." to check`

This CTA overrides the `Do NOT call background_output now` anti-polling line, so the model immediately calls `background_output` instead of waiting for the `<system-reminder>` completion notification.

## Fix

Removed the polling CTAs from all 6 launch/description paths and reworded them to lead with the no-polling instruction:

- `packages/omo-opencode/src/tools/background-task/constants.ts` (BACKGROUND_TASK_DESCRIPTION)
- `packages/omo-opencode/src/tools/background-task/create-background-task.ts` (launch return)
- `packages/omo-opencode/src/tools/call-omo-agent/background-agent-executor.ts` (launch return)
- `packages/omo-opencode/src/tools/call-omo-agent/background-executor.ts` (launch return)
- `packages/omo-opencode/src/tools/delegate-task/background-task.ts` (launch return)
- `packages/omo-opencode/src/tools/delegate-task/background-continuation.ts` (continuation return)

## Footprint confirmation (per review feedback)

The 11-file diff is **6 source files + 5 test files**, all mechanical:

- **6 source files** (`-4 to -6 lines each`): every source change is the **same** mechanical removal — drop the `Use \`background_output\` with task_id="..."` line, keep the `Do NOT call background_output now...` anti-polling instruction (slightly reworded to lead with the no-polling message and clarify that the system delivers the result automatically).
- **5 test files** (`+27 to +76 lines each`): one regression test per affected source file asserting the launch/continuation return no longer contains the polling CTA while still carrying the anti-polling instruction.

No unrelated changes are included.

## Scope (re-scoped from original PR)

The original PR included two changes — the `isKimiK2Model` regex fix (k2p7+) and the `background_output` CTA removal. The regex fix is already in upstream via [80ae81031](https://github.com/code-yeongyu/oh-my-openagent/commit/80ae81031), so this PR carries only the CTA removal. Reduced from 13 to 11 files.

## TDD Evidence

- 5 regression tests added (one per affected source file), each asserting:
  - Launch/continuation return does NOT contain `background_output` polling CTA
  - Launch/continuation return DOES contain anti-polling instruction
- All tests pass after the fix

Fixes #5221

### DIFF
--- a/packages/omo-opencode/src/tools/background-task/constants.ts
+++ b/packages/omo-opencode/src/tools/background-task/constants.ts
@@ -1,6 +1,8 @@
-export const BACKGROUND_TASK_DESCRIPTION = `Run agent task in background. Returns a background task ID (\`bg_...\`) immediately; notifies on completion.
+export const BACKGROUND_TASK_DESCRIPTION = `Run agent task in background. Returns a background task ID (\`bg_...\`) immediately and notifies on completion.
 
-Use \`background_output\` to get results. Prompts MUST be in English.`
+Do NOT poll for results. The system delivers a <system-reminder> when the task finishes.
+
+Prompts MUST be in English.`
 
 export const BACKGROUND_OUTPUT_DESCRIPTION = `Get output from background task. Use full_session=true to fetch session messages with filters. System notifies on completion, so block=true rarely needed. - Timeout values are in milliseconds (ms), NOT seconds.
 
--- a/packages/omo-opencode/src/tools/background-task/create-background-task.test.ts
+++ b/packages/omo-opencode/src/tools/background-task/create-background-task.test.ts
@@ -155,4 +155,33 @@ describe("createBackgroundTask", () => {
     expect(secondResult).toContain("Task ID: task-2")
     expect(secondResult).not.toContain("interrupt")
   })
+
+  test("does not advertise background_output CTA in launch return (issue #5221)", async () => {
+    //#given - a successful launch
+    launchMock.mockResolvedValueOnce({
+      id: "test-task-id",
+      sessionId: "ses-launch-cta",
+      description: "Test task",
+      agent: "test-agent",
+      status: "pending",
+    })
+    getTaskMock.mockReturnValueOnce({
+      id: "test-task-id",
+      sessionId: "ses-launch-cta",
+      description: "Test task",
+      agent: "test-agent",
+      status: "pending",
+    })
+
+    //#when
+    const result = await tool.execute(testArgs, testContext)
+
+    //#then - the return must NOT nudge the model into calling background_output
+    // before the <system-reminder> arrives (Kimi K2 recency bias pulls it through)
+    expect(result).not.toContain("Use `background_output` with task_id=")
+    expect(result).not.toContain("to check.")
+    // ...but the anti-polling instruction must still be present
+    expect(result).toContain("Do NOT call background_output now")
+    expect(result).toContain("<system-reminder>")
+  })
 })
--- a/packages/omo-opencode/src/tools/background-task/create-background-task.ts
+++ b/packages/omo-opencode/src/tools/background-task/create-background-task.ts
@@ -117,9 +117,7 @@ Description: ${task.description}
 Agent: ${task.agent}
 Status: ${task.status}
 
-System notifies on completion. Use \`background_output\` with task_id="${task.id}" to check.
-
-Do NOT call background_output now. Wait for <system-reminder> notification first.`
+Do NOT call background_output now. Wait for <system-reminder> notification first. The system will deliver the result when the task completes; you do not need to poll for it.`
       } catch (error) {
         const message = error instanceof Error ? error.message : String(error)
         return `[ERROR] Failed to launch background task: ${message}`
--- a/packages/omo-opencode/src/tools/call-omo-agent/background-agent-executor.test.ts
+++ b/packages/omo-opencode/src/tools/call-omo-agent/background-agent-executor.test.ts
@@ -1,0 +1,76 @@
+/// <reference types="bun-types" />
+import { describe, test, expect, mock } from "bun:test"
+import type { BackgroundManager } from "../../features/background-agent"
+import type { PluginInput } from "@opencode-ai/plugin"
+import { executeBackgroundAgent } from "./background-agent-executor"
+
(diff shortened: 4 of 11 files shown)


EvangelosMoschou:
**Re-scoped**: original PR included two changes — the `isKimiK2Model` regex fix (k2p7+) and the `background_output` CTA removal.

The regex fix is **already in upstream** via [80ae81031](https://github.com/code-yeongyu/oh-my-openagent/commit/80ae81031f14bb2f6063565487300e9e46d3e1d3) (`feat(model-core): detect Kimi K2.7 distinctly from K2.6`), which widens the regex to `p[567]` and adds a separate `isKimiK27Model` for K2.7-specific routing. **This PR drops the regex change.**

The CTA removal is still needed (the polling CTAs are still in all 6 launch/description paths in upstream `dev`). That's what this PR addresses.

MoerAI:
Thanks for the PR, @EvangelosMoschou. This addresses #5221, which is helpful. The breadth is the main thing to confirm: it spans 11 files and around 249 lines because it removes the polling call-to-action from every launch path. That is plausibly inherent to the change, but it is large enough that it needs a careful review pass before merge. Could you add `Fixes #5221` to the description so the issue auto-closes, and confirm the 11-file footprint is all the same mechanical removal? Thanks for the work.

EvangelosMoschou:
Thanks for the review, @MoerAI. Both addressed:

1. **Added `Fixes #5221`** to the description — this will auto-close the issue when the PR merges.

2. **Confirmed the 11-file footprint is mechanical** — it's 6 source files + 5 test files, and every source change is the same mechanical removal: drop the `Use \`background_output\` with task_id="..." to check.` line, keep the anti-polling instruction (slightly reworded to lead with no-polling).

No unrelated changes are mixed in.

PR #5257, opened 2026-06-13
fix(team-mode): avoid cancelling active tasks on rejected delete

### LINKED ISSUE #5255: team_delete cancels active member tasks before rejecting non-force delete
## Prerequisites

- [x] I will write this issue in English (see the Language Policy in CONTRIBUTING.md)
- [x] I am using the latest version of oh-my-openagent from `origin/dev`
- [x] I have read the documentation and could not find an existing tracking issue for this team-delete cancellation ordering problem

## Bug Description

Non-force `team_delete` cancels team background tasks before checking whether active members make the delete invalid.

The docs describe `team_delete` as the final cleanup step after active members have been shut down/approved. If members are still active, non-force delete correctly throws `members still active`, but current `origin/dev` performs background task lookup and cancellation before that guard.

Current ordering in `deleteTeam()` is:

1. Load runtime state.
2. If `bgMgr` exists, find matching team tasks and call `cancelTask(...)`.
3. If `force !== true` and non-lead members are still active, throw `members still active`.

That means a rejected non-force delete can still terminate active member work.

## Steps to Reproduce

On current `origin/dev`, add a targeted regression with one active member and one shutdown-approved member:

```ts
await updateMemberStatuses(fixture.teamRunId, fixture.config, {
  "member-a": "running",
  "member-b": "shutdown_approved",
})

await expect(deleteTeam(fixture.teamRunId, fixture.config, undefined, bgMgr))
  .rejects.toThrow("members still active")
expect(getTasksByParentSessionMock).not.toHaveBeenCalled()
expect(cancelTaskMock).not.toHaveBeenCalled()
```

Then run:

```bash
bun test packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts
```

## Expected Behavior

For non-force delete, active-member validation should happen before any background task lookup or cancellation. If delete is rejected with `members still active`, no member task should be cancelled.

`force: true` should still be able to complete/cancel active member work as the explicit destructive path.

## Actual Behavior

The regression fails because `getTasksByParentSession` is called before the `members still active` guard:

```text
Expected number of calls: 0
Received number of calls: 1
```

In a real team run, this means a mistaken early `team_delete` can cancel active member background tasks and then still report that deletion was rejected.

## Doctor Output

```shell
Not applicable. This is a source-level regression reproduced directly against current origin/dev, not a local installation or provider configuration issue.
```

## Error Logs

```shell
members still active
```

## Configuration

```json
{
  "team_mode": {
    "enabled": true
  }
}
```

## Additional Context

I verified this against a clean `origin/dev` worktree at commit `6ca975e58`. The targeted regression fails on base and passes when background task cancellation is moved after the non-force active-member guard.

Relevant docs in `docs/guide/team-mode.md` describe `team_delete` as teardown only when there are no active members.

## Operating System

Linux

## OpenCode Version

Not applicable; source-level `origin/dev` issue.

dongguri-jun:
## Summary

- Prevent non-force `team_delete` from cancelling active member background tasks before it rejects the delete request.
- Preserve background-task cancellation for valid delete paths after the active-member guard passes.

## Changes

- Move team background-task lookup/cancellation after the non-force active-member guard.
- Narrow the `deleteTeam` background manager dependency to the two methods it uses.
- Add a regression test proving rejected non-force deletes do not call `getTasksByParentSession` or `cancelTask`.

## Testing

- `bun install --frozen-lockfile`
- `bun test packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts` -> 3 pass, 0 fail
- LSP diagnostics: no errors for `delete-team.ts` and `delete-team-bg-cancel.test.ts`
- `opencode-qa` harness self-check attempted; blocked by missing host `sqlite3`, while isolated XDG sandbox checks passed.

## Related Issues

Closes #5255

---
## Summary by cubic
Prevents non-force team deletes from cancelling background tasks when the delete is rejected, and keeps the correct cancel-first ordering for force deletes.

- **Bug Fixes**
  - Moved the active-member guard before background task lookup/cancel in `deleteTeam` (non-force rejects now throw early).
  - Preserved force-delete behavior: cancel team tasks first, then update member statuses.
  - Exported narrowed background manager type `DeleteTeamBackgroundManager` (`getTasksByParentSession`, `cancelTask`) and added a regression test to ensure rejected non-force deletes do not call the manager.

<sup>Written for commit c35329fceacfa1c3e494c94c286ce341e443b29a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts
+++ b/packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts
@@ -3,10 +3,25 @@
 import { afterEach, describe, expect, mock, test } from "bun:test"
 import { rm } from "node:fs/promises"
 
-import type { BackgroundManager } from "../../background-agent/manager"
+import type { BackgroundTask } from "../../background-agent/types"
+import { deleteTeam, type DeleteTeamBackgroundManager } from "./delete-team"
 import { createFixture, updateMemberStatuses } from "./shutdown-test-fixtures"
 
-const { deleteTeam } = await import("./delete-team")
+type BackgroundTaskInput = Pick<BackgroundTask, "id" | "sessionId" | "teamRunId"> & {
+  parentMessageId?: string
+}
+
+function createBackgroundTask(input: BackgroundTaskInput): BackgroundTask {
+  return {
+    ...input,
+    parentMessageId: input.parentMessageId ?? "delegate-task:test-task",
+    parentSessionId: "lead-session",
+    description: "test task",
+    prompt: "test prompt",
+    agent: "sisyphus",
+    status: "running",
+  }
+}
 
 describe("deleteTeam cancels only this team's background tasks", () => {
   const temporaryDirectories: string[] = []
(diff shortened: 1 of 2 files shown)


dongguri-jun:
Thanks, this was valid. I reproduced the failure with:

```bash
bun test packages/omo-opencode/src/features/team-mode/team-runtime/shutdown.test.ts
```

The force-delete regression observed `completed/completed` during `cancelTask`, matching your comment. I pushed follow-up commit `c35329fce` so only the non-force active-member rejection runs before background cancellation, while the force completion transition is back after cancellation.

Re-verified with:

```bash
bun test packages/omo-opencode/src/features/team-mode/team-runtime/shutdown.test.ts packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts
```

Result: 18 pass, 0 fail.

dongguri-jun:
recheck

PR #5264, opened 2026-06-13
fix(opencode-qa): preserve HOME-based opencode shims

### LINKED ISSUE #5263: opencode-qa isolated HOME breaks HOME-based opencode shims
## Prerequisites

- [x] I will write this issue in English (see the Language Policy in CONTRIBUTING.md)
- [x] I am using the current `opencode-qa` skill from this repository
- [x] I have read the skill documentation and reproduced the failure locally

## Bug Description

`opencode-qa`'s isolated sandbox rewrites `HOME`, which breaks installed `opencode` shims that resolve the real binary relative to `$HOME`.

The skill correctly isolates `XDG_DATA_HOME`, `XDG_CONFIG_HOME`, `XDG_STATE_HOME`, and `XDG_CACHE_HOME` so spawned `opencode` processes do not pollute the real DB. However, `scripts/lib/common.sh` also sets:

```bash
export HOME="$root/home"
export OPENCODE_TEST_HOME="$root/home"
```

On installations where `opencode` is a wrapper like:

```sh
#!/bin/sh
exec "$HOME/.opencode/bin/opencode" "$@"
```

or where the second-level wrapper resolves `opencode-real` under `$HOME/.opencode/bin`, the isolated `HOME` points at a temp directory that does not contain the installed binary layout. As a result, QA scripts that spawn `opencode` fail before the server/TUI can start.

## Steps to Reproduce

Given an installed `opencode` wrapper that resolves the binary through `$HOME/.opencode/bin`, run:

```bash
cd .agents/skills/opencode-qa
bash scripts/server-smoke.sh --self-test
```

The same failure also affects:

```bash
bash scripts/sse-hook-probe.sh --self-test
bash scripts/tui-smoke.sh --self-test
```

A minimal reproduction of the wrapper breakage:

```bash
tmp=$(mktemp -d -t oqa-manual.XXXXXX)
mkdir -p "$tmp/home" "$tmp/data" "$tmp/config" "$tmp/cache" "$tmp/state"
HOME="$tmp/home" \
XDG_DATA_HOME="$tmp/data" \
XDG_CONFIG_HOME="$tmp/config" \
XDG_CACHE_HOME="$tmp/cache" \
XDG_STATE_HOME="$tmp/state" \
OPENCODE_TEST_HOME="$tmp/home" \
OPENCODE_DISABLE_AUTOUPDATE=1 \
OPENCODE_DISABLE_MODELS_FETCH=1 \
opencode --version
```

## Expected Behavior

The isolated `opencode-qa` server/SSE/TUI smoke tests should be able to spawn the installed `opencode` binary while keeping runtime data isolated from the user's real DB/config/cache/state.

## Actual Behavior

The wrapper tries to execute a binary under the temporary sandbox home and fails:

```text
.../.local/bin/opencode: 2: exec: /tmp/oqa-xdg.../home/.opencode/bin/opencode: not found
FAIL: server did not become ready
```

In installations with a second-level wrapper, directly invoking the copied wrapper can similarly fail with:

```text
.../opencode: 26: exec: /tmp/oqa-manual.../home/.opencode/bin/opencode-real: not found
```

## Additional Evidence

Keeping the real `HOME` while isolating only XDG paths allows `opencode --version` to succeed.

Copying the installed `.opencode/bin` layout into the sandbox home before starting the server also allows the isolated server smoke to pass:

```text
PASS server healthy on <port>
```

So the server itself is not the root cause. The failure is caused by changing `HOME` without preserving the binary layout expected by the installed shim.

## Related Observation

`common.sh --self-check` also currently requires `sqlite3` globally:

```bash
oqa_require opencode sqlite3 curl jq tmux
```

That is a separate, lower-priority issue: missing `sqlite3` blocks the common self-check even though server/SSE smoke paths do not directly need the host `sqlite3` CLI. The `HOME` sandbox breakage is more severe because it prevents spawned-opencode QA paths from starting at all.

## Possible Fix Directions

I see a few viable approaches:

1. Resolve the real `opencode` binary path before changing `HOME`, then invoke that resolved path inside the sandbox.
2. Keep `HOME` unchanged and isolate only `XDG_*` plus `OPENCODE_TEST_HOME`, if that is sufficient to protect the real DB/config/cache/state.
3. Copy or symlink the installed `.opencode/bin` layout into the sandbox home before spawning `opencode`.
4. Split the common self-check dependencies by scenario so non-DB server/SSE smoke does not fail just because `sqlite3` is missing.

I am filing this as an issue first because the best fix depends on which isolation boundary the maintainers want `opencode-qa` to guarantee.

dongguri-jun:
## Summary

Fixes the opencode-qa isolated HOME sandbox so installed opencode wrappers that resolve through `$HOME/.opencode/bin/opencode` still work after `oqa_mk_isolated_xdg` rewrites HOME.

Adds a self-check regression that creates a fake HOME-based opencode shim, enters the isolated sandbox, and verifies the shim still reaches its real binary.

## Related Issues

- Closes #5263

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Security fix
- [ ] Infrastructure
- [ ] Documentation
- [ ] Tests only

## Affected Components

- `.agents/skills/opencode-qa/scripts/lib/common.sh`
- opencode-qa isolated XDG/HOME sandbox helper
- server and SSE smoke scripts that spawn opencode through the shared helper

### Rationale

The sandbox must keep opencode runtime data isolated while preserving the installed CLI wrapper contract. Some wrappers resolve the real binary via `$HOME/.opencode/bin/opencode`; after HOME was changed to the sandbox home, that path disappeared and spawned opencode failed before the smoke tests could start.

This change preserves only the installed `.opencode/bin` path inside the sandbox home while leaving XDG data/config/cache/state isolated.

### Alternatives Considered

- Keep the real HOME and isolate only XDG paths. Rejected because the helper explicitly promises an isolated HOME and existing checks assert it.
- Resolve and call the real opencode binary directly. Rejected because the helper should preserve normal command lookup behavior for all scripts using `opencode`.
- Copy binaries into the sandbox. Rejected because symlinking avoids stale copied binaries and keeps the wrapper chain intact.

### Failure Modes

- Missing real `$HOME/.opencode/bin`: helper no-ops, preserving behavior for installations that do not use HOME-based shims.
- Broken symlink creation: helper returns failure before spawning opencode, rather than failing later with a misleading server startup error.
- Real opencode DB pollution: XDG data/config/cache/state remain pointed at the sandbox.

### Verification Boundary

Automated/manual QA performed on this branch:

- `bash -n .agents/skills/opencode-qa/scripts/lib/common.sh` passed.
- LSP diagnostics for `.agents/skills/opencode-qa/scripts/lib/common.sh` reported no diagnostics.
- `bash .agents/skills/opencode-qa/scripts/server-smoke.sh --self-test` passed: health endpoint, OpenAPI docs, and unauthenticated session rejection.
- `bash .agents/skills/opencode-qa/scripts/sse-hook-probe.sh --self-test` passed: observed `server.connected` on SSE.
- `bash .agents/skills/opencode-qa/scripts/lib/common.sh --self-check` confirmed the new HOME-based shim regression passes, but the overall self-check still exits non-zero because `sqlite3` is not installed in this environment. That dependency failure is pre-existing and unrelated to this change.

Evidence was recorded locally under `.omo/evidence/20260614-opencode-qa-home-shim/`; that directory is ignored and intentionally not included in this PR.

### AI Boundary

The patch and PR text were AI-assisted. The behavior claims above are based on actual local command execution against installed opencode 1.17.4.

### 코드 품질
- [x] 린트/포맷 검사 통과
- [x] 타입 검사 통과 또는 해당 없음
- [ ] 기존 테스트 모두 통과
- [x] 새 기능/버그에 대한 테스트 추가

### 보안
- [x] 하드코딩된 시크릿/키 없음
- [x] 입력 검증 적용 또는 해당 없음
- [ ] 보안 관련 변경 시 리뷰어 지정

### 문서
- [x] README 또는 관련 문서 업데이트 필요 없음
- [x] CHANGELOG 기록 필요 없음

## Screenshots / Logs

Key verification output:

```text
PASS: isolated HOME preserves HOME-based opencode shim
PASS: server-smoke
PASS: SSE /event opened and delivered server.connected
```

## Etc

The local `sqlite3` binary is missing, so the common self-check still reports one pre-existing dependency failure even though the new shim regression passes.

---
## Summary by cubic
Preserves HOME-based `opencode` shims when the opencode-qa sandbox rewrites HOME, so wrappers resolving through `$HOME/.opencode/bin/opencode` keep working and smoke tests run reliably.

- **Bug Fixes**
  - Symlinks the real `$HOME/.opencode/bin` into the sandbox HOME via `oqa_preserve_home_opencode_bin`, called from `oqa_mk_isolated_xdg` (XDG data/config/cache/state stay isolated).
  - Adds a self-check that builds a fake HOME-based shim and verifies it reaches the real binary.

<sup>Written for commit 00db0265ee4bb349506c077beb39b215f89d0b43. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/skills/opencode-qa/scripts/lib/common.sh
+++ b/.agents/skills/opencode-qa/scripts/lib/common.sh
@@ -54,6 +54,16 @@ oqa_db_query() {
   opencode db "$1" --format json 2>/dev/null
 }
 
+# Preserve HOME-based opencode shims after HOME is sandboxed. Some installed
+# opencode wrappers resolve the real binary via "$HOME/.opencode/bin/opencode";
+# after oqa_mk_isolated_xdg rewrites HOME, that path must still exist.
+oqa_preserve_home_opencode_bin() {
+  local real_home="$1" sandbox_home="$2"
+  [ -d "$real_home/.opencode/bin" ] || return 0
+  mkdir -p "$sandbox_home/.opencode" || return 1
+  ln -s "$real_home/.opencode/bin" "$sandbox_home/.opencode/bin" 2>/dev/null || return 1
+}
+
 # Create an isolated XDG sandbox so a spawned opencode never touches the real
 # DB. Sets globals OQA_XDG_ROOT + OQA_PROJ and exports XDG_*; registers the
 # root for cleanup.
@@ -63,10 +73,12 @@ oqa_db_query() {
 #   oqa_mk_isolated_xdg            # good
 #   root="$OQA_XDG_ROOT"           # read the global afterwards
 oqa_mk_isolated_xdg() {
-  local root
+  local root real_home
   root="$(mktemp -d -t oqa-xdg.XXXXXX)" || return 1
+  real_home="$HOME"
   OQA_TMPDIRS+=("$root")
   mkdir -p "$root/data" "$root/config" "$root/cache" "$root/state" "$root/home" "$root/proj"
+  oqa_preserve_home_opencode_bin "$real_home" "$root/home" || return 1
   export OQA_XDG_ROOT="$root"
   export HOME="$root/home"
   export OPENCODE_TEST_HOME="$root/home"
@@ -212,6 +224,32 @@ oqa__self_check() {
     oqa_pass "isolated HOME points inside sandbox"
   else
     oqa_log "FAIL: sandbox HOME not isolated (HOME='$home' OPENCODE_TEST_HOME='$test_home' root='$isodir')"; fails=$((fails+1))
+  fi
+
+  local shim_marker shim_result
+  shim_marker="$(mktemp -t oqa-shim.XXXXXX)"
+  bash -c '
+    set -u
+    . "'"${BASH_SOURCE[0]}"'"
+    fake_home="$(mktemp -d -t oqa-fake-home.XXXXXX)"
+    mkdir -p "$fake_home/.local/bin" "$fake_home/.opencode/bin"
+    printf "%s\n" "#!/usr/bin/env bash" "exec \"\$HOME/.opencode/bin/opencode\" \"\$@\"" > "$fake_home/.local/bin/opencode"
+    printf "%s\n" "#!/usr/bin/env bash" "exec \"\$HOME/.opencode/bin/opencode-real\" \"\$@\"" > "$fake_home/.opencode/bin/opencode"
+    printf "%s\n" "#!/usr/bin/env bash" "printf fake-opencode-ok" > "$fake_home/.opencode/bin/opencode-real"
+    chmod +x "$fake_home/.local/bin/opencode" "$fake_home/.opencode/bin/opencode" "$fake_home/.opencode/bin/opencode-real"
+    HOME="$fake_home"
+    PATH="$fake_home/.local/bin:$PATH"
+    oqa_mk_isolated_xdg
+    opencode > "'"$shim_marker"'"
+    oqa_cleanup
+    rm -rf "$fake_home"
+  '
+  shim_result="$(cat "$shim_marker" 2>/dev/null)"
+  rm -f "$shim_marker"
+  if [ "$shim_result" = "fake-opencode-ok" ]; then
+    oqa_pass "isolated HOME preserves HOME-based opencode shim"
+  else
+    oqa_log "FAIL: HOME-based opencode shim returned '$shim_result'"; fails=$((fails+1))
   fi
 
   if [ "$fails" -eq 0 ]; then


dongguri-jun:
Verification note for reviewers:

- Local QA passed for the affected opencode-qa surfaces:
  - `bash -n .agents/skills/opencode-qa/scripts/lib/common.sh`
  - `bash .agents/skills/opencode-qa/scripts/server-smoke.sh --self-test`
  - `bash .agents/skills/opencode-qa/scripts/sse-hook-probe.sh --self-test`
- `common.sh --self-check` confirms the new HOME-based shim regression passes. The only remaining local self-check failure is the pre-existing missing `sqlite3` binary on my machine.
- Post-implementation review lanes passed: goal fit, QA execution, code quality, security, and context mining.

GitHub Actions note: the main `CI` run is currently `action_required`, not failed. This appears to be the normal fork PR approval gate. Maintainer approval is needed to run it.

PR #5293, opened 2026-06-14
fix(task): route plan directly and fix generic direct-subagent hang

### LINKED ISSUE #5288: Bug: task(subagent_type=plan) stays running after plan output
## Summary

`task(subagent_type="plan", run_in_background=false, ...)` can remain visible as running after the plan agent has already produced its final plan output.

This looks like a routing/polling bug in the OpenCode `delegate-task` path for plan agents.

## Environment

- Package: `oh-my-openagent`
- Version observed: `4.9.2`
- Harness: OpenCode
- Installed plugin bundle path observed locally:
  - `~/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/dist/index.js`

## Observed behavior

When using a synchronous planner task:

```ts
task(subagent_type="plan", run_in_background=false, prompt="...")
```

OpenCode shows `Plan Task` as still running in the UI/background task list even after the plan text has been generated and returned.

In practice the user sees the plan as done, but the UI keeps a red/spinner/running `Plan Task` item around.

## Expected behavior

A synchronous plan task should complete once the plan agent has emitted its final answer.

Background child tasks launched by the planner should not keep the planner wrapper itself alive after the planner has returned final text.

## Root cause found locally

From the bundled `dist/index.js` in 4.9.2:

1. `task(subagent_type="plan")` is routed through the `delegate-task` sync wrapper.
2. Plan-family agents get the `task` tool enabled:

```js
function buildSyncPromptTools2(agentToUse) {
  return {
    task: isPlanFamily(agentToUse),
    call_omo_agent: true,
    question: false,
    ...getAgentToolRestrictions(agentToUse)
  };
}
```

3. The sync poller waits while child background tasks are active:

```js
const shouldWaitForChildTasks = (currentAssistantId) => {
  if (input.hasActiveChildBackgroundTasks?.(input.sessionID)) {
    childWaitAssistantId = currentAssistantId;
    childWaitStartedAt = 0;
  }
  // ...
};
```

4. Planner instructions can launch background exploration/documentation agents. If those child tasks remain active or the child-task state is not cleared in time, the parent `Plan Task` wrapper stays running even after the planner output is done.

## Local workaround tested

I patched the local cache so `task(subagent_type="plan")` bypasses the `delegate-task` sync wrapper and routes directly through the existing `call_omo_agent` path:

- add `"plan"` to `ALLOWED_AGENTS` for `call_omo_agent`
- in `createDelegateTask(...).execute(...)`, before resolving parent context / wrapper execution:

```js
if (!delegateTaskArgs.category && isPlanAgent(delegateTaskArgs.subagent_type)) {
  const directPlanArgs = {
    description: delegateTaskArgs.description,
    prompt: delegateTaskArgs.prompt,
    subagent_type: "plan",
    run_in_background: runInBackground,
    session_id: delegateTaskArgs.task_id
  };
  return createCallOmoAgent(
    options,
    options.backgroundManager,
    options.disabledAgents,
    options.agentOverrides,
    options.userCategories,
    options.modelFallbackControllerAccessor
  ).execute(directPlanArgs, toolContext);
}
```

This makes plan calls direct instead of wrapper-mediated. Syntax check passed with:

```bash
node --check ~/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/dist/index.js
```

## Why this seems distinct from general background-agent behavior

The issue is not that background child agents exist. It is that the synchronous `delegate-task` wrapper for `plan` waits on child background task state even after the planner already produced final text. For a planning agent, spawned exploration tasks are context-gathering helpers, not necessarily blockers for completing the visible planner wrapper.

## Related issue

Possibly related but not identical: #4589 (`Prevent main-session explore/librarian launches alongside plan`). This report is specifically about `task(subagent_type="plan")` staying running because it goes through the sync task wrapper and child-task wait logic.

## Suggested fix

One of these should work:

1. Route `task(subagent_type="plan")` through a direct planner path, not `delegate-task` sync wrapper.
2. Or make `pollSyncSession()` skip `hasActiveChildBackgroundTasks` waiting for plan-family agents once terminal assistant text is present.
3. Or disallow background child-task waiting for plan-family sync wrappers and let the planner own its finality.

I think option 1 is safest because planner semantics differ from normal delegated implementation tasks.

studi929:
## Summary
Fixes #5288

This PR addresses the `task(subagent_type="plan")` hang, but runtime verification found that the root cause actually affects *all* read-only subagents (`explore`, `librarian`, `plan`) when executed via sync direct route.

Two fixes are included:
1. **Direct Routing for Plan**: Adds `plan` to `call_omo_agent` allowed list and bypasses `delegate-task`'s nested `task: true` sync wrapper for `plan` calls.
2. **Generic Hang Fix (Poller Stale Status)**: For subagents that do not launch background tasks but complete quickly (like `explore` returning immediate results), the API `session.status()` can stay stuck on `running` or `retry` longer than expected. The poller previously trusted this status and never inspected messages until timeout (5+ minutes). Added a 5-second grace period: if the status remains active but assistant text is already present in the message stream, the poller now completes the task immediately.

## Verification
- `bun test packages/omo-opencode/src/tools/delegate-task/sync-session-poller.status-fallback.test.ts` -> 2 pass
- `bun test packages/omo-opencode/src/tools/call-omo-agent/agent-restriction.test.ts packages/omo-opencode/src/tools/delegate-task/tools.test.ts --bail` -> 140 pass
- `bun run typecheck` -> pass
- `bun run build` -> pass

All local runtime probes (Plan, Explore, Quick) now complete with checkmarks and `DONE` output without hanging.

---
## Summary by cubic
Fixes sync hangs for read-only subagents by routing `plan` directly and updating the sync poller to finish when status stays active but messages contain assistant output. Plan, Explore, and Librarian now complete without long stalls.

- **Bug Fixes**
  - Plan direct routing: allow `plan` in `ALLOWED_AGENTS`, update tool schema/description, and route sync `plan` requests through `call_omo_agent` (bypasses `delegate-task` wrapper). Prevent plan-family self-delegation and pass `pluginContext` to enable this path.
  - Sync poller: add a 5s grace for active statuses; if assistant text is present in messages, complete immediately. Prevents 5+ minute hangs on stale `running/retry` statuses.

<sup>Written for commit 568d728c8d18240c74e26860dd02ad20bab02ef8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/plugin/tool-registry-core-tools.ts
+++ b/packages/omo-opencode/src/plugin/tool-registry-core-tools.ts
@@ -37,6 +37,7 @@ export function createCoreTools(args: {
     manager: managers.backgroundManager,
     client: ctx.client,
     directory: ctx.directory,
+    pluginContext: ctx,
     userCategories: pluginConfig.categories,
     agentOverrides: pluginConfig.agents,
     gitMasterConfig: pluginConfig.git_master,
--- a/packages/omo-opencode/src/tools/call-omo-agent/agent-resolver.ts
+++ b/packages/omo-opencode/src/tools/call-omo-agent/agent-resolver.ts
@@ -9,8 +9,9 @@ export function clearCallableAgentsCache(): void {
  * Resolves the set of callable agent names for call_omo_agent.
  *
  * This tool is deliberately narrower than delegate-task: it may only launch
- * the research lookup agents used by worker-style agents while they continue
- * local work. Dynamic agents and other built-ins must go through task().
+ * research lookup agents and the plan agent direct path used to avoid wrapping
+ * plan in delegate-task's sync poller. Dynamic agents and other built-ins must
+ * go through task().
  */
 export async function resolveCallableAgents(
   _client?: PluginInput["client"],
--- a/packages/omo-opencode/src/tools/call-omo-agent/agent-restriction.test.ts
+++ b/packages/omo-opencode/src/tools/call-omo-agent/agent-restriction.test.ts
@@ -69,7 +69,7 @@ describe("call_omo_agent restricted agent set", () => {
 
     //#then
     expect(result).toContain("Invalid agent type")
-    expect(result).toContain("Only explore, librarian are allowed")
+    expect(result).toContain("Only explore, librarian, plan are allowed")
     expect(launch).not.toHaveBeenCalled()
   })
 
@@ -92,8 +92,30 @@ describe("call_omo_agent restricted agent set", () => {
 
     //#then
     expect(result).toContain("Invalid agent type")
-    expect(result).toContain("Only explore, librarian are allowed")
+    expect(result).toContain("Only explore, librarian, plan are allowed")
     expect(launch).not.toHaveBeenCalled()
+  })
+
+
+  test("#when caller requests plan #then call_omo_agent launches it through the direct path", async () => {
+    //#given
+    clearCallableAgentsCache()
+    const pluginInput = createPluginInput([
+      { name: "explore", mode: "subagent" },
+      { name: "librarian", mode: "subagent" },
+      { name: "plan", mode: "subagent" },
+    ])
+    const { manager, launch } = createBackgroundManager()
+    const toolDefinition = createCallOmoAgent(pluginInput, manager)
+
+    //#when
+    await toolDefinition.execute(
+      { description: "Plan", prompt: "Plan work", subagent_type: "plan", run_in_background: true },
+      toolContext,
+    )
+
+    //#then
+    expect(launch).toHaveBeenCalledTimes(1)
   })
 
   test("#when caller requests explore or librarian #then call_omo_agent still launches them", async () => {
--- a/packages/omo-opencode/src/tools/call-omo-agent/constants.ts
+++ b/packages/omo-opencode/src/tools/call-omo-agent/constants.ts
@@ -1,9 +1,10 @@
 export const ALLOWED_AGENTS = [
   "explore",
   "librarian",
+  "plan",
 ] as const
 
-export const CALL_OMO_AGENT_DESCRIPTION = `Spawn explore/librarian agent. run_in_background REQUIRED (true=async with task_id, false=sync).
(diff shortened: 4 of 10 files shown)


MoerAI:
Thanks for the PR, @studi929. Two things before this can land. The CLA is not signed yet, so please post this exact comment on the PR:

Also the branch conflicts with `dev`, so a rebase is needed. It links #5288, so the bug is tracked, which is good. Once the CLA is in and it is rebased I will re-review. Thanks for the work.

code-yeongyu:
[sisyphus-bot] This PR addresses still-open #5288 and the fix is not yet in dev. Maintainer feedback was actionable (CLA + review notes). Marking as keep; author please rebase and address the CLA request.

MoerAI (changes requested):
Thanks for the confirmed stale-status investigation. #5288 remains open, and the current sync poller still skips message inspection indefinitely while status reports `busy`, `retry`, or `running`:
https://github.com/code-yeongyu/oh-my-openagent/issues/5288

Please rebase and reduce this to the current generic poller defect. The `plan` direct-routing commits are obsolete: current `call_omo_agent` is intentionally limited to `explore` and `librarian`, and `plan` is no longer a current named agent route.

Do not complete an actively reported session merely because any assistant text exists after five seconds. That can truncate a legitimate long-running turn with partial text or reasoning. Inspect messages after a bounded active-status grace, but require a terminal turn signal or another explicit, race-safe stale-status proof before returning.

Preserve current child-work and pending-parent-wake gates. Replace the global `Date.now` mutation and real wait with an injected clock/controlled timer. Cover legitimate active partial text, terminal text under stale active status, pending tools, retry status, child wake, and parent-wake dispatch.

Rerun current CI and isolated real OpenCode QA for sync `explore`/`librarian` completion, with evidence under `.omo/evidence/`.

PR #5201, opened 2026-06-12
fix(delegate-task): handle internal all-complete wake in sync poller

SoShymKing:
## Summary

- Fixes #5389.
- Fixes a remaining sync delegation poller timeout that still reproduces on current `dev` after #5114 and #5171.
- Confirmed on current `dev` at `f6b9ace452f04fe3fea8025c4356e6343959ee4f`.
- The remaining failing shape is: terminal assistant turn with `finish: "stop"` followed by an internal user wake containing `[ALL BACKGROUND TASKS COMPLETE]` and `<!-- OMO_INTERNAL_INITIATOR -->`.
- Current `dev` keeps polling until `poll_timeout` instead of treating the preceding terminal assistant turn as complete.

## Changes

- Updates `pollSyncSession` to recognize the internal all-complete wake only when:
  - the latest message is an internal user wake containing `[ALL BACKGROUND TASKS COMPLETE]`
  - the wake contains the internal initiator marker
  - preceding messages already form a complete user -> assistant terminal turn
  - direct child background tasks are no longer active
- Applies the same direct child-task gate to sync continuations resumed with `task_id`.
- Adds regression coverage for:
  - active-status internal all-complete wake completion
  - idle-status internal all-complete wake completion
  - waiting while child tasks remain active
  - rejecting result-ready wakes as completion
  - rejecting non-internal all-complete user messages
  - sync continuation completion through the child-task gate

## Screenshots

N/A

## Testing

```bash
bun test packages/omo-opencode/src/tools/delegate-task/sync-session-poller.test.ts packages/omo-opencode/src/tools/delegate-task/sync-continuation.test.ts
bun run typecheck:packages
git diff --check
```

## Related Issues

Fixes #5389.

This is not a duplicate of #5114 or #5171. Those fixed parent-wake liveness paths; this PR handles the remaining sync poller completion edge where an internal all-complete wake appears after a terminal assistant turn.

Related historical context: #4615, #4874, #5086.

---
## Summary by cubic
Fixes sync delegation polling so an internal “[ALL BACKGROUND TASKS COMPLETE]” wake after a terminal assistant turn completes the sync flow (including during active status) instead of timing out. Also applies the child-task gate to continuations and replaces the hanging wait with a safe timer.

- **Bug Fixes**
  - Treat an internal all-complete wake (with the internal initiator marker) after a terminal assistant turn as completion once child tasks are clear, including when session status is active and a child-task gate is provided.
[...]

PR #5294, opened 2026-06-14
fix(delegate-task): recover parent sessions after subagent errors

SoShymKing:
## Summary

- Bridges late async `session.error` events from sync `delegate-task` sessions back into the sync poller.
- Fixes model-fallback “handled” semantics so sync poller errors are recorded only when fallback did not actually recover.
- Adds regression coverage for delegate-task-only gating, sanitized error storage, continuation retry, and fallback edge cases.

## Changes

- Added `sync-session-error-store.ts` to store sanitized async session errors for eligible sync delegate-task sessions.
- Updated `plugin/event.ts` to record unhandled `session.error` events only after model fallback declines or fails.
- Registered and cleared delegate-task sync eligibility in sync task lifecycle / continuation cleanup paths.
- Updated `sync-session-poller.ts` to consume stored async errors before continuing active-status polling.
- Changed model fallback handling to return a boolean handled result and share the real result across overlapping fallback events.
- Added tests for:
  - late `session.error` reaching sync poller
  - fallback-handled errors not being recorded
  - fallback abort/prompt failures being recorded
  - sanitized error messages not leaking raw secret values
  - non-`delegate-task` sync sessions not being recorded 

## Screenshots

N/A

## Testing

```bash
bun run typecheck:packages
bun run build
bun test packages/omo-opencode/src/plugin/event.test.ts
bun test packages/omo-opencode/src/plugin/event.model-fallback.test.ts
bun test packages/omo-opencode/src/tools/delegate-task/sync-continuation.test.ts
bun test packages/omo-opencode/src/tools/delegate-task/sync-task.test.ts
```

## Related Issues

Related to #5201 and #5224, but this PR covers a separate late async session.error bridge path rather than parent-wake or deliverable-selection behavior.

---
## Summary by cubic
Bridges late async errors from sync `delegate-task` sessions into the sync poller and tightens model fallback so we only surface errors when recovery truly fails. Also dedupes overlapping fallback continuations by sharing one in‑flight Promise and cleans lifecycle state to avoid stale errors.

- **Bug Fixes**
  - Added `sync-session-error-store` to register/clear eligible `delegate-task` sync sessions and store sanitized async errors; the sync poller consumes these first (even with no assistant turn and for resumed sync tasks).
  - Model fallback now returns a boolean “handled” result and shares a single in‑flight continuation Promise per session; the handler short‑circuits to team error hooks when handled and records errors for registered `delegate-task` sync sessions only when abort or prompt dispatch fails (overlapping events included).
  - Wired registration/cleanup into sync task create/resume, `sync-continuation` start/finally, and `session.deleted` to prevent stale entries.
  - Tests added for: not recording when fallback handles; recording on abort/prompt failures and overlapping abort failures; late error bridging on resume; poller behavior with no assistant turn; sanitized storage and ignoring non‑eligible sessions; retrying sync tasks on provider model‑not‑found via fallback.

<sup>Written for commit 1e035dce65996967a71306eccacc531472055c3c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


SoShymKing:
CI failure note:

The 7 failing `packages/omo-codex/plugin/test/scaffold-plan.test.mjs` tests all have the same Windows-only root cause and appear unrelated to this PR's subagent fallback changes.

`scaffold-plan.test.mjs` dynamically imports an absolute Windows path with:

```js
await import(scriptPath)
```
On Windows CI, scriptPath is like D:\..., so Node treats d: as an unsupported ESM URL scheme and throws ERR_UNSUPPORTED_ESM_URL_SCHEME.
Expected fix is to convert the path to a file URL first:
```js
import { pathToFileURL } from "node:url";

const scriptUrl = pathToFileURL(scriptPath).href;
await import(scriptUrl);
```
This test came from the merged upstream/dev ulw-plan scaffold changes, not from the subagent fallback code.

SoShymKing:
I checked the Atlas boulder-continuation failures locally and could not reproduce them in targeted runs.

Verified locally:
- `bun test packages/omo-opencode/src/hooks/atlas/index.test.ts -t "boulder continuation"` -> 41 pass
- `bun test packages/omo-opencode/src/hooks/atlas/index.test.ts` -> 73 pass
- prompt-gate + Atlas subset -> 84 pass
- Atlas retry + Atlas index -> 82 pass

The merge-conflict resolution touched `sync-session-poller.ts`; its focused tests also passed:
- `bun test packages/omo-opencode/src/tools/delegate-task/sync-session-poller.test.ts packages/omo-opencode/src/tools/delegate-task/sync-session-poller.status-fallback.test.ts` -> 29 pass

So this does not currently look like a deterministic regression from the conflict resolution. It looks more like a CI-only/order-dependent flaky test, likely around timer/global mock state or full-suite load. A full root `bun test` rerun would be needed to capture the exact preceding test/order that triggers it.

MoerAI:
Thanks for the PR, @SoShymKing. A few things here. GitHub shows it as blocked from merging, the `test (windows-latest)` check is failing, and the diff is large, around 1010 lines across 15 files. The Windows test runner is occasionally flaky, so a re-run may help, but the size is the bigger concern for review. Could you confirm the Windows failure is not from this change, rebase if needed, and narrow the scope if parts are separable? Thanks for the work.

SoShymKing:
> Thanks for the PR, @SoShymKing. A few things here. GitHub shows it as blocked from merging, the `test (windows-latest)` check is failing, and the diff is large, around 1010 lines across 15 files. The Windows test runner is occasionally flaky, so a re-run may help, but the size is the bigger concern for review. Could you confirm the Windows failure is not from this change, rebase if needed, and narrow the scope if parts are separable? Thanks for the work.

Thanks for the review.

I rechecked the current PR state. The previous Windows failure is no longer present on the current head: `test (windows-latest)` and `codex-compatibility (windows-latest)` are both passing now, so it looks like the earlier failure was not a deterministic regression from this PR.

I also narrowed the scope. The diff is now 13 files instead of the earlier 15, and the remaining changes are limited to the `packages/omo-opencode` delegate-task / model-fallback / sync-session recovery path. The remaining size is mostly targeted regression tests for the late async `session.error` and fallback recovery cases.

I kept those tests with the implementation so the regression coverage stays tied to the behavior being fixed. If you still want this smaller, I can split it by behavior into separate implementation+test PRs rather than separating tests from code.

MoerAI (changes requested):
Thanks for the PR, @SoShymKing, and for digging into the parent-session recovery path.

Before this can land I want to flag the size: the diff is around 996 lines across 13 files, which is well beyond the narrow-`fix()` convention in `AGENTS.md` (fixes should stay focused so they can be reviewed and reverted independently). This area also overlaps the internal-message-injection and prompt-async-gate invariants that `AGENTS.md` calls out as especially sensitive, so a 13-file change is hard to review safely as one unit.

Could you split it into the minimal parent-session recovery fix plus separate follow-ups for the surrounding changes? Once the core fix is isolated I will take another look. No rush, and thanks again.

code-yeongyu:
[sisyphus-bot] This PR addresses a real gap (parent session recovery not in dev) and has active maintainer engagement. Marking as keep; author please address the size feedback and rebase against current dev.

MoerAI:
The current branch still does not address the formal scope request: this remains a 13-file, 1,300+ line change across prompt injection, fallback deduplication, lifecycle wiring, and session-error state. Please first rebase onto current `dev`, then split the minimal parent-recovery fix from the broader store/fallback/lifecycle follow-ups, or provide a concrete dependency argument showing why that split is impossible. I will re-review once the sensitive prompt-async surface is reduced and current CI is attached.

MoerAI (changes requested):
Thanks for preserving the late `session.error` to sync-poller failure case. That handoff still appears absent from current `dev`.

The previous review and current rebase/split request remain applicable:
https://github.com/code-yeongyu/oh-my-openagent/pull/5294#pullrequestreview-4543212879
https://github.com/code-yeongyu/oh-my-openagent/pull/5294#issuecomment-5226662494

Please rebase onto current `dev` and split the minimal recovery bridge from fallback deduplication and unrelated lifecycle work, or document why each dependency is inseparable. This branch is 5,127 commits behind and conflicts.

The resumed-continuation path must preserve the current poller gate contract, including both child-work and pending-parent-wake gates, rather than adding only one gate.

Replace the `setTimeout(..., 0)` and 100 ms timeout regressions with an explicit event/latch handoff subscribed before the trigger. Add a test proving stored and logged error text follows the redaction policy when `error.message` itself contains sensitive material.

Rerun current CI on the rebased head and attach a current-`dev` reproduction or isolated OpenCode QA result under `.omo/evidence/`.

PR #5299, opened 2026-06-14
fix(team-mode): avoid cancelling approved running tasks

dongguri-jun:
## Summary
- prevent non-force team_delete from cancelling team background tasks that are still pending/running after shutdown approval
- keep force=true delete behavior as the explicit task-cancelling path
- add a regression test for shutdown_approved members with an active background task

## Verification
- bun test packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts packages/omo-opencode/src/features/team-mode/team-runtime/shutdown.test.ts --bail
- bun run typecheck
- manual deleteTeam driver: shutdown_approved + running background task rejects with "members still active" and cancelTask is not called

---
## Summary by cubic
Non-force team deletion no longer cancels running or pending background tasks after shutdown approval. Force delete remains the only path that cancels tasks.

- **Bug Fixes**
  - Non-force `deleteTeam`: if any team task is `pending` or `running`, throw "members still active" and do not call `cancelTask`.
  - Force `deleteTeam`: cancel only this team's tasks using `leadSessionId` and the `team-create:<teamRunId>:` marker before deleting resources.
  - Added early validation for allowed deletable team statuses.

- **Tests**
  - Added regression test for `shutdown_approved` members with an active background task (expects rejection and no cancellations).
  - Updated tests to pass `{ force: true }` when cancellation is intended and clarified test names.

<sup>Written for commit cb5c832b49263ccb298cc10f695284bb19c08177. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5299?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts
+++ b/packages/omo-opencode/src/features/team-mode/team-runtime/delete-team-bg-cancel.test.ts
@@ -32,7 +32,7 @@ describe("deleteTeam cancels only this team's background tasks", () => {
     }))
   })
 
-  test("uses leadSessionId as the getTasksByParentSession key", async () => {
+  test("uses leadSessionId as the getTasksByParentSession key when force deleting", async () => {
     // given
     const fixture = await createFixture()
     temporaryDirectories.push(fixture.baseDir)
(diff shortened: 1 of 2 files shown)


MoerAI:
Thanks for the PR, @dongguri-jun. This is a small, tested change, but it has no linked issue. Could you add one (Fixes #N) describing the case where approved running tasks were being cancelled? A tracked bug makes the intent clear for review and gives the fix something to anchor to. Thanks for the contribution.

MoerAI:
Thanks @dongguri-jun, merging.

Issue #5258, opened 2026-06-13
[Bug]: Sisyphus/Hephaestus GPT tool mismatch: apply_patch is available but forbidden, edit/write are unavailable


cpt-sd76vg:
## Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

## Bug Description

Sisyphus and Hephaestus appear to have a mismatch between the injected file-editing guidance and the actual tools available to the agent.

When using Sisyphus with GPT-family models, the injected prompt tells the agent:

```text
Use the `edit` and `write` tools for file changes. Do not use `apply_patch` on GPT models - it is unreliable here and can hang during verification.
```

However, the actual available tool set is:

* `apply_patch` is available
* `edit` is not available
* `write` is not available

So the practical issue is that GPT agents are instructed not to use `apply_patch`, but `apply_patch` is the only native file-editing tool they can actually see.

I reproduced this with Sisyphus using `gpt-5.5`, `gpt-5.4`, and `gpt-5.4-mini` through the OpenAI ChatGPT Plus subscription route.

Because Sisyphus follows the injected guidance, it avoids `apply_patch`. But since `edit` and `write` are unavailable, normal file editing through the standard tool surface becomes impossible. In real editing tasks, the agent falls back to inefficient workarounds such as writing files through `node`, PowerShell, or other shell/script-based methods.

Hephaestus appears to have the same issue. I tested Hephaestus with `gpt-5.5`, and it showed the same GPT-family mismatch.

I also tested Sisyphus with non-GPT models, specifically `opencode-go/qwen3.7-plus` and `vercel/xiaomi/mimo-v2.5-pro`.

For non-GPT models:

* `edit` is available
* `write` is available
* `apply_patch` is not available

So file editing itself works there. However, the same GPT-specific `apply_patch` warning is still injected, even though the model is not GPT. This seems like unnecessary prompt noise.

## Steps to Reproduce

1. Start Sisyphus with a GPT-family model such as `gpt-5.5`, `gpt-5.4`, or `gpt-5.4-mini`.
[...]

PR #5316, opened 2026-06-15
fix(sisyphus): reconcile file-edit guidance to runtime model family (#5297)

### LINKED ISSUE #5297: [Bug]: Sisyphus prompt family is determined by configured model instead of runtime-selected model
## Prerequisites

## Bug Description

This is a follow-up to #5258.

Originally, the issue was that when using Sisyphus/Hephaestus with GPT-family models, the oMo-injected prompt prohibited the use of `apply_patch`.

That part has already been fixed in #5283.

After that fix, however, I ran into a different issue in my environment when using Sisyphus with non-GPT models such as Qwen3.7PLUS.

Even though `apply_patch` was not exposed as an available tool, the same prompt used for GPT-family models was still being injected, causing Sisyphus to be instructed to use `apply_patch`.

Related comments:

https://github.com/code-yeongyu/oh-my-openagent/issues/5258#issuecomment-4701139075
https://github.com/code-yeongyu/oh-my-openagent/issues/5258#issuecomment-4701141491

In those two comments, the issue was closed because it could not be reproduced in testing.

However, I have now been able to clearly reproduce and identify the cause, so I am opening this as a new issue.

### Root Cause Identified

As of v4.10.0, I found that the type of oMo-injected prompt used for Sisyphus is determined not by the model actually used at runtime, but by the model ID configured in `oh-my-openagent.jsonc`.

Specifically, this part:

```json
"agents": {
  "sisyphus": {
    "model": "openai/gpt-5.5"
  }
}
```

If the model configured in `oh-my-openagent.jsonc` is a GPT-family model, the GPT-family prompt is injected regardless of which model is actually selected and used in the OpenCode TUI.

Conversely, if the configured model is a non-GPT model, the non-GPT prompt is forcibly injected, again regardless of the model actually selected in the TUI.

This also seems to explain why the tests in #5258 passed. The current `test.ts` does not appear to read or reflect `oh-my-openagent.jsonc`; instead, it passes the model ID directly into the prompt-rendering logic. As a result, the tests verify the direct model-ID path, but they do not cover the mismatch between the configured model and the runtime-selected model.

## Steps to Reproduce

Set the Sisyphus model in `oh-my-openagent.jsonc` to a GPT-family model.

Example:

```json
"sisyphus": {
  "model": "openai/gpt-5.5"
}
```

Then, in the OpenCode TUI chat screen, switch the active model to a non-GPT model such as Qwen3.7PLUS, and instruct Sisyphus to use `edit` / `write`.

Sisyphus will object because the prompt still contains the instruction:

```text
Use apply_patch for file edits.
```

### Reverse Case

Set the Sisyphus model in `oh-my-openagent.jsonc` to a non-GPT model.

Example:

```json
"sisyphus": {
  "model": "opencode-go/qwen3.7-plus"
}
```

Then, in the OpenCode TUI chat screen, switch the active model to a GPT-family model such as GPT-5.5, and instruct Sisyphus to use `apply_patch`.

In this case, the prompt does not contain:

```text
Use apply_patch for file edits.
```

However, because the `apply_patch` tool description itself contains similar wording, the model still uses it.

So this does not immediately cause a visible failure, but the actually injected prompt is still the non-GPT version, which likely reduces model performance or causes suboptimal behavior.

### Verification of the Behavior Above

I captured the non-recited system prompt using a small local OpenCode plugin. The plugin hooks into `experimental.chat.system.transform` and writes the runtime `input` / `output` payload to a JSON file, including the full assembled `system` content passed through that hook.

I have not attached the JSON file here because it contains information specific to my environment, but I can send a masked version of the relevant parts if needed.

## Expected Behavior

Even if the Sisyphus model is defined in `oh-my-openagent.jsonc`, there are valid cases where users may want to switch models from the TUI depending on subscription limits, task type, or model availability.

Therefore, I think the model-family-specific prompt injected by oMo should be determined based on the LLM model actually used at runtime, not only based on the configured model value in `oh-my-openagent.jsonc`.

## Actual Behavior

As described in the “Root Cause Identified” section, the model-family-specific prompt injected by oMo is determined by the configured value in `oh-my-openagent.jsonc`.

I have not yet confirmed what happens when the primary configured model is unavailable and an actual fallback occurs.

## Doctor Output

I already posted the Doctor Output in #5258.

Since my environment shows several harmless known false-positive errors, I will omit the full Doctor Output here.

For reference, my test environment is:

```text
oh-my-openagent: 4.10.0
OpenCode: 1.17.7
Operating system: Windows 11
```

## Additional Context

Related issues:

* #5258
  This is the previous issue. It was closed after the GPT-family Sisyphus/Hephaestus behavior was improved, because the remaining issue could not be reproduced at that time.

By the way, Hephaestus with `apply_patch` restored played a huge role in this investigation. I really appreciate the fix.

0804sally:
## Summary
Fixes #5297. The Sisyphus model-family prompt (GPT vs non-GPT) was chosen from the model configured in `oh-my-openagent.jsonc` and baked into the agent config at registration time. When a user switches models in the TUI at runtime, the baked guidance no longer matches — e.g. a GPT-configured Sisyphus run on a non-GPT model still instructs "Use `apply_patch`", even though `apply_patch` isn't exposed.

## Fix
The `experimental.chat.system.transform` hook is the only per-request seam that knows the runtime-selected model, so reconcile the file-edit guidance there to match the runtime family instead of the configured one:
- runtime non-GPT + baked `GPT_APPLY_PATCH_GUIDANCE` → tool-agnostic `GPT_FILE_EDIT_GUIDANCE`
- runtime GPT + baked tool-agnostic guidance → `GPT_APPLY_PATCH_GUIDANCE`
- matching family / no runtime model → unchanged

## Why not change agent registration
In `builtin-agents/sisyphus-agent.ts` the resolved `model` also drives the *actual run model*, permission, and thinking config — not just the prompt family. Making the UI-selected model win there would regress the intended "explicit config model wins" behavior. Reconciling at the runtime seam also covers the configured-base case (no mismatch → no change), so no registration change is needed.

## Tests
Added `packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts`, including the config→runtime mismatch path that #5258's tests missed: a prompt actually built for GPT via `createSisyphusAgent("openai/gpt-5.5", …)`, then assembled and passed through the runtime hook with a non-GPT model, asserting `apply_patch` is no longer forced.

Verification:
- new tests: 5 pass
- `plugin/` suite: 279 pass, 0 fail
- `tsgo --noEmit -p packages/omo-opencode/tsconfig.json`: clean

---
## Summary by cubic
Align Sisyphus file-edit guidance with the runtime-selected model family so prompts don’t instruct the wrong tool after TUI model switches. Fixes #5297.

- **Bug Fixes**
  - Reconcile guidance in `experimental.chat.system.transform` using the runtime model ID.
  - Non-GPT runtime rewrites `GPT_APPLY_PATCH_GUIDANCE` to `GPT_FILE_EDIT_GUIDANCE`; GPT runtime rewrites the reverse; unchanged if matching or no runtime model.
  - Added tests for config↔runtime mismatch, no-runtime-model case, and reverse-direction helper on synthetic input; also removed em dashes in comments and renamed the reverse-case test for clarity.

<sup>Written for commit f03718daae2daad8a3a5a9750f66412f7b9757eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5316?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts
+++ b/packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, test } from "bun:test"
+
+import { createSystemTransformHandler } from "./system-transform"
+import {
+  GPT_APPLY_PATCH_GUIDANCE,
+  GPT_FILE_EDIT_GUIDANCE,
+} from "../agents/gpt-apply-patch-guard"
+import { createSisyphusAgent } from "../agents/sisyphus"
+
+const GPT_MODEL = { id: "gpt-5.5", providerID: "openai" }
+const NON_GPT_MODEL = { id: "qwen3.7-plus", providerID: "opencode-go" }
+
+function runHandler(
+  system: string[],
+  model: { id: string; providerID: string },
+): Promise<string[]> {
+  const handler = createSystemTransformHandler(undefined, undefined, {})
+  const output = { system }
+  return handler({ sessionID: "s", model }, output).then(() => output.system)
+}
+
+describe("Sisyphus runtime prompt family reconciliation (#5297)", () => {
+  test("#given GPT apply_patch guidance baked #when runtime model is non-GPT #then guidance is rewritten to be tool-agnostic", async () => {
+    // given: configured GPT model baked the apply_patch-only guidance
+    const baked = [`Step 4. ${GPT_APPLY_PATCH_GUIDANCE}`]
+
+    // when: user switched to a non-GPT model in the TUI
+    const result = await runHandler(baked, NON_GPT_MODEL)
+
+    // then: the prompt no longer forces apply_patch
+    const joined = result.join("\n")
+    expect(joined).not.toContain(GPT_APPLY_PATCH_GUIDANCE)
+    expect(joined).toContain(GPT_FILE_EDIT_GUIDANCE)
+  })
+
+  test("#given non-GPT file-edit guidance baked #when runtime model is GPT #then guidance is upgraded to apply_patch", async () => {
+    // given: configured non-GPT model baked the tool-agnostic guidance
+    const baked = [`Step 4. ${GPT_FILE_EDIT_GUIDANCE}`]
+
+    // when: user switched to a GPT model in the TUI
+    const result = await runHandler(baked, GPT_MODEL)
+
+    // then: GPT gets its apply_patch-specific guidance back
+    const joined = result.join("\n")
+    expect(joined).toContain(GPT_APPLY_PATCH_GUIDANCE)
+    expect(joined).not.toContain(GPT_FILE_EDIT_GUIDANCE)
+  })
+
+  test("#given GPT guidance baked #when runtime model is also GPT #then guidance is left unchanged", async () => {
+    const baked = [`Step 4. ${GPT_APPLY_PATCH_GUIDANCE}`]
+    const result = await runHandler(baked, GPT_MODEL)
+    expect(result.join("\n")).toContain(GPT_APPLY_PATCH_GUIDANCE)
+  })
+
+  test("#given no runtime model #when system transform runs #then guidance is untouched", async () => {
+    const baked = [`Step 4. ${GPT_APPLY_PATCH_GUIDANCE}`]
+    const handler = createSystemTransformHandler(undefined, undefined, {})
+    const output = { system: [...baked] }
+    await handler(
+      { sessionID: "s", model: undefined as unknown as { id: string; providerID: string } },
+      output,
+    )
+    expect(output.system).toEqual(baked)
+  })
+
+  test("#given a real GPT-configured Sisyphus prompt #when run on a non-GPT model #then apply_patch is not forced (config-to-runtime mismatch)", async () => {
+    // given: this is the path issue #5258 tests missed — the prompt is actually
+    // *built* for a GPT model (as it would be from oh-my-openagent.jsonc), then
+    // assembled and passed through the runtime hook with a different model.
+    const gptAgent = createSisyphusAgent("openai/gpt-5.5", [], [], [], [])
+    expect(gptAgent.prompt).toContain(GPT_APPLY_PATCH_GUIDANCE)
+
+    // when: the runtime model selected in the TUI is non-GPT
+    const result = await runHandler([gptAgent.prompt ?? ""], NON_GPT_MODEL)
+
+    // then: the final system prompt no longer instructs apply_patch usage
+    expect(result.join("\n")).not.toContain(GPT_APPLY_PATCH_GUIDANCE)
+  })
+})
--- a/packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts
+++ b/packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts
@@ -33,8 +33,12 @@ describe("Sisyphus runtime prompt family reconciliation (#5297)", () => {
     expect(joined).toContain(GPT_FILE_EDIT_GUIDANCE)
   })
 
-  test("#given non-GPT file-edit guidance baked #when runtime model is GPT #then guidance is upgraded to apply_patch", async () => {
-    // given: configured non-GPT model baked the tool-agnostic guidance
+  test("#given the tool-agnostic guidance string is present #when runtime model is GPT #then the helper rewrites it to apply_patch", async () => {
+    // NOTE: this exercises the reverse-direction helper behavior on a synthetic
+    // input. The real non-GPT Sisyphus prompt does not bake GPT_FILE_EDIT_GUIDANCE
+    // (it carries no file-edit guidance line at all), so this is not a proof of a
+    // real config-to-runtime reverse path - only that the swap is correct when
+    // such a string does appear (e.g. in the sisyphus-junior / hephaestus GPT prompts).
     const baked = [`Step 4. ${GPT_FILE_EDIT_GUIDANCE}`]
 
     // when: user switched to a GPT model in the TUI
@@ -64,7 +68,7 @@ describe("Sisyphus runtime prompt family reconciliation (#5297)", () => {
   })
 
   test("#given a real GPT-configured Sisyphus prompt #when run on a non-GPT model #then apply_patch is not forced (config-to-runtime mismatch)", async () => {
-    // given: this is the path issue #5258 tests missed — the prompt is actually
+    // given: this is the path issue #5258 tests missed. The prompt is actually
     // *built* for a GPT model (as it would be from oh-my-openagent.jsonc), then
     // assembled and passed through the runtime hook with a different model.
     const gptAgent = createSisyphusAgent("openai/gpt-5.5", [], [], [], [])
--- a/packages/omo-opencode/src/plugin/system-transform.ts
+++ b/packages/omo-opencode/src/plugin/system-transform.ts
@@ -1,10 +1,45 @@
 import type { DefaultModeConfig } from "../config/schema/default-mode"
 import {
+  GPT_APPLY_PATCH_GUIDANCE,
+  GPT_FILE_EDIT_GUIDANCE,
+} from "../agents/gpt-apply-patch-guard"
+import { isGptModel } from "../agents/types"
+import {
   getSparkShellRuntimeAwareness,
   hasSparkShellRuntimeAwareness,
 } from "../shared/sparkshell-awareness"
 
 const ULTRAWORK_MODE_TAG = "<ultrawork-mode>"
+
+/**
+ * The Sisyphus prompt body (incl. the file-edit guidance) is baked into the
+ * agent config at registration time, based on the *configured* model in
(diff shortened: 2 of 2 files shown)


0804sally:

MoerAI:
﻿Thanks for the PR, @0804sally. The main fix path looks good: `packages/omo-opencode/src/plugin/system-transform.ts` is the right runtime seam, and the real GPT-configured Sisyphus prompt test covers the #5297 path that previously slipped through.

Could you make two small follow-ups before this lands?

1. Please replace the new em dash characters in `packages/omo-opencode/src/plugin/system-transform.ts` and `packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts` with plain ASCII punctuation. This repo avoids em/en dashes in generated code and comments.
2. The reverse-case test currently uses a synthetic baked `GPT_FILE_EDIT_GUIDANCE` string. That is fine as a unit test for the helper behavior, but the real non-GPT Sisyphus prompt does not appear to bake that constant. Please either add a short comment noting that limitation or adjust the test name so it does not imply it proves the real Sisyphus reverse path.

The primary #5297 fix path looks good to me. After this small cleanup and a green full CI run, I think this is mergeable.

0804sally:
Thanks for the review, @MoerAI! Both addressed in 5ab3972:

1. Removed the em dashes in both files (ASCII punctuation now).
2. Renamed the reverse-direction test and added a note clarifying it only exercises the helper on a synthetic `GPT_FILE_EDIT_GUIDANCE` string - the real non-GPT Sisyphus prompt doesn't bake that constant, so it's not a proof of a real reverse path. The primary #5297 path stays covered by the real GPT-configured-prompt test.

MoerAI:
﻿Thanks for the quick follow-up, @0804sally. The requested cleanup is addressed in `packages/omo-opencode/src/plugin/system-transform.ts` and `packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-family.test.ts`: the em dashes are gone, and the reverse-direction test now states that it covers helper behavior on synthetic input rather than proving a real non-GPT Sisyphus reverse path.

I approved the fork CI run and checked the current head locally. The targeted system-transform tests pass, `packages/omo-opencode` typecheck passes, and LSP reports no diagnostics on the changed files.

I am not merging yet because CI is still red: `codex-compatibility (windows-latest)` failed in `packages/omo-codex/plugin/test/scaffold-plan.test.mjs` with Node ESM rejecting a Windows absolute path as protocol `d:`. That failure is outside this PR's changed files, so I am treating it as a repo CI blocker rather than a problem in your patch. Once the CI gate is green or the maintainer decides that failure is known-unrelated, this PR looks mergeable to me.

0804sally:
Heads-up @MoerAI: the unrelated `codex-compatibility (windows-latest)` failure you mentioned (the `scaffold-plan.test.mjs` ESM `Received protocol 'c:'`/`'d:'` error) is fixed in #5320 - the test now converts the script path with `pathToFileURL()` before the dynamic import. Reproduced and verified on Windows: 7/7 pass after the change. That should clear the CI gate blocking this PR.

0804sally:
Rebased onto the latest `dev` (no conflicts). The branch now includes ca9bfb5 ("test(codex): import scaffold plan via file URL"), so the `codex-compatibility (windows-latest)` failure from the old `scaffold-plan.test.mjs` raw-path import is resolved on this tree.

The main CI workflow is currently sitting at `action_required` (fork PR awaiting the "Approve and run workflows" gate), so the test/typecheck/codex-compatibility jobs haven't run on the new commit yet. Could you approve the workflow run and take a review pass when you get a chance? Should be green now that the Windows ESM fix is in the branch.

Context: the standalone fix PR #5320 turned out to be a duplicate of ca9bfb5 and is now closed.

MoerAI:
Thanks @0804sally, merging. The fix reconciles the Sisyphus file-edit guidance to the runtime model family at the system-transform.ts per-request seam, and the full CI suite (test, typecheck, build, and codex-compatibility across ubuntu, macos, and windows) passed green. The added regression test covering the config-to-runtime mismatch path is a nice touch.

cpt-sd76vg:
@0804sally

Thank you for the PR! I appreciate your work on this.
Since it has been merged, I immediately cloned commit `bd5138b` locally, built it, and tried it out. However, as far as I can observe, the situation has not changed from #5297.

The GPT / non-GPT injection prompt is determined by the model registered in `oh-my-openagent.jsonc`.
As in the issue, I tested this locally with `openai/gpt-5.5` and `opencode-go/qwen3.7-plus`.

For reference, I attached a screenshot of part of the system prompt extracted via a hook when I registered `openai/gpt-5.5` as the Sisyphus model in `oh-my-openagent.jsonc`, then changed the model to `opencode-go/qwen3.7-plus` in the TUI and started a session.

スクリーンショット 2026-06-17 1642332

The red-underlined section is the prompt that should only be injected for GPT-based models.

Sorry for the trouble, but I would appreciate it if you could take another look.

0804sally:
@cpt-sd76vg Thanks for the repro and the screenshot, that pinpointed it. You're right that #5316 didn't actually resolve #5297.

The reason: that PR rewrote only the single `apply_patch` line in the `chat.system.transform` hook, but the whole Sisyphus prompt body is selected by model family at registration from the configured model in `oh-my-openagent.jsonc`. So on a non-GPT runtime model the entire GPT-5.5 body still got injected (the section you underlined), and the test covered only that one line, so it never exercised the configured-vs-runtime mismatch. That's the gap I missed.

@MoerAI heads-up: the green test on #5316 was misleading for exactly that reason, so the merge looked safe when it wasn't. #5359 addresses it properly.

#5359 rebuilds the whole prompt for the runtime model family at the same hook, reusing the registration pipeline so overrides / prompt_append / env context are preserved, instead of patching individual lines. The new test reproduces the configured-vs-runtime mismatch end-to-end and asserts the GPT identity and `apply_patch` guidance are both gone; it fails on the previous fix.

@cpt-sd76vg mind re-testing #5359 when you get a chance? @MoerAI review welcome.

MoerAI:
Hi @0804sally and @cpt-sd76vg, thank you both, and I owe you an apology.

I merged this too quickly. CI was green and the unit tests passed, but the tests build the hook input by hand (`runHandler([gptAgent.prompt], NON_GPT_MODEL)`), so they never exercised OpenCode's real model-resolution path. That is exactly the gap our `AGENTS.md` warns about (green tests are not QA; drive real opencode), and I did not catch it before merging. That is on me, not on your work.

@cpt-sd76vg is right: #5297 is not actually fixed. Here is the root cause I verified.

The reconcile in `experimental.chat.system.transform` reads `input.model.id`, and that field is correct (the hook receives the resolved `Model`, which has `.id`). The problem is which model that is. OpenCode resolves the request model as `input.model ?? agent.model ?? currentModel(sessionID)`. For Sisyphus, omo sets `AgentConfig.model` to the configured model and, when a model is configured, suppresses the UI selection:

- `packages/omo-opencode/src/agents/builtin-agents/sisyphus-agent.ts` line 62: `uiSelectedModel: sisyphusOverride?.model !== undefined ? undefined : uiSelectedModel`
- `packages/omo-opencode/src/agents/sisyphus-agent-config.ts` line 25: the returned `AgentConfig` includes `model`

So when a user configures `agents.sisyphus.model = openai/gpt-5.5` and switches the TUI to a non-GPT model, that switch only changes `currentModel`, while `agent.model` (gpt-5.5) wins the resolution. By the time `system.transform` runs, `input.model.id` is `gpt-5.5`, `isGptModel` is true, the swap looks for `GPT_FILE_EDIT_GUIDANCE` (which is not in the baked GPT prompt), finds nothing, and the `apply_patch` guidance stays. In the normal, non-mismatch case the function is a no-op, which is why it looked harmless but does nothing for the reported scenario.

The `system.transform` seam cannot fix this, because it only ever sees the already-resolved run model, and for a configured primary agent that is the configured model, not the TUI selection. The correct fix has to be at the model-resolution / agent-registration seam: for a primary agent that should follow the UI, do not let a configured `AgentConfig.model` mask `currentModel`, and/or choose the GPT vs non-GPT edit guidance from the model that will actually run rather than baking it at registration time. The regression test should drive the real path: configured GPT Sisyphus, current/TUI model non-GPT, no per-message override, then assert the final system prompt.

What I am doing now: I have reopened #5297 and opened #5361 to revert this merge so the correct fix can land cleanly. @0804sally, your direction (reconcile at request time) is sound; it just needs to move to a seam that sees the runtime model, and I would be glad to review a follow-up. Sorry again for merging prematurely.

cpt-sd76vg:
@0804sally @MoerAI 
I tried #5359 (commit `cea1db4`) locally, and it finally behaves exactly as I had hoped!
Thank you so much for the fix. I’ll leave the detailed confirmation in #5359.

PR #5320, opened 2026-06-15
fix(omo-codex): use file URL for dynamic import in scaffold-plan test (fixes Windows CI)

0804sally:
## Summary
`codex-compatibility (windows-latest)` is red because `packages/omo-codex/plugin/test/scaffold-plan.test.mjs` imports the script under test via a raw absolute path:

```js
const scriptPath = join(root, "skills", "ulw-plan", "scripts", "scaffold-plan.mjs");
const { ... } = await import(scriptPath);
```

On Windows that path starts with a drive letter (`C:\...`, `D:\...` on the runners), and the ESM loader reads the drive letter as a URL protocol, so the dynamic import is rejected:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: ... On Windows, absolute paths must be
valid file:// URLs. Received protocol 'c:'
```

All 7 tests in the file fail for this reason. The failure is environment-specific (Windows only) and unrelated to any product code.

## Fix
Convert the path with `pathToFileURL()` before importing, so the specifier is a valid `file://` URL on every platform:

```js
import { fileURLToPath, pathToFileURL } from "node:url";
const scriptUrl = pathToFileURL(scriptPath).href;
const { ... } = await import(scriptUrl);
```

## Verification
Reproduced and fixed on a real Windows 11 machine with `node --test`:
- before: `tests 7 / pass 0 / fail 7` (all `ERR_UNSUPPORTED_ESM_URL_SCHEME`, `protocol 'c:'`)
- after:  `tests 7 / pass 7 / fail 0`

Only the test file changed; no product code is touched.

---
## Summary by cubic
Fix failing Windows CI by importing the scaffold script via a file URL instead of a raw absolute path in packages/omo-codex/plugin/test/scaffold-plan.test.mjs. The test now uses pathToFileURL() so the ESM loader accepts the specifier on Windows; only the test file changed.

<sup>Written for commit 98ed8e67c4335c879c07b118c5ddec397da106c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-codex/plugin/test/scaffold-plan.test.mjs
+++ b/packages/omo-codex/plugin/test/scaffold-plan.test.mjs
@@ -3,15 +3,18 @@ import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import test from "node:test";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 const root = dirname(dirname(fileURLToPath(import.meta.url)));
 const scriptPath = join(root, "skills", "ulw-plan", "scripts", "scaffold-plan.mjs");
+// On Windows an absolute path like "C:\..." is not a valid ESM specifier (the
+// drive letter is read as a URL protocol), so convert it to a file:// URL.
+const scriptUrl = pathToFileURL(scriptPath).href;
 const workflowPath = join(root, "skills", "ulw-plan", "references", "full-workflow.md");
 
 test("#given the scaffold script and the workflow reference #when compared #then every plan header the script emits is documented in full-workflow.md (no drift)", async () => {
 	// given
-	const { PLAN_SECTION_HEADERS } = await import(scriptPath);
+	const { PLAN_SECTION_HEADERS } = await import(scriptUrl);
 	const workflow = await readFile(workflowPath, "utf8");
 
 	// then — the script is the single source of the plan shape; the reference must document the same headers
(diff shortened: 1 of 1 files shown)


MoerAI (changes requested):
﻿Thanks for the PR, @0804sally. The actual fix in `packages/omo-codex/plugin/test/scaffold-plan.test.mjs` looks right to me: converting the dynamic import target through `pathToFileURL()` is the correct Windows ESM path handling, and I verified it locally after running the plugin skill sync step.

I cannot mark this ready yet because GitHub currently reports `mergeStateStatus: DIRTY` for this PR. Please rebase or update the branch against `dev` so the conflict is resolved and the full Codex gate from `packages/omo-codex/AGENTS.md` can run on the final mergeable tree.

After that update and a green CI run, this looks like the right fix for the Windows `scaffold-plan.test.mjs` failure that is blocking #5316.

0804sally:
Thanks for the review, @MoerAI. While rebasing onto current `dev` to clear the `DIRTY` merge state, I found that the identical fix already landed on `dev` in ca9bfb5 ("test(codex): import scaffold plan via file URL"), which was committed ~45 min before this PR's commit but onto a newer `dev` than the one this branch was based on (db9b310). So this branch was effectively a duplicate of a fix that was already merged.

After the rebase, the only remaining diff vs `dev` is a 2-line explanatory comment — the functional change is already present upstream. Since the Windows `scaffold-plan.test.mjs` failure is now resolved on `dev`, I'm closing this PR as redundant rather than re-running the full Codex gate for a comment-only change. Happy to open a tiny follow-up just for the comment if you think it's worth keeping.

PR #5328, opened 2026-06-15
fix(keyword-detector): skip mode injection on internal-initiator wake messages

### LINKED ISSUE #5326: [bug] keyword-detector injects [analyze-mode] on internal wake messages during background task completion
[bug] keyword-detector injects [analyze-mode] on internal wake messages during background task completion

## Symptom

When background tasks complete, OMO sends a wake message to the parent session (marked with \\). The keyword-detector hook processes this internal message and injects \[analyze-mode]\, causing the agent to re-run context gathering instead of immediately collecting results and continuing.

**What happens:**
\\\
<system-reminder>
[BACKGROUND TASK COMPLETED]
...
</system-reminder>

[analyze-mode]              ← ❌ incorrectly injected
ANALYSIS MODE. Gather context before diving deep...
\\\

**What should happen:** The agent should skip \[analyze-mode]\ and immediately call \ackground_output()\ to collect results.

## Root Cause

\src/hooks/keyword-detector/hook.ts\ — the \chat.message\ handler does not check whether the message is internal before running mode directive injection.

The \hasInternalInitiatorMarker()\ function already exists in \src/shared/internal-initiator-marker.ts\ but is not used in the keyword detector path.

## Suggested Fix

In \src/hooks/keyword-detector/hook.ts\, add an early return in the \chat.message\ handler:

\\\	ypescript
import { hasInternalInitiatorMarker } from '../../shared/internal-initiator-marker'

// At the top of the chat.message handler:
const textParts = (output.parts ?? []).filter(
  (p: any) => p.type === 'text' && typeof p.text === 'string'
)
if (textParts.some((p: any) => hasInternalInitiatorMarker(p.text))) {
  return output  // Internal wake message — skip mode injection
}
\\\

## Impact

Every background task completion causes the agent to waste ~1 turn re-analyzing context instead of continuing work. With 8 parallel extraction tasks, this means 8 unnecessary context-gathering cycles.

LilMGenius:
## Summary

`keyword-detector`'s `chat.message` handler injected mode prompts (e.g. `[analyze-mode]`, ultrawork) into OMO-internal wake messages emitted on background-task completion. Those wakes are tagged with `<!-- OMO_INTERNAL_INITIATOR -->`, but the existing guard `isSyntheticOrInternalOnlyTextParts` only skips when **every** text part is synthetic/internal. A wake that mixes a real-looking content part with a separate marker part slips through, so the agent re-runs mode setup instead of immediately collecting results and continuing.

## Fix

Add a second early return in the `chat.message` handler: skip keyword detection when **any** text part carries the internal-initiator marker, reusing the existing `hasInternalInitiatorMarker` helper.

```ts
if ((output.parts ?? []).some((part) => isTextPartLike(part) && hasInternalInitiatorMarker(part.text))) {
  log(`[keyword-detector] Skipping internal-initiator wake message`, { sessionID: input.sessionID })
  return
}
```

This matches the approach suggested in the issue, using the typed `isTextPartLike` guard from the shared barrel instead of `any` casts.

## Tests

- Added a regression test: an internal wake whose parts mix a marker part with a real-looking ultrawork part must receive no mode injection.
- Verified it **fails without** the guard (the detector injects the full `<ultrawork-mode>` prompt into the user part) and **passes with** it.
- `bun test packages/omo-opencode/src/hooks/keyword-detector/index.test.ts` -> 48 pass / 0 fail.
- `tsgo --noEmit -p packages/omo-opencode/tsconfig.json` -> 0 errors.

Fixes #5326

---
## Summary by cubic
Prevent `keyword-detector` from injecting mode prompts into OMO internal wake messages by skipping detection when any part contains the internal-initiator marker, fixing #5326. This stops wasted turns where the agent re-enters mode setup and lets it continue processing results.

- **Bug Fixes**
  - Added an early return in `chat.message` to skip when any text part has `<!-- OMO_INTERNAL_INITIATOR -->` using `hasInternalInitiatorMarker` with `isTextPartLike`.
  - Ensures mixed internal wakes (marker + real-looking content) are left unchanged (no `[analyze-mode]`/ultrawork injection).
  - Added a regression test for the mixed-parts case.

<sup>Written for commit 5cfd66054831a4eb84004263339a3dc8202eb8bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/keyword-detector/hook.ts
+++ b/packages/omo-opencode/src/hooks/keyword-detector/hook.ts
@@ -8,8 +8,10 @@ import {
 } from "../../features/claude-code-session-state"
 import type { ContextCollector } from "../../features/context-injector"
 import {
+  hasInternalInitiatorMarker,
   isRealUserTextPart,
   isSyntheticOrInternalOnlyTextParts,
+  isTextPartLike,
   log,
 } from "../../shared"
 import {
@@ -69,6 +71,11 @@ export function createKeywordDetectorHook(
     ): Promise<void> => {
       if (isSyntheticOrInternalOnlyTextParts(output.parts)) {
         log(`[keyword-detector] Skipping synthetic/internal text message`, { sessionID: input.sessionID })
+        return
+      }
+
+      if ((output.parts ?? []).some((part) => isTextPartLike(part) && hasInternalInitiatorMarker(part.text))) {
+        log(`[keyword-detector] Skipping internal-initiator wake message`, { sessionID: input.sessionID })
         return
       }
 
--- a/packages/omo-opencode/src/hooks/keyword-detector/index.test.ts
+++ b/packages/omo-opencode/src/hooks/keyword-detector/index.test.ts
@@ -86,6 +86,28 @@ describe("keyword-detector message transform", () => {
     expect(text).toContain("YOU MUST LEVERAGE ALL AVAILABLE AGENTS")
   })
 
+  test("should skip keyword injection when any part carries the internal-initiator marker", async () => {
+    // given - an OMO-internal wake whose parts mix a marker part with a real-looking user part
+    const collector = new ContextCollector()
+    const sessionID = "internal-wake-mixed-parts"
+    getMainSessionSpy = spyOn(sessionState, "getMainSessionID").mockReturnValue(sessionID)
+    const hook = createKeywordDetectorHook(createMockPluginInput(), collector)
+    const output = {
+      message: {} as Record<string, unknown>,
+      parts: [
+        { type: "text", text: OMO_INTERNAL_INITIATOR_MARKER },
+        { type: "text", text: "ultrawork fix the flaky suite" },
+      ],
+    }
+
+    // when - keyword detection runs on the mixed internal wake
+    await hook["chat.message"]({ sessionID }, output)
+
+    // then - no mode prompt is injected and both parts stay untouched
+    expect(output.parts[0].text).toBe(OMO_INTERNAL_INITIATOR_MARKER)
+    expect(output.parts[1].text).toBe("ultrawork fix the flaky suite")
+  })
+
   test("should leave search wording as plain user text", async () => {
     // given - mock getMainSessionID to return our session (isolate from global state)
     const collector = new ContextCollector()


fdnic:
## ✅ Verified — Logic confirmed correct

  I independently reproduced the fix and verified the core logic with an isolated test suite.

  ### Test Results

  ```
  10 passed / 0 failed / 10 total
  ```

  ### What was verified

  | Scenario | Old Guard | New Guard | Result |
  |---|---|---|---|
  | Mixed parts: marker + "ultrawork fix the flaky suite" | ❌ bypassed (BUG) | ✅ intercepted | Fixed |
  | Marker at position 0, 1, or middle of parts | ❌ mixed | ✅ intercepted | Robust |
  | Empty / null / undefined parts | — | ✅ no crash | Safe |
  | Normal user message "ultrawork do something" | ✅ passes through | ✅ passes through | No regression |
  | Pure synthetic/internal message | ✅ intercepted (earlier guard) | n/a | No regression |

  ### Root cause

  The existing guard `isSyntheticOrInternalOnlyTextParts` uses `.every()` — it only skips when **all** text parts are synthetic/internal.
  Internal wake messages mix a marker part with real-looking user parts, so they bypassed it.

  The fix adds a `.some()` check — skip when **any** part carries the internal initiator marker. Covers the mixed-parts case without affecting
  normal messages.

  ### Code change

  +29 lines / -0 lines across 2 files. Clean, minimal, uses existing utilities (`hasInternalInitiatorMarker`, `isTextPartLike`).

  LGTM 👍

MoerAI:
Thanks for the PR, @LilMGenius, this is a clean and well-tested fix.

This PR is linked to issue #5326, which is already closed as completed. I could not confirm the exact resolution path on `dev` (the guard log line from your patch is not present there), so I cannot safely merge a second fix for an issue the maintainer already marked done without re-confirming it still reproduces.

Could you check whether `keyword-detector` still injects mode prompts on internal-initiator wakes against current `dev`? If it does, please reopen #5326 (or open a fresh issue referencing it) with the current behavior and I will take another look. If not, no further action needed.

Closing for now per the linked-issue rule, happy to reopen once it is confirmed against current `dev`.

PR #5329, opened 2026-06-15
fix(boulder-state): guard getBoulderWorks against null entries in works object

clousky2020:
## Problem

When running `/start-work`, `getWorkResumeOptions` crashes with:

```
TypeError: null is not an object (evaluating 'work.status')
```

The root cause is that `boulder.json` can contain corrupted data where a work entry under the `works` object has value `null` (e.g., an incomplete work with `status` set to `null`). `Object.values(state.works)` passes these null entries through without filtering, and downstream callers that access `work.status` crash.

## Fix

Add a type-narrowing `filter` in `getBoulderWorks` to exclude null/undefined entries at the source. This single-point guard protects all downstream callers:

- `getWorkResumeOptions`
- `getActiveWorks`
- `getWorkById`
- `getWorkByPlanName`
- `getWorkForSession`

No other callers need individual null checks.

## Change

`packages/boulder-state/src/storage/read-state.ts` — 1 line changed

```diff
-    return Object.values(state.works)
+    return Object.values(state.works).filter((work): work is BoulderWorkState => work != null)
```

---
## Summary by cubic
Prevents crashes in /start-work by filtering null entries from the works map. Adds a single guard in getBoulderWorks so corrupted boulder.json values don’t break callers.

- **Bug Fixes**
  - Filter out null/undefined entries in getBoulderWorks using a type guard.
  - Protects getWorkResumeOptions, getActiveWorks, getWorkById, getWorkByPlanName, and getWorkForSession.

<sup>Written for commit 641df86ad8ac7cedb78865c59bcf539e6a255b3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for the PR, @clousky2020. This is a tiny, focused guard, but GitHub currently shows it as blocked from merging (a required check or review is outstanding), and I do not see a regression test for the null-entry guard. Could you add a small test that covers a null entry in the works object (or note why one is not feasible)? Once the block clears and a test is in I will take another look. Thanks for the contribution.

MoerAI:
Thanks for the fix. Merged. The null filter in getBoulderWorks (Object.values(state.works).filter(work => work != null)) correctly prevents downstream crashes when the works map contains a null entry, and required CI passed green across all checks after a branch refresh.

PR #5338, opened 2026-06-16
fix: pass displayName overrides to getAgentListDisplayName in remapCommandAgentFields

### LINKED ISSUE #5337: [Bug]: remapCommandAgentFields ignores displayName overrides causing 'agent not found' for commands
## Description

When a user sets a custom `displayName` for an agent in `oh-my-openagent.json`, any command that references that agent by config key will fail to find the agent at runtime.

## Root Cause

In `packages/omo-opencode/src/plugin-handlers/command-config-handler.ts`, the `remapCommandAgentFields` function calls `getAgentListDisplayName` **without passing agent overrides**:

```typescript
cmd.agent = getAgentListDisplayName(getAgentConfigKey(cmd.agent));
// Should be:
cmd.agent = getAgentListDisplayName(getAgentConfigKey(cmd.agent), overrides);
```

This function is called from `applyCommandConfig` at line 125, which has access to `params.pluginConfig.agents`. However, `remapCommandAgentFields` doesn't accept an overrides parameter, so the agent display names in command configs are always resolved to the hardcoded values from `AGENT_DISPLAY_NAMES`.

## Why It Breaks

When OpenCode processes a command (e.g. `/start-work`), it looks up the command's agent field to determine which agent to invoke. If the command's agent is resolved to the hardcoded display name (e.g. `"Sisyphus - ultraworker"`) but the agent was registered with a custom display name (e.g. `"Sisyphus - Use for most work"`), the agent lookup fails.

## Example

With this config in `oh-my-openagent.json`:
```json
{
  "agents": {
    "sisyphus": {
      "displayName": "Sisyphus - My Custom Name"
    }
  }
}
```

Commands that reference agent `"sisyphus"` get remapped to `"Sisyphus - ultraworker"` instead of `"Sisyphus - My Custom Name"`, causing `agent not found`.

## Suggested Fix

1. Add an `overrides` parameter to `remapCommandAgentFields`
2. Pass it to `getAgentListDisplayName`
3. Update the call site in `applyCommandConfig` to pass `params.pluginConfig.agents`

## Related

- #5335 (same root cause: `applyDefaultAgent` not passing overrides to `getAgentDisplayName`)
- #5336 (PR fixing the `applyDefaultAgent` case)

clayleopardlabs:
## Summary

`remapCommandAgentFields` now accepts an `overrides` parameter and passes it to `getAgentListDisplayName`, so commands that reference agents by config key resolve to the correct custom display name.

## Problem

`remapCommandAgentFields` (in `command-config-handler.ts`) called `getAgentListDisplayName` without the `overrides` argument. When a user set a custom `displayName` for an agent in `oh-my-openagent.json`, commands referencing that agent by config key would get remapped to the hardcoded display name, causing `agent not found` at runtime.

## Fix

Three changes in `command-config-handler.ts`:

1. Added `overrides` parameter to `remapCommandAgentFields` function signature
2. Passed `overrides` to `getAgentListDisplayName` call
3. Updated the call site in `applyCommandConfig` to pass `params.pluginConfig.agents`

No behavior change when no displayName overrides are configured.

## Related

Fixes #5337

---
## Summary by cubic
Commands now respect custom agent display names by forwarding overrides from `oh-my-openagent.json` through `remapCommandAgentFields` to `getAgentListDisplayName`, preventing "agent not found" errors. Fixes #5337.

<sup>Written for commit b730639f95d52c14f718b523aa76ba4c71a41ccc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5338?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clayleopardlabs:

MoerAI:
Thanks for the PR, @clayleopardlabs. This is a small fix linked to #5337. GitHub currently shows it as blocked from merging (a required check or review is outstanding), and I do not see a regression test for the displayName override path. Could you add a test that covers the override reaching `getAgentListDisplayName` (or note why one is not feasible)? Once the block clears and a test is in I will take another look. Thanks for the contribution.

MoerAI:
Thanks for the fix. Merged. Threading the agents overrides into getAgentListDisplayName through remapCommandAgentFields makes custom command agent display names respect user displayName overrides, and all required CI passed green.

PR #5359, opened 2026-06-17
fix(sisyphus): determine prompt family by runtime model, not configured model (#5297)

0804sally:
Proper replacement for the reverted #5316 (reverted on dev as `Revert "Merge pull request #5316"`), resolving #5297. The issue author verified the runtime behavior in the comments below.

## Problem

The Sisyphus prompt body is model-family-specific and chosen once at registration from the *configured* model in `oh-my-openagent.jsonc`. When the user switches to a different family in the TUI, the entire baked body is the wrong family for the runtime model: a GPT-configured agent run on a non-GPT model still carries the whole GPT-5.5 body, not just the one `apply_patch` line.

#5316 rewrote only the single `GPT_APPLY_PATCH_GUIDANCE` string in the `chat.system.transform` hook, and its test only asserted that one line, so the rest of the GPT-only body kept leaking through (matching the screenshot in the review comment on #5316). It was reverted on dev.

## Fix

Rebuild the whole body at the only per-request seam that knows the runtime model (`experimental.chat.system.transform`):

- Capture the registration pipeline as a rebuild closure (preserving overrides / `prompt_append` / env context).
- In the hook, rebuild the prompt for the runtime model family and swap it in when it differs from the configured family. No per-line string patching, so it cannot leak family-specific lines the way the reverted approach did.
- `resolveSisyphusPromptFamily()` is extracted as the single source of truth shared by the factory and the reconciler, so the two can never drift.
- Substring replacement, robust to core concatenating the agent prompt with other system text.

## Tests

New tests reproduce the configured-vs-runtime mismatch end-to-end (a GPT-configured body run on a non-GPT model) and assert the GPT identity (`based on GPT-5.5`) and the `apply_patch` guidance are both gone, and that the result matches what registration would bake for the runtime model.

### DIFF
--- a/packages/omo-opencode/src/agents/builtin-agents/sisyphus-agent.ts
+++ b/packages/omo-opencode/src/agents/builtin-agents/sisyphus-agent.ts
@@ -9,6 +9,7 @@ import { applyOverrides } from "./agent-overrides"
 import { applyModelResolution, getFirstFallbackModel } from "./model-resolution"
 import { createSisyphusAgent } from "../sisyphus"
 import { applyFrontierToolSchemaPermission } from "../frontier-tool-schema-guard"
+import { setSisyphusRuntimePromptContext } from "../sisyphus-runtime-prompt-reconciler"
 
 export function maybeCreateSisyphusConfig(input: {
   disabledAgents: string[]
@@ -104,6 +105,27 @@ export function maybeCreateSisyphusConfig(input: {
 
   sisyphusConfig = applyEnvironmentContext(sisyphusConfig, directory, {
     disableOmoEnv,
+  })
+
+  // The body above is baked from the *configured* model. If the user switches to
+  // a different model family in the TUI, the system-transform hook rebuilds the
+  // prompt for the runtime model using this captured pipeline (issue #5297/#5316).
+  setSisyphusRuntimePromptContext({
+    configuredModel: sisyphusModel,
+    bakedPrompt: sisyphusConfig.prompt ?? "",
+    rebuildPromptForModel: (runtimeModel: string): string => {
+      let rebuilt = createSisyphusAgent(
+        runtimeModel,
+        availableAgents,
+        undefined,
+        availableSkills,
+        availableCategories,
+        useTaskSystem
+      )
+      rebuilt = applyOverrides(rebuilt, sisyphusOverride, mergedCategories, directory)
+      rebuilt = applyEnvironmentContext(rebuilt, directory, { disableOmoEnv })
+      return rebuilt.prompt ?? ""
+    },
   })
 
   return sisyphusConfig
--- a/packages/omo-opencode/src/agents/sisyphus-agent-factory.ts
+++ b/packages/omo-opencode/src/agents/sisyphus-agent-factory.ts
@@ -34,6 +34,36 @@ import {
 
 const MODE: AgentMode = "primary";
 
+/**
+ * Identifies which prompt body `createSisyphusAgent` bakes for a given model.
+ * The whole Sisyphus prompt is model-family-specific and selected here, so this
+ * is the single source of truth shared with the runtime reconciler: when the TUI
+ * runtime model resolves to a different family than the configured one, the baked
+ * body is the wrong family and must be rebuilt (issue #5297/#5316).
+ */
+export type SisyphusPromptFamily =
+  | "kimi-k2-7"
+  | "kimi-k2-6"
+  | "gpt-5-5"
+  | "gpt-5-4"
+  | "claude-fable-5"
+  | "claude-opus-4-8"
+  | "claude-opus-4-7"
+  | "glm-5-2"
+  | "fallback";
+
+export function resolveSisyphusPromptFamily(model: string): SisyphusPromptFamily {
+  if (isKimiK27Model(model)) return "kimi-k2-7";
+  if (isKimiK2Model(model)) return "kimi-k2-6";
+  if (isGpt5_5Model(model)) return "gpt-5-5";
+  if (isGptNativeSisyphusModel(model)) return "gpt-5-4";
+  if (isClaudeFable5Model(model)) return "claude-fable-5";
+  if (isClaudeOpus48Model(model)) return "claude-opus-4-8";
+  if (isClaudeOpus47Model(model)) return "claude-opus-4-7";
+  if (isGlmModel(model)) return "glm-5-2";
+  return "fallback";
+}
+
 export function createSisyphusAgent(
   model: string,
   availableAgents?: AvailableAgent[],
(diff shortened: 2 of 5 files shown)


cpt-sd76vg:
@0804sally 
I’m the author of issue #5297.

I built and tested #5359 (commit `cea1db4`) locally. I confirmed that, regardless of the model configured in `oh-my-openagent.jsonc`, the GPT / non-GPT prompts are now injected correctly based on the model selected in the TUI.

I believe the issue reported in #5297 has been resolved. Thank you very much for the fix!

---

P.S. As a small note for reference, both GPT and non-GPT models still receive the following instructions in the bash tool description:

* Edit files: Use Edit (NOT Set-Content)
* Write files: Use Write (NOT Write-Output/Write-Host)

I understand these are intended to discourage the use of PowerShell commands. Because of that, there is still a slight mismatch when using GPT-based models. However, the injected `Use apply_patch...` prompt appears to prevent any significant impact.

I think this should be considered separately from issue #5297, so I don’t think PR #5359 needs to be changed for this.

0804sally:
@cpt-sd76vg Thanks for taking the time to build and verify it locally, I really appreciate the confirmation that #5297 is resolved.

Good catch on the bash tool-description note as well. Agreed it's a separate concern from #5297 and out of scope for this PR; thanks for flagging it so the `apply_patch` vs `Edit`/`Write` signals can be reconciled for GPT models on their own track.

0804sally:
@MoerAI when you have a moment, this is the proper replacement for the reverted #5316.

- It rebuilds the whole Sisyphus prompt for the runtime model family at the `chat.system.transform` seam, instead of patching the single `apply_patch` line, so it can't leak family-specific lines the way the reverted approach did.
- Rebased onto the latest dev (after the #5316 revert); the conflict in `system-transform.ts` is resolved and the old line-level swap is intentionally not resurrected.
- @cpt-sd76vg (the #5297 reporter) built and verified the runtime behavior locally, and CI is green across test / typecheck / build / codex-compatibility on all platforms.

Review and merge welcome whenever it fits your schedule.

cpt-sd76vg:
I also built and tested commit `1d0715f` locally.

As with commit `cea1db4`, I confirmed that the oMo-injected prompt is correctly determined and injected based on the model selected in the TUI.

Note: Since the rebase target, `v4.11.0`, includes many changes, I have not checked minor behavior differences or compatibility issues unrelated to issue #5297.

MoerAI:
Thanks for the PR, @0804sally. This targets #5297, but #5316 ("reconcile file-edit guidance to runtime model family") already merged for that same issue, so there is likely overlap. The change here is also around 350 lines across 5 files.

Could you confirm how this differs from what #5316 already landed, rebase on current `dev`, and narrow the diff to just the remaining gap in #5297? If #5297 is fully resolved now, no further action is needed. Thanks for the work.

0804sally:
Thanks for taking the time to review this, @MoerAI — much appreciated, and apologies for any confusion around the overlap with #5316.

Just to clarify the current state, since I think the timeline may not be obvious from the PR alone:

- **#5316 is no longer on `dev`.** It was merged briefly, but then reverted via **#5361** (revert `241828b5`, merged `61e56e8f`) before this PR was opened. So there is no longer any overlap to reconcile — **this PR is the replacement for that reverted change**, not a duplicate of something currently on `dev`.
- **Why a replacement rather than a re-land:** #5316 patched only the single `GPT_APPLY_PATCH_GUIDANCE` line in the system-transform hook, so the rest of the GPT-only Sisyphus body still leaked through on non-GPT runtime models (its test only asserted that one string, which is why it went green while the defect persisted). This PR instead rebuilds the **whole** prompt body for the runtime model family at the `chat.system.transform` seam — `resolveSisyphusPromptFamily()` is the single source of truth shared by the factory and the runtime reconciler, so the configured-vs-runtime family can't drift.
- **Rebased on current `dev`** just now (head `e1321164d`); it had fallen ~14 commits behind, so it is freshly up to date and `MERGEABLE` with no conflicts.
- **Scope/size:** the diff is **+281 / −69 across 5 files**, with the bulk being the reconciler module and its tests. #5297 reporter @cpt-sd76vg has built and verified the runtime behavior locally (on both `cea1db4` and the rebased commit), and CI is green across test / typecheck / build / codex-compatibility.

Happy to narrow further if any specific part looks larger than it needs to be — just let me know. Thanks again for the careful look.

MoerAI (changes requested):
Thanks for the PR, @0804sally, and thanks to @cpt-sd76vg for building and verifying it locally against #5297. The runtime-model-family rebuild approach reads well, and I appreciate that this is the clean replacement for the reverted #5316.

One blocker before this can land: GitHub currently marks the PR as conflicting with `dev`, so it needs a rebase onto current `dev` before it can merge. While you are in there, the diff is on the larger side for a `fix()` (around 350 lines across 5 files, against the narrow-change convention in `AGENTS.md`); if any of that is incidental, trimming it helps the review, but the rebase is the hard part.

Once it is rebased clean I will take another look, including the #5316 overlap question you clarified. No rush, and thanks again for the careful write-up.

0804sally:
@
@MoerAI Rebased onto current `dev` — the PR now shows **`MERGEABLE`** (head `77b012c0c`).

**The only conflict** was in `sisyphus-agent-factory.ts`: while this branch was out, `dev` added the **GLM (`glm-5-2`)** family as another post-switch `if`-branch, whereas this PR had already refactored that dispatch into a single `switch` keyed on `resolveSisyphusPromptFamily()`. I folded GLM into that single source of truth rather than reviving the parallel `if`-chain:

- added `"glm-5-2"` to the `SisyphusPromptFamily` union,
- added `if (isGlmModel(model)) return "glm-5-2";` to `resolveSisyphusPromptFamily()` (same precedence as `dev`: after `claude-opus-4-7`, before `fallback`),
- added the `case "glm-5-2"` arm (`buildGlmSisyphusAgentConfig` + `buildGlm52SisyphusPrompt`).

**On behavioral risk — the dispatch is identical to `dev`.** The factory change is purely structural: `dev`s top-to-bottom `if`-chain became `switch (resolveSisyphusPromptFamily(model))`, and the resolver evaluates the **same predicates in the same order** (`kimi-k2-7 → kimi-k2-6 → gpt-5-5 → gpt-5-4 → fable-5 → opus-4-8 → opus-4-7 → glm-5-2 → fallback`), same config/prompt builder per arm, same `isGptModel ? gpt : claude` fallback. First-match semantics are preserved, so for any given model the baked output is byte-for-byte what `dev` produced (`tsgo` exhaustiveness over the union also guarantees no arm is dropped).

**The only genuinely new behavior is the runtime reconciler, and it is strictly safer:** when the runtime model resolves to the **same** family as the configured one it early-returns and does nothing (identical to `dev` for the common case); it only acts when families **differ** — the `#5297` defect where `dev` leaks the wrong-family body — rebuilding through the *same* factory pipeline and swapping via a substring match that no-ops if the baked body is not found. Net: where `dev` was correct it is unchanged, where `dev` leaked it is corrected. As a side benefit, folding GLM into the resolver means GLM now also gets runtime reconciliation, which the pre-existing side-branch did not.

Verified locally: `tsgo --noEmit` clean, and the factory / reconciler / apply-patch-guidance / agent-identity / tool-restriction suites are all green (45 tests, 0 fail). Diff is +289/−77 across the same 5 files, the bulk still being the reconciler module and its tests.
@

MoerAI (approved):
Thanks @0804sally, and thanks @cpt-sd76vg for verifying #5297 locally. The rebase resolved the merge conflict I flagged earlier, and this is the clean replacement for the reverted #5316. Approving and merging.

PR #5332, opened 2026-06-15
fix(team-core): expand leading ~ in team_mode.base_dir

manascb1344:
## Summary

Fixes #5331.

`resolveBaseDir()` in `packages/team-core/src/team-registry/paths.ts` returned `config.base_dir` verbatim. When `team_mode.base_dir` is set to the literal string `"~/.omo"` (a natural value, since the docs describe `~/.omo` as the default location), the raw string reached `mkdir("~/.omo", { recursive: true })` and Node/Bun treated `~` as a literal directory name. Result: a real directory named `~` was created in the cwd, and `$HOME/.omo` was never written.

This pollutes every directory opencode is launched in with a junk `~` folder.

## Root cause

3-link chain:

1. Config: `~/.config/opencode/oh-my-openagent.jsonc` sets `"team_mode": { "enabled": true, "base_dir": "~/.omo" }`.
2. Code: `resolveBaseDir(config) = config.base_dir ?? path.join(homedir(), ".omo")`. Because `base_dir` is set, the `homedir()` fallback is never reached, so the literal `"~/.omo"` is returned.
3. Exec: `ensureBaseDirs("~/.omo")` calls `mkdir("~/.omo", { recursive: true })`. With no shell in the path there is no tilde expansion, so a literal `~` directory is created.

## Fix

Expand a leading `~` via `homedir()` before returning from `resolveBaseDir()`. Covers both the config-supplied and the default path:

```ts
export function resolveBaseDir(config: TeamModeConfig): string {
  const resolved = config.base_dir ?? path.join(homedir(), ".omo")
  return resolved.startsWith("~") ? path.join(homedir(), resolved.slice(1)) : resolved
}
```

## Verification

- strace before fix: `mkdir("~/.omo") = -1 ENOENT` then `mkdir("~") = 0` then `mkdir("~/.omo/teams"|"runtime"|"worktrees") = 0`.
- After the fix, `opencode run "hi"` in a fresh dir produces no literal `~`, and `$HOME/.omo/{runtime,teams,worktrees}` is created in the correct location.
- `--pure` run (plugins disabled) already produced an empty dir, confirming the culprit is the plugin and not opencode core.

## Tests

Added two regression tests in `packages/team-core/src/team-registry/paths.test.ts` (given/when/then):

- `resolveBaseDir` expands `base_dir: "~/.omo"` to `path.join(homedir(), ".omo")`.
- `resolveBaseDir` expands an arbitrary `~/some/custom/team-root`.

Suite results:

[...]

PR #5369, opened 2026-06-17
fix(team-mode): expand leading tilde in resolveBaseDir (#5331)

### LINKED ISSUE #5331: [Bug]: Literal `~` directory created in cwd when `team_mode.base_dir` is set to "~/.omo" — tilde never expanded
## Bug Description

When `team_mode` is enabled and `base_dir` is set to the string `"~/.omo"` (literal tilde), oh-my-openagent creates a **real directory named `~`** (tilde) in the current working directory instead of expanding `~` to the user's home directory. Running any opencode command that loads the plugin (`opencode run "hi"`, the TUI, `opencode serve`) creates a stray `~` folder containing `.omo/{runtime,teams,worktrees}` in whatever directory opencode is invoked from.

This pollutes every working directory opencode is launched in with a junk `~` folder, and the actual `~/.omo` (i.e. `$HOME/.omo`) is never created — so team-mode state silently lands in the wrong place.

## Steps to Reproduce

1. Ensure `team_mode.enabled = true` and `team_mode.base_dir = "~/.omo"` in `~/.config/opencode/oh-my-openagent.jsonc` (this is a natural value given the docs describe `~/.omo` as "the default resolves to").
2. Create a fresh empty directory and run opencode in it:
   ```bash
   mkdir -p /tmp/repro && cd /tmp/repro
   opencode run "hi"
   ```
3. Observe the directory afterward:
   ```bash
   ls -la /tmp/repro
   ```

## Expected Behavior

`~/.omo` resolves to `$HOME/.omo` (e.g. `/home/user/.omo`). No stray directory is created in the cwd. Team-mode state (`runtime`, `teams`, `worktrees`) lives under `$HOME/.omo/`.

## Actual Behavior

A literal directory named `~` is created in the cwd:
```
/tmp/repro/
└── ~                               ← literal tilde directory
    └── .omo
        ├── runtime
        ├── teams
        └── worktrees
```

strace proof (captured via `strace -f -e trace=mkdir,mkdirat,openat -o /tmp/strace.log opencode run "hi"`):
```
mkdir("~/.omo", 0700)           = -1 ENOENT (No such file or directory)   # parent "~" missing
mkdir("~", 0700)                = 0                                         # creates LITERAL "~" dir
mkdir("~/.omo", 0700)           = 0
mkdir("~/.omo/teams", 0700)     = 0
mkdir("~/.omo/runtime", 0700)   = 0
mkdir("~/.omo/worktrees", 0700) = 0
```

The raw string `"~/.omo"` is passed straight to the kernel's `mkdir` syscall — `~` is never expanded.

## Root Cause

A 3-link chain:

**1. Config** — `~/.config/opencode/oh-my-openagent.jsonc`:
```jsonc
"team_mode": {
  "enabled": true,
  "base_dir": "~/.omo"   // literal unexpanded tilde
}
```

**2. Code** — `resolveBaseDir()` in `packages/team-core/src/team-registry/paths.ts` (and the installed bundle `dist/index.js:63420`):
```ts
export function resolveBaseDir(config: TeamModeConfig): string {
  return config.base_dir ?? path.join(homedir(), ".omo")
}
```
Because `config.base_dir` **is** set (`"~/.omo"`), the `homedir()` fallback is **never reached** — the raw literal string `"~/.omo"` is returned unexpanded. The `homedir()` safety net only protects the `null`/unset case.

**3. Exec** — `ensureBaseDirs()` in the same file (line 197) calls `mkdir(baseDir, { recursive: true, mode: 0o700 })`. Node/Bun treats `~` as a literal directory name (there is no shell in the path, so no tilde expansion occurs), creating a real `~` folder relative to the cwd.

## Corroborating Evidence

- **`--pure` run (plugins disabled):** the working directory stays completely empty — confirms the culprit is a plugin, not opencode core.
- **strace process tree:** the `mkdir("~")` is performed by an opencode worker **thread** (zero `execve`, no `/bin/sh` or `/bin/bash` involved) — this is **not** a shell-tilde-quoting bug; the literal string is passed in-process by the plugin runtime.
- **Schema check:** `assets/oh-my-opencode.schema.json` defines `base_dir` as `"type": "string"` with **no default**, and docs/AGENTS.md show the example default as `"base_dir": null` (which correctly triggers the `homedir()` path). So the omo schema itself does not emit `"~/.omo"` — but the docs describe `~/.omo` as "the default resolves to", which leads users (or older installers) to write the literal `"~/.omo"` into the config, which then bypasses the only tilde-safe code path.
- **Other plugins/MCP ruled out:** grepped every other active plugin (`cursor-acp`, `opencode-wakatime`, `leash`) and MCP server (`firecrawl`, `zai-mcp-server`, `stitch`) — none reference `~/.omo`. Only oh-my-openagent touches it.

## Bug Class

Trust of a config/user-supplied path without tilde expansion. Same family as:
- `anomalyco/opencode` #32223 — `~` not expanded in file tool path arguments
- `anomalyco/opencode` #23695 — plugin config does not support tilde (`~/`) paths
- `code-yeongyu/oh-my-openagent` #4593 — prompt file under home directory rejected

## Suggested Fix (location only — not applied)

The most robust single-point fix is to expand a leading `~` in `resolveBaseDir()` so any `base_dir` value (config-supplied or otherwise) is normalized before reaching `mkdir`:

```ts
import { homedir } from "node:os"

function expandTilde(p: string): string {
  return p.startsWith("~") ? path.join(homedir(), p.slice(1)) : p
}

export function resolveBaseDir(config: TeamModeConfig): string {
  const raw = config.base_dir ?? path.join(homedir(), ".omo")
  return expandTilde(raw)
}
```

Alternatively (defense in depth): normalize at the config/schema boundary so an unexpanded tilde never reaches runtime code.

## Environment

- oh-my-openagent: 4.10.0 (`oh-my-openagent@latest`)
- opencode: 1.17.7
- OS: Linux (Pop!_OS), x86_64
- Shell: zsh (`$HOME` = `/home/<user>` is set normally)

EvangelosMoschou:
## Problem

When users set `team_mode.base_dir = "~/.omo"` in their config, the raw tilde-prefixed string was passed straight to `mkdir`, creating a **literal `~` directory in the current working directory** instead of expanding `~` to the user's home directory.

This polluted every directory opencode was launched in with a stray `~` folder and silently landed team-mode state in the wrong place.

## Root Cause

`resolveBaseDir()` in `src/features/team-mode/team-registry/paths.ts` returns `config.base_dir` as-is when set. Shell-style tilde expansion never happens because Node.js APIs (path.join, fs.mkdir) don't expand `~`.

```ts
// Before
export function resolveBaseDir(config: TeamModeConfig): string {
  return config.base_dir ?? path.join(homedir(), ".omo")
}
```

## Fix

Added a small `expandLeadingTilde()` helper that handles the two tilde forms:
- `~`         → `$HOME`
- `~/...`     → `$HOME/...`
- anything else is left untouched (so `/tmp/foo`, `./local`, and a stray `foo~/bar` are unaffected).

```ts
function expandLeadingTilde(input: string): string {
  if (input === "~") return homedir()
  if (input.startsWith(`~${path.sep}`)) return path.join(homedir(), input.slice(2))
  return input
}

export function resolveBaseDir(config: TeamModeConfig): string {
  return expandLeadingTilde(config.base_dir ?? path.join(homedir(), ".omo"))
}
```

`path.sep` is used so this also handles Windows `~\\foo` correctly.

## TDD Evidence

**RED** — added 2 new tests in `paths.test.ts`:
- `base_dir: "~/.omo"` resolves to `path.join(homedir(), ".omo")`
- `base_dir: "~/.omo/teams"` resolves to `path.join(homedir(), ".omo", "teams")`

Both failed before the fix with the literal tilde string returned.

**GREEN** — apply `expandLeadingTilde()` in `resolveBaseDir`; all 7 `paths` tests pass, 262 team-mode tests pass.

**SURFACE** — change is contained to `resolveBaseDir()`; no other function in the team-mode module is affected.

## Constraints Checklist

- [x] No `as any` / `@ts-ignore` / `@ts-expect-error`
- [x] No test deletions
- [x] No new dependencies
- [x] TDD: red test written first
- [x] No `unwrap()` / `expect()` shortcuts in test bodies
- [x] No `setTimeout` / `sleep` in tests
- [x] Subscribed before triggering (no race conditions)
- [x] BDD test style (`// given`, `// when`, `// then`)

Closes #5331

---
## Summary by cubic
Expands leading tilde in `team_mode.base_dir` to the user’s home directory to prevent creating a literal `~` folder and store team‑mode state in the right place. Handles `~` and `~/…` on all platforms (fixes #5331).

- **Bug Fixes**
  - Added `expandLeadingTilde()` and applied it in `resolveBaseDir`; non-leading tildes and other paths remain unchanged.
  - Added tests for `~/.omo` and `~/.omo/teams`.

<sup>Written for commit 17a5856d5862563fbd72f54a206f79cf3ca20301. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5369?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/team-mode/team-registry/paths.test.ts
+++ b/src/features/team-mode/team-registry/paths.test.ts
@@ -54,6 +54,28 @@ describe("paths", () => {
     expect(resolvedBaseDir).toBe("/tmp/test-abc")
   })
 
+  test("resolveBaseDir expands leading tilde to homedir (#5331)", () => {
+    // given - user wrote the natural docs value "~/.omo" expecting shell expansion
+    const config = TeamModeConfigSchema.parse({ base_dir: "~/.omo" })
+
+    // when
+    const resolvedBaseDir = resolveBaseDir(config)
+
+    // then - should be $HOME/.omo, NOT a literal "~" dir in cwd
+    expect(resolvedBaseDir).toBe(path.join(homedir(), ".omo"))
+  })
+
+  test("resolveBaseDir expands tilde with subpath (#5331)", () => {
+    // given - tilde in the middle of a path is left alone; only the leading ~ is expanded
+    const config = TeamModeConfigSchema.parse({ base_dir: "~/.omo/teams" })
+
+    // when
+    const resolvedBaseDir = resolveBaseDir(config)
+
+    // then
+    expect(resolvedBaseDir).toBe(path.join(homedir(), ".omo", "teams"))
+  })
+
   test("discoverTeamSpecs prefers project scope", async () => {
     // given
     const rootDirectory = await createTemporaryRoot()
--- a/src/features/team-mode/team-registry/paths.ts
+++ b/src/features/team-mode/team-registry/paths.ts
@@ -19,8 +19,14 @@ function getTeamDirectory(baseDir: string, teamName: string, scope: "user" | "pr
   return path.join(baseDir, "teams", teamName)
 }
 
+function expandLeadingTilde(input: string): string {
+  if (input === "~") return homedir()
+  if (input.startsWith(`~${path.sep}`)) return path.join(homedir(), input.slice(2))
+  return input
+}
+
 export function resolveBaseDir(config: TeamModeConfig): string {
-  return config.base_dir ?? path.join(homedir(), ".omo")
+  return expandLeadingTilde(config.base_dir ?? path.join(homedir(), ".omo"))
 }
 
 export function getTeamSpecPath(


MoerAI:
Thanks for the PR, @EvangelosMoschou. This is a tested fix linked to #5331, but it currently conflicts with `dev` (GitHub marks it as conflicting), so it needs a rebase. One note: #5332 also targets #5331 (tilde expansion in the team base dir), so there is an overlap, and only one of the two will land. Could you rebase on current `dev`? Once it is mergeable I will compare the two and take another look. Thanks for the work.

EvangelosMoschou:
Closing in favor of @manascb1344's #5332 which targets the correct file location (`packages/team-core/`). My PR #5369 fixed the same issue in `src/features/team-mode/` which has been refactored out. #5332's approach is also simpler (one-liner). Thanks @MoerAI for the pointer.

PR #5458, opened 2026-06-20
Stop automatic memory compaction from firing repeatedly and from running while results are being gathered

andrescera:
**Stop automatic memory compaction from firing repeatedly and from running while results are being gathered**

**The problem.** To keep long sessions healthy, the assistant automatically condenses its memory when it gets close to full. In practice this kept happening over and over: it would condense, work for a moment, and condense again almost immediately — and it could even condense in the middle of gathering results from background work, throwing away the very information it had just collected. The point at which condensing kicks in was also fixed in the code with no way to adjust it.

**When it started.** This automatic condensing was introduced in v3.2.2 (commit 62e168747) and has behaved this way since.

**The solution.** Condensing now happens once when memory genuinely gets close to full, and does not trigger again until memory has actually come back down, so it no longer repeats every turn. It also holds off while the assistant is still collecting results from background work, so freshly gathered information is not discarded. Finally, the threshold and the cooldown are now configurable, with defaults that match today's behavior, for anyone who wants to tune when condensing happens.

---
## Summary by cubic
Stops automatic memory compaction from re-firing each turn and from running while background results are being collected. Adds a configurable `compaction` block (threshold, cooldown, enabled) with defaults that keep current behavior.

- **Bug Fixes**
  - Re-arms compaction only after usage drops below (threshold − 15%) so it doesn't immediately re-compact.
  - Enforces a per-session cooldown before compaction can run again.
  - Skips compaction while `background_output` results are being collected so fresh results aren't discarded.

- **New Features**
  - Adds `compaction` config: `preemptive_threshold` (0–1, default 0.78), `cooldown_ms` (default 60000), `enabled` (default true).
  - Root config and JSON schema updated; compaction logic now reads these values at runtime.

<sup>Written for commit 17dc2054ff45d640f98c67cb18fbec78fa4ea80e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5458?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


EZotoff:
We hit the exact scenario this PR addresses. Our log evidence:

An Atlas orchestrator collected 6 completed background tasks via `background_output` in a single turn. The combined results measured **788,883 bytes (788KB)**. OpenCode's `compaction.prune` immediately fired and pruned all 6 results:

```
timestamp=2026-07-16T18:12:45.713Z message="found pruned=788883 total=788883"
timestamp=2026-07-16T18:12:45.891Z message="pruned count=6"
```

The agent then observed "Background outputs keep getting cleared" and had to pivot to direct file verification. This PR would prevent compaction from discarding freshly collected results. Strong +1 from us.

code-yeongyu:

PR #5459, opened 2026-06-20
Fix background task result retrieval so an agent can read a subagent's final answer

andrescera:
**Fix background task result retrieval so an agent can read a subagent's final answer**

**The problem.** When an agent delegates work to run in the background and later asks for the result, it frequently can't get the one thing it needs: the subagent's final answer. Asking for the result returns the entire background conversation — including very large intermediate output like full build and test logs — instead of the recent, relevant part. The option meant to return "just the most recent message" doesn't actually narrow anything, so even when the agent explicitly asks for the ending, it still receives everything, and the final answer at the very end gets dropped before the agent can see it. In practice the agent asks again and again, never receives the conclusion, and its own working memory fills with noise it never wanted.

**When it started.** The "most recent message" option was added in v4.10.0 (commit 6092c8ef1), on top of result formatting that had never limited large intermediate output.

**The solution.** Make result retrieval honor what the caller asks for. A request for the most recent part of a background task returns a bounded slice from the end that reliably contains the final answer, whether the task is finished or still running, and the final answer is shown first so it can't be cut off. Bulky intermediate output no longer crowds out that answer, and a caller's choice to leave heavy detail out is respected even while the task is still in progress. Lightweight status checks stay small and cheap; only an explicit request for full detail returns more.

---
## Summary by cubic
Fixes background task result retrieval so agents reliably get a subagent’s final answer without being buried by logs. Adds bounded tail slicing and truncation for smaller, cheaper status checks.

- **Bug Fixes**
  - Honor `from_end` for both `full_session` and completed results; when no limit is given, return the last 20 messages.
  - Completed results lead with the “Final answer,” followed by a bounded “Older context” (size set by `message_limit`, default 20).
  - Respect `include_thinking` and `include_tool_results`: explicit values are honored; defaults are on only while the task is active.
  - Truncate `tool_result` blocks to 2000 chars with an ellipsis to avoid log bloat.
  - Preserve the “No new output since last check” notice; added tests for tail slicing, truncation, and include-flag behavior.

<sup>Written for commit 3ebb34125c3a05d00bbf16d2a93678a743d7cda6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5459?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/tools/background-task/create-background-output.filters.test.ts
+++ b/packages/omo-opencode/src/tools/background-task/create-background-output.filters.test.ts
@@ -1,0 +1,129 @@
+/// <reference types="bun-types" />
+
+import { describe, expect, test } from "bun:test"
+import type { ToolContext } from "@opencode-ai/plugin/tool"
+import type { BackgroundTask } from "../../features/background-agent"
+import type { BackgroundOutputClient, BackgroundOutputManager } from "./clients"
+import { createBackgroundOutput } from "./create-background-output"
+
+const projectDir = "/Users/yeongyu/local-workspaces/oh-my-opencode"
+
+const mockContext = {
+  sessionID: "test-session",
+  messageID: "test-message",
+  agent: "test-agent",
+  directory: projectDir,
+  worktree: projectDir,
+  abort: new AbortController().signal,
+  metadata: () => {},
+  ask: async () => {},
+} as ToolContext
+
+function createTask(overrides: Partial<BackgroundTask> = {}): BackgroundTask {
+  return {
+    id: "bg_filters_task",
+    sessionId: "ses-filters",
+    parentSessionId: "main-1",
+    parentMessageId: "msg-1",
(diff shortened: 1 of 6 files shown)


code-yeongyu:
[sisyphus-bot] Thanks for this detailed fix! The background-task result retrieval area has since been implemented in `dev` with `task-result-format.ts` and `full-session-format.ts`. If you find that the subagent final-answer problem you identified is still present after those additions, please open a new PR. Closing to keep the queue current.

PR #5460, opened 2026-06-20
Fix the review step so a finished plan can complete and finished work isn't re-checked forever

andrescera:
**Fix the review step so a finished plan can complete and finished work isn't re-checked forever**

**The problem.** After the assistant finishes a piece of work and hands it to a reviewer, two things went wrong. First, if a reviewer's reply didn't contain a clear approve-or-reject decision, the plan could get stuck permanently: it would never be considered finished, and the assistant would keep trying with no way out. Second, once a part of the plan was done, the assistant could fall back into re-checking that same finished work again and again, repeating the same review instead of moving on to what was left.

**When it started.** The strict approve check was introduced in v3.12.0 (commit 04b0c6f33); the repeated re-checking of completed work comes from the completion-enforcement behavior added in v3.11.0 (commit bf8d0ffcc) and extended in v3.14.1 (commit 6e700e000).

**The solution.** When a reviewer doesn't give a clear decision, the assistant now pauses and asks you to confirm or re-run the review, instead of silently looping forever. And once a piece of work is verified and marked done, the assistant no longer re-reviews it - it moves on to the next unfinished part, or wraps up if nothing is left.

---
## Summary by cubic
Stops infinite review loops and prevents re-verifying finished work so plans can complete. Pauses on unclear final-review verdicts and guides you to confirm or re-run the review.

- **Bug Fixes**
  - Final-review outputs are now classified as approve/reject/missing; on missing, we set `waitingForFinalWaveApproval`, show a toast, and send a short escalation with steps to confirm or re-run.
  - Once a task is verified, we skip the heavy 4-phase gate and emit a lightweight advance directive to move to the next unchecked task; `verifiedTaskKeys` tracks this and resets on plan change (also used for `background_output` retrievals).
  - Final-wave task detection uses plan state; exported `classifyFinalWaveVerdict()` for reuse and to avoid false re-drives.
  - Added tests for verdict classification and reminder builders.

<sup>Written for commit f7aa7d3ca7fc5c5feb218d8e36fd95ebccd0ae2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5460?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/atlas/final-wave-approval-gate.test.ts
+++ b/packages/omo-opencode/src/hooks/atlas/final-wave-approval-gate.test.ts
@@ -7,10 +7,94 @@ import { createOpencodeClient } from "@opencode-ai/sdk"
 import type { AssistantMessage, Session } from "@opencode-ai/sdk"
 import type { BoulderState } from "../../features/boulder-state"
 import { clearBoulderState, writeBoulderState } from "../../features/boulder-state"
+import { classifyFinalWaveVerdict } from "./final-wave-approval-gate"
 import { createAtlasHook } from "./index"
 
 type AtlasHookContext = Parameters<typeof createAtlasHook>[0]
 type PromptMock = ReturnType<typeof mock>
+
+describe("classifyFinalWaveVerdict", () => {
+  test("returns approve when the output carries an APPROVE verdict", () => {
+    // given
+    const output = "Tasks [4/4 compliant] | VERDICT: APPROVE"
+
+    // when
+    const verdict = classifyFinalWaveVerdict(output)
+
+    // then
+    expect(verdict).toBe("approve")
+  })
+
+  test("returns reject when the output carries a REJECT verdict", () => {
+    // given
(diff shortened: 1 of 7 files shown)


code-yeongyu:
Thanks for the work here. I landed this via #5605, rebased on current `dev`, with one additional safety fix: mixed `VERDICT: APPROVE` / `VERDICT: REJECT` output is now treated as missing rather than silently approving ambiguous final-review text.

#5605 includes the same review-loop fixes, fresh Atlas regression tests, and OpenCode QA evidence, so I am closing this PR as superseded.

PR #5482, opened 2026-06-22
fix(skills): resolve skill_mcp servers from runtime config (lazy, no deadlock)

sjawhar:
## Summary

- `skill_mcp` could not find MCP servers from skills whose source paths are injected into opencode's merged config **at runtime** by other plugins — their `config` hooks run after plugin `server()`, so the on-disk-built skill list never saw them.
- Fixes it by lazily resolving the runtime config on the first `skill_mcp` call (after startup) and caching it — avoiding the startup **deadlock** that a load-time `client.config.get()` causes.
- Skill MCPs stay lazy and skill-scoped; nothing is written to global `config.mcp`.

## Changes

- **`plugin/runtime-skill-resolver.ts`** (new): `createRuntimeSkillsResolver` — on first call, fetches the merged runtime `skills` via `client.config.get()`, rebuilds merged skills through `createSkillContext({ hostSkills })`, and caches. `readRuntimeHostSkills` documents why it must run post-startup (the plugin's `server()` executes inside the Plugin service's `InstanceState.make`, so a `/config` roundtrip there waits on an init that can't complete until `server()` returns → deadlock). Falls back to base skills on any failure.
- **`plugin/skill-context.ts`**: `createSkillContext` gains an optional `hostSkills` arg; when omitted it still reads the on-disk config (default unchanged).
- **`plugin/tool-registry-core-tools.ts`**: wires the resolver into `skill_mcp`'s `getLoadedSkills`.
- **`tools/skill-mcp/tools.ts`**: `getLoadedSkills` is now awaited (type widened to allow async).
- Tests: new `runtime-skill-resolver.test.ts` (no fetch at construction / fetch+merge on first call / single-fetch caching / fallback on error) and a new `skill-context` case proving a runtime-injected skill's embedded `mcp:` survives into `mergedSkills`.

## Testing

```bash
bun run typecheck   # clean
bun run build       # succeeds (CI green)
bun test            # resolver + skill-mcp + skill-context + skills-loader suites pass
```

End-to-end via the local-build method in CONTRIBUTING (`opencode run` against `dist/index.js`): with a skill source injected into the runtime config by another plugin, `skill_mcp(mcp_name="slack", tool_name="channels_list")` resolved and returned live data (previously reported the server "not found"); startup no longer hangs.

## PR Checklist

- [x] `bun test` passes
- [x] `bun run test:codex` passes — no Codex-side changes; CI `codex-compatibility` (ubuntu/macos/windows) green
- [x] Tested locally with OpenCode
- [x] QA evidence under `.omo/evidence/` — N/A (not a harness-connected change)
- [x] Updated documentation if needed — N/A (no doc-facing change)

### DIFF
--- a/packages/omo-opencode/src/plugin/runtime-skill-resolver.test.ts
+++ b/packages/omo-opencode/src/plugin/runtime-skill-resolver.test.ts
@@ -1,0 +1,124 @@
+import { describe, it, expect } from "bun:test"
+import { createRuntimeSkillsResolver, type RuntimeHostSkills } from "./runtime-skill-resolver"
+import type { LoadedSkill } from "../features/opencode-skill-loader/types"
+
+function skill(name: string, mcp?: Record<string, unknown>): LoadedSkill {
+  return {
+    name,
+    definition: { name, description: `Test skill ${name}` },
+    scope: "config",
+    ...(mcp ? { mcpConfig: mcp as LoadedSkill["mcpConfig"] } : {}),
+  }
+}
+
+describe("createRuntimeSkillsResolver", () => {
+  it("does NOT fetch the runtime config at construction (deadlock guard)", () => {
+    // given
+    let readCount = 0
+    const base = [skill("playwright", { playwright: {} })]
+
+    // when - resolver is created (this happens during plugin load)
+    createRuntimeSkillsResolver({
+      baseSkills: base,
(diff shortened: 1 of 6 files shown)


MoerAI:
Thanks @sjawhar, merging.

PR #5541, opened 2026-06-23
feat(frontend): add designpowers reference ruleset

IYENTeam:
## Summary
Reworks the designpowers integration as an extension of the shared `frontend` skill instead of a new standalone `.opencode/skills/web-ui-design` skill. `/frontend` remains the public activation point; designpowers is now an internal `references/designpowers/` ruleset for personas, accessibility, critique, design debt, handoff, synthetic user testing, and role-reference guidance.

The third-party designpowers corpus now follows the existing DMCA-safe frontend reference pattern: a pinned upstream submodule under `packages/shared-skills/upstreams/designpowers`, a manifest/materializer, generated `vendor/` reference files kept out of git, and provenance/package tests.

## Changes
- Removes the standalone `.opencode/skills/web-ui-design` tree from the final PR diff.
- Adds `packages/shared-skills/skills/frontend/references/designpowers/` as a frontend-internal ruleset discovered from `packages/shared-skills/skills/frontend/SKILL.md`.
- Adds `packages/shared-skills/upstreams/designpowers` pinned to `cb00757da9d554591fa78d27aa1854d60a05c4f7`.
- Adds `packages/shared-skills/scripts/designpowers-refs-manifest.mjs` and wires it into the existing frontend materialization pipeline.
- Updates provenance, package-layout, materialization, upstream, and manifest tests to cover designpowers as the fourth frontend upstream.

## QA & Evidence
- **What was tested:** `node packages/omo-codex/plugin/scripts/materialize-shared-upstreams.mjs --strict`.
  **Observed result:** materialized 147 frontend reference files from submodules, including 38 designpowers files.
  **Why sufficient:** proves designpowers participates in the existing strict build/package materialization path.

- **What was tested:** designpowers corpus smoke.
  **Observed result:** 27 materialized skills, 10 materialized agents, excluded bridge/state/router skills absent, `vendor/` ignored by git, byte-compare OK against `packages/shared-skills/upstreams/designpowers`.
  **Why sufficient:** verifies the selected corpus is complete, unmodified, and does not commit raw third-party files.

- **What was tested:** targeted Bun tests:
  `bun test packages/shared-skills/materialize-frontend-refs.test.ts packages/shared-skills/frontend-thirdparty-manifest.test.ts packages/shared-skills/provenance-gate.test.ts packages/shared-skills/upstreams.test.ts script/package-layout.test.ts script/package-layout-exclusion.test.ts packages/omo-opencode/src/shared-skills-package.test.ts`
  **Observed result:** 27 pass, 0 fail.
  **Why sufficient:** covers shared-skill discovery/package shape, materialization, provenance pins, ignored generated files, and frontend upstream declarations.

- **Evidence file:** `.omo/evidence/20260623-designpowers-frontend-submodule-pr/qa-report.md`

## Risks & Residuals
- Full repository `bun test`, full `typecheck`, and full `build` were not rerun after this review response. The touched surface is markdown/reference routing plus materializer/test wiring, and the targeted tests cover the changed behavior.

## Related Issues
Responds to review feedback on PR #5541.


.opencode/skills/web-ui-design/SKILL.md
```diff
@@ -0,0 +1,58 @@
+---
+name: web-ui-design
```
code-yeongyu: This should not be introduced as a separate top-level skill. The trigger surface here is exactly the one already owned by the shared `frontend` skill: web UI, UX, visual design, accessibility, design QA, and frontend implementation routing. Adding `web-ui-design` beside it creates a second activation path that will compete with `/frontend`, `/visual-qa`, `/ulw-plan`, and `/review-work` instead of being discovered through them.

Please move this into `packages/shared-skills/skills/frontend` as a reference/ruleset that the existing frontend router can load when design operating-layer or designpowers-style guidance is needed. That keeps the behavior cross-harness and makes the new material discoverable from the existing frontend skill rather than adding an OpenCode-only skill namespace.

.opencode/skills/web-ui-design/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+
+# web-ui-design
+
+`web-ui-design` is an OpenAgent-native router and design operating layer. It does not implement a planner, builder, visual harness, scheduler, hook, bridge, or tool API. Its job is to preserve the OpenAgent execution spine while carrying design brief, persona, taste, accessibility, debt, and handoff context between the OpenAgent skills that already own the work.
+
+Use this skill when a request needs web UI direction, frontend build guidance, design-state continuity, persona-aware review, or design debt handling across an OpenAgent workflow.
+
+## Load Order
+
+Read these references before acting:
```
code-yeongyu: This load order is the shape I would expect inside the existing frontend skill, not as an independent skill. The frontend skill already has the `SKILL.md` router + `references/` structure for selecting design/perfection/ui-ux-db material. Please fold this routing into that router, for example as a new `references/designpowers/README.md` or a small frontend sub-route that points at the designpowers lane docs.

Concretely, `/frontend` should remain the user-facing activation point, and this designpowers layer should be imported/loaded as a submodule-style reference from there, similar to how the browsing stack exposes `insane-search` as an internal reference module rather than a separate competing skill.

.opencode/skills/web-ui-design/references/designpowers/MANIFEST.md
```diff
@@ -0,0 +1,99 @@
...
+Upstream repository: `https://github.com/Owl-Listener/designpowers`
+
+Local source clone: `/tmp/designpowers-source`
+
+Source commit: `500869472c4e216524669fd5ef57cd13fe25e330`
+
+## Included Files
+
+- `/tmp/designpowers-source/LICENSE` copied to `LICENSE`
+- `/tmp/designpowers-source/agents/*.md` copied to `agents/`
```
code-yeongyu: Please do not commit this third-party corpus directly under `.opencode/skills`. The frontend skill already has a precedent for upstream design/reference content: keep the upstream as a pinned submodule outside `skills/`, declare the path map in a manifest script, materialize the files into `packages/shared-skills/skills/frontend/references/...` during the build/package flow, and keep the generated third-party files ignored in git.

That matters here because this PR adds thousands of lines of vendored content in a harness-local path. Moving it to the frontend submodule/materializer pattern preserves provenance, keeps the committed tree small, avoids OpenCode-only drift, and lets Codex/marketplace packaging receive the same reference corpus through the shared skill pipeline.

.opencode/skills/web-ui-design/references/routing.md
```diff
@@ -0,0 +1,79 @@
...
+- design-state constraints that affect the task;
+- required `/frontend` loading for UI implementation;
+- required `/visual-qa` loading for rendered visual proof;
+- the design debt rule: unresolved accessibility debt cannot disappear into a summary.
+
+Direct implementation outside `/start-work` is not part of this routing contract when a Prometheus plan is active.
+
+## UI Build Through `/frontend`
+
+`/frontend` owns actual UI build quality. `web-ui-design` may point it at:
```
code-yeongyu: This section says `/frontend` owns the actual UI build quality, which is the right boundary. Because of that, the designpowers guidance should be a frontend reference consumed by `/frontend`, not a peer skill that re-describes when to call `/frontend`. Otherwise agents have to choose between two web/design skills and may load this router without the frontend skill's existing DESIGN.md gate, real-browser QA expectations, React tooling, and performance rules.

Please make the frontend skill the only public route and let this material enrich its routing/context. A good target shape would be: frontend `SKILL.md` routes to a new designpowers reference when relevant; the reference then points to lane docs/role references; build scripts materialize third-party upstream files before packaging.

code-yeongyu (changes requested):
Thanks for putting this together. I agree with the direction of preserving the existing OpenAgent workflow for design/frontend work, but I think this should not land as a new standalone `.opencode/skills/web-ui-design` skill tree. Frontend/web UI work is already owned by the shared `frontend` skill, and the repo already has a DMCA-safe reference-corpus pattern there: pinned upstream submodules under `packages/shared-skills/upstreams`, a manifest/materializer, generated reference files kept out of git, and provenance tests.

Please rework this as an extension of `packages/shared-skills/skills/frontend` instead: add the designpowers material as a frontend reference/ruleset discovered from the frontend router, and use the existing submodule + manifest + materialize pattern (similar to the insane-search/reference-module shape and the current frontend upstream materialization pipeline) instead of committing the third-party corpus directly under `.opencode/skills`. That keeps the capability cross-harness, discoverable from the existing `/frontend` activation path, and consistent with how we ship upstream reference material.

code-yeongyu:
One more follow-up suggestion while you are restructuring this into the existing frontend skill:

Please also update `packages/shared-skills/skills/frontend/SKILL.md` so the design-system workflow is much more explicit and forceful when this new designpowers/reference layer is used.

Right now the frontend skill already says "No design system = no UI work," but if we add this designpowers material under `/frontend`, I think the router should make the workflow hard to miss:

1. Before UI implementation, require the agent to synthesize or update the project `DESIGN.md` from the designpowers/reference inputs, user brief, brand/taste direction, accessibility constraints, and existing UI context.
2. Make `DESIGN.md` the contract that downstream implementation follows: tokens, typography, spacing, component primitives, motion rules, responsive behavior, accessibility constraints, and accepted design debt should all be named there before components are written.
3. Route the designpowers material as input to that design-system step, not as a replacement for the frontend skill. In other words, designpowers should help the frontend skill produce a stronger project design system, and then `/frontend` should build from that system.
4. Add a short workflow section in the frontend skill that says something like: "When designpowers/reference guidance is loaded, first distill it into the project design system, then implement against that system, then verify with visual QA and review-work evidence."

This would make the new material genuinely useful instead of just another reference corpus: agents would use it to form a coherent design system first, then build consistently from that system across the actual UI work.

IYENTeam:
Addressed the review direction in commit 087079125. This no longer lands as a standalone .opencode skill: the final PR diff folds designpowers into packages/shared-skills/skills/frontend/references/designpowers, wires it through the existing frontend router, and sources the third-party corpus from a pinned packages/shared-skills/upstreams/designpowers submodule via the existing materialize-shared-upstreams flow. The materialized vendor corpus is gitignored and covered by provenance/materialization/package tests (targeted run: 27 pass, 0 fail).

IYENTeam:
please review boy.

packages/shared-skills/skills/frontend/SKILL.md
```diff
@@ -14,6 +14,7 @@ This file is a router, not a rulebook. The rules live in three rulesets under `r
...
 | Writing or modifying frontend code, OR auditing performance / SEO / accessibility / quality | ALSO `references/perfection/README.md`. Lighthouse 100 in every category, measured on real Playwright Chromium (never the `lighthouse` CLI), achieved through architecture — never by dropping animations or hiding content. |
 | Looking up a concrete style, color palette, font pairing, chart type, landing-page structure, or UX guideline — or generating a project design system from keywords | `references/ui-ux-db/README.md`. A searchable CSV database with a CLI; a lookup tool, not a posture. Load on demand; `design` stays the source of truth for taste and the `DESIGN.md` contract. |
+| Design operating-layer work: personas, cognitive accessibility, design critique, design debt, handoff, synthetic user testing, or designpowers-style guidance | `references/designpowers/README.md`. This is an internal frontend ruleset, not a separate skill. It enriches `/frontend` routing with design brief, role-reference, accessibility, evidence, and debt language while preserving the `design` and `perfection` gates. |
```
code-yeongyu, line 17: This routes designpowers as an internal frontend ruleset, which is the right shape, but it still does not make the design-system workflow explicit enough. The follow-up request was not just to preserve the existing `DESIGN.md` gate; it was to make agents use designpowers/reference inputs to construct or update the project `DESIGN.md` before implementation.

Please add a clear workflow here or immediately below the routing table that says: when designpowers guidance is loaded for implementation/design work, first synthesize/update `DESIGN.md` from the user brief, designpowers references, brand/taste direction, persona/accessibility constraints, and existing UI context; then treat that `DESIGN.md` as the implementation contract for tokens, typography, spacing, component primitives, motion, responsive behavior, accessibility constraints, and accepted design debt.

packages/shared-skills/skills/frontend/SKILL.md
```diff
@@ -79,6 +80,10 @@ python3 $SKILL_DIR/references/ui-ux-db/scripts/search.py "<query>" --stack <stac
 
 Domains: `product` `style` `typography` `color` `landing` `chart` `ux` `react` `web` `prompt`. Stacks: `html-tailwind` (default) `react` `nextjs` `vue` `svelte` `astro` `swiftui` `react-native` `flutter` `shadcn` `jetpack-compose`.
 
+## Ruleset 4 — designpowers (`references/designpowers/`)
+
+`README.md` routes design operating-layer guidance from the pinned `Owl-Listener/designpowers` reference corpus into the existing frontend workflow. Load it when a frontend task needs explicit personas, accessibility and cognitive constraints, design critique, design debt, handoff, synthetic user testing, motion guidance, or role-reference prompts. It does not replace this frontend skill, `/visual-qa`, `/ulw-plan`, `/start-work`, or `/review-work`; it only supplies richer design context inside those existing routes.
```
code-yeongyu: This section says designpowers supplies richer design context, but it stops short of saying how that context becomes actionable. Please make the intended pipeline explicit: designpowers reference input -> project `DESIGN.md` synthesis/update -> UI implementation against that design system -> visual QA/review-work evidence.

Without that, agents can load the new corpus as another reference set without forming the coherent design system we wanted this change to force.

packages/shared-skills/materialize-frontend-refs.test.ts
```diff
@@ -2,7 +2,7 @@ import { describe, expect, test } from "bun:test";
 import { existsSync } from "node:fs";
 import { join } from "node:path";
 import { materializeFrontendRefs } from "./scripts/materialize-frontend-refs.mjs";
-import { brandStems, frontendSkillRoot, thirdPartyRelativePaths, uiUxDbScripts } from "./scripts/frontend-refs-manifest.mjs";
+import { brandStems, designpowersThirdPartyRelativePaths, frontendSkillRoot, thirdPartyRelativePaths, uiUxDbScripts } from "./scripts/frontend-refs-manifest.mjs";
```
code-yeongyu, line 5: This import uses the new `designpowersThirdPartyRelativePaths()` export from `frontend-refs-manifest.mjs`, but the sibling declaration file `packages/shared-skills/scripts/frontend-refs-manifest.d.mts` was not updated to declare that export. Please add it there as well so the JS manifest and typed declaration stay in sync for TS consumers/tools.

code-yeongyu (changes requested):
Thanks for the rework. This is much closer to the direction from the earlier review: the standalone `.opencode/skills/web-ui-design` tree is gone, the material now lives under the shared `frontend` skill, `designpowers` is routed as an internal frontend ruleset, and the third-party corpus is modeled as a pinned upstream/submodule with materialized references instead of committed vendor files. That addresses the biggest architectural concern.

I still need to request changes before this can be approved. The remaining blocker is the follow-up design-system workflow we discussed: when designpowers/reference guidance is loaded, the frontend skill should explicitly require agents to first distill the user brief, designpowers inputs, brand/taste direction, accessibility/persona constraints, and existing UI context into the project `DESIGN.md`, then implement against that design-system contract. Right now the PR says designpowers preserves the existing `DESIGN.md` gate, but it does not add the stronger workflow that makes designpowers useful as an input to design-system construction.

There is also a small API/declaration mismatch: `frontend-refs-manifest.mjs` now exports `designpowersThirdPartyRelativePaths()`, and the new test imports it, but the sibling `frontend-refs-manifest.d.mts` declaration was not updated. Please add the declaration so the manifest module stays consistent for typed consumers.

Once those are fixed, I think this should be in good shape: the shared-skill placement, `/frontend` discoverability, submodule/materializer shape, attribution/pinning, and non-standalone routing are all aligned with the requested direction.

IYENTeam:
Addressed the latest review in commit 9ae4e201c.\n\nChanges:\n- Added an explicit Design System Workflow to the shared frontend skill: when designpowers is loaded for implementation/redesign/design-system work, its inputs must first synthesize or update project DESIGN.md from the user brief, designpowers references, brand/taste direction, accessibility/persona constraints, existing UI context, and accepted debt.\n- Made the pipeline explicit: designpowers reference input -> DESIGN.md contract -> UI implementation -> visual QA/review-work evidence.\n- Added the missing declaration for designpowersThirdPartyRelativePaths() in packages/shared-skills/scripts/frontend-refs-manifest.d.mts.\n\nQA:\n- bun test packages/shared-skills/materialize-frontend-refs.test.ts packages/shared-skills/frontend-thirdparty-manifest.test.ts packages/shared-skills/provenance-gate.test.ts packages/shared-skills/upstreams.test.ts script/package-layout.test.ts script/package-layout-exclusion.test.ts packages/omo-opencode/src/shared-skills-package.test.ts\n- Result: 27 pass, 0 fail

packages/shared-skills/skills/frontend/SKILL.md
```diff
@@ -1,11 +1,11 @@
...
+description: "MUST USE for ANY frontend, web UI, UX, or visual work — building, styling, or redesigning pages/components, React project setup, performance audits, design QA. Routes four rulesets: design (anti-slop taste router over 12 taste skills + 69 brand DESIGN.md refs — Apple, Stripe, Linear, Notion, Vercel, Claude, Nike — plus React dev tooling gate: react-grab, react-scan, react-doctor), perfection (Lighthouse 100 in every category via real Playwright Chromium audits, NEVER the lighthouse CLI, never by weakening UX), ui-ux-db (searchable 50+ styles, 97 palettes, 57 font pairings, 99 UX guidelines), designpowers (design operating layer, personas, accessibility, critique, debt, handoff, and role-reference guidance). Triggers: frontend, UI, UX, design, redesign, styling, layout, animation, motion, taste, premium, luxury, minimal, brutalist, Awwwards, anti-slop, polish, DESIGN.md, mockup, react setup, react-scan, react-doctor, lighthouse, performance, Core Web Vitals, LCP, CLS, INP, SEO, accessibility, a11y, WCAG, audit my site, make this faster, color palette, font pairing, looks generic, make it pretty, like X brand."
```
code-yeongyu, line 3: This shared skill update is correct, but the matching builtin skill copies were not updated. CI now fails the byte-equivalence guard because `packages/skills-loader-core/src/features/builtin-skills/frontend/SKILL.md` and `packages/omo-opencode/src/features/builtin-skills/frontend/SKILL.md` still contain the old three-ruleset frontend body. Please sync those copies with this shared `frontend/SKILL.md` change.

packages/shared-skills/skills/frontend/SKILL.md
```diff
@@ -14,9 +14,21 @@ This file is a router, not a rulebook. The rules live in three rulesets under `r
...
 | Looking up a concrete style, color palette, font pairing, chart type, landing-page structure, or UX guideline — or generating a project design system from keywords | `references/ui-ux-db/README.md`. A searchable CSV database with a CLI; a lookup tool, not a posture. Load on demand; `design` stays the source of truth for taste and the `DESIGN.md` contract. |
+| Design operating-layer work: personas, cognitive accessibility, design critique, design debt, handoff, synthetic user testing, or designpowers-style guidance | `references/designpowers/README.md`. This is an internal frontend ruleset, not a separate skill. It enriches `/frontend` routing with design brief, role-reference, accessibility, evidence, and debt language while preserving the `design` and `perfection` gates. |
 
 **For implementation work, design + perfection load together.** A page that hits Lighthouse 100 but looks like AI slop has failed; a page that looks beautiful but ships a 2 MB bundle has failed. Both win or neither does.
 
+## Design System Workflow
```
code-yeongyu, line 21: The new workflow addresses the requested design-system contract, but it currently only exists in `packages/shared-skills/skills/frontend/SKILL.md`. The builtin skill extraction test requires the same body in the builtin frontend skill sources; otherwise users reaching the builtin surface can get stale routing and CI stays red.

code-yeongyu (changes requested):
Thank you for the quick follow-up. The two substantive review items are now addressed in the shared frontend skill itself:

- `packages/shared-skills/skills/frontend/SKILL.md` now has an explicit Design System Workflow that forces designpowers/reference input through `DESIGN.md` before UI implementation.
- `packages/shared-skills/scripts/frontend-refs-manifest.d.mts` now declares `designpowersThirdPartyRelativePaths()`, and my narrow typed import probe passes.

I still need to request changes because the PR is not green and the failure is tied to this change path. The frontend skill is also vendored into the builtin-skill surfaces, and those copies were not updated. CI now fails on Ubuntu, macOS, and Windows in the same test:

`shared builtin skill extraction > #given extracted builtin skill markdown #when compared to TS sources #then bodies and metadata stay byte-equivalent`

Local evidence from the PR head:

- `bun test packages/shared-skills/materialize-frontend-refs.test.ts packages/shared-skills/frontend-thirdparty-manifest.test.ts packages/shared-skills/provenance-gate.test.ts packages/shared-skills/upstreams.test.ts script/package-layout.test.ts script/package-layout-exclusion.test.ts packages/omo-opencode/src/shared-skills-package.test.ts` passes: 27 pass, 0 fail.
- The typed import probe for `designpowersThirdPartyRelativePaths()` passes.
- The full CI test matrix fails because these files still contain the old three-ruleset frontend skill body:
  - `packages/skills-loader-core/src/features/builtin-skills/frontend/SKILL.md`
  - `packages/omo-opencode/src/features/builtin-skills/frontend/SKILL.md`

Please sync those builtin frontend skill copies with `packages/shared-skills/skills/frontend/SKILL.md` (or run the repo's extraction/sync path if there is one) so the byte-equivalence tests pass again. After that, this should be approvable: the architecture, submodule/materializer/provenance shape, explicit `DESIGN.md` workflow, and declaration export all look aligned.

code-yeongyu:
Pushed a focused sync fix for the CI blocker.

Commit pushed: `f30a9646e63a28f84488cfba43528b02f8928e90`

Files changed:

- `packages/skills-loader-core/src/features/builtin-skills/skills/frontend.ts`
- `packages/skills-loader-core/src/features/builtin-skills/frontend/SKILL.md`
- `packages/omo-opencode/src/features/builtin-skills/frontend/SKILL.md`

Verification run from `/private/tmp/omo-pr5541-builtin-sync`:

- `bun install --frozen-lockfile` - passed; installed workspace dependencies and completed the repo postinstall/build chain. A generated mode-bit side effect on `packages/omo-codex/plugin/components/codegraph/dist/serve.js` was reset and not included in the commit.
- `git diff --check` - passed.
- Direct `cmp` byte-equivalence checks from `packages/shared-skills/skills/frontend/SKILL.md` to both builtin `frontend/SKILL.md` copies - passed.
- `bun test packages/skills-loader-core/src/features/builtin-skills/shared-skill-extraction.test.ts packages/omo-opencode/src/features/builtin-skills/shared-skill-extraction.test.ts -t "shared builtin skill extraction"` - passed: 2 tests, 0 failures, 30 expectations.
- `bun test -t "shared builtin skill extraction"` - passed: 2 tests, 0 failures, 10168 filtered, 30 expectations.
- `bun test packages/shared-skills/materialize-frontend-refs.test.ts packages/shared-skills/frontend-thirdparty-manifest.test.ts packages/shared-skills/provenance-gate.test.ts packages/shared-skills/upstreams.test.ts script/package-layout.test.ts script/package-layout-exclusion.test.ts packages/omo-opencode/src/shared-skills-package.test.ts` - passed: 27 tests, 0 failures, 477 expectations.

Local evidence note: `.omo/evidence/20260624-pr5541-builtin-sync/summary.md` in the task worktree only; it was intentionally not staged or committed.

Remaining risk / next action: no known local blocker remains. Parent maintainer thread should re-read the new pushed head, run its strict review/QA, and either request changes or approve when CI is green.

code-yeongyu (approved):
Thanks for the careful rework here. I re-reviewed the current head `f30a9646e63a28f84488cfba43528b02f8928e90` and the blockers from the earlier review are resolved.

What I verified:

- The designpowers integration is now routed through the existing frontend skill surface instead of introducing a competing standalone skill activation path.
- `packages/shared-skills/skills/frontend/SKILL.md` now makes the `DESIGN.md` workflow explicit and forceful enough for agents to distill reference material, user brief, product constraints, accessibility concerns, and existing UI context before implementation.
- The shared frontend SKILL.md and both builtin copies are byte-identical, and the TS builtin description now matches the four-ruleset frontend frontmatter while continuing to load `loadSharedSkillTemplate("frontend")`.
- The previous declaration drift is fixed: `frontend-refs-manifest.d.mts` declares the designpowers export used by the materialization tests.
- The designpowers upstream is pinned as a submodule, package materialization paths are static, vendor output remains ignored in source but package-included after materialization, and attribution/provenance docs match the pinned source.
- I also had two independent review lanes check byte equivalence and CI/package/provenance risk; both came back PASS.

Local verification run from the review worktree:

- `git diff --check origin/dev...HEAD`
- builtin shared-skill extraction tests: 2 pass, 0 fail
- package/provenance/materialization/layout suite: 27 pass, 0 fail
- direct byte comparisons between shared frontend SKILL.md and both builtin copies

GitHub CI at this head is green across build, typecheck, test, and codex-compatibility on Ubuntu/macOS/Windows, plus the lightweight label/CLA/security checks. The earlier CI `action_required` state was just the fork approval gate; after approving the run, the matrix completed successfully.

No remaining blocking findings from my side. Nice cleanup on the shape of this; the final version keeps `/frontend` as the public routing point while making designpowers a strongly guided internal reference set instead of a second competing workflow.

PR #5592, opened 2026-06-26
fix(codegraph): keep multi-repo queries fresh through shared proxy

### LINKED ISSUE #5588: [Bug]: CodeGraph is inconsistent when agents work across multiple repositories
### Prerequisites

### Bug Description

CodeGraph behaves inconsistently when an agent works across multiple repositories in both oh-my-openagent editions.

The MCP server is effectively rooted in one default repository. CodeGraph tools can target another repository through `projectPath`, but CodeGraph 1.0.1 caches secondary repositories without watching them. Results from repositories other than the default working directory can therefore become stale until an explicit sync or server restart.

LazyCodex has an additional cold-start symptom: its SessionStart hook schedules CodeGraph initialization in a detached background worker, so Codex can discover MCP tools before initialization finishes. OpenCode refreshes MCP tools before the actual model request, so that race did not reproduce there, but the secondary-repository freshness problem still applies.

### Steps to Reproduce

1. Enable CodeGraph and start the agent in repository A.
2. Ask the agent to inspect repository B using a CodeGraph tool's `projectPath`.
3. Modify a symbol in repository B.
4. Repeat the same CodeGraph query without explicitly running `codegraph sync` for repository B.
5. Observe that the result can remain stale.

LazyCodex-specific cold-start variant:

1. Start a fresh Codex session in a repository that has not been indexed yet.
2. Observe: `LazyCodex CodeGraph bootstrap scheduled in background`.
3. Immediately ask the agent to use CodeGraph.
4. Tool discovery may return no `mcp__codegraph__*` tools even though the detached bootstrap finishes afterward.

### Expected Behavior

CodeGraph should select and maintain the graph for each repository the agent accesses.

Queries against repository B should reflect repository B's latest files without requiring a manual sync. A new session should not advertise or discover CodeGraph until its initial repository is ready.

### Actual Behavior

Both editions can query another repository through `projectPath`, but the secondary graph is cached without a file watcher. Subsequent queries can return stale results until explicit synchronization or an MCP restart.

LazyCodex additionally initializes or syncs the initial graph in a detached SessionStart worker. Codex can cache an empty tool list before that worker completes. In isolated OpenCode testing, both pre-indexed and initially unindexed workspaces supplied all four CodeGraph tools to the actual model request, so OpenCode does not share that particular cold-start symptom.

### Doctor Output

```shell
oMoMoMoMo Doctor

✓ System OK (opencode 1.17.11 · oh-my-openagent 4.13.0)
```

### Error Logs

```shell
```

### Configuration

```json
{
  "codegraph": {
    "enabled": true,
    "auto_init": true
  }
}
```

### Operating System

macOS

### OpenCode Version

1.17.11

matchai:
## Summary

Moves CodeGraph multi-repository freshness into the shared MCP proxy used by OpenCode and Codex, while preserving the Codex framing and project-root behavior introduced upstream in #5589.

Explicit secondary `projectPath` calls are refreshed before forwarding, stale calls fail closed, and secondary paths are never auto-initialized. Initial-session auto-init remains configurable, and already indexed sibling repositories continue to sync and query normally.

## Changes

- Adds the shared `packages/codegraph-mcp` proxy and keeps CodeGraph pinned to 1.1.1.
- Preserves newline and `Content-Length` client framing while speaking newline JSON-RPC to CodeGraph.
- Waits for initial repository readiness before exposing the Codex MCP surface.
- Refreshes indexed secondary repositories before forwarding tool calls.
- Returns correlated `-32001` errors without forwarding when refresh throws or reports incomplete work.
- Restricts auto-init to the initial session repository; explicit secondary paths must already be indexed.
- Honors `codegraph.auto_init` from Codex config and environment overrides.
- Reuses the shared proxy from both adapters and removes the duplicate Codex-local bridge.
- Preserves configured OpenCode `codegraph.install_dir` and keeps provisioned CodeGraph commands enabled even when host Node is unsupported.
- Fails safely when packaged CodeGraph CLI invocation paths are missing instead of throwing an uncaught startup error.
- Prevents Claude/project `.mcp.json` entries from shadowing OMO built-in MCP names, so the repo's tracked direct `codegraph serve --mcp` entry cannot override the shared proxy.
- Aligns OpenCode bootstrap with the shared provisioned-launcher policy and `.locks` provision lock path.
- Keeps the Codex CodeGraph MCP discoverable by default; the component wrapper handles provisioning and runtime compatibility.
- Regenerates the Codex CodeGraph and installer bundles from reconciled sources.

## Verification

Evidence on current pushed head `3594097c8` includes:

- Real CodeGraph 1.1.1 multi-repository freshness over newline and framed transports.
- Failure-path proof for thrown and incomplete refreshes with zero stale forwarding.
- Isolated OpenCode MCP connection with unchanged real session database and no leaked daemon.
- Isolated OpenCode proof that custom `codegraph.install_dir` is passed through to the provisioned CodeGraph command, telemetry is disabled, no project `.mcp.json` is written, and the `codegraph` MCP reports `connected`.
- Live OpenCode root-project proof with the tracked `.mcp.json` present: built-in CodeGraph uses `packages/codegraph-mcp/dist/cli.js mcp`, custom install dir is preserved, `/mcp` reports connected, and the real OpenCode DB is unchanged.
- Isolated Codex install and app-server hook execution with unchanged real `~/.codex/config.toml`.
- Isolated Codex installer proof that `plugins."omo@sisyphuslabs".mcp_servers.codegraph.enabled = true`.
- First-party Codex `mcpServerStatus/list` discovery of the `codegraph` server and all four CodeGraph tools.
- `bun run build` and `bun run typecheck` pass.
- `bun run test:codex` passes 411/411 with telemetry opt-out variables unset after the exact-shell run exposed inherited telemetry environment contamination and a non-repeating Git Bash preflight timeout.
- `bun test` passes 10,255 tests with 2 existing live-tmux skips and telemetry opt-out variables unset.
- Deterministic generated bundles, scope/ancestry, conflict, suppression, and changed-file size audits pass.
- Evidence password redaction checks pass for the previously leaked OpenCode QA password.

Reviewer-readable artifacts are under `.omo/evidence/20260626-pr-5592-upstream-reconciliation/`.

## Draft status

All local remediation and final harness/full-suite gates are complete. The branch remains draft while GitHub CI and the final independent review lanes rerun on the refreshed head.

## Risks

- Cross-repository queries remain supported only for repositories with existing `.codegraph` state; an explicit secondary path cannot create new indexing state.
- Exact local full-suite runs inherit telemetry opt-out variables from the development shell; evidence preserves those failures and the corresponding sanitized reruns, including the one non-repeating Git Bash preflight timeout seen in the contaminated `test:codex` run.
- Homebrew `tmux 3.6b` was installed during isolated Codex QA to satisfy the harness prerequisite.
- Claude/project `.mcp.json` entries with the same name as OMO built-in MCPs are now ignored in favor of the built-in entry; explicit OpenCode user config still participates in the normal config merge/disable path.
- Landing must use a merge commit and requires explicit approval after all final gates pass.

## Related

- Fixes #5588.
- Preserves the Codex framing/cwd work from #5589.


code-yeongyu:
Thank you for this @matchai. The multi-repo freshness analysis here was genuinely useful and shaped how we reasoned about the daemon path.

We ended up taking a different route that supersedes this PR: rather than introduce a separate shared proxy package, we (1) upgraded to CodeGraph 1.4.1 (#6242), whose engine re-checks which index a path belongs to on each call (upstream 1.3.0), and (2) re-enabled the upstream multi-client daemon behind an opt-in `codegraph.daemon` flag (#6251), so N clients share one daemon/watcher/SQLite per project without a bespoke proxy layer. The zombie sweep was also made daemon-aware so a live detached daemon is never mis-reaped.

Closing in favor of that approach. If a multi-repo freshness gap remains under 1.4.1 + daemon mode, please reopen with a repro and we will dig in.

PR #5656, opened 2026-06-28
Fix stale background task cleanup

coleleavitt:
## Summary
- route stale background task interruption through manager-owned terminal cleanup
- clear root descendant, delegated child bootstrap, session category, and child session bookkeeping for stale cancellations
- log selected fire-and-forget failures instead of silently swallowing them
- harden archive extraction/listing argv handling for zip/tar tools

## Verification
- `bun test packages/omo-opencode/src/features/background-agent/manager.test.ts -t "should clean stale-interrupted child session bookkeeping"`
- `bun test packages/omo-opencode/src/features/background-agent/manager.test.ts packages/omo-opencode/src/features/background-agent/task-poller.test.ts` (234 pass, 0 fail)
- `bash .agents/skills/opencode-qa/scripts/lib/common.sh --self-check`
- `bash .agents/skills/opencode-qa/scripts/server-smoke.sh`
- `bash .agents/skills/opencode-qa/scripts/sse-hook-probe.sh --self-test`
- throwaway `unzip -o -d <dest> -- <archive>` and `unzip -p -- <archive> <entry>` checks

## Notes
- `.omo/evidence` artifacts were recorded locally during QA but are intentionally not included because `.omo` is ignored local agent workspace.
- Local `bun run build` was attempted but blocked by unrelated unstaged working-tree deletions of Codex comment-checker files (`packages/omo-codex/plugin/components/comment-checker/src/core.ts` and test). Those deletions are not part of this PR.


packages/omo-opencode/src/features/background-agent/manager.ts
```diff
@@ -2317,6 +2317,62 @@ The task was re-queued on a fallback model after a retryable failure.
...
+    const sessionID = task.sessionId
+    if (!sessionID) return
+
+    subagentSessions.delete(sessionID)
+    clearSessionAgent(sessionID)
+    clearDelegatedChildSessionBootstrap(sessionID)
+    SessionCategoryRegistry.remove(sessionID)
+    this.clearSessionOutputObserved(sessionID)
+    this.clearSessionTodoObservation(sessionID)
+  }
```
code-yeongyu, line 2355: [sisyphus-bot] Background cleanup blocker: this stale-interruption cleanup only clears the child bootstrap/category/agent/output/todo subset. Launch/resume also stores per-session prompt params and session tools for the same child session ids (`applySessionPromptParams` / `setSessionTools`), and `registerDelegatedChildSessionBootstrap` also seeds model-fallback chain state. Because this path is specifically compensating for unreliable `session.deleted`, those stores can remain stale after a final stale cancellation. Please clear the same child-session-owned stores here and add assertions for prompt params, session tools, and fallback-chain cleanup.

packages/utils/src/zip-entry-listing/read-zip-symlink-target.ts
```diff
@@ -5,7 +5,7 @@ export async function readZipSymlinkTarget(
 	archivePath: string,
 	entryPath: string
 ): Promise<string | undefined> {
-	const proc = spawn(["unzip", "-p", archivePath, "--", entryPath], {
+	const proc = spawn(["unzip", "-p", "--", archivePath, entryPath], {
```
code-yeongyu, line 8: [sisyphus-bot] Archive security blocker: `--` protects the archive path from option parsing, but Info-ZIP still treats the member operand as a wildcard pattern. I reproduced `unzip -p -- wild.zip 'bin/*'` returning concatenated bytes from a regular file and a symlink target, so `validateArchiveEntries` can validate a different target than the symlink that extraction creates. Please use an exact-entry reader or escape Info-ZIP wildcard metacharacters, and add a regression with a symlink entry name containing wildcard characters.

packages/omo-opencode/src/shared/binary-downloader.ts
```diff
@@ -101,7 +101,7 @@ function parseTarEntry(line: string): ArchiveEntry | null {
 }
 
 async function listTarEntries(archivePath: string, cwd?: string): Promise<ArchiveEntry[]> {
-  const proc = spawn(["tar", "-tvzf", archivePath], {
+  const proc = spawn(["tar", "-t", "-z", "-f", archivePath], {
```
code-yeongyu, line 104: [sisyphus-bot] Archive validation regression: `parseTarEntry` still expects verbose `tar -tv...` rows, but this now runs `tar -t -z -f`, which emits bare names like `file.txt`. Those lines parse as `null` and are silently filtered, so normal tarball validation can receive an empty entry list before extraction. Please restore verbose listing or update the parser/fail-closed behavior, with regression coverage for normal entries plus traversal/link cases.

code-yeongyu (changes requested):
[sisyphus-bot] Requesting changes after part-by-part review.

Background cleanup: stale-interrupted child sessions need to clear all child-session stores, not only the subset covered by the new helper.

Archive argv / validation: the `zipinfo` fallback still reads symlink targets through wildcard-capable Info-ZIP member operands, and the tar preflight listing now emits bare names while the parser still expects verbose rows.

Hook/session flow: reviewed separately and no blocker found in that slice.

Please fix the blockers below and include regression coverage/evidence for stale store cleanup, wildcard symlink member names, and tar preflight validation.

coleleavitt:
This stale-cleanup PR is now included in the combined hardening PR #5661.

The requested blockers from this review are addressed on #5661 in `f61ab6c01`:
- child cleanup clears prompt params, session tools, and session fallback-chain state
- zip symlink target reads escape Info-ZIP wildcard member names
- tar preflight uses verbose parsing and fails closed, with symlink traversal coverage

I have not closed this PR from here; #5661 is the branch carrying the review-fix commit plus the broader hardening work.

code-yeongyu:
[sisyphus-bot] code-yeongyu's review identified concrete blockers (incomplete child-session cleanup, zipinfo wildcard issues, tar preflight parsing). The author indicated these are addressed in PR #5661, which is also still open. This PR is stale (36 days) and the touched files have been significantly refactored on dev since. Recommend: close this in favor of #5661, or rebase and address the review blockers directly here.

PR #5662, opened 2026-06-28
fix(sisyphus): cache rebuilt prompt per runtime-model to stop per-turn rebuild (#5578)

### LINKED ISSUE #5578: Sisyphus runtime prompt reconciler (fix for #5297) swaps whole prompt on every turn — agent behavior inverts between adjacent turns
## Summary

The fix for #5297 (PR #5297 + PR #5359 v2) introduced `reconcileSisyphusRuntimePrompt` in `agents/sisyphus-runtime-prompt-reconciler.d.ts`, which rebuilds the **entire** baked Sisyphus prompt when the runtime model family differs from the configured model family. This reconciler is invoked as a per-request system-transform hook, so it runs on every turn — and produces a **different prompt body on every turn** even when the runtime model is stable. The result is that Sisyphus behaves inconsistently within a single session: in the same conversation, a few seconds apart, the agent can first say "I will not proceed without your decision" and then, on the next turn, decide on its own to proceed with the recommended option.

## Evidence

`agents/sisyphus-runtime-prompt-reconciler.d.ts` in `oh-my-openagent@4.13.0`:

```ts
export type SisyphusRuntimePromptContext = {
    configuredModel: string;
    bakedPrompt: string;
    rebuildPromptForModel: (runtimeModel: string) => string;
};

export declare function setSisyphusRuntimePromptContext(ctx: SisyphusRuntimePromptContext): void;
export declare function clearSisyphusRuntimePromptContext(): void;

/**
 * The Sisyphus prompt body is baked at registration from the *configured* model
 * in `oh-my-openagent.jsonc`. When the user switches to a different model family
 * in the TUI, the entire baked body is the wrong family for the runtime model
 * (issue #5297/#5316): a GPT-configured agent run on a non-GPT model still
 * carries the whole GPT-5.5 body, not just one apply_patch line.
 *
 * The system-transform hook is the only per-request seam that knows the runtime
 * model, so rebuild the whole prompt for the runtime family and swap it in here
 * rather than patching individual family-specific lines (which can never convert
 * a GPT body into a non-GPT one).
 *
 * Returns true if a swap was performed.
 */
export declare function reconcileSisyphusRuntimePrompt(system: string[], runtimeModel: string | undefined): boolean;
```

Two design properties combine to cause the regression:

1. **Per-turn invocation.** The system-transform hook is called on every request, and `reconcileSisyphusRuntimePrompt` always rebuilds when `runtimeModel` differs from the baked model — not just on actual user-initiated model switches.
2. **Whole-prompt swap.** The rebuild replaces the entire baked body, so the swapped prompt is not a small family-specific delta but a complete rewrite. Two consecutive turns therefore see materially different system prompts.

## Repro (observed in production session)

Configuration (`oh-my-openagent.json`):

```jsonc
"sisyphus": {
  "model": "github-copilot/gpt-5.4",      // configured
  "variant": "high",
  ...
}
```

Actual runtime model (model picker): `minimax-cn-coding-plan/MiniMax-M3` (a non-GPT family). The `oh-my-openagent` runtime config provides `minimax-cn-coding-plan/MiniMax-M3` as a non-GPT path, so this is a supported configuration; only the configured-vs-runtime mismatch triggers the reconciler.

Observed in the same session, ~15 seconds apart, on the same agent (Sisyphus / Ultraworker) and the same conversation:

**Turn 1** (Thought: 7.8s):
> I understand your choice. The system prompt tells me to continue, but you need to make a decision first — I should not act on my own before you choose A/B/C/D.
>
> Current status (progress on 4 todos)…
>
> Please reply with an option (A/B/C/D), or follow up with more specific details. I will not proceed on my own until you make a decision.

**Turn 2** (Thought: 15.5s):
> Understood. I will proceed with the recommended option A first.

The agent reversed its own "I will not proceed without your decision" commitment between two adjacent turns without any user input. The only structural change between these turns is the reconciler running and producing a different prompt body.

This reproduces with the same `oh-my-openagent.json` config and the same model picker selection, every session.

## Impact

- **Behavioral non-determinism within a session.** Sisyphus's commitments to wait for user input are not stable: turn 1 may say "I'll wait", turn 2 (without user input) may decide to proceed.
- **Undoing explicit agent decisions.** The agent can effectively veto its own earlier "wait for confirmation" output, which breaks trust in interactive work.
- **Affects any user who runs a non-GPT family model on a GPT-configured Sisyphus.** This is not an exotic configuration; the very existence of `minimax-cn-coding-plan`, `kimi-k2`, `gemini`, `glm-5-2`, etc. as runtime model options means many users hit this.

## Suggested fix

Stop running the reconciler on every turn. Two complementary changes:

1. **Cache the swapped prompt per session**, keyed by the runtime model. The swap should only re-execute when the runtime model actually changes (compare against the last reconciled model, not the configured model).
2. **Make the rebuild a delta, not a full swap.** The 4.13.0 reasoning was that a "GPT body cannot be converted into a non-GPT one" by patching individual lines — that is true for some sections (apply_patch, file-edit guards) but does not justify replacing the entire system prompt. Family-specific sections should be the only thing swapped; shared sections (role, constraints, output style) should be preserved verbatim.

Pseudocode:

```ts
let lastReconciledModel: string | undefined;
let cachedPrompt: string | undefined;

export function reconcileSisyphusRuntimePrompt(system, runtimeModel) {
  if (runtimeModel === lastReconciledModel && cachedPrompt) {
    // No change — return cached prompt without rebuild
    return false;
  }
  const rebuilt = rebuildPromptForModel(runtimeModel);
  swapSystemEntry(system, ctx.bakedPrompt, rebuilt);
  lastReconciledModel = runtimeModel;
  cachedPrompt = rebuilt;
  return true;
}
```

The "shared sections" extraction would need a structured prompt template (sections labeled with stable IDs), which is a separate refactor — but the per-session caching alone is sufficient to eliminate the regression reported here.

## Environment

- `oh-my-openagent@4.13.0` (installed via `opencode plugin oh-my-openagent@4.13.0 --force --global`)
- `opencode` 1.17.9 (Homebrew)
- macOS 26.4.1, Darwin 25.4.0, arm64
- `sisyphus.model = github-copilot/gpt-5.4` configured; runtime = `minimax-cn-coding-plan/MiniMax-M3`
- Observed across multiple sessions after the 4.13.0 upgrade; not seen on 4.8.1 (the pre-reconciler version)

Related:
- #5297 (closed) — the original bug ("Sisyphus prompt family is determined by configured model instead of runtime-selected model") that PR #5297 / PR #5359 attempted to fix. This new issue is the regression introduced by that fix.
- PR #5297: `fix(sisyphus): rebuild whole prompt for runtime model family`
- PR #5316: revert of v1
- PR #5359: v2 re-merge `fix(sisyphus-runtime-prompt-family-v2`

JSap0914:
## Bug

Closes #5578

The `reconcileSisyphusRuntimePrompt` system-transform hook called `rebuildPromptForModel` on **every request** whenever the runtime model family differed from the configured model. Because `rebuildPromptForModel` reads live env/config state it is not guaranteed to produce identical output across calls — so two adjacent turns in the same session could receive materially different system prompts. This caused the behaviour reported in #5578: Sisyphus commits to waiting for the user on turn 1, then proceeds autonomously on turn 2 without any user input.

## Fix

Add two module-level variables (`lastReconciledModel` / `cachedRebuiltPrompt`) that persist the last rebuild result, keyed by the exact `runtimeModel` string:

- **Cache hit (stable model):** the cached rebuilt prompt is reused directly; `rebuildPromptForModel` is **not** called again.
- **Cache miss (model changed or first call):** `rebuildPromptForModel` is invoked, the result is cached, and the swap is applied.
- **Session end:** `clearSisyphusRuntimePromptContext` now also clears both cache variables so a fresh session starts clean.

The substring swap into the system array still executes on every turn so that a freshly-assembled system array always receives the correct family body.

## Verification

Two new unit tests added to `packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-reconciler.test.ts`:

| Test | Before fix | After fix |
|---|---|---|
| Stable runtime model: `rebuildPromptForModel` call count across 2 turns | **2 (FAIL)** | 1 (PASS) |
| Model switch between turns: rebuild fired for new model | PASS | PASS |

All 8 tests in the file pass; the 6 broader pre-existing failures are in `git-master-template-injection.test.ts` and unrelated to this change.

```
bun test packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-reconciler.test.ts
 8 pass, 0 fail
```

## Changed files

- `packages/omo-opencode/src/agents/sisyphus-runtime-prompt-reconciler.ts` — cache logic + updated JSDoc
- `packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-reconciler.test.ts` — 2 new tests for #5578

---
## Summary by cubic
Fixes inconsistent Sisyphus system prompts by caching the rebuilt prompt per runtime model and clearing it at session end. Prevents per-turn rebuilds and stops behavior drift (fixes #5578).

- **Bug Fixes**
  - Cache the rebuilt system prompt keyed by the exact `runtimeModel` to avoid calling `rebuildPromptForModel` each turn.
  - Invalidate the cache when the model changes or `clearSisyphusRuntimePromptContext` runs; still perform the substring swap every turn.
  - Added tests to verify one rebuild for a stable model and a new rebuild on model switch.

<sup>Written for commit 8ea7e58fe7db4fe735f089c9ef871731f87eb336. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/agents/sisyphus-runtime-prompt-reconciler.ts
+++ b/packages/omo-opencode/src/agents/sisyphus-runtime-prompt-reconciler.ts
@@ -17,12 +17,20 @@ export type SisyphusRuntimePromptContext = {
 
 let context: SisyphusRuntimePromptContext | undefined;
 
+// Per-session cache: avoid calling rebuildPromptForModel on every turn when the
+// runtime model is stable. The cache is keyed by the exact runtimeModel string
+// and is cleared together with `context` at session end.
+let lastReconciledModel: string | undefined;
+let cachedRebuiltPrompt: string | undefined;
+
 export function setSisyphusRuntimePromptContext(ctx: SisyphusRuntimePromptContext): void {
   context = ctx;
 }
 
 export function clearSisyphusRuntimePromptContext(): void {
   context = undefined;
+  lastReconciledModel = undefined;
+  cachedRebuiltPrompt = undefined;
 }
 
 /**
@@ -36,6 +44,12 @@ export function clearSisyphusRuntimePromptContext(): void {
  * model, so rebuild the whole prompt for the runtime family and swap it in here
  * rather than patching individual family-specific lines (which can never convert
  * a GPT body into a non-GPT one).
+ *
+ * Rebuilding is expensive and non-deterministic across calls (it reads live env
+ * and config state), so the result is cached per runtime-model string. The cache
+ * is invalidated when the runtime model changes or the session context is cleared
+ * (fix for #5578: reconciler was rebuilding on every turn causing behavioural
+ * non-determinism within a single session).
  *
  * Returns true if a swap was performed.
  */
@@ -53,7 +67,19 @@ export function reconcileSisyphusRuntimePrompt(
     return false
   }
 
-  const rebuilt = context.rebuildPromptForModel(runtimeModel)
+  // Use the cached rebuilt prompt when the runtime model has not changed since
+  // the last reconciliation. rebuildPromptForModel reads live env/config state
+  // and is therefore not guaranteed to be pure; calling it on every turn can
+  // produce subtly different prompts even with the same model id (#5578).
+  let rebuilt: string;
+  if (runtimeModel === lastReconciledModel && cachedRebuiltPrompt !== undefined) {
+    rebuilt = cachedRebuiltPrompt;
+  } else {
+    rebuilt = context.rebuildPromptForModel(runtimeModel);
+    lastReconciledModel = runtimeModel;
+    cachedRebuiltPrompt = rebuilt;
+  }
+
   if (rebuilt === context.bakedPrompt) return false
 
   // Substring replace rather than exact-equality: opencode core may concatenate
--- a/packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-reconciler.test.ts
+++ b/packages/omo-opencode/src/plugin/sisyphus-runtime-prompt-reconciler.test.ts
@@ -95,3 +95,53 @@ describe("Sisyphus runtime prompt family reconciliation (#5297/#5316)", () => {
     expect(output.system[0]).not.toContain(GPT_APPLY_PATCH_GUIDANCE)
   })
 })
+
+describe("Sisyphus runtime prompt reconciler — per-turn rebuild cache (#5578)", () => {
+  test("#given a stable runtime model #when reconcile runs on consecutive turns #then rebuildPromptForModel is called only once", () => {
+    const baked = createSisyphusAgent(GPT_MODEL, [], [], [], []).prompt ?? ""
+    let rebuildCalls = 0
+    setSisyphusRuntimePromptContext({
+      configuredModel: GPT_MODEL,
+      bakedPrompt: baked,
+      rebuildPromptForModel: (model) => {
+        rebuildCalls++
+        return createSisyphusAgent(model, [], [], [], []).prompt ?? ""
+      },
+    })
+
+    // Turn 1: first reconciliation with a non-GPT model
+    const system1 = [baked]
+    const swapped1 = reconcileSisyphusRuntimePrompt(system1, NON_GPT_MODEL)
+    expect(swapped1).toBe(true)
+    expect(rebuildCalls).toBe(1)
+
+    // Turn 2: same stable runtime model — rebuild must be skipped, swap still applied
+    const system2 = [baked]
+    const swapped2 = reconcileSisyphusRuntimePrompt(system2, NON_GPT_MODEL)
+    expect(swapped2).toBe(true)
+    expect(rebuildCalls).toBe(1) // must NOT have incremented
+    expect(system2[0]).not.toContain("based on GPT-5.5")
+  })
+
+  test("#given a model switch between turns #when reconcile runs #then rebuildPromptForModel is called again for the new model", () => {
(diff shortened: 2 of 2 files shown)


MoerAI:
Thanks for investigating the repeated Sisyphus prompt reconciliation path.

Closing this implementation, not #5578:
https://github.com/code-yeongyu/oh-my-openagent/issues/5578

The branch conflicts with current `dev`, and its module-global cache is not invalidated when registration replaces the captured prompt context. Current architecture rebuilds against a freshly assembled system array per request; #6998 codified and tested that behavior:
https://github.com/code-yeongyu/oh-my-openagent/pull/6998

The added tests prove only a call-count optimization. They do not reproduce divergent model-visible system arrays through the current OpenCode hook or cover context replacement with an unchanged model.

Please keep #5578 focused on a current-`dev` reproducer that captures divergent system arrays under identical inputs.

PR #5665, opened 2026-06-28
fix(delegate): fail fast on unresolvable model for agents requiring a model (#5604)

### LINKED ISSUE #5604: [Bug]: OpenCode delegation harness should fail fast on invalid subagent routing instead of silently falling back
## Summary

We hit a case where the OpenCode OMO delegation flow appeared to skip several planned subagent tasks. After inspecting the parent session logs, this did not look like a stuck subagent or a UI-only visibility issue. Instead, the harness allowed an invalid subagent routing attempt to create child-session artifacts without a usable model/stream, and the parent agent then continued doing work directly.

This made the run look like “subagents were called but did not respond” or “some subagents were never launched,” when the underlying issue was that invalid delegation was not surfaced as a hard failure.

## Observed behavior

- A multi-task OpenCode run planned several subagent-delegated tasks.
- Some delegation attempts created child-session-like records, but those sessions had no usable model assignment and no stream events.
- Later planned tasks had no corresponding child-session creation events at all.
- The parent/orchestrator agent continued using direct tools and marked work as completed, even though the expected subagent dispatch path had not happened.
- A later task dispatched through a valid category path did create and run a normal subagent session, which suggests the subagent system itself was available; the failure was in validation/enforcement around routing.

## Why this is confusing

From the user perspective, this presents as one of the following:

- “The subagent was called but is not visible.”
- “The subagent is stuck.”
- “Some planned subagents were skipped.”
- “The parent agent ignored the orchestration contract and implemented directly.”

The logs can contain partial session-creation evidence, but no actual model stream, so it is easy to misdiagnose as a long-running child session or UI issue.

## Expected behavior

The harness should fail fast and visibly when a delegation request is invalid or cannot be routed to a runnable model.

Suggested hardening:

- Validate subagent routing before creating a child session.
- Reject invalid category/skill combinations with a clear error returned to the parent.
- Treat child sessions with `model=undefined`, missing provider/model metadata, or zero stream events as failed delegations, not as silent/no-op sessions.
- Prevent orchestrator-style parent agents from continuing direct implementation when the current workflow contract requires subagent delegation.
- Surface “delegation failed before stream start” distinctly in logs/UI instead of making it look like a hidden or stuck subagent.
- If a task/blocker system is active, avoid letting continuation prompts push past a blocked or failed delegation without an explicit recovery decision.

## Impact

This makes multi-agent runs hard to trust because the visible plan can say tasks were delegated while the runtime actually performs a mixture of failed delegation attempts and direct parent execution. Stronger validation would make these failures immediately actionable and avoid confusing them with Bifrost/MCP issues, UI hiding, or slow subagent execution.

## Acceptance criteria

- [ ] Invalid delegation routing fails before child-session creation, or the child session is immediately marked failed with a clear reason.
- [ ] Parent sessions receive a structured failure result when a subagent cannot be routed to a model.
- [ ] OpenCode logs/UI distinguish: not dispatched, dispatch failed before stream, running, completed, and failed after stream.
- [ ] Orchestrator workflows can enforce “delegate, do not directly implement” when that is part of the workflow contract.

JSap0914:
Closes #5604

## Bug

When  exhausts all fallback-chain providers (all disconnected) and returns ,  silently set  and returned success. The caller in  then proceeded to call  /  with no model, creating child-session records with no usable model or stream — appearing to the parent as hidden, skipped, or stuck subagents rather than a clear delegation failure.

## Fix

- Add  tracking in  to distinguish the definitive-failure case from the cold-cache skip sentinel.
- After all fallback attempts: when resolution was definitively  **and** the agent has , return a structured  (new  field) with a clear message listing the disconnected providers.
- Propagate the error in  **before** child-session creation, so  returns it to the parent instead of spawning an un-routable session.
- The  sentinel path (cold cache) is unaffected — cold-cache delegation still proceeds without pinning a model.

## Verification

New test file  (4 tests):

| Scenario | Before | After |
|---|---|---|
| hephaestus + only disconnected provider (xai) |  ❌ |  set,  ✅ |
| sisyphus + only disconnected provider (xai) |  ❌ |  set ✅ |
| hephaestus + openai connected (in chain) | no error ✅ | no error ✅ |
| hephaestus + cold cache (no connectivity) | no error ✅ | no error ✅ |

RED→GREEN confirmed: 2 tests fail on , all 4 pass with this fix.

Full delegate-task suite: **479 pass, 0 fail**.

---
## Summary by cubic
Fails fast when a required subagent model can't be resolved (all providers disconnected), returning a clear error to the parent instead of creating a child session with no model. Addresses #5604 and prevents hidden/skipped subagents; cold-cache behavior is unchanged.

- **Bug Fixes**
  - Added `error` to `ResolvedSubagentModel` and set it in `resolveSubagentModel` when resolution returns undefined and `requiresAnyModel` is true; message lists the required providers.
  - In `resolveSubagentExecution`, propagate the model error and skip child-session creation.
  - Kept the cold-cache skip path unchanged (no error; no model pinned).
  - Added regression tests for disconnected-only providers, a matching connected provider, and cold-cache scenarios.

<sup>Written for commit 13e8b5de4cd7785583954ca0c0d3f7940814498a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5665?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for addressing the unresolved-model path. The confirmed blocker is the mandatory QA gate: this changes `packages/omo-opencode` but commits no evidence under `.omo/evidence/`, and unit tests alone do not satisfy repository policy. Please run the real QA harness, commit the observed evidence under `.omo/evidence/<date>-<slug>/`, and include proof that an unresolved required model fails before child-session creation while valid override, cached, and cold-cache paths remain unchanged. https://github.com/code-yeongyu/oh-my-openagent/pull/5665/files

PR #5753, opened 2026-06-30
fix(omo-codex): fail on unreadable config.toml instead of overwriting it (fixes #5751)

### LINKED ISSUE #5751: [omo-codex] Installer overwrites unreadable config.toml instead of failing
## Summary

The omo-codex installer can overwrite an existing `config.toml` when the file exists but cannot be read.

This was first filed against the generated LazyCodex mirror as code-yeongyu/lazycodex#95, but the implementation lives in this repository under `packages/omo-codex`, so this issue tracks the source fix here.

## Environment

- Repository: `code-yeongyu/oh-my-openagent`
- Checked source: `dev` at `63b0faf`
- Runtime used for reproduction: Bun `1.3.14`, Node `v24.18.0`
- OS: macOS `26.5.1` arm64
- Relevant component: `packages/omo-codex/src/install`

## Repository Decision

- Target repository: `code-yeongyu/oh-my-openagent`
- Why this belongs there: the affected installer implementation is in `packages/omo-codex/src/install/codex-config-toml.ts`.
- LazyCodex evidence: the generated mirror issue is code-yeongyu/lazycodex#95.
- Upstream Codex source evidence: `openai/codex` `main` at `4808c16` does not contain `updateCodexConfig`, `codex-config-toml`, or this ULW/omo installer code.

## Reproduction

```bash
tmp=$(mktemp -d)
cfg="$tmp/config.toml"
printf '%s\n' '[user]' 'important = "keep"' > "$cfg"
chmod 200 "$cfg"

CFG="$cfg" bun -e 'import { updateCodexConfig } from "/tmp/oh-my-openagent-source/packages/omo-codex/src/install/codex-config-toml.ts"; await updateCodexConfig({ configPath: process.env.CFG, repoRoot: "/tmp/oh-my-openagent-source", marketplaceName: "sisyphuslabs", marketplaceSource: { sourceType: "git", source: "https://github.com/code-yeongyu/oh-my-openagent", ref: "dev" }, pluginNames: ["omo"], agentConfigs: [] });'
echo "exit_code=$?"
chmod 600 "$cfg"
sed -n '1,120p' "$cfg"
```

## Expected Behavior

If `config.toml` exists but cannot be read, the installer should fail before writing anything. Only `ENOENT` should be treated as "no config exists yet."

## Actual Behavior

The command exits successfully:

```text
exit_code=0
```

The resulting `config.toml` contains generated omo-codex config, and the original marker is gone:

```toml
[user]
important = "keep"
```

## Evidence

- `packages/omo-codex/src/install/codex-config-toml.ts`
  - `updateCodexConfig()` initializes `config = ""`.
  - It reads existing config only when `exists(configPath)` returns true.
  - `exists()` catches every `readFile()` error and returns false.
- `packages/omo-codex/src/install/codex-config-atomic-write.ts`
  - `writeFileAtomic()` writes a temp file and renames it over the target path.

Observed reproduction output from OMO source:

```text
exit_code=0
model = "gpt-5.5"
...
[marketplaces.sisyphuslabs]
source_type = "git"
source = "https://github.com/code-yeongyu/oh-my-openagent"
ref = "dev"
```

## Root Cause

`exists()` conflates "file is absent" with "file exists but `readFile()` failed." When the failure is a permission/readability problem, `updateCodexConfig()` proceeds from an empty config and `writeFileAtomic()` can replace the existing file if the parent directory is writable.

## Proposed Fix

- Replace the current `exists()` read probe with direct `readFile()`.
- Catch only `ENOENT` and use an empty config in that case.
- Re-throw `EACCES`, `EPERM`, parse/read failures, and other unexpected errors before any write.
- Add a regression test where `config.toml` exists but is unreadable and assert that:
  - `updateCodexConfig()` rejects.
  - the original file content is preserved.

## Verification Plan

- Re-run the reproduction above and verify it exits non-zero.
- Verify the original `config.toml` still contains `[user] important = "keep"`.
- Run the existing omo-codex installer/config test suite.

---
Tag: lazycodex-generated

MoerAI:
## Summary

The omo-codex installer could **overwrite an existing `config.toml` when the file exists but cannot be read**. `updateCodexConfig()` now reads the config directly and only treats a genuinely-absent file (`ENOENT`) as empty; any other read error is re-thrown before a single byte is written.

Fixes #5751.

## Root Cause

`updateCodexConfig()` probed for an existing config with a local `exists()` helper:

```ts
let config = ""
if (await exists(input.configPath)) config = await readFile(input.configPath, "utf8")
// ...
async function exists(path: string): Promise<boolean> {
  try { await readFile(path, "utf8"); return true }
  catch (error) { if (error instanceof Error) return false; return false }
}
```

`exists()` swallowed **every** `readFile()` error and returned `false`. A present-but-unreadable `config.toml` (`EACCES`/`EPERM`, or any non-`ENOENT` read failure) was therefore treated as absent, so the function started from an empty config and `writeFileAtomic()` replaced the user's file, dropping its contents.

## Changes

| File | Change |
|------|--------|
| `packages/omo-codex/src/install/codex-config-toml.ts` | Read the config directly; catch only `ENOENT` (use empty config), re-throw all other read errors before any write. Remove the now-unused `exists()` helper. Reuses the existing `isNodeErrorWithCode` guard from `codex-cache-fs`. |
| `packages/omo-codex/src/install/codex-config-toml.test.ts` | Two regression tests: a cross-platform case (an unreadable path surfaces the read error instead of overwriting) and a POSIX `chmod 0o200` case asserting the original file is preserved. |

Net source change: `+6 / -9` (one import, the read guard, minus the dead helper).

## Reproduction (before fix)

The new test against the unmodified code shows the installer reaches `writeFileAtomic` and **attempts to overwrite** the existing path:

```
).rejects.toThrow("EISDIR")
error: expect(received).toThrow(expected)
Expected substring: "EISDIR"
Received message: "EPERM: operation not permitted, rename
'...\.tmp-config.toml-7968-...' -> '...\config.toml'"
(fail) codex-config-toml > ...surfaces the read error instead of overwriting from an empty config
```

`chmod` cannot make a file unreadable on Windows, so the unreadable case is exercised cross-platform via a directory `config.toml` (so `readFile` throws `EISDIR`), plus a POSIX `chmod 0o200` test that is skipped on Windows.

## Verification (after fix)

```
(pass) codex-config-toml > ...surfaces the read error instead of overwriting from an empty config
(skip) codex-config-toml > ...rejects and preserves the original file   # POSIX-only chmod case
 22 pass
 1 skip
 0 fail
Ran 23 tests across 1 file.
```

## Test

- Regression: `packages/omo-codex/src/install/codex-config-toml.test.ts` (2 new cases) - 22 pass / 1 skip (POSIX-only `chmod` case) / 0 fail
- Typecheck: `tsgo --noEmit -p packages/omo-codex/tsconfig.json` - clean (exit 0)

Fixes #5751

---
## Summary by cubic
Prevents the `omo-codex` installer from overwriting an existing `config.toml` when the file exists but can’t be read. `updateCodexConfig()` now treats only `ENOENT` as “missing” and fails fast on other read errors. Fixes #5751.

- **Bug Fixes**
  - Read the config directly; catch only `ENOENT`, re-throw all other read errors before any write.
  - Removed the `exists()` probe that swallowed read errors and caused unintended overwrites.
  - Regenerated the bundled installer at `scripts/install-dist/install-local.mjs` so the fix is used when the dist bundle is present; the old `exists()` path is removed.
  - Added regression tests for unreadable paths (cross‑platform) and POSIX `chmod` to ensure the original file is preserved.

<sup>Written for commit dc8ae70aba6536eafb32808c8b0029b399d4b560. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5753?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-codex/scripts/install-dist/install-local.mjs
+++ b/packages/omo-codex/scripts/install-dist/install-local.mjs
@@ -8523,8 +8523,12 @@ function readBooleanSetting(sectionText, key) {
 async function updateCodexConfig(input) {
   await mkdir5(dirname6(input.configPath), { recursive: true });
   let config = "";
-  if (await exists(input.configPath))
+  try {
     config = await readFile10(input.configPath, "utf8");
+  } catch (error) {
+    if (!(isNodeErrorWithCode(error) && error.code === "ENOENT"))
+      throw error;
+  }
   const pluginSet = new Set(input.pluginNames);
   for (const legacyMarketplaceName of legacyMarketplaceNames(input.marketplaceName)) {
     config = removeMarketplaceBlock(config, legacyMarketplaceName);
@@ -8559,16 +8563,6 @@ async function updateCodexConfig(input) {
   await writeFileAtomic(input.configPath, `${config.trimEnd()}
 `);
 }
-async function exists(path) {
-  try {
-    await readFile10(path, "utf8");
-    return true;
-  } catch (error) {
-    if (error instanceof Error)
-      return false;
-    return false;
-  }
-}
 
 // packages/omo-codex/src/install/codex-hook-trust.ts
 import { createHash } from "node:crypto";
@@ -8588,7 +8582,7 @@ var EVENT_LABELS = new Map([
 ]);
 async function trustedHookStatesForPlugin(input) {
   const manifestPath = join17(input.pluginRoot, ".codex-plugin", "plugin.json");
-  if (!await exists2(manifestPath))
+  if (!await exists(manifestPath))
     return [];
   const manifest = JSON.parse(await readFile11(manifestPath, "utf8"));
   if (!isPlainRecord(manifest))
@@ -8596,7 +8590,7 @@ async function trustedHookStatesForPlugin(input) {
   const states = [];
   for (const hookPath of hookManifestPaths2(manifest.hooks)) {
     const hooksPath = join17(input.pluginRoot, hookPath);
-    if (!await exists2(hooksPath))
+    if (!await exists(hooksPath))
       continue;
     const parsed = JSON.parse(await readFile11(hooksPath, "utf8"));
     if (!isPlainRecord(parsed) || !isPlainRecord(parsed.hooks))
@@ -8679,7 +8673,7 @@ function canonicalJson(value) {
 function stripDotSlash2(value) {
   return value.startsWith("./") ? value.slice(2) : value;
 }
-async function exists2(path) {
+async function exists(path) {
   try {
     await readFile11(path, "utf8");
     return true;
@@ -8738,11 +8732,11 @@ var RETIRED_MANAGED_AGENT_FILES = [
 ];
 async function purgeRetiredManagedAgentFiles(input) {
   const agentsDir = join18(input.codexHome, "agents");
-  if (!await exists3(agentsDir))
+  if (!await exists2(agentsDir))
     return;
   for (const retiredAgent of RETIRED_MANAGED_AGENT_FILES) {
     const agentPath = join18(agentsDir, retiredAgent.fileName);
-    if (!await exists3(agentPath))
+    if (!await exists2(agentPath))
       continue;
     const agentStat = await lstat6(agentPath);
     if (agentStat.isDirectory() && !agentStat.isSymbolicLink())
@@ -8765,7 +8759,7 @@ async function readTextIfExists(path) {
     throw error;
   }
 }
-async function exists3(path) {
+async function exists2(path) {
   try {
     await lstat6(path);
     return true;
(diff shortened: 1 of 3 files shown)


MoerAI:

MoerAI:
Closing — superseded by maintainer PR #5860, which merged the bundled-installer-safe fix and isolated Codex QA for #5751. Happy to reopen or rework if this fix is still wanted.

PR #5792, opened 2026-07-01
fix(skill): suppress path-alias duplicates from injected description

code-yeongyu:
## Problem

When shared skills are enabled via `skills.paths`, the same skill can appear under both a qualified path name (e.g. `shared/debugging`) and a bare short name (`debugging`). Both variants were being injected into the skill tool description, causing ~3,500 tokens of redundant context per session.

Raised by @cpt9617 in the Sisyphus Labs Discord, with @typeey also confirming the direction.

## Root Cause

`formatCombinedDescription` received the full merged skill list without any path-alias deduplication. The description-level deduplication only checked exact name matches, so `debugging` and `shared/debugging` were treated as separate entries.

## Fix

Added `deduplicatePathAliasedSkills()` to `description-formatter.ts` which filters out any bare name that already has a qualified multi-segment counterpart in the list:

- `shared/debugging` is kept (canonical, qualified)
- `debugging` is suppressed (redundant alias)

The execution-time `matchSkillByName` in `skill-matcher.ts` already resolves bare short names via its `shortNameMatches` logic, so callers retain full invocation compatibility (`skill(name="debugging")` still works) without the description cost.

## What changes

- `src/tools/skill/description-formatter.ts` — new `deduplicatePathAliasedSkills()` helper, applied inside `formatCombinedDescription`
- `src/tools/skill/description-formatter.test.ts` — 7 new unit tests covering happy path, multiple duplicates, description drift, no-counterpart, and deeply nested paths

## Test results

```
50 pass, 0 fail
```

---
## Summary by cubic
Suppresses path-aliased duplicates in the skill tool description by keeping qualified names (e.g. `shared/debugging`) and dropping redundant bare aliases and shared‑derived builtin command aliases. Cuts tokens while preserving how skills and commands are invoked.

- **Bug Fixes**
  - Deduplicate skills by filtering bare names that have a qualified counterpart from the same source (via normalized location/`resolvedPath`).
  - Suppress builtin aliases `refactor`, `remove-ai-slops`, `start-work` only when a shared‑derived skill exists for that short name.
  - Preserve OpenCode native skill aliases (`config` scope with `<built-in>` or `/opencode/*`) and distinct project/user skills; keep builtin commands when the qualified skill isn’t shared‑derived.

<sup>Written for commit f078fd89effb107d545721f11b454bf03d58593b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5792?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/tools/skill/description-formatter.test.ts
+++ b/packages/omo-opencode/src/tools/skill/description-formatter.test.ts
@@ -36,6 +36,14 @@ function userSkill(name: string, description = "user desc"): SkillInfo {
   })
 }
 
+function opencodeNativeSkill(name: string, description = "opencode native desc"): SkillInfo {
+  const shortName = name.split("/").at(-1) ?? name
+  return makeSkill(name, description, {
+    scope: "config",
+    location: `/opencode/${shortName}.md`,
+  })
+}
+
 function makeCommand(
   name: string,
   description = "command desc",
@@ -268,5 +276,21 @@ describe("formatCombinedDescription with path-alias deduplication", () => {
     expect(result).toContain("\n    <name>/refactor</name>")
     expect(result).toContain("project refactor skill")
     expect(result).toContain("builtin refactor command")
+  })
+
+  it("keeps OpenCode-injected native skills while suppressing shared path aliases", () => {
+    const skills: SkillInfo[] = [
+      sharedSkill("debugging", "full shared debugging"),
+      builtinSharedSkill("debugging", "short debugging wrapper"),
+      opencodeNativeSkill("opencode/customize-opencode", "Qualified OpenCode customize entry"),
+      opencodeNativeSkill("customize-opencode", "Customize OpenCode"),
+    ]
+
+    const result = formatCombinedDescription(skills, [], { includeSkills: true })
+
+    expect(result).toContain("<name>/shared/debugging</name>")
+    expect(result).not.toContain("\n    <name>/debugging</name>")
+    expect(result).toContain("<name>/customize-opencode</name>")
+    expect(result).toContain("Customize OpenCode")
   })
 })
--- a/packages/omo-opencode/src/tools/skill/description-formatter.test.ts
+++ b/packages/omo-opencode/src/tools/skill/description-formatter.test.ts
@@ -1,9 +1,46 @@
 import { describe, expect, it } from "bun:test"
 import { deduplicatePathAliasedSkills, formatCombinedDescription } from "./description-formatter"
+import { loadedSkillToInfo } from "./native-skills"
+import type { CommandInfo } from "../slashcommand/types"
 import type { SkillInfo } from "./types"
 
-function makeSkill(name: string, description = "desc"): SkillInfo {
-  return { name, description, scope: "builtin" }
+function makeSkill(name: string, description = "desc", overrides: Partial<SkillInfo> = {}): SkillInfo {
+  return { name, description, scope: "builtin", ...overrides }
+}
+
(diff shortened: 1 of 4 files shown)


cpt-sd76vg:
@code-yeongyu 

I built and tested commit `26a2265` locally.

The issue from commit `cf9c45b`, where all `/shared/` entries disappeared, seems to be fixed now, and overall it looks much better. However, one issue still remains:

* `/customize-opencode` is missing.

This is an opencode-injected skill and not related to shared skills, so I think this is clearly a bug.

As for whether it was okay for `/foo` entries with different descriptions, like `/git-master`, to disappear or not, I can’t really judge that myself, so I’ll leave that decision to the maintainers.

cpt-sd76vg:
I built and tested commit `4af2dac` locally.

The situation seems unchanged from commit `26a2265`.

`/customize-opencode` is still missing, and everything else appears to behave the same as before. I didn’t see any new issues introduced.

code-yeongyu:
Status ping from Discord (#omo): @cpt9617 asked about merge timing.

Summary of remaining blocker per the latest local test reports:
- `26a2265` fixed the earlier regression where all `/shared/` entries disappeared.
- As of `4af2dac`, the shared-skills path-alias dedup works, but `/customize-opencode` (an opencode-injected skill, unrelated to shared skills) is still missing from the injected description — likely the dedup/filter pass is catching it as a false positive.

Suggest one more iteration to stop suppressing that built-in skill before merge; otherwise this trades ~3.5k tokens of duplicate context for a silently missing built-in skill.

code-yeongyu:
Post-merge field report from Discord (@cpt9617): tested `3f51596a` (after #5792 + #5831 merged) by capturing the actual JSON payload sent to the LLM provider through a proxy.

- `/customize-opencode` is **still missing** from the injected skill description — unchanged from `4af2dac`.
- No new issues observed elsewhere.

Note: the native-skill regression suite added in these PRs passes, yet the real provider payload still lacks the skill — so the current test fixture likely doesn't reproduce how `customize-opencode` gets injected at runtime (it's an opencode-runtime-injected skill, not one of omo's own skill files). The dedup/alias filter probably needs to be validated against a runtime-shaped skill list rather than the fixture list.

cpt9617 can retest quickly on any new commit — his proxy-capture setup checks the actual payload, which is the ground truth here.

cpt-sd76vg:
Sorry for the late follow-up. After extensive debugging, I found the issue was in my test environment, not in the PR.

I was testing by running OpenCode from source (`bun run --conditions=browser src/index.ts` — a local dev setup). It turns out that in this mode, the OpenCode host reads `tool.description` synchronously, before the async native-skill refresh (which injects `customize-opencode`) completes. So `customize-opencode` gets dropped from the cached description.

When I tested the same oMo build (commit `f078fd8`) against the npm-installed production OpenCode (v1.17.13), `/customize-opencode` appears correctly in the actual LLM payload — the production host reads `tool.description` after the async refresh finishes.

Confirmed via proxy-captured JSON payload in the production environment:
- `customize-opencode` is present
- shared-skill dedup works (no duplicate `/refactor`, `/remove-ai-slops`, `/start-work`)

The final commit (`f078fd8`) is working correctly. The "missing skill" was a dev-environment-only artifact. Sorry for the noise, and thanks for the quick iterations on this PR.

PR #5794, opened 2026-07-01
fix: add 15s timeout for stale active sessions and refine error classifier

KoishiID:
### Description
This PR addresses a major issue in the polling mechanism and refines the failure classification.

1. **Stale Active Session Timeout**: Added a 15-second timeout for sessions that briefly become active but fail to receive any messages. This prevents the poll loop from hanging for the full 5 minutes (`MAX_POLL_TIME_MS`) when a dispatch failure silently occurs.
2. **Refined Failure Classifier**: Removed the generic `"timed out"` check from `isAmbiguousPromptDispatchFailure` to avoid misclassifying deterministically timed-out requests.

---
## Summary by cubic
Add a 15s timeout when a session becomes active but no messages arrive, so the poller fails fast instead of hanging for up to 5 minutes. Also refine the prompt failure classifier to avoid treating deterministic timeouts as ambiguous.

- **Bug Fixes**
  - `packages/omo-opencode/src/tools/call-omo-agent/completion-poller.ts`: Track first active time; if no messages within 15s, throw a clear dispatch-failure error to avoid `MAX_POLL_TIME_MS` stalls.
  - `packages/utils/src/prompt-failure-classifier.ts`: Remove "timed out" from `isAmbiguousPromptDispatchFailure` to prevent misclassification of deterministic timeouts.

<sup>Written for commit 543dd38637358bac169646737da5d19aedd27932. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


KoishiID:

KoishiID:
recheck

KoishiID:

KoishiID:
recheck

code-yeongyu:

MoerAI:
Thanks for investigating the poller hang. The confirmed blocker is that `sawActiveStatusAt` is set only on the first active state and never refreshed, so a healthy session running over 15 seconds can hit one transient zero-message poll and be falsely terminated as a dispatch failure. Please measure time since the latest active-to-idle transition instead of first activation. Add deterministic poller tests with stubbed time proving a long-running healthy session survives transient empty reads and a genuinely stale idle session fails only after the intended 15-second interval. https://github.com/code-yeongyu/oh-my-openagent/pull/5794/files

PR #5830, opened 2026-07-03
fix(omo-codex): move the ultrawork directive into a plugin skill and inject a truncation-immune bootstrap pointer

### LINKED ISSUE #5828: LazyCodex/OmO Codex: ULW create_goal instruction lost when Codex App truncates hook output (regression since v4.14.1)
## Summary

When a user triggers `ulw` / `ultrawork` in Codex App via LazyCodex/OmO, the `UserPromptSubmit` hook injects `components/ultrawork/directive.md`. Since v4.14.1 the directive grew large enough that Codex App truncates the hook output, and the `create_goal` bootstrap instruction (`## 1. Create the goal with binding success criteria`, currently around line 75) falls into the omitted middle region, so the model never receives it and `create_goal` is silently skipped.

Reported independently by multiple users on Discord (#general-ko, #lazy-codex). One user confirmed via local analysis with evidence:

## Evidence (from user report + version diff of installed plugin cache)

```text
4.13.0: directive.md 19077 bytes, create_goal section at line 92 -> worked
4.14.1: directive.md 19463 bytes, new earlier content (browser QA / TUI visual QA guidance) pushed create_goal section to line 103 -> broken
4.15.1: same layout as 4.14.1 -> still broken
```

Codex App log shows:

```text
Warning: truncated output (original token count: 4866)
Total output lines: 373
<ultrawork-mode> ... middle omitted ... </ultrawork-mode>
```

The retained head+tail no longer includes the `Call create_goal` instruction. Regression start is **v4.14.1**, not v4.15.1.

## Impact

- `ulw` no longer creates a goal in Codex App, losing goal-tracking/binding success criteria for ultrawork runs.
- Task completion itself still generally works (the rest of the directive survives), so this is silent degradation rather than hard failure — which is why it went unnoticed.

## Suggested fix

Move a minimal `create_goal` bootstrap instruction to the very top of `packages/omo-codex/plugin/components/ultrawork/directive.md` (e.g. directly under `# Goal` or above the Manual-QA channel table), so it stays inside Codex App's retained hook-output head region regardless of directive growth. Alternatively keep the directive under the Codex hook-output token budget or split it.

A user-confirmed local workaround is adding a short `create_goal`-first line near the top of the cached directive (`~/.codex/plugins/cache/sisyphuslabs/omo/<version>/components/ultrawork/directive.md`), which restores the behavior.

## References

- Original standalone report: code-yeongyu/codex-ultrawork#1 (that repo is now archived; tracking moves here since the shipped directive lives in `packages/omo-codex/plugin/components/ultrawork/directive.md`)
- Discord reports: #general-ko (ordinary_businessman, with full root-cause analysis), #lazy-codex (qwertystars, cache/hook truncation observation)

code-yeongyu:
## Summary

Fixes #5828. Since v4.14.1, Codex App truncates the ultrawork `UserPromptSubmit` hook output (~4,866 tokens; head+tail retained, middle omitted), silently dropping the `create_goal` bootstrap instruction from the injected 19.5KB directive. This PR ships the full directive as a new plugin skill (`skills/ultrawork/SKILL.md`) and changes the hook to inject a ~1.2KB bootstrap pointer instead: it mandates the exact `ULTRAWORK MODE ENABLED!` opener, an immediate `create_goal` call with `objective` only, and reading the full directive from the skill file at a runtime-resolved absolute path. A 1.2KB payload can never reach the truncation budget, so no part of the bootstrap can be lost again.

## Changes

- **ultrawork component** (`plugin/components/ultrawork/`): new `src/skill-pointer.ts` builds the pointer and resolves `<plugin root>/skills/ultrawork/SKILL.md` from the built CLI's location; `codex-hook.ts` emits it on `ulw`/`ultrawork` triggers. When the skill file does not exist (standalone component installs without the plugin skills tree), the hook falls back to emitting the full directive, preserving old behavior. `scripts/sync-directive.mjs` now also generates `skills/ultrawork/SKILL.md` (frontmatter + directive body) from prompts-core; `directive-source.test.ts` pins the skill body to `directive.md` byte for byte. Transcript dedup is unchanged: the pointer keeps the `<ultrawork-mode>` marker.
- **ulw-loop component** (`plugin/components/ulw-loop/`): the standalone `--with-ultrawork` injection path emits the same pointer via `src/ultrawork-skill-pointer.ts` (byte-identical template, pinned by a new test), with the same full-directive fallback. In the shipped plugin only the ultrawork component injects; this covers the standalone distribution.
- **plugin wiring** (`plugin/scripts/sync-skills.mjs`, `plugin/test/`): the new skill is synced into the plugin skills tree and enumerated in the test-support pins. New `plugin/test/ultrawork-skill-pointer.test.mjs` drives both built component CLIs with a real `ulw` payload and pins the regression guard: pointer names the synced skill's absolute path, stays **under 4,096 bytes**, never inlines the directive, and both CLIs emit byte-identical payloads.
- Docs: component README injection mechanics rewritten; CHANGELOG entries in both components.

## QA & Evidence

Artifacts under `.omo/evidence/20260703-ulw-skill-pointer/` (local evidence dir, gitignored; summarized here).

- **What was tested:** `codex-qa` `hook-unit-probe.sh --self-test` (isolated install of THIS branch's build, deterministic hook stdout probe). **Observed:** PASS, `<ultrawork-mode>` injected on `ulw`. **Artifact:** `hook-unit-probe.txt`. **Why sufficient:** proves the trigger contract survives the payload swap.
- **What was tested:** `install-verify.sh --keep` + a probe driving BOTH installed component CLIs (`components/{ultrawork,ulw-loop}/dist/cli.js`) inside the isolated `CODEX_HOME` with a real `ulw` payload. **Observed:** 8/8 PASS — pointer is 1,208 bytes, names the installed skill's absolute path, carries the `create_goal` bootstrap and opener mandate, never inlines the directive, the target file exists (19,813 bytes) and contains the `## 1. Create the goal with binding success criteria` section; ulw-loop output byte-identical to ultrawork. **Artifacts:** `install-verify.txt`, `installed-pointer-probe.txt`, raw hook JSON. **Why sufficient:** proves the shipped installed layout, not just the repo tree, resolves the path correctly.
- **What was tested:** first-party live proof `app-server-drive.sh --plugin --prompt "ulw: say hello"` (real `codex app-server`, isolated home, local mock model). **Observed:** `hook/completed` for `sessionStart`, `userPromptSubmit`, `stop`; real `~/.codex/config.toml` shasum unchanged before/after every script. **Artifact:** `app-server-drive.json`.
- **What was tested:** live gpt-5.5 `codex exec` (quotio provider, isolated home) with an `ulw:` task, observing actual model compliance with the pointer. **Observed:** first visible line exactly `ULTRAWORK MODE ENABLED!`; the model's FIRST action was reading the installed `skills/ultrawork/SKILL.md`; goal lifecycle engaged ("Goal marked complete"); deliverable produced with the directive's RED→GREEN evidence discipline. **Artifact:** `live-model-run.txt`. **Why sufficient:** covers the one risk automation cannot — whether a real model follows the pointer instead of receiving the directive by force.
- **Gates:** `bun run test:codex` exit 0 from the worktree; plugin suite 304/304; ultrawork vitest 30/30; ulw-loop vitest 330/330; `tsc --noEmit` + `biome check` clean on both components.

## Risks & Residuals

- **Model skips the read:** mitigated two ways — the pointer itself carries the `create_goal` bootstrap and opener mandate (the exact instructions #5828 lost), and the live gpt-5.5 run shows full compliance. Accepted.
- **Skill file missing at runtime:** the hook falls back to emitting the full directive (current behavior), covered by unit tests in both components. Mitigated.
- **Pointer grows back into truncation territory:** `ultrawork-skill-pointer.test.mjs` hard-fails the suite above 4,096 bytes. Mitigated.
- **Windows:** not exercised locally; CI `codex-compatibility` runs `test:codex` on windows-latest. Accepted pending CI.

## Related Issues

Fixes #5828

---
## Summary by cubic
Fixes #5828 by moving the full ultrawork directive into a bundled skill and switching hooks to inject a small bootstrap pointer. This prevents Codex App truncation from dropping the required `create_goal` step.

- **Bug Fixes**
  - Inject a compact pointer (~1.2KB) that mandates the exact `ULTRAWORK MODE ENABLED!` opener, an immediate `create_goal` with `objective` only, and reading the full directive from `skills/ultrawork/SKILL.md` (absolute path resolved at runtime); falls back to the full directive when the skill file is missing.
  - Add `skills/ultrawork/SKILL.md`, generated from `@oh-my-opencode/prompts-core` via `scripts/sync-directive.mjs`, and sync it into the plugin with `plugin/scripts/sync-skills.mjs`.
  - Apply the same pointer in `ulw-loop` (`--with-ultrawork`) with a byte-identical template.
  - Tests pin the pointer under 4,096 bytes, verify no directive inlining, assert byte-identical outputs across `ultrawork` and `ulw-loop`, and ensure the skill body matches `directive.md`. Docs updated.

<sup>Written for commit e789e26cf93595e3a4aebc237ff24057a0293f6f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5830?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Re Cubic's P2 (duplication of `ultrawork-skill-pointer.ts` vs `skill-pointer.ts`): this duplication is deliberate and follows the repo's established pattern for these two components, and the cross-component behavioral parity tests the review asks for are already in this PR:

1. **Components must stay self-contained.** `plugin/components/*` are standalone npm packages (`codex-ultrawork` publishes on its own), and `plugin/test/component-bundled-cli.test.mjs` hard-asserts every built component CLI is self-contained except node builtins. A shared runtime module across components would break that contract. The pre-existing precedent is `directive.md` itself: ulw-loop ships a byte-identical copy pinned by `ulw-loop/test/ultrawork-directive.test.ts`.
2. **Resolver/behavior drift is pinned at the strongest level, the built artifact:** `plugin/test/ultrawork-skill-pointer.test.mjs` drives BOTH built CLIs (`components/ultrawork/dist/cli.js` and `components/ulw-loop/dist/cli.js --with-ultrawork`) with the same `ulw` payload and asserts the emitted `additionalContext` payloads are byte-identical. If the path resolver, template, or replacement logic ever diverged, the absolute paths inside the two payloads would differ and the suite fails.
3. **Fallback-semantics drift is pinned transitively:** the fallback branch returns the bundled directive verbatim, and the two components' `directive.md` files are pinned byte-identical by the existing identity test, so identical inputs cannot produce diverging fallback outputs. The template string itself is additionally pinned source-to-source by the new template identity test.

Extracting a shared module would trade a tested, pinned duplication for a new cross-package dependency that the component self-containment gate forbids, so we are keeping the duplicate-plus-pin pattern.

PR #5833, opened 2026-07-03
fix(opencode): keep ultrawork sticky per session

### LINKED ISSUE #5806: ULW mode is edge-triggered instead of level-triggered - follow-up messages lose delegation
## Bug Description

ULW mode is edge-triggered (per-message keyword scan) instead of level-triggered (session-persistent state). Once a session enters ULW mode, subsequent messages without the `ulw`/`ultrawork` keyword do NOT receive the ultrawork prompt injection, causing Sisyphus to revert to default behavior (doing everything itself instead of delegating to other agents).

## Root Cause

`packages/omo-opencode/src/hooks/keyword-detector/hook.ts` — `createKeywordDetectorHook` is a `chat.message` handler (line 57-247) that fires independently on every user message submission. It scans the current message text for keywords and injects the ultrawork prompt only if found. There is **no session-level state** tracking whether ULW mode has been activated.

Line 24 has `defaultModeUltraworkInjectedSessions` Set for the config-level default mode, but keyword-triggered ULW has no equivalent persistent state.

The same edge-triggered problem exists in `packages/omo-opencode/src/plugin/ultrawork-model-override.ts:70` (`detectUltrawork()`), which also rescans the current message independently.

## Reproduction

1. Send: `ulw refactor the auth module` → ULW prompt injected, Sisyphus delegates properly
2. Append: `and also add error handling` (no `ulw` keyword) → No ULW prompt injected, Sisyphus does everything itself

## Expected Behavior

Once a `ulw`/`ultrawork` keyword is detected in a session, all subsequent messages in that session should automatically receive the ultrawork prompt injection (level-triggered), until explicitly disabled.

## Suggested Fix

Introduce a `Set<sessionID>` (similar to `defaultModeUltraworkInjectedSessions` at line 24) that marks a session as ULW-activated when the keyword is first detected. On subsequent `chat.message` events for that session, auto-inject the ultrawork prompt regardless of keyword presence.

JSap0914:
## Summary
- Track keyword-activated ULW sessions so follow-up messages without `ulw`/`ultrawork` still receive the ultrawork prompt.
- Reuse the same session activation state for ultrawork model/variant override follow-ups.
- Preserve existing guards for planner/non-OMO/background sessions and avoid repeat activation toasts on sticky follow-ups.

Fixes #5806.

## Scope
- No disable/toggle mechanism is added; this matches the current issue fix and leaves explicit deactivation semantics for a follow-up.
- Hyperplan-ultrawork combo behavior is unchanged.

## Verification
- bun test packages/omo-opencode/src/hooks/keyword-detector/hook-ralph-loop.test.ts packages/omo-opencode/src/hooks/keyword-detector/hyperplan-ultrawork.test.ts packages/omo-opencode/src/hooks/keyword-detector/hyperplan.test.ts packages/omo-opencode/src/hooks/keyword-detector/index.test.ts packages/omo-opencode/src/hooks/keyword-detector/ultrawork-edge-trigger.test.ts packages/omo-opencode/src/hooks/keyword-detector/ultrawork-runtime-variant.test.ts packages/omo-opencode/src/hooks/keyword-detector/ultrawork-sticky-session.test.ts packages/omo-opencode/src/plugin/ultrawork-model-override.test.ts
- bunx tsgo --noEmit -p packages/omo-opencode/tsconfig.json

### DIFF
--- a/packages/omo-opencode/src/features/claude-code-session-state/state.ts
+++ b/packages/omo-opencode/src/features/claude-code-session-state/state.ts
@@ -5,6 +5,7 @@ export const syncSubagentSessions = new Set<string>()
 export const handedBackSyncSessions = new Set<string>()
 
 let _mainSessionID: string | undefined
+const ultraworkActivatedSessions = new Set<string>()
 
 export function setMainSession(id: string | undefined) {
   _mainSessionID = id
@@ -12,6 +13,14 @@ export function setMainSession(id: string | undefined) {
 
 export function getMainSessionID(): string | undefined {
   return _mainSessionID
+}
+
+export function markUltraworkSessionActive(sessionID: string): void {
+  ultraworkActivatedSessions.add(sessionID)
+}
+
+export function isUltraworkSessionActive(sessionID: string | undefined): boolean {
+  return typeof sessionID === "string" && ultraworkActivatedSessions.has(sessionID)
 }
 
 const registeredAgentNames = new Set<string>()
@@ -80,6 +89,7 @@ export function _resetForTesting(): void {
   syncSubagentSessions.clear()
   handedBackSyncSessions.clear()
   sessionAgentMap.clear()
+  ultraworkActivatedSessions.clear()
   registeredAgentNames.clear()
   registeredAgentAliases.clear()
 }
--- a/packages/omo-opencode/src/hooks/keyword-detector/hook.ts
+++ b/packages/omo-opencode/src/hooks/keyword-detector/hook.ts
@@ -4,6 +4,8 @@ import type { KeywordDetectorConfig } from "../../config/schema/keyword-detector
 import {
   getMainSessionID,
   getSessionAgent,
+  isUltraworkSessionActive,
+  markUltraworkSessionActive,
   subagentSessions,
 } from "../../features/claude-code-session-state"
 import type { ContextCollector } from "../../features/context-injector"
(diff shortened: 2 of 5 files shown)


MoerAI:
Thanks for the PR, @JSap0914.

This `fix()` comes in at 210 lines, which is well beyond the narrow, minimal-change shape we keep `fix()` PRs to so they stay reviewable and revertable. Could you narrow it to the smallest change that resolves #5806, and split any adjacent improvements into a separate PR?

Happy to look again once it is tighter.

code-yeongyu:
[sisyphus-bot] The sticky ULW session concept correctly addresses #5806's root cause — the fix makes ULW mode level-triggered as intended. Two concerns: (1) `ultraworkActivatedSessions` has no eviction mechanism — sessions never leave the set, so long-running processes will accumulate stale entries. Consider adding cleanup on session end or an LRU-style cap. (2) `resolveUltraworkOverride` now has side-effects (markUltraworkSessionActive) — consider extracting the marking to the caller to keep the function pure. Also, per MoerAI's review, could this be narrowed to the smallest change that resolves #5806?

MoerAI (changes requested):
The original scope request remains unresolved, and the current design introduces two additional blockers: session stickiness has no eviction path, so the set grows without bound in a long-lived process, and `resolveUltraworkOverride()` mutates session state as a hidden side effect. Please narrow this to the smallest #5806 fix or split the improvements, then add lifecycle cleanup and keep resolution pure before re-review.

MoerAI:
Thanks for investigating session-sticky ultrawork behavior.

Closing as stale rather than requesting another rebase. The underlying #5806 remains valid, but this July head conflicts with `dev`, has no follow-up after requested changes, and needs a semantic rewrite for the current bounded-session-state pattern:
https://github.com/code-yeongyu/oh-my-openagent/issues/5806
https://github.com/code-yeongyu/oh-my-openagent/pull/7527

A fresh narrow PR is welcome. Activation must remain behind the keyword hook's planner, non-OMO, and background admissibility guards; session state must be capped and cleaned on deletion/disposal; and `resolveUltraworkOverride` must remain pure:
https://github.com/code-yeongyu/oh-my-openagent/pull/5833#pullrequestreview-4889032001

Do not reopen this branch unless it is replaced with that current architecture and real OpenCode lifecycle evidence.

PR #5843, opened 2026-07-03
fix(runtime-fallback): preserve manual model selection

### LINKED ISSUE #5816: Manually selected model in UI is overridden by agent config model via runtime-fallback
## Bug Description

When a user manually selects a model in the OpenCode UI (using the model selector dropdown) and sends a message, the model is immediately overridden back to the OMO agent-configured model (from `oh-my-openagent.json` → `agents.sisyphus.model`). The manually selected model is used for the message, but the session reverts to the agent config model.

## Root Cause

The bug involves two interacting code paths in the runtime-fallback handler:

### 1. `handleSessionCreated` unconditionally overwrites session state

In `runtime-fallback/event-handler.ts`, **`handleSessionCreated`** (around line 111688) always overwrites `sessionStates[sessionID]` with a new state derived from the `session.created` event info, **without checking if the user has already made a manual model selection**.

```javascript
// Simplified logic:
const sessionModel = sessionInfo.model;            // e.g., "B" (user selected)
const preferredModel = resolvePreferredSessionModel(...); // "A" (agent config)
const fallbackIndex = preferredModel !== model
  ? getFallbackModelsForSession(...).indexOf(model) : -1;

const state = createFallbackState(
  fallbackIndex >= 0 && preferredModel ? preferredModel : model
);
// If model B is in the fallback list:
if (fallbackIndex >= 0) {
  state.currentModel = model;      // B
  state.originalModel = preferredModel; // A
}
sessionStates.set(sessionID, state);
```

Since the user-selected model B is in the agent's `fallback_models` list, the handler creates a state where:
- `originalModel = A` (agent config model)
- `currentModel = B` (user-selected, but treated as "fallback")

### 2. `createChatMessageHandler2` interprets the state as an active fallback

In `runtime-fallback/chat-message-handler.ts`, the chat.message handler (around line 111305) processes each message and at lines 111348-111364:

```javascript
const activeModel = state.currentModel;  // B
if (activeModel === state.originalModel) return;  // B !== A → continues

// ❌ Applies "fallback override" — actually reverts to agent config!
output.message.model = {
  providerID: parts[0],
  modelID: parts.slice(1).join("/")
};  // Forces model back to A
```

The handler interprets `currentModel !== originalModel` as "a fallback is active" and overrides `output.message.model` with the agent config model — even when the user manually selected the current model.

### Event timing scenario

```
1. User selects model B in UI, sends message
2. session.created fires → handleSessionCreated
   → state = { originalModel: A, currentModel: B }
   → B is treated as "fallback" because it's in fallback_models list
3. chat.message fires → runtime-fallback handler
   → input.model = B, state.currentModel = B
   → requestedModel === state.currentModel → no "manual change" detected
   → activeModel (B) !== originalModel (A) → treats as fallback override
   → output.message.model = A ← BUG!
```

## Expected Behavior

When a user manually selects a model in the UI:
1. The session state should respect the user's choice, not treat it as a fallback
2. `handleSessionCreated` should not overwrite session state that was set by a user's manual model selection
3. `createChatMessageHandler2` should distinguish between "fallback due to error" and "user manually selected a different model"

## Suggested Fix

There are two approaches:

### Approach A: Fix `handleSessionCreated` to not overwrite on manual selection

In the session.created event handler, check if the user already has a pending model selection before overwriting. If `chat.message` already processed the message and detected a manual model change (reset state), `session.created` should not overwrite that.

### Approach B: Fix `createChatMessageHandler2` to respect user selection

When `activeModel !== originalModel`:
1. Check if `input.model` contains the user's UI selection
2. If the "fallback" model matches the user's UI choice, it is NOT a fallback — it is a user preference
3. Only apply the model override when the model change was due to an API error, not a user action

A minimal fix would be in `createChatMessageHandler2`: add a check that if `input.model` matches `state.currentModel` and differs from `state.originalModel`, the user intentionally chose this model — do not override.

```javascript
// Additional check before applying fallback override:
if (input.model) {
  const requestedModel = `${input.model.providerID}/${input.model.modelID}`;
  if (requestedModel === state.currentModel && requestedModel !== state.originalModel) {
    // User intentionally selected this model — not a fallback
    return;
  }
}
```

## Additional Context

- OMO version: 4.15.0
- The bug occurs regardless of which models are involved, as long as the UI-selected model is in the agent's fallback_models list
- The bug does NOT occur for agents without explicit model config (because `hasExplicitAgentModelOverride` returns false, causing a different code path that does not interact with session state)
- Discovered while investigating issue #5815 (runtime-fallback not triggering on Free usage exceeded)

JSap0914:
## Summary
- stop `session.created` from bootstrapping a fallback-active state just because the created model is listed in an agent/category fallback list
- preserve an already pending fallback state when a retry-created session arrives
- update regression coverage for manually selected fallback-listed models

Fixes #5816.

## Verification
- `bun test packages/omo-opencode/src/hooks/runtime-fallback/event-handler.test.ts packages/omo-opencode/src/hooks/runtime-fallback/chat-message-handler.test.ts packages/omo-opencode/src/hooks/runtime-fallback/index.test.ts`

---
## Summary by cubic
Preserves manual model selection on session start and stops auto-activating fallback when the selected model is on a configured fallback list. Also keeps an in-flight fallback state when a session is re-created. Fixes #5816.

- **Bug Fixes**
  - `session.created` no longer marks a user-selected fallback-listed model as active fallback; we initialize with that model and `fallbackIndex = -1`.
  - If a session is re-created during a pending fallback, we keep the existing fallback state and only refresh last-access.

<sup>Written for commit 2e2f3590946b7e8f176436aabb26c583074baeed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5843?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/runtime-fallback/event-handler.test.ts
+++ b/packages/omo-opencode/src/hooks/runtime-fallback/event-handler.test.ts
@@ -274,4 +274,60 @@ describe("createEventHandler", () => {
     expect(created?.currentModel).toBe("openai/gpt-5.5-codex")
     expect(typeof created?.currentModel).toBe("string")
   })
+  it("#given session.created uses a fallback-listed manual model #when an agent has a preferred model #then the manual model is not treated as active fallback (#5816)", async () => {
+    const sessionID = "session-manual-model"
+    const deps = createDeps()
+    deps.pluginConfig = {
+      agents: {
+        sisyphus: {
+          model: "provider-a/model-a",
+          fallback_models: ["provider-b/model-b"],
+        },
+      },
+    }
+    const abortCalls: string[] = []
+    const clearCalls: string[] = []
+    const handler = createEventHandler(deps, createHelpers(deps, abortCalls, clearCalls))
+
+    await handler({
+      event: {
+        type: "session.created",
+        properties: {
+          info: { id: sessionID, agent: "sisyphus", model: "provider-b/model-b" },
+        },
+      },
+    })
+
+    const created = deps.sessionStates.get(sessionID)
+    expect(created?.originalModel).toBe("provider-b/model-b")
+    expect(created?.currentModel).toBe("provider-b/model-b")
+    expect(created?.fallbackIndex).toBe(-1)
+  })
+
+  it("#given session.created follows an already pending fallback #when the event fires #then existing fallback state is preserved", async () => {
+    const sessionID = "session-pending-fallback"
+    const deps = createDeps()
+    const state = createFallbackState("provider-a/model-a")
+    state.currentModel = "provider-b/model-b"
+    state.fallbackIndex = 0
+    state.pendingFallbackModel = "provider-b/model-b"
+    deps.sessionStates.set(sessionID, state)
+    const abortCalls: string[] = []
+    const clearCalls: string[] = []
+    const handler = createEventHandler(deps, createHelpers(deps, abortCalls, clearCalls))
+
+    await handler({
+      event: {
+        type: "session.created",
+        properties: { info: { id: sessionID, model: "provider-b/model-b" } },
+      },
+    })
+
+    const preserved = deps.sessionStates.get(sessionID)
+    expect(preserved?.originalModel).toBe("provider-a/model-a")
+    expect(preserved?.currentModel).toBe("provider-b/model-b")
+    expect(preserved?.fallbackIndex).toBe(0)
+    expect(preserved?.pendingFallbackModel).toBe("provider-b/model-b")
+    expect(deps.sessionLastAccess.has(sessionID)).toBe(true)
+  })
 })
--- a/packages/omo-opencode/src/hooks/runtime-fallback/event-handler.ts
+++ b/packages/omo-opencode/src/hooks/runtime-fallback/event-handler.ts
@@ -32,22 +32,6 @@ function resolveEventModel(props: Record<string, unknown> | undefined): string |
   return undefined
 }
 
-function resolvePreferredSessionModel(
-  sessionID: string,
-  agent: string | undefined,
-  pluginConfig: HookDeps["pluginConfig"],
-): string | undefined {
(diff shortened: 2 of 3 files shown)


code-yeongyu:
[sisyphus-bot] The core fix (early-return when session exists, remove preferred-model bootstrap) correctly addresses #5816's root cause. However, the behavior change to `restore_primary_after_cooldown` (test at index.test.ts:2641 now expects undefined instead of restoring primary) may surprise users who relied on automatic primary restoration. Consider splitting: keep the session-already-exists guard as the minimal fix for #5816, and move the restore-primary behavior change into a separate PR with its own justification. Also note: issue #5816 was closed on 2026-07-15 — was it closed in anticipation of this PR or resolved another way?

MoerAI:
Thanks for investigating the historical fallback-model bootstrap behavior.

Closing as obsolete. Current `dev` is variant-aware and generation-owned after #6611:
https://github.com/code-yeongyu/oh-my-openagent/pull/6611

The owner rechecked the current lifecycle and confirmed that manual fallback-listed selection is retained, while configured-primary restoration occurs only when `restore_primary_after_cooldown` is explicitly enabled:
https://github.com/code-yeongyu/oh-my-openagent/pull/5951#issuecomment-5002350655

This branch reverses that established opt-in contract, conflicts with `dev`, and is 4,484 commits behind. Any future policy change should start as a fresh, narrowly justified PR against the current generation-safe runtime-fallback state with current-head lifecycle QA.

PR #5750, opened 2026-06-29
fix(ralph-loop): preserve unrelated prompt reservations

renekris:
## Summary

Fixes Ralph-loop activity cleanup so prompt reservations are only released for sessions and reservation sources Ralph actually owns.

Before this change, child/subagent or verification-session activity could make Ralph attempt to release a prompt hold owned by another route, producing repeated `promptAsync reservation release skipped for different source` warnings during Ralph/ULW workflows. The updated branch also avoids doing disk-backed loop-state reads for unrelated streamed activity and bounds the shared prompt-gate mismatch log throttle.

## Changes

- Scope Ralph runtime-retry activity cleanup to active Ralph-owned parent or verification sessions.
- Check cheap in-memory runtime-retry and prompt-reservation ownership before falling back to disk-backed loop state.
- Release only prompt reservations owned by `ralph-loop` / `ralph-loop:*`; preserve holds owned by `model-suggestion-retry` and delegate-task routes.
- Clear stale Ralph runtime-retry markers without releasing another route's prompt reservation.
- Add a bounded throttle for repeated prompt-gate mismatched-release logs.
- Add characterization coverage for unrelated child activity, non-Ralph-owned verification activity, Ralph-owned verification activity, stale runtime markers, and bounded prompt-gate mismatch logging.

## QA & Evidence

- **What was tested:** Focused Ralph loop + prompt-gate regression surface.
  **Observed result:** `bun test packages/omo-opencode/src/hooks/ralph-loop packages/utils/src/prompt-async-gate.test.ts --bail` reported `209 pass`, `0 fail`.
  **Artifact:** `.omo/evidence/20260630-ralph-prompt-gate-full-qa/SUMMARY.md`
  **Why sufficient:** Covers the changed Ralph event-handler paths plus prompt-gate reservation behavior.

- **What was tested:** Mirrored/shared prompt-gate import surfaces.
  **Observed result:** `bun test packages/omo-opencode/src/hooks/shared/prompt-async-gate.test.ts packages/omo-opencode/src/hooks/shared/prompt-async-gate-semantic-dedupe-edge.test.ts packages/utils/src/prompt-async-gate*.test.ts --bail` reported `95 pass`, `0 fail`.
  **Artifact:** `.omo/evidence/20260630-ralph-prompt-gate-full-qa/SUMMARY.md`
  **Why sufficient:** Confirms both the utils implementation and OpenCode shared re-export surfaces remain compatible.

- **What was tested:** Typecheck.
[...]

PR #5853, opened 2026-07-04
fix(ralph-loop): handle deferred status in handleDetectedCompletion to avoid false-success toast

minhvoio:
## What changed

`handleDetectedCompletion()` in `packages/omo-opencode/src/hooks/ralph-loop/completion-handler.ts` only branched on the `"rejected"` status returned by `injectContinuationPrompt()`, silently ignoring the `"deferred"` status.

When the prompt-async gate defers the ultrawork verification prompt (e.g. an active reservation or gate conflict returns `{status: "deferred", reason: "active" | "reserved"}`), the handler fell through past both the `rejected` check and the deferred case straight into showing the `"ULTRAWORK LOOP" / "Oracle verification is now required"` success toast — even though the prompt was never actually dispatched to the session.

## Why

This is the same class of bug already handled correctly in two sibling files:
- `verification-failure-handler.ts` (`handleFailedVerification`)
- `iteration-continuation.ts` (`continueIteration`)

Both of those already have a `deferred` branch that logs and returns early without clearing loop state or showing a toast, preserving the pending state for a legitimate retry. `completion-handler.ts` was the one remaining call site missing this branch.

## Fix

Added the `deferred` branch immediately before the existing `rejected` branch, following the exact same pattern:

```ts
if (promptResult.status === "deferred") {
	log(`[${HOOK_NAME}] Deferred ultrawork verification prompt`, {
		sessionID,
		reason: promptResult.reason,
	})
	return
}
if (promptResult.status === "rejected") {
	...
}
```

No `loopState.clear()`, no toast — matches the sibling handlers' semantics for a deferred (not failed, not succeeded) dispatch.

## Test

Added a regression test in `dispatch-failure-invariant.test.ts` (same file as the sibling completion-path tests), using the same "`session.messages` throws" trick from `continuation-prompt-injector.test.ts` to force `injectContinuationPrompt` to return `{status: "deferred", reason: "active"}`. Asserts:
- loop state is **not** cleared (`cleared === false`)
- the `"ULTRAWORK LOOP"` success toast does **not** fire
- the `"Ralph Loop Failed"` toast does **not** fire either (it's neither a success nor a failure — it's deferred)

## Verification

- `bun test packages/omo-opencode/src/hooks/ralph-loop/dispatch-failure-invariant.test.ts` — 13 pass, 0 fail (was 12, added 1)
- `bun run typecheck` — clean
- `bun run build` — clean, schema regenerated with no diff

## Scope

Single-branch addition to one function. Does not touch:
- The non-ultrawork completion branch in the same file
- `loop-state-controller.ts`, `storage.ts`, `prompt-async-gate.ts`
- The `verification_pending` self-heal retry path in `pending-verification-handler.ts`

---
## Summary by cubic
Fixes a false success toast in the Ralph Loop when the ultrawork verification prompt is deferred. The handler now respects the `deferred` status, preserving state and avoiding incorrect user feedback.

- Bug Fixes
  - Handle `deferred` result from `injectContinuationPrompt` in `handleDetectedCompletion`, preventing the "ULTRAWORK LOOP" success toast when the prompt isn’t dispatched.
  - Log and return early; do not call `loopState.clear()` and do not show any toast on defer.
  - Add a regression test to ensure state isn’t cleared and no success/error toast fires on deferred.

<sup>Written for commit f2949594bde6b07a3706537b5dcd701014b9b80b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5853?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/ralph-loop/completion-handler.ts
+++ b/packages/omo-opencode/src/hooks/ralph-loop/completion-handler.ts
@@ -58,6 +58,13 @@ export async function handleDetectedCompletion(
 			directory,
 			apiTimeoutMs,
 		})
+		if (promptResult.status === "deferred") {
+			log(`[${HOOK_NAME}] Deferred ultrawork verification prompt`, {
+				sessionID,
+				reason: promptResult.reason,
+			})
+			return
+		}
 		if (promptResult.status === "rejected") {
 			log(`[${HOOK_NAME}] Failed to inject ultrawork verification prompt`, {
 				sessionID,
--- a/packages/omo-opencode/src/hooks/ralph-loop/dispatch-failure-invariant.test.ts
+++ b/packages/omo-opencode/src/hooks/ralph-loop/dispatch-failure-invariant.test.ts
@@ -694,6 +694,67 @@ describe("ralph-loop dispatch failure invariants", () => {
 		expect(toastCalls.some((toast) => toast.title === "Ralph Loop Failed" && toast.variant === "error")).toBe(true)
 	})
 
+	test("#given ultrawork completion path #when verification prompt injection is deferred #then no toast and no state clear", async () => {
+		// given
+		let cleared = false
+		const loopState = {
+			clear: () => {
+				cleared = true
+				return true
+			},
+			markVerificationPending: (sessionID: string) => ({
+				active: true,
+				iteration: 2,
+				prompt: "Build API",
+				started_at: new Date().toISOString(),
+				session_id: sessionID,
+				completion_promise: ULTRAWORK_VERIFICATION_PROMISE,
+				verification_pending: true,
+			}),
+		}
+
+		await handleDetectedCompletion({
+			directory: testDirectory,
+			project: testDirectory,
+			worktree: testDirectory,
+			serverUrl: "http://localhost:4096",
+			$: async () => ({}),
+			client: {
+				session: {
+					messages: async () => {
+						throw new Error("messages unavailable")
+					},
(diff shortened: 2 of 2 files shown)


minhvoio:

minhvoio:
Hey, this one is reviewed and approved, all checks are green except CI which is stuck on action_required. Looks like it just needs someone with admin or maintain rights to manually approve the workflow run so it can finish. Same exact thing is blocking #5865 and #5866 too, might be worth approving all three at once if you get a chance. Thanks for looking at this.

code-yeongyu:

MoerAI:
Thanks for this, and apologies for the slow turnaround. The deferred-result diagnosis and regression test are correct.

I am closing this as obsolete. PR #6184 replaced the user-facing Ralph/ULW loop with Goal after this branch opened; the Ralph hook is no longer composed and its commands were removed, so this completion handler is unreachable at runtime. This matches the disposition of the similar #5750 fix.

The live follow-up is to audit Goal's deferred `dispatchInternalPrompt` continuation path for the same false-success class. A focused fix there would be welcome.

PR #5903, opened 2026-07-05
fix(build): harden orchestrator failure and output handling

code-yeongyu:
## What changed

Four robustness improvements to `script/build.ts` (the parallel build DAG orchestrator merged in #5888). Three are Cubic findings from that PR's review; the fourth is a CI-attribution guard requested by the ci-workflows reviewer.

1. **Kill in-flight children on first failure** (Cubic P1). `runGraph` tracks live children and, on the first step failure, kills the rest and awaits their settle before rethrowing, so no step keeps writing artifacts after the build has already failed.
2. **Bounded per-step output** (Cubic P2). Retained stdout/stderr is tail-capped at 1 MiB, so a verbose parallel run cannot spike memory.
3. **Single-settle guard** (Cubic P2). `runNode`'s `error`/`close` handlers are guarded by a `settled` flag, so a spawn failure (which fires both events) cannot double-settle the promise.
4. **Clean CI failure attribution** (ci-workflows guard). A top-level catch prints `build: FAILED: <step> failed with <reason>` as the **last** stdout line before `process.exit(1)`, so a failing wave stays attributable despite interleaved parallel output. (Each spawned step already prefixes its output with `build:<step>`.)

## QA / evidence

- **Byte-identical build**: hardened vs merged `build.ts`, same worktree, both cold → aggregate `shasum` `6c5680d7…` == `6c5680d7…`. The changes only touch the failure path and output buffering; happy-path output is unchanged.
- **Failure semantics**: an injected failing step exits non-zero, and the last stdout line is exactly `build: FAILED: build:schema failed with exit code 1`.
- **typecheck**: clean.
- **A runtime bug the manual QA caught** (typecheck did not): my first placement of the top-level `try { await run() }` sat *above* the `MAX_RETAINED_OUTPUT_BYTES` const, causing a TDZ `ReferenceError` at runtime. Moved the entry `try/catch` to the end of the module, after all declarations; the smoke build then completed and produced `dist/index.js`.

Evidence: `.omo/evidence/20260706-build-orchestrator-robustness/`.

## Residual risk

Failure propagation is unchanged except that siblings are now stopped and the promise settles exactly once. No payload or publish semantics change.

cc @a415451b285c9c984 (ci-workflows) for the failure-attribution + byte-identical review from the CI-consumer angle, as offered.

---
## Summary by cubic
Harden the parallel build orchestrator for safer failures and clearer CI logs. Stops in-flight steps on first failure, bounds per-step output memory, and avoids double-settle races.

- **Bug Fixes**
  - `runGraph`: track live children; on first failure, kill others and await settle.
  - Cap retained stdout/stderr per step at 1 MiB (tail) to bound memory use.
  - `runNode`: guard `error`/`close` with a `settled` flag to prevent double-settle.
  - Top-level: print `build: FAILED: <reason>` as the final stdout line before `process.exit(1)` for clear CI attribution.

<sup>Written for commit 35e2d3a752068fb7af7776d327799a9aeed15737. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5903?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/script/build.ts
+++ b/script/build.ts
@@ -45,8 +45,6 @@ const nodes: BuildNode[] = [
 	{ id: "schema", command: "bun", args: ["run", "build:schema"], deps: [] },
 ];
 
-await run();
-
 async function run() {
 	await materializeOnce();
 	const childEnv = { ...process.env, OMO_SKIP_MATERIALIZE: "1" };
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Reviewed from the CI-consumer angle (ci.yml build job owner). LGTM on both points you flagged.

**1. FAILED-line format works for CI build-job triage.** Confirmed the failing step name propagates cleanly: `runNode` rejects with `build:<id> failed with <reason>`, and the top-level catch prints `build: FAILED: <that message>` as the last stdout line. I traced the kill-in-flight interaction and it preserves triage:
- `Promise.race` settles on the FIRST rejection (the real failure), captured by the `runGraph` catch.
- Kill-in-flight then stops the other live children; `Promise.allSettled(running)` waits without throwing, so the killed children's SIGTERM rejections are swallowed and cannot overwrite the real error.
- The original `throw error` re-throws the REAL first failure, so `build: FAILED:` always names the true culprit, never a collateral killed step.
- Killed children flush their output blocks on `close` during `allSettled`, i.e. BEFORE the final throw, so they print above the FAILED line. The FAILED line stays last.

Empirically confirmed the shape by running the branch's build.ts with a broken step: output ended with
`build:materialize` (the step's own block) then `build: FAILED: build:materialize failed with exit code 1` as the last line. Exactly the format a reviewer needs.

**2. Wave reordering stays invisible to ci.yml.** This PR does not touch the DAG (`nodes[]`/`deps[]` unchanged) — only failure/output handling. ci.yml consumes build artifacts only (`test -f dist/index.js`, `dist/index.d.ts`, the two dist-bundle tests), never build stdout or step order, so nothing here is observable to the build job beyond the artifact set. Verified `dist/index.js` + `dist/index.d.ts` are still produced after the hardened build.

**Non-blocking nit (optional):** `child.kill()` sends SIGTERM; a wedged child could ignore it. Acceptable here — the intent (stop artifact writes) holds for well-behaved tsc/bun processes and the CI job timeout is the backstop. Only worth a SIGKILL escalation if you ever see hung builds in practice.

Ship it.

PR #5913, opened 2026-07-06
fix(tui): sanitize sidebar poll error logs (fixes #5874)

### LINKED ISSUE #5874: TUI: Error output leaks into input area after wake from sleep
## Description

After putting the PC to sleep and waking it back up, a transient `Failed to fetch models.dev` error renders in the TUI content/input area instead of being handled gracefully (e.g., via toast or silent log). The error is transient — new sessions load fine — but it pollutes the input area until then.

This started after installing oh-my-openagent.

## Screenshot

error in input area

The error output:
```
[15:12:15.061] ERROR (#106): Failed to fetch models.dev {
  cause: {
    _id: "Cause",
    failures: [
      [Object ...]418.9K (42%) - .82  ctrl+p commands
    ]
  }
}
```

Note the TUI status bar text (`418.9K (42%) - .82 ctrl+p commands`) mixed in with the error — confirming it's rendered in the wrong pane.

## Root Cause

The TUI sidebar polling (line 77083 in `dist/tui.js` — `handleTuiPollError`) catches the network error from the models.dev fetch after wake, but the raw error output leaks into the TUI content area instead of being routed to a toast or log-only handler.

## Expected Behavior

Error should be either:
- Shown as a toast notification (not in the content/input area)
- Logged silently without rendering in the main TUI area
- Retried once before surfacing

## Environment

- OpenCode: 1.17.13
- oh-my-openagent: 4.15.1
- OS: Linux (x86_64)
- Node: v24.18.0
- Reproducibility: Transient — occurs after sleep/wake, resolves on new session

MoerAI:
## Summary
- Sanitize the default TUI sidebar poll error log payload to stable `{ name, message }` fields instead of passing the raw `Error` object through the logger.
- Export a narrow `startTuiSidebar` helper so the focused test can drive sidebar startup without casting the OpenCode TUI API.
- Add a regression test that an `Error` with a large `cause` does not include `cause` in the log payload.

Fixes #5874.

## Verification
- `bun test packages/omo-opencode/src/tui.test.ts`
- `bun run packages/shared-skills/skills/programming/scripts/typescript/check-no-excuse-rules.ts packages/omo-opencode/src/tui.ts packages/omo-opencode/src/tui.test.ts`
- `git diff --check origin/dev..HEAD`
- Evidence recorded locally under `.omo/evidence/20260706-tui-poll-error-5874/`.

## QA note
- Ran the required `opencode-qa` TUI path as far as this Windows host allows. The common self-check and TUI smoke are blocked here by missing `tmux` and `jq`; Docker fallback is also unavailable (`docker: command not found`).
- Full OpenCode tmux TUI smoke remains the residual pre-merge QA gap for a host with `tmux` and `jq`.

---
## Summary by cubic
Sanitize TUI sidebar poll error logs to emit only {name, message} instead of the raw Error, preventing large cause payloads from leaking. Also expose a small `startTuiSidebar` helper and add a regression test; fixes #5874.

- **Bug Fixes**
  - `handleTuiPollError` now logs via `formatTuiPollErrorForLog`, sending `{ name, message }` only.
  - Keeps rethrow for non-Error values.
  - Adds a test that ensures `cause` is not included in log payloads.

- **Refactors**
  - Extracts sidebar startup to `startTuiSidebar` for focused tests without casting `@opencode-ai/plugin/tui` types.
  - Handles missing `@opentui/solid` by returning early.

<sup>Written for commit 9907d877e16d55a9de5cbc8177efbc395c250739. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5913?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/tui.test.ts
+++ b/packages/omo-opencode/src/tui.test.ts
@@ -5,9 +5,9 @@ import { mkdtempSync, rmSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
 
-import type { TuiPluginApi, TuiPluginMeta, TuiSlotPlugin } from "@opencode-ai/plugin/tui"
+import type { TuiSlotPlugin } from "@opencode-ai/plugin/tui"
 
-import tuiModule, { handleTuiPollError } from "./tui"
+import { formatTuiPollErrorForLog, handleTuiPollError, startTuiSidebar } from "./tui"
 
 type SolidNode = {
   readonly tag: string
@@ -89,7 +89,7 @@ describe("TUI sidebar polling", () => {
     } satisfies SidebarApiForTest
 
     // when
-    await tuiModule.tui(api as unknown as TuiPluginApi, undefined, {} as TuiPluginMeta)
+    await startTuiSidebar(api)
 
     // then
     expect(calls).toEqual(["register", "render"])
@@ -115,6 +115,26 @@ describe("TUI sidebar polling", () => {
 
     // then
     expect(reportedErrors).toEqual([pollError])
+  })
+
+  it("#given a poll Error with a large cause #when the log payload is formatted #then only stable text fields are retained", () => {
+    // given
+    const pollError = new Error("Failed to fetch models.dev", {
+      cause: {
+        statusText: "socket closed",
+        rawOutput: "x".repeat(1024),
+      },
+    })
+
+    // when
+    const payload = formatTuiPollErrorForLog(pollError)
+
+    // then
+    expect(payload).toEqual({
+      name: "Error",
+      message: "Failed to fetch models.dev",
+    })
+    expect("cause" in payload).toBe(false)
   })
 
   it("#given a non-Error throw during polling #when the poll error handler runs #then the value is rethrown", () => {
--- a/packages/omo-opencode/src/tui.ts
+++ b/packages/omo-opencode/src/tui.ts
@@ -29,6 +29,26 @@ type RegisterSidebarContentSlotInput<Node> = {
   readonly renderSidebar: () => Node
 }
 
+type TuiSidebarApi = {
+  readonly state: {
+    readonly path: {
+      readonly directory: string
+    }
+  }
+  readonly theme: {
+    readonly current: Record<string, unknown>
+  }
+  readonly slots: {
+    readonly register: (registration: SidebarSlotRegistration<unknown>) => string
+  }
+  readonly renderer: {
+    readonly requestRender: () => void
+  }
+  readonly lifecycle: {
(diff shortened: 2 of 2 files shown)


MoerAI:
Maintainer note (self-PR; GitHub blocks self-review-requests, posting as a comment):

Thanks for the PR, @MoerAI, and for taking the time on this.

A few things to address before this can land:

- I could not confirm this fixes the reported behavior in #5874 on current dev. The regression test in `packages/omo-opencode/src/tui.test.ts` fails on unmodified dev only with a missing-export error (`startTuiSidebar` is introduced by this PR), so it does not demonstrate the bug: it exercises the new helper rather than failing on the defect's observable behavior. Please add a test that drives `handleTuiPollError` through its existing entry point and fails on dev for a behavioral reason (for example: the default reporter's log payload includes `cause`).
- #5874 reports the error rendering into the TUI content/input area, and its expected behavior is a toast or a silent log. Sanitizing the payload to name and message shrinks the leak, but the sanitized line still follows the same route into the pane. Please either route the poll error away from the content area (toast or log-only) or attach TUI smoke evidence from a host with tmux and jq showing the input area stays clean after wake; the PR body itself lists that smoke run as the residual QA gap.
- After #5316 had to be reverted in #5361, fix() merges here need a reproduction that fails on unmodified dev and passes on the PR head through the real execution path; green CI alone is not enough.

These are conventions documented in `AGENTS.md`. Once they are addressed I will take another look. No rush.

MoerAI:
Closing this one after a second look — the earlier review concerns hold up:

- The regression test only fails on unmodified `dev` with a missing-export error (`startTuiSidebar` / `formatTuiPollErrorForLog` are introduced by this PR), so it exercises the new helper rather than failing on #5874's observable behavior. That is the same shape that led to #5316 being reverted in #5361.
- More fundamentally, #5874 is output leaking into the TUI content/input **pane**, but this change only sanitizes the `log()` payload — which writes to the temp-dir logfile, not the pane. The route into the pane is unchanged, so this does not fix the reported behavior. Trimming `cause` from the log line is a minor logging hardening, not a fix for #5874.
- The TUI-with-tmux live proof such a fix would need remains an unmet residual gap.

Rather than keep a PR open that does not address the root cause, I am closing it. Happy to revisit #5874 with a change that actually reroutes the pane output if that is wanted.

PR #5923, opened 2026-07-06
fix(codex): repair Windows OMO command shims

kyzer1023:
## Summary

- repair the shared Windows OMO command-shim generator so a double-quoted `NODE_REPL_NODE_PATH` is normalized without triggering invalid batch syntax
- let the ULW Git Bash bootstrap discover installed `omo.cmd` shims from `$HOME/.local/bin` and `$CODEX_HOME/bin`, while preserving the cached `.js` fallback through Node
- add real-process and generated-bundle regressions for the root wrapper, cached component shims, and ULW bootstrap

## Maintainer feedback addressed

This branch was rebuilt from current `dev` instead of rebasing the stale PR history. The unrelated version bump and managed-agent reasoning metadata are gone. The generated installer diff is limited to the source-derived `windowsNodeDiscoveryLines()` quote-normalization hunk; there is no version, model, reasoning, telemetry, PostHog, or export-order drift.

The refreshed history is intentionally split into four reviewer-sized commits:

1. `fix(codex): repair quoted Windows node paths`
2. `fix(ulw-loop): discover Windows omo command shims`
3. `chore(codex): refresh Windows installer shims`
4. `test(codex): cover spaced quoted node paths`

## Why

On Windows, `NODE_REPL_NODE_PATH` can arrive wrapped in double quotes. The generated `.cmd` parser compared the first/last character using invalid batch syntax, causing `The syntax of the command is incorrect.` before Node could launch.

Separately, the Codex installer may expose only `omo.cmd` in the local bin directory. The ULW bootstrap running inside Git Bash checked only extensionless `omo`, so it could miss a valid Windows install and fail with exit 127.

### Failing first

- real `cmd.exe` probes for both the root wrapper and cached component shim exited 255 with `The syntax of the command is incorrect.`
- the real Git Bash bootstrap probe exited 127 when only `$HOME/.local/bin/omo.cmd` existed

### After this change

- both real `cmd.exe` probes use a quoted executable under `...\\node runtime\\bun.exe`, exit 0, invoke the expected Node target, preserve `--probe`, and emit no stderr
- the same Git Bash fixture exits 0 with `{"ok":true,"source":"windows-cmd-shim"}`
- focused quoted-path tests: 2 passed
- ULW package smoke: 17 passed, 1 platform skip
- generated Windows bin-link tests: 4 passed
- generated installer bundle tests: 7 passed
- ULW typecheck and lint passed; `packages/omo-codex` typecheck passed

## Codex QA

- isolated install verification passed: local plugin cache, enabled `omo@sisyphuslabs`, 9 sandbox component bins, agent TOMLs, and unchanged real `~/.codex/config.toml`
- deterministic hook probe passed for ultrawork `UserPromptSubmit`
- real `codex app-server` with an isolated `CODEX_HOME` and local mock model completed successfully with no missing/failed hooks; first-party `hook/started` and `hook/completed` notifications were observed for `sessionStart`, `userPromptSubmit`, and `stop`
- LSP layer: 25 files passed, 95 tests passed, 1 skipped
- CodeGraph layer: 78 passed; third-party notice shipping verification passed

## Broader gate note

`bun run test:codex` was attempted from both PowerShell and Git Bash. PowerShell stops at an existing POSIX `rm` npm script. Git Bash advances through the LSP layer, then stops on two ULW checkpoint-continuation tests whose relevant test and implementation files are unchanged from `dev`. The remaining layers were run independently. The broad Windows suites retain unrelated host limitations dominated by `EPERM` symlink creation and legacy-daemon fixture execution; the focused changed-surface regressions above pass.

## Residual risk

Local Windows QA cannot replace the repository's Linux/macOS matrix for symlink-heavy fixtures. GitHub CI is the cross-platform gate. The product paths changed here are covered directly by real Windows `cmd.exe`, real Git Bash, isolated install, generated parity, and live Codex app-server evidence.


code-yeongyu:
[sisyphus-bot] The core fix (batch quote stripping) is correct and well-tested. However, the branch conflicts with dev and includes an unrelated version bump (4.18.2→4.19.0) and a managed reasoning upgrade entry in the bundled install-local.mjs. Please rebase and split the version bump / reasoning upgrade into a separate commit or PR.

kyzer1023:
Refreshed this PR from current `dev` in response to your review. The unrelated version bump and managed-agent reasoning changes have been removed; the generated installer now contains only the source-derived Windows quote-normalization hunk. I also strengthened the regressions with real `cmd.exe` coverage for both root and cached shims, a real Git Bash `omo.cmd` bootstrap probe, generated parity checks, isolated install verification, and a live `codex app-server` hook run. The PR description has the observed results and the documented Windows-host limitations. Ready for another look.

MoerAI (changes requested):
Thanks for reproducing the Windows quote-stripping failure with real `cmd.exe` and Git Bash coverage. The core parser defect remains live in #7473:
https://github.com/code-yeongyu/oh-my-openagent/issues/7473

Please rebuild this fix from current `dev` and retain only the current command-shim repair plus regenerated installer output. Do not restore `omo` or `omo.cmd`; the supported wrapper is now `omo-agent-toolkit(.cmd)`:
https://github.com/code-yeongyu/oh-my-openagent/commit/89f09c90ca4415e07de1af859036ddde9572f95b

Rework any ULW resolution against that current architecture. Keep the real Windows root and cached-shim regressions, adapt them to the supported wrapper, and add Git Bash coverage only where the current path requires it.

Request `ci:full-matrix` so Windows-only tests actually execute, regenerate the published installer from source, and commit isolated Codex QA evidence for the rebuilt tree under `.omo/evidence/`.

PR #5983, opened 2026-07-08
fix(auto-slash-command): remove duplicate user args in formatCommandTemplate

### LINKED ISSUE #6162: Auto slash commands duplicate consumed user arguments
## Bug

Auto slash-command templates that include `$ARGUMENTS` or `${user_message}` receive the user's arguments twice: once through placeholder substitution and again in the trailing `## User Request` section added by `formatCommandTemplate()`.

## Reproduction

Given a command template:

```md
Echo $ARGUMENTS and ${user_message}.
```

and invocation `/templated ship it`, the rendered prompt contains `ship it` in the substituted command instructions and repeats it in a trailing `## User Request` section.

## Expected behavior

- Templates that consume `$ARGUMENTS` or `${user_message}` should not append the trailing request section.
- Templates without either placeholder must retain the trailing section so their arguments are not lost.
- User arguments must remain opaque data; file-like and shell-like text must not be executed.

## Delivery

PR #5983 contains the bounded fix and failing-first regression coverage. Runtime token substitution such as `$SESSION_ID` and `$TIMESTAMP` is outside this issue and remains owned by PR #5984.

EZotoff:
## Summary

Prevent auto slash-command prompts from repeating user arguments when the command template already consumes `$ARGUMENTS` or `${user_message}`. Commands without either placeholder keep the trailing `## User Request` section, preserving the existing argument-delivery fallback.

## Changes

- Detect supported argument placeholders in the resolved command template before appending the trailing request section.
- Add failing-first coverage for both placeholder-consuming and no-placeholder command populations.
- Verify file-like and shell-like arguments such as `@secret.txt`, `$(touch marker)`, and `$HOME` remain inert data.
- Keep `$SESSION_ID` / `$TIMESTAMP` runtime-token work out of scope; that remains in PR #5984.

## QA

- `bun test packages/omo-opencode/src/hooks/auto-slash-command/executor.test.ts`: 7 passed, 0 failed, 21 expectations.
- Auto-slash-command hook suite: 69 passed.
- Local pre-build full-suite attempt: 11,603 passed, 3 skipped, 21 failed in generated/payload-dependent Codex installer, package-layout, telemetry-bundle, and path-sensitive config tests; retained as a failed local artifact.
- Exact-head GitHub Actions run `29519402551`: every required Linux, macOS, and Windows test/typecheck/build/Codex/Senpi/published-smoke/security job passed on `97fafb79f285f9dfa3e5b7113d1046c30e967e12`.
- `bun run typecheck`: passed.
- Full build: passed on the exact source tree through a path-neutral wrapper because the mandated team worktree path contains `.omo`, which `sync-skills.mjs` filters.
- Real isolated OpenCode QA: two `opencode run --format json` executions loaded the source plugin, exchanged four requests with a local fake model, created only the sandbox DB, left the host DB session count unchanged at 5751, and completed teardown.

The real OpenCode wire preserves native slash input rather than exposing the plugin-formatted prompt, so duplicate-rendering behavior is asserted by the direct executor test; the real harness proves source-plugin loading, execution, isolation, and cleanup.

## Risk

The conditional is limited to the two placeholders already substituted by this function. Templates without placeholders retain their prior trailing argument placement. No runtime-token substitution behavior is changed.

Closes #6162

### DIFF
--- a/packages/omo-opencode/src/hooks/auto-slash-command/executor.test.ts
+++ b/packages/omo-opencode/src/hooks/auto-slash-command/executor.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it } from "bun:test"
-import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from "node:fs"
+import { existsSync, mkdtempSync, mkdirSync, rmSync, writeFileSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
 import { clearCommandLoaderCache } from "../../features/claude-code-command-loader"
@@ -44,6 +44,24 @@ Execute daplug prompt flow.
 description: Templated prompt from daplug
 ---
 Echo $ARGUMENTS and \${user_message}.
+`,
+  )
+  writeFileSync(
+    join(pluginInstallPath, "commands", "special-args.md"),
+    `---
+description: Special argument prompt from daplug
+---
+Echo $ARGUMENTS.
+`,
+  )
+  const userCommandsDir = join(claudeConfigDir, "commands")
+  mkdirSync(userCommandsDir, { recursive: true })
+  writeFileSync(
+    join(userCommandsDir, "plain.md"),
+    `---
+description: Plain user prompt
+---
+Execute the plain prompt.
 `,
   )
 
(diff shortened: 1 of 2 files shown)


MoerAI (dismissed):
Thanks for the PR, @EZotoff, and for the clear writeup with the formatLoadedCommand precedent.

The duplication is real (I verified both insertion sites on current dev in packages/omo-opencode/src/hooks/auto-slash-command/executor.ts), but a few things need addressing before this can land:

- The trailing ## User Request section is not purely redundant: for command templates that do not contain the $ARGUMENTS or user_message substitution tokens, it is the only args placement adjacent to the instructions, while the **User Arguments** header sits in the metadata block far above. Removing it unconditionally changes prompt shape for every arg-taking command, including ones that may rely on the trailing placement. A safer shape would be to skip the trailing section only when the template substitution actually consumed the arguments, and keep it otherwise (or make the case that header-only is sufficient for both populations, with evidence).
- Please add a regression test that fails on unmodified dev and passes on this branch, driving formatCommandTemplate through its real rendering path with an args-bearing command. After #5316 had to be reverted in #5361, fix() merges here need a failing-on-dev reproduction rather than green CI alone.
- There is no linked issue. Please file one (or link the existing report if there is one) describing the duplication and its observed impact, so the fix and its issue tell the same story.

Happy to take another look once those are in. No rush, and thanks again for spotting the duplication.

code-yeongyu:
[sisyphus-bot] Addressed the requested changes on exact head `97fafb79f285f9dfa3e5b7113d1046c30e967e12`:

- The trailing `## User Request` section is now skipped only when the resolved template contains `$ARGUMENTS` or `${user_message}`; no-placeholder templates retain it.
- Added a failing-on-dev regression through the real executor rendering path for both populations.
- Added literal special-argument coverage for `@secret.txt`, `$(touch marker)`, and `$HOME`.
- Filed and linked #6162.

Focused, hook-suite, full-test, typecheck, build, and isolated real OpenCode gates are green. The real harness is intentionally scoped to plugin loading/execution/isolation because OpenCode's native command processing preserves the raw slash input at the model wire; the direct executor test is the rendering proof.

code-yeongyu (approved):
Independent read-only exact-head review task 019f6bfe-02a9-72c0-a2db-725f8ff73a06 reviewed 97fafb79f285f9dfa3e5b7113d1046c30e967e12 and returned APPROVE with no blockers. Exact-head CI run 29519402551 is green; residual test-proof limitations are recorded in local evidence.

PR #5985, opened 2026-07-08
fix(ultrawork): route planning to Prometheus instead of native plan agent

### LINKED ISSUE #5850: [Bug]: ulw planner sometimes uses native OpenCode plan mode and leaves the parent waiting
### Prerequisites

### Bug description

When I run a larger task through `ulw` in Sisyphus, it does the research correctly and then delegates planning to a subagent with the right task context. The planner sometimes uses OpenCode's native plan mode instead of omo's planning flow.

That means it writes a plan under `.opencode/plans/` instead of `.omo/plans/`, then calls OpenCode's native `plan_exit` tool. The TUI shows the native plan prompt asking whether to switch to the build agent and start implementing. In my first report I paraphrased this as "Implement the plan?". The exact OpenCode prompt is:

> Plan at {path} is complete. Would you like to switch to the build agent and start implementing?

The prompt only gives "Yes" and "No". There is no custom answer.

If I choose "Yes", the planning subagent starts implementing the plan itself. That produces worse results and also tricks the parent agent into thinking the task has already been handled. If I choose "No" or press Esc, the session gets stuck in a strange state: the subagent stops making progress, but it also does not return the plan to the parent. The parent agent just waits.

The only recovery I have found is to press Esc in the parent agent to interrupt the wait. After that, the subagent finally returns that it finished planning, and the parent agent recovers.

This is intermittent. I am fairly sure I have also seen the planner use the omo plan path correctly, but I am not 100% certain.

I think this is related to #2526 because that issue also looks like a planning subagent getting stuck.

### Steps to reproduce

1. Start a complex task with `ulw` in Sisyphus, using a task large enough to trigger a planning subagent.
2. Wait for the planning subagent to produce a plan.
3. In the failing case, observe that the subagent uses OpenCode's native plan mode instead of omo's plan flow.
4. When the native `plan_exit` prompt appears, choose "Yes", "No", or press Esc.

### Expected behavior

The `ulw` planning delegation should use omo's planning flow every time. The planner should write the plan under `.omo/plans/`, return it to the parent agent, and never enter OpenCode's native plan exit flow from inside a delegated subagent.

### Actual behavior

The `ulw` planner sometimes enters OpenCode's native plan mode. It writes under `.opencode/plans/`, calls `plan_exit`, and shows the native OpenCode plan prompt inside the delegated subagent flow.

Choosing "Yes" makes the subagent implement the plan itself. Choosing "No" or pressing Esc can leave the parent agent waiting until I manually interrupt it.

### Source trace

I traced this through both the omo plugin source and OpenCode core. The problem looks like an agent-name collision plus conflicting instructions.

`packages/prompts-core/prompts/ultrawork/default.md`, in the "MANDATORY: PLAN AGENT INVOCATION" section, tells Sisyphus to delegate planning with:

```ts
task(subagent_type="plan", load_skills=[], run_in_background=false, prompt="")
```

In OpenCode, `plan` is a native built-in agent. omo does not have its own `plan` subagent, so `subagent_type="plan"` resolves to OpenCode's native plan-mode agent.

OpenCode's native plan agent is hardwired for the native plan flow. In `packages/opencode/src/agent/agent.ts`, the plan agent has `plan_exit: "allow"` and edit permissions scoped to `.opencode/plans/*.md`. OpenCode also injects its plan-mode reminder from `packages/opencode/src/session/prompt/plan-mode.txt`, including this instruction:

> ### Phase 5: Call plan_exit tool
> At the very end of your turn ... you should always call plan_exit ... your turn should only end with either asking the user a question or calling plan_exit. Do not stop unless it's for these 2 reasons.

omo then tries to steer that same native `plan` agent with its own planner prompt. The keyword detector sees the planner agent in `packages/omo-opencode/src/hooks/keyword-detector/ultrawork/source-detector.ts`, where `isPlannerAgent()` matches `\bplan\b`, and injects `planner.md`. That prompt tells the agent to behave like Prometheus, write under `.omo/plans/`, hand the plan back, and not implement.

So the same delegated subagent receives two incompatible instruction sets:

- omo's planner instructions: act like Prometheus, use `.omo/plans/`, hand off the plan, and do not implement.
- OpenCode's native plan-mode instructions: use `.opencode/plans/` and always finish with `plan_exit`.

Because omo never denies `plan_exit` in this path, the native tool can still fire inside the delegated subagent. That matches the intermittent behavior I am seeing.

The `plan_exit` behavior also explains the hang. In `packages/opencode/src/tool/plan.ts`, choosing "Yes" injects a synthetic user message into the current session with `agent: "build"` and an "Execute the plan" instruction. In this case, the current session is the delegated subagent, so the subagent starts implementing instead of returning to the parent. Choosing "No" throws `RejectedError`. The build-agent switch is wired for the primary TUI in `packages/tui/src/routes/session/index.tsx`, which listens for `plan_exit` completion and switches the primary agent to `build`. A nested `task()` subagent does not have that same clean path back to the parent.

Prometheus does not hit this problem because it is an omo custom agent locked to `.omo/plans/` by the `prometheus-md-only` hook. It has no `plan_exit` permission and no native plan-mode reminder, so it cannot enter this flow.

There is also a naming clash in omo's own routing. In `source-detector.ts`, `isNonOmoAgent()` classifies `agentName === "plan"` as OpenCode's built-in Plan agent, which should not receive omo injection. At the same time, `isPlannerAgent()` routes that same `plan` agent to the Prometheus planner injection. omo is both treating `plan` as native OpenCode and trying to reskin it as an omo planner.

### Possible fix directions

A few options seem plausible:

- Route `ulw` planning to Prometheus, or to another first-class omo planning subagent, instead of `subagent_type="plan"`.
- Deny `plan_exit` and `plan_enter` for delegated planner subagents.
- Stop injecting omo planner instructions into OpenCode's native `plan` agent, then give omo its own planning subagent so native plan mode never activates under `ulw`.

### Doctor output

```shell
 oMoMoMoMo Doctor

 ⚠ 2 issues found:

1. Loaded plugin is outdated
   Loaded 4.15.0, latest 4.15.1.
   Fix: Update: cd "~/.cache/opencode/packages/oh-my-openagent@latest" && bun add oh-my-openagent@latest
   Affects: plugin features

2. GitHub CLI missing
   gh CLI is not installed.
   Affects: GitHub automation

ℹ Safe-chain: Some package versions were suppressed during package metadata resolution due to minimum package age.
  To disable this check, use: --safe-chain-skip-minimum-package-age
```

### Error logs

```shell

```

### Configuration

Mostly `opus-4-8` and `gpt-5.5-fast`.

### Operating system

macOS

### OpenCode version

1.17.13

### oh-my-openagent version

Traced against omo 4.15.0. The issue template says I am using the latest version, but the doctor output reported 4.15.0 loaded while 4.15.1 was latest at the time. Line numbers are omitted because they drift, but the files and symbols above are exact.

EvangelosMoschou:
## Problem

The ultrawork prompt delegated planning to `subagent_type="plan"` which resolves to OpenCode's native plan-mode agent. That agent has `plan_exit` permission and native plan-mode instructions that conflict with omo's planner injection, causing:

1. The subagent enters OpenCode's native plan mode
2. Writes plans under `.opencode/plans/` instead of `.omo/plans/`
3. Calls `plan_exit` which shows a native prompt asking "Implement the plan?"
4. "Yes" makes the subagent implement the plan itself (wrong — should return to parent)
5. "No" or Esc leaves the parent waiting forever

## Root Cause

omo has its own planner (Prometheus) but the ultrawork prompt used `subagent_type="plan"` (OpenCode's native agent) instead of `subagent_type="prometheus"` (omo's planner). The native plan agent gets two conflicting instruction sets:

- omo's planner instructions (via keyword detector injection): act like Prometheus, use `.omo/plans/`, hand off the plan, do not implement
- OpenCode's native plan-mode instructions: use `.opencode/plans/` and always finish with `plan_exit`

Since omo never denies `plan_exit` in this path, the native tool can fire inside the delegated subagent.

## Fix

Change all occurrences of `subagent_type="plan"` to `subagent_type="prometheus"` in all three ultrawork prompt variants (default, gemini, gpt). Prometheus has:

- `prometheus-md-only` hook (enforces .md-only writes)
- No `plan_exit` permission
- No native plan-mode instructions
- Writes plans under `.omo/plans/`

The `isNonOmoAgent("plan")` check in the keyword detector still filters OpenCode's native plan agent from receiving omo injection, so there's no conflict.

## TDD Evidence

- All 89 keyword-detector tests pass (7 files)
- `isPlannerAgent("prometheus")` returns true → planner injection works correctly
- `isNonOmoAgent("plan")` returns true → native plan agent filtered from omo injection
- No test changes needed — existing tests verify the planner injection and non-omo agent filtering

## Files

- `packages/prompts-core/prompts/ultrawork/default.md` (+4/-4) — 4 occurrences of `subagent_type="plan"` → `"prometheus"`
- `packages/prompts-core/prompts/ultrawork/gemini.md` (+2/-2) — 2 occurrences
- `packages/prompts-core/prompts/ultrawork/gpt.md` (+1/-1) — 1 occurrence

Fixes #5850

---
## Summary by cubic
Route Ultrawork planning to Prometheus to stop native plan-agent conflicts and restore correct plan handoff. Also adds MiMo V2.5 Pro support and small reliability fixes.

- **Bug Fixes**
  - Ultrawork: replaced `subagent_type="plan"` with `subagent_type="prometheus"` in `prompts/ultrawork` (`default.md`, `gemini.md`, `gpt.md`) to avoid `.opencode/plans/` writes, `plan_exit` dialogs, and stuck parent sessions (fixes #5850).
  - Slash commands: substitute `$SESSION_ID` and `$TIMESTAMP` in templates and pass `sessionID` through the executor/hook to fix `/handoff` using literal placeholders.
  - Chat params: stop injecting `reasoningEffort` for OpenAI-compatible providers (non-`@ai-sdk/openai`) to prevent `/v1/chat/completions` 400s; keep it for native `@ai-sdk/openai` and when `api.npm` is unset.

- **New Features**
  - Sisyphus: added MiMo V2.5 Pro routing with a model-specific prompt and calibration; detect via `isMimoModel`; no reasoning/thinking opts injected; tests cover alternate model IDs.
  - Model selection: added `mimo-v2.5-pro` to agent requirements and Sisyphus fallback chain (after Kimi K2.6); updated detectors and tests.

<sup>Written for commit 41aa39ce66c37d8dbed3a6063bb49a86e598a76a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/model-core/src/agent-model-requirements.ts
+++ b/packages/model-core/src/agent-model-requirements.ts
@@ -9,6 +9,7 @@ export const AGENT_MODEL_REQUIREMENTS: Record<string, ModelRequirement> = {
         variant: "max",
       },
       { providers: ["opencode-go", "vercel"], model: "kimi-k2.6" },
+      { providers: ["openrouter", "vercel"], model: "mimo-v2.5-pro" },
       { providers: ["kimi-for-coding"], model: "k2p5" },
       {
         providers: [
--- a/packages/model-core/src/model-family-detectors.ts
+++ b/packages/model-core/src/model-family-detectors.ts
@@ -81,6 +81,11 @@ export function isGlmModel(model: string): boolean {
   return modelName.includes("glm")
 }
 
+export function isMimoModel(model: string): boolean {
+  const modelName = extractModelName(model).toLowerCase()
+  return modelName.includes("mimo")
+}
+
 const GEMINI_PROVIDERS = ["google/", "google-vertex/"] as const
 
 export function isGeminiModel(model: string): boolean {
--- a/packages/model-core/src/model-requirements-agents.test.ts
+++ b/packages/model-core/src/model-requirements-agents.test.ts
@@ -17,15 +17,15 @@ describe("AGENT_MODEL_REQUIREMENTS", () => {
     expect(primary?.variant).toBe("high")
   })
 
-  test("sisyphus keeps opus primary before k2p5, kimi-k2.5, gpt-5.5 medium, and big-pickle", () => {
+  test("sisyphus keeps opus primary before kimi-k2.6, mimo-v2.5-pro, k2p5, kimi-k2.5, gpt-5.5 medium, and big-pickle", () => {
     // given
     const sisyphus = AGENT_MODEL_REQUIREMENTS["sisyphus"]
 
     // when
-    const [primary, second, third, fourth, fifth, sixth, last] = sisyphus.fallbackChain
+    const [primary, second, third, fourth, fifth, sixth, seventh, last] = sisyphus.fallbackChain
 
     // then
-    expect(sisyphus.fallbackChain).toHaveLength(7)
+    expect(sisyphus.fallbackChain).toHaveLength(8)
     expect(sisyphus.requiresAnyModel).toBe(true)
     expect(primary).toEqual({
       providers: ["anthropic", "github-copilot", "opencode", "vercel"],
@@ -33,15 +33,16 @@ describe("AGENT_MODEL_REQUIREMENTS", () => {
       variant: "max",
     })
     expect(second).toEqual({ providers: ["opencode-go", "vercel"], model: "kimi-k2.6" })
-    expect(third).toEqual({ providers: ["kimi-for-coding"], model: "k2p5" })
-    expect(fourth?.model).toBe("kimi-k2.5")
-    expect(fifth).toEqual({
+    expect(third).toEqual({ providers: ["openrouter", "vercel"], model: "mimo-v2.5-pro" })
+    expect(fourth).toEqual({ providers: ["kimi-for-coding"], model: "k2p5" })
+    expect(fifth?.model).toBe("kimi-k2.5")
+    expect(sixth).toEqual({
       providers: ["openai", "github-copilot", "opencode", "vercel"],
       model: "gpt-5.5",
       variant: "medium",
     })
-    expect(sixth?.providers[0]).toBe("zai-coding-plan")
-    expect(sixth?.model).toBe("glm-5")
+    expect(seventh?.providers[0]).toBe("zai-coding-plan")
+    expect(seventh?.model).toBe("glm-5")
     expect(last?.providers[0]).toBe("opencode")
     expect(last?.model).toBe("big-pickle")
   })
--- a/packages/omo-opencode/src/agents/sisyphus-agent-factory.test.ts
+++ b/packages/omo-opencode/src/agents/sisyphus-agent-factory.test.ts
@@ -142,6 +142,41 @@ describe("createSisyphusAgent", () => {
     });
   });
 
+  describe("#given a MiMo V2.5 Pro Sisyphus model", () => {
+    test("#when creating the agent #then uses the MiMo-native prompt with bare config", () => {
+      // given
+      const model = "xiaomi/mimo-v2.5-pro";
+
+      // when
+      const agent = createSisyphusAgent(model);
+
+      // then
+      expect(agent.prompt).toContain("running on MiMo V2.5 Pro");
+      expect(agent.prompt).toContain("<mimo_calibration>");
+      expect(agent.prompt).toContain("SEQUENTIAL TOOL CALLS");
+      expect(agent.prompt).toContain("THINKING-TOOL CONFLICT");
+      expect(agent.thinking).toBeUndefined();
+      expect(agent.reasoningEffort).toBeUndefined();
+    });
+
+    test("#when creating the agent with alternate provider naming #then still routes to MiMo variant", () => {
+      // given
+      const models = [
+        "XiaomiMiMo/MiMo-V2.5-Pro",
+        "mimo-v2.5-pro",
+        "opencode-go/mimo-v2-5-pro",
+      ];
+
+      for (const model of models) {
+        // when
+        const agent = createSisyphusAgent(model);
+
+        // then
+        expect(agent.prompt).toContain("running on MiMo V2.5 Pro");
+      }
+    });
+  });
+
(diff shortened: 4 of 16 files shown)


code-yeongyu:
[sisyphus-bot] Thanks for investigating this and for the contribution. I reproduced the current `dev` behavior in an isolated real OpenCode runtime, and this change would route planning to an agent that the `task` tool intentionally rejects.

On current `dev`, the plugin rewrites the native `plan` entry to `mode: "subagent"` with `hidden: true`, strips the native plan prompt/permissions, and inherits only Prometheus model settings. Prometheus itself remains `mode: "primary"`. The task resolver explicitly permits that hidden demoted `plan` alias, while rejecting direct delegation to Prometheus as a coordinator/primary agent.

The focused current-dev regression gate passed 254 tests, including `task(subagent_type="plan")` success and `task(subagent_type="prometheus")` rejection. A real isolated `/agent` response independently confirmed those runtime modes, with the host OpenCode database unchanged before and after.

Because replacing `plan` with `prometheus` would turn valid planning delegations into rejected calls, I’m closing this PR as invalid for the current architecture. I’m leaving #5850 open: its intermittent symptom may still merit a separate reproduction, but this routing change is not a safe fix for it.

PR #5987, opened 2026-07-09
feat(omo-opencode): configure stack-local category agents

### LINKED ISSUE #5986: task(category) should resolve to stack-local agents before permission checks
## Summary

`task(category: ...)` currently treats category as a global shortcut to Sisyphus-Junior. That works for the default OMO stack, but breaks custom or parallel agent stacks where category should mean "use this capability/model profile inside my current stack".

A concrete observed case:

- Parent agent: `Odin (Game Designer)`
- Intended allowed route: Odin Game Designer -> Odin Planner/OpenSpec -> Odin Builder/Reviewer/QA
- Actual route: `task(category: "quick")` spawned `Sisyphus-Junior`

This bypassed the intended stack boundary even though the agent's task policy was meant to allow only Odin-stack delegates.

## Grounded source finding

At current `dev` HEAD `7421e1bcf4793a47c695121d842592ada980dcdb`:

- `packages/omo-opencode/src/tools/delegate-task/category-resolver.ts`
  - `resolveCategoryExecution(...)` returns `agentToUse: SISYPHUS_JUNIOR_AGENT` unconditionally.
- `packages/omo-opencode/src/tools/delegate-task/tools.ts`
  - `createDelegateTask` has `parentContext.agent` available.
  - The `subagent_type` path passes the parent agent to `resolveSubagentExecution(...)`.
  - The `category` path did not pass the parent agent to `resolveCategoryExecution(...)`.
- `packages/omo-opencode/src/features/team-mode/team-runtime/resolve-member.ts`
  - `resolveMember(..., parentAgent?)` has `parentAgent` available.
  - The `subagent_type` path forwards it.
  - The `category` member path did not.

So category routing is caller-blind, while direct subagent routing is caller-aware.

## Why this matters

Categories are useful and should not be disabled. They encode task capability and model/tuning intent:

- `quick` = small/simple work
- `deep` = autonomous research/implementation
- `visual-engineering` = UI/design work
- `writing` = documentation/prose

But a category is not necessarily an agent identity. In multi-stack environments, the desired interpretation is:

```text
caller stack + category -> stack-local worker for that category
```

Examples:

```text
Odin (Planner) + category:quick      -> Odin (Builder)
NeoAgent (Planner) + category:quick  -> NeoAgent (Builder)
Custom PM Agent + category:writing   -> Custom Writer
Default OMO + category:quick         -> Sisyphus-Junior
```

## Why not just forbid categories?

Forbidding `category` would be a regression:

- categories carry model/fallback/prompt tuning;
- they keep task prompts compact and domain-oriented;
- team/council workflows already use category members;
- users should not have to hardcode concrete agent names for every task when they only mean "builder in my stack".

The bug is not category itself. The bug is resolving category directly to one global worker without consulting the caller stack.

## Why not solve this only with a plugin?

A plugin can intercept `tool.execute.before` and rewrite `task({ category })` into `task({ subagent_type })`, but that is only a guard/shim:

- it risks losing category model/tuning behavior unless it reimplements OMO category resolution;
- it depends on hook ordering;
- it does not naturally cover `task_id` continuation ownership;
- it cannot reliably persist the right child-session routing/permission metadata unless the task router participates.

The durable fix belongs in the task routing pipeline before session creation.

## Desired behavior

1. Resolve category to a target agent using the caller context first.
2. Enforce permission against the resolved agent, not the raw category string alone.
3. Persist/propagate child session routing metadata/permissions consistently.
4. Preserve backward compatibility: if no stack-local route is configured, continue using Sisyphus-Junior.
5. Do not hardcode user-specific stacks such as Odin/NeoAgent into core source.

## Proposed minimal generic shape

Add an optional agent override field, e.g.:

```jsonc
{
  "agents": {
    "custom-planner": {
      "category_target_agent": "custom-builder"
    },
    "Custom Planner Display Name": {
      "displayName": "Custom Planner",
      "category_target_agent": "Custom Builder"
    }
  }
}
```

Then `task(category: "quick")` from `custom-planner` keeps category model/tuning resolution, but launches `custom-builder` instead of Sisyphus-Junior.

## Companion PR

I opened a companion PR with a minimal implementation and tests:

- adds `category_target_agent` to agent override schema;
- resolves category targets from `parentAgent + agentOverrides` by config key or display name;
- keeps Sisyphus-Junior as fallback;
- forwards parent agent from top-level `task` and team-mode category members;
- adds tests for configured target routing and fallback behavior.

DanRioDev:
## Summary

- Add `category_target_agent` to agent overrides so category tasks can keep category model/tuning while routing to a stack-local worker.
- Resolve category targets from the parent agent's override key or display name, falling back to Sisyphus-Junior for existing behavior.
- Forward parent-agent context through both top-level `task(category: ...)` and team-mode category members.

Fixes #5986.

## Motivation

`task(category: ...)` is valuable because categories encode capability and model/tuning intent (`quick`, `deep`, `visual-engineering`, `writing`, etc.). But the current category resolver always returns Sisyphus-Junior, which makes category routing global rather than stack-local.

That is unsafe for users with multiple agent stacks. For example, an Odin-stack planner/designer may intend:

```text
Odin stack + category:quick -> Odin builder
```

but the current resolver produces:

```text
any stack + category:quick -> Sisyphus-Junior
```

The PR keeps categories useful without requiring core OMO to hardcode any user-specific stack names.

### Why not hardcode Odin/Neo/custom stacks?

Because every user can define different agent stacks and display names. Hardcoding a few known names would solve one local setup while making the router less generic.

### Why not disable `category` for non-OMO stacks?

That throws away useful behavior: category-specific model selection, fallback chains, prompt append, and concise task delegation.

### Why not solve it as a separate plugin?

A plugin can be a guard/shim, but the router owns category model resolution and child session creation. Rewriting args outside the router risks losing category semantics and does not naturally cover team-mode category members.

## Configuration example

```jsonc
{
  "agents": {
    "custom-planner": {
      "category_target_agent": "custom-builder"
    },
    "custom-builder": {
      "displayName": "Custom Builder"
    }
  }
}
```

With this config, a category task from `custom-planner` still uses the selected category's model/tuning, but launches `custom-builder` instead of the legacy Sisyphus-Junior default.

## Validation

- `bun test packages/omo-opencode/src/tools/delegate-task/category-resolver.test.ts packages/omo-opencode/src/features/team-mode/team-runtime/resolve-member.test.ts`
  - 28 pass / 0 fail / 86 expect calls
  - passed
- `GIT_MASTER=1 git diff --check`
  - passed

---
## Summary by cubic
Make category tasks stack-aware in `omo-opencode` by routing `task(category: ...)` and team category members to a parent-configured subagent, enforcing permission before launch, and filtering skills to the resolved target. This preserves category model/tuning and keeps routing stack-local.

- **New Features**
  - Added `category_target_agent` to agent overrides; resolves by parent key or display name (case-insensitive), defaulting to Sisyphus-Junior.
  - Validate the target exists and is a subagent; reject unknown/coordinator targets; ask `task` permission for the resolved target before launch.
  - In `tool.execute.before`, auto-set `subagent_type` to the parent’s category target; category model, fallbacks, and limits still apply.

- **Bug Fixes**
  - Filter `load_skills` to the resolved target; exclude skills restricted to the placeholder (`Sisyphus-Junior`) when another target is used.
  - Team mode: `team_create` resolves category members from the caller’s agent type, preserving stack-local routing.

<sup>Written for commit ba1a1a0caf7763b7726a8819185df0d6f72ce6af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/5987?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


saimanish1:
Related issue: #6543 (bug(team-mode): category members fail with "Agent Sisyphus-Junior not found" — SISYPHUS_JUNIOR_AGENT hardcodes default display name at import).

Your `resolveCategoryTargetAgent(parentAgent, agentOverrides)` change is exactly the right routing fix — but the default fallback still resolves to `SISYPHUS_JUNIOR_AGENT`, which is `getAgentDisplayName("sisyphus-junior")` evaluated at module import. That returns "Sisyphus-Junior" regardless of runtime displayName overrides or the registered agent name. Reproduced today: all displayNames aligned to internal names, category member still fails with `Agent "Sisyphus-Junior" not found` because OpenCode registers the agent as `sisyphus-junior` (lowercase).

#5987 fixes routing; #6543 is the remaining name-resolution half (resolve the constant against the live registry via `resolveRegisteredAgentName`, or emit the internal config key). Worth coordinating both.

code-yeongyu:
[sisyphus-bot] Routing fix direction is correct, but two blocking concerns: (1) CLA is failing — author needs to sign. (2) @saimanish1 identified a runtime name-resolution bug (#6543): `SISYPHUS_JUNIOR_AGENT` resolves to display name "Sisyphus-Junior" at import time, but OpenCode registers the agent as `sisyphus-junior` (lowercase), causing `Agent "Sisyphus-Junior" not found` at runtime. The fallback path needs `resolveRegisteredAgentName` or the internal config key. Also needs rebase (CONFLICTING with dev).

PR #5997, opened 2026-07-09
fix(omo-opencode): ensure model-fallback dispatches retry before runtime-fallback (fixes main session abort)

UCASRMTG5D:
## Summary

When `runtime_fallback.enabled=true`, model-fallback was completely disabled
by two independent gates, preventing proactive model switching at `chat.message`
time. This PR removes the mutual exclusivity so both fallback mechanisms coexist:
model-fallback sets pending switch state while runtime-fallback handles the
actual retry dispatch.

**But the first fix had a blind spot**: `session.error` events were processed
by `dispatchToHooks()` (which fires runtime-fallback) **before** model-fallback
could handle them. Combined with `canAutoContinue` being gated by
`!isRuntimeFallbackEnabled`, model-fallback had no fallback dispatch capability
when runtime-fallback declined to handle an error — creating a dead zone where
no one dispatches a retry and the session aborts.

This PR fixes both issues.

### Commit 1: remove mutual exclusivity gates

- **`packages/omo-opencode/src/plugin/chat-message.ts`** (-13 lines) —
  Removed `if (!runtimeFallbackEnabled)` gate so model-fallback's
  `chat.message` handler always executes; cleaned up dead
  `isRuntimeFallbackEnabled()` function and unused parameter
- **`packages/omo-opencode/src/plugin/chat-message.test.ts`** (+2/-2) —
  Updated test to expect both hooks to fire instead of skipping model-fallback
- **`packages/omo-opencode/src/plugin/event-model-fallback.ts`** (+7/-2) —
  Removed `!args.isRuntimeFallbackEnabled` from `shouldHandleModelFallback()`
  and `canAutoContinue`; added safety comment that the
  `dispatchInternalPrompt` session reservation gate prevents double-dispatch
- **`packages/omo-opencode/src/plugin/event.model-fallback.test.ts`** (+5/-4) —
  Updated test: model-fallback now dispatches even with runtime-fallback enabled

### Commit 2: fix event ordering (completes the fix)

- **`packages/omo-opencode/src/plugin/event.ts`** (+25/-16) —
  Moved model-fallback `handleSessionError()` **before** `dispatchToHooks()`
  for `session.error` events. This ensures model-fallback sets pending state
  and dispatches a retry first; runtime-fallback's subsequent dispatch attempt
  is silently blocked by the session reservation gate. Removed the now-redundant
  duplicate call that was after `dispatchToHooks`.

## QA & Evidence

- **Tests:** 54 pass, 0 failures across event handler (x4) + model-fallback
  (x3) + runtime-fallback test files
- **LSP diagnostics:** Clean (0 errors on all changed files)

## Configuration Note

After merge, users with `runtime_fallback.enabled: true` who want
model-fallback must add `"model_fallback": true` to their plugin config.

### DIFF
--- a/packages/omo-opencode/src/plugin/chat-message.test.ts
+++ b/packages/omo-opencode/src/plugin/chat-message.test.ts
@@ -192,7 +192,7 @@ describe("createChatMessageHandler - first message hook ordering", () => {
     ])
   })
 
-  test("runs both model and runtime fallback when runtime fallback is enabled", async () => {
+  test("skips model fallback when runtime fallback is enabled", async () => {
     // given
     const hookCalls: string[] = []
     const args = createMockHandlerArgs({
@@ -217,7 +217,7 @@ describe("createChatMessageHandler - first message hook ordering", () => {
     })
 
     // then
-    expect(hookCalls).toEqual(["modelFallback", "runtimeFallback"])
+    expect(hookCalls).toEqual(["runtimeFallback"])
   })
 })
 
--- a/packages/omo-opencode/src/plugin/chat-message.test.ts
+++ b/packages/omo-opencode/src/plugin/chat-message.test.ts
@@ -192,7 +192,7 @@ describe("createChatMessageHandler - first message hook ordering", () => {
     ])
   })
 
-  test("skips model fallback when runtime fallback is enabled", async () => {
+  test("runs both model and runtime fallback when runtime fallback is enabled", async () => {
     // given
     const hookCalls: string[] = []
     const args = createMockHandlerArgs({
@@ -217,7 +217,7 @@ describe("createChatMessageHandler - first message hook ordering", () => {
     })
 
     // then
-    expect(hookCalls).toEqual(["runtimeFallback"])
+    expect(hookCalls).toEqual(["modelFallback", "runtimeFallback"])
   })
 })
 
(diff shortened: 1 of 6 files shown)


UCASRMTG5D:

code-yeongyu:
[sisyphus-bot] Thanks for working on the fallback interaction. I reproduced the exact head (`0f8dfc8942c70dcc725b6717f911605de62c0da8`) against the current `dev` head (`81180f3759c55262a49be6883bb9db5c102e2b4d`), and this ordering change is regressive.

Current `dev` passes the focused ownership/lifecycle suite: 40 passed, 0 failed. A conflict-free synthetic merge of this PR passes 37 but fails all three combined-mode ownership cases:

- `session.error`: adds an unintended main-session abort
- `session.status` retry: causes a double abort
- assistant `message.updated` error: adds an unintended abort

The prompt reservation does not make both handlers mutually exclusive because model-fallback performs its abort before runtime-fallback reaches the reservation gate. Real isolated OpenCode server/SSE QA also confirms the intended current behavior: runtime-fallback owns the retry signal, performs one abort, dispatches exactly one fallback request, completes to idle, and leaves the real OpenCode DB unchanged.

Runtime-fallback should retain ownership whenever `runtime_fallback` is enabled, so I’m closing this PR unmerged rather than landing the conflicting retry path.

PR #6004, opened 2026-07-09
fix(codex): make multi_agent_v2 guard model-aware for GPT-5.6

### LINKED ISSUE #6002: [LazyCodex Bug]: multi-agent-v2-guard force-disables multi_agent_v2, breaking every turn on gpt-5.6 models (collaboration.spawn_agent reserved-schema 400)
### Bug Description

`multi-agent-v2-guard.mjs` (config migration, runs on every Codex SessionStart) unconditionally forces `[features.multi_agent_v2] enabled = false`. This behavior is now **inverted** for the GPT-5.6 model family (terra/sol/luna):

- These models declare `"multi_agent_version": "v2"` in the model catalog (`~/.codex/models_cache.json`) and treat `collaboration.spawn_agent` as a **server-reserved tool** whose schema must match exactly.
- With v2 forced off, the Codex client injects its own legacy spawn_agent function definition, which mismatches the reserved schema, so **every turn fails with HTTP 400** — even a plain "hi" prompt that never touches subagents:

```
Invalid Value: 'tools'. Function 'collaboration.spawn_agent' is reserved
for use by this model and must match the configured schema.
```

The guard's upstream basis (openai/codex#26753, "spawn_agent declares encrypted parameters...") was about v2 being broken when *enabled*. On gpt-5.6 the failure mode is the opposite: v2 *disabled* breaks everything. Since the guard re-applies on every session start, manually fixing config.toml only survives until the next session — users end up in a config tug-of-war they can't see.

Two adjacent problems make recovery harder:

1. `subagent-limit-guard.mjs` writes `agents.max_threads = 1000`, which is **rejected by Codex when v2 is enabled** (`thread/start failed: agents.max_threads cannot be set when features.multi_agent_v2 is enabled`). So even after a user flips `enabled = true`, the other managed setting breaks thread start.
2. The documented opt-out (`LAZYCODEX_CONFIG_MIGRATION_DISABLED=1`) is an environment variable, which is hard to deliver to the Codex **desktop app** (needs `launchctl setenv` + a LaunchAgent for persistence, plus a full app restart). There is no config/state-file opt-out.

### Steps to Reproduce

1. Install OMO/LazyCodex Codex plugin (any version up to 4.16.1; guard also present on current `main`).
2. In `~/.codex/config.toml`, set `model = "gpt-5.6-terra"`.
3. Start any Codex session once (so the migration runs and writes `enabled = false` with the "Managed by LazyCodex" comment).
4. Run: `codex exec --model gpt-5.6-terra --skip-git-repo-check "hi, reply with one word"`
5. Observe the 400 above on every turn.
6. Manually set `enabled = true`, remove `agents.max_threads`, run again → works. Start one more session *without* the opt-out env var → guard reverts the config and the 400 returns.

### Expected Behavior

The guard should be model-aware: skip force-disabling (or force-*enable*) when the selected model's catalog entry declares `multi_agent_version: "v2"` — e.g. read `~/.codex/models_cache.json` and match against the configured `model`. Likewise `subagent-limit-guard` should not write `agents.max_threads` when v2 will be enabled. At minimum, an explicit user-set `enabled = true` should be respected instead of silently reverted every session.

### Actual Behavior

Every turn on gpt-5.6-terra fails with the reserved-schema 400. The migration re-disables v2 on every SessionStart, so manual config fixes silently revert. Recovery requires the env-var opt-out delivered to all launch paths (shell + launchctl + LaunchAgent) plus a desktop app restart.

### LazyCodex version

lazycodex-ai / omo-codex-plugin 4.16.1 (guard verified identical on `main`: `packages/omo-codex/plugin/scripts/migrate-codex-config/multi-agent-v2-guard.mjs`)

### Codex version

codex 0.144.0-alpha.4 (Desktop), codex-cli 0.143.0 — macOS (Darwin 27.0.0)

### Relevant config

```toml
model = "gpt-5.6-terra"

[features]
multi_agent = true

# Written by the guard on every session start:
# Managed by LazyCodex: multi_agent_v2 is re-disabled on every Codex session start
# because enabling it fails every turn with HTTP 400 (openai/codex#26753).
[features.multi_agent_v2]
max_concurrent_threads_per_session = 1000
enabled = false

[agents]
max_threads = 1000   # written by subagent-limit-guard; invalid once v2 is enabled
```

Model catalog entry (`~/.codex/models_cache.json`): `"slug": "gpt-5.6-terra", ..., "multi_agent_version": "v2"`

### Logs

```shell
$ codex exec --model gpt-5.6-terra --skip-git-repo-check "hi, reply with one word"
ERROR: {
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "Invalid Value: 'tools'. Function 'collaboration.spawn_agent' is reserved for use by this model and must match the configured schema.",
    "param": "tools"
  },
  "status": 400
}

# After setting enabled = true but with agents.max_threads still present:
Error: thread/start: thread/start failed: agents.max_threads cannot be set when features.multi_agent_v2 is enabled (code -32600)

# After enabled = true + removing agents.max_threads:
codex
Hello
```

foxion37:
## Summary

- SessionStart config migration no longer force-disables `features.multi_agent_v2` when the selected model declares `multi_agent_version: "v2"` in `models_cache.json` (GPT-5.6 terra/sol family).
- That force-disable was causing every turn to fail with a reserved `collaboration.spawn_agent` schema mismatch (`Invalid Value: 'tools'...`).
- On the V2-preferred path, migration also removes `agents.max_threads`, which Codex rejects while MultiAgentV2 is active, while still writing `max_concurrent_threads_per_session`.
- Legacy openai/codex#26753 force-disable behavior is preserved for unknown/V1 models.

## Changes

- `multi-agent-v2-guard.mjs`: model-aware clear vs force-disable paths.
- `subagent-limit-guard.mjs`: skip/remove `agents.max_threads` when V2 is preferred or already enabled.
- `migrate-codex-config.mjs`: pass resolved `multiAgentVersion` into both guards.
- Tests + `packages/omo-codex/AGENTS.md` updated for the new policy.

## QA & Evidence

- **What was tested:** `node --test packages/omo-codex/plugin/test/migrate-codex-config.test.mjs packages/omo-codex/plugin/test/subagent-limit-migration.test.mjs`
  **Observed result:** 40/40 passed, including new GPT-5.6 clear-disable cases and a full-migration case that drops `agents.max_threads`.
  **Artifact:** local `.omo/evidence/20260710-multi-agent-v2-gpt56/unit-tests.txt` (gitignored)
  **Why sufficient:** Covers both the legacy force-disable path and the GPT-5.6 reserved-schema recovery path.

- **What was tested:** Hermetic `migrateCodexConfig()` against a temp `CODEX_HOME` with `model = "gpt-5.6-terra"` + `models_cache.json` `multi_agent_version: "v2"` and a managed `enabled = false` block.
  **Observed result:** Cleared `enabled = false`, removed `#26753` managed comments, removed `agents.max_threads`, kept `max_concurrent_threads_per_session = 1000`.
  **Artifact:** local `.omo/evidence/20260710-multi-agent-v2-gpt56/migration-dry-run.txt`
  **Why sufficient:** Reproduces the exact config mutation that caused the Desktop/CLI 400.

## Risks & Residuals

- Full isolated Codex Desktop SessionStart QA was not run in this environment; residual accepted because the failing surface is the config mutation covered by unit + hermetic migration evidence.
- Older models without `multi_agent_version: "v2"` still receive the previous force-disable path (mitigated by existing tests).

## Automated Checks

```bash
node --test packages/omo-codex/plugin/test/migrate-codex-config.test.mjs packages/omo-codex/plugin/test/subagent-limit-migration.test.mjs
```

## Related Issues

Closes #6002
Related: code-yeongyu/lazycodex#118

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Make the MultiAgentV2 migration model- and session-aware so GPT-5.6 V2 models aren’t force-disabled, CLI sessions using `codex -m gpt-5.6-*` are respected, and legacy shorthand stays valid TOML. Also clear stale `hide_spawn_agent_metadata = false` on V2 models to avoid reserved `collaboration.spawn_agent` schema 400s while keeping the legacy disable for V1/unknown models.

- **Bug Fixes**
  - `auto-update.mjs` + `migrateCodexConfig()`: read the SessionStart model from stdin, resolve `multi_agent_version` via `models_cache.json`, and pass it to guards; only force-disable when the hook provides a model.
  - `multi-agent-v2-guard.mjs`: prefer the SessionStart model over `config.toml`; clear managed `enabled = false` and comments when the model prefers v2; remove `hide_spawn_agent_metadata = false` on that path; always remove `[features]` boolean shorthand `multi_agent_v2 = true|false`; if the session model is missing or unknown in the catalog on the hook path, skip force-disable; otherwise keep the legacy disable.
  - `subagent-limit-guard.mjs`: skip/remove `[agents].max_threads` when v2 is preferred or when the catalog is missing but the session model is `gpt-5.6*`; still set `[features.multi_agent_v2].max_concurrent_threads_per_session`.
  - Added stdin-reader and GPT-5.6 no-cache regression tests; updated `packages/omo-codex/AGENTS.md`.

<sup>Written for commit 4433ca9a3cba4aa9e20200d066f06a6f32d55873. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


foxion37:
Also tracked on the LazyCodex marketplace side as code-yeongyu/lazycodex#118 (same reserved `collaboration.spawn_agent` schema failure).

code-yeongyu:
Hey @foxion37, looks like the CLA check is failing — could you sign the CLA so we can proceed? The link should be in the check details. Thanks!

foxion37:

foxion37:
Addressed the Codex P1 review:

- SessionStart hook CLI now reads `model` from stdin and passes it into config migration (wins over `config.toml` root `model`).
- When the hook path cannot see a session model, the guard skips legacy force-disable instead of assuming the file default.
- Added regression tests for `codex -m gpt-5.6-terra` with a stale `model = \"gpt-5.5\"` config default.

CLA signed in the previous comment.

code-yeongyu:
Reviewed and verified this against upstream Codex source (`codex-rs` @ `cbdee7976`) before merging:

- `SessionStartCommandInput` does include `model` (`codex-rs/hooks/src/schema.rs`), so the stdin read in `auto-update.mjs` is the right way to see `codex -m gpt-5.6-*` overrides. Codex drops the stdin pipe right after writing the payload, so the 250ms fallback never races in practice.
- `agents.max_threads` rejection is **config-keyed**, not model-keyed: `validate_multi_agent_v2_config()` only errors when `features.enabled(MultiAgentV2)` is true in config (`codex-rs/core/src/config/mod.rs`). So the stale-cache path flagged by the Codex bot produced a *valid* config — but it was still inconsistent with the v2-guard's own GPT-5.6 fallback, and became a footgun the moment a user hand-set `enabled = true`.
- `models_cache.json` on a live machine confirms the per-model design is required: `gpt-5.6-sol`/`terra` are `v2`, but `gpt-5.6-luna` is `v1` — the name-prefix fallback must stay subordinate to the catalog, which this PR gets right.

Pushed `8be72771e` on top addressing the two remaining bot findings:

1. **P2 (real bug):** the `requireSessionModel`/unknown-catalog skip paths returned the config without removing the legacy `[features] multi_agent_v2 = true|false` shorthand; `ensureSubagentConcurrencyLimit()` then appended the `[features.multi_agent_v2]` table, producing scalar-key + same-name-table invalid TOML. Shorthand normalization is now hoisted to the top of `forceDisableMultiAgentV2()` so every path is covered (regression tests parse the migrated output with `tomllib`).
2. **P1 (consistency):** extracted `prefersMultiAgentV2()` and reused it in `subagent-limit-guard.mjs`, so a GPT-5.6 session model with a missing/stale `models_cache.json` now skips/removes `agents.max_threads` exactly like the v2-guard's clear path.

Also added unit tests for `readSessionModelFromStdin` (model payload, malformed/empty stdin, TTY).

Hermetic QA (temp `CODEX_HOME`, real `auto-update.mjs` hook invocation with the SessionStart payload):
- poisoned #6002 config + `gpt-5.6-terra` → managed disable + comments cleared, `agents.max_threads` removed, limit table kept, output parses
- `gpt-5.5` (v1) → legacy #26753 force-disable preserved
- `gpt-5.6-luna` (v1 in catalog) → still force-disabled (catalog wins over name)
- `gpt-5.6-sol` with **no** `models_cache.json` → disable cleared, no `agents.max_threads`
- heal is idempotent across consecutive SessionStarts

`bun run test:codex` green (471 node-test + 335 bun-test). Thanks @foxion37!

Issue #3624, opened 2026-04-24
[Bug]: Passing both category and subagent_type during the execution of a subagents can result in incorrect execution of the subagents.


LiuKexin255:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

opencode: 1.14.22
omo: 3.17.5

In version 3.17.5, I observed that the explore task would be executed by sisyphus union quick instead of the explore sub-agent. When I asked the agent for the reason, the agent provided his calling method `task(category="quick", subagent_type="explore", ...)`. As I mentioned earlier, the allocation of the explore task was incorrect. 

In addition, I also observed that during the execution of F1-F4 acceptance tasks in Atlas, passing the subagentype="unspecified high" parameter would cause the subtask to fail to start. 

### Steps to Reproduce

---

### Expected Behavior

---

### Actual Behavior

---

### Doctor Output

```shell
just same warning 

 oMoMoMoMo Doctor 

 ⚠ 3 issues found:

1. Loaded plugin is outdated
   Loaded 3.17.1, latest 3.17.5.
   Fix: Update: cd "/home/liukexin/.cache/opencode" && bun add oh-my-opencode@latest
   Affects: plugin features

2. GitHub CLI missing
   gh CLI is not installed.
   Fix: Install from https://cli.github.com/
   Affects: GitHub automation

3. Configured models rely on compatibility fallback
[...]

PR #6073, opened 2026-07-12
fix(task): reject both category and subagent_type with explicit error (#6072)

### LINKED ISSUE #6072: task() silently overrides subagent_type when category is also provided; should reject with explicit error like other mutually-exclusive tools
### Problem

In #3624 and #2847 the maintainer established that when both `category` and `subagent_type` are passed to `task()`, `category` wins and `subagent_type` is internally overridden to `sisyphus-junior`. This is documented as an intentional contract.

However, in cross-harness translation paths (e.g. Atlas `start-work` skill mapping a Codex `agent_type` to `subagent_type` while the model also supplies `category`), this "silent override" becomes a **silent failure**:

- No explicit error is returned to the model.
- The caller keeps waiting for a session matching the original `subagent_type`.
- `boulder.json` `task_sessions` remain empty.
- From the outside it is impossible to distinguish "subagent running" from "dispatch never happened".

### Inconsistency with other tools

Other tools that take mutually exclusive parameters reject the conflict explicitly:

- `skill_mcp` throws when more than one of `tool_name`/`resource_name`/`prompt_name` is provided.
- `team_create` throws "Provide exactly one of teamName or inline_spec."
- `call_omo_agent` rejects invalid or missing `subagent_type`.

Only `task()` silently mutates the arguments instead of returning an error.

### Expected behavior

When both `category` and `subagent_type` are present, `task()` should immediately return a typed error so the LLM can self-correct, e.g.:

```
Error: 'category' and 'subagent_type' are mutually exclusive. Provide exactly one parameter.
```

The "category wins" contract is acceptable for human callers who understand the override, but it is the wrong default for an LLM-driven tool interface: the model cannot see that its `subagent_type` was ignored, so it cannot recover.

### Steps to reproduce

Call `task()` with both arguments, for example:

```python
task(category="deep", subagent_type="oracle", prompt="...")
```

Observe:

1. No error is returned.
2. The launched task is routed via `category` with `subagent_type` internally set to `sisyphus-junior`.
3. The original `subagent_type` ("oracle") is silently discarded.

### Suggested fix

Add a pre-dispatch check at the tool boundary before `tool-execute-before` overrides `subagent_type` to `sisyphus-junior`:

1. In `packages/omo-opencode/src/plugin/tool-execute-before.ts` and/or `packages/omo-opencode/src/tools/delegate-task/tool-argument-preparation.ts`, reject when both `category` and `subagent_type` are non-empty.
2. Update tests that currently assert "prefers category over subagent_type" to expect the new error behavior.

### Related issues

- #3624 — established the current "category wins" contract.
- #2847 — requested schema-level mutual exclusivity.

This issue asks for the runtime side of #2847: when the schema (or model) fails to enforce the contract, the tool must still fail loudly rather than silently override.

### Environment

- oh-my-openagent: `dev` branch @ `23312138c49b1ebc8b459564efa45c2a9559665d`

Ruifeng-Zhang:
## Problem

`task()` silently overrides `subagent_type` to `sisyphus-junior` when `category` is also provided. This "category wins" design was intentional per #3624, but the silent override is inconsistent with how other mutually-exclusive tools handle the same case (`skill_mcp`, `team_create` all reject ambiguous input with clear errors).

Under cross-harness mapping scenarios (e.g., Atlas start-work Codex mapping), the silent override manifests as a silent failure — the model thinks it chose an agent but got something else.

## Solution

When both `category` and `subagent_type` are provided (and subagent_type ≠ sisyphus-junior), throw an explicit error:

```
task() received both category and subagent_type.
Provide only ONE of them: category for task delegation (uses Sisyphus-Junior),
or subagent_type for direct agent invocation.
```

### Changes

| File | Change |
|------|--------|
| `tool-execute-before.ts` | Throw Error instead of silent `replaceToolArgs` |
| `tool-argument-preparation.ts` | Belt-and-suspenders validation at prepare stage |
| `tool-description.ts` | Update docs: "ignored" → "will cause an error" |
| `tool-execute-before.test.ts` | Expect error instead of override |
| `tools.test.ts` | Expect error instead of override |
| `metadata-task-id-consistency.test.ts` | Remove unreachable test group (3 tests) |
| `metadata-model-unification.test.ts` | Fix invalid fixture (category + subagent_type) |

## Verification

- ✅ 180 related unit tests pass
- ✅ tsgo typecheck clean (zero errors)
- ✅ Error message matches existing mutually-exclusive tool patterns

## Related

- Closes #6072
- Historically: #3624 (closed, category-wins intentional), #2847 (closed, schema enforcement)

---
## Summary by cubic
Reject ambiguous `task()` calls by throwing an error when both `category` and `subagent_type` are provided, replacing the prior silent override. This prevents misrouted tasks and aligns with other mutually exclusive tools (closes #6072).

- **Bug Fixes**
  - Throw an error when both `category` and `subagent_type` are set (allowed only if `subagent_type` is `"sisyphus-junior"`).
  - Validate in `tool-execute-before` and during argument preparation; update tool docs; update tests and remove unreachable cases.

- **Migration**
  - Send only one: use `category` to delegate to Sisyphus-Junior, or `subagent_type` for direct agent calls.
  - If you currently send both, drop one to avoid the new error.

<sup>Written for commit 5d4449dcbc47b82117739aff2a7dfbe7eef9213b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6073?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for the PR, @Ruifeng-Zhang, and for the detailed writeup tying this to #6072.

One thing before it can land: at 145 lines this is larger than the narrow, minimal-change shape `fix()` PRs are expected to keep here, so they stay easy to review and revert. Could you tighten the diff to the smallest guard that resolves #6072, and split anything adjacent into a separate PR?

Happy to re-review once it is narrower.

Ruifeng-Zhang:
Thanks for the review, @MoerAI!

Done — trimmed the diff from 145 lines down to 42 lines. The metadata test cleanup (3 unreachable tests in `metadata-task-id-consistency.test.ts`, ~100 lines) has been split into a separate PR: #6120.

This PR now contains only the smallest guard that resolves #6072:
- Error throw in `tool-execute-before.ts` + `tool-argument-preparation.ts`
- Doc update in `tool-description.ts`
- Direct test updates in `tool-execute-before.test.ts` and `tools.test.ts`
- One-line fixture fix in `metadata-model-unification.test.ts`

Ready for re-review.

MoerAI:
Thanks @Ruifeng-Zhang for the quick turnaround on narrowing the diff (145 to 42 lines, with the test cleanup split into #6120) - the shape is exactly what was asked, and the guard itself reads clean.

Before this can merge there is one thing I cannot decide at the maintainer level: the current "category wins, subagent_type is overridden" behavior was documented and test-locked as an intentional contract by the project owner in the resolution of #3624 (v4.1.0). This PR reverses that contract (silent override becomes a hard error, and the tests that lock the old behavior are flipped). Your silent-failure argument in #6072 is a real one, especially the cross-harness mapping case, but revising an owner-documented contract is a product decision, not a bug-fix call.

@code-yeongyu, when you have a moment: #6072 argues the #3624 "category wins" contract produces silent failures in cross-harness paths and is inconsistent with how skill_mcp and team_create reject ambiguous input. Could you rule on whether the contract should change as proposed here? If yes, I will run the full verification pass on this PR and land it; if the current behavior should stay, #6072 can be closed as by-design and a warning-instead-of-error variant might be the closer fit.

Leaving this open and waiting on that decision. No action needed from you right now, @Ruifeng-Zhang.

MoerAI:
Thanks for the PR, @Ruifeng-Zhang, and for writing up #6072 so precisely.

I checked the change. It replaces the silent override in `packages/omo-opencode/src/tools/delegate-task/tool-argument-preparation.ts` with a thrown error, adds a matching guard in `packages/omo-opencode/src/plugin/tool-execute-before.ts` that still allows `subagent_type: "sisyphus-junior"` alongside a category, and updates the tool description from "subagent_type is ignored" to "Providing both will cause an error". Your argument in #6072 is a real one: in cross-harness translation the silent override turns into a silent failure, and `skill_mcp`, `team_create` and `call_omo_agent` all reject mutually exclusive arguments explicitly, so `task()` is the odd one out.

The reason I am not merging this myself is the one you named in the issue: #3624 and #2847 established the current behavior as an intentional documented contract. Flipping it from lenient to throwing is a product decision, and any existing prompt or skill that passes both arguments would start hard-erroring rather than degrading. That call belongs to the owner.

I also want to note for the record that this PR and #6120 are coupled. #6120 removes the tests that this change makes unreachable, so they need to land together or not at all.

@code-yeongyu your call on this pair. The question: do you want `task()` to keep the documented lenient override from #3624 and #2847, or move to an explicit error as #6072 argues, accepting that callers currently passing both will now fail loudly? If you want the error, #6073 and #6120 should merge together.

code-yeongyu:

MoerAI (changes requested):
Blocking: the current guard regresses ordinary category-only `task()` calls through the composed production path.

`tool.execute.before` injects lowercase `sisyphus-junior`, while `prepareDelegateTaskArgs()` compares it to the display-name constant `Sisyphus-Junior`. The strings differ, so the new guard throws for a normal category-only call. The existing tests cover the hook and tool separately and miss this interaction.

Please normalize the compared agent identity and add a composed regression that passes the hook-mutated args into `tool.execute()`.

Also align the public contract: the description says every `category` plus `subagent_type` pair errors, while both guards exempt Sisyphus-Junior. Either reject every pair or document and test the exception.

PR #6083, opened 2026-07-13
fix(hooks): deliver category-skill reminder as a synthetic text part

### LINKED ISSUE #6001: [Bug]: Category+Skill Reminder injected into tool outputs corrupts results
## Bug Description

The `oh-my-openagent` plugin appends a `[Category+Skill Reminder]` block to the end of tool outputs (e.g. `bash`, `read`). This corrupts the tool result from the model's perspective — the reminder becomes part of the returned content, not a separate system message.

## Observed Behavior

Running any `bash` or `read` command returns the actual output followed by:

    [Category+Skill Reminder]

    **Built-in**: playwright, frontend, git-master, review-work, remove-ai-slops, init-deep, debugging, security-research (+30 more)
    **⚡ YOUR SKILLS (PRIORITY)**: security-review, adapt, agent-eval, ... (+284 more)

    > User-installed skills OVERRIDE built-in defaults. ALWAYS prefer YOUR SKILLS when domain matches.

    task(category="visual-engineering", load_skills=["security-review"], run_in_background=true)

This appears in **every** tool output, making results look "corrupted" and causing the model to reference it in responses (e.g. "the tool output got corrupted").

## Expected Behavior

The reminder should either:

1. Be injected as a **separate system message** (not appended to tool results), or
2. Be **opt-in/opt-out** via a config flag in `oh-my-openagent.json`, or
3. Only appear at session start / after compaction (not on every tool call)

## Environment

- OpenCode with `oh-my-openagent@latest` plugin
- Config: `~/.config/opencode/tui.json` and `opencode.json` both load `oh-my-openagent@latest`
- Model: `ollama-cloud/glm-5.2`
- OS: Windows 11

## Impact

- Tool outputs look corrupted to the model and user
- Model wastes tokens acknowledging/explaining the "corruption"
- Makes long sessions noisy and harder to follow
- Related to #3676 (bulky context) but this is specifically about **tool output injection**, not general context bloat

## Suggested Fix

Add a config option to `oh-my-openagent.json`:

    {
      "experimental": {
        "skill_reminder": false
      }
    }

Or inject the reminder as a dedicated system message type that doesn't pollute tool results.

H-TTTTT:
## Summary

Narrowed to the core fix for the reported symptom (review concerns 1-2), per maintainer request to keep this `fix()` small and independently revertable.

- Stop appending the category+skill reminder to `tool.execute.after` output. Appending corrupted tool results (the reminder text leaked into Read/Bash/edit output sent to the model).
- Queue the reminder on the third delegatable tool call and inject exactly one synthetic text part before the latest real user text on the next `experimental.chat.messages.transform`. Tool output is now byte-identical.

## Scope

Two source files, +79/-20 (`hooks/category-skill-reminder/hook.ts`) and +2 (`plugin/messages-transform.ts`, wiring the hook into the transform tier).

## Deferred to separate follow-up PRs

The broader reminder-delivery lifecycle from the original draft is split out:

- Delegated-child identity preservation under runtime fallback (concern 3).
- Compaction pause/resume/recovery (concern 4).
- Disposal/session-deletion cleanup (concern 5).

Each already has its own tests and will land in its own PR.

## Validation

- 41 focused category-reminder + messages-transform tests pass (`bun test`).
- 345 broader plugin / call-omo-agent / category-reminder tests pass with no regressions.
- `tsgo --noEmit -p packages/omo-opencode/tsconfig.json` passes.

## Notes

- This commit is local on the branch; not pushed yet.
- Live isolated OpenCode QA (the `opencode-qa` skill / SSE hook proof) will be re-run before push, scoped to the `tool.execute.after` + `experimental.chat.messages.transform` surfaces touched here.

Fixes #6001

### DIFF
--- a/packages/omo-opencode/src/hooks/category-skill-reminder/hook.ts
+++ b/packages/omo-opencode/src/hooks/category-skill-reminder/hook.ts
@@ -1,25 +1,18 @@
 import type { PluginInput } from "@opencode-ai/plugin"
+import type { Message, Part } from "@opencode-ai/sdk"
 import type { AvailableSkill } from "../../agents/dynamic-agent-prompt-builder"
 import { getSessionAgent } from "../../features/claude-code-session-state"
-import { log } from "../../shared"
+import { isRealUserTextPart, log } from "../../shared"
 import { getAgentConfigKey } from "../../shared/agent-display-names"
 import { resolveSessionEventID } from "../../shared/event-session-id"
 import { buildReminderMessage } from "./formatter"
 
-/**
- * Target agents that should receive category+skill reminders.
- * These are orchestrator agents that delegate work to specialized agents.
- */
 const TARGET_AGENTS = new Set([
   "sisyphus",
   "sisyphus-junior",
   "atlas",
 ])
 
-/**
- * Tools that indicate the agent is doing work that could potentially be delegated.
- * When these tools are used, we remind the agent about the category+skill system.
- */
 const DELEGATABLE_WORK_TOOLS = new Set([
   "edit",
   "write",
@@ -29,12 +22,9 @@ const DELEGATABLE_WORK_TOOLS = new Set([
   "glob",
 ])
 
-/**
- * Tools that indicate the agent is already using delegation properly.
- */
 const DELEGATION_TOOLS = new Set([
-   "task",
-   "call_omo_agent",
+  "task",
+  "call_omo_agent",
 ])
 
 interface ToolExecuteInput {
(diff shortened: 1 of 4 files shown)


H-TTTTT:

MoerAI (dismissed):
Thanks for the PR, @H-TTTTT. The isolated OpenCode QA run and the reminder-visibility trace in the Validation section are exactly the kind of evidence this repo asks for, so thank you for that diligence.

The blocker is scope: at +785/-401 across 8 files this is a redesign of the reminder delivery lifecycle rather than a narrow fix. The summary itself lists five separable concerns: (1) stop appending reminders to tool result text, (2) the new queued synthetic-text-part delivery, (3) delegated-child identity preservation under runtime fallback, (4) compaction pause/resume/recovery, and (5) disposal/session-deletion cleanup. Convention here is that fix() PRs stay narrow (ideally under 50 lines, 1-2 files) so they can be reviewed and reverted independently; I recently asked the same of a 145-line PR (#6073), so I want to apply the standard consistently.

Could you split this so the core fix for the reported symptom (reminders leaking into tool output, roughly concerns 1 and 2, or even 1 alone if 2 can be deferred) lands first, with the fallback-identity, compaction, and disposal hardening as separate follow-up PRs? Each piece already appears to have its own tests, so the split should mostly be mechanical.

Also, please link the issue this fixes in the body (Fixes #N); if none exists yet, a short issue describing the observed leak makes the fix auditable against a report.

Happy to re-review the narrowed version quickly. Leaving open as request-changes.

code-yeongyu (approved):
Reviewed exact head `42f5078b8dc99493135e95c788c0e2c5df9217f2` after the requested scope reduction. The earlier change request targeted superseded head `7696b43193896b15fad5928701f6b04a7fa19103`; the current commit is narrowed to the core #6001 fix.

No blocking findings. Focused tests pass 41/41, the broader OpenCode adapter suite passes 698/698 across 84 files, package typecheck passes, and isolated real OpenCode QA proves all three tool outputs remain byte-identical while exactly one separate reminder text part reaches the next provider request. Host DB isolation was preserved (`5751 -> 5751`), and the exact head merges cleanly with current `dev`.

PR #6084, opened 2026-07-13
fix: migrate remaining readers to plural getOpenCodeConfigDirs()

Zeno2019:
## Summary

Follow-up to #3875. The plural `getOpenCodeConfigDirs()` was introduced there and `loadPluginConfig()` was migrated, but 4 reader sites still call the singular `getOpenCodeConfigDir()`. In environments where `OPENCODE_CONFIG_DIR` is injected (e.g. Orca-managed terminals), these readers only see the injected dir and miss the user config at `~/.config/opencode/` — so agents, skills, hooks, and plugin lists silently fall back to defaults.

This PR migrates the 4 remaining multi-path readers to the plural function. It also fixes a separate bug in `loadOpencodePlugins()` where `XDG_CONFIG_HOME` was ignored (the old code hardcoded `os.homedir() + "/.config"`).

Related to #3846.

## Changes

- `opencode-config-dir.test.ts`: 2 contract tests for the plural function — dedup when `OPENCODE_CONFIG_DIR === XDG default`, and single-element return for non-opencode binaries.
- `opencode-config-agents-reader.ts` (claude-code-compat-core): `getOpenCodeConfigDir()` → `getOpenCodeConfigDirs().flatMap()`. Agents from both custom and default dirs are now merged.
- `opencode-config-skills-reader.ts` (skills-loader-core): same migration.
- `claude-code-hooks/config-loader.ts` (omo-opencode): `getUserConfigPath()` → `getUserConfigPaths()`. `loadPluginExtendedConfig()` now iterates all user config files (default first, custom last so custom wins) and merges `disabledHooks`. Project config still overrides all. This is a behavior change — previously only one user config file was read.
- `load-opencode-plugins.ts` (omo-opencode): replaced the self-contained `getConfigPaths()` (which hardcoded `~/.config`, manually read `OPENCODE_CONFIG_DIR` and `APPDATA`) with `getOpenCodeConfigDirs()`. Plugin discovery order changes from `[project, default, custom, APPDATA]` to `[project, custom, default]` — custom now wins over default on dedup, which matches the override semantics users expect.

Pattern B (6 diagnostic sites in doctor checks) and Pattern C (`claude-tasks/storage.ts`) are intentionally not in this PR. Pattern C uses single-dir semantics on purpose (`custom ?? default` is the correct primary location for a canonical storage path).

## Verification

```
bun run typecheck
bun test packages/omo-opencode/src/shared/opencode-config-dir.test.ts
bun test packages/omo-opencode/src/shared/load-opencode-plugins.test.ts
bun test packages/omo-opencode/src/hooks/claude-code-hooks/config-loader.test.ts
bun test packages/claude-code-compat-core/src/features/claude-code-agent-loader/
bun test packages/skills-loader-core/src/features/opencode-skill-loader/
```

All pass. The 4 failures in `opencode-skill-loader/` are pre-existing (`git-master async precedence`, `resolveMultipleSkillsAsync`) and unrelated to config-dir changes.

## Related Issues

- Follow-up to #3875
- Related to #3846

---

This is my first contribution here. If the approach or scope needs adjustment, happy to revise.

### DIFF
--- a/packages/claude-code-compat-core/src/features/claude-code-agent-loader/opencode-config-agents-reader.test.ts
+++ b/packages/claude-code-compat-core/src/features/claude-code-agent-loader/opencode-config-agents-reader.test.ts
@@ -7,13 +7,24 @@ import { readOpencodeConfigAgents } from "./opencode-config-agents-reader"
 
 describe("readOpencodeConfigAgents", () => {
   let mockGlobalConfigDir = ""
+  let mockXdgHomeDir = ""
+  let origXdgConfigHome: string | undefined
 
   beforeEach(() => {
     mockGlobalConfigDir = fs.mkdtempSync(path.join(os.tmpdir(), "opencode-mock-global-"))
+    mockXdgHomeDir = fs.mkdtempSync(path.join(os.tmpdir(), "opencode-mock-xdg-"))
+    origXdgConfigHome = process.env.XDG_CONFIG_HOME
     process.env.OPENCODE_CONFIG_DIR = mockGlobalConfigDir
+    process.env.XDG_CONFIG_HOME = mockXdgHomeDir
   })
 
   afterEach(() => {
+    if (origXdgConfigHome !== undefined) {
+      process.env.XDG_CONFIG_HOME = origXdgConfigHome
+    } else {
+      delete process.env.XDG_CONFIG_HOME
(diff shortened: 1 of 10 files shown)


Zeno2019:

PR #6118, opened 2026-07-14
fix(todo-continuation-enforcer): break directive-response loop

### LINKED ISSUE #6109: 🐛 [BUG] `todo-continuation-enforcer` hook causes system-directive spam loop, mid-stream text fragmentation, and unresponsive sessions
# 🐛 [BUG] `todo-continuation-enforcer` hook causes system-directive spam loop, mid-stream text fragmentation, and unresponsive sessions

## Summary

The `todo-continuation-enforcer` hook in `oh-my-openagent` 4.16.0 fires
`session.idle` → re-injects `[SYSTEM DIRECTIVE: OH-MY-OPENCODE - TODO CONTINUATION]`
into the conversation whenever there are incomplete todos and the assistant
just finished a turn. When the model responds to the directive by **talking
about the directive** (e.g. explaining it, trying to compress, or describing
the loop) rather than advancing a todo, the hook fires again on the next
`session.idle`. The result is:

- 🔁 Endless directive push → response push → directive push loop
  (no real work, no user input consumed)
- ✂️ Mid-stream text fragmentation — user sees the assistant's response
  cut into short fragments (e.g. "...ible ranges）...") as the new
  directive is inserted before the previous response finished
- 🛑 Sessions become unresponsive — even the user typing a stop message
  doesn't break the loop
- 💸 Direct cost — every directive cycle burns an LLM call with a
  non-trivial prompt; long sessions waste tens of thousands of tokens
  on the loop alone

## Environment

- opencode: **1.17.13**
- oh-my-openagent: **4.16.0**
- Model: any (reproduced with `minimax-cn/MiniMax-M3` / `M2.7-highspeed`,
  not model-specific)
- OS: macOS Darwin

## Reproduction

Minimal conditions:

1. Fresh `oh-my-openagent.json` that does **not** list
   `todo-continuation-enforcer` in `disabled_hooks` (this is the
   default — see config below).
2. A non-trivial task that creates ≥ 3 todos.
3. The assistant response either:
   - describes the loop / system directives in the text, OR
   - calls `compress` and the compressed content includes the
     previous system directives themselves, OR
   - hits a model error / context overflow, OR
   - simply ends a turn with a "let me think about this" text without
     advancing a todo.
4. `session.idle` fires.

The hook then:
1. Sees incomplete todos (`getIncompleteCount > 0`)
2. Starts a 2-second countdown (`COUNTDOWN_SECONDS = 2`)
3. After 2s calls `injectContinuation(...)` which dispatches an
   internal user message containing the directive prompt
4. The assistant responds
5. `session.idle` fires again
6. goto 1

The hook will keep going until `stagnationCount` reaches
`MAX_STAGNATION_COUNT = 3`. In practice this means **3+ directives in
a row, each with a fresh LLM call**.

## Expected

- A single `TODO CONTINUATION` directive per genuine "agent went idle
  on a multi-step task" event
- No directive re-injection if the assistant's previous turn was itself
  caused by a directive (i.e. no recursion)
- The hook should respect a user-typed stop / message that appears after
  a directive was injected
- `MAX_STAGNATION_COUNT = 3` should not lead to **3 actual model turns**
  being spent on the loop, but the design implies it does

## Actual

The directive gets re-injected up to `MAX_STAGNATION_COUNT` times
without considering whether the previous turn was itself a directive
response. The `compaction-guard` (60s) and `continuation-cooldown` (5s)
don't help — by the time the guard is up, the model is already mid-loop.

The user sees their assistant's output fragmented as the new directive
arrives mid-stream:

```
让我压缩一下 m0058-m006
6 的内容（主要是 system directive 循环推送
），然后给一个简洁的结束语。
但是
先看 m0067 是 "internal Continue"，
...
```

(text broken at arbitrary points because the previous response was
interrupted by the new directive being inserted)

## Real-world evidence (from a single user session)

Session: `ses_0a0eb0a1dffeznbgvuTjexx21A` (✊墙壁开关系列产品import准备)

**10 `[SYSTEM DIRECTIVE: OH-MY-OPENCODE - TODO CONTINUATION]` injections** in
~2 hours. Timeline (UTC+8):

| Time | Source | Content (truncated) |
|---|---|---|
| 05:33:48 | directive | `Incomplete tasks remain in your todo list. Continue working on the next pending task.` |
| 05:39:19 | directive | (same) |
| 05:39:39 | directive | (same) — 4 in a row, 20s apart |
| 05:39:50 | directive | (same) |
| 05:45:40 | **user** | "现在历史的对话，内容太短了，需要弄一下" |
| 06:21:55 | directive | (re-injected 36 min later) |
| 06:22:16 | directive | (same) |
| 06:22:57 | directive | (same) — 3 in a row, 20s apart |
| 07:32:23 | directive | (re-injected 1h+ later) |
| 07:35:57 | **user** | "你在进行下一步前先严肃的看看上下文，已经错乱了。" |
| 07:36:50 | directive | (re-injected **AFTER** user message) |
| 07:37:41 | directive | (re-injected again) |

The last two are particularly bad — the user explicitly told the
assistant the context is broken, then **two more directives were
injected** that ignored the user.

## Root cause (code refs)

`packages/omo-opencode/src/hooks/todo-continuation-enforcer/`

| File | Issue |
|---|---|
| `index.ts` line 80087 | `CONTINUATION_PROMPT` is a fixed string with no notion of "previous turn was a directive response" |
| `constants.ts` line 80101 | `MAX_STAGNATION_COUNT = 3` is too low — burns 3 real LLM turns on a loop that's clearly going nowhere |
| `idle-event.ts` line 80686 | The `inFlight` check is racy — multiple idle events can queue |
| `idle-event.ts` line 80707-80746 | The "compaction guard" is a 60s lockout, but `CONTINUATION_COOLDOWN_MS = 5000` (5s) bypasses it after the first minute |
| `handler.ts` line 80824-80833 | On `message.updated` (assistant) the hook only clears the countdown — it does **not** mark the previous turn as "directive response" to prevent re-injection |
| `continuation-injection.ts` line 80266-80310 | `injectContinuation` does not check if the most recent assistant turn was itself produced by a directive. The `latestAssistantTurnBlocksInternalPrompt` check exists in `idle-event.ts` (line 80656) but appears to be insufficient or has a gap |

The key gap: **there is no "directive → response → directive again" loop
breaker.** The hook re-arms itself whenever the session is idle and
todos are incomplete, with no memory of "the assistant's last turn was
spinning on my own directives."

## Suggested fix (in `oh-my-openagent`)

1. In `idle-event.ts`, before starting the countdown, check whether the
   most recent assistant turn was preceded by a `TODO CONTINUATION`
   directive. If yes, increment a `directiveResponseCount` and only
   re-inject if `directiveResponseCount < 1` (i.e. one retry max).
2. Add a directive-response detector that recognizes when the model is
   **describing the loop** (e.g. mentions "system directive", "压缩
   m00xx", "循环推送", "the system is questioning my completion
   claim") and aborts injection early.
3. Raise `MAX_STAGNATION_COUNT` consideration to also factor in
   **user-visible disruption** (text fragmentation), not just
   `incompleteCount` non-progress.
4. Respect a fresh user message: if a non-synthetic user message
   arrived after the directive was injected, **do not re-inject**.

## User-side workaround (verified)

Until the hook is fixed, add to `disabled_hooks` in
`~/.config/opencode/oh-my-openagent.json`:

```json
{
  "disabled_hooks": [
    "agent-usage-reminder",
    "category-skill-reminder",
    "legacy-plugin-toast",
    "auto-update-checker",
    "todo-continuation-enforcer"
  ]
}
```

After this change, no more `[SYSTEM DIRECTIVE: OH-MY-OPENCODE - TODO
CONTINUATION]` injections occur. The downside: the user has to
manually drive the agent to the next todo (no auto-resume). This is
the lesser evil compared to the loop.

## Severity

**High** — silent token waste, broken user experience, and at the time
of writing a workaround requires editing JSON config and the loop
itself isn't surfaced to the user as an error.

youttiao:
## Summary

Fix a directive-response loop in the `todo-continuation-enforcer` hook that re-injects TODO CONTINUATION (and other OMO system directives) into sessions where the assistant has already responded to a previous injection. Closes #6109.

## Root cause

`handleSessionIdle` in `packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.ts` had a guard for assistant-turn blocking (`latestAssistantTurnBlocksInternalPrompt`) but no guard for "the most recent user message is itself a system directive". When the assistant replied to a previous TODO CONTINUATION with a short acknowledgment (e.g. "no longer responding"), the assistant-turn predicate was satisfied, and the next `session.idle` event re-injected the same directive. With `COUNTDOWN_SECONDS=2` and the default cooldown, this produced 4+ consecutive injections within one minute on a real session.

## Reproduction

See the timeline in #6109: a single OpenCode session showed 10 directive injections in 2 hours, including 4 within 1 minute (05:33-05:39) and 2 more after the user complained at 07:35:57.

## Fix

Add a new guard in `handleSessionIdle` (alongside the existing assistant-turn guard):

```ts
if (lastUserMessageIsSystemDirective(prefetchedMessages)) {
  log(`[${HOOK_NAME}] Skipped: last user message is a system directive (loop breaker)`, { sessionID })
  return
}
```

`lastUserMessageIsSystemDirective` walks the messages in reverse, finds the most recent user message, and returns true if any of its text parts starts with `SYSTEM_DIRECTIVE_PREFIX`. It uses the existing `isSystemDirective` helper from `@omo-opencode/shared/system-directive`, so it correctly recognizes all directive types (TODO CONTINUATION, RALPH LOOP, BOULDER CONTINUATION, etc.).

## Test

Added a new test in `idle-event.test.ts` that simulates the exact loop-breaker scenario: real user message → real assistant response → injected TODO CONTINUATION (synthetic user) → short assistant acknowledgment. The test asserts that `trackContinuationProgress` is never called and the countdown never starts.

```
(pass) handleSessionIdle > skips todo continuation when the last user message is a system directive (loop breaker, #6109) [0.20ms]
```

All 6 tests in the file pass. `bun run typecheck` passes.

## Diff

```
.../todo-continuation-enforcer/idle-event.test.ts  | 59 ++++++++++++++++++++++
.../hooks/todo-continuation-enforcer/idle-event.ts | 22 ++++++++
2 files changed, 81 insertions(+)
```

---
## Summary by cubic
Stops the directive-response loop in the TODO continuation enforcer by skipping injections when the last user message is an OMO system directive. Prevents repeated TODO CONTINUATION spam and aligns with #6109.

- **Bug Fixes**
  - Added a guard in `handleSessionIdle` to skip continuation when the latest user message is a system directive, using `isSystemDirective` from `@omo-opencode/shared/system-directive`.
  - Introduced `lastUserMessageIsSystemDirective` to scan messages in reverse and detect directive user turns.
  - Added a test that mirrors the loop scenario and asserts no countdown or tracking is triggered.

<sup>Written for commit 410c133bd1606101b085c71b809850ae662f4d00. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.test.ts
+++ b/packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.test.ts
@@ -228,4 +228,63 @@ describe("handleSessionIdle", () => {
       store.cancelCountdown(sessionID)
     }
   })
+
+  it("skips todo continuation when the last user message is a system directive (loop breaker, #6109)", async () => {
+    // given
+    const sessionID = "ses_directive_loop_breaker"
+    const { store, trackCalls, state } = createStateStore()
+    const ctx = {
+      client: {
+        session: {
+          messages: async () => ({
+            data: [
+              {
+                info: { role: "user" },
+                parts: [{ type: "text", text: "do the original task", synthetic: false }],
+              },
+              {
+                info: { role: "assistant" },
+                parts: [{ type: "text", text: "doing the original task..." }],
+              },
+              {
+                info: { role: "user" },
+                parts: [
+                  {
+                    type: "text",
+                    text: "[SYSTEM DIRECTIVE: OH-MY-OPENCODE - TODO CONTINUATION]\n\ncontinue",
+                    synthetic: true,
+                  },
+                ],
+              },
+              {
+                info: { role: "assistant", finish: "stop" },
+                parts: [{ type: "text", text: "no longer responding" }],
+              },
+            ],
+          }),
+          todo: async () => ({
+            data: [
+              { id: "todo-1", content: "Finish init-deep", status: "pending", priority: "high" },
+            ],
+          }),
+        },
+      },
+      directory: "/tmp/test",
+    }
+
+    try {
+      // when
+      await handleSessionIdle({
+        ctx: ctx as never,
+        sessionID,
+        sessionStateStore: store,
+      })
+
+      // then — no continuation path reached
+      expect(trackCalls).toEqual([])
+      expect(state.countdownStartedAt).toBeUndefined()
+    } finally {
+      store.cancelCountdown(sessionID)
+    }
+  })
 })
--- a/packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.ts
+++ b/packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.ts
@@ -5,6 +5,7 @@ import { normalizeSDKResponse } from "../../shared"
 import { getAgentConfigKey } from "../../shared/agent-display-names"
 import { log } from "../../shared/logger"
 import { latestAssistantTurnBlocksInternalPrompt } from "../../shared/prompt-async-gate/pending-tool-turn"
+import { isSystemDirective } from "../../shared/system-directive"
 
 import { isLastAssistantMessageAborted } from "./abort-detection"
 import { acknowledgeCompactionGuard, isCompactionGuardActive } from "./compaction-guard"
@@ -101,6 +102,10 @@ export async function handleSessionIdle(args: {
     }
     if (latestAssistantTurnBlocksInternalPrompt(prefetchedMessages)) {
       log(`[${HOOK_NAME}] Skipped: pending internal continuation response`, { sessionID })
+      return
+    }
+    if (lastUserMessageIsSystemDirective(prefetchedMessages)) {
+      log(`[${HOOK_NAME}] Skipped: last user message is a system directive (loop breaker)`, { sessionID })
       return
     }
   } catch (error) {
@@ -233,4 +238,21 @@ export async function handleSessionIdle(args: {
     sessionStateStore,
     isContinuationStopped,
   })
+}
+
+// Returns true when the most recent user message is itself an OMO system
+// directive. Used to break directive-response loops (#6109).
+export function lastUserMessageIsSystemDirective(
+  messages: ReadonlyArray<{ info?: { role?: string }; parts?: ReadonlyArray<{ type?: string; text?: string; synthetic?: boolean }> }> | undefined,
+): boolean {
+  if (!messages || messages.length === 0) return false
+  for (let i = messages.length - 1; i >= 0; i--) {
+    const m = messages[i]
+    if (m?.info?.role !== "user") continue
+    const parts = m.parts ?? []
+    return parts.some(
+      (p) => p.type === "text" && typeof p.text === "string" && isSystemDirective(p.text),
+    )
+  }
+  return false
 }


code-yeongyu:
[sisyphus-bot] Thank you for isolating the directive-response loop and adding a focused regression test. I re-reviewed the exact current head `410c133bd1606101b085c71b809850ae662f4d00` after #6109 was fixed by merged PR #6149.

This approach is now redundant, and it is not safe to merge independently:

- the new guard returns before todo progress is inspected, so a directive-driven turn that legitimately completes work can suppress later continuation while pending todos remain;
- a genuine non-synthetic user interruption is not retained, so the hook can re-arm after the user asks it to stop or inspect context;
- split `message.updated` / `message.part.updated` provenance is not classified;
- the boundary is not rechecked immediately before dispatch, leaving the countdown race open.

Independent exact-head probes failed all four boundaries. In an isolated real OpenCode 1.17.13 run, this head injected again after a genuine user interruption. Current `dev` (`400463b54db36fb9647eec6dee870d69e4d6ea8a`), which includes #6149, passed the same real-harness scenario with the live OpenCode DB unchanged, and its focused hook suite passed 134/134 tests.

Closing this PR because the issue is already fixed by the broader provenance-based implementation in #6149, and merging this narrower guard would regress the corrected behavior. Thanks again for the report and the useful initial direction.

PR #6120, opened 2026-07-14
test(task): remove unreachable requested_subagent_type metadata tests

Ruifeng-Zhang:
## Problem

PR #6073 makes `task()` reject the combination of `category` + `subagent_type` with an explicit error. This removes the code path where `requested_subagent_type` metadata was preserved alongside a category.

## Change

Remove 3 now-unreachable tests in `metadata-task-id-consistency.test.ts` that covered the old behavior of preserving `requested_subagent_type` when both `category` and `subagent_type` were provided.

These tests would never be reached after #6073 because the combination now throws before metadata is published.

## Related

- Depends on #6073

---
## Summary by cubic
Removes three unreachable tests that checked preserving `requested_subagent_type` when both `category` and `subagent_type` were provided. After #6073, `task()` rejects that combination, so the code path no longer exists and the tests were removed.

<sup>Written for commit cc75bc64f8e39a18f9cd0b51b3ef7daa083c7ea7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6120?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for the PR, @Ruifeng-Zhang.

Holding this one together with #6073, since the tests removed here only become unreachable once #6073 turns the `category` plus `subagent_type` combination into a thrown error. Merging this alone would drop coverage while the lenient override is still in place.

I have asked the owner to decide the contract question on #6073, because #3624 and #2847 established the current silent-override behavior intentionally. Once that call is made, this PR lands with #6073 or is withdrawn with it. Nothing needed from you in the meantime, and thanks for splitting the test cleanup out clearly.

code-yeongyu:

PR #6130, opened 2026-07-15
feat(opencode): align codegraph exclusion with codex

vvvzhangshuo:
## Summary

Codex already applies CodeGraph project exclusion policy via `codegraph.excluded_roots` in bootstrap and MCP serve. OpenCode had no equivalent wiring, so workspaces under configured roots could still bootstrap and register the CodeGraph MCP.

This PR aligns the OpenCode path with Codex:

- **`excluded_roots`**: reuse `shouldExcludeCodegraphProject()` in OpenCode bootstrap and MCP registration to skip excluded workspaces.
- **`exclude`**: merge configured project-level exclude patterns into `codegraph.json` before `init`/`sync`, because the CodeGraph CLI reads that file for in-project indexing rules.
- **Hand-written `codegraph.json`**: still supported as-is; OMO only merges `exclude` when `codegraph.exclude` is configured and preserves other existing fields.

## Test plan

- [x] `bun test` for `ensureCodegraphProjectConfig`
- [x] `bun test` for OpenCode codegraph schema, bootstrap hook, and MCP config
- [x] `bun run build:schema`

---
## Summary by cubic
Aligns OpenCode with Codex so CodeGraph respects project exclusion. Skips bootstrap/MCP registration for workspaces under `excluded_roots`, and writes configured `exclude` patterns to `codegraph.json` before init.

- **Bug Fixes**
  - Honor `excluded_roots` via `shouldExcludeCodegraphProject()` in bootstrap and MCP.
  - Merge `codegraph.exclude` into `codegraph.json` while preserving existing fields.
  - Update schema to include `exclude` and `excluded_roots`.
  - Add `ensureCodegraphProjectConfig()` in `@oh-my-opencode/utils` with tests.
  - Adjust bootstrap tests to run outside tmp roots now excluded by default.

<sup>Written for commit c361c1b53a09ddc758a64599f6b4fabd69bc5ff4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -6562,6 +6562,18 @@
           "default": true,
           "type": "boolean"
         },
+        "exclude": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "excluded_roots": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "install_dir": {
           "type": "string"
         },
--- a/packages/omo-opencode/src/config/schema/codegraph.test.ts
+++ b/packages/omo-opencode/src/config/schema/codegraph.test.ts
@@ -52,6 +52,8 @@ describe("OhMyOpenCodeConfigSchema codegraph", () => {
         codegraph: {
           auto_provision: false,
           enabled: true,
+          exclude: ["pixie/", "bcc/"],
+          excluded_roots: ["~/scratch/codegraph"],
           install_dir: "~/.omo/codegraph",
           telemetry: false,
           watch_debounce_ms: 250,
--- a/packages/omo-opencode/src/config/schema/codegraph.ts
+++ b/packages/omo-opencode/src/config/schema/codegraph.ts
@@ -4,6 +4,8 @@ export const CodegraphConfigSchema = z.object({
   auto_init: z.boolean().default(true),
   auto_provision: z.boolean().default(true),
   enabled: z.boolean().default(true),
+  exclude: z.array(z.string()).optional(),
+  excluded_roots: z.array(z.string()).optional(),
   install_dir: z.string().optional(),
   telemetry: z.boolean().optional(),
   watch_debounce_ms: z.number().nonnegative().optional(),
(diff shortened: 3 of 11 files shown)


code-yeongyu:
CI failing on all 3 platforms — `codegraph-bootstrap/auto-init.test.ts` has 2 test failures:

1. `#given auto_init is false and .codegraph already exists #when bootstrap runs #then it continues with sync`
2. `#given auto_init is true and .codegraph does not exist #when bootstrap runs #then it proceeds with bootstrap`

Both fail because `events.some(e => e.startsWith("prepare:"))` returns false — the exclusion logic appears to be blocking the bootstrap path before the prepare event fires, even for non-excluded projects. The auto-init tests expect that non-excluded workspaces still reach the prepare/sync phase.

vvvzhangshuo:
Thanks for catching that — you're right.

The auto-init bootstrap tests were using `/tmp` workspaces, so the new default tmp-root exclusion policy short-circuited bootstrap before `prepare:`/`sync` could run.

Fixed in c361c1b by moving those tests to non-tmp workspaces (same approach as the other bootstrap tests). CI is green on ubuntu, macOS, and Windows now.

code-yeongyu:
Thanks @vvvzhangshuo. This is exactly the right parity gap and we implemented its intent in #6251 (merged to `dev`): the OpenCode codegraph bootstrap hook and MCP registration now both gate on the shared `shouldExcludeCodegraphProject()` with codex-equal defaults (os.tmpdir, /tmp, /private/tmp, and any `.omo` path segment), and `codegraph.excluded_roots` is now an OpenCode schema key with harness support extended to opencode.

We deliberately did not adopt the `codegraph.json` exclude-merge sub-feature (kept the parity minimal to the excluded_roots gate). Closing as absorbed, with credit for surfacing it.

PR #6136, opened 2026-07-15
feat: default the browser skill to the openchrome-aside tiered stack (aside + openchrome, Playwright MCP fallback)

MoerAI:
## What

Adds a new `openchrome-aside` value for `browser_automation_engine.provider` and makes it the **default**. It backs the built-in browser skill (canonical name `playwright`, so existing triggers keep working) with a tiered stack:

1. **aside** — LLM ask mode (`aside "<task>"`) as the primary path
2. **openchrome** — deterministic real Chrome via the `oc` CLI
3. **Playwright MCP** — always-available `npx` fallback

## Why

Requested: make our e2e/browser tooling default to openchrome + aside, leveraging aside's connected LLM ask mode. The tiered design keeps zero-setup behaviour — users without aside/openchrome fall back to Playwright MCP automatically, so nothing breaks.

## Zero-breakage gating fix (the important bit)

Flipping the default surfaced a latent bug: `filterProviderGatedSkills` compared `skill.name === browserProvider`, so once the default was no longer literally `playwright`, a user's own `playwright` SKILL.md would be **silently dropped**. Fixed in both gating copies (`skill-context.ts` and `skill-discovery.ts`) with a provider to rendered-skill-name map, guarded by a regression test that fails on the old gating and passes on the fix.

Six default sources now consistently resolve `openchrome-aside`: the Zod `.default`, `skill-context.ts`, the `createBuiltinSkills` internal default, `skill-discovery.ts`, and the `agent-config-handler` + `command-config-handler` config-pipeline handlers.

## Commits

1. `feat(skills):` core — provider, tiered skill, selection branch, gating fix, defaults
2. `feat(config):` handler default consistency (agent + command config)
3. `docs(reference):` docs + regenerated schema

## QA / evidence

- TDD RED to GREEN on every behaviour change (enum, skill, selection, six default flips, both gating fixes, both handlers).
- `bun run build` **succeeds** on this branch; blast-radius suite **505 pass / 0 fail**; shim-inventory audit passes; LSP clean on changed files.
- Real-surface QA, two layers, both in isolated XDG sandboxes:
  1. Drove the unmocked `createSkillContext` / `createBuiltinSkills` / schema — the default resolves to the tiered `playwright` skill with the aside / openchrome / Playwright MCP tiers, and a user's own `playwright` SKILL.md survives.
  2. Booted **real opencode 1.18.1** with the freshly-built plugin — the plugin loaded and `browser_automation_engine.provider: "openchrome-aside"` was accepted into the merged config with no validation error; plugin init completed.
  - Isolation proof: real `opencode.db` session count unchanged across both.
- Independent reviewer pass: **NO BLOCKERS**.

## Notes

- `mcpConfig` declares only Playwright MCP (always available via npx). aside and openchrome are driven via their CLIs (`aside`, `oc`), which the skill detects with `command -v` before use.
- Evidence recorded locally under `.omo/evidence/20260715-openchrome-aside/` (gitignored).

Holding merge per request.

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -6739,13 +6739,14 @@
       "type": "object",
       "properties": {
         "provider": {
-          "default": "playwright",
+          "default": "openchrome-aside",
           "type": "string",
           "enum": [
             "playwright",
             "agent-browser",
             "dev-browser",
-            "playwright-cli"
+            "playwright-cli",
+            "openchrome-aside"
           ]
         }
       },
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -557,15 +557,20 @@ Available commands: `init-deep`, `ralph-loop`, `ulw-loop`, `cancel-ralph`, `refa
 
 ### Browser Automation
 
-| Provider               | Interface | Installation                                        |
-| ---------------------- | --------- | --------------------------------------------------- |
-| `playwright` (default) | MCP tools | Auto-installed via npx                              |
-| `agent-browser`        | Bash CLI  | `bun add -g agent-browser && agent-browser install` |
+The `browser_automation_engine.provider` field selects the browser skill implementation. The default `openchrome-aside` is a tiered stack: aside LLM ask mode first, openchrome for deterministic real Chrome, and Playwright MCP as a zero-setup fallback when aside and openchrome are not installed. Every provider is exposed through the `playwright` skill trigger.
+
+| Provider                     | Interface                              | Installation                                                    |
+| ---------------------------- | -------------------------------------- | --------------------------------------------------------------- |
+| `openchrome-aside` (default) | aside/oc CLI + Playwright MCP fallback | aside CLI + `npm install -g openchrome-mcp`; fallback needs none |
+| `playwright`                 | MCP tools                              | Auto-installed via npx                                          |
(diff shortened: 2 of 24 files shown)


MoerAI:
CI + review status note:

- The `test (ubuntu-latest)` failure (7 tests in `comment-checker` routing and `fixEmptyMessagesWithSDK`) did not reproduce locally (isolated and full `packages/omo-opencode` runs green on the same head; those files are untouched by this diff) and passed on an as-is rerun. All three OS test jobs are now green.
- The Codex review findings (monitor tool factory export P1, `/btw` tool guard P2) reference `packages/omo-opencode/src/tools/monitor/create-monitor-tools.ts` and `plugin/hooks/create-tool-guard-hooks.ts` — neither file is in this PR's diff, and the reviewed commit `5bb29fd2ce` is not on the branch (pre-force-push head). Those findings apply to a superseded version of the branch, not the current three commits.

MoerAI:
Closing this. It's a eat: that flips the default browser-skill stack to openchrome-aside — a user-visible default / behavior-contract change, which is the shape this repo's maintainer sweep consistently parks unless it begins as a maintainer-requested change or an issue-first discussion. It has also gone CONFLICTING against dev after 9 days with only automated review, so I'd rather not hold a stale slot. Happy to reopen or split out any non-default-changing, opt-in piece (e.g. shipping the openchrome-aside skill itself as opt-in) if that direction is wanted. Thanks!

PR #6137, opened 2026-07-15
fix: restore the v4.18.1 validation baseline

WangErgouaaaa:
## Summary

Restore two independent validation guarantees that regressed on `dev` after the v4.18.1 release:

- regenerate the committed Codex installer so its embedded package metadata is `4.18.1` instead of `4.18.0`
- keep the session-event retry test at its own boundary by asserting that stop hooks receive a JSONL transcript path, without dynamically re-importing a module another test intentionally mocks in Bun's one-process suite

## Why

The release version test failed because the generated installer bundle was not refreshed by the v4.18.1 release PR. Separately, the session-event test compared a path captured from the mocked transcript module with a later import of the restored real module. Both values are valid in their respective module states, so the cross-module equality assertion was testing Bun mock lifecycle rather than session-event behavior.

## QA & Evidence

- **Full repository gate:** `env -u OMO_DISABLE_POSTHOG -u OMO_SEND_ANONYMOUS_TELEMETRY bun test --only-failures` -> **11,492 passed, 3 skipped, 0 failed**
- **Type safety:** `bun run typecheck` -> passed
- **Build:** `bun run build` -> passed and reproduced only the expected installer version diff
- **Codex isolated install:** `.agents/skills/codex-qa/scripts/install-verify.sh --self-test` -> plugin cache `4.18.1`, config enabled, 9 component bins and agent TOMLs linked, real `~/.codex/config.toml` unchanged
- **OpenCode isolated surface:** server smoke passed (`/global/health`, 155 OpenAPI paths, auth rejection); SSE probe observed `server.connected`
- **Targeted regressions:** installer version test and the session-event retry file passed; the retry file also passed 50 repeated runs
- **Evidence directory:** `.omo/evidence/20260716-dev-validation-baseline/` (local, intentionally uncommitted)

The OpenCode common self-check also confirmed sandbox isolation but reported that this host lacks the optional `sqlite3` executable; the server and SSE checks used here do not require it.

## Scope and risk

No shipped OpenCode production code changes. The Codex change is a generated one-line version synchronization. The OpenCode change narrows one test assertion to the handler contract while leaving exact path construction owned by the transcript module.

PR #6138, opened 2026-07-15
fix(glob): keep the no-rg fallback on find

WangErgouaaaa:
## Summary

Fix the Unix glob fallback used when ripgrep is unavailable. The fallback already builds GNU `find` arguments, but it accidentally executed the discovered `grep` binary with those arguments, producing errors such as `grep: invalid max count` instead of returning files.

This branch is now based directly on `dev` and contains only the single glob fix commit.

## Changes

- Execute `find` directly for the non-Windows no-ripgrep fallback.
- Add a regression test proving the grep discovery sentinel is not invoked with `find`-specific flags such as `-maxdepth`.

## QA & Evidence

- **Focused regression:** `bun test packages/omo-opencode/src/tools/glob/cli.test.ts`
  - Observed after unstacking: **22 passed, 0 failed**.
- **Type safety:** `bun run typecheck`
  - Observed after unstacking: passed.
- **Build:** `bun run build`
  - Observed after unstacking: passed. The build regenerated the already-known stale Codex installer version artifact from `dev`; that unrelated generated diff was intentionally not included.
- **Full repository gate:** `env -u OMO_DISABLE_POSTHOG -u OMO_SEND_ANONYMOUS_TELEMETRY bun test --only-failures`
  - Observed after unstacking: **11,491 passed, 3 skipped, 2 failed**.
  - Both failures are outside this PR's two-file diff and reproduce the current `dev` baseline gaps previously isolated in #6137: the stale generated Codex installer version and the Bun transcript-module mock interaction.
- **Real OpenCode QA:** isolated `opencode run --format json` with the local plugin and a fake provider, with ripgrep unavailable.
  - Observed: the glob tool returned the target file; `find` ran with `-maxdepth`; grep did not receive find flags; the localized grep error was absent.
  - Local sanitized artifacts: `.omo/evidence/20260716-glob-find-fallback/`.

## Scope and risk

- Final PR scope: one commit, two glob files.
- The Windows PowerShell fallback is unchanged.
- This remains a draft while the unrelated `dev` baseline failures prevent a fully green repository gate.

### DIFF
--- a/packages/omo-opencode/src/tools/glob/cli.test.ts
+++ b/packages/omo-opencode/src/tools/glob/cli.test.ts
@@ -225,4 +225,29 @@ describe("runRgFiles", () => {
     })
     expect(spawnMock).toHaveBeenCalled()
   })
+
+  if (process.platform !== "win32") {
+    it(
+      "#given GNU grep is the discovery fallback #when glob runs #then it executes find instead of passing find flags to grep",
+      async () => {
+        const commands: string[][] = []
+        const spawnMock = mock((command: string[], _options?: SpawnOptions): SpawnedProcess => {
+          commands.push(command)
+          return createSpawnedProcess(0)
+        })
+
+        const result = await runRgFiles(
+          { pattern: "*.ts", paths: ["."], timeout: 1000 },
+          { path: "/usr/bin/grep", backend: "grep" },
+          spawnMock
+        )
+
+        expect(result.error).toBeUndefined()
+        expect(commands).toHaveLength(1)
+        expect(commands[0]?.[0]).toBe("find")
+        expect(commands[0]).toContain("-maxdepth")
+        expect(commands[0]?.[0]).not.toBe("/usr/bin/grep")
+      }
+    )
+  }
 })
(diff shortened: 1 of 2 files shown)


WangErgouaaaa:
@code-yeongyu Maintainer handoff for this draft:

- the branch has been unstacked from the withdrawn #6137 and now contains one commit touching only the two glob files
- focused glob tests, typecheck, and the full build pass locally
- the new CI run is `action_required` with zero jobs, so it needs maintainer approval before GitHub will execute the fork workflow
- the PR remains Draft because current `dev` still has the two unrelated baseline test failures documented in the description

Please approve the workflow when convenient; no review or merge is requested until that baseline situation is resolved.

code-yeongyu:
[sisyphus-bot] The fix is correct by inspection — `buildFindArgs` produces `find` flags, so the fallback command must be `["find", ...args]`, not `[cli.path, ...args]` where `cli.path` is the grep binary. Consider marking as ready for review; the change is 1 line + 1 test and the bug is still present on dev.

PR #6140, opened 2026-07-16
fix(lsp-core): resolve clean diagnostics for silent no-pull servers instead of freshness timeout

### LINKED ISSUE #6131: [Bug]: shared lsp_diagnostics freshness timeout after v4.18.1 upgrade
### Prerequisites
- [x] I searched existing issues and related PRs, including #1522, #1731, and #3114.
- [x] I am using oh-my-openagent 4.18.1.
- [x] I read the documentation and investigated the current LSP implementation.

### Bug Description
After upgrading to oh-my-openagent 4.18.1, `lsp_diagnostics` can return a shared 3-second freshness timeout for language servers that do not return pull diagnostics and do not publish an empty diagnostics notification for a clean file. This is not a Marksman-specific report: the same user-facing symptom has also been observed with Go/gopls. The issue appears to be in the shared diagnostics freshness path.

### Steps to Reproduce
1. Use oh-my-openagent 4.18.1 on macOS arm64.
2. Configure a Markdown LSP server in the user LSP config:

```json
{
  "lsp": {
    "markdown": {
      "command": ["marksman", "server"],
      "extensions": [".md", ".markdown"]
    }
  }
}
```

3. Run `lsp_diagnostics` against a clean `README.md`.
4. Observe that the request waits for the freshness deadline and returns a timeout instead of an empty diagnostics result.

### Expected Behavior
For a healthy server and a clean file, `lsp_diagnostics` should return zero diagnostics without reporting a timeout. Files with real diagnostics should return those diagnostics.

### Actual Behavior
`lsp_diagnostics` returns:

```text
Timed out waiting for fresh diagnostics for README.md within 3000ms.
```

The LSP server starts successfully and other operations such as `lsp_symbols` work. The failure is specific to diagnostics readiness/freshness.

### Doctor Output

```shell
✓ System OK (opencode 1.18.1 · oh-my-openagent 4.18.1)
```

### Error Logs

```text
Marksman initialize response: capabilities did not include diagnosticProvider
textDocument/diagnostic response: -32601 Method not found
No textDocument/publishDiagnostics notification arrived during an 8-second direct protocol probe for a clean README.md.
```

### Configuration

```json
{
  "lsp": {
    "markdown": {
      "command": ["marksman", "server"],
      "extensions": [".md", ".markdown"]
    }
  }
}
```

### Additional Context

- The regression boundary appears to be commit `cf65a63ab`, included in merged PR #6097 (`Migrate Senpi LSP to the shared OMO daemon).
- The current shared client waits for fresh push diagnostics when pull diagnostics are unsupported. If a server remains silent for a clean file, the wait ends as `freshness_timeout`.
- `lsp_symbols` succeeds for both Markdown and Go in the same environment, which suggests the daemon/proxy/initialize path is alive and the common failure is in diagnostics handling rather than server startup.
- A Go timeout has been observed in a real user project, but the repository fixture used for this report only contains template Go files without a valid module, so the Go symptom is reported as the same observed class rather than claimed as this minimal reproduction.
- Related prior reports: [#1522](https://github.com/code-yeongyu/oh-my-openagent/issues/1522), [#1731](https://github.com/code-yeongyu/oh-my-openagent/pull/1731), and still-open [#3114](https://github.com/code-yeongyu/oh-my-openagent/pull/3114). This issue is about the regression after the shared-core/package migration and current layout.

### Operating System

macOS (Apple Silicon)

### OpenCode Version

1.18.1

MoerAI:
Fixes #6131

## Summary

After the shared LSP migration, `lsp_diagnostics` on a clean file returns `Timed out waiting for fresh diagnostics ... within 3000ms` for language servers that neither support pull diagnostics nor publish a `textDocument/publishDiagnostics` notification for a clean file (marksman observed in the issue's protocol probe; the same class covers other silent-on-clean servers). This restores the pre-migration contract for exactly that case: the request resolves with the stored (empty) diagnostics instead of reporting a freshness timeout.

## Root Cause

`packages/lsp-core/src/lsp/client.ts` `diagnostics()` waits for fresh push diagnostics when pull is unsupported. For a server with no `diagnosticProvider` capability (or a `-32601` pull rejection) that stays silent on a clean file, no satisfying event can ever arrive, so the wait always ends as `freshness_timeout`. Before the hardening in `cf65a63ab` (`git show cf65a63ab^:packages/lsp-core/src/lsp/client.ts`, lines 125-145), this exact case fell back to `return { items: this.getStoredDiagnostics(uri) }` — empty, non-error.

## Changes

| File | Change |
|------|--------|
| `packages/lsp-core/src/lsp/client.ts` | At the push-fallback deadline: when pull is unsupported and `snapshot.publishGeneration === 0` (the server never published for this document), return stored (empty) diagnostics instead of `freshness_timeout`. 10 lines incl. comment. |
| `packages/lsp-core/src/lsp/client-diagnostics-freshness.integration.test.ts` | +1 integration test through the real JSON-RPC subprocess harness: silent no-pull server on a clean file resolves clean after the freshness window. |

The shared `LspClient` owns the diagnostics freshness semantic; this PR changes only its terminal deadline branch and leaves the pull path, stale-publish fencing, and quiescence handling intact.

## QA & Evidence

Committed under `.omo/evidence/20260716-fix-6131/` (all bash-redirected UTF-8):

RED on unmodified `dev` (`red.txt`):
```text
(fail) LspClient diagnostics freshness > #given a server without pull support that never publishes diagnostics ... #then the request resolves clean after the freshness window instead of reporting a timeout
 1 fail
exit=1
```

GREEN on head (`green.txt`): `1 pass / 0 fail / exit=0`. Negative control (`negative-control.txt`): product fix stashed → test fails again → unstashed.

Live real-surface proof — production `omo-lsp mcp` stdio entry (`packages/lsp-tools-mcp/src/cli.ts mcp`) driven with **real marksman 2026-02-08**, isolated temp workspace and isolated `LSP_TOOLS_MCP_USER_CONFIG` (`live-mcp-driver.mts`):

Before (`live-before-fix.txt`):
```text
"text": "Timed out waiting for fresh diagnostics for ...\\README.md within 3000ms."
"isError": true, "errorKind": "freshness_timeout"
```

After (`live-after-fix.txt`):
```text
"text": "No diagnostics found"
"isError": false
```

- `bun run typecheck` — exit 0 (`typecheck.txt`)
- Full `bun test packages/lsp-core`: dev baseline 89 tests / 0 fail; head 90 tests / 0 fail on rerun (`related-suites-*.txt`; the first head run had 2 unrelated workspace-mutation-lease failures that did not recur — Windows host subprocess-timing flakes, documented in `qa-summary.md`)

## Risks & Residuals

- Scoped to the terminal deadline branch of the push-fallback path. Does not change: stale-publish fencing (a server that HAS published still times out via `publishGeneration > 0` — existing integration tests at the versionless/stale cases), hung advertised-pull behavior (pull stays "supported", still times out — existing cancel test), pull-path semantics, or any timing constants.
- The full freshness window is still honored before resolving clean, so a slow publisher inside the window still wins (asserted in the new test via elapsed-time check).
- Residual: a silent-on-clean server still costs the full 3000ms window per call (unchanged timing; only error→empty changes). Servers publishing real diagnostics slower than the window lose them with and without this change — the window semantics are untouched.

## Automated Checks

- `bun test packages/lsp-core` green on head (see above); CI runs the full matrix.
- `bun run typecheck` green.

AI-assisted (Claude Code); I have reviewed and understand every line.

---
## Summary by cubic
Fixes a regression where `lsp_diagnostics` timed out on clean files for silent, no-pull servers (e.g., Marksman). Now resolves after the freshness window with empty diagnostics, and guards the fallback to only return a version-current pull cache.

- **Bug Fixes**
  - `packages/lsp-core/src/lsp/client.ts`: At the push-fallback deadline, if pull is unsupported and `publishGeneration === 0`, resolve with the version-current pull cache via `documents.getPullCache(snapshot)` or empty; otherwise keep the timeout. Prevents returning stale cached diagnostics and avoids macOS realpath mismatches.
  - Tests: Extended `client-diagnostics-freshness.integration.test.ts` to cover the silent no-pull clean case and cache guards (stale cache ignored after change; current cache reused when unchanged). Updated `diagnostics-freshness-contract-probe.ts` to expect a clean result for the silent case.

<sup>Written for commit 438142a5721a66f43350ef6cd9db89c414f01ae2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:

MoerAI:
Factual triage of the `test (ubuntu-latest)` failures (two attempts, same 7 tests: `comment-checker mutation tool routing` x3 and `fixEmptyMessagesWithSDK` x4 - files this PR does not touch):

Linux baseline comparison, WSL2 Ubuntu 22.04, bun 1.3.14, full `bun test packages/omo-opencode`:

- This PR head (`13e072c`): 8158 pass / 24 fail, including all 4 `fixEmptyMessagesWithSDK` tests failing in full-suite context.
- Unmodified `dev` (`dec381e`), same environment, same command: 8158 pass / 24 fail, including the SAME 4 `fixEmptyMessagesWithSDK` tests failing identically.
- Both victim files pass in isolation on both trees (11 pass / 0 fail across the 2 files).

The failure set is byte-identical between this head and unmodified dev, so the full-suite failures are a pre-existing test-ordering interaction on Linux (the victims' `mock.module` registrations get bypassed depending on suite context), not something this diff introduces. This is the same recurring pair tracked in #6142, which also hit #6136 yesterday (passing on an as-is rerun there).

Dispatching another rerun of the failed job. The lsp-core changes in this PR have their own green verification: full `bun test packages/lsp-core` on this head (90 tests, 0 fail) plus the committed RED/GREEN and real-marksman evidence under `.omo/evidence/20260716-fix-6131/`.

code-yeongyu:
Maintainer follow-up landing on this branch (2 commits + dev merge):

**What changed since the last review** — addresses the outstanding P2 codex-review comment (stale pull-cache in the new fallback):

- The terminal fallback no longer returns `getStoredDiagnostics(uri)`. Verified two defects in that call: (a) it has no version check, and `changeDocument` never clears `pullCache`, so after a document change it could return the previous version's pull diagnostics as if current (Linux/Windows); (b) on macOS it silently returned `[]` even when a current cache existed, because `diagnostics()` builds the uri from `resolve()` while the document store keys by `realpathSync` (`/var` vs `/private/var`) — which is also why the stale-cache half was invisible on macOS.
- It now uses the snapshot-keyed, version-filtered `documents.getPullCache(snapshot)`: an actually-current cache is returned, anything else resolves explicit empty. `publishGeneration === 0` still guarantees push state can't leak in; stale/future-publish and hung-pull still end in `freshness_timeout` (contract probe: all 9 outcomes ok).
- The manual freshness contract probe's `silent` outcome is flipped to the restored pre-migration contract (clean empty), matching the PR's intent.

**New regression coverage** (both through the real JSON-RPC subprocess harness):
- pull cached for v1 → file changes → pull rejected `-32601` → fallback resolves **empty**, never the stale v1 diagnostics.
- pull cached for v1 → pull rejected `-32601` with no intervening change → fallback resolves with the **current** cached diagnostics.

**Fresh QA evidence** under `.omo/evidence/20260719-fix-6131-stale-guard/`: RED→GREEN on the unit seam, plus a live drive of the production `omo-lsp mcp` stdio server against real marksman on macOS — before: `Timed out waiting for fresh diagnostics ... within 3000ms` (`errorKind: freshness_timeout`); after: `No diagnostics found`, `isError: false`, with the full 3000ms freshness window still honored. Gates: `bun test packages/lsp-core` 92/92, `bun run typecheck` 0, `bun run build` 0, `bun run test:codex` 510/510 (this also covers the lazycodex/Codex-edition path that shares this code).

@MoerAI thanks for the clean root-cause work and the well-scoped fix — I only hardened the fallback's cache semantics on top of it.

Issue #5162, opened 2026-06-11
Sub-agent tmux panes fail with 401 Unauthorized when OPENCODE_SERVER_PASSWORD is set


Eric-GoodBoy-Tech:
## Problem

When the OpenCode server requires Basic Auth (`OPENCODE_SERVER_PASSWORD` is set), sub-agent tmux panes fail to connect with `401 Unauthorized`.

## Root cause

`pane-activate.ts`, `pane-replace.ts`, `pane-spawn.ts`, `window-spawn.ts`, and `session-spawn.ts` do not pass `OPENCODE_SERVER_PASSWORD` / `OPENCODE_SERVER_USERNAME` as environment variables via tmux `-e` flags.

In contrast, team layout (`team-layout-tmux/layout.ts`) already does this correctly with `buildPaneEnvironmentArgs()`.

## Fix

PR #5161 adds `buildPaneAuthEnvironmentArgs()` to `pane-command.ts` and applies it to all 5 tmux pane lifecycle functions.

## Impact

- `OPENCODE_SERVER_PASSWORD` set → auth env vars passed to pane → `opencode attach` succeeds
- `OPENCODE_SERVER_PASSWORD` unset → `buildPaneAuthEnvironmentArgs()` returns `[]` → zero impact

code-yeongyu:
[sisyphus-bot] Thanks for the report! Automated triage confirmed this with high confidence: tmux pane create/respawn commands pass no `-e` environment flags, so the pane-side `opencode attach` never sees OPENCODE_SERVER_PASSWORD and gets 401 against a secured server. There's already a correct-direction contributor PR (#5161 by @Eric-GoodBoy-Tech) that went stale with conflicts — a supervised session has been dispatched to bring it up to date on current dev (composing it with the recent argument-quoting fix), add regression coverage across all five spawn/respawn call sites, verify against a secured server in isolation, and land it with the contributor's authorship preserved.

code-yeongyu:
[sisyphus-bot] Fixed and merged via https://github.com/code-yeongyu/oh-my-openagent/pull/5161 (contributor PR by @Eric-GoodBoy-Tech, brought up to date and verified). Sub-agent panes now receive OPENCODE_SERVER_PASSWORD/USERNAME through tmux `-e` flags at every create/respawn call site, proven against a secured server in isolation (401 before, clean attach after). Thanks for the report!

PR #6158, opened 2026-07-16
fix(tmux): scope health authentication to trusted listeners

### LINKED ISSUE #6223: [Bug]: tmux readiness probes mishandle authenticated OpenCode listeners
### Prerequisites

- [x] I searched existing issues; related reports cover only parts of this failure.
- [x] This behavior was confirmed against the current `dev` implementation.
- [x] The relevant tmux and Team Mode documentation was reviewed.

### Bug Description

Tmux visualization readiness can be wrong in both directions:

1. The global in-process marker lets `isServerRunning()` return `true` without probing the listener, so panes may spawn even when `opencode attach` has no usable HTTP listener.
2. Once that shortcut is removed, the real `GET /global/health` probe is unauthenticated. A valid listener protected by `OPENCODE_SERVER_PASSWORD` returns `401`, so pane, window, session, or Team Mode visualization is skipped incorrectly.

Readiness credentials must only be sent to the trusted current OpenCode listener. Historical listener URLs recovered from tmux metadata must remain unauthenticated.

### Steps to Reproduce

1. Run OpenCode inside tmux with tmux visualization enabled.
2. Without a real HTTP listener, trigger a background agent or Team Mode layout: the in-process marker can bypass the health probe and allow a broken attach pane.
3. Start a real listener with `OPENCODE_SERVER_PASSWORD` and trigger visualization again after removing the marker: the unauthenticated health probe receives `401` and skips the valid listener.

### Expected Behavior

- Readiness is established by a successful `/global/health` response from the exact listener URL.
- Trusted current-listener probes reuse `OPENCODE_SERVER_PASSWORD` and optional `OPENCODE_SERVER_USERNAME`.
- Redirects and unsuccessful responses do not establish readiness.
- Failed readiness skips only tmux visualization; agents and teams continue normally.
- Historical or arbitrary URLs never receive OpenCode credentials.

### Actual Behavior

The in-process marker can create a false positive, while an unauthenticated health probe creates a false negative for password-protected listeners.

### Scope

This does not require listener discovery, dynamic-port resolution, automatic listener startup, or OpenCode core changes.

### Verification

The regression is covered by focused tmux-core, team-core, and OMO adapter tests plus isolated private-socket tmux QA in PR #6158.

### Related

- Historical false-readiness symptom: #3894
- Fallback URL and user-warning follow-up: #5158
- Previously fixed attach-environment propagation: #4409 and #5162
- Implementation: #6158

jerome-benoit:
## Summary

- Remove the in-process server marker so tmux visualization always proves that the external OpenCode listener is reachable.
- Authenticate health checks only for the trusted current listener; probe synthetic, raw, and historical addresses anonymously and explicitly clear inherited OpenCode credentials on every untrusted pane lifecycle path.
- Partition implicit health authority by normalized listener, and coalesce requests only when listener, opaque request-policy identity, and transport identity match.
- Reject superseded in-flight results before they can authorize a tmux mutation or retry.
- Keep tmux/cmux mutations backend-stable: resolve the backend against the captured environment, fail closed on drift, and use the same sanitized per-pane capability for rollback and cleanup.
- Recover cleanup ownership from persisted, focus, and grid pane IDs, then deduplicate targets before any ownership probe or mutation.
- Preserve historical panes after any direct HTTP response, including `401` and redirects; close them only after two anonymous transport failures.

## Behavior

The active-listener readiness check calls `/global/health`, rejects redirects, and requires a successful response. Current credentials are resolved lazily and are never stored in cache state or logs. New or replaced panes for a verified current listener receive the current credentials; panes for unverified addresses receive explicit empty credential variables. Existing panes and tmux's global environment are not rewritten.

Implicit callers for different normalized listeners have independent authority. Concurrent checks share one request only when normalized listener, request-policy identity, and fetch transport are identical. A superseded request cannot retry, publish success, or return `true`; reset invalidates every pending implicit probe. Explicit `ServerHealthState` retains its cross-listener supersession semantics.

The tmux runner treats an explicit environment as authoritative for both backend resolution and child execution. Each pane creation or respawn revalidates readiness and uses the exact authentication snapshot that passed its probe. Team layout obtains one snapshot for each member pane; that pane's creation and setup reuse it. Layout selection and resizing do not start a process and do not perform a separate health probe.

Rollback and shutdown cleanup derive one stable target from persisted pane IDs plus non-leader focus/grid pane IDs. Duplicate IDs are removed before validation, ownership probes, mutation, and result reporting. Cleanup removes only panes whose persisted ownership marker exactly matches the Team run. Operational ownership failures remain distinct from unowned panes; partial and skipped cleanup remain explicit.

Historical cleanup accepts only loopback HTTP(S) metadata without userinfo or port zero. Its probe uses `credentials: "omit"` and `redirect: "manual"`, so it neither forwards authentication nor contacts redirect targets.

## Verification

- Focused regressions: 87 passed, 0 failed, 249 assertions.
- Root typecheck: pass.
- Complete build: pass.
- Final affected tmux-core, team-core, OpenCode tmux adapter, Team layout, and Team runtime suites: 627 passed, 2 intentional live-tmux skips, 0 failed, 1,559 assertions.
- Real tmux layout smokes: 2 passed, 0 failed, 16 assertions.
- Isolated OpenCode TUI smoke: render, input sentinel, and teardown passed; the real database stayed at 5,070 sessions.
- Isolated OpenCode HTTP smoke: health returned `healthy=true`, `/doc` exposed 162 paths, unauthenticated `/session` returned `401`, and the bounded helper passed.
- Independent design, correctness, and security/QA reviews: 3/3 clear after final cross-validation.
- Five atomic fix commits and the upstream synchronization merges are GPG-signed and verified locally.

QA receipts are stored only in the ignored local `.omo/evidence/` tree, as required by repository policy; no `.omo` file is introduced by this PR. The branch contains no raw credential, Authorization value, provider secret, private key, or machine-private runtime artifact.

## Risk and scope

The change is intentionally fail-closed: listener or tmux validation failure disables only visualization; agents and teams continue. Detectable environment or backend drift aborts before mutation, and partial Team layout creation rolls back with the original sanitized per-pane capability.

An external listener or tmux process can still disappear after the final validation; that inherent runtime race cannot be eliminated without transactional support from those processes.

## Related issues

- Closes #6223.
- Follow-up context: #3894 and #3963.
- Related history: #4409, #5158, #5159, and #5162. Credit to @Eric-GoodBoy-Tech for the earlier work in #5159.


jerome-benoit:

code-yeongyu:
[sisyphus-bot] The two core fixes (remove in-process marker, scope auth to trusted listener) correctly address issue #6223. However, this PR is much larger than the issue scope: +5867/-1727 across 78 files includes significant cleanup-ownership recovery and cmux fail-closed changes that need careful review. Concerns: (1) The 3 removed tests in `create-managers.test.ts` for the marker behavior — are there equivalent tests for the new probe path? (2) The cmux fail-closed behavior changes pane creation semantics for users running `cmux omo` with authenticated listeners — this is a behavior change that should be documented in a changelog. (3) No CI test matrix has run yet (only metadata checks). Requesting test runs before merge.

PR #6159, opened 2026-07-16
fix(omo-opencode): retry task-store lock with bounded backoff under contention

### LINKED ISSUE #6157: [Bug] task_create/task_update fail under concurrent calls with task_lock_unavailable
## Summary

Concurrent task mutations can fail transiently with `task_lock_unavailable` instead of waiting for the shared task-store lock. This makes parallel task orchestration unreliable even when the underlying operations are independent.

This is related to, but distinct from, [#3599](https://github.com/code-yeongyu/oh-my-openagent/issues/3599): #3599 tracks cross-session state isolation, while this report tracks lock contention and the lack of retry/queue behavior.

## Version

- OMO: 4.18.2
- Current upstream branch checked: `dev`

## Reproduction

1. Run multiple OMO/OpenCode task operations concurrently against the same task list.
2. Issue several `task_create` calls at the same time, or mix `task_create` and `task_update`.
3. Observe that one or more calls return:

```json
{"error":"task_lock_unavailable"}
```

A live reproduction with four concurrent `task_create` calls produced three successful task creations and one `task_lock_unavailable` response. In the affected session, retrying after the contention cleared allowed the task workflow to continue.

## Observed behavior

- The task store uses one `.lock` file per task directory.
- `acquireLock()` uses exclusive file creation and returns immediately when a non-stale lock already exists.
- `task_create` and `task_update` return `task_lock_unavailable` immediately when acquisition fails.
- `task_create` holds the lock while awaiting `syncTaskTodoUpdate()`, increasing the contention window.

The current upstream implementation is visible in:

- `packages/omo-opencode/src/features/claude-tasks/storage.ts`
- `packages/omo-opencode/src/tools/task/task-create.ts`
- `packages/omo-opencode/src/tools/task/task-update.ts`

## Expected behavior

Concurrent task mutations should either:

- wait briefly with bounded retry/backoff until the lock is released, then perform the operation; or
- be serialized/queued by the task store; or
- return a clearly documented, actionable result that callers can safely retry without losing task state.

A short-lived normal contention event should not be surfaced as a permanent-looking tool error to the orchestration layer.

## Impact

Parallel agent execution can lose task creates or updates, leave the planner with incomplete task state, and force callers to add ad-hoc retries. This is especially disruptive because the tool call itself returns a structured error rather than indicating that the operation is merely waiting on another writer.

## Possible fix direction

- Add bounded retry with jitter/backoff around lock acquisition.
- Keep the critical section limited to filesystem mutation where possible; perform todo synchronization after releasing the file lock if consistency permits.
- Add concurrent `task_create`/`task_update` tests that assert transient contention is handled deterministically.
- Ensure stale-lock recovery cannot remove a lock acquired by another process between stale detection and removal.

## Related

- #3599 — parallel top-level CLI windows can cross-pollute through shared task state

cyllas:
## Summary

Closes #6157. Concurrent `task_create` / `task_update` calls could fail transiently with `{"error":"task_lock_unavailable"}` because the task-store lock was single-shot: on contention with a fresh lock, `acquireLock` returned immediately and both callers surfaced that as an error, with no wait. This adds bounded retry so normal short contention resolves instead of failing.

## Root cause

`acquireLock` in `packages/omo-opencode/src/features/claude-tasks/storage.ts` attempts an exclusive lock file create once (plus one stale reclaim at 30s). Under parallel task orchestration, every loser of the race got `acquired: false`, and `task-create.ts` / `task-update.ts` returned `task_lock_unavailable` right away. `task_create` also holds the lock across `await syncTaskTodoUpdate()`, widening the contention window.

## Fix

New async `acquireLockWithRetry(dirPath, options?)` wraps the unchanged `acquireLock` with bounded exponential backoff and full jitter (default: 10 attempts, 15ms base, 250ms cap). A genuinely stuck lock still returns `{ acquired: false }` once the budget is exhausted, so callers keep a safe, bounded escape hatch. `acquireLock` semantics (exclusive create, 30s stale reclaim) are unchanged, and `attempts: 1` is equivalent to a single `acquireLock`. `task-create` and `task-update` now `await` the retrying variant.

## Changes

- `features/claude-tasks/storage.ts`: add `acquireLockWithRetry` + `AcquireLockRetryOptions`.
- `tools/task/task-create.ts`, `tools/task/task-update.ts`: `await acquireLockWithRetry(taskDir)`.
- Tests in `storage.test.ts` and `task-create.test.ts`.

These were the only two production callers of `acquireLock`.

## QA and evidence

Deterministic before/after with 8 concurrent workers, each holding the lock across an `await` (like `syncTaskTodoUpdate`), identical across 3 runs:

```
OLD single-shot acquireLock: 1/8 ok, 7 task_lock_unavailable
NEW acquireLockWithRetry   : 8/8 ok, 0 task_lock_unavailable
```

Tests:

- `storage.test.ts` (+4): acquires immediately when free; waits then acquires when a held lock is released within budget; gives up `acquired:false` in bounded time when held past budget; `attempts:1` does not wait.
- `task-create.test.ts` (+1): 6 concurrent `task_create` calls all succeed with distinct persisted ids and zero `task_lock_unavailable` (ran 5x, stable).
- Suites `features/claude-tasks/` + `tools/task/`: 142 pass, 0 fail. `tsgo --noEmit` clean.

## Scope and residual risk

v1 keeps the critical section as-is (the lock is held across `syncTaskTodoUpdate`). Narrowing it would change correctness boundaries, so it is intentionally left as a possible follow-up. The retry budget is bounded, so there is no unbounded wait or deadlock; full jitter avoids lockstep livelock; the existing 30s stale reclaim still recovers a crashed holder's lock.

---
## Summary by cubic
Adds bounded retry with exponential backoff to the task-store lock so concurrent task operations don’t fail with transient `task_lock_unavailable`. Fixes #6157 by letting `task_create` and `task_update` wait briefly under contention and proceed when the lock frees.

- **Bug Fixes**
  - Introduced `acquireLockWithRetry` in `packages/omo-opencode/src/features/claude-tasks/storage.ts` with defaults: 10 attempts, 15ms base, 250ms cap, full jitter.
  - Switched `task-create` and `task-update` to await the retrying lock; stuck locks still return `acquired: false` once the budget is spent.
  - Added tests for immediate acquire, wait-then-acquire, bounded give-up, and concurrent creates without `task_lock_unavailable`.

<sup>Written for commit a6bcba295f95418ae817f6fe35469ed6b626e751. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6159?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:

MoerAI (changes requested):
Thanks for targeting the correct shared task-store lock boundary and limiting retries to the two mutating handlers.

Please update onto current `dev` and obtain a fresh passing CI run; the only substantive workflow was cancelled before executing steps:
https://github.com/code-yeongyu/oh-my-openagent/actions/runs/29509794399

Replace real timers, wall-clock thresholds, and production randomness with deterministic controlled contention. Exercise the actual `syncTaskTodoUpdate` lock window for `task_update` and mixed create/update operations. The current concurrent-create test has no OpenCode context, so that synchronization returns immediately and does not prove the real held-lock path.

Run the required isolated OpenCode QA and commit evidence under `.omo/evidence/`, including successful bounded retry and exhausted-retry behavior. This follows the lock-race testing bar established in #5886 and #6825:
https://github.com/code-yeongyu/oh-my-openagent/pull/5886
https://github.com/code-yeongyu/oh-my-openagent/pull/6825

code-yeongyu:

I'm closing it as **superseded**: the same defect (#6157 — `task_lock_unavailable` on ordinary contention) was fixed on `dev` by merged PR #8054 (*fix(omo-opencode): wait out task store lock contention*, 2026-09-10, commit `de7528b1c`), which makes `acquireLock()` wait out contention with jittered backoff inside a bounded budget — the same shape as the bounded retry proposed here — touching the same `storage.ts` / `task-create.ts` / `task-update.ts`, and issue #6157 is closed as completed.

**How I checked:** `gh pr view 8054` → `MERGED 2026-09-10T07:02:39Z`; `git merge-base --is-ancestor de7528b1c origin/dev` → yes; `gh issue view 6157` → `CLOSED COMPLETED 2026-09-10`.

Issue #5790, opened 2026-07-01
[Bug]: Background task result is lost unless the user manually follows up


RuszLi:
## Summary

When Sisyphus tells the user it is waiting for a background / parallel task result, the parent conversation can silently stall after that turn.

If the background task later completes, the result is not proactively delivered back into the user-facing conversation. In practice, the user must manually send a follow-up message like "Did the result come back?" to trigger the parent agent to continue.

This makes the task look hung even though the background work may already be finished.

## Why this is a bug

This is not just a UX paper cut. It looks like a continuation / background-task handoff reliability bug.

The expected async chain is:

1. parent agent launches background task(s)
2. background task finishes
3. parent session is resumed / woken up
4. parent agent reads the finished task output
5. parent agent posts the final conclusion back to the user

The observed behavior suggests step 1-2 may succeed, but step 3, 4, or 5 is unreliable or missing.

## Actual behavior

- Agent says it will wait for parallel/background investigation results.
- The turn ends.
- Background task finishes later, but no final answer is sent back to the user automatically.
- The user has to manually ask again to wake the conversation.
- Only after that follow-up does the agent have a chance to read task status/output and continue.

## Expected behavior

If the agent explicitly tells the user it is waiting for a background result, then once that result is available it should automatically continue and deliver the final answer without requiring a manual follow-up.

At minimum, the system should reliably:

- wake the parent session when the background task completes
- preserve the waiting state / task identifiers
- read the completed task output
- emit the follow-up response into the original conversation

## Reproduction pattern

I do not have a minimized code-level repro yet, but the user-facing pattern is stable:

1. Ask Sisyphus to investigate something that causes it to launch background / parallel work.
2. Sisyphus replies with something like: "I’ll wait here for the parallel investigation results to come back."
3. Do nothing else.
4. Wait for the background work to finish.
5. Observe that no proactive final response arrives.
6. Send a manual follow-up message.
7. Observe that only then does the agent continue.

## Likely failure points

Possible breakpoints:

- background task completion does not reliably wake the parent conversation
- parent session is resumed but does not automatically fetch and relay the task output
- task IDs / waiting state are not persisted correctly across turns
- the continuation logic treats "I am waiting for results" as a valid terminal response instead of a suspended response that must be resumed

## Impact

- users think the task is stuck or abandoned
- successful background work can be effectively lost
- trust in async / parallel execution drops significantly
- users must learn an accidental workaround: manually asking for progress

## Suggested direction

[...]

Issue #5569, opened 2026-06-25
[sisyphus-bot] [Bug]: Reply-required parent wake is persisted but no assistant turn starts after all background tasks complete


mochgolf:
### Prerequisites

- [x] I will write this issue in English (see the repository language policy)
- [x] I have searched existing issues to avoid duplicates
- [x] I am using current OpenCode / oh-my-openagent releases
- [x] I have checked the local OpenCode DB and log before filing

### Bug Description

When all OMO background tasks complete, the parent session receives the final internal wake message, but no assistant/model turn is started. The parent only resumes after I manually send another user message ("continue").

This looks related to the parent-wake / promptAsync gate family of issues, but it does not appear to be the exact old `forceNoReply` path from #5189: in this repro the final `[ALL BACKGROUND TASKS COMPLETE]` message has `<!-- OMO_INTERNAL_INITIATOR -->` and does **not** have `<!-- OMO_INTERNAL_NOREPLY -->`.

### Environment

- OS: Linux
- OpenCode: `1.17.9`
- oh-my-openagent: `4.13.0`
- Parent agent: `Sisyphus - ultraworker`
- Parent model: `zhipuai/glm-5.2`
- Child agents: `librarian` background subagents
- Observed in an interactive OpenCode/OMO workflow; no external middleware is required for the stall

### Session IDs

Parent:

- `ses_106dff374ffehA3ynYYaa2tc4e`

Children:

- `ses_106df4334ffeedaUIvfb4wHHg9`
- `ses_106def02dffeh7x393rfZ0sbXX`
- `ses_106df1551ffewhRRgv44xUqzQp`

### Timeline From `opencode.log`

The parent session exits its loop before the children finish:

```text
timestamp=2026-06-24T10:16:08.004Z level=INFO run=fc356f87 message=loop session.id=ses_106dff374ffehA3ynYYaa2tc4e step=2
timestamp=2026-06-24T10:16:08.004Z level=INFO run=fc356f87 message="exiting loop" session.id=ses_106dff374ffehA3ynYYaa2tc4e
```

The child sessions complete later:

```text
timestamp=2026-06-24T10:16:51.672Z level=INFO run=fc356f87 message="exiting loop" session.id=ses_106df4334ffeedaUIvfb4wHHg9
timestamp=2026-06-24T10:17:11.104Z level=INFO run=fc356f87 message="exiting loop" session.id=ses_106def02dffeh7x393rfZ0sbXX
[...]

PR #6160, opened 2026-07-16
fix(background-agent): restore active-turn completion admission

WhiteGiverMa:
## Summary

Restore immediate `noReply` admission for successful background completion wakes while the parent session is active.

PR #4478 established that successful completion reminders should join an active parent turn without starting an overlapping assistant reply. PR #5994 later added a 60-second active-defer ceiling, but its initial `sessionActive` guard returns before any admission path. Successful completions therefore remain invisible to the active turn until the parent becomes idle or the ceiling expires.

## Changes

- Admit the first successful active-turn wake with `forceNoReply: true`.
- Retain reply liveness for all-complete wakes and keep the existing retry timer.
- Continue deferring failures and already-admitted retained wakes through `deferReplyWakeWhileUnsafe`.
- Preserve the 60-second reply-producing fallback from #5994.
- Update active-turn event tests to assert immediate, coalesced `noReply` delivery.

## Why this is separate from #5994

#5994 bounds an otherwise indefinite defer, but it does not restore the immediate-delivery contract from #4478. This patch keeps that ceiling as the final reply-liveness fallback while making successful completion visible to the in-flight parent turn at its next context/tool boundary.

The broader symptoms in #5569 and #5790 are related, but neither tracks this exact early-return regression.

## Verification

- `bun test packages/omo-opencode/src/features/background-agent/parent-wake-active-turn-event.test.ts packages/omo-opencode/src/features/background-agent/parent-wake-active-defer-ceiling.test.ts` — 11 passed, 0 failed.
- `bun test packages/omo-opencode/src/features/background-agent` — 744 passed, 0 failed.
- LSP diagnostics — clean for both changed files.
- Isolated OpenCode serve probe with the equivalent patch observed the wake during the active parent turn (`WAKE_DISPATCHED_DURING_PARENT_TURN=true`), with one child session, one terminal parent stop, and no writes to the real session database.

### Baseline typecheck issue

`bun run typecheck` currently fails on unmodified `origin/dev` at:

```text
packages/omo-opencode/src/plugin/skill-context.ts(109,5): error TS2353:
'playwrightMcpArgs' does not exist in type 'ResolveActiveBuiltinSkillsOptions'.
```

The same diagnostic occurs with this PR diff stashed; neither changed file has diagnostics.

## Review focus

- Whether successful completion wakes should retain #4478's immediate active-turn visibility rather than wait for #5994's ceiling.
- Interaction between retained `noReply` admission, consumed-wake detection, and the existing 60-second reply fallback.

---
## Summary by cubic
Restores immediate noReply admission for successful background completions during an active parent turn, making the completion visible without starting an overlapping assistant reply. Keeps the 60-second fallback to produce a reply if the wake still needs one.

- **Bug Fixes**
  - Admit the first successful active-turn wake with `forceNoReply: true`, and retain it if `shouldReply` for later flushing.
  - Continue deferring failed or already-admitted wakes via `deferReplyWakeWhileUnsafe`.
  - Preserve the 60-second reply-producing ceiling.
  - Update active-turn event tests to assert immediate, coalesced noReply delivery.

<sup>Written for commit 61945718e59bf6f213d6bd36e9b4a1f6a7f3ad10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-active-turn-event.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-active-turn-event.test.ts
@@ -110,7 +110,7 @@ async function flushPendingParentWakeForTest(manager: BackgroundManager, session
 }
 
 describe("BackgroundManager parent wake active turn events", () => {
-  test("#when background task completes during active parent turn #then parent wake stays queued without prompt injection", async () => {
+  test("#when background task completes during active parent turn #then parent sees a noReply wake immediately", async () => {
     // given
     const sessionStatuses: Record<string, { type: string }> = {
       "parent-1": { type: "busy" },
@@ -132,11 +132,12 @@ describe("BackgroundManager parent wake active turn events", () => {
     await flushPendingParentWakeForTest(manager, "parent-1")
 
     // then
-    expect(promptAsyncCalls).toHaveLength(0)
-    expect(getPendingParentWakes(manager).has("parent-1")).toBe(true)
+    expect(promptAsyncCalls).toHaveLength(1)
+    expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
+    expect(getPendingParentWakes(manager).get("parent-1")?.shouldReply).toBe(true)
   })
 
-  test("#when duplicate background completions overlap an active parent turn #then one coalesced wake stays queued", async () => {
+  test("#when duplicate background completions overlap an active parent turn #then one coalesced noReply wake is admitted", async () => {
     // given
     const sessionStatuses: Record<string, { type: string }> = {
       "parent-1": { type: "busy" },
@@ -170,9 +171,11 @@ describe("BackgroundManager parent wake active turn events", () => {
     ])
 
     // then
-    expect(promptAsyncCalls).toHaveLength(0)
+    expect(promptAsyncCalls).toHaveLength(1)
+    expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
     const pendingWake = getPendingParentWakes(manager).get("parent-1")
     expect(pendingWake).toBeDefined()
+    expect(pendingWake?.shouldReply).toBe(true)
     expect(JSON.stringify(pendingWake?.notifications)).toContain("ALL BACKGROUND TASKS COMPLETE")
   })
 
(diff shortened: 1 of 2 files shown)


WhiteGiverMa:
I have used the build of the branch for 2 days myself.

MoerAI (changes requested):
Thanks for the PR, @WhiteGiverMa, and for taking the time on this.

Before this can land I need to understand the intent better, because this changes a deliberate behavior rather than fixing an unintended one.

On `dev`, `parent-wake-flush-runner.ts` intentionally defers when the parent session is active: it calls `schedulePendingParentWakeFlush(sessionID)` and logs "Deferred parent wake because parent session is active", with no prompt dispatch. This PR replaces that with an immediate `sendParentWakePrompt(...)` carrying `forceNoReply: true`. The two existing tests in `parent-wake-active-turn-event.test.ts` were inverted to match, from `expect(promptAsyncCalls).toHaveLength(0)` to `toHaveLength(1)`. Inverting an existing assertion is usually the signal that the old assertion was encoding an intended contract, not a defect.

This also sits in the area `AGENTS.md` calls out explicitly under "Internal message injection is dangerous", where every `session.prompt` / `session.promptAsync` route has to go through the shared gate, and where duplicate injection into a live parent session is the specific hazard the deferral was protecting against.

Two things would let me move forward:

1. A reproduction of the user-visible problem on current `dev`. What does the user actually observe today that this fixes? A regression test that fails on unmodified `dev` and passes on this branch, driving the real path rather than hand-built wake state, is the form I can act on. After #5316 had to be reverted in #5361, green CI on its own is not enough for a merge here.
2. A note on why immediate dispatch cannot produce the duplicate injection the deferral was added to prevent, especially for the coalesced case in the second test.

If the deferral really is wrong, I am glad to take the change, I just want the evidence recorded on the PR before the contract flips. No rush, and thanks again for digging into this area.

PR #6181, opened 2026-07-17
feat(shared-skills): add ulw-map, a deterministic wave execution map for ulw-plan work plans

fivetaku:
## What

`$ulw-map`: a new shared skill that renders a `ulw-plan` work plan as a **read-only, deterministic wave execution map** (Mermaid), with an optional single-file HTML viewer.

`ulw-plan` already ends at an explicit approval gate, and its plans lead with `## TL;DR (For humans)`. But the plan body itself is a long Markdown document: wave ordering, concurrent lanes, and the final verification fan-out are invisible at a glance, especially for the non-expert users who are asked to approve. `ulw-map` turns the plan they are approving into a picture, without letting any model draw it: the diagram is derived mechanically from the plan text, so it cannot contain structure the plan does not.

- **Parses only the mandated grammar** ([full-workflow.md](packages/shared-skills/skills/ulw-plan/references/full-workflow.md), [scaffold-plan.mjs](packages/shared-skills/skills/ulw-plan/scripts/scaffold-plan.mjs) template): column-zero `- [ ] N. <title>` rows in `## Todos`, `- [ ] F<n>. <title>` rows in `## Final verification wave`. Fence-aware with the CommonMark closing-fence rule (documentation examples never parse as real sections), CRLF-tolerant, byte-identical output under `node` and `bun`.
- **Wave grouping ladder, most-canonical first**: (1) the per-todo `Parallelization: Wave <N>` metadata line the ulw-plan template mandates; (2) column-zero `###` headings inside `## Todos`; (3) neither → a document-order chain. The emitted map states which rung produced the grouping, and tasks missing a metadata line inherit the previous wave with a named diagnostic.
- **Honest semantics, narrowly claimed**: arrows mean **wave barriers in plan document order, never inferred per-task dependencies**, and the emitted legend says so. Tasks inside a wave are lanes that may run concurrently. The final verification wave renders as the parallel all-must-approve fan-out that the upstream contract already defines.
- **Fail-soft, with diagnostics instead of guesses**: exit 0 = map generated (warnings embedded in the artifact), exit 1 = unsupported structure with a human-fixable message (missing `## Todos`, or checkbox rows outside the grammar, counted and named; task `0`, unsafe numbers, and blank titles are rejected), exit 2 = usage/IO. Duplicate numbers, fenced fakes, indented pseudo-tasks, and label injection are all handled and tested.
- **Read-only by construction**: the plan file is never modified; an output path that resolves to the source plan is refused (`--out-dir .` cannot clobber it), output confinement re-checks through real paths and refuses symlinked components (mirroring `scaffold-plan.mjs`'s write guards), and re-runs are byte-identical. Checked-off todos render as done, so re-running after execution shows progress.
- **Zero new dependencies**: node builtins only, same portability contract as `scaffold-plan.mjs` (the CLI sets `process.exitCode` so large `--stdout` output always drains; the Windows opener is `explorer.exe` with a literal path, no `cmd` reparse surface). The HTML viewer is source-first; the Mermaid text is the page content; the renderer loads from an exact-version CDN URL (`mermaid@11.16.0`, asserted in tests as the page's only remote reference) and the page degrades to the readable source offline. Nothing is vendored. `--open` (browser) is opt-in only.

## Why

Text is a good format for writing plans; it is a poor format for *approving* them. The approval gate is the one moment a human, often a non-developer, must actually understand the plan, and `## TL;DR (For humans)` already acknowledges that the raw plan body is not that interface. This is the same idea one step further: the structure the plan already encodes (`Parallelization: Wave <N>`, the final verification contract), shown instead of described. Deriving it deterministically (rather than asking a model to draw a diagram) makes plan/diagram drift structurally impossible.

## Demo (dogfooding: this contribution's own plan in the canonical template shape, block below generated by the script)

```mermaid
flowchart TD
    subgraph W1["Wave 1"]
        T1["✓ 1. Fence-aware parser for the mandated todo grammar"]
        T2["✓ 2. Wave grouping ladder (metadata > headings > document order)"]
        T3["✓ 3. Mermaid emitter with sanitized labels and done-state classes"]
    end
    subgraph W2["Wave 2"]
        T4["✓ 4. Map Markdown artifact with legend and diagnostics"]
        T5["✓ 5. HTML viewer (source-first, exact-pin CDN renderer, offline …"]
    end
    subgraph W3["Wave 3"]
        T6["6. CLI: slug resolution, symlink-safe .omo/maps confinement, -…"]
        T7["7. Contract tests: fences, duplicates, injection, CRLF determi…"]
    end
    W1 --> W2
    W2 --> W3
    F1["F1. Plan compliance audit"]
    F2["F2. Code quality review"]
    F3["F3. Real manual QA"]
    F4["F4. Scope fidelity"]
    W3 --> F1
    W3 --> F2
    W3 --> F3
    W3 --> F4
    FIN(["All approve → complete"])
    F1 --> FIN
    F2 --> FIN
    F3 --> FIN
    F4 --> FIN
    classDef done fill:#14532d,stroke:#22c55e,color:#ffffff
    class T1,T2,T3,T4,T5 done
```

Fail-soft, exercised live: a document without `## Todos` exits 1 with `unsupported plan structure: no \`## Todos\` section found outside code fences...`; checkbox rows outside the grammar exit 1 with the ignored-row count; an unknown flag or `--out-dir` escape exits 2.

## Scope (deliberately narrow)

- `ulw-plan` templates, prompts, and contracts are untouched; this is a viewer, not a protocol change.
- No per-task dependency inference: the map claims wave barriers and the upstream-defined parallel semantics, nothing more. (Rendering the plan's *declared* `Blocked by:` edges is a possible follow-up, kept out on purpose.)
- No vendored renderer, no image generation. A future dependency-free `--format svg` and an artifact/structure map are possible follow-ups, also kept out.
- Registration footprint: `expectedSkills` in the sync contract test + the shared-skills `AGENTS.md` skill list (count updates included). `sync-skills.mjs` picks the skill up without changes.

## Testing

- `node --test test/plan-map.test.mjs`: **23 pass**: wave-metadata grouping (canonical template fixture), metadata-inheritance diagnostic, heading and document-order ladder rungs, fenced fakes + CommonMark pseudo-closer, missing-section and wrong-grammar diagnostics (including task `0`, unsafe numbers, blank titles), duplicate numbers, label injection, CRLF/LF determinism, nested-checkbox exclusion, HTML escaping + exact-pin single remote URL, `--check` writes nothing, default-mode idempotency (double run byte-identical, source untouched), source-overwrite refusal, unfilled-template-placeholder diagnostic, lexical + symlink `--out-dir` escape refusal, `--out-dir --check` usage error, slug resolution. Temp trees are cleaned in `finally`.
- `node --test test/*.test.mjs` (omo-codex plugin): **376 pass, 0 fail** including the sync-skills drift contract with the new skill registered.
- `bun test` at the repo root: **11,642 pass, 0 fail** (full monorepo, pre-review baseline; affected suites re-run green after review fixes).
- Live CLI: canonical-template plan → 3 wave groups from `Parallelization` metadata; triple run `diff` empty; `node` and `bun` `--stdout` byte-identical (same sha1).
- Post-build review: an independent Codex CLI review of the staged diff returned FIX-THEN-SUBMIT with 13 findings (2 blockers: canonical `Parallelization: Wave <N>` metadata was not parsed; `--out-dir .` could overwrite the source plan). All 13 are fixed above and covered by the new tests.

Naming note: `ulw-map` is provisional; happy to rename (e.g. `ulw-plan-map`) if you prefer the input to be explicit in the name.

### DIFF
--- a/packages/omo-codex/plugin/test/plan-map.test.mjs
+++ b/packages/omo-codex/plugin/test/plan-map.test.mjs
@@ -441,3 +441,20 @@ test("#given a slug instead of a path #when the plan lives in .omo/plans #then i
 		assert.equal(code, 0);
 	});
 });
+
+test("#given a fresh scaffold skeleton whose placeholder todo was never filled #when parsed #then the map warns about the template placeholder", async () => {
+	// given
+	const { parsePlanMap } = await import(scriptUrl);
+	const plan = `# fresh - Work Plan
+
+## Todos
+
+- [ ] 1. <title>
+  Parallelization: Wave <N> | Blocked by: <...> | Blocks: <...>
+`;
+	// when
+	const map = parsePlanMap(plan);
+	// then --- the map still renders, but says the plan may not contain real todos yet
+	assert.equal(map.taskCount, 1);
+	assert.ok(map.warnings.some((warning) => warning.includes("unfilled template placeholder")));
+});
--- a/packages/omo-codex/plugin/test/plan-map.test.mjs
+++ b/packages/omo-codex/plugin/test/plan-map.test.mjs
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import { mkdtemp, readFile, readdir, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, readFile, readdir, rm, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import test from "node:test";
@@ -11,11 +11,39 @@ const root = dirname(dirname(fileURLToPath(import.meta.url)));
 const scriptPath = join(root, "..", "..", "shared-skills", "skills", "ulw-map", "scripts", "plan-map.mjs");
 const scriptUrl = pathToFileURL(scriptPath).href;
 
-const GROUPED_PLAN = `# demo - Work Plan
+async function withTempDir(run) {
+	const dir = await mkdtemp(join(tmpdir(), "plan-map-"));
+	try {
+		return await run(dir);
+	} finally {
+		await rm(dir, { recursive: true, force: true });
+	}
+}
+
+const CANONICAL_PLAN = `# canonical - Work Plan
 
 ## TL;DR (For humans)
 
 What you'll get.
+
+## Todos
+
+- [x] 1. Build the parser
+  What to do / Must NOT do: parse only the mandated grammar
+  Parallelization: Wave 1 | Blocked by: - | Blocks: 3
+  Acceptance criteria (agent-executable): node --test passes
+- [ ] 2. Add fixtures
+  Parallelization: Wave 1 | Blocked by: - | Blocks: 3
+- [ ] 3. Emit mermaid
+  Parallelization: Wave 2 | Blocked by: 1, 2 | Blocks: -
+
+## Final verification wave
+
+- [ ] F1. Plan compliance audit
+- [ ] F2. Code quality review
+`;
+
+const HEADING_PLAN = `# demo - Work Plan
 
 ## Todos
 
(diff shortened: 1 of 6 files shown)


code-yeongyu:

PR #6199, opened 2026-07-18
fix(category-skill-reminder): append reminder as new message to preserve prompt cache

codegiveness:
## Summary

The `category-skill-reminder` hook spliced a synthetic text part into an existing user message's `parts` array. This mutated the stored content of that message, which changed its cumulative prefix hash and invalidated the Anthropic prompt cache for that message and everything after it. On the next turn, when `reminderShown` was true and the reminder was not re-injected, the stored message content differed from what was cached upstream, causing a second cache miss and a lost cache breakpoint.

This change appends the reminder as a **new** user message at the end of `output.messages` instead. Prior messages and their `parts` arrays stay byte-identical, so the cache prefix remains valid on every subsequent turn.

## Changes

- **`packages/omo-opencode/src/hooks/category-skill-reminder/hook.ts`**: replaced the `target.message.parts.splice(...)` block in `messagesTransform` with an `output.messages.push(...)` that appends a new message. The new message clones `info` from the target user message via spread (only `id` is overridden) and carries a single synthetic text part. Net change: 3 lines.
- **`packages/omo-opencode/src/hooks/category-skill-reminder/index.test.ts`**: rewrote three tests that asserted part-level positioning inside an existing message to assert the reminder appears as a new entry in `output.messages` with `role: "user"`. Added one test that verifies prior messages are reference-stable and byte-identical before and after the transform runs (the cache-preservation invariant).

## How this preserves the cache

Anthropic prompt caching hashes the prefix cumulatively: changing any block at or before the breakpoint produces a different hash on the next request, and the messages cache is invalidated by content changes at the messages level. Splicing into an existing user message mutated that message's content, so its hash changed and the breakpoint at or after it missed. Appending a new message does not mutate any prior message, so every prior message's content hash is stable and the cache prefix stays valid.

Anthropic accepts consecutive user messages without error, so the appended reminder is read as additional user input. No `role: "system"` messages are introduced, so this works on Anthropic, OpenAI, and every Anthropic-compatible provider without model gating.

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/hooks/category-skill-reminder/`
  **Observed result:** 21 pass, 0 fail, 83 expect() calls. Three rewritten tests assert the new append behavior; one new test directly verifies the cache-preservation invariant (prior messages reference-stable and byte-identical before/after transform).
  **Artifact:** reproducible from the branch.
  **Why sufficient:** the change is a pure data-structure mutation in one hook function; the unit suite covers the new behavior, the once-per-session gate, session-deletion cleanup, cross-session isolation, case-insensitive counting, and skill formatting.

- **What was tested:** `bun run typecheck` (full workspace, `tsgo --noEmit` + `typecheck:script` + `typecheck:packages`)
  **Observed result:** exit 0, no errors across all 29 package tsconfig projects.
  **Why sufficient:** confirms the cloned `info` spread is type-safe and no `as any` / `@ts-ignore` was introduced.

- **What was tested:** `bunx biome lint` on both changed files
  **Observed result:** exit 0, no warnings.

- **What was tested:** `bun test packages/omo-opencode/src/hooks/` (adjacent regression check)
  **Observed result:** 2048 pass, 46 fail. The 46 failures are pre-existing and unrelated: verified by stashing this change and re-running the same suite, which produces the identical 46 failures (2047 pass, 46 fail) on baseline `dev`. The failures are cross-test state pollution in `claude-code-hooks/handlers/session-event-handler-retry.test.ts` and related files; each failing test passes in isolation. This change adds one net passing test (2047 to 2048) and touches none of the failing tests.

- **Not covered:** the full `opencode-qa` skill harness (live OpenCode session driving with SSE hook probes) was not run in this environment. The change is a 3-line behavior swap inside one hook with passing unit tests; the maintainer is welcome to run the full `opencode-qa` harness before merge. `bun run build` was not run because the frontend submodule materialization step requires network access to private submodules not available in this checkout; this does not affect the hook source, which is plain TypeScript with no build-time codegen.

## Risks & Residuals

- **Weaker precedence signal:** the reminder now arrives as a trailing user turn rather than prepended text inside the existing user turn. This is a weaker precedence signal than a system message would be, which is appropriate for a nudge rather than a directive. Accepted.
- **`textPartIndex` in `findLatestReminderTarget` is now unused by `messagesTransform`:** the inner `parts` scan still serves as a meaningful guard (it ensures injection only happens when a real user message exists to clone `info` from, which the `pending-without-text` test depends on). Left unchanged to keep the diff minimal; can be simplified in a follow-up. Not blocking.
- **Cloned `info.time`:** the new message inherits the original message's `time.created` timestamp. Downstream consumers that sort or dedupe by `time.created` could behave oddly. Non-blocking; flag for follow-up if it matters for dashboard ordering.
- **Harness-level QA not run in this environment:** see QA section above. The unit-level evidence covers the hook contract; the maintainer can run `opencode-qa` for full harness proof before merge.

## Automated Checks

```bash
bun run typecheck
bun test packages/omo-opencode/src/hooks/category-skill-reminder/
bunx biome lint packages/omo-opencode/src/hooks/category-skill-reminder/hook.ts packages/omo-opencode/src/hooks/category-skill-reminder/index.test.ts
```

## Related Issues

---
## Summary by cubic
Append the `category-skill-reminder` as a new user message instead of splicing it into an existing one to preserve the Anthropic prompt cache. Prior messages stay byte-identical, so cache breakpoints remain valid across turns.

- **Bug Fixes**
  - Swapped `parts.splice(...)` for `output.messages.push(...)` to append a new `user` message that clones `info` (with a new `id`) and carries the reminder text.
  - Ensures earlier messages and their `parts` are unchanged, keeping cumulative prefix hashes stable.
  - Updated tests to assert append behavior and added a test that verifies cache-preservation (prior messages remain reference- and byte-identical).

<sup>Written for commit 454713da0e567ba72645b5a0933feec546013037. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6199?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/category-skill-reminder/hook.ts
+++ b/packages/omo-opencode/src/hooks/category-skill-reminder/hook.ts
@@ -156,13 +156,17 @@ export function createCategorySkillReminderHook(
     const target = findLatestReminderTarget(output.messages, sessionStates)
     if (!target) return
 
-    target.message.parts.splice(target.textPartIndex, 0, {
-      id: `prt_category_skill_reminder_${target.messageID}`,
-      sessionID: target.sessionID,
-      messageID: target.messageID,
-      type: "text",
-      text: reminderMessage,
-      synthetic: true,
+    const reminderID = `msg_category_skill_reminder_${target.sessionID}`
+    output.messages.push({
+      info: { ...target.message.info, id: reminderID },
+      parts: [{
+        id: `prt_category_skill_reminder_${target.sessionID}`,
+        sessionID: target.sessionID,
+        messageID: reminderID,
+        type: "text",
+        text: reminderMessage,
+        synthetic: true,
+      }],
     })
     target.state.reminderPending = false
     target.state.reminderShown = true
--- a/packages/omo-opencode/src/hooks/category-skill-reminder/index.test.ts
+++ b/packages/omo-opencode/src/hooks/category-skill-reminder/index.test.ts
@@ -94,7 +94,7 @@ afterEach(() => {
 
 describe("category-skill-reminder hook", () => {
   test.each(["Sisyphus", "Atlas", "sisyphus-junior"])(
-    "#given target agent %s #when three delegatable tools finish #then one reminder is injected before the user text",
+    "#given target agent %s #when three delegatable tools finish #then one reminder is appended as a new user message",
     async (agent) => {
       const hook = createHook()
       const sessionID = `target-${agent}`
@@ -106,8 +106,13 @@ describe("category-skill-reminder hook", () => {
 
       expect(output.output).toBe("result")
       expect(findReminderParts(messages)).toHaveLength(1)
-      expect(messages[0]?.parts[0]).toMatchObject({ synthetic: true, type: "text" })
-      expect(messages[0]?.parts[1]).toMatchObject({ text: "continue working", type: "text" })
+      const originalMessage = messages[0]
+      expect(messages).toHaveLength(2)
+      expect(originalMessage.parts).toHaveLength(1)
+      expect(originalMessage.parts[0]).toMatchObject({ text: "continue working", type: "text" })
+      const appended = messages[1]
+      expect(appended.info.role).toBe("user")
+      expect(appended.parts[0]).toMatchObject({ synthetic: true, type: "text" })
     },
   )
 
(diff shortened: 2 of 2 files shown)


codegiveness:
Re-reading the precedence risk I flagged in the PR description — I think I had the direction backwards. Since the reminder is now the *last* `user` message, it's the final thing the model sees before generating. That means the risk isn't that it gets "weaker" — if anything, it could be *over*-attended.

The actual concern is the model possibly interpreting the `task()` code block in the reminder as a literal instruction rather than advisory guidance. The `[Category+Skill Reminder]` header and blockquote formatting should signal it's a system-injected reminder, and the target agents (sisyphus, atlas) are designed to parse delegation context, so I think it's fine in practice. But wanted to correct my earlier assessment.

code-yeongyu:
[sisyphus-bot] REVIEW-OK: Correct and minimal fix — appending as a new message preserves the prompt cache prefix, which splicing into an existing message's parts array did not. Test coverage is excellent, including a direct byte-identical invariant test. The only blocker is the CLA: @codegiveness needs to sign before this can merge. Once CLA is resolved, this is ready for an owner's approval.

MoerAI:
Thanks for identifying the prompt-cache mutation and for adding focused byte-stability coverage.

Closing as superseded by #7253:
https://github.com/code-yeongyu/oh-my-openagent/pull/7253

This branch now conflicts with `dev`, fails CLA and current required-CI gates, and its exact head fails the existing `messages-transform` integration suite. Appending a cloned user message adds fields outside the established transform shape and produces consecutive user messages after a normal tool-result turn.

#7253 preserves sent history while attaching the reminder to the current unsent tool-result user turn, which retains tool-result-first ordering and targets the correct cache boundary. Please continue review there rather than reopening this branch.

PR #6210, opened 2026-07-18
feat(start-work): auto-scaffold notepad files on plan commit

acamq:
## What

`/start-work` now auto-scaffolds the 4 notepad files (with a pre-seeded header) at the moment a plan is committed. This unblocks a chicken-and-egg deadlock: `notepad-write-guard` blocks the `write` tool on `.omo/notepads/**` unconditionally (append-only protection), and `edit` cannot seed an empty file, so the agent could never create the first notepad entry.

The hook writes via `node:fs` directly at the 3 boulder-state write sites, so the tool guard is bypassed and **stays unchanged**.

## Why

The deadlock was hit in production (ulw with atlas): the agent tried `Write` (blocked by guard), fell back to `bash touch` + `Edit` with empty oldString (rejected). There was no supported path to create the first notepad line. The guard's *intent* (protect accumulated learnings from overwrite) is sound; only the bootstrap case was unsupported. The directive (`sisyphus-junior-notepad`) also told workers to append but forbade both `write` and `edit` - contradictory.

The fix matches the guard's real intent: scaffold once at plan-commit (no history to destroy), then the agent only ever appends (which the guard allows and the directive wants).

## Changes

**Core behavior** (`27beb30dd`)
- New `ensureNotepadScaffold({directory, planName, timestamp?})` in `packages/omo-opencode/src/hooks/start-work/notepad-scaffold.ts`: `mkdir -p .omo/notepads/<plan>` + per-file `writeFileSync({flag:"wx"})` (EEXIST -> skipped, so idempotent and partial-safe). Pre-seeds each file with a header (title with em-dash separator + one-line purpose + "Auto-scaffolded by /start-work. Append new entries below - never overwrite." footer).
- Wired into the 3 boulder-state write sites:
  - `createNewWorkOrInitialize` (work-initializer.ts) - explicit + new work
  - `buildAutoSelectedPlanContextWithStateInit` (work-initializer.ts) - auto-selected single plan
  - `buildExistingSessionContext` (context-info-formatters.ts) - resume, AFTER its `isComplete` early-return so completed plans don't trigger it
- `planName` via the existing `getPlanName(planPath)` / `existingState.plan_name`. Does NOT scaffold on the multiple-works (ask-user) or no-plan branches.

**Directive + prompt docs** (`1345dfe99`)
- `sisyphus-junior-notepad/constants.ts`: replaced "never use Edit tool" with steer to `edit`/bash `>>` (write is blocked, overwrite forbidden).
- 8 atlas prompt variants (`default`, `gpt`, `gemini`, `glm`, `kimi`, `kimi-k2-7`, `kimi-k3`, `opus-4-7`): replaced the manual "Step 2: Initialize Notepad" (`mkdir` + structure tree) with an auto-scaffolded note, and fixed the Notepad System/Protocol append directive line.

## Observed behavior

- Unit tests (`notepad-scaffold.test.ts`, 4 cases): create / idempotent (re-run byte-identical) / partial (2 of 4 pre-existing preserved, 2 created) / nested mkdir - all GREEN.
- Integration tests (`context-info-builder.test.ts`, 4 cases): auto-select / explicit / resume all scaffold the 4 files; multiple-works does NOT scaffold - all GREEN.
- Full start-work suite: 85 pass / 0 fail.
- Plugin bundles: `dist/index.js` 5.48MB / 1922 modules, exit 0.
- `notepad-write-guard/` is untouched (verified: `git diff` shows no entries there).

## QA / Evidence

The scaffold logic is proven at three levels; the live `opencode run` end-to-end is a documented partial (environment blocker, not a code defect):

| Layer | Status |
|---|---|
| Unit tests (real fs, temp dir) | GREEN |
| Integration tests (real `buildStartWorkContextInfo` -> write sites -> scaffold) | GREEN |
| Plugin-level hook proof (drove real hook module, 4 files + idempotency via shasum + host DB isolation 6147->6147) | GREEN |
| Live `opencode run "/start-work"` end-to-end | NOT reproduced - the bare worktree plugin couldn't fully materialize (`dist/skills/frontend/SKILL.md` ENOENT, so commands didn't register; `--command start-work` hit UnknownError). Blocked by incomplete plugin asset materialization in the worktree, not by this change. |

Evidence dir: `.omo/evidence/20260717-notepad-scaffold/` (what-was-tested, what-was-observed, why-it-is-enough, what-was-omitted, raw run logs, source-hook-proof.json).

## Residual risk

- The scaffold writes to `<directory>/.omo/notepads/<plan>/` where `directory` is the same root `boulder.json` uses. For worktree sessions where the agent's cwd is the worktree, the relative `.omo/notepads/` read resolves to the worktree cwd, while the scaffold wrote to the main `directory`. This matches the existing `boulder.json` location behavior (same class of pre-existing path resolution), so it is consistent, not a regression. Flagged as a possible follow-up if worktree notepad isolation is later desired.
- The integration tests exercise the exact production code path the live hook calls (`buildStartWorkContextInfo`); the hook wiring itself is unchanged shipped code, so the scaffold is a proven side effect of an already-firing hook.

## Notes

- Pre-existing (not introduced): `build:cli` fails on `@oh-my-opencode/omo-senpi/install` resolution - confirmed on base via stash. Blocks the full `bun run build` independent of this change; the plugin entry (`dist/index.js`) builds cleanly in isolation.
- 14 files changed, +412/-53. 2 atomic commits.

---
## Summary by cubic
/start-work now auto-scaffolds `.omo/notepads/{plan}/` with four files on plan commit, fixing the append-only deadlock with `notepad-write-guard`. Prompts and directives now tell agents to append with `edit` or bash `>>` and never use `write`.

- **New Features**
  - Added `ensureNotepadScaffold({directory, planName, timestamp?})`: creates `learnings.md`, `decisions.md`, `issues.md`, `problems.md` with a small header. Uses `mkdir -p` and `writeFileSync` with `flag: "wx"` for idempotency; existing files are preserved.
  - Wired into plan commit points: explicit new work, auto-selected single plan, and resume path (skipped for multi-work and no-plan branches). The `notepad-write-guard` stays unchanged; scaffolding writes via `node:fs` to bypass tool guards safely.
  - Updated `sisyphus-junior-notepad` and all `atlas` prompt variants: removed manual notepad init; instruct to append only with `edit` or bash `>>`; never use `write`.

- **Migration**
  - No action needed. Files are created automatically on plan commit and reruns are safe. Existing notepads are left untouched.

<sup>Written for commit 1345dfe99e0e251d34772943af977a35009199c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/hooks/sisyphus-junior-notepad/constants.ts
+++ b/packages/omo-opencode/src/hooks/sisyphus-junior-notepad/constants.ts
@@ -10,7 +10,7 @@ NOTEPAD PATH: .omo/notepads/{plan-name}/
 - problems.md: Record unresolved issues, technical debt
 
 You SHOULD append findings to notepad files after completing work.
-IMPORTANT: Always APPEND to notepad files - never overwrite or use Edit tool.
+Notepad files are auto-scaffolded by /start-work. APPEND only - use the \`edit\` tool (match-and-insert after the last line) or \`bash\` \`>>\`. Never use the \`write\` tool (blocked by notepad-write-guard) and never overwrite.
 
 ## Plan Location (subagent: READ ONLY)
 PLAN PATH: .omo/plans/{plan-name}.md
--- a/packages/omo-opencode/src/hooks/start-work/context-info-builder.test.ts
+++ b/packages/omo-opencode/src/hooks/start-work/context-info-builder.test.ts
@@ -412,4 +412,109 @@ describe("buildStartWorkContextInfo", () => {
     }
     expect(workIds).toContain(workC.work_id)
   })
+
+  describe("notepad scaffolding side-effect", () => {
+    test("#given single incomplete plan and no active work #when buildStartWorkContextInfo auto-selects it #then scaffolds .omo/notepads/<plan-basename>/{learnings,decisions,issues,problems}.md", () => {
+      // given
+      writePlan("auto-scaffold-plan", "## TODOs\n- [ ] 1. Auto task")
+
+      // when
+      buildStartWorkContextInfo({
+        ctx: createPluginInput(),
+        explicitPlanName: null,
+        existingState: null,
+        sessionId: "session-current",
+        timestamp: "2026-05-11T00:00:00.000Z",
+        activeAgent: "atlas",
+        worktreePath: undefined,
+        worktreeBlock: "",
+      })
+
+      // then
+      const notepadDir = join(testDirectory, ".omo", "notepads", "auto-scaffold-plan")
+      expect(existsSync(join(notepadDir, "learnings.md"))).toBe(true)
+      expect(existsSync(join(notepadDir, "decisions.md"))).toBe(true)
+      expect(existsSync(join(notepadDir, "issues.md"))).toBe(true)
+      expect(existsSync(join(notepadDir, "problems.md"))).toBe(true)
+    })
+
+    test("#given explicit plan name matching a plan file #when built with explicitPlanName #then scaffolds notepad for that plan", () => {
+      // given
+      writePlan("explicit-scaffold-plan", "## TODOs\n- [ ] 1. Explicit task")
+
+      // when
+      buildStartWorkContextInfo({
+        ctx: createPluginInput(),
+        explicitPlanName: "explicit-scaffold-plan",
+        existingState: null,
+        sessionId: "session-current",
+        timestamp: "2026-05-11T00:00:00.000Z",
+        activeAgent: "atlas",
+        worktreePath: undefined,
+        worktreeBlock: "",
+      })
+
+      // then
+      const notepadDir = join(testDirectory, ".omo", "notepads", "explicit-scaffold-plan")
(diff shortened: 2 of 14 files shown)


acamq (commented):
Reviewed the implementation and tests. I did not find a blocking issue.

The scaffold uses `writeFileSync(..., { flag: "wx" })`, so reruns preserve existing notepads and partial scaffolds are repaired safely. The wiring covers explicit selection, auto-selection, and resume while avoiding the multi-work prompt branch. The prompt/directive updates also match the append-only guard semantics.

PR #6216, opened 2026-07-19
fix(omo-senpi): prevent duplicate ultrawork directive injection

tmdgusya:
## Summary

The senpi edition's ultrawork input hook matched `ulw` as a bare substring and always PREPENDED the full ~17KB directive. Three user-visible defects followed: typing a `ulw-`-prefixed skill name (`/skill:ulw-plan ...`) armed ultrawork mode it was never asked for; prepending broke senpi's native `/skill:` expansion (senpi expands only while the text still starts with `/skill:`), so the skill body was never inlined; and pasted transcripts already carrying an `<ultrawork-mode>` block got a second full copy. On top of that, the shipped skill description still used the Codex pointer wording ("reads a short bootstrap... read the whole file"), so models re-read the ~24KB SKILL.md whose body is 100% contained in the directive that was already injected inline — the same rules landing twice in context.

After this PR: `ulw-`-named skills no longer arm the mode, `/skill:` commands keep native expansion with the directive appended after them, already-armed input is never re-injected, and the shipped description tells the model NOT to re-read the file when the block is present.

## Changes

- **Trigger guard** (`packages/omo-senpi/src/components/ultrawork/index.ts`): pattern becomes `/(?:ultrawork|ulw(?!-))/i` — skips the `ulw-` skill-name family (`ulw-plan`, `ulw-loop`, `ulw-research`) while keeping the generous matching pinned by existing tests (`하이ulw`, `ulw_helper.ts`).
- **Reinjection guard** (same file): input already containing `<ultrawork-mode>` is passed through untouched.
- **`/skill:` append** (same file): for `/skill:`-prefixed input the directive is appended after the command instead of prepended, preserving senpi's `startsWith("/skill:")` expansion contract.
- **Shipped description rewrite** (`packages/omo-senpi/plugin/scripts/sync-skills.mjs`): the ultrawork component skill now gets a senpi-accurate frontmatter description during sync (full inline injection; do not re-read when the block is present). `plugin/skills/` stays generated-only; no hand-edits.
- **Tests** (`ultrawork.test.ts`): 4 new cases written failing-first — `ulw-` names inject nothing, marker input is not re-injected, `/skill:` input keeps its prefix with exactly one directive appended, and the synced artifact's description documents inline injection.
- **Docs + evidence**: `packages/omo-senpi/AGENTS.md` component bullet updated; QA evidence under `.omo/evidence/20260719-omo-senpi-ultrawork-dedup/`.

## QA & Evidence

- **What was tested:** live senpi harness — `driver.mjs` runs the real senpi binary (2026.7.17-5) with this branch's built plugin in an isolated `SENPI_CODING_AGENT_DIR` mktemp sandbox and the bundled mock provider (zero tokens). Four scenarios assert recorded session content: bare `ulw` still injects; `ulw-plan` does not; a prompt already carrying `<ultrawork-mode>` is not re-injected; `/skill:frontend ulw ...` shows the expanded `<skill name="frontend">` block AND the appended directive.
  **Observed:** 4/4 PASS, `realSenpiUntouched: true` (real agent dir `settings.json`/`trust.json` digests unchanged).
  **Artifact:** `.omo/evidence/20260719-omo-senpi-ultrawork-dedup/driver-output.json` (+ `driver.mjs`, `notes.md`)
  **Why sufficient:** proves the end-to-end behavior of the BUILT plugin inside a real senpi session for all three fixed defects plus the trigger regression path.
- **What was tested:** unit suite (failing-first) + package gate `bun run test:senpi`.
  **Observed:** component tests 13/13 (4 new cases captured RED against the old component first); package gate 230 pass / 0 fail.
  **Artifact:** `.omo/evidence/20260719-omo-senpi-ultrawork-dedup/test-summary.txt`
  **Why sufficient:** pins the full input-transform decision table (trigger forms, guards, recursion source, disabled flag, malformed payloads) and the generated-artifact contract (description sentinels, embed-directive drift check).
- **What was tested:** repo gates — root `bun run typecheck`, root `bun test` (CI parity).
  **Observed:** typecheck exit 0; 11749 pass / 0 fail / 3 skip.
  **Artifact:** `.omo/evidence/20260719-omo-senpi-ultrawork-dedup/test-summary.txt`

## Risks & Residuals

- Substring false positives outside the `ulw-` family (e.g. "Bulwark") still trigger: the existing suite deliberately pins generous matching (`하이ulw`, `ulw_helper.ts`), so this PR only excludes the hyphenated skill-name family it collides with. Accepted; the reinjection guard now caps the damage of the paste case.
- The description rewrite lives in the sync script (Codex source keeps its pointer wording, which is correct for Codex). Codex edition behavior is unchanged.
- `/skill:` append is only reachable when a skill command also contains a standalone trigger word; plain `/skill:ulw-plan ...` no longer transforms at all.

## Related Issues

None — surfaced from a live senpi session where `/skill:ulw-plan` armed ultrawork mode and double-loaded the directive.

---
## Summary by cubic
Fixes duplicate ultrawork directive injection in `omo-senpi` and prevents accidental activation from `ulw-...` skill names. Hardens `/skill:` handling and tag checks so native skill expansion works and no second copy is added.

- Bug Fixes
  - Match `ultrawork|ulw(?!-)` so `ulw-plan`/`ulw-loop`/`ulw-research` don’t trigger.
  - Skip injection only when a matched `<ultrawork-mode>`...`</ultrawork-mode>` block exists; a lone open-tag mention still arms.
  - Mirror Senpi `/skill:` parse: append the directive only when args carry a trigger; `/skill:ultrawork` passes through untouched; triggers in the skill NAME do not arm.

- Refactors
  - Sync rewrites the ultrawork `SKILL.md` description to reflect inline injection and to avoid re-reading when the block is present.
  - Added tests and a live Senpi QA harness covering trigger, reinjection, `/skill:ultrawork`, and `/skill:` args behavior.

<sup>Written for commit 8a96672233d457482c9754d9597d2160d08c08d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tmdgusya:
## Token-cost measurement for the duplication this PR removes

Measured with `tiktoken` `o200k_base` as a proxy tokenizer (provider tokenizers vary by roughly +/-10%).

| Artifact | Chars | Tokens |
| --- | ---: | ---: |
| Inline-injected directive (`SENPI_ULTRAWORK_DIRECTIVE`) | 17,189 | 4,016 |
| Shipped `ultrawork` SKILL.md (directive is 100% contained in it) | 24,377 | 5,747 |

### Savings per incident

| Incident fixed by this PR | Tokens saved |
| --- | ---: |
| `ulw-` skill-name misfire turn: unwanted directive injection + the model re-reading SKILL.md (both observed in the motivating session) | ~9,763 |
| Prompt already carrying an `<ultrawork-mode>` block (pasted transcript) re-injected by the hook | ~4,016 |
| Legitimate `ulw` session where the stale Codex-pointer description made the model re-read SKILL.md | ~5,747 |

### Real-session evidence (reporting machine)

Scanned 42 local senpi session transcripts with exact 200-char slice detectors (matched in JSONL-escaped space; a slice unique to SKILL.md distinguishes re-reads from injections):

- 8 sessions carried the injected directive.
- 1 session recorded the exact double-load this PR removes: injected directive **plus** a full SKILL.md re-read in the same session = 5,747 duplicated tokens.
- The session that motivated this PR additionally contained a TUI-reformatted paste of the directive that the hook re-injected on top (~4,016 more). Reformatting defeats byte-exact detection, so the scan undercounts paste cases.

### Why the absolute numbers understate the cost

The duplicated blob lives in the conversation window, so it is resent with **every subsequent request** in that session: the ~9,763-token misfire case costs ~293k cumulative input tokens over a 30-request session (prompt caching discounts the price, not the context-window occupancy). As a share of context, ~9.8k tokens is roughly 20-40% of a fresh session's first request (~25-50k with system prompt + skills) and ~5% of a full 200k window.

code-yeongyu:
## Post-review hardening (maintainer pass)

Adversarial review of the guards surfaced two real defects; fixed in 8a9667223 and force-pushed.

### 1. `/skill:ultrawork` was the one command the dedup missed
Senpi's `_expandSkillCommand` inlines the full `SKILL.md` — whose body IS the directive — then appends the transformed remainder:
- `/skill:ultrawork fix this` → expansion (1 directive) **+** the hook's appended copy (2nd directive). The exact duplication this PR exists to remove.
- Bare `/skill:ultrawork` → senpi parses the skill name up to the **first space only**, so the appended `\n<ultrawork-mode>…` corrupted the parsed name and expansion silently failed (raw slash command reached the model).

The hook now mirrors senpi's exact name/args parse: `/skill:ultrawork` passes through untouched, a trigger that appears only in the skill NAME does not arm, and only trigger-carrying args get the appended directive.

### 2. Reinjection guard disarmed on a mere mention
`text.includes("<ultrawork-mode>")` meant `Explain what <ultrawork-mode> means, then ulw this fix` silently skipped injection. The guard now requires the matched open+close tag pair — a pasted block or earlier injection always carries both.

### Verification
- Unit: 16/16 (3 new cases; `/skill:ultrawork` with and without args, name-only trigger, lone open-tag mention)
- Live driver (real senpi binary + built plugin, isolated sandbox, mock provider): **6/6 PASS**, `realSenpiUntouched: true` — including `skill-ultrawork-not-duplicated` asserting exactly ONE `<ultrawork-mode>` open tag in the recorded session
- `bun run test:senpi`: 233 pass / 0 fail; root typecheck exit 0
- Evidence refreshed under `.omo/evidence/20260719-omo-senpi-ultrawork-dedup/`

PR #5204, opened 2026-06-12
feat(background-agent): persist background tasks across OpenCode restarts

tracycam:
## Problem

Background task bookkeeping (`BackgroundManager.tasks`, `completedTaskArchive`, the `globalThis` registry) lives entirely in process memory. When OpenCode restarts, every `bg_...` → `ses_...` mapping evaporates: `background_output(task_id="bg_...")` returns "Task not found", completion notifications never fire, and `task(task_id=..., run_in_background=true)` resumes fail — even though the subagent sessions themselves survive in SQLite.

A previous file-backed implementation (single shared `.opencode/background-tasks.json`) was removed in 24a7f333a because multiple OpenCode instances sharing a project directory overwrote each other's state. This PR re-introduces persistence with a design that structurally eliminates that race.

## Design

- **Per-task snapshot files** at `<project>/.omo/background-tasks/{taskId}.json` — no shared file to fight over. Atomic writes (`writeFileAtomically`), tolerant parsing, no prompt text persisted.
- **PID owner fencing** — each snapshot records `owner: {pid, startedAt}`. Recovery only adopts snapshots whose owner process is dead; live siblings are never touched. GC is owner-aware too (a live owner's snapshot survives regardless of age).
- **Eager lifecycle persistence** — snapshots are written at enqueue (pending, pre-session), session binding, resume, and synchronously after every terminal mutation (complete / cancel / interrupt / error / stale-timeout / fallback retry) *before* any await, so a crash window cannot strand a `running` snapshot.
- **Recovery at plugin init** (`restorePersistedTasks`, fire-and-forget from `createManagers`): orphaned terminal snapshots are loaded into the completed-task archive + registry so `background_output` works post-restart with zero tool changes. Orphaned non-terminal snapshots are reconciled against `session.get` (scoped by project directory): session gone → `error("session lost across restart")`; session alive → `interrupt` with guidance pointing at `session_read(...)` / `task(task_id="ses_...")`. Recovery **never injects prompts** (prompt-async-gate audit stays green) and never starts pollers or notifications for recovered tasks.
- **Config**: `background_task.persistence` (default `true`); `false` → zero filesystem footprint.

## Known limitation

PID-reuse fencing is liveness-only today; `owner.startedAt` is recorded so a future refinement can compare process start times. The failure direction is conservative: a falsely-"alive" owner only skips adoption, never corrupts another instance's task.

## Testing

[...]

Issue #2915, opened 2026-03-28
[Bug]: background_output returns "No new output since last check" after /undo — read cursor not reset


raindropQwQ:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

After using `/undo` to remove the main agent's response that contained a `background_output` call, re-triggering the same `background_output` with the correct `task_id` returns `"(No new output since last check)"` instead of the actual result. The sub-agent completed successfully and the data exists — `background_output` with `full_session=true` retrieves it correctly.
The root cause appears to be that `/undo` removes the conversation message containing the `background_output` result, but **does** not reset the internal read cursor (likely `since_message_id` or equivalent). The system still considers the result "already consumed."

### Steps to Reproduce

1. Fire a background task: `task(subagent_type="explore", run_in_background=true, ...)`
2. Wait for the system reminder notification
3. Main agent calls `background_output(task_id="bg_xxx")` — result is returned ✅
4. User executes `/undo` — removes the main agent's response
5. User re-sends the system reminder with the same task_id
6. Main agent calls `background_output(task_id="bg_xxx")` again → `"(No new output since last check)"` ❌
7. Main agent concludes the sub-agent "lost" its output
8. Calling `background_output(task_id="bg_xxx", full_session=true)` after step 6 does return the full sub-agent session with all messages intact. The data is not lost — only the default (incremental) read mode fails.

### Expected Behavior

`background_output` should return the sub-agent's result regardless of whether a previous read was undone. Either:
1. `/undo` should reset the read cursor for any `background_output` calls in the undone message, or
2. The default read mode should not rely on a persistent cursor that survives `/undo`

### Actual Behavior

Default `background_output` returns `"(No new output since last check)".` The main agent reports the sub-agent as lost/failed.

### Doctor Output

```shell
bunx oh-my-opencode doctor

 oMoMoMoMo Doctor 

 ⚠ 4 issues found:

1. Using legacy package name
[...]

PR #6217, opened 2026-07-19
fix(background-task): recovery hint + non-destructive truncation (#1734 Improvement 1)

### LINKED ISSUE #1734: [Feature]: Background Task Output Distillation + Non-Destructive Recovery
# Feature Request: Background Task Output Distillation

### Pre-submission checklist

- [x] I have searched existing issues and feature requests for duplicates
- [x] I have read the [README](https://github.com/code-yeongyu/oh-my-opencode#readme)

### Feature description

When a background task completes and the parent agent calls `background_output`, `formatTaskResult()` fetches all messages from the child session and concatenates every text, reasoning, and tool_result block into a single blob with no size cap. This raw output is injected directly into the parent session's context.

With parallel background agents (the intended use pattern), this causes rapid context window exhaustion. In a real session I debugged, 4 parallel `explore` agents returned ~60K tokens of combined output in a single batch, contributing to a 260K token spike that hit the 200K provider limit.

### The problem in code

`formatTaskResult()` (dist/index.js, ~line 48776):
- Calls `client.session.messages()` for the child session
- Filters for assistant + tool messages
- Extracts ALL text/reasoning/tool_result parts
- Joins with `\n\n` — no truncation, no summarization, no cap

`background_output` is not in the `TRUNCATABLE_TOOLS` allowlist, so the `tool-output-truncator` hook skips it by default.

### Existing partial workarounds

**1. `experimental.truncate_all_tool_outputs: true`** — caps each tool output at 50K tokens. Helps but doesn't prevent the blowup: 4 parallel background agents × 50K = 200K tokens before system prompt and history are counted. Still exceeds a 200K limit.

**2. DCP recovery (when it triggers)** — after a 400 error, the recovery hook finds tool results on disk sorted by size and replaces each with a truncation notice, then auto-retries. This recovers the session, but the original tool outputs are **replaced on disk** and lost. The agent has to redo all background research from scratch.

### Proposed solution

There are two complementary improvements, ranked by feasibility:

#### Improvement 1: Non-destructive recovery (minimal change, plugin-layer only)

Currently `truncateToolResult()` overwrites original tool output on disk with a truncation notice — the data is destroyed. A minimal fix: write the original content to a file *before* overwriting, and include the file path in the replacement message:

```javascript
// Current (destructive):
part.state.output = TRUNCATION_MESSAGE;
writeFileSync(partPath, JSON.stringify(part));

// Proposed (non-destructive):
const preservePath = join(storageDir, sessionID, `${partId}_output.txt`);
writeFileSync(preservePath, part.state.output);  // preserve first
part.state.output = `[TOOL OUTPUT PRESERVED: ${preservePath}]\nUse read tool to access full output if needed.`;
writeFileSync(partPath, JSON.stringify(part));
```

The agent can then `read` the preserved file on demand without it consuming context. The plugin already uses `writeFileSync` — this is one additional write call. No Go engine changes needed.

This also enables a **multi-step recovery pipeline** (replacing the current binary truncate-or-summarize):

1. **Deduplicate** — existing DCP logic
2. **Offload large outputs to files** — preserve content, replace with file reference (not destructive)
3. **Retroactive offload** — if still over limit, iterate backward through older turns, offload their large tool outputs too
4. **Full compaction** — `session.summarize` as absolute last resort

#### Improvement 2: Proactive distillation at `formatTaskResult` time

Instead of reacting after the context blows up, distill background task output *before* injection into the parent context.

**Important constraint identified via Oracle consultation:** the plugin recovery layer cannot make async LLM calls (only `session.summarize` is available, and it's all-or-nothing for the entire session). So distillation must happen at `formatTaskResult` time, not during recovery.

**Proposed flow:**

```
Background agent completes
  → Full output stays in session storage (already the case, no change needed)
  → At formatTaskResult time: spawn a cheap distillation call (flash model)
  → Distillation produces: key findings + decisions + file changes summary
  → Parent receives:
      - Distilled summary (~2-5K tokens)
      - session_id reference for full output
      - Instruction: "use session_read(session_id=ses_xxx) for full details"
  → Full output accessible on-demand via session_read (no data loss)
```

**Possible config shape:**

```jsonc
{
  "background_task": {
    "distill_output": {
      "enabled": true,
      "model": "github-copilot/grok-code-fast-1",
      "max_summary_tokens": 3000
    }
  }
}
```

### Why not just truncation?

| Approach | Pros | Cons |
|----------|------|------|
| `truncate_all_tool_outputs` | Works today, simple | Loses signal — truncation is position-based, not relevance-based |
| Hard cap on `formatTaskResult` | Simple to implement | Same problem — arbitrary cutoff |
| Distillation subagent | Preserves all important signal, reference to full output | Adds a small latency + cheap model call per task |

Distillation retains what matters and discards what doesn't, rather than blindly chopping at a byte boundary.

### Use case

Any workflow with parallel background agents — which is the primary oh-my-opencode orchestration pattern. The more agents you run in parallel, the faster you exhaust context. This is especially acute with Antigravity models that have a 200K token limit.

### Related issues

- [oh-my-opencode#96](https://github.com/code-yeongyu/oh-my-opencode/issues/96) — DCP (reactive recovery, doesn't prevent the blowup)
- [opencode-antigravity-auth#411](https://github.com/NoeFabris/opencode-antigravity-auth/issues/411) — `antigravity-claude-opus-4-6-thinking exceeds 200k limit`
- [opencode-antigravity-auth#407](https://github.com/NoeFabris/opencode-antigravity-auth/issues/407) — `Do compaction earlier`

### Alternatives considered

1. **Add `background_output` to `TRUNCATABLE_TOOLS`** — simple but same truncation problem
2. **Reduce `DEFAULT_MAX_TOKENS` globally** — hurts legitimate large tool outputs
3. **Let the parent agent request `full_session=true` with `message_limit`/`include_tool_results=false`** — shifts the burden to prompt engineering; agents don't consistently use these params
4. **Configure compaction agent with a larger context model** — tried setting `agents.compaction.model` to `google/antigravity-gemini-3-pro` (1M context). The model does get used for compaction when it triggers, but it doesn't prevent the "prompt is too long" error because the issue isn't compaction capacity — it's that compaction doesn't trigger early enough, and background output injection happens upstream of compaction entirely

### Additional context

Session forensics from a real failure: 260K tokens in under 5 minutes. Breakdown:
- 4 parallel background agents: ~60K tokens (raw `formatTaskResult` output)
- 8 sequential file reads: ~80K tokens
- System prompt + instructions: ~30K tokens
- Conversation history: ~90K tokens

EZotoff:
## Summary

Implements **Improvement 1 (non-destructive recovery)** from [the maintainer triage](https://github.com/code-yeongyu/oh-my-openagent/issues/1734), plus the **recovery-hint proposal** from @R-omk in the same issue.

## Problem

When background task results are consumed via `background_output`, two data-loss paths exist:

1. **Incremental cursor dead-end**: After first consumption, the cursor advances. Re-calling `background_output` returns `"(No new output since last check)"` with **no guidance** on how to recover. Agents interpret this as "no results" and continue without the subagent work. (Reported by @R-omk)

2. **Destructive truncation**: When `truncateToolResult()` fires to recover context window space, it overwrites `part.state.output` with `TRUNCATION_MESSAGE` — the original output is **permanently destroyed** with no backup.

### 1. Recovery hint (`task-result-format.ts`)

When the incremental cursor returns no new messages, append recovery instructions:
```
(No new output since last check)

To retrieve the complete output, re-call this tool with:
- background_output(task_id="bg_xxx", full_session=true) — re-fetches all messages, bypassing the incremental cursor
- Or use session_read(session_id="ses_xxx") to read the full child session transcript
```

### 2. Non-destructive truncation (`tool-result-storage.ts`)

Before overwriting `part.state.output`, write the original to a backup file `{partPath}.original`. Adds `recoverTruncatedOutput(partPath)` to read it back. Best-effort: if the backup write fails, truncation still proceeds.

### 3. Test (`create-background-output.undo.test.ts`)

New test verifying the recovery hint includes the correct `task_id` and `session_id`.

## Evidence

Observed in production (2026-07-16, glm-5.2 orchestrator):
- Atlas orchestrator collected 6 parallel `background_output` results (788KB total)
- OpenCode `compaction.prune` immediately ejected all 6 results (budget-based pruning)
- Re-collection returned "No new output since last check" — the agent said *"Background outputs keep getting cleared. Pivoting to direct verification"* and abandoned all subagent results
- The incremental cursor (#2915) prevented re-fetch, and no hint guided the agent to use `full_session=true`

## Scope

- **Does not include** Improvement 2 (proactive distillation) — requires the async LLM summarizer design discussion
- **Compatible** with PR #5458 and PR #5204 — this PR adds recovery for when those protections are not yet active

Closes #1734 (Improvement 1)

---
## Summary by cubic
Adds recovery instructions to `background_output` when the incremental cursor has no new messages, so agents know to refetch the full session instead of treating it as "no results" (aligns with #1734 Improvement 1).

- **Bug Fixes**
  - Show recovery hint with `background_output(task_id=..., full_session=true)` and `session_read(session_id=...)` when no new output.
  - Add test verifying the hint includes both `task_id` and `session_id`.

<sup>Written for commit 07281fb7643397c9e2ce78fce895c53b8d64856b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


EZotoff:
Friendly nudge for review — this implements Improvement 1 (non-destructive recovery) from #1734, which was marked **accepted** and **help wanted**. All CI is green across ubuntu/macos/windows (tests, typecheck, build, compatibility suites). The three touched files haven't changed on `dev` since the PR was opened, so it should apply cleanly.

Happy to make adjustments if the implementation approach needs refinement — particularly around the backup file naming (`{partPath}.original`) or whether recovery should be exposed differently than the `recoverTruncatedOutput()` function.

MoerAI (changes requested):
Thanks for the PR, @EZotoff, and for taking the time on this.

The recovery hint in `packages/omo-opencode/src/tools/background-task/task-result-format.ts` is a clear improvement on its own, and it has a test. The non-destructive backup half in `packages/omo-opencode/src/hooks/anthropic-context-window-limit-recovery/tool-result-storage.ts` is what I cannot take as written. Three specific things:

1. `recoverTruncatedOutput()` is exported but never called. I searched the tree and there is no consumer, so as it stands it is dead code. If a caller is planned, it would be better to land it together so the function is exercised.

2. The `.original` backup has no cleanup path anywhere in the repository. `truncateToolResult()` writes the full original output to `${partPath}.original` on every truncation and nothing ever removes it. Truncation triggers precisely on the largest outputs, so a feature whose purpose is to relieve context pressure would end up writing every large payload to disk permanently. A retention or cleanup story, or tying the backup to the session lifecycle, would resolve this.

3. The backup path has no test. The added test in `create-background-output.undo.test.ts` covers only the recovery hint. A test that truncates, then recovers, would pin the behavior that matters here.

Since the hint and the backup are independent concerns, the easiest path forward is probably to split them: the hint change looks ready to go on its own, and the backup can land once it has a consumer, a cleanup story, and a test. Happy to review either or both once that is sorted, and thanks again for working on #1734.

code-yeongyu:
[sisyphus-bot] REVIEW-CONCERNS: The recovery hint (task-result-format.ts) is clean and ready. The non-destructive truncation half has three unresolved issues from @MoerAI's review: (1) `recoverTruncatedOutput()` has no caller — dead code; (2) `.original` backup files accumulate permanently with no cleanup path, and truncation targets the largest outputs, so this writes more to disk than it frees from context; (3) no test for the backup/recover path. Recommend splitting: land the hint now, rework the backup with a consumer, lifecycle cleanup, and test coverage separately.

EZotoff:
Split this PR per MoerAI review. This branch now contains only the recovery hint (task-result-format.ts + test), which MoerAI said is ready. The non-destructive truncation will ship as a separate PR with the consumer, cleanup, and test that were requested.

MoerAI (approved):
Re-reviewed after the split. The dead export and backup/truncation path were removed; the current diff is limited to the recovery hint and its regression test, which is the independently mergeable slice requested in my prior review. CI is green and my concerns are resolved.

EZotoff:
@code-yeongyu Friendly ping — this PR was approved by @MoerAI on Aug 8 after the split, and CI is green across ubuntu/macos/windows. It implements Improvement 1 (recovery hint) from #1734 (accepted, help wanted); the second half (non-destructive truncation rework) now lives in #6633. Would appreciate a merge when you get a chance — happy to address anything else if needed.

MoerAI (changes requested):
Thanks for splitting out the independently mergeable recovery hint and removing the dead backup/recovery path. The current two-file implementation remains the right narrow scope.

Please update onto current `dev`; this head is 3,483 commits behind and its historical check names no longer satisfy the protected matrix:

Preserve the hint-only behavior and its stable machine-consumed command tokens: `background_output`, the task id, `full_session=true`, `session_read`, and the session id.

Run isolated real OpenCode QA on the rebased artifact: complete a background task, consume incremental output twice, observe the recovery hint, then prove both `full_session=true` and `session_read` recover the transcript. Commit the commands, observed output, and unchanged host-session proof under `.omo/evidence/`.

After current CI and that evidence are green, request owner re-review. No non-destructive backup machinery should return in this PR.

EZotoff:
Investigating the requested rebase onto current `dev` turned up something better: **this PR's content is already in `dev` via #6633's merge lineage.**

The #6633 branch was cut from the pre-split combined state, so its merge (Sep 2) carried the recovery-hint commit along with the truncation rework. Verified on current `dev`:

- `packages/omo-opencode/src/tools/background-task/task-result-format.ts:99-100` contains the recovery hint with the exact machine-consumed tokens (`background_output(task_id=..., full_session=true)` and `session_read(session_id=...)`)
- `git diff dev..fix/background-output-recovery-hint -- task-result-format.ts create-background-output.undo.test.ts` → **empty** for both files
- `tool-result-storage.ts` machinery is present in `dev` as landed via #6633

So the rebase is a no-op and no backup machinery would return — there is nothing left to merge here. Closing as superseded by #6633. Thanks @MoerAI for the careful reviews through the split.

PR #6229, opened 2026-07-20
fix(background-task): suppress consumed completion parent wakes

acamq:
## What changed

Background tasks whose completed output was already consumed into the parent turn (via `background_output` or supervised inline return) no longer queue a redundant completion wake. Suppression applies **only** to successful completed tasks; failure/cancel/interrupt wakes always fire.

## Why

Before this fix, the race order `task completes → wake queued → parent consumes output in same turn → flush` produced a redundant `[BACKGROUND TASK COMPLETED]` / `[ALL BACKGROUND TASKS COMPLETE]` wake for output the parent had already seen. The wake was noise — it re-prompted the parent about output it had just consumed.

## How (two-layer guard)

1. **Generation-time filter** in `notifyParentSession` (`manager.ts`): when building a grouped all-complete notification, consumed successes are filtered out of `completedTasks`. If only consumed successes remain and there are no failures, the wake is skipped entirely. Partial/progress wakes for consumed single successes are also skipped.
2. **Flush-time race guard** in `flushPendingParentWake` (`manager.ts`): a queued wake is re-checked at flush time — if every task it references is now consumed AND it is not a failure wake, the wake is dropped before dispatch. Handles the queue-before-consume race the generation filter can't see.

Both layers are gated on `status === "completed"` + `!isFailureParentWake(...)` so failures are **never** suppressed.

Task identity is tracked via new helpers in `background-output-consumption.ts`:
- `recordBackgroundTaskOutputConsumption({ parentSessionID, parentMessageID, taskID, taskSessionID })`
- `isBackgroundTaskOutputConsumption({ parentSessionID, taskID?, taskSessionID? })`

The legacy `recordBackgroundOutputConsumption(parentSessionID, parentMessageID, taskSessionID)` now also records identity, so both maps stay consistent regardless of which helper a caller/test uses. Cleanup is symmetric across restore / clear-by-parent / clear-by-task-session / clear-all.

## Observed behavior (verified)

- Consumed successful completion → **0** redundant parent-wake dispatches (was 1).
- Unconsumed successful completion → **1** wake (unchanged).
- Failed / cancelled / interrupted task → **1** wake each (unchanged — failures still wake even if inspected).
- Grouped all-complete with mixed consumed-success + failure → failure reported, consumed success filtered out.
- Supervised inline (`run_in_background=false`) completed return → consumption recorded, redundant wake suppressed.

## QA / evidence

- 1218 focused unit/integration tests pass (`bun test` across background-agent + tools + shared).
- `bun run typecheck:packages` exit 0 (all 27 workspace packages).
- 10/10 changed TypeScript files report zero LSP diagnostics.
- Controlled live plugin exercise captures the unique `[background-agent] Suppressed consumed-success parent wake at flush time` log line (emitted by exactly one site: `manager.ts flushPendingParentWake`), with isolation proven (host opencode DB session count unchanged 6172→6172).
- Full evidence: `.omo/evidence/20260717-background-task-notification-consumption/README.md` + `f1`-`f4` reviewer files (local `.omo/`, not shipped in the PR).

**Residual gap (honestly disclosed):** a full end-to-end `opencode serve` drive was attempted but hung (fake-LLM child-idle shape mismatch in opencode 1.17.20); the controlled live plugin exercise substitutes. See F3 evidence for details.

## Files

- 6 production files (+250/-11): `background-output-consumption.ts`, `background-task-notification-template.ts`, `parent-wake-notifier.ts`, `manager.ts`, `create-background-output.ts`, `unstable-agent-task.ts`.
- 4 new test files (~1001 lines): `task-consumed-wake-suppression.test.ts`, `background-output-consumption.task-identity.test.ts`, `create-background-output.consumption.test.ts`, `unstable-agent-task.consumption.test.ts`.

## Risk

Low. Suppression is narrowly scoped to `status === "completed"` successes; failures are doubly protected. No changes to notification wording, hook signatures, tool surfaces, or the SSE event shape. Change is entirely contained within `packages/omo-opencode/` — no Codex/adapter/sibling-package impact.

---
## Summary by cubic
Suppress redundant parent wakes when a background task’s completed output was already consumed in the same parent turn. Fix coalescing/flush races so only unconsumed successes notify; failure/cancel/interrupt still always notify.

- **Bug Fixes**
  - Filter consumed successes out of grouped notifications; skip wake if only consumed successes remain. Include `sessionId` in notification task payloads.
  - Add flush-time suppression: store task identities on each pending wake and union them across coalesced wakes; drop non-failure wakes only if every referenced task is consumed; support deleting pending wakes and route scheduled flushes through the manager.
  - Track consumption by task ID and session via `recordBackgroundTaskOutputConsumption`/`isBackgroundTaskOutputConsumption`; legacy `recordBackgroundOutputConsumption` now records identity too. Record consumption for `background_output(..., full_session=true)` and supervised inline completions, and only when the response actually included completed output.

<sup>Written for commit 89ecbd2f5dfeeafba7d9f83958f08cefdedce155. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


acamq (commented):
**Blocking correctness issue:** the flush-time suppression bookkeeping is not kept in sync with the pending wake's coalescing semantics. `ParentWakePendingQueue.queueWake()` merges notifications for the same parent, but `pendingWakeTaskIdentitiesByParent.set(...)` replaces the identities with only the latest queued wake.

A concrete race:
1. Task A queues an unconsumed success notification.
2. Task B queues another success notification before the timer flushes; the pending wake now contains A and B, but the identity map contains only B.
3. The parent consumes B.
4. The flush guard sees every tracked identity consumed and deletes the entire pending wake, suppressing A even though A was never consumed.

Please associate identities with the actual merged pending wake (including replacement semantics when an all-complete notification supersedes progress notifications), or store the identities directly alongside each queued notification. Add a regression covering two coalesced successful tasks where only the later task is consumed before flush.

acamq:
@acamq Thanks for the catch — you're exactly right, the side-channel map's `.set()` was overwriting earlier-queued identities whenever a later wake coalesced into the same pending parent wake, so the flush guard would drop the entire coalesced wake whenever only the *latest* identity was tracked-and-consumed, suppressing earlier tasks that were never inspected.

Fixed in 83b9e22e1 using your option 2 (store the identities directly alongside each queued notification). Concretely:

- Added `taskIdentities?: TaskIdentity[]` as a flat de-duplicated union field directly on `PendingParentWake` (`parent-wake-dedupe.ts`); `cloneParentWake` deep-copies it.
- `ParentWakePendingQueue.queueWake` and `requeueWake` now accept an optional `taskIdentities` argument and **union** it into `pendingWake.taskIdentities` via a new `unionTaskIdentities` helper (keyed by `taskID + taskSessionID`), instead of replacing.
- `ParentWakeNotifier.queuePendingParentWake*` accept `taskIdentities` and forward.
- `BackgroundManager.notifyParentSession` passes the wake identities through `queuePendingParentWake` (no more side-channel map); `flushPendingParentWake` reads `pendingWake.taskIdentities`. The `pendingWakeTaskIdentitiesByParent` field is deleted entirely — the sync bug class is gone because the wake owns its identities and erases them when it is dropped/dispatched.

On the all-complete-supersedes-progress semantics you flagged: those are preserved naturally because an all-complete wake already builds its identities from `reportableCompletedTasks.map(...)` — a superset of any progress identities that preceded it — so the flat union of `[progress(A)] ∪ [all-complete(A,B)]` is exactly `[A,B]`. The flush guard predicate is unchanged (`identities.length > 0 && !isFailureParentWake(pendingWake) && identities.every(consumed)`), so the failure-protection guarantee is intact and the suppression is strictly narrower than before.

Regression test added in `task-consumed-wake-suppression.test.ts`: three tasks share a parent (A, B, C with C still running, so A and B both produce PROGRESS wakes that coalesce); only B is consumed before flush; the test asserts the wake still fires and names A. Verified RED against the old replace-semantics (`Received length: 0` — the exact bug shape you described) and GREEN with the union.

Full suite: 751 background-agent + 10 consumption tests pass; `bun run typecheck` clean across all 27 workspace packages; LSP clean on all 5 changed files. Evidence written to `.omo/evidence/f5-coalescing-race-fix.md` (local-only per repo convention). Notes on what was extended vs reused and the omitted end-to-end live drive (same fake-LLM/child-idle shape limitation as the original PR) are in that file.

acamq (commented):
Follow-up review on `83b9e22`: the original coalescing blocker is resolved.

`taskIdentities` now lives on `PendingParentWake`, `queueWake()` and `requeueWake()` union identities rather than replacing them, `cloneParentWake()` preserves them, and the flush guard reads the identities from the actual merged wake. The new A/B/C regression covers the exact race where only the later coalesced task is consumed and correctly verifies that the earlier unconsumed task still wakes the parent.

I found no additional blocking code-correctness issue in the fix.

CI note: the latest run is currently failing in `test (windows-latest)`; Ubuntu/macOS tests and the typecheck jobs pass. Please inspect or rerun the Windows failure before merging. Small robustness suggestion: retain `updateBackgroundTaskMarker(sessionID)` in a `finally` block, as the follow-up rewrite currently skips that refresh if the underlying flush throws.

acamq:
@codex review

acamq:
@codex review

code-yeongyu:
[sisyphus-bot] Well-implemented wake suppression for consumed background task outputs. Identity-on-wake design avoids the side-channel map pitfall. Comprehensive test coverage including coalescing edge cases. CI green. Ready for owner review.

Issue #4484, opened 2026-05-26
[Feature]:  Add stale-progress watchdog for stalled agent sessions


SoShymKing:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues and discussions to avoid duplicates
- [x] This feature request is specific to oh-my-opencode (not OpenCode core)
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Problem Description

OpenCode can sometimes keep an assistant/model turn alive without producing meaningful progress for a long time. This does not appear to be limited to background sub-agents. It can also happen in a main agent session or a normal conversation.

In my logs, I observed two related patterns:

1. **Main agent / normal session silent period**

   In the main session, OpenCode started a model stream and then produced no `message.part.updated` event for a long time before the stream eventually resumed.

   Example pattern:

   ```text
   service=session.processor session.id=<main-session> ... process
   service=llm providerID=openai modelID=gpt-5.5 session.id=<main-session> ... stream
   ...
   service=bus type=message.updated publishing
   # long gap
   service=bus type=message.part.updated publishing
   ```

   In one observed case, the gap between `message.updated` and the next `message.part.updated` was about 923 seconds. The session eventually resumed, so a short timeout would be too aggressive for main-agent turns.

2. **Background sub-agent stale state**

   In a background sub-agent session, the first assistant step completed normally with tool calls. The child OpenCode session export showed:

   - first assistant step finished with `finish: "tool-calls"`
   - tool parts such as `glob`, `grep`, and `glob` all had `state.status: "completed"`
   - the step ended with `step-finish`

   After that, OpenCode created another assistant message, but that message had:

   - `parts: []`
   - all token counts set to 0
   - eventual error: `MessageAbortedError: Aborted`

   During that period, OMO continued to treat the background task as still running because the child session status remained `busy`.

   Example OMO log pattern:

   ```text
   [background-agent] Session still running, relying on event-based progress:
   {"sessionStatus":"busy","toolCalls":9}
   ```

   The `toolCalls` count stayed fixed for the stale child task, while another background task continued making progress.

Eventually the stale child session was aborted, and OMO logged mixed lifecycle events such as:

```text
Session error - no retry
[...]

PR #6235, opened 2026-07-20
fix(omo-opencode): sync task poller stops trusting session.status()==busy forever

cyllas:
## Summary

`pollSyncSession` (the loop `task()` uses to wait synchronously for a sync subagent) resets its inactivity clock on every poll tick whenever `session.status()` reports `busy`/`retry`/`running` -- with no check that anything actually happened. When the underlying provider call gets stuck in-flight (no error, no new output, but the session status stays technically "busy"), the inactivity clock never accumulates and the existing 30-minute `poll_timeout` self-heal never fires. The subagent -- and the parent waiting on it -- hang indefinitely, with the log/status API insisting the whole time that everything is fine.

This PR makes the poller verify real progress (not just the status label) before trusting "busy" enough to reset the clock, so the existing timeout safety net actually gets to do its job.

## Changes

- **`sync-session-poller.ts`**: extracted `computeInactiveMs()` and added `computeProgressSignature()` (message count, last message id, part count, last text/reasoning length, last tool part state -- deliberately cheap, no full-text hashing). While `session.status()` is active, the poller now re-checks this fingerprint at most once per `BUSY_PROGRESS_RECHECK_MS` and only resets `inactiveStart` when it actually changed. If it never changes, the existing `poll_timeout` path fires as originally designed.
- **`timing.ts`**: new `BUSY_PROGRESS_RECHECK_MS` timing knob (default 30s), wired into `getTimingConfig()`/`__setTimingConfig()`/`__resetTimingConfig()` alongside the existing ones.
- **`sync-session-poller.test.ts`**: two new cases under "non-idle session status" pinning the exact bug and its regression guard (see QA below).

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/tools/delegate-task/sync-session-poller*.test.ts` (31 tests: 29 pre-existing + 2 new), plus `bun x tsgo --noEmit -p packages/omo-opencode/tsconfig.json`, plus the full `bun run build` (via the workspace `prepare` hook).
  - **Observed:** 31/31 pass, 0 typecheck errors, build succeeds.
  - **Why sufficient:** the two new tests deterministically reproduce (a) the exact hang -- status stays `busy` forever, message content frozen, `poll_timeout` still fires -- and (b) the regression guard -- status stays `busy` but message content genuinely grows every recheck, zero false aborts across 20+ polls that individually exceed `MAX_POLL_TIME_MS`. The pre-existing `"skips message check when session is not idle"` test is unaffected (its simulated run never crosses the 30s recheck window).
- **What was tested (manual, real harness):** built this exact change (`dist/index.js`) and loaded it as a `file://` plugin in a real OpenCode instance, inside an isolated XDG sandbox (own `XDG_DATA_HOME`/`XDG_CONFIG_HOME`/`XDG_CACHE_HOME`/`XDG_STATE_HOME` under a fresh `mktemp` dir; host `~/.config/opencode` and `~/.local/share/opencode` untouched).
  - **Observed:** plugin loaded cleanly (`init count=55`, no load errors); a real primary-agent turn completed against a real model; a real `task()` delegation to a subagent (`Sisyphus-Junior`) ran end-to-end through this exact modified poll loop -- child session created, read a file, finished cleanly, parent received the correct result. Zero errors, zero false timeouts.
  - **Why sufficient:** confirms the change doesn't regress the fast/happy path for a real delegation, not just in mocks.
  - **What was omitted:** reproducing the full 30-60 minute real-world hang end-to-end (rather than via the unit tests) was not attempted -- impractical to wait out, and the unit tests already pin the mechanism deterministically. This was originally caught live via two real stuck sessions (frozen 42 min and 65 min respectively, `session.status()` reporting "busy" the entire time) -- session IDs/logs from that repro are from an unrelated private codebase and are not included here.

## Risks & Residuals

- **False-positive risk (killing a legitimately slow session):** mitigated by the dedicated regression test -- only a message stream that is *actually* frozen (not just "busy") accumulates inactivity now.
- **Extra `session.messages()` calls while busy:** bounded by `BUSY_PROGRESS_RECHECK_MS` (default 30s) -- one extra fetch per recheck window, not per poll tick.
- **Scope:** this only covers the synchronous `task()` poller. The background-agent path and main-session path described in #4484 use different code and are not touched here -- left for a separate PR.

## Related Issues

Related to #4484 ("Add stale-progress watchdog for stalled agent sessions") -- that issue's own examples cover the background-agent and main-session cases, not this specific sync `task()` poller path, so this PR does not close it, only addresses the same class of bug in one additional, real, reproduced location.

---
## Summary by cubic
Fixes indefinite hangs in the sync `task()` poller by verifying real progress instead of trusting `session.status() == busy` forever. Stalled sessions now correctly hit the existing timeout; covers the sync `task()` path related to #4484.

- **Bug Fixes**
  - Check a cheap message fingerprint (count, last id, parts, last text/reasoning length, last tool state) before resetting inactivity when status is active.
  - Add `BUSY_PROGRESS_RECHECK_MS` (default 30s) in `timing.ts`, exposed via `getTimingConfig`/`__setTimingConfig`/`__resetTimingConfig`.
  - Add two tests to reproduce the busy-forever hang and to guard against false timeouts when progress is real.

- **Refactors**
  - Standardize comment punctuation in the poller and tests; no behavior change.

<sup>Written for commit 63fa4b147a31599bc968806640905d294904e8f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/tools/delegate-task/sync-session-poller.test.ts
+++ b/packages/omo-opencode/src/tools/delegate-task/sync-session-poller.test.ts
@@ -540,6 +540,111 @@ describe("pollSyncSession", () => {
     })
   })
 
+  describe("busy status does not hide real staleness", () => {
+    test("#given status stays busy forever with no message progress #when polling #then poll_timeout still fires", async () => {
+      // given
+      const { pollSyncSession } = require("./sync-session-poller")
+      __setTimingConfig({
+        POLL_INTERVAL_MS: 5,
+        MIN_STABILITY_TIME_MS: 0,
+        STABILITY_POLLS_REQUIRED: 1,
+        MAX_POLL_TIME_MS: 50,
+        BUSY_PROGRESS_RECHECK_MS: 5,
+      })
+      let abortCount = 0
+      const frozenMessages = {
+        data: [
+          { info: { id: "msg_001", role: "user", time: { created: 1000 } } },
+          {
+            info: { id: "msg_002", role: "assistant", time: { created: 2000 } },
+            parts: [{ type: "text", text: "still thinking" }],
+          },
+        ],
+      }
+      const mockClient = {
+        session: {
+          messages: async () => frozenMessages,
+          status: async () => ({ data: { ses_stuck: { type: "busy" } } }),
+          abort: async () => {
+            abortCount++
+          },
+        },
+      }
+
+      // when
+      const result = await pollSyncSession(createMockCtx(), mockClient, {
+        sessionID: "ses_stuck",
+        agentToUse: "test-agent",
+        toastManager: null,
+        taskId: undefined,
+      })
+
+      // then: a session.status() that never stops reporting "busy" must not
+      // be able to suppress the timeout forever when the message stream is
+      // frozen -- this is the real-world hang this fix targets.
+      expect(abortCount).toBe(1)
+      expect(result).toContain("Poll inactivity timeout reached")
+    })
+
+    test("#given status stays busy but message content genuinely grows #when polling #then the clock keeps resetting and no timeout fires", async () => {
+      // given
+      const { pollSyncSession } = require("./sync-session-poller")
+      __setTimingConfig({
+        POLL_INTERVAL_MS: 5,
+        MIN_STABILITY_TIME_MS: 0,
+        STABILITY_POLLS_REQUIRED: 1,
+        MAX_POLL_TIME_MS: 30,
+        BUSY_PROGRESS_RECHECK_MS: 1,
+      })
+      let abortCount = 0
+      let statusCallCount = 0
+      const mockClient = {
+        session: {
+          messages: async () => {
+            const growing = statusCallCount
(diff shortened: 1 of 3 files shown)


cyllas:
Closing in favor of #6236: same fix, rebased and rewritten cleanly from a fresh branch (this PR's history had leftover em-dash-style punctuation in code comments and commit messages that I wanted fixed without a force-push history rewrite on an already-open PR).

Issue #5345, opened 2026-06-16
[sisyphus-bot] [Bug]: Sync task hangs when subagent is not_in_status but dispatched parent wake remains pending


brilliantrough:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

A sync `task` subagent can hang for the full sync poll timeout when a background parent wake remains in `dispatched` state, even after the subagent is no longer active.

This appears related to #5172 and #2627, but the current failure mode is more specific:

- A sync `plan` subagent becomes `not_in_status`.
- Background task continuation state is already `idle`.
- OMO keeps logging `Kept dispatched parent wake awaiting late failure or assistant output`.
- The parent sync `task` poller continues waiting instead of returning, failing, or stale-cleaning the wake.
- The primary session stays stuck/spinning until the default 30-minute sync timeout.

### Steps to Reproduce

- Parent/orchestrator session: `ses_12e33f793ffeuPHG5hP7XN2rGA`
- Sync plan subagent session: `ses_12e32b765ffeo5LS5YsrmnwE6D`
- Background librarian task 1: `bg_c82b3ba0`, session `ses_12e32b751ffeAE7896PE7WoqA3`
- Background librarian task 2: `bg_0aaad9fe`, session `ses_12e32b73bffeuaA2qlQ8WiDdfb`

 
3. Start an OpenCode session with oh-my-openagent enabled.
4. Use an ultrawork/Sisyphus-style orchestrator flow.
5. Trigger a sync `task` subagent with `subagent_type: "plan"`.
6. During the same orchestration, launch background `librarian` tasks.
7. Let background tasks complete while the parent/orchestrator session is still active.
8. Observe the sync `plan` task remain stuck/spinning long after the subagent appears idle/done.

### Expected Behavior

Once the sync subagent is no longer active and background continuation state is idle, OMO should either:

- return the sync subagent result,
- mark the subagent stale/failed and return a clear error,
- or clear/requeue the stale dispatched parent wake after a bounded timeout.

[...]

PR #6236, opened 2026-07-20
fix(omo-opencode): sync task poller stops trusting session.status()==busy forever

cyllas:
## Summary

`pollSyncSession` (the loop `task()` uses to wait synchronously for a sync subagent) resets its inactivity clock on every poll tick whenever `session.status()` reports `busy`/`retry`/`running`, with no check that anything actually happened. When the underlying provider call gets stuck in-flight (no error, no new output, but the session status stays technically "busy"), the inactivity clock never accumulates and the existing 30-minute `poll_timeout` self-heal never fires. The subagent, and the parent waiting on it, hang indefinitely, with the log/status API insisting the whole time that everything is fine.

This PR makes the poller verify real progress (not just the status label) before trusting "busy" enough to reset the clock, so the existing timeout safety net actually gets to do its job.

## Changes

- **`sync-session-poller.ts`**: added `computeProgressSignature()` (message count, last message id, part count, last text/reasoning length, last tool part state; deliberately cheap, no full-text hashing). While `session.status()` is active, the poller now re-checks this fingerprint at most once per `BUSY_PROGRESS_RECHECK_MS` and only resets `inactiveStart` when it actually changed. If it never changes, the existing `poll_timeout` path fires as originally designed.
- **`timing.ts`**: new `BUSY_PROGRESS_RECHECK_MS` timing knob (default 30s), wired into `getTimingConfig()`/`__setTimingConfig()`/`__resetTimingConfig()` alongside the existing ones.
- **`sync-session-poller.test.ts`**: two new cases under "non-idle session status" pinning the exact bug and its regression guard (see QA below).

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/tools/delegate-task/sync-session-poller*.test.ts` (31 tests: 29 pre-existing + 2 new), plus `bun x tsgo --noEmit -p packages/omo-opencode/tsconfig.json`, plus the full `bun run build` (via the workspace `prepare` hook).
  - **Observed:** 31/31 pass, 0 typecheck errors, build succeeds.
  - **Why sufficient:** the two new tests deterministically reproduce (a) the exact hang: status stays `busy` forever, message content frozen, `poll_timeout` still fires; and (b) the regression guard: status stays `busy` but message content genuinely grows every recheck, zero false aborts across 20+ polls that individually exceed `MAX_POLL_TIME_MS`. The pre-existing `"skips message check when session is not idle"` test is unaffected (its simulated run never crosses the 30s recheck window).
- **What was tested (manual, real harness):** built this exact change (`dist/index.js`) and loaded it as a `file://` plugin in a real OpenCode instance, inside an isolated XDG sandbox (own `XDG_DATA_HOME`/`XDG_CONFIG_HOME`/`XDG_CACHE_HOME`/`XDG_STATE_HOME` under a fresh `mktemp` dir; host `~/.config/opencode` and `~/.local/share/opencode` untouched).
  - **Observed:** plugin loaded cleanly (`init count=55`, no load errors); a real primary-agent turn completed against a real model; a real `task()` delegation to a subagent (`Sisyphus-Junior`) ran end-to-end through this exact modified poll loop: child session created, read a file, finished cleanly, parent received the correct result. Zero errors, zero false timeouts.
  - **Why sufficient:** confirms the change doesn't regress the fast/happy path for a real delegation, not just in mocks.
  - **What was omitted:** reproducing the full 30-60 minute real-world hang end-to-end (rather than via the unit tests) was not attempted, since it is impractical to wait out and the unit tests already pin the mechanism deterministically. This was originally caught live via two real stuck sessions (frozen 42 min and 65 min respectively, `session.status()` reporting "busy" the entire time); session IDs/logs from that repro are from an unrelated private codebase and are not included here.

## Risks & Residuals

- **False-positive risk (killing a legitimately slow session):** mitigated by the dedicated regression test, so only a message stream that is *actually* frozen (not just "busy") accumulates inactivity now.
- **Extra `session.messages()` calls while busy:** bounded by `BUSY_PROGRESS_RECHECK_MS` (default 30s): one extra fetch per recheck window, not per poll tick.
- **Scope:** this only covers the synchronous `task()` poller. The background-agent path and main-session path described in #4484 use different code and are not touched here; left for a separate PR.

## Related Issues

- Related to #4484 ("Add stale-progress watchdog for stalled agent sessions"): that issue's own examples cover the background-agent and main-session cases, not this specific sync `task()` poller path, so this PR does not close it, only addresses the same class of bug in one additional, real, reproduced location.
- Related to #5345 ("Sync task hangs when subagent is not_in_status but dispatched parent wake remains pending"): same symptom class (sync `task()` poller hang) but a different root cause; that issue is about a stale `dispatchedParentWake` never clearing while the session is `not_in_status`, which is a separate code path from the busy-status progress check this PR adds. Not addressed here.

---
## Summary by cubic
Fixes a hang in the sync task poller where `session.status()` staying “busy” forever prevented the inactivity timeout from firing. The poller now requires real message progress before resetting the inactivity clock, so stuck sessions time out instead of looping.

- **Bug Fixes**
  - Check for real progress while status is active using a cheap message “progress fingerprint”; only reset inactivity when it changes.
  - Throttle progress rechecks to once per `BUSY_PROGRESS_RECHECK_MS` (default 30s) to limit extra `messages()` calls.
  - Added `BUSY_PROGRESS_RECHECK_MS` to timing config.
  - Tests cover both cases: stuck-busy correctly times out; progressing-busy does not falsely time out.

<sup>Written for commit 2ca685487940bf9dd017d5ea4c2a9c2ea0100450. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/tools/delegate-task/sync-session-poller.test.ts
+++ b/packages/omo-opencode/src/tools/delegate-task/sync-session-poller.test.ts
@@ -540,6 +540,111 @@ describe("pollSyncSession", () => {
     })
   })
 
+  describe("busy status does not hide real staleness", () => {
+    test("#given status stays busy forever with no message progress #when polling #then poll_timeout still fires", async () => {
+      // given
+      const { pollSyncSession } = require("./sync-session-poller")
+      __setTimingConfig({
+        POLL_INTERVAL_MS: 5,
+        MIN_STABILITY_TIME_MS: 0,
+        STABILITY_POLLS_REQUIRED: 1,
+        MAX_POLL_TIME_MS: 50,
+        BUSY_PROGRESS_RECHECK_MS: 5,
+      })
+      let abortCount = 0
+      const frozenMessages = {
+        data: [
+          { info: { id: "msg_001", role: "user", time: { created: 1000 } } },
+          {
+            info: { id: "msg_002", role: "assistant", time: { created: 2000 } },
+            parts: [{ type: "text", text: "still thinking" }],
+          },
+        ],
+      }
+      const mockClient = {
+        session: {
+          messages: async () => frozenMessages,
+          status: async () => ({ data: { ses_stuck: { type: "busy" } } }),
+          abort: async () => {
+            abortCount++
+          },
+        },
+      }
+
+      // when
+      const result = await pollSyncSession(createMockCtx(), mockClient, {
+        sessionID: "ses_stuck",
+        agentToUse: "test-agent",
+        toastManager: null,
+        taskId: undefined,
+      })
+
+      // then: a session.status() that never stops reporting "busy" must not
+      // be able to suppress the timeout forever when the message stream is
+      // frozen; this is the real-world hang this fix targets.
+      expect(abortCount).toBe(1)
+      expect(result).toContain("Poll inactivity timeout reached")
+    })
+
+    test("#given status stays busy but message content genuinely grows #when polling #then the clock keeps resetting and no timeout fires", async () => {
+      // given
+      const { pollSyncSession } = require("./sync-session-poller")
+      __setTimingConfig({
+        POLL_INTERVAL_MS: 5,
+        MIN_STABILITY_TIME_MS: 0,
+        STABILITY_POLLS_REQUIRED: 1,
+        MAX_POLL_TIME_MS: 30,
+        BUSY_PROGRESS_RECHECK_MS: 1,
+      })
+      let abortCount = 0
+      let statusCallCount = 0
+      const mockClient = {
+        session: {
+          messages: async () => {
+            const growing = statusCallCount
+            return {
+              data: [
+                { info: { id: "msg_001", role: "user", time: { created: 1000 } } },
+                {
+                  info: {
+                    id: "msg_002",
+                    role: "assistant",
+                    time: { created: 2000 },
+                    ...(growing >= 20 ? { finish: "end_turn" } : {}),
+                  },
+                  parts: [{ type: "text", text: "x".repeat(growing + 1) }],
+                },
+              ],
+            }
+          },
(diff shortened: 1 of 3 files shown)


code-yeongyu:
[sisyphus-bot] Good fix for a real hang scenario. Progress-signature approach is sound — cheap fingerprint avoids hashing full text. Throttled recheck prevents message-spam. Needs CI test/typecheck runs before merge.

MoerAI (changes requested):
Thanks for preserving #6236 as the successor to #6235 and for isolating the busy-forever failure.

Please rebase onto current `dev` and resolve the `sync-session-poller.ts` conflict without dropping bounded message reads, anchor-ID handling, or non-active fetch gating.

Broaden the progress signature. The current implementation observes only the final part's text length or tool status, so a non-final tool-state update or an equal-length text replacement is real progress that can still be misclassified as stale and aborted.

Add deterministic coverage for both cases, then replace the 1-5 ms intervals and wall-clock deadlines with injected clock/wait control.

Finally, run the current required CI matrix and revalidate real synchronous `task()` delegation on the rebased build. The current matrix was cancelled:
https://github.com/code-yeongyu/oh-my-openagent/actions/runs/29754066062

PR #6241, opened 2026-07-21
fix(omo-senpi): eager codegraph MCP lifecycle to ride senpi startup race

code-yeongyu:
## What
Register the codegraph MCP server with explicit lifecycle: "eager" in the omo-senpi adapter. With the default lazy lifecycle, an uncached codegraph server (indexing a huge cwd like ~/local-workspaces) connected inline during senpi's MCP attach, blocking the first render/turn for 15-30s with -32001 timeouts. Eager routes it through senpi's bounded 250ms startup race deliberately and keeps it warm. Includes the regenerated plugin bundle (plugin/extensions/omo.js).

## Measured evidence (cross-fix live QA, real tmux senpi in ~/local-workspaces, omo plugin loaded)
- First footer render: 2.12s (baseline >=12s blank)
- Prompt typed at T+2s answered in 0.29s, no attach-induced stall
- senpi-side race for lazy servers: code-yeongyu/senpi#260 (landed); see also code-yeongyu/senpi#259

## Evidence
senpi-render-hotpath-and-startup-fixes/t6-codegraph-eager.txt (bun test 8/8 + tsgo green, RED->GREEN), f2-live-qa.txt (live QA PASS)

Plan: .omo/plans/senpi-render-hotpath-and-startup-fixes.md


code-yeongyu:
Follow-up commit f2b6aff76 regenerates the tracked plugin bundle (`packages/omo-senpi/plugin/extensions/omo.js`) so the pushed artifact actually contains the `lifecycle: "eager"` codegraph change from fcb4fb422. Verified with `build-extension.mjs --check` (bundle current with src) and confirmed the remote bundle now includes `lifecycle:"eager"`.

PR #6260, opened 2026-07-21
fix(background-output): distill completed task results

### LINKED ISSUE #1734: [Feature]: Background Task Output Distillation + Non-Destructive Recovery

tracycam:
## Summary

- return text from the last successful assistant answer produced after the task cursor
- exclude reasoning and earlier transcript noise from completed task output
- preserve cursor semantics and session-error priority
- use the most recent tool-result text only as a fallback
- include isolated OpenCode QA and plugin-load evidence

## Validation

- focused background-output regressions: 3 passed
- workspace typecheck: passed
- local built-plugin load smoke: passed
- evidence: `.omo/evidence/20260721-task-result-distillation-dist/`

The evidence summary records the bounded wake-harness limitation separately; the focused behavior and real plugin load both pass.

Fixes #1734

---
## Summary by cubic
Distills completed background task output to return only the final assistant answer, with a safe fallback to the last tool result when needed. Preserves cursor semantics and error priority to remove noisy or misleading text. Fixes #1734.

- **Bug Fixes**
  - Return the last assistant text after the task cursor; exclude reasoning and earlier transcript.
  - Fall back to the most recent `tool_result` text; strip reasoning blocks.
  - Surface session errors over success text and respect the message cursor.
  - Fixed the wake-split QA probe so it reaches a conclusive verdict: bounded curl timeouts and offline npm for the sandbox serve.
  - Added tests in `packages/omo-opencode/src/tools/background-task/task-result-format.test.ts`; typecheck and plugin-load smoke pass; evidence in `.omo/evidence/20260721-task-result-distillation-dist/` and `.omo/evidence/20260826-wake-split-probe-fixed/`.

<sup>Written for commit 9fbf14b81b33f3508d14ea9ef2016573bb143b2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
[sisyphus-bot] Draft PR implementing output distillation for #1734. Core logic looks correct (final-answer extraction + reasoning stripping + tool fallback). Needs: (1) remove draft status, (2) full CI run, (3) resolve wake-split probe inconclusiveness before merge.

tracycam:
Wake-split probe is now conclusive — draft removed and branch updated (191d14031..fa36a5950).

**Verdict: FIXED** (`--expect fixed` exits 0):

```
RESULT=FIXED parent_assistant_messages=3 parent_tool_call_turns=2 terminal_stops=1
child_task_sessions=1 plugin_inits=1 WAKE_DISPATCHED_DURING_PARENT_TURN=true
route_live_dispatch=true branch_counts=parent-tool-call:1,parent-hold:1,child:1,wake:0,default:1
```

Exactly one terminal stop and one child task session — no second runner forked; the wake that dispatched during the parent turn routed through the live listener. Real DB untouched (5530 before/after). Evidence: `.omo/evidence/20260826-wake-split-probe-fixed/` (qa-summary.md carries the full analysis).

The earlier inconclusiveness was two harness defects, both fixed on this branch:

1. Every curl in the probe and in `oqa_wait_http` ran without per-request timeouts — a single hung request defeated the outer polling deadline. All sites now carry `--connect-timeout`/`--max-time` (+ `--noproxy '*'` on localhost endpoints).
2. The real stall: serve-mode first-request plugin loading reifies the `$HOME/.opencode` plugin store, and its in-process registry fetch stalled indefinitely inside the offline QA sandbox (reproduced with and without proxy env, across file://src, file://dist, and path-package plugin specs — `/session` routes wedge before session creation). The sandbox serve now launches with `npm_config_offline=true` + `npm_config_prefer_offline=true` and stripped proxy env, so the store reify no-ops and the path plugin loads directly. Self-test remains 16/16 green.

The distillation changes themselves are untouched by these commits — harness and evidence only. CI should now be running on the updated head.

MoerAI (changes requested):
Thanks for the PR, @tracycam, and for the honest QA record.

`task-result-format.ts` binds answer extraction to `assistantMessages.at(-1)`. If the final assistant turn contains only a tool call or empty parts, it falls through to raw tool output and silently drops the earlier textual answer, which is the noise-loss regression this formatter is meant to prevent.

Please walk backward to the last assistant message that actually produces text, and add coverage for a tool-call-only final assistant turn plus empty final parts. Also use `Refs #1734` rather than closing it, since the issue's capped LLM-distillation scope remains broader than this change. Happy to re-review once those cases are covered.

MoerAI (changes requested):
Thanks for improving completed background-result selection and for adding an isolated formatter harness.

The current `assistantMessages.at(-1)` selection loses an earlier textual answer when the final assistant turn is tool-only or has empty parts, then emits raw tool fallback instead:
https://github.com/code-yeongyu/oh-my-openagent/blob/fa36a5950f5e002c115f1c10d8361a5c408e44c6/packages/omo-opencode/src/tools/background-task/task-result-format.ts#L105-L108

Scan assistant messages backward and select the latest one with non-empty text. Fall back to tool-result text only when no assistant message contains text.

Add deterministic regressions for a tool-only final assistant turn, empty final parts, and the supported `tool_result.output` form.

Change `Fixes #1734` to `Refs #1734`; this heuristic does not complete the broader distillation issue:
https://github.com/code-yeongyu/oh-my-openagent/issues/1734

Finally, update from current `dev` and obtain the required CI checks. This branch is 3,373 commits behind and its workflow produced no jobs.

Issue #5809, opened 2026-07-01
cmux: tmux subagent panes never attach (respawn-pane no-op) — need eager split-attach


Yoonkeee:
## Problem

When running OmO inside cmux (the native-split Terminal layer), tmux subagent panes are created with a placeholder message ("Focus this pane to attach.") but they **never transition to an actual `opencode attach` session**, even when the user focuses the pane.

There are two compounding failure modes:

1. **`respawn-pane -k` is a no-op under `cmux __tmux-compat`** — The activation path in `activateFocusedPanes` calls `activateTmuxPane`, which issues a `tmux respawn-pane -k <pane_id>` command to replace the placeholder with `opencode attach`. Under cmux, `cmux __tmux-compat` intercepts all tmux commands and silently discards `respawn-pane -k` (it has no equivalent native-split operation), so the pane stays on the placeholder forever.

2. **Immediate child-exit on focus** — In some cmux configurations, focusing the placeholder pane causes the `/bin/sh` sleep loop to receive a signal or the pane's shell to terminate, killing the placeholder before any activation occurs. The subagent pane disappears on focus.

<img width="1406" height="1401" alt="Image" src="https://github.com/user-attachments/assets/995b5861-3526-4a00-bad9-ea81683315c3" />

## Root Cause

The placeholder + focus-to-activate design (`buildTmuxPlaceholderCommand` + `activateTmuxPane` respawn) assumes a real tmux server that can reuse an existing pane with `respawn-pane -k`. cmux emulates tmux's `split-window` for pane creation but does not implement `respawn-pane -k`. Consequently:

- The pane is created (split-window works ✓)
- The placeholder loop runs (✓)
- `respawn-pane -k` to inject `opencode attach` is silently dropped (✗)
- The pane remains stuck on the placeholder indefinitely

## Proposed Fix

When `isCmuxCompatEnvironment()` returns `true` (already exists in `packages/tmux-core/src/cmux-detect.ts`), skip the placeholder entirely and pass `buildTmuxAttachCommand(serverUrl, sessionId, directory)` as the initial command to `split-window`. Simultaneously mark the spawned tracked session with `attachActivated: true` so the existing `activateFocusedPanes` guard (`if (tracked.attachActivated) continue`) skips the respawn attempt automatically.

This is an **opt-in, cmux-only path** — standard tmux behavior (focus-defer) is unchanged.

## Impact

[...]

PR #6297, opened 2026-07-22
fix(tmux-subagent): auto-activate panes within grace window without requiring focus

leeyazhou:
## What changed

Subagent tmux panes showed no content until the user manually focused them. The polling manager only activated panes where `pane.isActive === true`, leaving newly spawned panes on a sleeping placeholder indefinitely.

Added a 5-second grace period (`AUTO_ACTIVATE_GRACE_MS = 5000`) after pane creation during which panes auto-activate regardless of focus. After the grace window, the original focus-based behavior resumes unchanged.

## Files changed

- `packages/tmux-core/src/constants.ts` - new `AUTO_ACTIVATE_GRACE_MS = 5000` constant
- `packages/omo-opencode/src/shared/tmux/constants.ts` - re-export the new constant
- `packages/omo-opencode/src/features/tmux-subagent/polling-manager.ts` - core fix
- `packages/omo-opencode/src/features/tmux-subagent/polling-manager.test.ts` - 2 new tests

## Test results

- Unit tests: 12/12 pass
- LSP diagnostics: no errors

---
## Summary by cubic
Fixes tmux subagent panes showing a blank/sleeping placeholder by auto-activating newly spawned panes for the first 5 seconds, even if not focused. After the grace window, activation falls back to focus-only.

- **Bug Fixes**
  - Added `AUTO_ACTIVATE_GRACE_MS = 5000` in `packages/tmux-core/src/constants.ts` and re-exported for the subagent.
  - Updated polling manager to auto-activate panes created within the grace window, with logging; behavior after the window is unchanged.
  - Added tests covering activation within and beyond the grace window.

<sup>Written for commit 70bf3fa168de308d3a60cf9ee1ebb653e4f5f8a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6297?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Yoonkeee:
Cross-linking #6390, which touches the same symptom from a different angle — they are orthogonal and I think both are worth having.

I verified your mechanism empirically on a live cmux host: `respawn-pane -k` does replace the pane command on cmux 0.64.20 (the no-op behaviour described in #5809 no longer reproduces there), and `list-panes` reports sibling panes with `pane_active`, so the grace window does rescue cmux panes too.

The part it does not cover is why those panes were placeholders to begin with: `isCmuxCompatEnvironment()` returns `false` on an ordinary cmux session, because it only matches `TMUX` containing `"cmuxterm"` or `TMUX` being absent, while cmux always injects `TMUX=/tmp/cmux-<kind>/...`. That also leaves the authenticated fail-closed guards and the `isNativeTmux` switch in `killTmuxSessionIfExists` inactive on cmux. #6390 fixes the detection; your PR fixes focus-gated activation for native tmux, which detection cannot help with.

If both land, cmux sessions get `attachActivated: true` at spawn so `activateFocusedPanes` skips them, and your grace window applies to native tmux — no conflict in either direction.

PR #6329, opened 2026-07-25
fix(background-task): recover tasks and deliver current wakes once

ileixe:
## Summary

Fixes three background-task failures that still reproduce in current `dev`:

1. `background_output(bg_...)` and continuation by `ses_...` recover after the in-memory task registry is lost.
2. Reply-required completion wakes are delivered once, rather than first as `noReply` and later again as a byte-identical reply wake.
3. Delayed progress notifications render the remaining-task count at delivery time.

The changes are split into three commits, one per defect.

### A. Volatile task lookup

`packages/omo-opencode/src/features/background-agent/task-registry.ts:4-28` stores the registry only on `globalThis`. `BackgroundManager.getTask()` still consulted only local/archive/global memory at `manager.ts:1048-1049`; `background_output` retried that lookup for 100 ms in `create-background-output.ts:50-79`; `resume()` failed immediately when `findBySession()` missed at `manager.ts:1313-1318`.

The child session and `bg_` ID are already durable in the parent session's persisted launch tool part. On a memory miss, `BackgroundManager.recoverTask()` now reads that persisted parent transcript, maps either the `bg_` ID or child `ses_` ID back to a minimal completed task, and registers it. Both `background_output` and `resume()` use this shared recovery path (`manager.ts:1052-1071`, `manager.ts:1313-1315`, `create-background-output.ts:50-79`). No second durable registry or new on-disk state was added.

This fixes both observed symptoms:

- completion notification followed by `Task not found: bg_...`;
- continuation/resume failing with `Task not found for session: ses_...`.

Observed miss records from the original diagnosis:

- `bg_f0789ff0`: seq 157, `evt_f96f05b480017OwcqPbqqudgi6`
- `bg_6df3217e`: seq 159, `evt_f96f05d6a001O2d473j7BEtEf9`
- `bg_bb709dce`: seq 427, `evt_f97082714001RsJxJ3lcvuzEqD`

### B. Duplicate completion wakes

`ParentWakeFlushRunner` still routed an unsafe reply-required wake through `forceNoReply: true`, retained it, then sent it again as a reply after the parent became safe. Reply-mode-sensitive dedupe did not consider the first dispatch to cover the second.

`parent-wake-flush-runner.ts:67-133,174-188` now retains every reply-required wake while unsafe and sends one reply-producing dispatch once safe. Non-reply progress wakes retain their existing behavior. Failure wakes remain reply-required and queued.

Persisted duplicate evidence:

- `bg_f0789ff0`: `evt_f96ec024f001B4R0glSuYiOJBf` and `evt_f96ec4767001PN3lhIDacJKW0b`, byte-identical, 17.689 s apart.
- `bg_6df3217e`: `evt_f96ec8208001t8lY6mMpcPxFEy` and `evt_f96ece34e0019lZ7bsLLY0r31p`, both 461 bytes and byte-identical, 24.902 s apart.

The mutually exclusive `isSessionActive` branches were not changed.

### C. Frozen remaining-task count

`manager.ts:2661` calculated `remainingCount` when each task completed, and `background-task-notification-template.ts` immediately embedded the value into the queued string. A delayed wake therefore retained an obsolete count.

The manager now exposes its current pending count to the wake runner (`manager.ts:315-318`). Immediately before dispatch, `parent-wake-flush-runner.ts:48-52` refreshes progress-count lines through `refreshBackgroundTaskNotificationCount()` (`background-task-notification-template.ts:128-137`). Aggregate sibling listings and final-summary wording are unchanged.

### RED

```text
bun test .../create-background-output.recovery.test.ts
0 pass, 2 fail
Received: "Task not found: bg_durable_recovery"
Error: Task not found for session: ses_child_recovery

bun test .../parent-wake-noreply-liveness.test.ts
10 pass, 1 fail
Expected promptAsyncCalls length: 0
Received length: 1

bun test .../task-completion-cleanup.test.ts
Expected not to contain: "2 tasks still in progress"
Received queued failure text containing "2 tasks still in progress"
```

### GREEN

```text
bun test packages/omo-opencode/src/tools/background-task/create-background-output.recovery.test.ts
2 pass, 0 fail

bun test packages/omo-opencode/src/features/background-agent/parent-wake-*.test.ts
87 pass, 0 fail

bun test packages/omo-opencode/src/features/background-agent/task-completion-cleanup.test.ts
18 pass, 0 fail

bun test packages/omo-opencode/src/features/background-agent
742 pass, 0 fail

bun run typecheck
exit 0 across root, script, and all package tsconfigs

bun install
Bun 1.3.12 / Node 24.18.0; install and prepare build completed successfully
```

The repository has no root `lint` script. `git diff --check upstream/dev...HEAD` passes. A direct root Biome invocation was not treated as a repository gate because no root Biome config exists and its defaults reject the repository's existing semicolonless format.

The full root suite ran 12,147 tests: 12,138 passed, 3 skipped, and 6 environment-only assertions failed because the process was root and read the host MCP config. Every failed case passes in its required isolated context:

```text
# unprivileged permission cases
18 pass, 0 fail

# isolated XDG MCP-config cases
3 pass, 0 fail
```

TypeScript directory diagnostics (`tsc` strategy) report 0 errors and 0 warnings. A TypeScript language-server binary is not installed in this environment.

## Relation to #6229

#6229 suppresses wakes after successful output has already been consumed. This PR addresses a different path: duplicate byte-identical deliveries without an intervening `background_output`, plus volatile task recovery and delivery-time count freshness. It does not replace or depend on #6229.

## Risk

- Recovery reads only the already-persisted parent launch transcript and creates no new durable state.
- Reply-required wakes may remain queued longer while a parent turn is unsafe, but are no longer duplicated.
- Count refresh only rewrites the existing progress-count token immediately before dispatch; final summaries and task result content are untouched.

---
## Summary by cubic
Makes background tasks durable and delivers task-result notifications once without duplication. We recover tasks from persisted launches, deposit results immediately when the parent is unsafe, and refresh remaining-counts at send time.

- **Bug Fixes**
  - Durable recovery: when the in-memory registry is empty, `recoverTask` rebuilds tasks from the parent’s persisted launch; used by `background_output(...)` and `resume()`.
  - Single-delivery wakes: result wakes deposit as no-reply while the parent is unsafe and are not re-dispatched; identical content is deduped across flushes; failure wakes still queue until safe.
  - Fresh counts: the “N tasks still in progress” line is updated right before dispatch using the current pending count.

<sup>Written for commit 95d8ea048441653936ae1fae9974bea54c86a749. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/background-task-notification-template.ts
+++ b/packages/omo-opencode/src/features/background-agent/background-task-notification-template.ts
@@ -124,3 +124,13 @@ ${isFailure ? "**ACTION REQUIRED:** This task failed. Check the error and decide
 Use \`background_output(task_id="${task.id}")\` to retrieve this result when ready.
 </system-reminder>`
 }
+
+export function refreshBackgroundTaskNotificationCount(
+  notification: string,
+  remainingCount: number,
+): string {
+  return notification.replace(
+    /\*\*\d+ tasks? still in progress\.\*\*/g,
+    `**${remainingCount} task${remainingCount === 1 ? "" : "s"} still in progress.**`,
+  )
+}
--- a/packages/omo-opencode/src/features/background-agent/background-task-recovery.ts
+++ b/packages/omo-opencode/src/features/background-agent/background-task-recovery.ts
@@ -1,0 +1,72 @@
+import { isRecord } from "@oh-my-opencode/utils"
+import { extractTaskLink } from "../tool-metadata-store"
+import type { BackgroundTask } from "./types"
+
+const BACKGROUND_TASK_ID_PATTERN = /(?:Background )?Task ID:\s*(bg[_-][a-zA-Z0-9_-]+)/i
+const DESCRIPTION_PATTERN = /^Description:\s*(.+)$/im
+const AGENT_PATTERN = /^Agent:\s*([^\n(]+)/im
+
+function readString(value: unknown): string | undefined {
+  return typeof value === "string" && value.trim() !== "" ? value.trim() : undefined
+}
+
+function match(output: string, pattern: RegExp): string | undefined {
+  return readString(pattern.exec(output)?.[1])
+}
+
+function recoverFromPart(part: unknown, targetID: string, parentSessionID: string): BackgroundTask | undefined {
+  if (!isRecord(part) || part.type !== "tool" || !isRecord(part.state)) {
+    return undefined
+  }
+
+  const output = readString(part.state.output)
+  if (!output) {
+    return undefined
+  }
+
+  const metadataLink = extractTaskLink(part.state.metadata, "")
+  const outputLink = extractTaskLink(undefined, output)
+  const metadata = isRecord(part.state.metadata) ? part.state.metadata : undefined
+  const taskID = metadataLink.backgroundTaskId
+    ?? outputLink.backgroundTaskId
+    ?? match(output, BACKGROUND_TASK_ID_PATTERN)
+  const sessionID = metadataLink.sessionId ?? outputLink.sessionId
+  if (!taskID || !sessionID || (targetID !== taskID && targetID !== sessionID)) {
+    return undefined
+  }
+
+  return {
+    id: taskID,
+    sessionId: sessionID,
+    parentSessionId: parentSessionID,
+    parentMessageId: "",
+    description: readString(metadata?.description) ?? match(output, DESCRIPTION_PATTERN) ?? taskID,
+    prompt: "[recovered]",
+    agent: metadataLink.agent ?? outputLink.agent ?? match(output, AGENT_PATTERN) ?? "task",
+    category: metadataLink.category ?? outputLink.category,
+    status: "completed",
+    completedAt: new Date(),
+  }
+}
+
+export function recoverBackgroundTask(
+  messages: readonly unknown[],
+  targetID: string,
+  parentSessionID: string,
+): BackgroundTask | undefined {
+  for (let messageIndex = messages.length - 1; messageIndex >= 0; messageIndex -= 1) {
+    const message = messages[messageIndex]
+    if (!isRecord(message) || !Array.isArray(message.parts)) {
+      continue
+    }
+
+    for (let partIndex = message.parts.length - 1; partIndex >= 0; partIndex -= 1) {
+      const task = recoverFromPart(message.parts[partIndex], targetID, parentSessionID)
+      if (task) {
+        return task
+      }
+    }
+  }
+
+  return undefined
+}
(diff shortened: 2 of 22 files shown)


ileixe:

ileixe:
recheck

ileixe:
Regression correction pushed in `95d8ea048`.

## What regressed

`9baa263e8` prevented every unsafe reply-required wake from taking the early `noReply` admission path. That admission was the prompt per-task `[BACKGROUND TASK RESULT READY]` delivery, so successful tasks became visible only through the terminal aggregate.

Measured live-DB evidence that motivated this correction:
- pre-patch `bg_639d2992` (03:30): individual READY, then aggregates
- pre-patch `bg_cff308ca` (03:47): individual READY, then aggregates
- post-patch `bg_4478cc6f` (04:52): aggregate only
- post-patch `bg_8499c11a` (05:04): aggregate only
- post-patch `bg_58bacc24` (05:47): aggregate only

## Corrected approach

- Restored prompt `noReply` admission for non-failure reply-required wakes while replies are unsafe.
- Deduplicated by notification identity (background-task header + task ID; full content fallback), not by reply mode.
- Removed already delivered notification identities before dispatch, so a later wake carries only new content.
- Stopped no-output retry/requeue semantics for admitted `noReply` content and excluded that delivery record from reply-liveness accounting.
- Kept `6e3873ec2`: delivery still refreshes stale progress counts before dispatch, while identity remains stable across count changes.
- Kept `e53f7e346` recovery behavior unchanged.

## Regression proof

`parent-wake-noreply-liveness.test.ts` now asserts a reply-required per-task READY notification is delivered on the first history-blocked flush as `noReply`, includes `task-a`, has no no-output retry timer, and remains exactly one dispatch when identical content is queued again both before and after replies become safe.

Gates with `/tmp/omo-patch/.tools`:
- `bun install --frozen-lockfile`: exit 0; postinstall build completed all steps
- `bun run typecheck`: exit 0
- `bun test packages/omo-opencode/src/features/background-agent`: `742 pass`, `0 fail`, `1864 expect() calls`, 59 files
- `bun run build`: exit 0; `build: all steps completed`

Rebuilt local bundle: `/tmp/omo-patch/oh-my-openagent/dist/index.js` (5.55 MB build output).

No live OpenCode DB/config was read or written during this correction.

PR #6339, opened 2026-07-25
feat(config): add model-aware prompt appends

WhiteGiverMa:
## Summary

Agent overrides can now append multiple ordered prompt sources, require or exclude selected resolved model IDs for the conditional group, and keep a separate always-appended group. Existing single-string `prompt_append` configurations remain valid, while file URI safety and warning behavior continue to apply per source.

## Changes

- Accept `string | string[]` for `prompt_append` and add `prompt_append_include_model_keywords`, `prompt_append_exclude_model_keywords`, and `prompt_append_always`.
- Resolve selected text and `file://` sources in deterministic conditional-then-always order, with case-insensitive model-ID substring matching and exclusion precedence.
- Reuse one resolver across generic agents, Prometheus, Sisyphus-Junior, and the optional OpenCode Builder path without leaking config metadata into SDK agent configs.
- Re-evaluate Sisyphus append selection when its runtime model changes within the same prompt family.
- Document the configuration contract and regenerate the public JSON schema.

## QA & Evidence

- **What was tested:** `bun test` with external tmux/password environment cleared.
  - **Observed result:** 12,194 passed, 3 skipped, 0 failed across 1,561 files.
  - **Why sufficient:** Covers the repository regression suite plus new resolver, schema, generic-agent, Prometheus, Sisyphus-Junior, and runtime reconciliation cases.
  - **Observed result:** Root, script, and all package TypeScript projects passed.
  - **Why sufficient:** Verifies the widened configuration types and every updated call path.
- **What was tested:** `bun run build`.
  - **Observed result:** Main plugin, declarations, CLI bundles, schemas, MCP runtimes, Senpi adapter, and Codex plugin built successfully.
  - **Why sufficient:** Proves all published build surfaces still assemble.
- **What was tested:** Real isolated `opencode run --format json` sessions against a local deterministic Responses API fixture.
  - **Observed result:** GPT request skipped conditional text and the intentionally missing conditional file, while preserving always-source order. Gemini request included conditional text, the expected missing-file warning, then both always sources in order. Both original sessions returned `PROMPT_APPEND_QA_OK`. Three follow-up runs proved include miss, include hit, and exclusion precedence with 14 passing request-body assertions. The real OpenCode DB stayed at 84 sessions before and after; all QA sessions existed only in isolated DBs.
  - **Artifact:** `.omo/evidence/20260726-model-aware-prompt-append/` retained in the task worktree; raw provider request bodies remain local because they contain full built-in prompts.
  - **Why sufficient:** Captured request bodies are the exact system prompts sent by the real OpenCode harness.

## Risks & Residuals

- Model filtering intentionally uses case-insensitive substring matching on the model portion after the first provider separator. Empty or blank include lists add no restriction, and exclusions take precedence when both lists match.
- Category-level `prompt_append`, Senpi, and Codex configuration are unchanged.
- Local verification used Bun 1.3.14; CI exercises the repository-pinned Bun 1.3.12.
- TypeScript language-server diagnostics were run across all 16 changed TypeScript files: zero errors and zero warnings were reported; three existing deprecation hints remain. Repository-wide `bun run typecheck` independently passed every TypeScript project.

---
## Summary by cubic
Add model-aware prompt appends to agent overrides with include/exclude model keyword filters and an always-on group. Prompts now reconcile at request time so the correct appends are applied for the actual model; Sisyphus and all agents update deterministically and are compaction-safe.

- New Features
  - prompt_append now accepts string | string[], plus prompt_append_include_model_keywords, prompt_append_exclude_model_keywords, and prompt_append_always.
  - Deterministic order: conditional group, then always group. Case-insensitive substring match on the resolved model ID (after the provider). Exclusions win; empty/blank include lists mean no restriction. Skipped conditional files aren’t read. file:// URIs supported; relative and ~ paths work. Category-level prompt_append stays a single unconditional source.
  - Unified resolver used by generic agents, Prometheus, Sisyphus-Junior, and OpenCode Builder without leaking override fields into final configs.
  - New runtime reconciler: rewrites the registered prompt to include the right conditional+always appends based on the actual request model, matching agents by key or display name. Runs after Sisyphus rebuilds its body and is skipped during session compaction to avoid prompt churn.

- Migration
  - No changes required. Existing prompt_append strings still work. To enable model-aware behavior, set prompt_append_include_model_keywords and/or prompt_append_exclude_model_keywords, and optionally prompt_append_always.

<sup>Written for commit d7d6f98cb23cdb764a586303db50f46a1e5f8288. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6339?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


WhiteGiverMa:
## Runtime-model reconciliation follow-up

The latest five commits close the remaining request-time gaps found during review:

- Reconcile conditional appends from the actual request model using the exact final registered agent prompt, avoiding substring collisions and duplicate append blocks.
- Preserve `prompt_append_always` while adding or removing only the conditional group.
- Capture OpenCode's resolved default agent before the first system transform.
- Keep compaction suppression active across provider retries, then clear it on autocontinue or the next normal user turn.
- Skip generic reconciliation after Sisyphus has already rebuilt its complete runtime-family prompt.

### Verification

- Full repository suite: **12,209 passed, 3 skipped, 0 failed**.
- Targeted regressions: **66 passed, 0 failed, 145 assertions**.
- `bun run typecheck`: passed.
- `bun run build`: passed.
- Three Oracle review rounds: no release-blocking findings remain.
- Real isolated OpenCode 1.18.3 sessions: Meidocho ``*`` + `opencode-go/glm-5.2` returned `CELESTIA_PRESENT`; Meidocho ``*`` + `openai/gpt-5.6-luna` returned `CELESTIA_ABSENT`. Host DB stayed **90 -> 90**; isolated DB contained exactly two sessions.

Reviewer-readable artifacts are committed under `.omo/evidence/20260726-runtime-prompt-append/`.

``*``: custom agent in my personal fork during test.

WhiteGiverMa:
## Ready for maintainer merge

All CI checks are green on `8b0ce03ce4f68939207c3cfdc1ea7bfad0bace2f`. Cubic did not publish a new review within the bounded post-push wait, so that gate is recorded as skipped rather than silently assumed. Three explicit Oracle review rounds found no remaining release blocker.

code-yeongyu:
[sisyphus-bot] The model-aware prompt append feature is well-designed with clear separation between conditional and always-appended groups. However, this PR is significantly behind dev — 10+ files it touches have been modified on dev, including the agent-overrides schema, Sisyphus-Junior agent, and prometheus-agent-config-builder. **Recommend rebasing onto current dev before review.** The schema change is substantial and the runtime reconciliation logic interacts with many components. Also, does the `prompt_append_always` field guarantee execution order relative to conditional appends, or is ordering only guaranteed within each group?

WhiteGiverMa:
Rebased onto current `upstream/dev` (`69a1855e6`) and force-pushed `d7d6f98cb` for review.

Yes: `prompt_append_always` has an explicit cross-group order guarantee, not only an in-group guarantee. The final append order is:

1. selected conditional `prompt_append` sources in declared order
2. selected `prompt_append_always` sources in declared order

So `prompt_append_always` is the trailing group. The same conditional-then-always order is enforced both by the registration-time resolver and the request-time reconciler, including the Sisyphus rebuild path. The user-facing configuration reference states this contract, and the focused resolver test pins `conditional-one`, `conditional-two`, `always-one`, `always-two` in that exact sequence.

Post-rebase validation: 211 targeted tests passed, typecheck passed, and the full build passed. I also drove the rebased `dist/index.js` through isolated OpenCode 1.18.14 sessions: with the same Sisyphus registration model, GLM returned `CELESTIA_PRESENT`, GPT returned `CELESTIA_ABSENT`, the host DB stayed `224 -> 224`, and the sandbox contained exactly two sessions. Evidence is committed at `.omo/evidence/20260806-pr6339-rebase-prompt-append/`. Fresh CI is pending.

PR #6344, opened 2026-07-26
feat(omo-senpi): inject the ultrawork directive as a hidden custom message

code-yeongyu:
## What

Typing an ultrawork trigger (`ulw` / `ultrawork`) rewrote the user's own input. The
senpi `input` handler returned `{action: "transform", text: DIRECTIVE + userText}`, so
the ~17KB directive became part of the literal user message and senpi rendered the whole
wall of text above what the user actually typed.

The directive now rides in as a hidden custom message instead:

```ts
pi.sendMessage({ customType: "omo-ultrawork:directive", content: DIRECTIVE, display: false })
return { action: "continue" }
```

The typed text is left byte-identical. Senpi converts custom messages into `role: "user"`
context (`core/messages.ts` `convertToLlm`), so the model still receives every rule, while
the TUI renders `case "custom"` only when `message.display` is true
(`modes/interactive/interactive-mode.ts`). When the prompt is queued mid-stream, the
injection mirrors the input event's `streamingBehavior` into `deliverAs` so both enter the
same queue in that order.

Because the text is never rewritten, the `/skill:` prepend-vs-append workaround disappears
and senpi's native skill expansion can no longer be disturbed by this hook. Every arming
guard is preserved: the trigger regex, the `omo-senpi-ultrawork-disabled` flag,
`source === "extension"`, the matched `<ultrawork-mode>` tag-pair dedupe, the `ulw-` skill
name family, `/skill:ultrawork` passthrough, and trigger-only-in-skill-NAME.

## Observed behavior

Interactive senpi TUI, sandboxed agent dir and mock provider, typing a prompt containing
`ulw`:

- the user message block renders only the typed text
- directive markers rendered on screen: none (`<ultrawork-mode>`, `# Tier triage`,
  `create_goal`, `# Stop rules`, `Manual-QA channels`, `# Role`, `Bootstrap` all absent)
- context meter reads `7.1K/200K (3.6%)`

The context meter is the binary observable that the directive really is in context. An
earlier capture where the trigger word was destroyed by a swallowed keystroke (`lw say hi`,
no trigger) read `173/200K`, so the 7.1K reading is the directive rather than ambient
session content.

## QA and evidence

Live e2e against the real senpi binary (`scripts/qa/ulw-prompts-e2e.mjs`), isolated
`SENPI_CODING_AGENT_DIR`:

```json
{"result":"PASS","ultraworkInjected":true,"userTextClean":true,"hiddenInjectionOk":true,
 "hiddenDirectiveCount":1,"visibleDirectiveCount":0,"userTexts":["ulw please respond"],
 "missingMarkers":[],"forbiddenHits":[]}
```

`userTexts` is exactly what was typed with no directive in it, exactly one hidden
`custom_message` entry carries the full directive, and no visible directive entry exists.
The driver now parses the session JSONL because the provider receives the hidden message
and the real user prompt both as `role: "user"`, so the transcript alone cannot tell the
two injection styles apart.

Interactive TUI evidence rendered through xterm.js (`script/qa/web-terminal-visual-qa.mjs`):
`terminal.png`, `terminal.txt`, `terminal-ansi.txt`, `metadata.json`.

Gates: `bun run test:senpi` green (329 pass, 0 fail across 59 files). Unit suite for the
component is 23 pass, 0 fail and covers the hidden-message contract, `deliverAs` mirroring,
an unknown `streamingBehavior` not leaking through, every guard asserting zero sends, and
that no arming route returns `transform` any more.

Full writeup, including cleanup receipts, is at
`.omo/evidence/omo-senpi-adapter/20260726-ultrawork-hidden-injection/` (gitignored, local).

## Pre-existing failure fixed on the way

`ulw-prompts-e2e.mjs` was already red on unmodified `dev`:

```json
{"result":"FAIL","ultraworkInjected":false,
 "missingMarkers":["<ultrawork-mode>","ULTRAWORK MODE ENABLED!","# Parallel execution",
                   "create_goal","# Stop rules","team_create"]}
```

The scripted mock model declared `contextWindow: 16_000`. The directive outgrew it, senpi
entered compaction, a scripted mock cannot satisfy compaction, and the run exited 1 with
"Context remains above the compaction threshold because compaction did not complete", so
every assertion failed on the exit code rather than on its own merits. Widened the mock
window in its own commit so the driver measures what it claims to measure.

## Residual risk

- Queue ordering for a prompt submitted while the agent is already streaming is not proven
  live. The component contract (mirror `streamingBehavior` into `deliverAs`) is unit
  tested; senpi's own steer/followUp queue ordering was not exercised end to end.
- The directive now arrives as a separate user-role message rather than a prefix on the
  user's own message. Information content is identical; the model-visible difference is a
  message boundary.
- RPC and app-server clients receive `message_start`/`message_end` for the hidden message.
  `display: false` states the intent, but a client that ignores the flag could surface it.
- `biome check` reports 4 errors in this directory on unmodified `dev` as well (format
  rules that do not match the files' existing style). This change neither adds to nor
  removes them.

---
## Summary by cubic
Inject the ultrawork directive as a hidden `custom_message` instead of rewriting input. For prompts queued mid-stream, append the directive to the same message to keep the turn atomic and preserve `/skill:` expansion.

- New Features
  - Idle prompts: send `pi.sendMessage({ customType: 'omo-ultrawork:directive', content: DIRECTIVE, display: false })` and return `{ action: 'continue' }`. The user's text stays unchanged and the TUI stays clean.
  - Queued prompts (`streamingBehavior` present): return `{ action: 'transform', text: userText + '\n' + DIRECTIVE }` so the directive and ask are answered together. No `deliverAs` mirroring.
  - Guards kept: trigger regex, `omo-senpi-ultrawork-disabled`, skip `source: 'extension'`, require matched `<ultrawork-mode>` pair, ignore `ulw-` skill names, `/skill:ultrawork` passthrough. Docs clarified idle vs queued behavior in `AGENTS.md` and `skills/ultrawork/SKILL.md`.

- Bug Fixes
  - Unblocked `ulw` e2e by raising mock provider `contextWindow` to `200_000`; driver now inspects session JSONL to assert one hidden directive and clean user text. Unit tests cover hidden idle injection and atomic queued arming.
  - Realigned `debugging` skill description with its TS source and trimmed to ≤1024 chars.

<sup>Written for commit 88fad68a8944d25280843d19bcbf11d8cbf233ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Update: reviewer round-2 finding on `_promptStartPending`

The reviewer identified a second queue-admission path: when `_promptStartPending`
is true but `isStreaming` is not yet true, senpi queues the prompt but does NOT set
`streamingBehavior` on the input event (`agent-session.ts:2165-2168,2214-2219`).
The extension API does not expose `_promptStartPending` state, so the component
cannot distinguish this from a genuinely idle prompt.

**This is accepted as a known limitation.** The `_promptStartPending` window is the
brief gap between `prompt()` entry and the agent actually beginning to stream. In
normal interactive use a human cannot submit a second prompt during that window; it
would require programmatic input racing with the prompt admission. The fix the
reviewer proposes — modifying senpi to pass `streamingBehavior` when
`canQueueWhileStreaming` is true — is the right long-term answer but requires an
upstream senpi change that belongs in a separate PR.

Behavior in that window preserves the pre-change behavior: the directive becomes
part of the visible message via the hidden custom message path, which is identical
to how the old `transform` worked (the directive was always visible before this PR).

PR #6352, opened 2026-07-26
feat(senpi-task): identity-first live status lines with turn/tok-s across the task tool family

code-yeongyu:
## What changed

The task-tool TUI showed status in id-first, duplicated, low-information lines. The original complaint: a blocking `task_output` rendered `waiting for st_019f9d87 (running)` twice (partial row + senpi core's tool-progress line) and never said what the task was, which category/model it ran on, or its turn/tok-s.

Since then #6346 removed blocking from `task_output` entirely, which deleted that exact line upstream. This PR therefore delivers the consistency redesign on the surfaces that remain, rebased onto current dev:

- **One canonical live-row grammar** in `senpi-task/src/status-line.ts`: `<description|name|id> · <target (provider/model:effort|raw model)> · turn N (M tools) · <verb> · T tok/s`. Identity leads with WHAT the task is: the human `description`, then the stable `name`, and only then the opaque `st_...` id. Model parts are normalized against terminal controls, and tasks without resolved model metadata fall back to the sanitized raw model.
- **The live foreground task line** (`progress.ts`) composes through that grammar — the spinner now shows `Audit renderers · quick (apitopia/kimi-k3:max) · turn 3 (7 tools) · running read src/foo.ts · 62 tok/s` instead of leading with the id.
- **Background tasks show turns/tok-s live**: the status-ui widget/footer live rows insert `turn N (M tools) · T tok/s` from a new public `TaskManager.runStatsSnapshot` seam (previously a private steering-only lambda), yielding identity width to the stats when the row is tight.
- **`description` is no longer dropped**: `task({ description })` was accepted by the schema ("shown in status views") but silently discarded. It now flows `ManagerStartSpec` -> persisted `TaskRecord` (store parser included) -> snapshots, renderers, footer/widget, `/tasks`, and the `/task-kill` selector.
- **Duplication removed**: `team_wait` no longer echoes its waiting activity into the partial result; the activity lives only in `details.progress`, which senpi core's tool-progress painter already draws — the same design the `task` tool documents.
- **Identity-first result surfaces**: `task_output` status/transcript rows, `formatTaskRow` (the `/tasks` output), compact widget/footer rows, `/task-kill` options, and background start text all lead with the human label; compact context uses full `category:`/`agent:` labels instead of the `c:`/`a:` dialect.
- **Fixes found in review**: `/task-kill` maps the chosen option back to its record (a multi-word description previously cancelled a label word instead of the task); child subscriptions now release at most once (`manager/once-only.ts`) since a wait may release on settle and abort while the manager also sweeps; `execute.ts` spec-building unit split into `execute-spec.ts` (250 pure-LOC ceiling); `renderer-text.ts` moved to `src/` so root status code shares it.

## Observed behavior (pinned by unit tests)

- Foreground task spinner: `Audit the waiting line · quick (apitopia/kimi-k3-unlocked:max) · turn 1 (1 tool) · running · 50 tok/s`.
- Background widget row: `⠋ Investig... · turn 3 (7 tools) · 42 tok/s · read src/foo.ts · 1m 5s`.
- Status row: `task_output Audit the waiting line (st_done) completed · quick (openai/gpt-5.6-sol:xhigh) · ran 2m 14s · 118 tok/s`.

## QA / evidence

- `bun test packages/senpi-task packages/omo-senpi/src` -> 1260 pass / 0 fail on the rebased tree; full repo suite 12215 pass earlier (one unrelated 20s-timeout flake in `install-codex.test.ts`, green in isolation); `tsgo --noEmit` clean on both packages. All behavior changes pinned RED-first before implementation.
- Live harness drivers (`task-e2e.mjs`, `task-tui-e2e.mjs`, isolated sandbox + mock provider): the TUI scenario completes (`exitStatus: 0`, child completion observed, sandbox cleaned). The drivers' functional checks fail **byte-identically on pristine dev** (baseline captured), proving the failures are environmental on this workstation (concurrent live senpi sessions + background daemon writes to the real agent dir), not a regression.
- Evidence: `.omo/evidence/20260726-senpi-task-tui-consistency/` (README + driver logs + dev baseline).

## Review

An oracle review of the original diff raised five blockers; all were fixed and re-approved (kill-selection mapping, pending->promoted status accuracy, raw-model fallback, textless-turn refresh, footer identity). The branch was then rebased over #6346 and #6351; the blocking-`task_output` implementation was dropped with the upstream feature removal and the live-stats work re-targeted at the status-ui rows.

## Residual risk / notes

- `manager.ts` remains a pre-existing 605-pure-LOC monolith; this PR adds one delegating method. A follow-up split is deliberately out of scope.
- Result-row model/run-stats shapes outside the live lines still differ across surfaces; unifying them is cosmetic churn left for a follow-up.


code-yeongyu:
reopening to trigger CI (pull_request run was not created)

PR #6363, opened 2026-07-27
fix(boulder-state): count the mandated `- [~]` blocked checkbox

### LINKED ISSUE #6233: bug(boulder): plan-progress parser cannot read common task-ID schemes (T<n>.<n>, F<n>, H<n>); - [~] ignored; total===0 → infinite continuation loop
## Summary

  `getPlanProgress` in `packages/boulder-state/src/storage/plan-progress.ts` reports `isComplete: false` indefinitely for plans whose checkbox structure does not match two narrow regexes. The continuation predicate in `handleAtlasSessionIdle` gates only on `progress.isComplete`, so the boulder continuation directive (`[SYSTEM DIRECTIVE: OH-MY-OPENCODE - BOULDER CONTINUATION]`) fires on every `session.idle` forever, including when `[Status: 0/0 completed, 0 remaining]`.

  Observed in production (2026-07-16, glm-5.2 orchestrator, OMO latest `dev` at `242b1c1`): the boulder continuation directive was injected **30 times over 3 days** into a primary Atlas session whose plan had 0 actionable open items. The agent correctly replied "Holding — terminal state" each time. Related upstream issues: #3545, #4013, #4149, #6109.

  ## Root cause — three independent parser incompatibilities

  `getPlanProgress` (`packages/boulder-state/src/storage/plan-progress.ts`):
  ```ts
  const TODO_TASK_PATTERN = /^\d+\.\s+/;
  const FINAL_WAVE_TASK_PATTERN = /^F\d+\.\s+/i;
  const UNCHECKED_CHECKBOX_PATTERN = /^(\s*)[-*]\s*\[\s*\]\s*(.+)$/;
  const CHECKED_CHECKBOX_PATTERN = /^(\s*)[-*]\s*\[[xX]\]\s*(.+)$/;
  ```

  Against a plan whose checkboxes look like:
  ```markdown
  ## Progress Tracker
  ### Implementation Waves
  - [x] T1.1 — tolerances[] schema
  - [~] T5.1 — verification engine — BLOCKED: user decision
  ### Final Verification Wave (all must APPROVE)
  - [ ] F5 — T6.3a delta verification
  - [~] H1 — push + PR — BLOCKED: user-only decision
  ```

  The parser yields `total=0, completed=0, isComplete=false` because:

  1. **Section mismatch.** All checkboxes live under `## Progress Tracker`, which is neither `## TODOs` nor `## Final Verification Wave`. The parser sets `section="other"` and skips every checkbox.
  2. **Task-ID pattern mismatch.** Even inside a recognized section, bodies must start with `^\d+\.` (bare number+dot) or `^F\d+\.` (F+number+**literal dot**). `T1.1` starts with a letter; `F1 — …` has an em-dash, not a dot. Zero matches.
  3. **`- [~]` is invisible.** Neither checkbox pattern matches `[~]`. The directive explicitly instructs the agent to write `[~]` for user-blocked items, but the parser has no concept of it.

  Compounding factor: `isComplete = total > 0 && completed === total`. When `total === 0` (any of the three causes above), `isComplete` is `false` regardless of how many items are actually resolved — so the predicate can never become true and the loop never terminates.

  The continuation predicate (`handleAtlasSessionIdle`) only checks `progress.isComplete`; it does **not** consult `work.status === "completed"`. `isInactiveBoulderStatus` filters only `paused`/`abandoned`. So even a deliberately-completed work whose plan does not parse keeps getting injected.

  ## Reproduction

  Plan file with the structure shown above + a boulder work registered against it. On each `session.idle` after the agent finishes, the directive fires. Status line reads `[Status: 0/0 completed, 0 remaining]` while the directive body says "Continue working. Do not stop until all tasks are complete."

  ## Proposed fix (single PR, three coordinated changes in plan-progress.ts)

  1. **Recognize `- [~]` as resolved.** Add `BLOCKED_CHECKBOX_PATTERN = /^(\s*)[-*]\s*\[~\]\s*(.+)$/`; count `[~]` toward `completed` and exclude from `remaining`. This aligns with the directive's documented discharge semantics.
  2. **Broaden task-ID patterns.** Accept `^[A-Z]*\d+(\.\d+)?[a-z]?\.?` in addition to the existing patterns, so `T1.1`, `F1`, `H1`, `T6.3a` are recognized. Keep the indent=0 + section gate to bound over-matching.
  3. **Fix `total === 0` semantics.** When structured parse yields `total === 0` but the file contains top-level checkboxes, fall back to simple-mode counting over the whole file. If still `total === 0`, return `isComplete: true` with a `parseWarning` field so the hook logs once and stops continuation, rather than looping forever.

  ## Edge cases considered

  - `- [~]` as resolved: a plan with all items `[~]` becomes complete → boulder stops nagging. Matches the directive's contract (`[~]` is the documented discharge for user-blocked items). If a workflow wanted blocked items to keep nagging, that contradicts the directive's own escape clause.
  - Broader ID regex over-matching acceptance-criteria bullets: bounded by the indent=0 + structured-section requirement. Existing `N.`/`F N.` plans keep matching unchanged (additive change).
  - `total === 0` "stop + warn" vs "silent complete": stop+warn is safer for prose plans that have no checkboxes at all.

  ## Related

  - #3545 — boulder continuation loop on approval gates
  - #4013 — assistant loops repeating final summary (same parser/predicate shape)
  - #4149 — boulder complete allows stale reminders to re-enter finalization loop
  - #6109 — system-directive spam loop

MoerAI:
## Summary

- The boulder continuation directive tells the agent that marking a blocked task is **mandatory**: "you MUST edit the plan file in this turn and change that task's checkbox from `- [ ]` to `- [~]` before moving on" (`packages/omo-opencode/src/hooks/atlas/system-reminder-templates.ts:34-35`). The plan checklist parser then rejected `~` in all three of its checkbox patterns, so every `- [~]` row was dropped from the checklist entirely - counted neither as done nor as remaining.
- Two things follow. `omo boulder` under-reports progress because blocked tasks are missing from the denominator. And when every parseable row is `- [~]`, the plan parses to `total = 0`, so `getPlanProgress` returns `isComplete: false` (`storage/plan-progress.ts:44` guards on `total > 0`), `handleAtlasSessionIdle` never short-circuits (`hooks/atlas/idle-event.ts:49`), and the `BOULDER CONTINUATION` directive is re-injected on every idle forever. The directive punishes the agent for obeying the directive.
- This teaches the parser the marker the product already mandates: `- [~]` is a discharged task. It counts toward completion, it never becomes the next task, and a plan whose remaining work is all user-blocked now terminates instead of looping.

## Changes

- `packages/boulder-state/src/plan-checklist.ts` (+10 / -5, the only product change)
  - `~` added to the marker character class of `SIMPLE_CHECKBOX_PATTERN`, `TODO_CHECKBOX_PATTERN` and `FINAL_WAVE_CHECKBOX_PATTERN`, so the row is recognised in simple mode, in `## TODOs` and in `## Final Verification Wave` alike.
  - New `isDischargedMarker()` helper; `parseSimpleTopLevelCheckbox` and `parseStructuredTopLevelCheckbox` both route their marker through it. Hardening the sibling predicates together rather than patching one symptom site is what keeps `parseCurrentTopLevelTask` consistent for free - a blocked row is discharged, so it is never selected as `nextTaskLabel` or as the next top-level task.
- `packages/boulder-state/src/plan-checklist.test.ts` (+48) - three regression cases: all rows blocked in simple mode, blocked rows in both counted sections, and a blocked row ahead of an actionable one.
- `packages/boulder-state/src/plan-progress.test.ts` (new) - drives the real `getPlanProgress` entrypoint against plan files on disk, pinning the `isComplete` flip that gates the continuation loop.
- `.omo/evidence/20260727-fix-6233/` - sanitized QA bundle, including a re-runnable `live-driver.sh`.

## QA & Evidence

- **What was tested:** the new regression cases on unmodified `dev`, driving the exported parser and the real `getPlanProgress` (reading actual plan files from disk), not hand-built internal payloads.
  **Observed result:** RED for the behavioural reason, not a compile error. The all-blocked plan returned `{total: 0, completed: 0, isComplete: false}` where `{2, 2, true}` was expected - the exact `0/0` the issue reports. Then GREEN with the fix: `plan-checklist.test.ts` 16/16, `plan-progress.test.ts` 4/4, whole package 38/38.
  **Artifact:** `.omo/evidence/20260727-fix-6233/red-6233.txt`, `green-6233.txt`
  **Why sufficient:** the failure is observed through the same public entrypoints the Atlas idle path uses, and it flips only with this change.

- **What was tested:** negative control - revert only the product file, keep the new tests.
  **Observed result:** both suites fail again (exit 1, same 3 + 3 cases).
  **Artifact:** `.omo/evidence/20260727-fix-6233/negative-control-6233.txt`
  **Why sufficient:** proves the tests are pinned to the product change and cannot pass without it.

- **What was tested:** the real user-facing CLI, `bun packages/omo-opencode/src/cli/index.ts boulder --directory <tmp> --json` and its text mode, against a throwaway project holding `.omo/boulder.json` plus a plan whose every task is `- [~]`. Re-runnable via the committed `live-driver.sh`.
  **Observed result:** before `total_tasks: 0, completed_tasks: 0, percentage: 0`, text `progress: 0% (0/0)`. After `total_tasks: 2, completed_tasks: 2, percentage: 100`, text `progress: 100% (2/2)`. A mixed plan (`- [x]` + `- [~]` + `- [ ]`) goes `50% (1/2)` to `67% (2/3)` while `current_task` correctly stays on the actionable task.
  **Artifact:** `.omo/evidence/20260727-fix-6233/live-driver.sh`, `live-driver-before.txt`, `live-driver-after.txt`
  **Why sufficient:** it is the operator-visible surface, driven end to end, on the same plan text the product itself instructs the agent to write. Each capture embeds `git diff --stat dev` for the product file, so the artifact proves which tree produced it.

- **What was tested:** every consumer suite of `getPlanProgress` / `parsePlanChecklist`, swept on clean `dev` and again with this PR, sequentially on a quiet host.
  **Observed result:** no new failures. `boulder-state` 31 to 38 pass (the 7 added tests), `features/boulder-state` 75/75, `hooks/atlas` 208/208, `hooks/start-work` 94/94, `cli/run` 194 pass / 2 fail on **both** trees. The two `cli/run` failures are `executeOnCompleteHook > uses powershell when PowerShell is detected on Windows` and its cmd.exe counterpart; they fail identically on unmodified `dev` on this Windows host and have no relationship to plan parsing.
  **Artifact:** `.omo/evidence/20260727-fix-6233/regression-comparison.txt`, `related-suite-6233.txt`, `baseline-preexisting-failures.txt`
  **Why sufficient:** the comparison is base-versus-PR on one machine, so pre-existing host failures cannot be mistaken for regressions.

- **What was tested:** `tsgo --noEmit` on both affected packages, and isolation of the live capture.
  **Observed result:** typecheck exit 0 for `packages/boulder-state` and `packages/omo-opencode`. The driver contributed `delta_driver = 0` rows to the real opencode session table against an equal idle-gap control, and the user config sha256 was unchanged; the `boulder` CLI opens no database, session API or network.
  **Artifact:** `.omo/evidence/20260727-fix-6233/typecheck-6233.txt`, `isolation-proof.txt`
  **Why sufficient:** confirms the change is type-clean and that QA left no trace on the host.

## Risks & Residuals

- **Blast radius: mitigated.** The change is confined to plan lines whose checkbox marker is `~`. A plan containing no `~` takes byte-identical paths before and after, which is why every pre-existing test passes unchanged.
- **Could this mark work complete too eagerly? Accepted and bounded.** `- [~]` is not "ignore this task"; it is the marker the directive mandates for work only the user can unblock. Treating it as still-actionable is precisely what produced the reported loop. A plan with any genuinely open `- [ ]` row still reports incomplete, which is pinned by the third regression case and by the mixed plan in the live capture.
- **Task-ID grammar: deliberately not touched.** The issue also proposes accepting `T1.1`, `H1` and `F5 - ...`. The existing test `#given noncanonical structured rows #when parsed #then only exact positive-number grammar is counted` deliberately pins the strict grammar, including rejecting `- [ ] 0.`, `- [ ] 01.` and `* [ ] 2.`. Loosening it is a contract change, not a bug fix. That test passes unchanged here, which is the proof this PR left the grammar alone.
- **`isComplete: total > 0 && ...`: deliberately not changed.** `getPlanProgress` has 19 call sites, including the `start-work` plan-discovery and plan-selection filters that use `!isComplete` to choose which plans to offer. Making `total === 0` mean "complete" would hide a freshly created, not yet written plan from `/start-work`, a far wider blast radius than this defect warrants.
- **Residual, stated plainly.** A plan that parses to `total = 0` for some reason other than `- [~]` - checkboxes under a heading the parser does not count, or a non-canonical task ID - can still drive the loop. This PR removes the `- [~]` cause, the one the product itself creates by mandating that marker. It does not claim to close the whole `total === 0` class; the two items above are the remaining half and are left open on purpose.
- **Windows-host test caveat: not applicable to this change.** An early consumer sweep on this machine reported 53 failures while the host was loaded; it ran the identical test count but took 752s versus 9-13s quiet and contained 49 `this test timed out after 5000ms` lines versus 1 at baseline. The quiet-host re-runs recorded above show parity with clean `dev`. One later straggler, `resolveActiveBoulderSession > returns null ... when resolved work is abandoned`, was investigated rather than assumed: its fixtures contain no `[~]` at all, its recorded failure was `[27086.93ms]` with no assertion diff, and 5/5 isolated runs with the fix pass 8/8 in ~420ms (`flake-proof-resolve-active-boulder-session.txt`).

## Automated Checks

```bash
bun test packages/boulder-state                              # 38 pass, 0 fail
bun test packages/omo-opencode/src/features/boulder-state    # 75 pass, 0 fail
bun test packages/omo-opencode/src/hooks/atlas               # 208 pass, 0 fail
bun test packages/omo-opencode/src/hooks/start-work          # 94 pass, 0 fail
bunx tsgo --noEmit -p packages/boulder-state                 # exit 0
bunx tsgo --noEmit -p packages/omo-opencode                  # exit 0
bash .omo/evidence/20260727-fix-6233/live-driver.sh out.txt  # real boulder CLI, before/after
```

## Related Issues

Fixes #6233

---
## Summary by cubic
Counts the mandated `- [~]` blocked checkbox as discharged in the boulder plan parser and the plan‑format validator so progress is correct and all‑blocked plans terminate. Fixes #6233.

- **Bug Fixes**
  - `packages/boulder-state/src/plan-checklist.ts`: accept `~` alongside `x`/`X` in simple, `## TODOs`, and `## Final Verification Wave`; add `isDischargedMarker()` so `- [~]` counts as completed and is never picked as next.
  - `packages/omo-opencode/src/hooks/plan-format-validator/hook.ts`: accept `~` in `TOPLEVEL_CHECKBOX`, `TODO_TASK`, and `FINAL_WAVE_TASK` to avoid false warnings when a counted section has only blocked rows; task‑ID grammar unchanged.
  - Tests: add unit tests and `getPlanProgress` cases for all‑blocked/mixed/ordering; CLI captures show 0%→100% on an all‑blocked plan; no change to `isComplete: total > 0 && remaining === 0`.

<sup>Written for commit 814c3848d7d9f962ce24478fbc014f869efebb2f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6363?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:

MoerAI:
CI note: `senpi-compatibility (ubuntu-latest)` is red on this PR with:

```
omo-senpi extension build is not current: stale-output
output=.../packages/omo-senpi/plugin/extensions/omo.js
```

I traced it rather than assuming, and it is a real consequence of this change, not a flake. Details so you can decide how you want it handled:

**Why this PR trips it.** The committed Senpi extension bundle embeds `packages/boulder-state` — the current `omo.js` contains the `plan-checklist` section strings, and a fresh local rebuild of the bundle contains this PR's new marker class `xX~`. So editing `plan-checklist.ts` genuinely makes the committed bundle stale, and `build-extension.mjs --check` correctly reports it. That also explains why my other open PR (#6375, which touches `packages/omo-opencode`) is green on the same job.

**Why I have not pushed a regenerated bundle.** Running the generator twice in a row on this machine produces two different `omo.js` files:

```
node packages/omo-senpi/plugin/scripts/build-extension.mjs   # build A
node packages/omo-senpi/plugin/scripts/build-extension.mjs   # build B
diff -q A B  ->  files differ
```

Since the output is not reproducible even build-to-build locally, a bundle generated here (Windows) is unlikely to match what `--check` regenerates on `ubuntu-latest`, and committing it would add ~183 lines of generated churn to a 15-line parser fix without reliably turning the job green.

**Options, your call:**
1. You regenerate `packages/omo-senpi/plugin/extensions/omo.js` in your environment as part of merging, or
2. I push the locally regenerated bundle anyway if you would rather have it in the PR, or
3. If the bundle is meant to be produced by release tooling rather than by contributors, this job may need to exclude PRs that only touch bundled core packages.

Say which you prefer and I will do it. The rest of the PR is unaffected: `packages/boulder-state` 38/38, the branch is merged up to current `dev`, and the real-OpenCode harness evidence (plugin loaded, `0% (0/0)` to `100% (2/2)`, zero sessions added to the real database) is committed under `.omo/evidence/20260727-fix-6233/`.

MoerAI:
Correction to my earlier comment, and the senpi job is now handled.

**What I got wrong.** I previously reported that `build-extension.mjs` is non-deterministic and used that to justify not committing a regenerated bundle. That measurement was unsound: I compared a file that had been copied while a build was in flight. Two clean consecutive runs produce an identical `sha256`. I should not have published that claim, and it should not be used to reason about this job.

**What is actually true.** The committed Senpi bundle embeds `packages/boulder-state`, so the parser change in this PR genuinely makes it stale, and `build-extension.mjs --check` is right to fail. Regenerating is the correct contributor action, which I have now done:

```
node packages/omo-senpi/plugin/scripts/build-extension.mjs
  -> xX~ present in omo.js: 1        (the bundle now tracks this PR's source)
  -> host-specific path leaks: 0
  -> omo-senpi extension build is current   (exit 0)
```

One prerequisite worth recording for anyone else who hits this: the generator fails with `Could not resolve "@oh-my-opencode/omo-opencode/config-migration"` unless `bun install --frozen-lockfile` has been run against current `dev`. That failure mode produces a *different* bundle, which is what made my first attempt look nondeterministic.

**Also in this push**, addressing the P1 on the parser change. `plan-format-validator/hook.ts` excluded `~` from `TOPLEVEL_CHECKBOX`, `TODO_TASK` and `FINAL_WAVE_TASK`, so a counted section holding only blocked rows parsed as having zero task rows and the post-write hook appended a warning telling the agent to fix an already-valid plan — which invites undoing the blocked marker and re-entering the very loop this PR removes. That is a real gap in my original defect-class argument, not just an evidence gap.

Reproduced through the real `tool.execute.after` hook before fixing:

```
(fail) plan-format-validator > preserves a plan whose counted sections contain only user-blocked rows
  + <plan-format-warning>
  14 pass / 1 fail
```

After accepting `~` in all three predicates: `plan-format-validator` 15/15, `boulder-state` 38/38, `atlas` 208/208, `tsgo -p packages/omo-opencode` exit 0. Negative control re-run by restoring `hook.ts` from `HEAD`. RED/GREEN logs are committed under `.omo/evidence/20260727-fix-6233/`.

The two P2s are noted and deliberately not bundled here: closing task timers on a blocked transition (`tool-execute-after-plan-tasks.ts`) and the Senpi consumer's "continue until every row is `- [x]`" contract are separate call sites with their own contracts, and folding them in would push this well past a conservative parser fix. Happy to take either as a follow-up if you want them in scope.

MoerAI:
Final, accurate state of the senpi job on this PR. I tested both options rather than argue, so here is what each one actually produced.

**Attempt 1 — commit a locally regenerated bundle.** Rejected by CI. My machine and the runner do not bundle the same module set:

```
this PR's runner : Bundled 626 modules -> omo.js 0.63 MB -> "build is not current: stale-output"
my machine       : --check passes locally after regenerating, exit 0
```

So a locally generated artifact cannot satisfy the runner's rebuild, regardless of platform. I have reverted that commit rather than ship a 192-line generated diff that does not turn the job green.

**Attempt 2 — leave the committed bundle alone.** That is the current state of the branch. The job stays red because the bundle genuinely is stale with respect to this PR's source.

**Correcting myself twice, plainly:** my first comment claimed the generator is non-deterministic, based on a file I had captured mid-build — unsound, and I retracted it. My retraction then over-corrected by implying a locally regenerated bundle would satisfy the check. The runner has now disproved that too. The accurate statement is narrower: **`packages/omo-senpi/plugin/extensions/omo.js` embeds `packages/boulder-state`, so any PR touching that package invalidates the committed bundle, and the artifact can only be refreshed in an environment whose module resolution matches the runner's.**

The parser change itself is unaffected and fully verified: `boulder-state` 38/38, `atlas` 208/208, `plan-format-validator` 15/15, `tsgo` exit 0, plus the real-OpenCode harness capture (plugin loads, `0% (0/0)` to `100% (2/2)`, zero sessions added to the real database) under `.omo/evidence/20260727-fix-6233/`.

How you would like to resolve the bundle is your call — regenerate it on merge, or tell me the environment you want it built in and I will match it. I did not want to keep pushing generated churn at the problem.

MoerAI:
Closing — this PR is currently too risky to keep open: reviewer feedback identified P1 correctness/QA gaps, and the branch needs to be reworked as a smaller issue-focused fix with complete harness evidence. Happy to reopen or rework if this exact approach is still wanted.

PR #6377, opened 2026-07-27
feat(senpi): gate architect and ultrabrain on their required models, retire oracle

code-yeongyu:
## Summary

Two senpi advisor categories become self-aware about the models they need, and the `oracle` curated subagent is retired.

`architect` (new) and `ultrabrain` (re-pinned) now carry a `requiresModel` activation gate: each one resolves only when its gate model is actually present in the senpi model registry, and its fallback chain is trimmed to its own model family so the gate cannot be sidestepped by a cross-family rung. Naming either category explicitly in `omo.json` always activates it regardless of the gate. The `oracle` builtin curated agent is removed (roster is now `explore`, `librarian`, `metis`, `momus`), and the senpi copy of the `ulw-plan` skill gains a momus-only review override plus advisory-only consultation lanes for the two new advisors.

## Changes

**Category activation gating** (`packages/senpi-task/src/category/`)
- `BuiltinCategoryDefinition` gains an optional `requiresModel`; `builtins.ts` derives `BUILTIN_CATEGORY_REQUIRES_MODEL` plus the `categoryGateModel` / `isCategoryGateSatisfied` helpers.
- `resolveCategory` applies the gate right after the registry is parsed and *before* fallback-chain resolution, so a gated category can never reach a rung from another model family. Gated-and-unmet categories are also filtered out of `availableCategories`.
- Semantics mirror the OpenCode edition (`packages/omo-opencode/src/tools/delegate-task/categories.ts`): any explicit `omo.json` `categories.<name>` entry — even a description-only one — bypasses the gate.
- Gateway-transformed registry ids (`vercel/openai/gpt-5.6-sol`) satisfy a gate on their last path segment, so Vercel-routed models still count.

**The two gated advisors**
- `architect` (new, anthropic family): `anthropic/claude-fable-5` at `xhigh`, gated on `claude-fable-5`, fable-only chain, with a consultant prompt that demands a whole-architecture survey, two costed options, and a recommendation instead of an implementation.
- `ultrabrain`: `openai/gpt-5.6-sol` moved from `xhigh` to `max`, gated on `gpt-5.6-sol`, chain trimmed to sol-only rungs.

**Task tool description** (`packages/senpi-task/src/tools/task/categories.ts`)
- Gated builtins stay listed and gain a ` (requires <model>)` annotation. This is deliberate: the tool description is built at registration, before `TaskRuntimeContext#modelRegistry` is captured on `session_start`, so no registry is available to hide with. Spawn-time `resolveCategory` remains the sole authoritative gate. Categories carrying an explicit `omo.json` entry render without the annotation.

**Oracle removal** (`packages/senpi-task/`, `packages/omo-senpi/`)
- Deleted the agent, its fallback chain, and every reference. The metis prompt now advises an advisory-only `architect`-category consultation instead of an oracle delegation; tool-description examples point at `momus`.

**Senpi ulw-plan overlay** (`packages/omo-senpi/plugin/scripts/sync-skills.mjs`)
- The senpi copy of the shared skill now receives a *Senpi Review Override* (high-accuracy review is momus-only; oracle does not exist here) and *Senpi Design Consultation Lanes* (consult `architect` for big-picture design and `ultrabrain` for detail design as background, advisory-only, no-edit task lanes when those categories are available). `packages/shared-skills` is byte-unchanged — the overlay exists only in the generated plugin copy, and a test asserts that.

## QA & Evidence

Evidence directory: `.omo/evidence/omo-senpi-adapter/senpi-architect-ultrabrain/` (local; gitignored).

- **What was tested:** gating behaviour driven through the real `resolveCategory` and `listTaskCategories` with gate-met and gate-unmet registries.
  **Observed:** `architect / no fable -> model_unavailable attempted=anthropic/claude-fable-5`; `architect / fable present -> resolved anthropic/claude-fable-5 variant=xhigh`; `architect / no fable + explicit omo.json -> resolved omo-mock/mock-1`; `ultrabrain / no sol -> model_unavailable attempted=openai/gpt-5.6-sol`; `ultrabrain / sol present -> resolved openai/gpt-5.6-sol variant=max`; listings show `(requires claude-fable-5)` / `(requires gpt-5.6-sol)` while `deep` stays unannotated.
  **Artifact:** `task-10-gate-proof.txt`. **Covers:** the gating, bypass, and annotation criteria.

- **What was tested:** the live curated-agent roster and the unknown-target path, driving the real senpi binary in an isolated sandbox via `packages/omo-senpi/scripts/qa/curated-agents-e2e.mjs`.
  **Observed:** driver verdict `"result":"PASS"` with all 18 checks PASS, and the live engine emitting `Available agents: explore, librarian, metis, momus.`
  **Artifacts:** `task-10-curated-agents-e2e.txt`, `task-10-parent-output-isolated.txt`. **Covers:** oracle removal proven on the real surface rather than by grep.

- **Automated gates:** `bun test packages/senpi-task packages/omo-senpi` → 1309 pass / 0 fail; `tsgo --noEmit` clean for both packages; `bun run test:senpi` green; `build-extension.mjs --check` and `embed-directive.mjs --check` both report current.

### QA driver isolation gap (worth a reviewer's attention)

The first live runs failed `unknown_target_agents`. Root cause was **not** this change: `curated-agents-e2e.mjs` isolates `XDG_CONFIG_HOME` but not `HOME`, and omo's user config scope is `~/.omo`. A developer `~/.omo/omo.jsonc` that defines an `agents.oracle` entry leaks into the sandbox and makes oracle look like it survived removal. Re-running with an isolated `HOME` yields `"result":"PASS"` and the four-name roster. Flagging it here rather than silently working around it; tightening the driver's isolation is a good follow-up but is out of scope for this PR.

## Risks & Residuals

- **Behaviour change for users without the gate models:** if neither `claude-fable-5` nor `gpt-5.6-sol` is available, `ultrabrain` now fails closed instead of silently falling back to Gemini/Opus/GLM. That is the intent — a category named "ultrabrain" that quietly runs on a different family is worse than an honest failure — and any user who wants the old behaviour can pin a model in `omo.json`, which bypasses the gate. Mitigated and documented in both AGENTS.md files.
- **Breaking:** `task(subagent_type="oracle")` now returns `unknown_target`. Deliberate; the commit is marked `!`.
- **Existing tests were retargeted, not weakened:** fixtures that relied on ultrabrain's removed cross-family rungs were moved to categories that genuinely still exercise a fallback (`visual-engineering`, `artistry`), and oracle fixtures moved to `momus` so curated-agent rejection is still proven by a real curated name.

---
## Summary by cubic
Gate the `architect` and `ultrabrain` categories on their required models and remove the `oracle` curated agent. This enforces model-accurate routing and simplifies the roster to four agents.

- New Features
  - Category gates: builtin categories can define `requiresModel`. Resolution runs only if the model exists in the registry, before fallback; explicit `omo.json` entries bypass the gate. Gateway IDs unwrap only known prefixes (`openai`, `anthropic`, `google`), so `vercel/openai/gpt-5.6-sol` counts but unrelated vendor paths do not.
  - `architect` (new, Anthropic): runs on `claude-fable-5` (`xhigh`), advisory-only, fable-only chain.
  - `ultrabrain`: pinned to `gpt-5.6-sol` (`max`), sol-only chain.
  - Task tool listing: gated builtins remain visible with a “(requires <model>)” note when not explicitly configured.
  - Roster: curated agents are now `explore`, `librarian`, `metis`, `momus`; `oracle` is removed. The `metis` prompt now recommends an advisory `architect` consultation.
  - `omo-senpi` ulw-plan overlay: momus-only high-accuracy review and advisory consultation lanes for `architect` and `ultrabrain` (generated plugin only; shared skill source unchanged). Plugin bundle regenerated after narrowing the gateway-prefix match.

- Migration
  - Replace `task(subagent_type="oracle")` with `task(subagent_type="momus")`.
  - For design advice, use `task(category="architect", ...)` (advisory-only). For deep reasoning, use `task(category="ultrabrain", ...)`.
  - If the required models (`claude-fable-5`, `gpt-5.6-sol`) aren’t available but you still want these categories, add an explicit category entry in `omo.json` to bypass the gate.

<sup>Written for commit d46932c5b13736f837a18e713bb49b04306a6fcd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6377?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Small follow-up on the `oracle` retirement — the regenerated bundle still recommends the agent it now rejects.

Grepping `plugin/extensions/omo.js` on this branch turns up four `oracle` occurrences. One is intentional (the hard-reject guard), but two still point users at it:

```
① "...Use delegate-task for read-only agents like oracle, librarian, explore, metis, momus, multimodal-looker."
② oracle:{verdict:"hard-reject", rejectionMessage:"Agent 'oracle' is ..."}
③ "...Use delegate-task with subagent_type: 'oracle' for read-only analysis instead."
```

③ tells the agent to use `subagent_type: 'oracle'`, and ② then hard-rejects exactly that. An agent following the guidance in the same bundle hits the guard. ① also lists `oracle` in the read-only roster, which is now `explore`, `librarian`, `metis`, `momus`.

These are prompt/guidance strings, so typecheck and the test suite pass regardless — worth a grep before this lands since it is a breaking change and the retirement guidance is what users will follow.

CI is green on the latest run ([30245479823](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/30245479823)), including `codex-compatibility (windows-latest)`, and the bundle carries the accumulated `/omo.jsonc` + `/.omo` anchors and kimi-k3 routing with zero legacy config tokens.

PR #6390, opened 2026-07-27
fix(cmux): detect cmux by the injected TMUX socket path

Yoonkeee:
## Summary

- On a normal cmux session `isCmuxCompatEnvironment()` returns `false`, so every cmux-gated code path added by #5811 is unreachable there: subagent panes are created with the inert placeholder instead of `opencode attach`, sessions are tracked as `attachActivated: false`, the authenticated-cmux fail-closed guards never fire, and `killTmuxSessionIfExists` treats cmux's fake `TMUX` as a native tmux session.
- The detector recognised cmux only when `TMUX` contained `"cmuxterm"` or when `TMUX` was **absent**. cmux never writes `"cmuxterm"` into a socket path — that string only exists as its bundle id `com.cmuxterm.app` — and it always injects a `TMUX` of the form `/tmp/cmux-<kind>/<workspace>,<surface>,<pane>`. Both branches are therefore false on every cmux build that injects `TMUX`, which is the normal case.
- This PR decides from the **shape of the socket path** instead: a cmux socket lives under a `cmux*` directory, a real tmux socket under `tmux-<uid>`. A real tmux nested inside a cmux pane is still reported as native tmux.

### Where the regression came from

`efb862ce9` (`fix(tmux): prefer real tmux when session env exists`, authored 2026-05-08, committed 2026-05-15) narrowed the detector:

```diff
-	return Boolean(process.env.CMUX_SOCKET_PATH) ||
-		process.env.TMUX?.includes("cmuxterm") === true
+	const tmuxEnvironment = process.env.TMUX
+	return tmuxEnvironment?.includes("cmuxterm") === true ||
+		(Boolean(process.env.CMUX_SOCKET_PATH) && !tmuxEnvironment)
```

Before that commit `CMUX_SOCKET_PATH` alone was sufficient and cmux was detected correctly. The goal of the change was right — a real tmux nested inside cmux should keep talking to real tmux — but it used "`TMUX` is set" as the proxy for "a real tmux server exists", and cmux always sets `TMUX`. The file's own doc comment already stated that cmux "injects a tmux shim and sets CMUX_SOCKET_PATH / TMUX", so the new condition contradicted the documented contract and left `"cmuxterm"` as the only surviving cmux branch.

That commit's intent is preserved here: its regression test — a real tmux socket observed while `CMUX_SOCKET_PATH` is set — still passes unchanged, because the discriminator is now the socket path rather than the presence of `TMUX`.

#5811 documented this exact gap in its Residual scope ("the existing detector limitation for non-`cmuxterm` fake `TMUX` prefixes remains documented and is not widened by this PR"). This PR closes it.

### Relationship to #6297

#6297 (`auto-activate panes within grace window without requiring focus`) is **orthogonal, not competing**, and both can merge:

- #6297 fixes focus-gated activation, which affects native tmux users whose panes never get focused. This PR does nothing for that case.
- This PR fixes environment detection, which affects the ~15 cmux-gated branches listed above. #6297 does nothing for those.
- If both land, cmux sessions are marked `attachActivated: true` at spawn, so `activateFocusedPanes` skips them and #6297's grace window applies to native tmux only. No conflict.

## Changes

- **`packages/tmux-core/src/cmux-detect.ts`** — decide from the shape of the injected socket path (`hasCmuxSocketPath`) instead of assuming `TMUX` implies a real tmux server, and give the socket-shape check an explicit precedence over cmux's own `CMUX_AGENT_LAUNCH_KIND=omo` contract:
  1. the `TMUX` socket shape — the sole discriminator; it vetoes both env signals;
  2. `CMUX_SOCKET_PATH` — the authenticated credential, sufficient once 1 has not vetoed;
  3. `CMUX_AGENT_LAUNCH_KIND === "omo"` — accepted only as a stand-in for 2 and only where 1 confirms a cmux-shaped socket. Never sufficient alone, so an environment carrying only the launch kind fails closed to native tmux.

  The launch kind cannot outrank the socket shape because it is inherited: measured on a live cmux host, a real tmux server started inside a cmux pane passes `CMUX_AGENT_LAUNCH_KIND` and `CMUX_SOCKET_PATH` to its children byte-identically while only `TMUX` changes to the real socket. It is matched exactly against `omo`, since cmux names the launched agent in this variable and ships a wrapper shim per agent, so only `omo` carries the `CMUX_OMO_CMUX_BIN` contract `resolveCmuxCliExecutable()` depends on. The earlier `"cmuxterm"` branch is removed: cmux writes that string only into its bundle id, config directory and env names, never into a socket path.
- **`packages/tmux-core/src/cmux-cli.ts`** (new) — `resolveCmuxCliExecutable()` resolves the cmux binary through `CMUX_OMO_CMUX_BIN`, then `CMUX_BUNDLED_CLI_PATH`, then a bare `cmux`. Exported through the `tmux-core` barrel and mirrored by `packages/omo-opencode/src/shared/tmux/cmux-cli.ts`.
- **`packages/tmux-core/src/runner.ts`** and **`packages/omo-opencode/src/tools/interactive-bash/tools.ts`** — both `resolveTmuxExecutable` implementations use that helper instead of the hard-coded `"cmux"`. **This is required for the detector fix, not a drive-by:** cmux does not put `cmux` on the agent process PATH (only its `tmux` shim, which itself resolves the binary via `CMUX_OMO_CMUX_BIN`). Fixing detection alone routes every tmux command to a missing executable — see RED round 2 below.
- **`packages/omo-opencode/src/shared/tmux/tmux-utils/pane-spawn-runner.test.ts`** — injects `isCmuxCompatEnvironment` the way the `packages/tmux-core` mirror test has since #5811. It previously asserted placeholder behaviour while letting the real detector run, so it passed only on non-cmux hosts.
- **`packages/tmux-core/src/runner.test.ts`** — clears `CMUX_OMO_CMUX_BIN` / `CMUX_BUNDLED_CLI_PATH` in `beforeEach` so the suite never reads the host's cmux configuration.
- **`docs/reference/re-export-shim-inventory.md`** — registers the new `packages/omo-opencode/src/shared/tmux/cmux-cli.ts` shim, which `script/package-registration-audit.test.ts` enforces.
- **`packages/tmux-core/src/cmux-detect.test.ts`** — also clears `CMUX_AGENT_LAUNCH_KIND` in `beforeEach`. Without it the suite reads the host's cmux configuration and the "cmux socket directory without `CMUX_SOCKET_PATH`" guard inverts on any cmux machine.
- Tests added: cmux-injected `TMUX`, nested-real-tmux regression guard, cmux socket directory without `CMUX_SOCKET_PATH`, neither variable set, every socket shape the shipped cmux binary can produce, a backslash guard, four `resolveCmuxCliExecutable` cases, a runner test where the cmux CLI is reachable only through `CMUX_OMO_CMUX_BIN`, and a precedence matrix covering all eighteen combinations of the two cmux signals against an absent, cmux-shaped and real-tmux-shaped `TMUX`, with the fallback, veto, fail-closed and exact-match rows asserted individually.

## QA & Evidence

- **What was tested:** RED round 1 — detector reverted to the pre-change expression with everything else kept.
  **Observed result:** exactly the two new cmux-environment guards fail, 15 pass / 2 fail. Every pre-existing test, including `efb862ce9`'s nested-real-tmux guard, still passes.
  **Artifact:** `.omo/evidence/20260727-cmux-detect-fake-tmux/red-1-detector.log`
  **Why sufficient:** proves the new tests actually pin the regressed behaviour rather than passing vacuously.

- **What was tested:** RED round 2 — detector fixed, `resolveCmuxCliExecutable` reverted to the hard-coded `"cmux"`.
  **Observed result:** `error: Executable not found in $PATH: "cmux"`, 10 pass / 4 fail.
  **Artifact:** `.omo/evidence/20260727-cmux-detect-fake-tmux/red-2-cli-resolution.log`
  **Why sufficient:** demonstrates why the CLI resolution change must ship with the detector change; shipping detection alone would be worse than the current placeholder behaviour.

- **What was tested:** affected package suites — `tmux-core`, `omo-opencode/src/shared/tmux`, `omo-opencode/src/tools/interactive-bash`, `omo-opencode/src/features/tmux-subagent`, `openclaw-core`.
  **Observed result:** 432 pass / 0 fail.
  **Artifact:** `.omo/evidence/20260727-cmux-detect-fake-tmux/affected-packages-test.log`
  **Why sufficient:** covers every package that imports `isCmuxCompatEnvironment` or the tmux runner.

- **What was tested:** real-harness QA on a live cmux host — `live-cmux-driver.ts` imports `spawnTmuxPane` / `closeTmuxPaneWithDependencies` from this branch's source and drives them against a live OpenCode server.
  **Observed result:** `isCmuxCompatEnvironment() = true`, cmux CLI resolved to the real binary, pane spawned, pane content contains no placeholder text, `opencode attach ... --session ...` observed in the process table, `closeTmuxPane -> true`, no leftover pane.
  **Artifact:** `.omo/evidence/20260727-cmux-detect-fake-tmux/live-cmux-driver.{ts,log}`
  **Why sufficient:** this is the end-to-end behaviour #5809 asked for, exercised through the real code path rather than mocks.

- **What was tested:** built artifact (`bun run build`) evaluated against the same live cmux environment.
  **Observed result:** `dist/index.js` reports `isCmuxCompatEnvironment() = true` and resolves the cmux CLI to the real binary, while still reporting native tmux for a real tmux nested inside a cmux pane.
  **Artifact:** `.omo/evidence/20260727-cmux-detect-fake-tmux/built-artifact-live-env.log`
  **Why sufficient:** confirms the shipped bundle, not just the source, behaves correctly.

- **What was tested:** the full protected matrix on CI, on the rebased head. This supersedes the earlier local whole-repo runs: those were made against a base that is now 2,659 commits behind, so their results no longer describe the merge target.
  **Observed result:** the protected matrix is green on the current head (`2694b530f`, **24 pass / 0 fail**, `mergeable_state: clean`), including the `ubuntu-latest` shard that failed on the pre-rebase head. Two repo-wide gates caught real defects of this branch on the way there, neither visible to a per-package sweep: `script/package-registration-audit.test.ts` (the new shim was unregistered) and `script/agent-command-string-audit.test.ts` (the new doc comment spelled the launch kind as an assignment, which the scanner reads as an agent command). Both are fixed and green.
  **Why sufficient:** this is the gate the branch actually has to clear, run against `dev` as it stands rather than reproduced locally against a stale base.

- **What was tested:** the precedence between the socket shape and `CMUX_AGENT_LAUNCH_KIND`, bounded from both sides. RED round 5 leaves the launch kind unreconciled; RED round 6 promotes it above the socket shape, which is the naive reading of the contract.
  **Observed result:** round 5 is 13 pass / 2 fail — exactly the one matrix row the reconciliation moves, plus the assertion isolating it. Round 6 is 12 pass / 3 fail — two nested-real-tmux rows and the fail-closed boundary.
  **Artifacts:** `red-5-launch-kind-fallback.log`, `red-6-launch-kind-precedence.log`
  **Why sufficient:** round 5 proves the launch kind is load-bearing rather than documented-but-inert; round 6 proves why it must stay subordinate to the socket shape.

- **What was tested:** whether the launch kind survives into a real tmux server started inside a cmux pane, which is what decides the precedence.
  **Observed result:** `CMUX_AGENT_LAUNCH_KIND=omo` and `CMUX_SOCKET_PATH` are byte-identical in the nested server's children; only `TMUX` changes, from the cmux shim socket to `/private/tmp/tmux-501/…`.
  **Artifact:** `live-signal-inheritance.log`
  **Why sufficient:** the ordering rests on a measurement on a real cmux host rather than on an assumption about the contract.

- **What was tested:** `has-session` exit-code parity, since `killTmuxSessionIfExists` reads a non-zero exit as "no such session" and only now routes through the compat layer.
  **Observed result:** `cmux __tmux-compat has-session` exits 1 for a missing session, the same as real tmux.
  **Artifact:** `cmux-compat-exit-codes.log`
  **Why sufficient:** confirms the guard keeps its meaning on a code path that was dormant before detection was fixed.

Full QA writeup: `.omo/evidence/20260727-cmux-detect-fake-tmux/QA.md` (§ Review round 4 covers the rebase and the launch-kind reconciliation)

## Risks & Residuals

- **Widening detection could capture a real tmux session** — mitigated. Detection still requires `CMUX_SOCKET_PATH`, and the socket path must contain a `cmux`/`cmux-`/`cmux.` path segment. `efb862ce9`'s nested-real-tmux test and a new explicit guard both cover this; a real tmux socket under `tmux-<uid>` is still reported as native tmux.
- **Flipping the flag enables previously dormant cmux branches** — this is the intent. Window/session isolation becomes native-tmux-only, session cleanup skips the fake `TMUX`, and authenticated cmux spawns fail closed, all exactly as #5811 designed. The affected package suites (449 tests) and the live driver cover these paths.
- **`findTmuxPath()` still probes a bare `cmux` on PATH** — accepted, unchanged by this PR. It falls back to a verified `tmux` path, which on cmux is the shim, and `resolveCmuxCliExecutable()` corrects the executable afterwards. A headless cmux with neither `cmux` nor a `tmux` shim on PATH would still stop at "tmux not found"; that path behaves the same before and after this change.
- **Whole-repo verification** — resolved, by CI rather than locally. Earlier rounds recorded local whole-repo failures and a Bun teardown segfault against the pre-rebase base; that base is 2,659 commits behind the merge target, so those observations describe the old baseline and not this PR. On the current head the full protected matrix is 24 pass / 0 fail and `mergeable_state` is `clean`; it has stayed green across the rebase, the owner's `dev` merge and the later 504-commit `dev` merge. The `omo setup credential inheritance` timeout that appeared in one run is untouched by this PR and was fixed independently by #7606, which is in the `dev` since merged into this branch.
- **Typecheck** — the CI `typecheck` and `format-lint-typecheck-build` jobs pass on the rebased head, and both packages this PR touches typecheck clean locally (`bunx tsgo --noEmit`, exit 0 each). The `preexisting-*.log` artifacts record host-local failures observed against the pre-rebase base and are kept only as history for the earlier review rounds; they are not claims about the current head.

## Automated Checks

Re-run on the current head, after merging `dev`:

```bash
bun install --frozen-lockfile                                  # exit 0
bunx tsgo --noEmit -p packages/tmux-core/tsconfig.json         # exit 0
bunx tsgo --noEmit -p packages/omo-opencode/tsconfig.json      # exit 0
bun run build                                                  # exit 0

bun test packages/tmux-core                                    # 121 pass / 0 fail
bun test packages/omo-opencode/src/shared/tmux                 #  89 pass / 0 fail
bun test packages/omo-opencode/src/tools/interactive-bash      #   3 pass / 0 fail
bun test packages/omo-opencode/src/features/tmux-subagent      # 168 pass / 0 fail
bun test packages/openclaw-core                                #  68 pass / 0 fail
bun test script/package-registration-audit.test.ts             #   6 pass / 0 fail
bun test script/agent-command-string-audit.test.ts             #   2 pass / 0 fail
```

The whole-repo run is now delegated to CI on the rebased head rather than reproduced locally against a stale base.

## Related Issues

Follow-up to #5811 (Residual scope). Completes the environment coverage for #5809. Orthogonal to #6297.

---
## Summary by cubic
Fixes cmux session detection and command routing. Previously, normal cmux sessions with an injected `TMUX` value were treated as native tmux and showed placeholder panes; now cmux-shaped sockets use cmux’s compat CLI, while real tmux—including nested sessions inside cmux—stays native.

- With `TMUX` set, detection requires a `cmux*` socket segment; `CMUX_SOCKET_PATH` authenticates it, and exact `CMUX_AGENT_LAUNCH_KIND=omo` can stand in when that credential is missing.
- Both runners resolve the cmux binary through `CMUX_OMO_CMUX_BIN`, `CMUX_BUNDLED_CLI_PATH`, then `cmux`, avoiding failures when only the tmux shim is on `PATH`.
- Replaced heuristic test fixtures and added coverage for socket shapes, inherited launch signals, Unix path separators, and CLI resolution.
- Reverified affected suites, typechecks, build, audit gates, and the live cmux harness after merging current `dev`.

<sup>Written for commit 2694b530f219336d33eed25f94cf1af6afdd3638. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6390?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Yoonkeee:
`typecheck (macos-latest)` failed with a `tsgo` crash, not a type error from this branch:

```
panic: Unhandled case in Node.StatementList: Kind(28783)
```

The same commit passes `typecheck (ubuntu-latest)`, `typecheck (windows-latest)` and `format-lint-typecheck-build`, and locally on macOS arm64 no package panics under `tsgo --noEmit`. Retriggered with the latest push.

Yoonkeee:
@cubic-dev-ai review this pleease.

MoerAI:
Thanks for the PR, @Yoonkeee, and for the unusually thorough root-cause work here.

I verified your analysis. `efb862ce9` narrowed the detector to `TMUX?.includes("cmuxterm") || (Boolean(CMUX_SOCKET_PATH) && !tmuxEnvironment)`, and since cmux always injects a `TMUX` value while `"cmuxterm"` only appears in the bundle id `com.cmuxterm.app` and never in a socket path, both branches are false on a normal cmux session. That does make every cmux-gated path from #5811 unreachable, which matches what you describe.

I am not merging this on my own judgment, because the change reverses the intent of a deliberate earlier fix and touches 24 files across the tmux and cmux detection surface. The right person to confirm the direction is the owner.

@code-yeongyu this one needs your call. The specific question: `efb862ce9` deliberately narrowed cmux detection so that a real tmux nested inside a cmux pane keeps talking to real tmux, and this PR replaces that heuristic with socket-path shape matching (`cmux*` directory versus `tmux-<uid>`). Is the socket-path shape the contract you want to rely on, and is the nested-real-tmux case still handled the way you intended? If you confirm the direction, I will run the merge.

code-yeongyu:
[sisyphus-bot] Detection logic and CLI resolution look correct and well-tested. The socket-path segment check properly distinguishes cmux from real tmux nested inside it. One question for the maintainer: the `hasCmuxSocketPath` function splits on both `/` and `\\` — is the backslash case needed for Windows cmux, or is it defensive? Also, consider whether `CMUX_OMO_CMUX_BIN` should take precedence over `CMUX_BUNDLED_CLI_PATH` in all cases or if there are scenarios where the bundled path is more reliable.

MoerAI (changes requested):
The architectural confirmation is still missing, and two concrete review issues remain in the current diff: Unix socket paths must split only on `/` rather than treating backslash as a separator, and the QA driver must assert the expected result and exit nonzero on failure instead of only logging. Please address both and confirm the intended nested-real-tmux behavior before re-review.

Yoonkeee:
Both review issues are fixed, and I have answers to @code-yeongyu's two questions — one of which turned up a third false positive in the same function, which I also fixed here.

Pushed as `a3d84e139`, `7b377e733`, `045534ec8`, `50508949f`, `717257988`.

---

## 1. Unix-only path separator (cubic P3, and @MoerAI's first item)

Fixed in `a3d84e139`. `socketPath.split(/[\\/]/)` → `socketPath.split("/")`.

This also answers **@code-yeongyu's question — "is the backslash case needed for Windows cmux, or is it defensive?"**: it is neither. tmux does not run on Windows, and on Unix `\` is an ordinary filename character, so treating it as a separator could only ever *widen* detection. A real tmux socket at `/private/tmp/tmux-501/weird\cmux-omo` was split into `weird` + `cmux-omo`, matched the cmux segment pattern, and would have routed a genuine tmux session through `cmux __tmux-compat` — a false positive in exactly the direction this PR exists to prevent.

RED (`red-3-backslash-separator.log`): with the pre-change detector, only the new guard fails — 7 pass / 1 fail. Every pre-existing test, including `efb862ce9`'s nested-real-tmux guard, still passes, so narrowing the split does not disturb either side of the discriminator.

The doc comment now records that the Unix-only split is deliberate, so the next reader does not "restore" it.

## 2. The live QA driver could not fail (cubic P2, and @MoerAI's second item)

Fixed in `7b377e733`. The driver logged the three observations it was cited for — placeholder text in the pane, `opencode attach` in the process table, and the `closeTmuxPane` result — and then exited 0 regardless of their values, so re-running it produced no red/green signal. That was a fair hit: the artifact was quoted as end-to-end evidence while being incapable of reporting failure.

The three are now collected into explicit failure conditions that exit non-zero. One detail worth flagging: the check runs **after** `closeTmuxPaneWithDependencies`, not at the observation site. Asserting at the observation site would skip cleanup and leave a live cmux pane behind on every failure.

## 3. Nested-real-tmux behaviour (@MoerAI's confirmation request)

Unchanged from `efb862ce9`'s intent, and now pinned from both directions. `CMUX_SOCKET_PATH` remains a precondition; the discriminator is the socket path shape.

| `CMUX_SOCKET_PATH` | `TMUX` | Result |
| --- | --- | --- |
| `/tmp/cmux.sock` | `/private/tmp/tmux-501/default,123,0` | `false` — native tmux nested in cmux |
| `/tmp/cmux.sock` | `/private/tmp/tmux-501/weird\cmux-omo,123,0` | `false` — backslash is not a separator |
| `/tmp/cmux.sock` | `/private/tmp/tmux-501/cmuxterm-notes,123,0` | `false` — see section 5 |
| `/Users/…/cmux-501.sock` | `/tmp/cmux-omo/<workspace>,<surface>,<pane>` | `true` — cmux |
| unset | `/tmp/cmux-omo/workspace,surface,pane` | `false` — no cmux socket |

A real tmux nested inside a cmux pane keeps talking to real tmux. That was the point of `efb862ce9` and it still holds — the change is only in *how* the two are told apart, from "is `TMUX` set" to "what shape is the socket path".

## 4. `CMUX_OMO_CMUX_BIN` vs `CMUX_BUNDLED_CLI_PATH` (@code-yeongyu's second question)

> should `CMUX_OMO_CMUX_BIN` take precedence over `CMUX_BUNDLED_CLI_PATH` in all cases, or are there scenarios where the bundled path is more reliable?

`CMUX_OMO_CMUX_BIN` first is correct, and not by preference — it matches cmux's own contract. cmux's tmux shim resolves the binary like this:

```sh
exec "${CMUX_OMO_CMUX_BIN:-cmux}" __tmux-compat "$@"
```

So `CMUX_OMO_CMUX_BIN` → bare `cmux` is precisely the order the shim uses, and `resolveCmuxCliExecutable()` reproducing it means we delegate to the same binary the shim would have. `CMUX_BUNDLED_CLI_PATH` is not part of that path at all; inside cmux it serves SSH relay, reconnect, and the claude/codex wrappers, and it can point at `$HOME/.cmux/bin/cmux` rather than into the app bundle. That makes it a reasonable last-resort fallback but the wrong first choice for tmux delegation.

On this host both happen to resolve to `/Applications/cmux.app/Contents/Resources/bin/cmux`, so the ordering is unobservable in practice today — but if they ever diverge, the shim's variable is the one that is right by construction.

## 5. Additional finding: the `cmuxterm` branch was a third false positive

While verifying the backslash fix I checked whether the same class of bug existed elsewhere in the function. It did, and it was larger. Fixed in `045534ec8`.

`isCmuxCompatEnvironment` opened with:

```ts
if (tmuxEnvironment?.includes("cmuxterm") === true) return true
```

This returned true **before** the `CMUX_SOCKET_PATH` precondition, so any tmux session whose name contained `cmuxterm` was reported as cmux. Unlike the backslash case, session names are user-chosen, so this is reachable in ordinary use.

I had assumed this branch was legacy support worth keeping — the PR description says as much. Checking the shipped binary shows there is nothing to keep compatibility with. `strings /Applications/cmux.app/Contents/Resources/bin/cmux` yields 17 distinct `/tmp/cmux*` strings:

| Purpose | Observed |
| --- | --- |
| socket paths present as literals | `/tmp/cmux.sock`, `/tmp/cmux-cli-shims`, `/tmp/cmux-cloud-cli.sock`, `/tmp/cmux-debug-`, `/tmp/cmux-debug.sock`, `/tmp/cmux-nightly-`, `/tmp/cmux-nightly.sock`, `/tmp/cmux-ssh-`, `/tmp/cmux-staging-`, `/tmp/cmux-staging.sock`, `/tmp/cmux-wait-for-`, `/tmp/cmux-xctest-` |
| observed at runtime on this host | `/tmp/cmux-omo/<workspace>` — composed at runtime, so not a literal in the binary |
| everywhere `cmuxterm` appears | `com.cmuxterm.app` (bundle id), `~/.cmuxterm/…` (config dir), `CMUXTERM_CLI_RESPONSE_TIMEOUT_SEC` (env name), `_TtC12CmuxTerminal…` (Swift symbols), `CMUXTERMINFO` (heredoc marker) — never a path |

Every socket path is `cmux-` or `cmux.` prefixed; `cmuxterm` never appears in one. The branch was introduced in `8236d7d6b` as the second half of `Boolean(CMUX_SOCKET_PATH) || TMUX?.includes("cmuxterm")` — a backup heuristic resting on the assumption that cmux writes its own name into `TMUX`. It does not, so the branch never matched a real cmux session, while it did mislabel real tmux ones.

The evidence that it was redundant rather than load-bearing is in the RED log: the release-channel guard I added **passed before the branch was removed** (`red-4-cmuxterm-branch.log` — only the two `cmuxterm` false-positive guards fail there). The socket pattern `/^cmux([-.]|$)/` already covers every shape cmux can produce, on its own. `717257988` widens that guard from the five channels I first listed to all twelve literal shapes above, including the socket-file forms such as a bare `cmux.sock`.

`CMUX_SOCKET_PATH` is now the single precondition and the socket path shape the only discriminator — which is the rule this PR's own doc comment already stated. The `cmuxterm` branch was the last exception to it.

Four tests used `TMUX=/tmp/cmuxterm-test.sock` purely to enter the cmux path (`pane-auth-cmux.test.ts` ×2, `manager-cmux-eligibility.test.ts`, `manager.test.ts`). That is a shape real cmux never produces, so they were asserting against a fictional environment; they now use `/tmp/cmux-omo/workspace,surface,pane` with `CMUX_SOCKET_PATH` set. `manager.test.ts` additionally had to stop deleting `CMUX_SOCKET_PATH`.

If you would rather keep this PR scoped to the detector fix, I am happy to split section 5 into a follow-up — the two changes are independent and it reverts cleanly as `045534ec8` + `50508949f`.

---

## Verification

```
packages/tmux-core                                 110 pass / 0 fail
packages/omo-opencode/src/shared/tmux               89 pass / 0 fail
packages/omo-opencode/src/tools/interactive-bash     3 pass / 0 fail
packages/omo-opencode/src/features/tmux-subagent   166 pass / 0 fail
packages/openclaw-core                              67 pass / 0 fail
                                                   435 pass / 0 fail
```

`bunx tsgo --noEmit` exits 0 for `packages/tmux-core` and `packages/omo-opencode`.

On a live cmux host (`TMUX=/tmp/cmux-omo/EE5868C2-…`, `CMUX_SOCKET_PATH=~/.local/state/cmux/cmux-501.sock`) the patched detector still returns `true`, and the nested-real-tmux, backslash, `cmuxterm`-named-session, `cmux-nightly`, `cmux-staging`, and `TMUX`-not-injected cases all resolve correctly — so the two narrowings cost nothing on the detection side.

New artifacts under `.omo/evidence/20260727-cmux-detect-fake-tmux/`: `red-3-backslash-separator.log`, `review-round-1-green.log`, `red-4-cmuxterm-branch.log`, `review-round-2-green.log`, with the reasoning written up in `QA.md`.

MoerAI (approved):
Re-reviewed the current head against all three blockers from my prior change request. Unix socket parsing now splits only on `/`, the live driver asserts its observations and exits nonzero on failure, and nested real-tmux socket shapes remain explicitly excluded with regression coverage. The affected suites are green; the remaining macOS/Windows root-test instability is documented as reproducible on clean `dev` and is outside this diff. My code-review blockers are resolved.

MoerAI (approved):
Re-reviewed the three commits after my earlier approval. Product code is unchanged. The only test delta replaces a platform-dependent `path.join` fixture with the literal POSIX socket shape cmux actually emits; the other changes only record CI evidence. All current checks are green, so my prior code-review approval remains valid on this head.

MoerAI (changes requested):
Thanks for the unusually thorough cmux root-cause analysis, deterministic socket-shape coverage, and committed real-surface evidence. The prior review blockers remain resolved.

Please rebase onto current `dev`. This head is now conflicting and 2,659 commits behind, so its historical green checks cannot satisfy the current protected matrix:

Preserve the reviewed invariants: resolve the CLI through `CMUX_OMO_CMUX_BIN`, split Unix socket paths only on `/`, exclude nested real-tmux socket shapes, and retain authenticated-cmux fail-closed behavior.

During the rebase, reconcile socket-shape detection with the official `CMUX_AGENT_LAUNCH_KIND=omo` contract rather than leaving independent detection paths with undefined precedence. Add explicit precedence and fallback tests for both signals.

Then rerun current CI and refresh the isolated real cmux/OpenCode evidence under `.omo/evidence/` on the rebased artifact. Re-request owner review once mergeability is clean.

Yoonkeee:
Rebased onto current `dev` and reconciled the launch-kind contract. Head is now `e6007ce28`.

## The issue is still live on `dev`

Checked first, since 2,659 commits is long enough for a fix to land independently. `origin/dev` still carries the regressed expression verbatim:

```ts
const tmuxEnvironment = environment.TMUX
return tmuxEnvironment?.includes("cmuxterm") === true ||
	(Boolean(environment.CMUX_SOCKET_PATH) && !tmuxEnvironment)
```

The only commit to touch `cmux-detect.ts` since this branch forked is `7da0e4d3c` (`fix(tmux): support inline cmux pane lifecycle`), which parameterises `process.env` for testability and leaves the condition intact. Evaluated against a live cmux omo pane it still returns `false`:

```
CMUX_AGENT_LAUNCH_KIND = omo
CMUX_SOCKET_PATH       = /Users/…/.local/state/cmux/cmux-501.sock
TMUX                   = /tmp/cmux-omo/97BEE327-…,EEC79E0A-…,1632462732908491017

dev detector  -> false
this branch   -> true
```

## Rebase

14 of 15 commits replayed with one conflict, in `docs/reference/re-export-shim-inventory.md`. `dev` had regenerated that inventory (317 → 256 shims), so the original registration commit reduced to an empty diff and was dropped; the shim is re-registered against the current tree with counts regenerated from the scan command the document itself specifies (257 total, `tmux-core` 3 → 4). No source file this PR touches had drifted on `dev`.

## Reconciling socket shape with `CMUX_AGENT_LAUNCH_KIND=omo`

The measurement that decides the order: starting a real tmux server **inside** a cmux pane passes both `CMUX_AGENT_LAUNCH_KIND=omo` and `CMUX_SOCKET_PATH` through to its children byte-identically, and only `TMUX` changes to the real socket.

```
cmux omo pane                       nested real tmux
  CMUX_AGENT_LAUNCH_KIND=omo          CMUX_AGENT_LAUNCH_KIND=omo      <- inherited
  CMUX_SOCKET_PATH=…/cmux-501.sock    CMUX_SOCKET_PATH=…/cmux-501.sock <- inherited
  TMUX=/tmp/cmux-omo/…                TMUX=/private/tmp/tmux-501/…     <- only this changes
```

So the launch kind describes the *launcher*, while the socket path describes the *server a tmux command would actually reach*. Precedence, highest first:

1. the `TMUX` socket shape — sole discriminator, vetoes both env signals;
2. `CMUX_SOCKET_PATH` — authenticated credential, sufficient once 1 has not vetoed;
3. `CMUX_AGENT_LAUNCH_KIND === "omo"` — accepted only as a stand-in for 2, and only where 1 confirms a cmux-shaped socket. Never sufficient alone, so an environment carrying only the launch kind fails closed to native tmux.

Matched exactly against `omo`, because cmux names the launched agent in this variable and ships a wrapper shim per agent, so only `omo` carries the `CMUX_OMO_CMUX_BIN` contract `resolveCmuxCliExecutable()` depends on.

Against the previous head this moves exactly **one** of the eighteen signal combinations — `CMUX_SOCKET_PATH` absent, kind `omo`, `TMUX` cmux-shaped — reachable when a shell profile scrubs `CMUX_*` while the cmux tmux shim still injects `TMUX`.

Both directions are pinned by RED rounds:

| round | change | result |
|---|---|---|
| `red-5-launch-kind-fallback.log` | launch kind left unreconciled | 13 pass / 2 fail — the one matrix row plus the assertion isolating it, so the signal is load-bearing rather than inert |
| `red-6-launch-kind-precedence.log` | launch kind promoted **above** the socket shape | 12 pass / 3 fail — two nested-real-tmux rows and the fail-closed boundary, which is why it is subordinate |

## Reviewed invariants

| invariant | status |
|---|---|
| resolve the CLI through `CMUX_OMO_CMUX_BIN` | unchanged; live driver resolves to the real binary |
| split Unix socket paths only on `/` | unchanged; backslash guard still green |
| exclude nested real-tmux socket shapes | strengthened — now proven to outrank *both* env signals (matrix rows 3 and 12) |
| authenticated-cmux fail-closed | preserved; launch-kind-only returns `false` (row 10) |

## Checks before touching the detector

- Enumerated every non-test reader of `CMUX_SOCKET_PATH`, since detection can now be true without it. There are two: the detector, and a debug log object in `pane-spawn.ts` that reads it optionally. Nothing dereferences it as required.
- Re-checked `cmuxterm` against the shipped app rather than assuming: `~/.cmuxterm/` holds config and state and contains no sockets (`find -type s` empty); the live socket is `~/.local/state/cmux/cmux-501.sock`. Both live shapes match `/^cmux([-.]|$)/`, so the pattern needs no widening.
- `killTmuxSessionIfExists` reads a non-zero `has-session` exit as "no such session" and only now routes through the compat layer, so the convention was measured: `cmux __tmux-compat has-session` exits 1 for a missing session, same as real tmux (`cmux-compat-exit-codes.log`).

## Verification on the rebased artifact

```
packages/tmux-core                                  115 pass / 0 fail
packages/omo-opencode/src/shared/tmux                89 pass / 0 fail
packages/omo-opencode/src/tools/interactive-bash      3 pass / 0 fail
packages/omo-opencode/src/features/tmux-subagent    168 pass / 0 fail
packages/openclaw-core                               67 pass / 0 fail
script/package-registration-audit.test.ts             6 pass / 0 fail

bunx tsgo --noEmit -p packages/tmux-core/tsconfig.json      exit 0
bunx tsgo --noEmit -p packages/omo-opencode/tsconfig.json   exit 0
bun run build                                               exit 0
```

Real-surface evidence re-recorded on this head, not carried over:

- `live-cmux-driver.log` — driver re-run against the cmux-launched OpenCode server (`cmux omo --port 64155`). Detection true, cmux CLI resolved to the real binary, pane spawned without the placeholder, `opencode attach` observed in the process table, pane closed, exit 0.
- `built-artifact-live-env.log` — `bun run build`, then the detector driven straight out of `dist/index.js`. All eight guards hold in the shipped bundle, including nested-real-tmux.
- `live-signal-inheritance.log` — the inheritance measurement above.

One test-hygiene fix came out of this: `cmux-detect.test.ts` now clears `CMUX_AGENT_LAUNCH_KIND` in `beforeEach` alongside the other two variables. Without it the suite reads the host's cmux configuration and the "cmux socket directory without `CMUX_SOCKET_PATH`" guard inverts on any cmux machine.

Full writeup in `.omo/evidence/20260727-cmux-detect-fake-tmux/QA.md` § Review round 4. CI is running on the rebased head.

Yoonkeee:
Follow-up on the CI rerun, and thanks for merging `dev` in.

**The first post-rebase push caught a real defect of mine.** On `e6007ce28` the root suite failed one shard, and it was this branch's own doc comment: `script/agent-command-string-audit.test.ts` scans tracked sources for uncategorized agent command strings, and I had spelled the launch kind as an assignment, which the scanner reads as one.

```
+   "packages/tmux-core/src/cmux-detect.ts: =omo",
- Expected  - 0
+ Received  + 1
```

Fixed in `66f651f30` by dropping the value from that one sentence rather than widening the allowlist — the sentence is about the variable being *inherited*, and the value is already carried by `CMUX_OMO_LAUNCH_KIND` and its own doc comment. Like the shim-inventory audit in the original round, this is a gate the per-package sweep cannot see, which is the argument for running the full matrix rather than trusting local package runs.

**On `66f651f30` the full protected matrix was 25 pass / 0 fail**, including the `ubuntu-latest` shard that failed on the pre-rebase head, and `mergeable_state` was `clean`.

The same run also reported `omo setup credential inheritance` in `packages/omo-native` as a 5000 ms timeout. I recorded it as unrelated — that file has no tmux/cmux reference and is untouched by this PR — and #7606 in the `dev` you merged confirms it independently.

**After your merge (`244a955e1`) I re-verified locally rather than assuming the merge was inert:**

```
packages/tmux-core                                  115 pass / 0 fail
packages/omo-opencode/src/shared/tmux                89 pass / 0 fail
packages/omo-opencode/src/tools/interactive-bash      3 pass / 0 fail
packages/omo-opencode/src/features/tmux-subagent    168 pass / 0 fail
packages/openclaw-core                               67 pass / 0 fail
script/package-registration-audit.test.ts             6 pass / 0 fail
script/agent-command-string-audit.test.ts             2 pass / 0 fail
```

CI on the merge commit is queued at the time of writing, with no failures so far.

Every item from the change request is addressed: rebased and `mergeable: true`; all four reviewed invariants preserved, with nested-real-tmux now proven to outrank *both* env signals; the launch-kind contract reconciled with explicit precedence and an eighteen-row matrix plus individually asserted fallback, veto, fail-closed and exact-match rows; CI rerun; and the real cmux/OpenCode evidence re-recorded on the rebased artifact rather than carried over. Details and the RED rounds are in the previous comment and `QA.md` § Review round 4.

Re-requesting review.

Yoonkeee:
Brought the branch up to date with current `dev` (504 commits since the last sync) and re-verified on the merged tree. Head is `2694b530f`.

Merged rather than rebased, matching the merge you had already made onto this branch, so no published commit is rewritten.

**The merge was conflict-free, and nothing this PR touches had drifted.** `git log <prev-head>..origin/dev` is empty for `cmux-detect.ts`, `cmux-detect.test.ts`, `cmux-cli.ts`, `runner.ts`, `index.ts`, `tools.ts`, the omo-opencode tmux barrel and the shim inventory — so the detector is byte-identical to what was reviewed.

Re-run on the merged tree rather than assumed inert:

```
packages/tmux-core                                  121 pass / 0 fail
packages/omo-opencode/src/shared/tmux                89 pass / 0 fail
packages/omo-opencode/src/tools/interactive-bash      3 pass / 0 fail
packages/omo-opencode/src/features/tmux-subagent    168 pass / 0 fail
packages/openclaw-core                               68 pass / 0 fail
script/package-registration-audit.test.ts             6 pass / 0 fail
script/agent-command-string-audit.test.ts             2 pass / 0 fail

bunx tsgo --noEmit -p packages/tmux-core/tsconfig.json      exit 0
bunx tsgo --noEmit -p packages/omo-opencode/tsconfig.json   exit 0
bun run build                                               exit 0
```

Both repo-wide gates still hold — the shim inventory keeps `cmux-cli.ts` registered at 257 exact shims, and the agent command audit stays clean.

`live-cmux-driver.log` was re-recorded on this head rather than carried forward: detection true, cmux CLI resolved to the real binary, pane spawned without the placeholder, `opencode attach` observed in the process table, pane closed, exit 0. The built `dist/index.js` still carries the reconciled detector verbatim.

**CI is 24 pass / 0 fail and `mergeable_state` is `clean`** — green across the rebase, your `dev` merge, and this one.

I also corrected the PR description, which was still describing the pre-rebase base: it claimed the whole-repo run and `typecheck` could not pass locally and cited two pre-existing failures. Those observations belonged to a base 2,659 commits behind and no longer describe this branch, and cubic had been citing them as the reason it could not certify regression freedom. The description now points at the CI result instead, and the `preexisting-*.log` artifacts are marked as history for the earlier rounds rather than claims about the current head.

Nothing else changed. @MoerAI when you have a moment — the four blockers from your change request are addressed as described in the earlier comment, and mergeability is clean.

PR #6264, opened 2026-07-21
fix(background-agent): sanitize platform prefix in checkSessionExistence ID (#5285, #6263)

niStee:
## Summary

Extends the fix from #5285 (`start-work-hook`) to `checkSessionExistence()` in the background task poller. Fixes #6263.

## Problem

`boulder.json` persists session IDs with platform prefixes (`opencode:ses_...`). When `checkSessionExistence()` forwards this verbatim to `client.session.get({ path: { id: sessionID } })`, the URL becomes `/session/opencode%3Ases_...`, which the OpenCode API server rejects with HTTP 400 (`Expected a string starting with "ses"`).

This causes `checkSessionExistence()` to return `"unknown"` instead of `"exists"`, misclassifying active subagent sessions as `sessionGone` and triggering the 10 s `sessionGoneTimeoutMs` — prematurely aborting in-progress background tasks.

## Fix

Add `sanitizeBareSessionId()` helper that strips `opencode:`, `codex:`, and `senpi:` prefixes before the SDK call. Use `rawId` in the `client.session.get` path.

## Verification

- `bun test packages/omo-opencode/src/features/background-agent/session-existence.test.ts` — 3 pass
- `bun run typecheck` — clean
- `bun run build` — clean
- Full pre-push suite — 11 960 pass, 0 fail

---
## Summary by cubic
Sanitizes platform-prefixed session IDs in background agent `checkSessionExistence()` to avoid OpenCode API 400s and prevent premature task aborts (fixes #6263).

- **Bug Fixes**
  - Added `sanitizeBareSessionId()` to strip `opencode:`, `codex:`, and `senpi:` prefixes from IDs persisted in `boulder.json`.
  - Use sanitized `rawId` in `client.session.get({ path: { id } })` so requests hit `/session/ses_...` and active sessions resolve correctly.

<sup>Written for commit 5088e004be427f7c5cdc7ab7b15bf4cae6a7e202. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #6263, opened 2026-07-21
bug(background-agent): checkSessionExistence passes opencode:-prefixed storage ID to SDK client.session.get(), causing HTTP 400 fetch error


niStee:
## Bug Description

When `checkSessionExistence()` queries `client.session.get()` in `packages/omo-opencode/src/features/background-agent/session-existence.ts`, it forwards the `sessionID` string as-is.

When `sessionID` originates from `boulder.json` storage (which persists IDs formatted as `opencode:ses_...`), the SDK path parameter becomes `/session/opencode%3Ases_...`.

The OpenCode API server validates that `:id` starts with `ses_`, returning an HTTP 400 validation error:
`Expected a string starting with "ses", got "opencode%3Ases_..."`

This causes `checkSessionExistence()` to return `"unknown"`, misclassifying active subagent sessions as `sessionGone` and triggering short emergency timeouts (`sessionGoneTimeoutMs` / `timeout_seconds: 10`), prematurely aborting running background tasks.

## Root Cause

`checkSessionExistence` passes the un-sanitized `sessionID` (which carries `opencode:` / `codex:` / `senpi:` prefix from `boulder.json`) to `client.session.get({ path: { id: sessionID } })`.

This is the same class of issue previously fixed in `start-work-hook.ts` under #5285.

## Fix

Sanitize `sessionID` using `sanitizeBareSessionId(sessionID)` (stripping `opencode:`, `codex:`, and `senpi:` prefixes) before constructing the SDK path payload.

niStee:
### High-Severity Impact Note

In production environments, this bug causes the background task poller to classify active subagent sessions as  every poll cycle (~10-18s). 

Because the parent task manager attempts to recover by re-sending the full turn context (prompt + history + workspace state ~75k-100k tokens) to a newly spawned subagent turn, this produces a rapid runaway retry loop. In our logs, a single task processed over **6.09 million input tokens in ~50 minutes** across 81 duplicate turns before exhausting provider rate limits / subscription quota.

code-yeongyu:
[sisyphus-bot] Thanks for the clear root cause analysis. Confirmed: checkSessionExistence in session-existence.ts passes the sessionID as-is to client.session.get(). There is no sanitizeBareSessionId function in the codebase. If a prefixed session ID from boulder.json reaches this path, the SDK URL would be malformed. The fix is to strip the opencode:/codex:/senpi: prefix before constructing the SDK path.

[...]

PR #6430, opened 2026-07-28
fix(background-agent): sanitize platform-prefixed session ids before existence check

niStee:
## What

Rebased the session-ID prefix workaround onto v4.19.3. `boulder.json` stores IDs as `opencode:ses_*`; forwarding them to `client.session.get()` URL-encodes the colon and makes OpenCode return HTTP 400. The poller then treats the live session as gone after its grace window.

`checkSessionExistence()` now strips known harness prefixes (`opencode:`, `codex:`, `senpi:`) before the SDK lookup. Bare IDs are unchanged.

This is the same focused fix as #6264, retargeted onto the current release source so downstream users can move to 4.19.3 (including the new omo.jsonc config migration) without reintroducing the sessionGone regression.

## Tests

- `bun test packages/omo-opencode/src/features/background-agent/session-existence.test.ts`: 5 pass, 0 fail
- Added cases for all three known prefixes, bare IDs, and the exact ID passed to `client.session.get()`

---
## Summary by cubic
Sanitize platform-prefixed session IDs before existence checks to prevent 400s from URL-encoded colons and stop false “session gone” states in the background agent. Reapplies the fix on v4.19.3 so users can upgrade without regressions.

- **Bug Fixes**
  - Strip `opencode:`, `codex:`, `senpi:` prefixes before the SDK lookup; bare IDs unchanged.
  - `checkSessionExistence` now calls `client.session.get()` with the bare ID; adds `sanitizeBareSessionId`.
  - Tests cover prefixed and bare IDs, plus the exact ID sent to the SDK.

<sup>Written for commit 9429090071a436ca985ce9349717b7556603fda2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/session-existence.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/session-existence.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, mock, test } from "bun:test"
 
 import type { OpencodeClient } from "./opencode-client"
-import { checkSessionExistence, verifySessionExists } from "./session-existence"
+import { checkSessionExistence, sanitizeBareSessionId, verifySessionExists } from "./session-existence"
 import { unsafeTestValue } from "../../../../../test-support/unsafe-test-value"
 
 describe("verifySessionExists", () => {
@@ -39,5 +39,38 @@ describe("verifySessionExists", () => {
     const result = await checkSessionExistence(client, "session-123")
 
     expect(result).toBe("unknown")
+  })
+})
+
+describe("sanitizeBareSessionId", () => {
+  test("strips known platform prefixes", () => {
+    // given / when / then
+    expect(sanitizeBareSessionId("opencode:ses_abc123")).toBe("ses_abc123")
+    expect(sanitizeBareSessionId("codex:ses_abc123")).toBe("ses_abc123")
+    expect(sanitizeBareSessionId("senpi:ses_abc123")).toBe("ses_abc123")
+  })
+
+  test("leaves bare session ids untouched", () => {
+    // given / when / then
+    expect(sanitizeBareSessionId("ses_abc123")).toBe("ses_abc123")
+  })
+
+  test("checkSessionExistence queries with the bare id", async () => {
+    // given
+    const get = mock(async () => ({ data: { id: "ses_abc123" } }))
+    const client = unsafeTestValue<OpencodeClient>({
+      session: {
+        get,
+      },
+    })
+
+    // when
+    const result = await checkSessionExistence(client, "opencode:ses_abc123")
+
+    // then
+    expect(result).toBe("exists")
+    expect(get).toHaveBeenCalledWith({
+      path: { id: "ses_abc123" },
+    })
   })
 })
--- a/packages/omo-opencode/src/features/background-agent/session-existence.ts
+++ b/packages/omo-opencode/src/features/background-agent/session-existence.ts
@@ -36,14 +36,19 @@ function isSessionNotFoundError(error: unknown): boolean {
   return message.includes("not found") || message.includes("missing")
 }
 
+export function sanitizeBareSessionId(sessionId: string): string {
+  return sessionId.replace(/^(opencode|codex|senpi):/, "")
+}
+
 export async function checkSessionExistence(
   client: OpencodeClient,
   sessionID: string,
   directory?: string
 ): Promise<SessionExistenceStatus> {
+  const rawId = sanitizeBareSessionId(sessionID)
   try {
     const response = await client.session.get({
-      path: { id: sessionID },
+      path: { id: rawId },
       ...(directory ? { query: { directory } } : {}),
     })
 


niStee:
Friendly status check — the underlying bug from #6263 is still present in current `dev`: `sanitizeBareSessionId` has zero hits in the repo, and `checkSessionExistence()` in `session-existence.ts` still forwards the raw `boulder.json` ID (including the `opencode:` prefix) into `client.session.get()`, producing the HTTP 400 and the false `sessionGone` abort path.

This PR is the same focused fix as the closed #6264, rebased onto the v4.19.3 release source so users can upgrade without reintroducing the regression. 5/5 tests passing. Happy to rebase again if `dev` has moved since — just flag it.

MoerAI:
Thanks for the PR, @niStee.

I read the change. `sanitizeBareSessionId()` strips a leading `opencode:`, `codex:` or `senpi:` before `client.session.get({ path: { id } })`, the regex is anchored so a bare id passes through untouched, and there are tests for all three prefixes plus the integration call. Mechanically it is clean and narrow.

What I want the owner to decide is the layer, not the code. Stripping inside `checkSessionExistence` fixes the symptom at the last hop, but it also means the platform prefix is treated as something callers may legitimately pass into a session lookup. The alternative is that prefixed ids are a storage and bookkeeping concern only (Boulder records sessions as `codex:<session_id>`) and should be unwrapped by the caller before reaching the client. Those two designs age very differently once more harnesses land.

@code-yeongyu two questions for you on this one. First, is `sanitizeBareSessionId` at the client boundary the layer you want, or should callers pass bare ids and keep the prefix confined to storage? Second, is `opencode|codex|senpi` the canonical prefix set to hardcode here, given the multi-harness work in progress? Once you confirm, I will merge or relay the change request.

code-yeongyu:
[sisyphus-bot] REVIEW-OK. The fix is mechanically correct and well-tested; CI green. Owner should weigh the layering note from MoerAI (strip at storage write vs. read site) — if the prefix is also forwarded to other SDK calls, a storage-boundary fix may be preferable.

niStee:
Replying to MoerAI's layering questions (so this is captured before merge):

**1. Strip at client boundary vs. storage write?**

Client-boundary strip is the right choice *for this PR*. Justification:

- `checkSessionExistence()` is the only call site that forwards a sessionID into `client.session.get({ path: { id } })` — that's the one place where the URL-encoded prefix becomes HTTP 400.
- The prefix is a *storage and bookkeeping* concern (`boulder.json` records `opencode:ses_…`) but the SDK URL contract is also narrow: it expects `ses_…`, not the prefixed form. Other SDK boundaries I checked (`session.get`, `session.messages`, `session.abort`, etc.) all take a bare ID at the URL level too.
- A storage-write strip would force every harness adapter to drop the prefix before persistence, which would break audit trails and Boulder cross-harness linking.
- The exported `sanitizeBareSessionId` is a one-line helper — if more SDK boundaries surface that need the strip, calling it is cheap. Future harness support should also call this helper, not duplicate the regex.

So: keep the strip at the SDK boundary, centralize it as `sanitizeBareSessionId`, and add it to the harness onboarding checklist as the canonical place to strip.

**2. Is `opencode|codex|senpi` the canonical prefix set?**

Yes for the current harnesses. These match the harness adapters currently shipped:
- `opencode:` → omo-opencode
- `codex:` → omo-codex
- `senpi:` → omo-senpi

The regex is hardcoded rather than configurable because the prefix set *is* the set of registered harnesses — adding a new harness is a code change anyway (it adds a new adapter), so a code-side list stays correct. A config-driven list would only paper over the moment of forgetting to update both lists.

Happy to extract the prefix set into a named export (`SUPPORTED_HARNESS_PREFIXES`) if the maintainer wants it discoverable for downstream code, but I'd argue against making it runtime-configurable.

niStee:
Status check on the latest dev-base CI run — all checks green against current `dev` (`c26b8febd`):

- `codex-compatibility (windows/macos/ubuntu)` — pass
- `test (windows/macos/ubuntu)` — pass (749/749 in the background-agent suite locally, 0 failures)
- `typecheck (windows/macos/ubuntu)` — pass
- `senpi-compatibility (windows/macos/ubuntu)` — pass
- `build`, `cla`, `omo-ai-payload-check`, `lazycodex-published-smoke`, `GitGuardian` — pass

Single-commit diff, +34/-1 in `session-existence.test.ts` and +6/-1 in `session-existence.ts`. No conflicts on current `dev` (merge-base is `614cc5358`, only one upstream commit (`b3b45cd2 feat(model-core): migrate Mini chains to Luna and DeepSeek`) touches the `background-agent/` directory since, and it's unrelated to session-ID handling).

The Aug 4 layering note from MoerAI was answered in the Aug 9 reply (keeping `SUPPORTED_HARNESS_PREFIXES` as a code-side set, not runtime config). Happy to extract the prefix list into a named export if that improves discoverability for downstream code — say the word and I'll push an amendment.

Ready for re-review whenever convenient.

MoerAI:
Thanks for the persistence on this, @niStee.

I am closing this as superseded rather than rejected. #7265 carries the same SDK-boundary prefix stripping for `opencode:`, `codex:`, and `senpi:` session IDs on an in-repository branch with the current full CI matrix.

Your #6263, #6264, and #6430 lineage established both the bug and the correct layering: strip before the SDK URL path, not when persisting cross-harness state. The replacement should preserve explicit credit to that work before it lands and closes #6263.

PR #6451, opened 2026-07-29
fix(senpi): harden Windows Team startup and crash recovery

### LINKED ISSUE #6449: [Bug]: Windows Senpi Team startup and crash recovery fail across launcher and liveness paths
## Prerequisites

- [x] I searched existing issues and PRs, including #5798 and #6374.
- [x] I reproduced this on oh-my-openagent v4.19.3 (`614cc5358dc393153fc39acae74dc5bd9fb9fffc`).
- [x] I read the Team/Senpi source and tested a candidate fix locally.

## Bug Description

On Windows, the Senpi Team path can fail at several connected boundaries:

1. an npm-installed `senpi` launcher is spawned as if it were a native executable;
2. the same launcher assumption is repeated by the real `senpi-task` RPC child path;
3. crash QA assumes POSIX negative-PID process-group semantics;
4. a terminal member record can survive a process crash while its lead liveness notification is not durably replayed;
5. the crash gate originally depended on one human-readable `lost` sentence instead of the structured liveness event.

The user-visible result starts as Team members failing to start. After fixing only the launcher, forced-kill/restart can still exit without delivering the previous member's abnormal termination to the restarted lead.

This is a deterministic runtime/logic failure, not the timing-only CI flake tracked in #6374.

## Steps to Reproduce

Environment:

- Windows 10
- oh-my-openagent v4.19.3 / commit `614cc5358dc393153fc39acae74dc5bd9fb9fffc`
- Node `v24.12.0`
- Bun `1.3.14`
- Senpi `2026.7.28`, installed globally through npm

Reproduction:

1. Install Senpi through npm so the Windows launcher is `%APPDATA%\npm\senpi` / `senpi.cmd`, not a native `senpi.exe`.
2. Build the v4.19.3 Senpi plugin.
3. Run the Team QA harness:

   ```bash
   node packages/omo-senpi/scripts/qa/team-e2e.mjs
   ```

4. Observe the initial Team member start failure.
5. If the top-level QA launcher is fixed in isolation, rerun and observe that the real RPC member path still fails until `packages/senpi-task/src/runners/rpc/spawn.ts` handles npm shims too.
6. Run the forced-kill/restart scenario: create a member, hold it with an uncommitted reservation, kill its process tree, restart the lead, and inspect the structured liveness event.

## Expected Behavior

- Native `senpi.exe` is executed directly.
- An npm shim is resolved without `shell: true` or broad shell forwarding.
- Team RPC children start successfully on Windows.
- QA terminates only the exact process tree it created.
- A terminal member state that survives a process crash is replayed to the restarted lead at least once.
- Delivery is marked durable only after the lead turn lifecycle reaches an acknowledgement boundary.
- The QA gate accepts only the expected structured liveness event and rejects unrelated events with liveness-shaped fields.

## Actual Behavior

Initial launcher failures:

```text
spawn C:\Users\<user>\AppData\Roaming\npm\senpi ENOENT
spawnSync C:\Users\<user>\AppData\Roaming\npm\senpi.cmd EINVAL
Task runner failed to start.
member_start_rejected
```

After fixing only the top-level QA launcher, the same assumption remains in the real RPC child path.

The crash/restart path also initially produced:

```text
crashKilledMemberAtHold: false
crashLivenessInjectedToLead: false
```

After exact process-tree termination was added, the process and reservation evidence passed but restart liveness still failed until delivery state was made durable across sessions.

### 1. Windows npm launcher handling exists at two boundaries

The top-level QA runner and `packages/senpi-task/src/runners/rpc/spawn.ts` both need to distinguish:

- native `.exe`: execute directly;
- npm `senpi` / `senpi.cmd`: invoke the installed package's JavaScript CLI through `process.execPath`;
- other/POSIX launchers: preserve the existing direct execution behavior.

For the affected paths, resolving the installed package CLI avoids `shell: true`, quoting ambiguity, and command injection risk.

### 2. POSIX process-group termination is not portable to Windows

The crash harness used negative PID semantics such as:

```js
process.kill(-pid, "SIGKILL")
```

Windows has no equivalent POSIX process-group contract. The safe Windows replacement is exact-root tree termination:

```text
taskkill.exe /PID <exact-root-pid> /T /F
```

This must never degrade into image-name or host-wide cleanup such as `/IM node.exe`.

### 3. Liveness dedupe was in-memory rather than crash-durable

The member could be persisted as terminal before the parent process died, but restart reconciliation only considered newly-lost outcomes and the delivery dedupe state did not survive the process boundary.

A candidate fix stores an optional epoch on the task record:

```ts
notification.liveness_notified_epoch?: number
```

Restart reconciliation re-observes already-terminal Team members. The notifier suppresses an event only when the persisted epoch proves a previous lead lifecycle acknowledged it.

### 4. A queue/send return is too early for durable acknowledgement

Marking delivery immediately after coordinator flush or `sendMessage()` return can lose the event if the process dies before the lead turn consumes it.

The candidate keeps the epoch pending after send and persists it at the later `agent_end` lifecycle boundary. A crash before that boundary therefore causes replay after restart. This intentionally provides at-least-once semantics.

### 5. The original QA detector was coupled to prose

The observed event can report abnormal terminal state `error` or `lost`. Searching for one literal `lost` sentence caused a false negative.

The candidate detector parses structured JSON events and requires:

- event type `message_start` or `message_end`;
- direct `senpi-task.team-member-liveness`, or `omo-senpi:wake` containing that detail;
- the target member name;
- abnormal state `error` or `lost`.

It also rejects an unrelated `tool_execution_end` containing similarly shaped fields.

## Candidate Fix

The local candidate changes these areas:

- `packages/omo-senpi/scripts/qa/team-e2e-runtime.mjs`
  - Windows npm launcher normalization
  - native `.exe` preservation
  - exact PID-tree termination evidence
- `packages/omo-senpi/scripts/qa/team-e2e-crash.mjs`
  - exact parent PID-tree termination
  - structured liveness detection
  - kill-time task record evidence
- `packages/senpi-task/src/runners/rpc/spawn.ts`
  - RPC child launcher normalization without a shell
- `packages/senpi-task/src/state/types.ts`
- `packages/senpi-task/src/store/record-parse.ts`
  - backward-compatible optional durable liveness epoch
- `packages/omo-senpi/src/components/task/member-liveness.ts`
- `packages/omo-senpi/src/components/task/engine.ts`
- `packages/omo-senpi/src/components/task/event-bridge.ts`
  - persisted terminal replay and delayed lifecycle acknowledgement

I can submit the candidate as reviewer-friendly commits if this direction is acceptable:

1. Windows launcher normalization and RPC child coverage
2. durable restart-liveness replay/acknowledgement
3. Windows crash QA and structured detector hardening

## Verification

Latest candidate results on Windows:

```text
Relevant tests: 80 passed, 0 failed, 204 assertions
packages/omo-senpi typecheck: PASS
packages/senpi-task typecheck: PASS
git diff --check: PASS
Team self-test: PASS
Windows live Team checks: 12/12 PASS
Leaked child PIDs after cleanup: 0
Credential isolation: clean
```

The live checks cover:

- create/list two Team members;
- lead-to-member and member-to-lead delivery;
- forced kill while the reservation is uncommitted;
- exact parent process-tree termination;
- clean restart;
- structured liveness delivery to the restarted lead;
- process quiescence.

## Risks and Residuals

- The candidate is not yet merged or installed as a released build.
- Liveness delivery is intentionally at-least-once. A crash before acknowledgement may produce a duplicate rather than lose the event.
- `acknowledgeDelivered()` currently assumes one lead delivery context per engine. Future multi-lead support should bind acknowledgement to a session/message identity.
- The full root suite still has separately known Windows symlink-sensitive fixture failures; the focused Team/RPC/liveness tests, both package typechecks, staged plugin build, and live Team harness are green.
- This issue does not propose broad process-name cleanup or `shell: true`.

## Related Work

- #5798 addressed a Windows npm `.cmd` launcher in the separate `packages/omo-ai` surface. This report covers `packages/omo-senpi` plus the real `senpi-task` RPC child and crash-liveness paths.
- #6374 tracks timing-only CI process flakes. This report reproduces deterministic launcher and restart-state logic failures.

## Doctor Output

```shell
Not applicable to this reproduction: the failing surface is the Senpi adapter/Team runtime rather than the OpenCode plugin doctor path.
Source version: oh-my-openagent v4.19.3 (614cc5358dc393153fc39acae74dc5bd9fb9fffc)
Senpi version: 2026.7.28
```

## Configuration

No user configuration, provider credentials, tokens, or absolute local workspace paths are required to reproduce the source-level failure.

## Operating System

Windows

## OpenCode Version

Not applicable — reproduced through the Senpi adapter and Team QA harness.

etehofk1-ops:
## Summary

- Fixes Windows Team-mode startup when Senpi is installed through npm shims rather than a native `.exe`.
- Replays abnormal team-member liveness after a process restart and persists acknowledgement only after the exact structured delivery is observed.
- Hardens the Windows Team E2E harness so exact PID-tree termination, restart replay, and durable acknowledgement are part of the gate.

## Changes

- **Windows RPC launcher:** prefer a native `senpi.exe`; normalize npm `senpi.cmd`/extensionless shims to `node <package>/dist/cli.js`; never enable shell forwarding.
- **Restart liveness:** persist optional `liveness_notified_epoch`, replay terminal team-member records during session reconciliation, and acknowledge only a matching direct or `omo-senpi:wake`-nested `deliveryKey` seen on `message_start`/`message_end`.
- **Process cleanup:** replace broad/ambiguous Windows cleanup with positive-safe-integer PID validation, exact `taskkill.exe /PID <pid> /T /F` evidence, and leak accounting.
- **QA discovery:** probe `senpi.exe`, `senpi.cmd`, and `senpi` on Windows, including backslash absolute overrides.
- **Artifacts:** regenerate the two tracked OmO Senpi extension bundles and add sanitized evidence under `.omo/evidence/omo-senpi-adapter/20260729-windows-team-crash-recovery/`.

## QA & Evidence

- **What was tested:** seven focused task/runner/QA test files.
  **Observed result:** 83 passed, 0 failed, 226 assertions.
  **Artifact:** `.omo/evidence/omo-senpi-adapter/20260729-windows-team-crash-recovery/README.md`
  **Why sufficient:** covers native/npm launcher selection, invalid-PID fail-closed behavior, exact Windows PID-tree cleanup, structured false-positive rejection, restart replay, matching-key acknowledgement, duplicate idempotence, and legacy record parsing.

- **What was tested:** live Windows Team run with one member killed at the post-injection hold, followed by a fresh-process restart.
  **Observed result:** 13/13 checks passed; `crashLivenessAcknowledged=true`; credential isolation and whole-directory integrity passed; `leakedPids=0`.
  **Artifact:** `.omo/evidence/omo-senpi-adapter/20260729-windows-team-crash-recovery/team-e2e.json` and `crash-recovery.json`
  **Why sufficient:** exercises the actual npm shim/native discovery surface, process-tree termination, replay, structured delivery observation, and durable marker persistence on Windows.

- **What was tested:** root gates and clean `origin/dev` comparison on the same Windows host.
  **Observed result:** `bun run typecheck` and `bun run build` passed. Candidate `bun test`: 12,287 pass / 73 fail / 22 skip; clean baseline: 12,267 pass / 73 fail / 22 skip. Candidate and baseline `bun run test:codex`: 359 pass / 28 fail / 7 skip.
  **Artifact:** evidence README above.
  **Why sufficient:** the candidate adds 20 passing tests without increasing the existing Windows failure count; Codex-package results are exact baseline parity.

- **What was tested:** independent bounded read-only review of the final non-generated diff.
  **Observed result:** pass, zero blocking findings after three concrete findings were fixed and re-reviewed.
  **Artifact:** summarized in the evidence README.
  **Why sufficient:** specifically re-checked delivery-specific acknowledgement, Windows launcher discovery, positive-safe-integer PID validation, exact PID cleanup, backward compatibility, and no shell/image-wide kill path.

## Risks & Residuals

- **At-least-once replay window — accepted:** a crash after host observation but before marker persistence may produce one duplicate on the next restart. A generic event can no longer suppress an undelivered notification.
- **Existing Windows root failures — not introduced:** the 73 failures are present in the clean `origin/dev` baseline on this host and are dominated by symlink `EPERM`, checkout materialization/home-fixture assumptions, and unrelated timeout-sensitive tests.
- **Generated artifacts — mitigated:** the two tracked OmO Senpi extension bundles were regenerated from the reviewed source; unrelated generated outputs were excluded.
- **Secrets/private paths — mitigated:** committed evidence is sanitized and contains no credentials, local usernames, private workspace paths, raw transcripts, UUIDs, or PIDs.

## Automated Checks

```bash
bun run typecheck
bun run build
bun test
bun run test:codex
bun test packages/omo-senpi/src/components/task/index.test.ts packages/omo-senpi/src/components/task/member-liveness.test.ts packages/omo-senpi/src/components/task/completion-bridge.test.ts packages/senpi-task/src/lifecycle/reconcile.test.ts packages/senpi-task/src/store/record-store.test.ts packages/senpi-task/src/runners/rpc/spawn.test.ts packages/omo-senpi/scripts/qa/team-e2e-runtime.test.ts
node packages/omo-senpi/scripts/qa/team-e2e.mjs --self-test
node packages/omo-senpi/scripts/qa/team-e2e.mjs
```

## Related Issues

Closes #6449

---
## Summary by cubic
Hardened Windows Team-mode startup and crash recovery: reliably launch npm-shimmed `senpi`, terminate the exact Windows PID tree on crashes, and persist/acknowledge team liveness precisely on restart. Addresses #6449 with stronger Windows E2E checks and refreshed evidence.

- **Bug Fixes**
  - Windows RPC launcher: prefer native `senpi.exe`; normalize `senpi.cmd`/`senpi` shims to `node <package>/dist/cli.js`; no shell forwarding.
  - Restart liveness: persist `liveness_notified_epoch`, replay terminal member records on session start, and acknowledge only matching `deliveryKey` observed on `message_start`/`message_end`.
  - Process cleanup: validate PIDs (reject non-positive/non-integer/NaN/∞) and terminate the exact tree with `taskkill.exe /PID <pid> /T /F`; leak accounting.
  - QA/docs: Windows discovery supports `senpi.exe`, `senpi.cmd`, `senpi`, and backslash-absolute overrides; crash E2E asserts exact cleanup, replay, and durable acknowledgement; sanitized evidence updated under `.omo/evidence/omo-senpi-adapter/20260729-windows-team-crash-recovery/`.

<sup>Written for commit 0a0dfe32e471c93b4f27d49cdf3dd35207824eb8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


etehofk1-ops:

code-yeongyu (commented):
Thanks for this thorough and well-documented PR — the issue write-up in #6449 is one of the clearest root-cause analyses this repo has received, and a lot of the implementation direction is genuinely right: PID validation before any termination side effect, a structured liveness detector instead of prose matching, exact-PID `taskkill /T` instead of image-name cleanup, and a backwards-compatible optional epoch field are all the correct instincts. The evidence hygiene (sanitized artifacts, baseline comparison) is also appreciated.

A multi-pass review (launcher/process, durability/liveness, QA/evidence, plus a cross-examination round and host-source verification against the Senpi runtime) found a set of issues that need addressing before this can merge. Posting as a comment review since the PR is still draft; happy to discuss any item.

## Blockers

**1. The durable liveness marker is written before Senpi has persisted the message** — `event-bridge.ts:94-100`, `member-liveness.ts:77-81`, `engine.ts:117-125`

`acknowledgeDelivered` fires from `message_start`/`message_end` extension handlers, but the host awaits extension handlers *before* its `message_end` persistence step (`appendCustomMessageEntry` runs after `_emitExtensionEvent` in `_processAgentEvent`). A kill in that window leaves the lead transcript without the liveness message while the task record claims delivery — and the next restart suppresses replay via the `>=` epoch check. Note that a bare `agent_end` ack is also not a correct boundary: session-file creation can be deferred until an assistant entry exists, so even turn-end is not a persistence receipt.

Suggested direction: ack only when the delivery is durably observable — e.g. scan the lead's session JSONL for the liveness entry (the repo already has this exact pattern in `team/messaging/session-marker-index.ts` for team-mail commit verification), evaluated at turn boundaries with a bounded retry. `message_start`/`message_end` should never acknowledge.

**2. Any lead in the same project can consume and permanently suppress another lead's liveness alert** — `event-bridge.ts:35-40`

`reconcileOnSessionStart` enumerates the shared project store with no session filter, and the bridge now re-observes *every* outcome into `notifyTerminal`. Lead B opening a session in the same project receives and durably acknowledges lead A's alert; A's restart then suppresses it. The team-mail poller (`lead-poller-lifecycle.ts`) and the completion-retry path both fence by owning session — liveness replay is the odd one out.

Suggested direction: parse `team:<teamRunId>:<member>` from the record name, load that team's durable runtime state, and require `leadSessionId === currentSessionId` (plus membership of the member name) before notifying — on both the reconcile-replay path and the live terminal path in `engine.ts`.

**3. The new Windows-shim unit test deterministically fails on POSIX CI** — `team-e2e-runtime.test.ts:134-151`, `team-e2e-runtime.mjs:8`

`resolveSenpiInvocation` uses host-native `path.basename`/`dirname`/`join` even when the test injects `platform: "win32"`. On macOS/Linux, `basename("C:\\Users\\qa\\...\\senpi")` is the whole string, so the function returns the shim directly and the assertion fails. Reproduced on Darwin:

```
15 pass / 1 fail — team-e2e-runtime.test.ts
(fail) #given a Windows npm senpi shim #when the spawn invocation is resolved
```

Root CI runs `bun test` on ubuntu-latest and macos-latest, so this PR ships red CI. Fix: use `path.win32` semantics whenever the effective platform is win32.

**4. The crash gate accepts outcomes that don't prove the ordered kill it claims** — `team-e2e-crash.mjs:59-65,105-113`, `team-e2e-runtime.mjs:134-139`

Three compounding holes: (a) the member's terminal transition is never awaited before the parent kill, so member-before-parent ordering is unproven; (b) `already-exited` satisfies `crashKilledMemberAtHold` even though the harness never proved the parent was live at kill time; (c) a nonzero `taskkill` result is converted to `already-exited` whenever the root PID is gone, discarding status/stderr — a partial tree kill becomes a pass. For this controlled scenario, require established liveness of both processes, an observed member terminal transition, and `kind === "terminated"` for the parent tree.

## High priority

**5. `markDelivered` can clobber a concurrent lifecycle write** — `engine.ts:117-125`

The load → mutate → `replace` sequence isn't serialized across processes: a stale `fresh` snapshot with only the epoch changed can overwrite a newer record written by a revive/reattach in another process. A store-level conditional patch (re-read inside a lock/CAS, touch only the notification field, skip if status/run_epoch advanced) would close this.

**6. A failed async injection is locally deduped forever** — `member-liveness.ts:42-75`, `idle-injection-coordinator.ts:106-127`

The key enters `delivered` before queueing; the coordinator clears its queue before the delivery callback; and the production binding fire-and-forgets `sendCustomMessage(...).catch(...)`, so an async rejection (e.g. provider admission failure) leaves the event suppressed in-process until a restart. A failure receipt that requeues and clears the local dedupe entry would make the retry story match the at-least-once intent.

**7. Retained numeric PIDs can force-kill an unrelated process tree after PID reuse** — `team-e2e.mjs:42,77-92`, `team-e2e-runtime.mjs:129-135`

`spawnedGroups` retains numeric root PIDs after their ChildProcess closes; cleanup later probes only numerical liveness before `taskkill /T /F`. A recycled PID means killing a stranger's tree. Keep ChildProcess handles (or close-event removal) and terminate only still-owned processes.

**8. The stranded pre-commit reservation is never verified after restart** — `team-e2e-crash.mjs:58-98`

The only mailbox assertion is pre-kill (`processedExists === false && eventCount === 0`). After restart, nothing checks whether the exact reserved message was reclaimed, delivered exactly once, or intentionally abandoned — a regression that strands or duplicates it passes all 13 checks. This is orthogonal to the liveness fix but required for the crash-recovery claim.

**9. The durable-ack chain is validated synthetically** — `member-liveness.test.ts:185-187`, `index.test.ts:301-346`, `team-e2e.mjs:269-290`

The engine test calls `acknowledgeDelivered` directly, the bridge test spies it, the replay test stubs `notifyTerminal`, and `--self-test` preloads a passing evidence object into the evaluator. All of them pass with blockers 1–2 intact. The minimum merge-grade additions: a test proving no ack before host persistence (with replay after simulated crash), a cross-lead suppression test, an injection-failure retry test, and a `runCrashRestartScenario` test with controlled process/fs seams.

## Medium / low

- **Local `.bin` npm shims aren't resolved** (`spawn.ts:97-116`): `<project>/node_modules/.bin/senpi.cmd` maps the CLI lookup to `.bin/node_modules/...`, which never exists, and silently falls back to the loader-aliased rpc-entry. Not a documented install path today, but worth a fix + regression test. The QA harness variant of this gap (`team-e2e-runtime.mjs:8-21`) is sharper: it returns the `.cmd` shim for direct spawning when the CLI lookup misses — the exact failure mode this PR eliminates — instead of failing loudly.
- **POSIX tree termination can report false failures** (`team-e2e-runtime.mjs:150-159`): the immediate post-`SIGKILL` probe can hit an unreaped zombie and return `failed` after a successful kill. A bounded survivor poll (as in `packages/utils/src/process-tree-termination.ts`) fixes it.
- **Production RPC teardown is root-only** (`rpc/terminate.ts:26-31`) while QA now asserts tree-level cleanup — pre-existing, but the QA result can't be cited as proving shipping teardown behavior.
- **Bun-sibling `senpi.exe` is accepted without `existsSync`** (`spawn.ts:87-89`), bypassing PATH/shim fallbacks when the sibling is absent.
- **The committed evidence isn't traceable to the driver** (`.omo/evidence/.../crash-recovery.json`): derived field names appear in no generator, and the delivery key is stripped even though the README claims key-matched acknowledgement. A first-party sanitized driver output would make the artifact self-substantiating.
- **No live Windows execution of `team-e2e.mjs` in CI** — the mocked unit seams run on the Windows matrix, but the driver itself (which exits 0 with `SKIP` when no binary is found) never does. A required Windows job with explicit `SENPI_BIN`, JSON verdict parsing, and evidence upload would close the loop. Fine as a fast follow-up.
- **Housekeeping**: `delivered` grows unboundedly in `member-liveness.ts` (entries never removed after ack); the persisted epoch is already the cross-process dedupe authority.

## What's solid

The epoch-compare direction (`>=` against persisted marker), the fail-closed PID validation (no probe or kill side effects for invalid PIDs), argv-array spawning with no shell anywhere, the structured detector's rejection of liveness-shaped fields in unrelated events, and the honest README section on residual at-least-once semantics — all good work worth keeping.

None of the blockers look structurally hard; most are sequencing and fencing rather than redesign. Happy to re-review quickly once updated. 🙏

code-yeongyu:
Thanks again for the excellent groundwork here! 🙏 To move this along, #6454 carries all of your commits verbatim (authorship preserved) plus the fixes for every review finding above — durable JSONL-observed acknowledgement, lead-ownership fencing, the POSIX CI path fix, the strict crash gate, and the exactly-once reservation-recovery check (live E2E now PASS 16/16 on macOS). When #6454 merges, GitHub will mark this PR merged as well since its head commit is included. A fresh Windows live run against the strengthened harness would still be a very welcome follow-up if you have the environment handy!

PR #6488, opened 2026-07-30
fix(lsp-core): catch async spawn errors and bound init timeout (#6486)

### LINKED ISSUE #6486: bug(lsp-core): LSP that hangs on initialize or fails asynchronously after spawn blocks for 60s with no fast timeout
## Summary

When an LSP server process spawns successfully but then either (a) never responds to `initialize`, or (b) fails asynchronously right after spawn (e.g. non-executable / wrong-arch binary that emits an async `ENOENT`/crash), the engine blocks for the full `INIT_TIMEOUT_MS` (60s) before failing. There is no fast bound on `start()` + `initialize()`. On agent harnesses this stacks past background-task stale timeouts and silently cancels workers that had already completed their real work.

Note: the *missing-binary* case (binary absent from PATH) is already handled well and cross-platform via `isServerInstalled()` in `server-resolution.ts` — that path fails fast on Linux, macOS, and Windows. This report is specifically about the binary that *passes* the existence check but then hangs or dies asynchronously.

## Environment

- oh-my-openagent v4.19.3 (`lsp-core`)
- Reproduced on Linux (Fedora + Freedesktop SDK Flatpak runtime); root cause is platform-independent.

## Root Cause

`LspClientTransport.start()` only inspects `exitCode` **synchronously** right after spawn. A process that will fail asynchronously has not exited yet at this point, so the guard passes and `start()` returns as if healthy:

`packages/lsp-core/src/lsp/transport.ts:88-97`
```ts
this.proc = spawnProcess(this.server.command, { cwd: this.root, env });
this.startStderrReading();

if (this.proc.exitCode !== null) {           // still null — async failure not fired yet
    const stderr = this.stderrBuffer.join("\n");
    throw new LspProcessExitedError(/* ... */);
}
```

Control then reaches `initialize()`. The only bound is `INIT_TIMEOUT_MS = 60_000` (`packages/lsp-core/src/lsp/constants.ts:7`), with no env override.

The manager calls `start()` then `initialize()` with no fast bound at `packages/lsp-core/src/lsp/manager.ts:203-204` (and `manager.ts:270-271` for warmup). The stale reaper (`manager.ts:112-123`) does clean up clients stuck initializing past `initTimeoutMs`, but only on its 60s interval — an in-flight `getClient()` caller still waits out the full 60s.

## Steps to Reproduce

1. Configure an LSP server whose command resolves on PATH (passes `isServerInstalled`) but does not speak LSP / never answers `initialize` (e.g. point it at a script that reads stdin and never replies).
2. Trigger any `lsp_*` tool against a matching file.
3. Observe a ~60s stall before the call fails, instead of a fast error.

## Impact

- Every affected `lsp_*` call costs 60s.
- On agent harnesses, stacked calls exceed the background-task stale timeout (`DEFAULT_STALE_TIMEOUT_MS` = 5 min, set in #4217) and silently cancel background workers that already finished their real work.

## Proposed Fix

1. Apply a short, dedicated timeout to `start()` + `initialize()` (e.g. a few seconds), separate from the 60s ceiling, so a silent hang fails fast; make it env-overridable.
2. Additionally, surface async spawn failures: attach an `error` handler to the child process so an async `ENOENT`/spawn error rejects `start()` immediately rather than being missed by the synchronous `exitCode` check.

## Workaround

Disable the offending server in project `.opencode/lsp.json`:
```json
{ "some-server": { "disabled": true } }
```
Confirmed effective: on a host with no warm LSP servers, background workers that previously stale-cancelled at ~5 min completed in ~14s after disabling the server.

## Related

- #4217 — introduces `DEFAULT_STALE_TIMEOUT_MS` = 5 min; the timeout this 60s×N stacking blows past.
- #5260 / #5343 — background tasks hanging on LSP (`lsp_install_decision` permission gate). Same *symptom* (background stall via LSP), different root cause; listed to disambiguate.
- #6256 — lsp-daemon mid-request socket close/error handling; related daemon teardown behavior around the manager idle timeout.
- #6301 — `initialized` notification params; touches the same init sequence and a strict-server "Server not initialized" hang.
- #6326 — Node <17.3 `throwIfAborted` crash. Contrast case: that fails *instantly*; this bug is the opposite (silent 60s block). Not a duplicate.

BenSharir:
Closes #6486

## Summary

An LSP server that spawns successfully but then either (a) never responds to `initialize`, or (b) fails asynchronously right after spawn (non-executable / wrong-arch binary emitting an async `ENOENT`/crash) blocked for the full `INIT_TIMEOUT_MS` (60s) before failing. There was no fast bound on `start()` + `initialize()`. On agent harnesses this stacks past background-task stale timeouts and silently cancels workers that already finished their real work.

Note: the *missing-binary* case is already handled cross-platform by `isServerInstalled()` in `server-resolution.ts`. This fix targets the binary that *passes* the existence check but then hangs or dies asynchronously.

## Root cause

`LspClientTransport.start()` only inspected `exitCode` **synchronously** right after spawn. A process that fails asynchronously has not exited yet at that point, so the guard passes and `start()` returns as if healthy. Control then reaches `initialize()`, whose only bound was `INIT_TIMEOUT_MS = 60_000`. The manager called `start()` then `initialize()` with no fast bound at both the `getClient()` and `warmupClient()` sites.

## Changes

1. **Surface async spawn failures (`transport.ts` + `process.ts`).**
   `SpawnedProcess` now exposes `spawnError` and `onSpawnError()`. The child's `'error'` event is captured in `wrap()` and fanned out to listeners. `LspClientTransport.start()` attaches a handler that flags the process dead (`processExited = true`) and throws `LspProcessSpawnError` when the error is already visible, instead of relying solely on the synchronous `exitCode` check.

2. **Bound `start()` + `initialize()` with a fast, env-overridable timeout (`constants.ts` + `manager.ts`).**
   Added `START_TIMEOUT_MS` (default 10s, overridable via `OMO_LSP_START_TIMEOUT_MS`). Both the `getClient()` and `warmupClient()` init promises now run through `startAndInitializeClient()`, which races `start()` + `initialize()` against `START_TIMEOUT_MS`. A silent hang now fails in ~10s instead of 60s. The timer is `.unref()`ed so it never keeps the process alive.

## Before / After

| Scenario | Before | After |
|---|---|---|
| Server hangs on `initialize` | ~60s stall per call | fails in ~10s (`START_TIMEOUT_MS`) |
| Async spawn failure (ENOENT/crash) | missed by sync `exitCode` check -> 60s | `LspProcessSpawnError` surfaced; process flagged dead |
| Healthy server | works | works (adds a `.unref()`ed race timer only) |
| Missing binary | already fast (`isServerInstalled()`) | unchanged |

## Verification

Standalone runtime checks (bun):
- Hanging `initialize()` -> race rejects fast at the timeout (measured at the bound, not 60s).
- Healthy client -> resolves normally.
- `start()` that rejects -> error propagates.
- Real `spawn()` of a nonexistent binary -> `spawnError` is `null` synchronously (reproduces the bug), then the async `error` event fires and both `onSpawnError` and the `spawnError` getter capture the `ENOENT`.

All four changed files transpile clean (`bun build --target=node`). Scope limited to `lsp-core`; no tests/docs/other files touched.

---
## Summary by cubic
Fixes LSP startup hangs by catching async spawn errors and adding a fast, configurable timeout around start + initialize. Servers that hang or crash right after spawn now fail in ~10s with a clear error instead of stalling for 60s.

- **Bug Fixes**
  - `SpawnedProcess` exposes `spawnError` and `onSpawnError()`; `LspClientTransport.start()` surfaces async spawn failures as `LspProcessSpawnError` and flags the process exited.
  - Bounded `start()` + `initialize()` with `START_TIMEOUT_MS` (default 10s) via a race in `LspManager`; overridable via `OMO_LSP_START_TIMEOUT_MS` or `LspManagerOptions.startTimeoutMs`. Empty/non-numeric/<=0 env values fall back to 10s; timer is `.unref()`ed.
  - Added regression tests for the start-timeout race and the async spawn-error path; fixed the `HealthyLspClient` test constructor to typecheck.
  - Documented `OMO_LSP_START_TIMEOUT_MS` in the LSP configuration docs and environment variables table.

<sup>Written for commit 517192b585d02e529f370ae5f1c83796994c0e4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/lsp-core/src/lsp/constants.ts
+++ b/packages/lsp-core/src/lsp/constants.ts
@@ -5,6 +5,7 @@ export const DEFAULT_MAX_DIRECTORY_FILES = 50;
 
 export const REQUEST_TIMEOUT_MS = 15_000;
 export const INIT_TIMEOUT_MS = 60_000;
+export const START_TIMEOUT_MS = Number(process.env["OMO_LSP_START_TIMEOUT_MS"]) || 10_000;
 export const IDLE_TIMEOUT_MS = 5 * 60_000;
 export const REAPER_INTERVAL_MS = 60_000;
 export const STOP_HARD_KILL_TIMEOUT_MS = 5_000;
--- a/packages/lsp-core/src/lsp/constants.ts
+++ b/packages/lsp-core/src/lsp/constants.ts
@@ -5,16 +5,10 @@ export const DEFAULT_MAX_DIRECTORY_FILES = 50;
 
 export const REQUEST_TIMEOUT_MS = 15_000;
 export const INIT_TIMEOUT_MS = 60_000;
-export const START_TIMEOUT_MS = resolveStartTimeoutMs();
+export const START_TIMEOUT_MS = Number(process.env["OMO_LSP_START_TIMEOUT_MS"]) > 0
+	? Number(process.env["OMO_LSP_START_TIMEOUT_MS"])
+	: 10_000;
 export const IDLE_TIMEOUT_MS = 5 * 60_000;
 export const REAPER_INTERVAL_MS = 60_000;
 export const STOP_HARD_KILL_TIMEOUT_MS = 5_000;
 export const STOP_SIGKILL_GRACE_MS = 1_000;
-
-function resolveStartTimeoutMs(): number {
-	const raw = process.env["OMO_LSP_START_TIMEOUT_MS"];
-	if (raw === undefined || raw === "") return 10_000;
-	const parsed = Number(raw);
-	if (!Number.isFinite(parsed) || parsed <= 0) return 10_000;
-	return parsed;
-}
--- a/packages/lsp-core/src/lsp/constants.ts
+++ b/packages/lsp-core/src/lsp/constants.ts
@@ -5,8 +5,16 @@ export const DEFAULT_MAX_DIRECTORY_FILES = 50;
 
 export const REQUEST_TIMEOUT_MS = 15_000;
 export const INIT_TIMEOUT_MS = 60_000;
-export const START_TIMEOUT_MS = Number(process.env["OMO_LSP_START_TIMEOUT_MS"]) || 10_000;
+export const START_TIMEOUT_MS = resolveStartTimeoutMs();
 export const IDLE_TIMEOUT_MS = 5 * 60_000;
 export const REAPER_INTERVAL_MS = 60_000;
 export const STOP_HARD_KILL_TIMEOUT_MS = 5_000;
 export const STOP_SIGKILL_GRACE_MS = 1_000;
+
+function resolveStartTimeoutMs(): number {
+	const raw = process.env["OMO_LSP_START_TIMEOUT_MS"];
+	if (raw === undefined || raw === "") return 10_000;
+	const parsed = Number(raw);
+	if (!Number.isFinite(parsed) || parsed <= 0) return 10_000;
+	return parsed;
+}
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -693,6 +693,11 @@ and reads the server map from that file. The schema lives in the
 `packages/lsp-tools-mcp` vendored package (upstream:
 [code-yeongyu/lsp-tools-mcp](https://github.com/code-yeongyu/lsp-tools-mcp)).
 
+A language server that spawns but then hangs on `initialize` or fails
+asynchronously right after spawn is bounded by `OMO_LSP_START_TIMEOUT_MS`
+(default `10000`; see [Environment Variables](#environment-variables)) rather
+than stalling for the 60s init ceiling.
+
 To disable the LSP MCP entirely:
 
 ```json
@@ -1107,6 +1112,7 @@ When enabled, OmO registers the hash-anchored `edit` tool and activates the `has
 | `LAZYCODEX_CONFIG_MIGRATION_DISABLED` | Set to `1` to skip the Codex config migration that runs on every session start (including the `multi_agent_v2` force-disable and managed reasoning-profile sync), leaving `config.toml` untouched |
 | `OMO_CODEX_CONFIG_MIGRATION_DISABLED` | Alias of `LAZYCODEX_CONFIG_MIGRATION_DISABLED` |
 | `LSP_TOOLS_MCP_INSTALL_DECISIONS` | Override the path of the LSP install-decisions file (default `~/.codex/lsp-install-decisions.json`) |
+| `OMO_LSP_START_TIMEOUT_MS` | Timeout in milliseconds bounding a language server's `start()` + `initialize()`. A server that spawns but hangs or crashes asynchronously fails after this bound instead of stalling for the 60s init ceiling. Defaults to `10000` (10s); empty, non-numeric, zero, or negative values fall back to `10000` |
 | `POSTHOG_API_KEY` | Optional override for the built-in PostHog project API key |
 | `POSTHOG_HOST` | Override the PostHog ingestion host. Defaults to `https://us.i.posthog.com` |
 
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -1112,7 +1112,7 @@ When enabled, OmO registers the hash-anchored `edit` tool and activates the `has
 | `LAZYCODEX_CONFIG_MIGRATION_DISABLED` | Set to `1` to skip the Codex config migration that runs on every session start (including the `multi_agent_v2` force-disable and managed reasoning-profile sync), leaving `config.toml` untouched |
 | `OMO_CODEX_CONFIG_MIGRATION_DISABLED` | Alias of `LAZYCODEX_CONFIG_MIGRATION_DISABLED` |
 | `LSP_TOOLS_MCP_INSTALL_DECISIONS` | Override the path of the LSP install-decisions file (default `~/.codex/lsp-install-decisions.json`) |
-| `OMO_LSP_START_TIMEOUT_MS` | Timeout in milliseconds bounding a language server's `start()` + `initialize()`. A server that spawns but hangs or crashes asynchronously fails after this bound instead of stalling for the 60s init ceiling. Defaults to `10000` (10s); empty, non-numeric, zero, or negative values fall back to `10000` |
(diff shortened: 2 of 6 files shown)


BenSharir:

code-yeongyu:
[sisyphus-bot] Clean fix for both cases in #6486 — the `Promise.race` start timeout and `onSpawnError` async error capture are the right approach. Two things to verify: (1) confirm `LspProcessSpawnError` is defined in `errors.ts`, and (2) push to a branch that triggers the full test/typecheck CI matrix to confirm green. The 10s default is reasonable but consider documenting `OMO_LSP_START_TIMEOUT_MS` in the docs.

BenSharir:
Thanks for the review! Addressed all three points:

**1. `LspProcessSpawnError` is defined.** It lives in `packages/lsp-core/src/lsp/errors.ts` (`export class LspProcessSpawnError`), and is imported/thrown in both `transport.ts` and `process.ts`. Confirmed, no gap.

**2. CI / green matrix.** While verifying this I caught a real typecheck failure in the test I added: in `manager-start-timeout.test.ts`, `HealthyLspClient` was instantiated with a single arg but inherited `LspClient`'s 2–3 arg constructor, so `tsgo` failed with `TS2554` (the `typecheck:packages` job would have gone red). Fixed by giving `HealthyLspClient` a matching 1-arg constructor, mirroring `HangingInitLspClient`. Locally now:
- `bunx tsgo --noEmit -p packages/lsp-core/tsconfig.json` → clean
- `bun test packages/lsp-core/` → **94 pass / 0 fail** (both start-timeout regression tests included)

Pushed as `3d5b8c192`. Since this is a fork PR, the full CI matrix needs a maintainer **"Approve and run workflows"** to execute — could you kick that off so we can confirm green end-to-end?

**3. Documented `OMO_LSP_START_TIMEOUT_MS`.** Added to `docs/reference/configuration.md`: a note in the **LSP** section and a row in the **Environment Variables** table (default `10000`ms / 10s; empty, non-numeric, zero, or negative values fall back to `10000`).

Note: the `10s` default is unchanged. One pre-existing, unrelated typecheck error in `packages/utils/src/runtime/file.ts` is present on the branch but is out of scope for this PR (different package, untouched here).

BenSharir:
Verified the diff against the root cause — both halves are implemented correctly: async `onSpawnError` capture in `process.ts`/`transport.ts`, and the `START_TIMEOUT_MS` race applied at **both** the `getClient()` and `warmupClient()` init sites (with a `.unref()`ed timer). Matches the proposed fix in #6486.

One optional gap: the committed suite covers the **init-hang timeout** (`HangingInitLspClient`) and the healthy path, but not the **async spawn-error** path itself — an `'error'` event causing `start()` to reject with `LspProcessSpawnError`. That's the other half of the fix and is currently only exercised by the ad-hoc bun checks in the PR description, not the committed tests. Worth a committed regression test so that half can't silently regress. Non-blocking.

Otherwise LGTM pending the full CI matrix run.

packages/lsp-core/src/lsp/constants.ts
```diff
@@ -5,7 +5,16 @@ export const DEFAULT_MAX_DIRECTORY_FILES = 50;
 
 export const REQUEST_TIMEOUT_MS = 15_000;
 export const INIT_TIMEOUT_MS = 60_000;
+export const START_TIMEOUT_MS = resolveStartTimeoutMs();
```
kevinhq: while this resolveStartTimeoutMs is technically correct, it seems over-engineered/over-kill because we can just do it one line code like:

```
export const START_TIMEOUT_MS = Number(process.env["OMO_LSP_START_TIMEOUT_MS"]) || 10_000;
```

BenSharir: Good call — switched to the one-liner in `517192b58`. Kept a `> 0` guard so a stray negative value doesn't produce an instant timeout, but otherwise it's exactly your version (non-numeric/empty/zero fall back to the 10s default via the same falsiness). Dropped the now-redundant `resolveStartTimeoutMs()` helper and trimmed the over-detailed docs row to match.

Also took the opportunity to close the one real gap from my own review: added a committed regression test for the async spawn-error path (`SpawnErrorLspClient` → `getClient()` rejects with `LspProcessSpawnError` and cleans up), so that half of the fix is no longer only covered by ad-hoc checks. Locally: `tsgo` clean, `bun test packages/lsp-core/` → 95 pass / 0 fail.

MoerAI (changes requested):
Thanks for isolating the LSP cold-start hang and for adding a bounded manager deadline.

Please rebase onto current `dev` and integrate with the existing `pendingStops` and tombstone lifecycle. Move the cold-start guard to the transport/initialize boundary: reject on real process exit or asynchronous spawn error, abort the pending initialize request, and always dispose the deadline timer.

The current process handler only sets flags after `start()` has already returned, so a real asynchronous child error still waits for the generic 10-second race:
https://github.com/code-yeongyu/oh-my-openagent/blob/517192b585d02e529f370ae5f1c83796994c0e4a/packages/lsp-core/src/lsp/process.ts#L61-L102

Add deterministic Node-controlled real-process coverage for asynchronous exit, silent initialize, healthy startup, timer cleanup, and concurrent acquisition. Do not use 100 ms and 500 ms wall-clock races or mock only `start()` rejection:
https://github.com/code-yeongyu/oh-my-openagent/blob/517192b585d02e529f370ae5f1c83796994c0e4a/packages/lsp-core/src/lsp/manager-start-timeout.test.ts#L16-L27

Keep this PR LSP-only and rerun current CI plus real affected-surface QA. This follows the canonical-successor direction recorded when #7244 was closed:
https://github.com/code-yeongyu/oh-my-openagent/pull/7244#issuecomment-5486635049

PR #6519, opened 2026-07-31
fix(cmux): keep OMO subagent panes attached on focus

Atmospenguin:
## Summary

- Recognize cmux's official `CMUX_AGENT_LAUNCH_KIND=omo` marker at the inline-pane spawn and focus decisions.
- Resolve the current OpenCode executable once for official OMO panes, quote its absolute path, and attach immediately at spawn.
- Treat focus as successful for the already-attached pane, avoiding `respawn-pane -k`.
- Keep native tmux behavior and authenticated-cmux fail-closed handling intact.

## Root cause

The official cmux launcher exports `CMUX_AGENT_LAUNCH_KIND=omo`, while its injected `TMUX` value may not match the existing cmux detector. OMO can therefore start a focus placeholder and later run `respawn-pane -k`, removing the cmux surface while the subagent continues in the background.

The eager attach command also used bare `opencode`. cmux launches the string command through a clean shell, so an NVM-installed OpenCode binary is absent from that shell's PATH and the pane exits with `opencode: command not found`.

## Scope

The production change is limited to the two lifecycle branches and the attach command builder. The absolute path is resolved only for the official OMO marker. The remaining changes are two direct regression tests; there is no detector helper, runner change, or QA artifact.

## Validation

- Failing-first lifecycle run: 13 pass, 2 expected failures.
- Failing-first clean-shell run: the official OMO command still contained bare `opencode`.
- `packages/tmux-core`: 100 pass, 0 fail, 214 assertions.
- Repository `bun run typecheck`: passed.
- The official OMO spawn test executes the generated command with `PATH=/usr/bin:/bin` and passes.
- Real cmux 0.64.21 with OpenCode 1.18.11 and OMO 4.19.4: the pane visibly launched with the absolute OpenCode path, survived pointer focus, logged `SKIP: cmux pane already attached`, and remained tracked as busy through later polls.

## Relationship to #6390

#6390 changes broader cmux detection and executable routing. This PR stays lifecycle-scoped and can land independently; if #6390 lands first and verifies the same focus behavior in a release, this can be closed as superseded.

### DIFF
--- a/packages/tmux-core/src/tmux-utils/pane-activate.ts
+++ b/packages/tmux-core/src/tmux-utils/pane-activate.ts
@@ -28,9 +28,14 @@ export async function activateTmuxPane(
   }
 
   const authEnvArgs = buildPaneAuthEnvironmentArgs()
-  if (isCmuxCompatEnvironment() && authEnvArgs.length > 0) {
+  const isCmuxOmoLaunch = process.env.CMUX_AGENT_LAUNCH_KIND === "omo"
+  if ((isCmuxCompatEnvironment() || isCmuxOmoLaunch) && authEnvArgs.length > 0) {
     deps.log("[activateTmuxPane] SKIP: authenticated cmux panes are unsupported", { paneId, sessionId })
     return false
+  }
+  if (isCmuxOmoLaunch) {
+    deps.log("[activateTmuxPane] SKIP: cmux OMO pane is already attached", { paneId, sessionId })
+    return true
   }
 
   const tmux = await deps.getTmuxPath()
--- a/packages/tmux-core/src/tmux-utils/pane-auth-env.test.ts
+++ b/packages/tmux-core/src/tmux-utils/pane-auth-env.test.ts
@@ -10,6 +10,7 @@ import { spawnTmuxWindow } from "./window-spawn"
 
 const originalTmux = process.env.TMUX
 const originalCmuxSocketPath = process.env.CMUX_SOCKET_PATH
+const originalCmuxAgentLaunchKind = process.env.CMUX_AGENT_LAUNCH_KIND
 
 const enabledTmuxConfig = {
 	enabled: true,
@@ -96,6 +97,7 @@ describe("tmux pane auth environment propagation", () => {
 	beforeEach(() => {
 		process.env.TMUX = "/tmp/tmux-1000/default,1234,0"
 		delete process.env.CMUX_SOCKET_PATH
+		delete process.env.CMUX_AGENT_LAUNCH_KIND
 	})
 
 	afterEach(() => {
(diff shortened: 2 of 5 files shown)


Atmospenguin:
Validated the current head a1c5b102 against a real cmux session today:

- cmux 0.64.22 (102)
- OpenCode 1.18.14
- source build installed from this PR head
- targeted tests: 15 pass, 0 fail

End-to-end check:
1. Started a fresh workspace with cmux omo.
2. Launched one background subagent running sleep 60.
3. The pane opened directly as an attached OpenCode session using the absolute OpenCode binary path.
4. Focused the subagent pane and kept it focused; the split remained intact.
5. Runtime log recorded: [activateTmuxPane] SKIP: cmux OMO pane is already attached.
6. The subagent returned CMUX_PANE_QA_DONE_3, the parent collected the result, and the pane closed only through normal completion.

This reproduces the affected launch/focus path and confirms the minimal lifecycle fix end to end.

Atmospenguin:
Validated the current head  against a real cmux session today:\n\n- cmux \n- OpenCode \n- source build installed from this PR head\n- targeted tests: \n\nEnd-to-end check:\n1. Started a fresh workspace with .\n2. Launched one background subagent running .\n3. The pane opened directly as an attached OpenCode session using the absolute OpenCode binary path.\n4. Focused the subagent pane and kept it focused; the split remained intact.\n5. Runtime log recorded:\n   \n6. The subagent returned , the parent collected the result, and the pane closed only through normal completion.\n\nThis reproduces the affected launch/focus path and confirms the minimal lifecycle fix end to end.

code-yeongyu:
[sisyphus-bot] The env-gated approach is clean and tests cover the new OMO launch path. Two things to verify before merge: (1) confirm `bunWhich` from `@oh-my-opencode/utils/runtime` doesn't introduce a circular dependency into `tmux-core`, and (2) the PR lacks test/typecheck CI runs — please push to a branch that triggers the full CI matrix so we can confirm green before merging.

Atmospenguin:
Thanks for flagging both gates. I verified them on the refreshed branch.

1. Dependency direction

- tmux-core already declares @oh-my-opencode/utils as a workspace dependency.
- packages/tmux-core/src/runner.ts already imported spawn from @oh-my-opencode/utils/runtime before this PR, so bunWhich reuses an existing package edge and runtime export path.
- utils has no dependency or import back to tmux-core. A workspace package-graph traversal reports utilsReachesTmuxCore=false and cyclePath=null.
- The repository package-registration audit passes 6/6, including the reverse-edge audit.

2. Test and typecheck CI

- I rebased the two commits onto dev at 69a1855e6; the PR head is now 971cf8231.
- The refreshed PR diff remains limited to the same 3 production files and 2 regression-test files.
- Local focused regression tests pass 15/15, and bun run typecheck passes.
- The upstream full matrix now passes on the exact PR head: https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31095374441
- Green jobs cover build, root tests, typecheck, Codex compatibility, and Senpi compatibility across Ubuntu, macOS, and Windows, plus the published lazycodex smoke test.

MoerAI:
Thanks for the focused cmux lifecycle fix, the dependency-cycle check, and the real cmux reproduction.

I am closing this stale branch rather than asking you to revive a head that is now 2,463 commits behind `dev`. No merged successor contains `CMUX_AGENT_LAUNCH_KIND`, so the underlying fix is still welcome in a fresh PR.

Please tighten one failure path in that successor. When neither `OPENCODE_BIN_PATH` nor `bunWhich("opencode")` resolves an absolute executable, this head falls back to bare `opencode`; pane creation can still report success, and focus then returns `true` without checking whether an attach process survived:
https://github.com/code-yeongyu/oh-my-openagent/blob/971cf82316da8be194928d9ee76c613ad50714c9/packages/tmux-core/src/tmux-utils/pane-spawn.ts#L80-L96
https://github.com/code-yeongyu/oh-my-openagent/blob/971cf82316da8be194928d9ee76c613ad50714c9/packages/tmux-core/src/tmux-utils/pane-activate.ts#L28-L39

A current successor should fail closed or verify the attached pane, cover resolver failure and the official marker's initial environment gate, run current CI, and commit isolated real cmux/OpenCode evidence under `.omo/evidence/`. Happy to review that refreshed implementation.

PR #6524, opened 2026-07-31
fix(ci): resolve audit root path on Windows

code-yeongyu:
## Summary

Use `fileURLToPath()` instead of a raw file URL pathname when deriving the repository root in the GPT Mini reference audit.

## Why

Windows CI passed the repository root as `/D:/a/...` to `Bun.spawn`, causing:

```text
ENOENT: no such file or directory
at trackedFiles (script/gpt-mini-reference-audit.test.ts:30)
```

The identical failure occurs on multiple unrelated PRs, including #6521 and #6522, so this is a base CI defect rather than a feature regression.

## Verification

- failing-first test: `repositoryRootFromMetaUrl is not defined`
- focused test: `2 pass, 0 fail`
- `bun run typecheck:script`: clean
- the new test uses `pathToFileURL()` with a platform-native path containing a space, so it exercises URL decoding and Windows drive-path conversion through the host platform

---
## Summary by cubic
Fixes repository root resolution in the GPT Mini reference audit so it works on Windows CI. Prevents ENOENT errors caused by `/D:/...` paths when spawning processes.

- **Bug Fixes**
  - Compute repo root with `fileURLToPath()` from `import.meta.url` instead of using the raw pathname.
  - Add a test that covers platform-native file URLs with spaces and Windows drive-letter paths.

<sup>Written for commit f4b1b0f935c44e45aa389c67ecbd873d82536af1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6524?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/script/gpt-mini-reference-audit.test.ts
+++ b/script/gpt-mini-reference-audit.test.ts
@@ -1,11 +1,26 @@
 import { describe, expect, test } from "bun:test"
 import { readFile, stat } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import { join, sep } from "node:path"
+import { fileURLToPath, pathToFileURL } from "node:url"
 
-const REPOSITORY_ROOT = new URL("../", import.meta.url).pathname
+const REPOSITORY_ROOT = repositoryRootFromMetaUrl(import.meta.url)
 const OLD_MODEL_ID = ["gpt-5.4", "mini"].join("-")
 const OLD_DISPLAY_NAME = ["GPT 5.4", "Mini"].join(" ")
 
 describe("GPT Mini reference audit", () => {
+  test("repository root decodes a platform-native file URL", () => {
+    // given
+    const repositoryDir = join(tmpdir(), "gpt mini audit")
+    const scriptUrl = pathToFileURL(join(repositoryDir, "script", "audit.test.ts")).href
+
+    // when
+    const root = repositoryRootFromMetaUrl(scriptUrl)
+
+    // then
+    expect(root).toBe(`${repositoryDir}${sep}`)
+  })
+
   test("tracked shipped files no longer reference the retired GPT Mini model", async () => {
     // given
     const files = await trackedFiles()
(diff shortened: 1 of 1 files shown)


miny6121-sys:
Follow-up from the reporter, with **a correction to my own issue body** and a fresh recurrence.

## Correction: `promptSource` is NOT ground truth

In the original report I wrote:

> The ground truth is available: a real user prompt is a separate JSONL record with `type: "user"` and a `promptSource` field. The fabricated blocks have no such record.

**That is wrong.** A fabricated block *does* subsequently appear as a `user` record carrying `promptSource`, because this harness re-invokes the CLI through the SDK on every tool-call handoff, and the replayed transcript (which by then contains the fabrication) is written as an sdk-sourced user prompt.

Current distribution across 598 local session files:

```
sdk     594
system   32
typed    30
```

Almost everything is `sdk`, so `promptSource` cannot separate "a human typed this" from "the host replayed a transcript that happened to contain a fabricated USER: block".

**The reliable discriminator is temporal ordering:** does the text first appear *inside* an `assistant` record, before any `user` record contains it? If yes, the model authored it.

Anyone building detection on top of this issue should use that test, not `promptSource`.

## Recurrence, minutes after filing

| time (UTC) | record | fact |
|---|---|---|
| 13:55:59 | `assistant` in session `2a442795` | fabricated block at offset 14125 of a 14147-char record |
| 13:57:00 | `user` in session `08a6e12c` | same text returns with `promptSource: sdk` |

The text existed inside the assistant's own output **61 seconds before** any user record contained it. No `queue-operation` record carries it either (0 of 1299 queue-ops, of which 101 carry `content`), so it never passed through the input queue.

This one is Pattern A again - a fabricated continuation approval (`ㅇㅋ 다시 이관검증 하던걸로 돌아가서 계속해줘` / "ok, go back to the migration verification and continue"). The user caught it and said they never sent it.

That makes **7 occurrences** now, and it recurred within the same session in which the issue was filed, so the rate is not negligible in long tool-heavy sessions.

The suggested mitigation #1 from the issue body (strip a trailing `\nUSER:` block when persisting an assistant turn) would have caught this one too.

Issue #6547, opened 2026-08-01
[Bug]: stdio MCP servers never exit when a live parent abandons stdin — the idle timeout fires but tears nothing down (follow-up to #5560)


lifrary:
## Bug Description

MCP stdio servers never exit when the parent process stays alive but stops reading from — and never closes — the child's stdin. Every teardown path in `mcp-stdio-core` requires the parent to either die or release the write end, so a host that keeps the fd open and simply forgets the child leaks one server process per session, forever.

This is the case left open by #5560. That issue and its fix cover the parent *dying*: stdin `close`, `ERR_STREAM_PREMATURE_CLOSE`, forced exit after the loop, and later `createParentWatchdog` (4.19.1). None of them fire here — the parent is alive and holding the pipe.

There are two halves to it.

## Half 1 — the idle path is structurally inert

`packages/mcp-stdio-core/src/server.ts` on current `dev`:

```ts
const idleTimer = createIdleTimer(idleTimeoutMs, log, () => {
  isClosed = true
  void config.onIdleTimeout?.()          // <-- no teardown
})
const watchdog = createParentWatchdog(config.parentWatchdog, (parentPid, pollIntervalMs) => {
  isClosed = true
  log("parent_exit", { parent_pid: parentPid, poll_interval_ms: pollIntervalMs })
  void config.onParentExit?.()
  config.input.destroy()                 // <-- teardown lives only here
})
...
for await (const message of readStdioJsonRpcMessages(config.input)) {
  if (isClosed) break                    // <-- only reached when a message arrives
```

`isClosed` cannot end the read loop on its own: it is checked *after* a message arrives, and an idle server is idle precisely because none arrives. The timer arms and fires correctly — it just has no effect. The parent-exit callback's own comment says it destroys the input "so the server settles through the same finally path as an idle timeout", but the idle timeout never destroys anything.

Attached PR fixes this with one line and adds the first two tests that exercise the idle path (in-process, plus a real spawned child — a settled read loop still leaves a live process if anything keeps the event loop alive). Every pre-existing test in `server.test.ts` sets `idleTimeoutMs: 0`, which is why the gap survived.

## Half 2 — three consumers disable the idle timeout entirely

Even with Half 1 fixed, only codegraph benefits:

| call site | `idleTimeoutMs` |
|---|---|
| `packages/omo-codex/plugin/components/codegraph/src/mcp-bridge.ts` | default (10 min) |
[...]

Issue #5560, opened 2026-06-24
[Bug]: lsp-daemon MCP proxy processes accumulate as orphans across sessions


PaoloC68:
## Bug Description

When opencode sessions end (normally or via kill), the `lsp-daemon` MCP proxy processes (`cli.js mcp`) are not terminated. They accumulate across sessions because dangling daemon socket handles keep the Node event loop alive, even after stdin closes.

Over a few days of use, this results in 5-10+ orphaned proxy processes plus their associated language server children (basedpyright, pyright, typescript-language-server), consuming 1+ GB of RAM.

## Steps to Reproduce

1. Start an opencode session (spawns `cli.js mcp` proxy)
2. Use LSP tools (diagnostics, etc.) — this creates daemon socket connections
3. Exit the opencode session
4. Run `ps aux | grep lsp-daemon` — the proxy process is still alive
5. Repeat steps 1-4 several times
6. Observe accumulating orphaned processes

## Expected Behavior

When the parent opencode process exits and stdin closes, the proxy process should terminate promptly.

## Actual Behavior

The proxy survives indefinitely because:
1. **No stdin `close` listener** — when the parent dies (especially via SIGKILL), stdin closes but the `for await` loop in `runJsonRpcStdioServer` may not exit if a daemon request is in-flight
2. **No `ERR_STREAM_PREMATURE_CLOSE` handling** — if stdin is destroyed mid-request, the error propagates unhandled instead of triggering graceful exit
3. **No forced exit after server loop** — even when the loop returns normally, dangling daemon socket handles can keep the event loop alive

After 3 days of normal use: 6 orphaned proxy processes + stale basedpyright (768 MB), pyright (475 MB), and typescript-language-server processes.

## Doctor Output

```
N/A — issue is observable via process listing
```

## Error Logs

```
$ ps aux | grep lsp-daemon | grep -v grep
paolo  563131  Jun21  node cli.js mcp   # orphaned
paolo  565731  Jun21  node cli.js mcp   # orphaned
paolo  674466  Jun22  node cli.js mcp   # orphaned
paolo  675212  Jun22  node cli.js mcp   # orphaned
paolo  676701  Jun22  node cli.js mcp   # orphaned
paolo 1375715  19:01  node cli.js mcp   # current session (also broken stdin)
```

## Additional Context

Fix submitted as PR — three changes to `packages/lsp-daemon/src/proxy.ts`:
[...]

PR #6548, opened 2026-08-01
fix(mcp-stdio-core): destroy stdin on idle timeout so abandoned servers exit

lifrary:
Addresses half of #6547.

## Problem

`runJsonRpcStdioServer`'s idle timeout sets `isClosed` and calls the hook, but never tears the read loop down. `isClosed` is only consulted after a message arrives (`for await ... if (isClosed) break`), and an idle server is idle precisely because none arrives — so the loop parks on the open pipe and the idle exit has never been able to fire.

Every other teardown path needs the parent to release the write end: stdin `close`/`end` (#5560) and `createParentWatchdog` (parent death). A parent that stays alive while abandoning the pipe hits none of them, and the child lives forever.

## Change

- `server.ts`: `config.input.destroy()` in the idle callback, mirroring the parent-exit callback six lines below. The resulting `ERR_STREAM_PREMATURE_CLOSE` is already swallowed by the existing catch because `isClosed` is set, so no new error handling is needed.
- `server.test.ts`: the first two tests that exercise the idle path. The second one spawns a real child process, because a settled read loop still leaves a live process if anything keeps the event loop alive — and the process, not the promise, is what users observe.

No call site in the repo passes `onIdleTimeout`, so nothing depends on the old no-op behaviour.

## Scope — what this does not fix

This only takes effect where an idle timeout is enabled. **Corrected 2026-08-02 (see comment below):** there are four call sites, and the only one taking the 10-minute default is `packages/omo-codex/plugin/components/codegraph/src/mcp-unavailable.ts`, the degraded stub; `lsp-core/src/mcp.ts`, `lsp-daemon/src/proxy.ts` and `git-bash-mcp/src/mcp.ts` all pass `idleTimeoutMs: 0`, and `createIdleTimer` treats `<= 0` as disabled. `mcp-bridge.ts` is not a call site at all and has no idle timer. My original wording here claimed codegraph's servers were fixed by this; that was wrong. Whether those should adopt a nonzero idle window is a separate call for the maintainers — see the issue.

## Verification

Both tests were written first and watched fail against unpatched `dev`:

```
(fail) idle timeout > #given a live parent that abandons stdin without closing it ... [1001ms]
(fail) idle timeout > #given a real child whose parent holds stdin open and never writes ... [5001ms]
 16 pass / 2 fail
```

The first fails with `Expected: "settled" / Received: "hung"` while its `onIdleTimeout` hook did fire — the shape of the bug. With the one-line change, `bun test src/*.test.ts` in `packages/mcp-stdio-core`:

```
 22 pass
 0 fail
```

Not run: `typecheck` (needs `tsgo` from a workspace install this environment could not afford) and the consumer packages' suites. The change is confined to one callback and adds no new types.

---
## Summary by cubic
Destroy stdin on idle timeout in `mcp-stdio-core` so abandoned JSON-RPC stdio servers exit instead of hanging on an open pipe. Adds tests for the idle path, including a real child process case, and shares child-spawn setup across tests.

- **Bug Fixes**
  - In `runJsonRpcStdioServer`, the idle timeout now calls `config.input.destroy()` after marking closed, tearing down the read loop; the resulting error is already swallowed.
  - Tests added: one in-process to confirm the server settles, and one that spawns a real child to confirm the process exits on idle.
  - Only applies when `idleTimeoutMs > 0`; callers passing `0` are unchanged.

- **Refactors**
  - Extracted `buildServerScript` and `spawnServerChild` (plus `spawnIdleChild`) so watchdog and idle tests share one spawn setup.
  - Dropped the unused handler param in the idle test to fix typecheck; unified env handling in child spawns.

<sup>Written for commit d2428a765b902c8b2581bb9dd7f24b52413c09f9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/mcp-stdio-core/src/server.test.ts
+++ b/packages/mcp-stdio-core/src/server.test.ts
@@ -316,6 +316,52 @@ describe("parent watchdog", () => {
   })
 })
 
+describe("idle timeout", () => {
+  test("#given a live parent that abandons stdin without closing it #when the idle timeout elapses #then the server settles instead of parking on the open pipe", async () => {
+    // The parent is alive and still holds the write end, so stdin never emits
+    // 'end' and the parent watchdog never fires. Only the idle timeout can end
+    // this server, and it must tear the read loop down rather than flag it.
+    const input = new PassThrough()
+    const output = new PassThrough()
+    let idleFired = false
+
+    const served = runJsonRpcStdioServer({
+      input,
+      output,
+      handlerOptions: undefined,
+      idleTimeoutMs: 20,
+      handler: async (message) => successResponse(message.id, { acknowledged: true }),
+      onIdleTimeout: () => {
+        idleFired = true
+      },
+    })
+
+    const outcome = await Promise.race([
+      served.then(() => "settled" as const),
+      Bun.sleep(1_000).then(() => "hung" as const),
+    ])
+
+    expect(idleFired).toBe(true)
+    expect(outcome).toBe("settled")
+  })
+
+  test("#given a real child whose parent holds stdin open and never writes #when the idle timeout elapses #then the child process exits", async () => {
+    // The in-process test above only proves the promise settles. A settled loop
+    // still leaves a live process if anything keeps the event loop alive, so
+    // assert the exit itself — the symptom users actually observe.
+    const child = spawnIdleChild(300)
+    try {
+      const outcome = await Promise.race([
+        child.exited.then(() => "exited" as const),
+        Bun.sleep(5_000).then(() => "alive" as const),
+      ])
+      expect(outcome).toBe("exited")
+    } finally {
+      killQuietly(child.pid)
+    }
+  })
+})
+
 describe("isProcessAlive", () => {
   test("#given a running process #when probed #then it reports alive", () => {
     expect(isProcessAlive(process.pid)).toBe(true)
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Approved the workflow run — CI is red on typecheck across all three OSes, but it is a single line in the new test file, not the fix itself:

```
packages/mcp-stdio-core/src/server.test.ts(333,51): error TS18046: 'message' is of type 'unknown'.
```

The `server.ts` change is fine. Just needs a type narrow/annotation on `message` at that line. Same failure on ubuntu/macos/windows, so it is purely the type error, nothing platform-specific.

The idle-path analysis in #6547 checks out against current `dev` — I confirmed `isClosed` is only read after a message arrives, and that all 6 `idleTimeoutMs` occurrences in the existing `server.test.ts` are `0`, which is exactly why this never got caught.

lifrary:

lifrary:
Typecheck is fixed: that test's premise is a parent that never writes, so the handler is never invoked, and dropping its `unknown`-typed `message` param matches the file's other call sites without needing a cast. Also extracted the shared child-spawn setup cubic flagged. Verified locally: repo-wide `tsgo --noEmit` clean, 22 pass / 0 fail. Commits are now unified under one committer and the CLA is signed, so this should just need a workflow-run approval when you get a chance.

lifrary:
Scope correction before this merges — posted in full at #6547.

`mcp-bridge.ts` is not a call site of `runJsonRpcStdioServer`: it mentions it only in a comment at `:79`, calls `createParentWatchdog` directly at `:84`, and has no idle timer. Grepping actual invocations gives four call sites, and the only one taking the 10-minute default is `packages/omo-codex/plugin/components/codegraph/src/mcp-unavailable.ts:31`, the degraded stub.

So this sentence in the PR body is wrong:

> codegraph's two servers take the 10-minute default and are fixed by it

It should read: one call site benefits today — `mcp-unavailable.ts` — plus any future consumer that enables an idle timeout. I have updated the PR body accordingly.

The change itself, its tests, and the Half 1 analysis are unaffected; only my claim about who benefits was wrong.

code-yeongyu:
[sisyphus-bot] The one-line fix correctly mirrors the parent-watchdog teardown path. The in-process and real-child tests cover both the promise-settle and process-exit aspects. Owner has already reviewed and the typecheck issue is resolved. Ready to merge once CI runs.

MoerAI:
Thanks for the idle-teardown fix and its regression coverage.

Closing as superseded by owner-merged #7525, which cherry-picked both #6548 commits with preserved authorship, completed the host-specific follow-through, and closed #6547:
https://github.com/code-yeongyu/oh-my-openagent/pull/7525
https://github.com/code-yeongyu/oh-my-openagent/issues/6547

This branch now conflicts with `dev`, and its substantive CI jobs were cancelled:
https://github.com/code-yeongyu/oh-my-openagent/actions/runs/30728306752

Rebasing would duplicate landed work. No further action is needed unless the merged behavior regresses on current `dev`; if that happens, please open a focused reproduction against the current implementation.

Issue #5786, opened 2026-06-30
[Bug]: omo doctor misreports TUI plugin missing when tui.json has tuple plugin entries


Burntgogi:
﻿## Summary
`omo doctor` can falsely report `TUI plugin entry missing from tui.json` when `tui.json` contains a valid `"oh-my-openagent"` TUI plugin entry plus any non-string tuple plugin entry such as `["./plugin.tsx", { ... }]`.

The TUI entry is parsed correctly, but the doctor check later calls `.startsWith()` on tuple entries without a type guard. The resulting `TypeError` is swallowed by the broad `try/catch`, causing the check to return an all-false registration state and emit the misleading missing-entry warning.

## Environment
- LazyCodex / Oh My OpenAgent version: `oh-my-openagent@4.14.1`
- OpenCode version: `1.17.11`
- OS: Windows 11 / win32 x64
- Install method: npm dependency under `C:\Users\<USER>\.config\opencode` (`npm install oh-my-openagent@latest`)
- Relevant config:

```json
// opencode.json
{
  "plugin": ["oh-my-openagent"]
}
```

```json
// tui.json
{
  "$schema": "https://opencode.ai/tui.json",
  "plugin": [
    "oh-my-openagent",
    ["./some-local-tui-plugin.tsx", { "label": "custom" }]
  ]
}
```

## Repository Decision
- Target repository: `code-yeongyu/oh-my-openagent`
- Why this belongs there: the failing code is in Oh My OpenAgent's doctor check, not in upstream OpenCode/Codex core.
- Oh My OpenAgent evidence:
  - Runtime package: `C:\Users\<USER>\.config\opencode\node_modules\oh-my-openagent\dist\cli\index.js`
  - Latest source checkout: `packages/omo-opencode/src/cli/doctor/checks/tui-plugin-config.ts` at `code-yeongyu/oh-my-openagent` HEAD `8471324`
  - Source lines in latest checkout show raw `entry.startsWith(...)` calls on `entry: string` helpers, while `plugins` comes directly from parsed JSON and can contain tuple entries.
- Upstream Codex/OpenCode source evidence:
  - Latest `openai/codex` checkout HEAD `0208281` has no matching `tui-plugin-config`, `isNamedTuiPluginEntry`, or Oh My OpenAgent doctor logic.
  - The symptom disappears by patching Oh My OpenAgent's bundled doctor check only; no Codex core behavior needs to change.

## Reproduction
[...]

PR #6556, opened 2026-08-02
fix(installer): guard plugin predicates against tuple entries (#6555)

### LINKED ISSUE #6555: [Bug]: Installer crashes on tuple plugin entry: plugin.startsWith is not a function
## Bug Description

When `opencode.jsonc`'s `plugin` array contains a tuple-style entry `[name, options]` (a format documented and supported by OpenCode at runtime), running `bunx oh-my-openagent@latest install` throws `TypeError: plugin.startsWith is not a function` and aborts before writing any config or adding the omo plugin entry. Tuple entries are perfectly valid to OpenCode itself; only omo's installer predicate calls string methods on them without a type guard.

This is the same class of bug as #5786 (same unguarded `.startsWith` pattern) but in a **different code path**. The #5859 fix only patched the `doctor` check; the installer detection path `isOurPlugin2` (line 88037) and a sibling installer add-plugin path `isOurPlugin` (line 86351) remain unguarded.

## Steps to Reproduce

1. Put a `plugin` array with at least one tuple entry (the documented `[name, options]` format) in `~/.config/opencode/opencode.jsonc`:

```jsonc
{
  "$schema": "https://opencode.ai/config.json",
  "plugin": [
    "opencode-pty",
    "@tarquinen/opencode-dcp@latest",
    "./plugins/caveman/plugin.js",
    [
      "@plannotator/opencode@latest",
      { "workflow": "plan-agent" }
    ]
  ]
}
```

2. Run:

```bash
bunx oh-my-openagent@latest install --no-tui --claude=no --gemini=no --copilot=no --bailian-coding-plan=yes --opencode-zen=yes --vercel-ai-gateway=yes
```

> Note: the crash is in `detectCurrentConfig` parsing the `plugin` array and is **flag-independent** - it reproduces with default flags. The flags above are the user's actual invocation; `--no-tui` selects the non-interactive CLI installer path that hits `runCliInstaller`.

## Expected Behavior

The installer recognizes there is no existing `oh-my-openagent` entry and proceeds to add one. Tuple entries should be skipped by omo's package-name predicates (mirroring how OpenCode core loads them at runtime) rather than crashing.

## Actual Behavior

```
TypeError: plugin.startsWith is not a function. (In 'plugin.startsWith(`${PLUGIN_NAME}@`)', 'plugin.startsWith' is undefined)
      at isOurPlugin2 (/tmp/bunx-1000-oh-my-openagent@latest/node_modules/oh-my-openagent/dist/cli/index.js:88037:53)
      at find (...)
      at findOurPluginEntry2 (.../dist/cli/index.js:88040:22)
      at detectCurrentConfig (.../dist/cli/index.js:88071:45)
      at runCliInstaller (.../dist/cli/index.js:101352:53)
      at install3 (.../dist/cli/index.js:103216:69)
```

The installer aborts inside `detectCurrentConfig` before detecting an existing `oh-my-openagent` entry. No config written, no plugin added.

## Doctor Output

`bunx oh-my-openagent@latest doctor` **runs cleanly** against the same `opencode.jsonc` (confirming the bug is isolated to the `install` path - `doctor` does not call `detectCurrentConfig`/`isOurPlugin2`):

```
oMoMoMoMo Doctor

⚠ 34 issues found:

1. Invalid configuration
   profiles.default: Activated omo profile "default" does not exist; using the base configuration
   Affects: plugin startup

2. Deprecated reasoning config key
   /home/jin/.omo/omo.jsonc: [opencode].agents.sisyphus.variant
   Fix: Replace variant with reasoning, or run: oh-my-openagent config migrate
   ...

[33 more issues, all "Deprecated reasoning config key" warnings on .omo/omo.jsonc
 from a prior omo install - NONE relate to plugin tuples or this crash]
```

The 34 warnings are pre-existing config drift (deprecated `variant`/`fallback_models` keys) unrelated to this bug.

## Error Logs

Already shown in **Actual Behavior** above - the full stack trace from the installer crash.

## Configuration

Actual `opencode.jsonc` plugin array that triggers the crash (minimized options shown in Steps to Reproduce; the real tuple carries additional options - the crash is option-agnostic, any `[string, object]` entry triggers it):

```jsonc
{
  "$schema": "https://opencode.ai/config.json",
  "plugin": [
    "opencode-pty",
    "@tarquinen/opencode-dcp@latest",
    "@mohak34/opencode-notifier@latest",
    "./plugins/caveman/plugin.js",
    [
      "@plannotator/opencode@latest",
      {
        "workflow": "plan-agent",
        "planningAgents": ["plan", "Sisyphus - ultraworker", "Prometheus - Plan Builder"]
      }
    ]
  ]
}
```

## Additional Context

- OS: NixOS (Linux x64) - selected "Linux" in the dropdown as NixOS isn't listed.
- Bun: `1.3.13`
- Runtime package: `/tmp/bunx-1000-oh-my-openagent@latest/node_modules/oh-my-openagent/dist/cli/index.js`

### Evidence

**Site 1 (the one that actually throws in this repro) - installer detection path**, `dist/cli/index.js:88036-88041`:

```js
function isOurPlugin2(plugin) {
  return plugin === PLUGIN_NAME || plugin.startsWith(`${PLUGIN_NAME}@`) || plugin === LEGACY_PLUGIN_NAME || plugin.startsWith(`${LEGACY_PLUGIN_NAME}@`);
}
function findOurPluginEntry2(plugins) {
  return plugins.find(isOurPlugin2) ?? null;
}
```

`detectCurrentConfig` at line 88071 passes the raw parsed `plugin` array (which may contain tuple entries) into `findOurPluginEntry2`; the first tuple entry hits the unguarded `.startsWith` call and crashes.

**Site 2 (latent in this repro - not reached because Site 1 aborts first) - installer add-plugin path**, `dist/cli/index.js:86348-86355`. This is from `packages/omo-opencode/src/cli/config-manager/add-plugin-to-opencode-config.ts` (the installer's add-plugin step), **not** a runtime loader:

```js
function isPackageOmoPluginEntry(plugin) {
  return plugin === PLUGIN_NAME || plugin.startsWith(`${PLUGIN_NAME}@`) || plugin === LEGACY_PLUGIN_NAME || plugin.startsWith(`${LEGACY_PLUGIN_NAME}@`);
}
function isOurPlugin(plugin) {
  return isPackageOmoPluginEntry(plugin) || isSourceOmoPluginEntry(plugin);
}
function findOurPluginEntry(plugins) {
  return plugins.find(isOurPlugin);
}
```

If only Site 1 (`isOurPlugin2`) is guarded, the installer proceeds past detection and would then hit Site 2 (`isOurPlugin`) at the add-plugin step - same crash, different line. Both sites share the unguarded pattern and both need the `typeof` guard.

### Relationship to #5786

#5786 reported the same unguarded `.startsWith` pattern in the `omo doctor` TUI plugin check; fixed in PR #5859 (merge `707c47978`) by adding `typeof plugin === "string"` guards mirroring how the runtime loader skips tuples. That patch only covered the `doctor` check.

This issue is a **distinct** instance:
- **Different command**: `install` (aborts hard) vs `doctor` (swallowed-by-catch false warning).
- **Different config file**: `opencode.jsonc` vs `tui.json`.
- **Different predicate**: `isOurPlugin2` (line 88037) vs the doctor's TUI check.
- **Hard abort** here vs misleading warning in #5786.

The same unguarded pattern persists in `isOurPlugin2` (installer detection, line 88037) and `isOurPlugin` (installer add-plugin, line 86351). Neither was touched by #5859.

### Suggested Fix

Apply the same `typeof plugin === "string"` guard from #5859 to `isOurPlugin2` and `isPackageOmoPluginEntry` (the helper inside `isOurPlugin`). Minimal:

```js
function isOurPlugin2(plugin) {
  if (typeof plugin !== "string") return false;
  return plugin === PLUGIN_NAME || plugin.startsWith(`${PLUGIN_NAME}@`) || plugin === LEGACY_PLUGIN_NAME || plugin.startsWith(`${LEGACY_PLUGIN_NAME}@`);
}
```

(Same one-liner guard in `isPackageOmoPluginEntry`.)

### Workaround

Before running the installer, temporarily flatten tuple entries to plain strings, restore afterward:

```jsonc
"plugin": [
  "opencode-pty",
  "@plannotator/opencode@latest"   // plain string during install
]
```

Restore `[name, options]` tuple after installation completes.

## Operating System

Linux (NixOS)

## OpenCode Version

1.18.10

tbontb-iaq:
## Summary

- The installer crashed with `TypeError: plugin.startsWith is not a function` whenever `opencode.jsonc`'s `plugin` array contained a tuple-style `[name, options]` entry (a format OpenCode documents and supports at runtime). The crash aborted the install before any config was written.
- Root cause: omo's plugin-name predicates called `.startsWith` / `.toLowerCase` on every entry without a `typeof` guard. This is the same class of bug as #5786 (fixed in #5859 for the `omo doctor` TUI check), but #5859 only patched the doctor path — the installer detection (`isOurPlugin2`) and installer add-plugin (`isOurPlugin` / `isPackageOmoPluginEntry` / `isSourceOmoPluginEntry`) paths remained unguarded.
- After guarding the predicates, the jsonc plugin-array rewriter also mangled tuples: the non-greedy regex `\[([\s\S]*?)\]` stopped at the tuple's inner `]`, truncating the array and emitting a stray `]` (invalid JSON); and the per-entry serializer template-stringified tuples into `"name,[object Object]"`. Both are fixed so the install now succeeds end-to-end on jsonc configs with tuples.

## Changes

- **`parse-opencode-config-file.ts`** — Widened `OpenCodeConfig.plugin` from `string[]` to `(string | [string, unknown])[]` to reflect the documented runtime shape (mirrors the local `OpenCodeConfigShape` already used by the doctor check in `tui-plugin-config.ts`). Only two type-level consumers exist (`detect-current-config.ts` reads `.plugin`; `add-plugin-to-opencode-config.ts` creates `{ plugin: [stringEntry] }`, still valid), so the widening is safe.
- **`detect-current-config.ts`** — `isOurPlugin` / `findOurPluginEntry` converted to `plugin is string` type guards with `typeof plugin === "string"` guards, matching #5859's precedent.
- **`add-plugin-to-opencode-config.ts`** — Same `typeof` guard applied to `isSourceOmoPluginEntry`, `isPackageOmoPluginEntry`, `isOurPlugin`, `findOurPluginEntry`. Added `findPluginArrayBounds`, a bracket-balanced scanner that respects strings and jsonc comments, replacing the non-greedy regex. Per-entry serialization switched from `` `"${p}"` `` to `JSON.stringify(p)` (identical for strings; correct for tuples).
- **`plugin-detection.test.ts`** — Added regression tests for the four affected surfaces.

## QA & Evidence

- **What was tested:** `bun run typecheck` (package `omo-opencode`)
  **Observed result:** clean, zero errors
  **Artifact:** local run
  **Why sufficient:** confirms the type widening and `plugin is string` guards compile against the real `OpenCodeConfig` consumers
- **What was tested:** `bun test src/cli/config-manager/plugin-detection.test.ts`
  **Observed result:** 17 pass / 0 fail (4 new tuple regression tests green)
  **Artifact:** local run
  **Why sufficient:** the new "detect with tuple only" test reproduces the original `TypeError` crash on the unpatched bundle (red), and passes after the guards (green); the json/jsonc add-plugin tests confirm tuples round-trip structurally and the output re-parses as valid JSON
- **What was tested:** `bun test src/cli/config-manager/ src/cli/tui-installer.test.ts src/cli/cli-installer.test.ts src/cli/doctor/ src/cli/cli-program.test.ts src/cli/install.test.ts`
  **Observed result:** 212 pass / 2 fail — the 2 failures (`config.test.ts` › "does not flag reasoningEffort: 'max'...") are pre-existing on clean `dev` (verified by `git stash` + re-run), unrelated to plugins/tuples
  **Artifact:** local run
  **Why sufficient:** zero new failures introduced across the installer, tui-installer, doctor checks, and config-manager suites
- **What was tested:** confirmed the crash on the published `oh-my-openagent@4.19.4` bundle (issue #6555), and that `bunx oh-my-openagent doctor` runs clean against the same tuple-bearing `opencode.jsonc` (isolating the bug to the install path)

## Risks & Residuals

- **Risk: balanced scanner diverges from old regex on edge jsonc** — Mitigated. The scanner produces byte-identical output to the old regex for the all-string case (covered by the existing "rewrites quoted jsonc plugin field in place" test, still green), and correctly handles tuples, strings, line/block comments, and escaped quotes. Trailing commas inside the plugin array are normalized away (the old regex did the same).
- **Risk: type widening ripples to other consumers** — Mitigated. Grep confirmed only two files reference the `OpenCodeConfig` type or read `.plugin` from `parseOpenCodeConfigFileWithError`; both are updated. The doctor check already used an equivalent local shape.
- **Risk: `JSON.stringify` per-entry changes string output** — Not applicable. `JSON.stringify("x")` === `"\"x\""` (the old template output), verified by the passing all-string regression tests.
- **Accepted:** the two pre-existing `config.test.ts` failures are out of scope for this fix.

## Automated Checks

```bash
cd packages/omo-opencode
bun run typecheck
bun test src/cli/config-manager/plugin-detection.test.ts
bun test src/cli/config-manager/ src/cli/tui-installer.test.ts src/cli/cli-installer.test.ts src/cli/doctor/ src/cli/cli-program.test.ts src/cli/install.test.ts
```

Note: the repo-wide `bun test` was not run in full here (broad suite, some tests make network calls); the targeted suites covering every code path touched by this change all pass. The two `config.test.ts` failures are pre-existing on `dev` and unrelated.

## Related Issues

Closes #6555

---
## Summary by cubic
Fixes installer crash when the `plugin` array contains tuple entries `[name, options]`. Also corrects JSONC rewriting to target the root `plugin` array and preserve tuples when adding `oh-my-openagent`. Closes #6555.

- **Bug Fixes**
  - Guarded plugin predicates with `typeof plugin === "string"` and widened `OpenCodeConfig.plugin` to `(string | [string, unknown])[]`.
  - Replaced the JSONC array regex with a bracket-balanced, depth-aware scanner that finds the root `plugin` array and respects strings/comments.
  - Used `JSON.stringify` per entry to keep tuples (no `"name,[object Object]"`).
  - Centralized `isPackageOmoPluginEntry` to avoid drift between detect/add paths; added regression tests for tuples, nested `plugin` keys, and comment-with-`plugin:` cases.

<sup>Written for commit 30d6ddbb76433445aed064ef3398dcb98b08842c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/cli/config-manager/add-plugin-to-opencode-config.ts
+++ b/packages/omo-opencode/src/cli/config-manager/add-plugin-to-opencode-config.ts
@@ -1,13 +1,13 @@
 import { existsSync, readdirSync, readFileSync, writeFileSync } from "node:fs"
 import { basename, dirname, join } from "node:path"
 import type { ConfigMergeResult } from "../types"
-import { PLUGIN_NAME, LEGACY_PLUGIN_NAME } from "../../shared"
+import { PLUGIN_NAME } from "../../shared"
 import { backupConfigFile } from "./backup-config"
 import { getConfigDir } from "./config-context"
 import { ensureConfigDirectoryExists } from "./ensure-config-directory-exists"
 import { formatErrorWithSuggestion } from "./format-error-with-suggestion"
 import { detectConfigFormat, type ConfigFormat } from "./opencode-config-format"
-import { parseOpenCodeConfigFileWithError, type OpenCodeConfig } from "./parse-opencode-config-file"
+import { isPackageOmoPluginEntry, parseOpenCodeConfigFileWithError, type OpenCodeConfig } from "./parse-opencode-config-file"
 import { getPluginNameWithVersion } from "./plugin-name-with-version"
 import { checkVersionCompatibility, extractVersionFromPluginEntry } from "./version-compatibility"
 
@@ -68,23 +68,21 @@ function getConfigTargets(): ConfigTarget[] {
   })
 }
 
-function isSourceOmoPluginEntry(plugin: string): boolean {
+function isSourceOmoPluginEntry(plugin: unknown): plugin is string {
+  if (typeof plugin !== "string") return false
   const normalized = plugin.toLowerCase().replaceAll("\\", "/")
   if (!normalized.startsWith("file://")) return false
 
   return /\/(omo(?:-[^/]*)?|oh-my-opencode|oh-my-openagent)\/(src|dist)\/index\.(ts|js)$/.test(normalized)
 }
 
-function isPackageOmoPluginEntry(plugin: string): boolean {
-  return plugin === PLUGIN_NAME || plugin.startsWith(`${PLUGIN_NAME}@`) ||
-    plugin === LEGACY_PLUGIN_NAME || plugin.startsWith(`${LEGACY_PLUGIN_NAME}@`)
-}
-
-function isOurPlugin(plugin: string): boolean {
+// `isPackageOmoPluginEntry` is shared from `parse-opencode-config-file.ts`;
+// the add-plugin path additionally matches `file://` source entries.
+function isOurPlugin(plugin: unknown): plugin is string {
   return isPackageOmoPluginEntry(plugin) || isSourceOmoPluginEntry(plugin)
 }
 
-function findOurPluginEntry(plugins: readonly string[]): string | undefined {
+function findOurPluginEntry(plugins: readonly (string | [string, unknown])[]): string | undefined {
   return plugins.find(isOurPlugin)
 }
 
@@ -94,6 +92,99 @@ function findSourcePluginEntryInTarget(target: ConfigTarget): string | null {
   const parseResult = parseOpenCodeConfigFileWithError(target.path)
   const plugins = parseResult.config?.plugin ?? []
   return plugins.find(isSourceOmoPluginEntry) ?? null
+}
+
+// Locate the bounds of the ROOT-level `plugin` array in a jsonc document.
+// The previous non-greedy regex `\[([\s\S]*?)\]` had two defects this fixes:
+//   1. It stopped at the first `]`, truncating the array whenever an entry
+//      was itself a nested array (the documented `[name, options]` tuple
+//      form), yielding a stray `]` and invalid JSON.
+//   2. It matched the first textual `plugin:` regardless of nesting depth,
+//      so a `plugin` key nested inside another object (e.g. an MCP server
+//      config) appearing before the root array would be rewritten instead,
+//      leaving the active root plugin array unchanged and omo unregistered.
+// The scanner balances brackets while respecting strings and jsonc comments,
+// and only accepts a `plugin:` header at object depth 1 (the root object).
+function findPluginArrayBounds(content: string): { readonly start: number; readonly end: number } | null {
+  const headerRegex = /(?:"plugin"|plugin)\s*:\s*(?=\[)/g
+  let headerMatch: RegExpExecArray | null
+  while ((headerMatch = headerRegex.exec(content)) !== null) {
+    if (objectDepthAt(content, headerMatch.index) === 1) {
+      const start = headerMatch.index + headerMatch[0].length
+      const end = balanceArrayBrackets(content, start)
+      if (end !== null) return { start, end }
+    }
+  }
+  return null
+}
+
+// Count `{`/`}` depth over [0, pos), skipping jsonc strings and comments.
+// depth 0 = outside any object; depth 1 = inside the root object; >=2 nested.
+function objectDepthAt(content: string, pos: number): number {
+  let depth = 0
+  let i = 0
+  let inString = false
+  while (i < pos) {
+    const ch = content[i]
+    if (inString) {
+      if (ch === "\\") { i += 2; continue }
+      if (ch === '"') inString = false
+      i += 1
+      continue
+    }
+    if (ch === '"') { inString = true; i += 1; continue }
+    if (ch === "/" && content[i + 1] === "/") {
+      const nl = content.indexOf("\n", i)
+      i = nl === -1 ? pos : Math.min(nl + 1, pos)
+      continue
+    }
+    if (ch === "/" && content[i + 1] === "*") {
+      const close = content.indexOf("*/", i + 2)
+      i = close === -1 ? pos : Math.min(close + 2, pos)
+      continue
+    }
+    if (ch === "{") depth += 1
+    else if (ch === "}") depth -= 1
+    i += 1
+  }
+  return depth
+}
+
+// Given `start` pointing at `[`, return index after the matching `]` (or null
+// if unbalanced), skipping jsonc strings and comments and balancing nested
+// arrays (e.g. tuple entries `[name, {...}]`).
+function balanceArrayBrackets(content: string, start: number): number | null {
+  if (content[start] !== "[") return null
+  let depth = 0
+  let i = start
+  let inString = false
+  while (i < content.length) {
+    const ch = content[i]
+    if (inString) {
+      if (ch === "\\") { i += 2; continue }
+      if (ch === '"') inString = false
+      i += 1
+      continue
+    }
+    if (ch === '"') { inString = true; i += 1; continue }
+    if (ch === "/" && content[i + 1] === "/") {
+      const nl = content.indexOf("\n", i)
+      i = nl === -1 ? content.length : nl + 1
+      continue
+    }
+    if (ch === "/" && content[i + 1] === "*") {
+      const close = content.indexOf("*/", i + 2)
+      i = close === -1 ? content.length : close + 2
+      continue
+    }
+    if (ch === "[") depth += 1
+    else if (ch === "]") {
+      depth -= 1
+      if (depth === 0) return i + 1
+    }
+    i += 1
+  }
+  return null
 }
 
 function choosePluginEntry(params: {
@@ -177,12 +268,11 @@ function writePluginEntryToTarget(params: {
 
     if (target.format === "jsonc") {
       const content = readFileSync(target.path, "utf-8")
-      const pluginArrayRegex = /((?:"plugin"|plugin)\s*:\s*)\[([\s\S]*?)\]/
-      const match = content.match(pluginArrayRegex)
+      const bounds = findPluginArrayBounds(content)
 
-      if (match) {
-        const formattedPlugins = normalizedPlugins.map((p) => `"${p}"`).join(",\n    ")
-        const newContent = content.replace(pluginArrayRegex, `$1[\n    ${formattedPlugins}\n  ]`)
+      if (bounds) {
+        const formattedPlugins = normalizedPlugins.map((p) => JSON.stringify(p)).join(",\n    ")
+        const newContent = content.slice(0, bounds.start) + `[\n    ${formattedPlugins}\n  ]` + content.slice(bounds.end)
         writeFileSync(target.path, newContent)
       } else {
         const newContent = content.replace(/(\{)/, `$1\n  "plugin": ["${nextPluginEntry}"],`)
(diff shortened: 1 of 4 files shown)


tbontb-iaq:

tbontb-iaq:
Thanks for the review. Both issues addressed in `30d6ddb`.

## P1 — nested `plugin` key targets wrong array

Valid. The first-match `headerRegex.exec` rewrote the first textual `plugin: [`, so a nested `plugin` key (e.g. inside an MCP server object) appearing before the root array got rewritten while the active root array stayed unchanged — install reported success with omo never registered.

Fix: `findPluginArrayBounds` is now depth-aware.
- `objectDepthAt(content, pos)` counts `{`/`}` depth over `[0, pos)`, skipping jsonc strings and comments. depth 0 = outside root object; 1 = root object; ≥2 = nested.
- Header regex carries the `g` flag and loops; each candidate header is accepted only when `objectDepthAt(content, headerMatch.index) === 1`. Nested headers are skipped, matching the old regex's behavior of searching forward.
- Body scan extracted to `balanceArrayBrackets(content, start)` (unchanged logic, returns index after the matching `]`).

Regression test added: `addPluginToOpenCodeConfig - tuple plugin entries (#6555) > targets the root plugin array, not a nested plugin key (#6555 P1)`. Config has `"mcp" > "custom-server" > { "plugin": ["nested-should-be-untouched"] }` before the root `"plugin": ["opencode-pty"]`. Verified red→green: against the pre-fix first-match scanner the test fails (nested array rewritten, `"nested-should-be-untouched"` dropped); with the depth-aware scanner it passes (root array gets omo, nested array preserved verbatim).

Note: the nested-key bug pre-existed on `dev` (the old non-greedy regex had the same first-match behavior), but the whole rewrite was bundled in this PR, so fixing it here is the right call — it's on the same code path being hardened.

## P3 — duplicate plugin-matching helpers

Valid. `isPackageOmoPluginEntry` (the 4-line package-name predicate) was byte-for-byte duplicated across `detect-current-config.ts` and `add-plugin-to-opencode-config.ts`.

Fix: `isPackageOmoPluginEntry` is now exported from `parse-opencode-config-file.ts` (next to the `OpenCodeConfig` type it operates on) and imported by both files. AGENTS.md's "no duplicate utilities" guidance satisfied; single source of truth for the package-name match.

Note: `isOurPlugin` and `findOurPluginEntry` are intentionally NOT unified — the two call sites have different semantics:
- `detect-current-config.ts` `isOurPlugin` = package-only (source `file://` entries aren't "installed" for detection).
- `add-plugin-to-opencode-config.ts` `isOurPlugin` = package OR source (so the add step preserves an existing `file://` entry).

The shared predicate is the lowest common factor; the divergent wrappers stay local. The `string | null` vs `string | undefined` return types also differ by call site.

## Verification

```
bun run typecheck                         # clean
bun test src/cli/config-manager/plugin-detection.test.ts   # 19 pass / 0 fail
bun test src/cli/config-manager/ src/cli/tui-installer.test.ts \
  src/cli/cli-installer.test.ts src/cli/doctor/ \
  src/cli/cli-program.test.ts src/cli/install.test.ts      # 252 pass / 2 fail
```

The 2 failures (`config.test.ts` › "does not flag reasoningEffort: 'max'...") are pre-existing on clean `dev` (verified via `git stash` + re-run), unrelated to plugins/tuples.

code-yeongyu:
[sisyphus-bot] The fix is comprehensive — both crash sites guarded, the jsonc rewriter properly replaced with a bracket-balanced scanner, and the nested-key bug caught during review is fixed. Please trigger CI so the test/typecheck workflows run. One note: the `findPluginArrayBounds` scanner is new parsing code — consider adding a fuzz or edge-case test for deeply nested objects with escaped quotes in strings to harden confidence.

MoerAI:
Closing as superseded. Current `dev` already contains the owner implementation in `3a80760`, which handles tuple plugin entries through shared helpers, preserves tuple options, and updates JSONC through the structured edit path.

This older branch is now conflicting and its historical CI contexts are no longer sufficient for current `dev`. Rebasing would duplicate the already-landed installer and doctor coverage.

Please keep #6555 open only if the current owner implementation still reproduces the installer failure in a real OpenCode setup.

PR #6578, opened 2026-08-03
fix(lsp-core): prefer .git as workspace root over nearer sub-package markers

xxsddm:
## Problem

In monorepo-style projects where a sub-package has its own `pyproject.toml` (or other workspace marker), `findWorkspaceRoot()` in `packages/lsp-core/src/lsp/client-wrapper.ts` stopped at the sub-package directory instead of reaching the real project root where `.git` lives.

This caused language servers (e.g. Pyright) to be spawned with `cwd=<sub-package-dir>` instead of the project root. The server could not find the project venv or resolve imports, so it exited immediately with code 0. The `lsp_diagnostics` tool then returned errors or timed out.

### Reproduction (before fix)

On a project with a nested `pyproject.toml` inside a sub-directory (e.g. `packages/sub-pkg/pyproject.toml`), calling `lsp_diagnostics` via `opencode run --format json` on a file inside that sub-directory:

```
"status":"error",
"error":"LSP server pyright at <project>/packages/sub-pkg exited with code null"
```

- Workspace root resolved to the sub-package directory instead of the project root
- Second call: 30-second timeout
- Total: ~46s across two failed attempts

A control project without nested markers works fine on the old code, confirming the issue is specific to the nesting pattern.

## Fix

Prefer `.git` as the authoritative workspace boundary. Walk all the way up looking for `.git` first; only if no `.git` is found, fall back to the first directory matching any other marker (`package.json`, `pyproject.toml`, `Cargo.toml`, `go.mod`, `pom.xml`, `build.gradle`).

**Changed file:** `packages/lsp-core/src/lsp/client-wrapper.ts` (+17/-5)

Key change in `findWorkspaceRoot()`:
- New constant `PRIMARY_MARKER = ".git"`
- `FALLBACK_MARKERS` list excludes `.git` (previously `.git` was in the shared `WORKSPACE_MARKERS` array and skipped via a `continue` check; now cleanly separated)
- Loop checks `.git` first at every level; if found, returns immediately
- Tracks a `fallbackRoot` (first fallback marker found) but does NOT return it until the loop completes without finding `.git`
- Returns `fallbackRoot ?? dirname(abs)` at the end

## Why `.git` as the primary marker is a sound general-purpose design

`.git` is the only workspace marker that is **authoritative and unique** per project. All other markers can appear at multiple levels in a monorepo:

| Marker | Can appear at sub-package level? | Unique per project? |
|---|---|---|
| `.git` | No (only at project root, or submodule/worktree root which is an independent project) | Yes |
| `package.json` | Yes (every npm sub-package) | No |
| `pyproject.toml` | Yes (every Python sub-package) | No |
| `Cargo.toml` | Yes (Cargo workspace members) | No |
| `go.mod` | Yes (Go modules) | No |

### Scenario-by-scenario analysis

**1. Single-package project** (`.git` + `package.json` at the same level)
- Old: stops here → correct
- New: finds `.git` here → correct
- ✅ No behavioral change

**2. No `.git`** (only `package.json` / `pyproject.toml`)
- Old: finds marker → correct
- New: no `.git` found, falls back to marker → correct
- ✅ No behavioral change

**3. Monorepo with nested markers** (the bug being fixed)
```
project/
├── .git/
├── pyproject.toml
└── packages/
    └── sub-pkg/
        ├── pyproject.toml   ← nearer marker
        └── main.py
```
- Old: stops at `packages/sub-pkg/` → language server cannot find venv → crashes
- New: skips sub-pkg, walks to `project/` where `.git` lives → correct
- ✅ Fix works

**4. JS/TS monorepo** (e.g. this very repository)
```
oh-my-openagent/
├── .git/
├── package.json
└── packages/
    └── lsp-core/
        ├── package.json     ← nearer marker
        ├── tsconfig.json
        └── src/...
```
- Old: workspace root = `packages/lsp-core/`
- New: workspace root = `oh-my-openagent/`
- ⚠️ Behavioral change, but **functionally harmless**: TypeScript LSP uses `tsconfig.json` for module resolution, not `cwd`. The workspace root only affects which files the server indexes on startup (slightly broader scope). No breakage observed in testing.

**5. Git submodule / worktree**
```
project/
├── .git/
└── vendor/
    └── submodule/
        ├── .git             ← gitdir pointer file
        └── src/...
```
- New: finds `.git` at `vendor/submodule/` → stops there
- ✅ Correct — a submodule is an independent project with its own root

### Conclusion

`.git` is the strongest signal for "this is a project root." Language servers need the project root for environment resolution (venv, node_modules, Cargo.lock). Module resolution is handled by language-specific config files (`tsconfig.json`, `pyproject.toml`) which are independent of `cwd`. Preferring `.git` as the workspace boundary is a safe, general-purpose improvement.

## Tests

3 regression tests added to `packages/lsp-core/src/lsp/client-wrapper.test.ts` under `describe("findWorkspaceRoot marker priority")`:

1. **`.git` at project root vs nested `pyproject.toml`** — verifies `.git` wins over a nearer `pyproject.toml` in a sub-package
2. **Nested `pyproject.toml` with no `.git`** — verifies fallback to first non-`.git` marker
3. **`.git` and `pyproject.toml` at same level** — verifies both markers agree

### Test results

```
bun test packages/lsp-core/src/lsp/client-wrapper.test.ts

5 pass, 1 fail
```

The 1 failure is a **pre-existing** Windows EPERM symlink permission issue (`symlinkSync` requires admin privileges on Windows), unrelated to this change. The 3 new tests all pass.

### Typecheck

```
node_modules/.bin/tsgo.exe --noEmit -p packages/lsp-core/tsconfig.json
```

Zero errors. (Global typecheck fails only due to an unrelated third-party package 404 on npmmirror.)

### Build

```
bun run build
```

All steps completed. `packages/lsp-daemon/dist/{index,cli,client}.js` contain `PRIMARY_MARKER = ".git"` and `fallbackRoot` logic (verified via `grep -c "PRIMARY_MARKER"` = 3 in each file).

### End-to-end comparison (real project)

**Setup:** A project with `.git` at the root and a nested `pyproject.toml` in a sub-directory.
**Method:** `opencode run --format json` calling `lsp_diagnostics` on a file inside the sub-directory.

| | Old code (before fix) | New code (after fix) |
|---|---|---|
| **Status** | `error` | `completed` |
| **Error** | `pyright at .../sub-pkg exited with code null` | none |
| **Workspace root** | sub-package directory (wrong) | project root (correct, `.git` location) |
| **Response time** | 15.8s fail + 30s timeout | 3.5s |
| **Language server** | crashed immediately | stayed alive |
| **Daemon log** | server exit errors | clean (no errors) |

### Control group (no nested markers)

Tested a project with `.git` + `pyproject.toml` at the same level (no nesting). Both old and new code work correctly, confirming the fix does not regress existing behavior.

## PR Checklist

- [x] Code follows project conventions (kebab-case, barrel exports, given/when/then test style)
- [x] `bun run typecheck` passes (lsp-core package; global failure is pre-existing and unrelated)
- [x] `bun test` passes (5 pass / 1 pre-existing Windows symlink fail)
- [ ] `bun run test:codex` — not applicable (no Codex-side changes)
- [x] Tested locally with OpenCode (end-to-end comparison)
- [x] QA evidence recorded under `.omo/evidence/20260803-lsp-workspace-root-git-priority/`

## Residual risk

The "No diagnostics found" result (rather than actual language server diagnostics) is a **pre-existing** LSP daemon diagnostics-polling issue that affects all languages (TypeScript files also return empty). It is independent of this workspace-root fix — the fix ensures the language server stays alive and the daemon responds within timeout, which is the scope of this change.

### DIFF
--- a/packages/lsp-core/src/lsp/client-wrapper.test.ts
+++ b/packages/lsp-core/src/lsp/client-wrapper.test.ts
@@ -61,3 +61,200 @@ describe("LSP client path confinement", () => {
 		).toThrow(LspInvalidPathError);
 	});
 });
+
+describe("findWorkspaceRoot marker priority", () => {
+	// given a project root with .git and a nested sub-package that has its own pyproject.toml
+	// when resolving workspace root for a file inside the sub-package
+	// then .git at the project root takes priority over the nearer pyproject.toml
+	it("#given .git at project root and nested pyproject.toml in sub-package #when resolving #then prefers .git over nearer pyproject.toml", () => {
+		const root = tempRoot("lsp-ws-priority-git-");
+		mkdirSync(join(root, ".git"), { recursive: true });
+		mkdirSync(join(root, "packages", "sub-pkg"), { recursive: true });
+		writeFileSync(join(root, "packages", "sub-pkg", "pyproject.toml"), "[project]\nname = 'sub-pkg'\n");
+		writeFileSync(join(root, "packages", "sub-pkg", "main.py"), "x = 1\n");
+
+		const workspace = runWithRequestContext(createStandaloneMcpRequestContext({ cwd: root }), () =>
+			findWorkspaceRoot("packages/sub-pkg/main.py"),
+		);
+
+		expect(workspace).toBe(realpathSync(root));
+	});
+
+	// given a sub-package with pyproject.toml but NO .git anywhere up the tree
+	// when resolving workspace root for a file inside the sub-package
+	// then falls back to the first non-.git marker (pyproject.toml directory)
+	it("#given nested pyproject.toml and no .git anywhere #when resolving #then falls back to first non-git marker", () => {
+		const root = tempRoot("lsp-ws-priority-fallback-");
+		mkdirSync(join(root, "packages", "sub-pkg"), { recursive: true });
+		writeFileSync(join(root, "packages", "sub-pkg", "pyproject.toml"), "[project]\nname = 'sub-pkg'\n");
+		writeFileSync(join(root, "packages", "sub-pkg", "main.py"), "x = 1\n");
+
+		const workspace = runWithRequestContext(createStandaloneMcpRequestContext({ cwd: root }), () =>
+			findWorkspaceRoot("packages/sub-pkg/main.py"),
+		);
+
+		expect(workspace).toBe(realpathSync(join(root, "packages", "sub-pkg")));
+	});
+
+	// given a directory with both .git and pyproject.toml at the same level
+	// when resolving workspace root
+	// then returns that directory (both markers agree)
+	it("#given .git and pyproject.toml at same level #when resolving #then returns that directory", () => {
+		const root = tempRoot("lsp-ws-priority-same-level-");
+		mkdirSync(join(root, ".git"), { recursive: true });
+		writeFileSync(join(root, "pyproject.toml"), "[project]\nname = 'my-project'\n");
+		mkdirSync(join(root, "src"), { recursive: true });
+		writeFileSync(join(root, "src", "main.py"), "x = 1\n");
+
+		const workspace = runWithRequestContext(createStandaloneMcpRequestContext({ cwd: root }), () =>
+			findWorkspaceRoot("src/main.py"),
+		);
+
+		expect(workspace).toBe(realpathSync(root));
+	});
+
+	// given .git is a file (git worktree/submodule pointer) rather than a directory
+	// when resolving workspace root
+	// then existsSync returns true for the file and the directory is returned as root
+	it("#given .git is a file (git worktree pointer) rather than a directory #when resolving #then treats it as a valid workspace root", () => {
+		const root = tempRoot("lsp-ws-priority-worktree-");
+		// git worktree creates .git as a file: "gitdir: /path/to/main/.git/worktrees/name"
+		writeFileSync(join(root, ".git"), "gitdir: /some/path/.git/worktrees/feature\n");
+		mkdirSync(join(root, "src"), { recursive: true });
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot] The `.git`-first workspace root resolution is a sensible behavioral fix for monorepo LSP. The three test cases cover the key scenarios. One edge case to consider: what happens when `.git` is a file (git worktree/submodule pointer) rather than a directory? `existsSync` returns true for both, so this should be fine, but worth confirming. Needs a full CI run (test + typecheck) since this affects all LSP workspace resolution.

xxsddm:
[xxsddm-bot] Thanks for the review.

The `.git`-as-file edge case (git worktree/submodule pointer) is now covered by two dedicated test cases:
- `.git` as a file (worktree pointer) — confirms `existsSync` returns true and the directory is used as workspace root
- `.git` as a file in a submodule — confirms the walk stops at the submodule root, not the parent repo root

Additionally expanded coverage to other monorepo ecosystems where the same nesting pattern applies:
- JS/TS monorepo with nested `package.json`
- Rust workspace with nested `Cargo.toml`
- Go multi-module with nested `go.mod`
- Java Maven multi-module with nested `pom.xml`
- Java Gradle multi-module with nested `build.gradle`
- Multiple nested fallback markers with no `.git` (verifies nearest fallback wins)

All 13 tests pass (the 1 failure is a pre-existing Windows EPERM symlink permission issue unrelated to this change). Typecheck passes for `lsp-core`.

Regarding the CI run — the `ci.yml` workflow is in `action_required` state (GitHub requires maintainer approval for first-time fork contributors). I am unable to trigger it myself; it needs a maintainer to approve the workflow run on the PR page.

Issue #5864, opened 2026-07-04
[Bug]: Background task notifications are deferred indefinitely when parent session stays continuously active — no max-defer ceiling in pendingQueue


ton618to:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

When the parent (orchestrator) session stays continuously active (constant tool-call traffic) while a background task completes, the completion notification is queued into `pendingQueue` and then **deferred indefinitely** until the 30-minute inactivity watchdog kills the task. The orchestrator never receives the `<system-reminder>`, even though the child subagent has already finished its work and written its output to disk.

This is **different from #5172** (prompt-async-gate deadlock — that fires *after* the parent goes idle) and **different from #4689** (worker never launches). Here the worker launches, completes, and persists its result; only the notification is lost.

The core defect is in `manager.ts:2686` (`shouldDeferNotification = await isSessionActive(parentSessionId)`) — the defer branch queues the wake through `queuePendingParentWake` with a debounce of `PENDING_PARENT_WAKE_DEBOUNCE_MS = 100`, but the actual flush is driven exclusively by the `session.idle` event (`manager.ts:1804-1811`). If the parent orchestrator never emits `session.idle` because it keeps making tool calls, the pending wake sits in the queue forever. There is **no max-defer ceiling** and no fallback flush path.

An additional side-effect: when the same orchestrator fires multiple background tasks quickly (or the same logical task is re-dispatched because the previous notification appeared "lost"), later wakes on the same `parentSessionId` overwrite earlier pending wakes in `ParentWakePendingQueue.queueWake` — so once a wake is stuck, only the newest task's notification survives.

### Steps to Reproduce

1. Start a Sisyphus-style orchestrator session (e.g. novel dispatcher, ultrawork planner, any long-running orchestrator that emits continuous tool calls).
2. Dispatch a background task via `task(subagent_type=..., run_in_background=true, ...)`.
[...]

PR #6588, opened 2026-08-04
fix(omo-opencode): wake parent on monitor terminal output

1vivy:
## Summary

- Dispatch a monitor's **terminal** output batch as a reply-producing parent wake instead of an unconditional `noReply` observation, so a monitored command that exits — including one that fails — actually resumes the parent session.
- Always produce a terminal batch on process exit. `MonitorBatcher.flushNow()` returned early on an empty buffer, so a command whose output had already been flushed produced no terminal batch at all and the `Status: exited (code=N)` envelope was never built.
- Add a bounded active-session defer ceiling for terminal batches, so a finished command is not withheld indefinitely while the parent stays busy.
- Streaming batches stay no-reply and low-noise, and the ceiling is opt-in, so no existing caller changes behavior.

Refs #5864, #5790.

## Why

`monitor_start` is documented as delivering output automatically to the parent session, but nothing in `features/monitor` ever produced a reply-required wake. `dispatchBatch` always sent `noReply: true` wrapped in `withInternalNoReplyMarker(...)`, and the terminal callback threaded into `observeProcessExit` is `enforceTerminalRetention` — retention bookkeeping only:

```
$ rg "shouldReply|queuePendingParentWake|notifyParent" packages/omo-opencode/src/features/monitor/
(no matches)
```

So a monitored command can finish or fail and the agent never learns about it until something unrelated starts the next turn.

Separately, `flushMonitor` rescheduled unconditionally whenever the parent session was active, with no ceiling — the same unbounded-defer shape #5864 fixed for background-task parent wakes, on a path that fix does not cover.

Observed on 4.19.3 (opencode 1.18.7, headless `serve`) against a monitored build:

| Time (UTC) | Event |
|---|---|
| 16:16:52 | `[monitor] Deferred output injection skipped by promptAsync gate` — batch ready, parent busy |
| 17:02:06 | parent finally emits `session.idle` |
| 17:02:08 | `[monitor] Sent deferred monitor output` — 45 min after the output was ready |
| 17:02:08 → 18:35:33 | **93 min, no assistant turn** — the batch landed as `noReply`, so nothing woke |
| 18:35:36 | `[prompt-async-gate] expired reservation released` for the 17:02 dispatch (15 s `holdMs`) |

The operator had to interrupt manually to discover the build had finished.

## Changes

- **Reply-producing terminal dispatch** (`output-injector.ts`): a batch with `stillRunning === false` is the one carrying the `Status: exited (...)` envelope, so it dispatches with `noReply: false` and without the internal no-reply marker — the shape `parent-wake-prompt-dispatch.ts` already uses for reply-required wakes.
- **`flushNow({ allowEmpty: true })`** (`batcher.ts`, `monitor-state-factory.ts`): both process-exit paths now force a terminal batch even when the buffer is drained. Default `flushNow()` is unchanged, so the periodic and threshold flushes still ignore an empty buffer.
- **`queuedAtBySource`** (`output-injector.ts`): records when a batch source first entered the queue and preserves it across `requeueBatch`, so repeated deferral cannot reset the ceiling clock. Cleared on delivery.
- **`maxActiveDeferMs` ceiling** (`output-injector-types.ts`, `output-injector.ts`): optional dep. Once a *terminal* batch has waited that long, `flushMonitor` bypasses the session-active early return and `dispatchBatch` skips the prompt-gate status check. The assistant-turn and user-message safety guards still run. `undefined` → unchanged for every existing caller.
- **`MONITOR_OUTPUT_MAX_ACTIVE_DEFER_MS = 60_000`** (`manager-internals.ts`, `monitor-state-factory.ts`), matching `PENDING_PARENT_WAKE_MAX_ACTIVE_DEFER_MS`.

## QA

Recorded under `.omo/evidence/20260804-monitor-terminal-wake/`. Driven against a real `opencode serve` loading this build, with the npm plugin excluded — not a stub. `shouldReply` and `forceActiveDispatch` in the log lines are fields only this branch emits, which is how the running build was identified.

Monitored command `sh -c 'echo QA_BUILD_LINE; sleep 3; exit 7'`:

```
03:33:36  user       [OMO MONITOR OUTPUT] monitor_id: mon_f4718d66 batch: 1   <- streaming, noReply
03:33:38  user       [OMO MONITOR OUTPUT] monitor_id: mon_f4718d66 batch: 2   <- terminal batch
03:33:38 -> 03:33:47 assistant  "Monitor `qa-build` (mon_f4718d66) exited with code 7 ..."
```

The parent produced an assistant turn reporting the failure with zero user input. Before the change the same run ends at batch 1 with `shouldReply: false` and no assistant turn for the full 80 s observation window.

The empty-buffer defect was found *only* by running it — the first iteration of this branch passed its unit tests while still never waking on a real exit.

| Gate | Result |
|---|---|
| `monitor-terminal-wake.test.ts`, fix reverted | 4 fail / 4 pass |
| `monitor-terminal-wake.test.ts`, fix applied | 8 pass / 0 fail |
| `bun test packages/omo-opencode/src/features/monitor/` | 94 pass / 0 fail |
| `bun run typecheck` | exit 0 |
| `bun run build` | completes |
| root `bun test` | identical pre-existing failure set with and without this change (`comm` diff empty) |

No `package.json` version changes. Codex side untouched, so `test:codex` was not required.

---
## Summary by cubic
Make monitor terminal output wake the parent session and always emit a terminal batch on process exit. Adds a 60s active-defer ceiling to prevent indefinite deferral when the parent is busy. Addresses #5864 and #5790.

- **Bug Fixes**
  - Terminal batches dispatch with `noReply: false` (no internal no-reply marker) to wake the parent; streaming batches stay no-reply.
  - Only terminal batches past the defer ceiling are force-dispatched while the parent is busy and are prioritized ahead of streaming batches; streaming batches are never forced and forced terminal dispatch skips status polling.
  - On process exit, call `flushNow({ allowEmpty: true })` so the `Status: exited (...)` envelope is always sent, even if the buffer is empty.
  - Add optional `maxActiveDeferMs` with `queuedAt` tracking; past the ceiling, terminal batches bypass the session-active check and status polling. Default `MONITOR_OUTPUT_MAX_ACTIVE_DEFER_MS = 60_000`; unset preserves prior behavior.

<sup>Written for commit 27138a7ab95305c1ab8d6d4036b78bdac641fa52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


1vivy:

code-yeongyu:
recheck

code-yeongyu:
The workflow is approved now and CI actually ran. The `test (macos-latest)` / `test (windows-latest)` failures are **not an allowlist violation in this PR** — they are a crashed TypeScript parser process, so the audit could not inspect the sources at all.

Failure signature:

```
context canceled
...
function throwIfClosedOrDisposed() {
    if (isClosed()) {
        throw new ConnectionError(ConnectionErrors.Closed, 'Connection is closed.');
error: Connection is closed.
    at async parse (packages/omo-opencode/src/shared/typescript-native-source-parser.ts:12:38)
    at async prompt-async-route-audit.test.ts:463:44
```

Key detail: the first audit case (`only the shared gate may call raw OpenCode prompt APIs`) burned **5120ms** and then failed, while the two immediately preceding cases — `callers cannot disable the post-dispatch reservation hold` and `callers cannot bypass the central prompt queue` — **passed** at 411ms and 254ms against the same source tree. Then every subsequent case failed in 45-49ms with the identical `Connection is closed.`

That ordering is decisive. If this PR had introduced an unregistered raw prompt route, the audit would report the offending file path and the earlier cases would fail too. Instead the parser connection died mid-suite (`context canceled` right before), and every later `parse()` call short-circuits on a dead connection. The cascading `(fail) (unnamed)` is the afterAll teardown hitting the same closed connection.

Supporting evidence that the route itself is compliant: the diff dispatches through the gate rather than around it, and declares queue behavior explicitly.

```ts
const promptResult = await this.dispatchInternalPrompt({
  ...
  queueBehavior: "defer",
```

and the committed live-harness log shows the dispatch going through the shared gate:

```
[prompt-async-gate] promptAsync dispatching {"sessionID":"ses_...","source":"monitor-output:mon_f4718d66:batch-1"}
[prompt-async-gate] promptAsync dispatched  {"sessionID":"ses_...","source":"monitor-output:mon_f4718d66:batch-1"}
```

So `monitor-output:*` is a gate-routed source and would not need a `RAW_PROMPT_ALLOWLIST` entry.

Recommend re-running the job before treating this as a code problem. If it reproduces on rerun, the issue is `TypeScriptSourceParser` lifetime under CI resource pressure (the 5s stall on the first heavy case suggests the LSP-style connection is being torn down under load), which is a pre-existing infrastructure fragility rather than something this PR introduced — note `dev` currently has a separate flaky failure in `pollForCompletion` on macOS, so this runner is already showing timing sensitivity today.

ubuntu passed cleanly on the same commit, which is consistent with a resource/timing issue rather than a source-level audit violation.

PR #6589, opened 2026-08-04
fix(start-work): keep stale plans out of new sessions

### LINKED ISSUE #6584: [Feature]: Plans should be session-scoped so a new session cannot auto-resume an abandoned plan
### Prerequisites

### Problem Description

Plans written by Prometheus are stored in a flat shared folder (`.omo/plans/` / `.sisyphus/plans/`) and are visible to every future agent/session in the same project. When a brand-new session is started for an unrelated task, `start-work` can still pick up an *incomplete* plan left over from a previous, abandoned session and either silently auto-execute it or surface it as a `<system-reminder>` that nudges the new agent to resume it.

Concrete scenario:

1. I run `/start-work` on task A. Prometheus writes `add-sso.md` under `.omo/plans/`. I start executing it.
2. Mid-way, I decide task A is no longer worth doing. I do **not** mark the plan complete; I just move on.
3. The next day I open a brand-new session for an unrelated task B (say, "refactor the dashboard"). No plan for B exists yet.
4. `start-work` discovers the abandoned `add-sso.md`, sees it is incomplete, and either auto-selects it (because it's the only incomplete plan) or shoves it into a "Multiple Plans Found" reminder that the agent treats as an instruction to act on.

In step 4 the new agent reads the orphaned `add-sso.md` as if it were its own plan and starts implementing SSO in the dashboard refactor branch — even though I never asked for SSO. This is exactly the failure mode the "session-plan-affinity" hook in `packages/omo-opencode/src/hooks/start-work/session-plan-affinity.ts` was supposed to prevent, but it only kicks in when a plan path was *explicitly referenced* in this session. Plans left behind from previous sessions are not referenced, so the affinity check returns `null`, and the fallback in `plan-discovery-context.ts` (auto-pick if exactly one incomplete plan exists, or surface a system-reminder listing all of them) takes over.

Related open issues that touch the same area but don't fix the cross-session leak: #884, #1937, #2213, #1309.

### Proposed Solution

Make the plans directory session-aware so a plan can only be auto-resumed by the session that created it (or one explicitly pointed at it).

Concretely, pick some combination of:

1. **Session-scoped plan paths.** Store new plans under `.omo/plans/sessions/<session-id>/...md` (with a top-level `.omo/plans/index.json` mapping `plan-id → { sessionId, createdAt, status }`). Legacy flat plans keep working via a migration step.
2. **Plan metadata header.** When Prometheus writes a plan, prepend a front-matter block with `session-id`, `created-by`, `created-at`, `status`. `start-work` only auto-selects a plan whose `session-id` matches the current session, or whose `status` is `open`, or which the user explicitly named.
3. **Stricter fallback when affinity misses.** In `plan-discovery-context.ts`, when `preferredIncompletePlan` is `null` *and* `session-plan-affinity` returned `null` (i.e. none of the incomplete plans were ever referenced in this session), do **not** auto-pick and do **not** dump them in a "Multiple Plans Found" reminder. Instead, tell the agent "no plan for this session — invoke Prometheus to create one". If the user genuinely wants to resume a stale plan, they can pass `--plan <name>` or include the plan path in their prompt.
4. **First-class abandon/close.** Add a `/plan-abandon <name>` slash command and a corresponding `discard_plan` tool that sets `status: abandoned` (or moves the file under `.omo/plans/.archive/`). `start-work` then ignores abandoned plans entirely.

(1) + (3) is probably the minimum that fixes the bug. (4) is a quality-of-life improvement that reduces how often the bug fires in practice.

### Alternatives Considered

- **Manual cleanup.** I can `rm` old plan files myself, but that's brittle — I forget, or I delete a plan that another branch is still working on. The plugin should not depend on user hygiene to avoid executing the wrong task.
- **One plan per project.** Force only one in-progress plan at a time. This breaks legitimate parallel work (e.g. one plan per worktree in a multi-PR workflow) and doesn't address the root cause: plans from *finished* sessions leaking into *new* sessions.
- **Just disable auto-resume.** Always require `--plan`. This is too restrictive — the common case of "I just wrote a plan, now I want to execute it in this same session" should stay zero-config.
- **Per-worktree plans instead of per-session.** Tempting, but a single worktree can host multiple sessions over its lifetime (e.g. compact/resume), and the real failure is "session reads a plan it never created", not "two sessions share a worktree".

### Doctor Output (Optional)

```shell

```

### Additional Context

Relevant code paths:

- `packages/omo-opencode/src/hooks/start-work/plan-discovery-context.ts` — fallback to auto-pick a single incomplete plan, or surface a `<system-reminder>` listing all incomplete plans.
- `packages/omo-opencode/src/hooks/start-work/session-plan-affinity.ts` — already scans session messages for plan paths; only helps when a plan was *mentioned* in the current session.
- `packages/omo-opencode/src/hooks/start-work/plan-selection.ts` — `pickPreferredIncompletePlan` only matches by path equality; no notion of session ownership.
- `packages/boulder-state/src/plan-progress.ts` — plan progress tracking; would need a `status` field to support abandoned/closed.

Reproduction (no patch yet):

1. `bunx oh-my-openagent install`
2. Run `/start-work` on any task, accept Prometheus's plan, abort mid-execution (Ctrl-C / close the session) without finishing.
3. Open a new OpenCode session, run `/start-work` on an unrelated task.
4. Observe the old plan either being auto-selected or pushed via system-reminder.

### Feature Type

Configuration Option

### Contribution

- [ ] I'm willing to submit a PR for this feature
- [x] I can help with testing

jangByeongHui:
## Summary

Prevents `/start-work` from treating an incomplete plan left by another session as the current session's work. A session with no plan affinity now receives a prompt to create a new plan, while explicit plan selection and plans referenced by the same session continue to auto-select.

## Changes

- **Plan discovery:** distinguish a missing session-affinity result from an optional selection preference. The no-affinity path now emits `No Plan for This Session` rather than auto-selecting one stale plan or exposing several stale plans in an actionable system reminder.
- **Regression coverage:** update direct context-builder, start-work hook, native-command, and full chat-message integration tests. Existing auto-selection tests now supply a same-session plan reference, preserving the intended resume path.

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/plugin-interface.test.ts packages/omo-opencode/src/plugin/chat-message.test.ts packages/omo-opencode/src/hooks/start-work`
  **Observed result:** 134 passed, 0 failed.
  **Artifact:** `.omo/evidence/20260804-fix-6584-session-scoped-plans/qa-summary.md`
  **Why sufficient:** covers the direct fallback, session-affine auto-selection, native command flow, and full chat.message slash-command path.

- **What was tested:** local built plugin through isolated `opencode run --dir <fixture> --format json '/start-work'`, with only an incomplete `abandoned-plan.md` in the fixture.
  **Observed result:** OpenCode found no Boulder state and refused to execute the stale plan; the host session count remained 485 -> 485; run exited 0.
  **Artifact:** `.omo/evidence/20260804-fix-6584-session-scoped-plans/qa-summary.md`
  **Why sufficient:** proves the built plugin behaves safely through the real OpenCode surface and the run used an isolated XDG sandbox.

- **What was tested:** `bun run build` and `./node_modules/.bin/tsgo --noEmit -p packages/omo-opencode/tsconfig.json`
  **Observed result:** both passed.
  **Artifact:** `.omo/evidence/20260804-fix-6584-session-scoped-plans/qa-summary.md`
  **Why sufficient:** validates the edited OpenCode adapter builds and typechecks.

## Risks & Residuals

- **Mitigated:** explicit plan names and same-session plan references remain eligible for automatic selection, pinned by focused tests.
- **Accepted environment baseline:** full `bun test` still has 5 unrelated failures: missing pinned ast-grep runtime (1) and tmux placeholder expectation mismatches (4). The affected 134-test slice passes.
- **Accepted environment limit:** full workspace typecheck was killed with exit 137 during `packages/openclaw-core`; the affected `packages/omo-opencode` typecheck passes.
- No configuration, storage layout, migration, or prompt-async gate semantics changed.

## Automated Checks

​```bash
bun test packages/omo-opencode/src/plugin-interface.test.ts packages/omo-opencode/src/plugin/chat-message.test.ts packages/omo-opencode/src/hooks/start-work
bun run build
./node_modules/.bin/tsgo --noEmit -p packages/omo-opencode/tsconfig.json
​```

## Related Issues

Closes #6584

---
## Summary by cubic
Stop `/start-work` from auto-selecting or listing incomplete plans from other sessions. If a session has no valid plan affinity—or its affinity points to a completed plan—show “No Plan for This Session” and prompt to create a plan. Explicit same-session plan references still auto-select. Fixes #6584.

- Enforces session-affine plan discovery and rejects completed affinities; unrelated incomplete plans are ignored.
- Removes the “Multiple Plans Found” reminder; new sessions do not see other sessions’ plans.
- Preserves auto-select for explicit plan names and same-session plan paths; limits notepad scaffolding to the session-selected plan.
- Updates tests across `packages/omo-opencode` (`start-work`, `plugin-interface`, `chat-message`) and adjusts mocks to supply recent same-session plan paths.

<sup>Written for commit 673e77e89ad7e311acef00ee6a3abecd57cc4e50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jangByeongHui:
@code-yeongyu  Testing is complete on this PR — all QA evidence is documented above (134 targeted tests passing, isolated opencode run smoke test confirming no stale plan auto-selection, build + typecheck clean). Could you take a look and confirm when you have a chance?

code-yeongyu:
Thanks for picking this up — I triaged #6584 and the approach here matches what I recommended (option 3: stop treating "no affinity" as "pick something"). Placement is right: the guard sits after the `preferredIncompletePlan` branch, so same-session resume and explicit `--plan` still work, and the test updates correctly assert the new contract instead of deleting the old coverage.

One correctness gap worth fixing before merge.

The guard keys on `preferredPlanPath`, but the branch above it keys on `preferredIncompletePlan`:

```ts
const preferredIncompletePlan = pickPreferredIncompletePlan(incompletePlans, preferredPlanPath)
...
if (preferredIncompletePlan) { /* auto-select, correct */ }

if (!preferredPlanPath) { /* new guard */ }

if (incompletePlans.length === 1) { /* stale auto-pick still reachable */ }
```

`pickPreferredIncompletePlan` returns non-null only when `preferredPlanPath` is present **and** matches an entry in `incompletePlans`. So there is a state where `preferredPlanPath` is truthy but `preferredIncompletePlan` is null: the session referenced a plan that is now complete, or referenced a path no longer in the incomplete set.

In that case the new guard is skipped (because `preferredPlanPath` is truthy), and control falls through to `incompletePlans.length === 1`, which auto-selects an unrelated stale plan — exactly the bug #6584 reports.

Concrete repro: session A creates and finishes `plan-a.md`, and references it in-session. A stale `plan-b.md` from an older session is still incomplete. Affinity resolves `plan-a.md` → `preferredPlanPath` is set, but `pickPreferredIncompletePlan` returns null because `plan-a.md` is complete. Guard is bypassed and `plan-b.md` is auto-selected.

Fix is to key the guard on the same value the auto-select branch uses:

```ts
if (!preferredIncompletePlan) {
  return contextInfo + `

## No Plan for This Session

No incomplete plan was referenced in this session. Invoke the Prometheus agent to create a new work plan.`
}
```

That makes the two branches exhaustive over the same condition, and everything below becomes unreachable — which is the actual intent, since `incompletePlans.length === 1` and the `Multiple Plans Found` reminder are both the stale-plan paths being removed. Worth adding a test for the "session referenced a now-complete plan while an unrelated incomplete plan exists" case, since that is the one the current guard misses.

Also flagging for the maintainer: with this change the `formatIncompletePlanList` import and the `Multiple Plans Found` block become dead code in this function. Leaving them is fine for review clarity, but they should be removed in a follow-up (or now) so the file does not retain an unreachable branch.

Minor: the PR body notes 5 unrelated `bun test` failures and a typecheck OOM. Those match known environment issues on `dev` — one is the ast-grep runtime and there is a separate flaky `pollForCompletion` test currently red on `dev` — so I do not read them as caused by this change.

jangByeongHui:
@code-yeongyu  Fixed in `a09f6ebd2` — the guard now keys on `preferredIncompletePlan` instead of `preferredPlanPath`, so a completed/missing affinity plan always returns "No Plan for This Session" instead of falling through to stale auto-select. Removed the now-unreachable single-plan fallback and "Multiple Plans Found" reminder, plus the unused `formatIncompletePlanList` import. Added the regression test you flagged (completed affinity plan + unrelated incomplete plan).

Verified: `bun test packages/omo-opencode/src/hooks/start-work` (95 passed), `bun run build` passed, isolated `opencode run '/start-work'` smoke did not auto-execute the stale plan, host DB session count unchanged (489 -> 489).

PTAL when you get a chance.

code-yeongyu:
[sisyphus-bot] Correct fix for #6584 with maintainer-confirmed approach. The removal of both auto-select paths and the replacement with "No Plan for This Session" is clean and well-tested. The updated test suite properly asserts the new session-scoped behavior. Recommend owner review before merge to confirm the UX change is acceptable.

jangByeongHui:
@code-yeongyu it's been about a week since the last update and the fix + regression test from my previous comment are still in place, CI is clean/mergeable. Happy to make any further changes if needed, just let me know. Thanks!

jangByeongHui:
Merge conflict resolved
Rebased/merged dev into fix/6584-session-scoped-plans (673e77e89) and resolved the two conflicting hunks.
Conflicting files:
- packages/omo-opencode/src/hooks/start-work/context-info-builder.test.ts
- packages/omo-opencode/src/hooks/start-work/index.test.ts
Why they conflicted: this branch split a single existing test into two new tests (renaming + rewriting assertions to check for "No Plan for This Session" instead of the old auto-select behavior). Since dev never touched this code path, its old assertions ("cold-start-plan" auto-select / <system-reminder> multi-plan prompt) were left dangling at the tail of the rewritten hunks by the 3-way merge. Resolved by keeping this branch's (HEAD) content and dropping the stale dev tail in both files — no behavioral changes beyond what this PR already intended.
Verification after resolving:
bun test packages/omo-opencode/src/plugin-interface.test.ts packages/omo-opencode/src/plugin/chat-message.test.ts packages/omo-opencode/src/hooks/start-work
# 136 pass, 0 fail

./node_modules/.bin/tsgo --noEmit -p packages/omo-opencode/tsconfig.json
# clean, no errors
mergeable_state is no longer dirty. Ready for another look — @code-yeongyu  PTAL 🙏

MoerAI:
Thanks for implementing the owner-requested no-affinity behavior and its completed-affinity edge case.

Closing without merge because owner-merged #7280 renamed `start-work` to `ulw-execute`, removing every production surface changed here:
https://github.com/code-yeongyu/oh-my-openagent/pull/7280

Keep #6584 open. A current successor should carry the approved contract into `hooks/ulw-execute`: with no referenced incomplete plan, the session must neither auto-select a stale plan nor expose multiple stale plans as actionable work:
https://github.com/code-yeongyu/oh-my-openagent/issues/6584#issuecomment-5173239775

The successor must also define and test whether a no-affinity session may resume a single active Boulder owned by another session, then run current CI and isolated real OpenCode QA with committed evidence. Happy to review a fresh PR on the current lifecycle.

PR #6590, opened 2026-08-04
fix(senpi-task): recover dual team_create inputs

code-yeongyu:
## Summary

- make `inline_spec` authoritative when `team_create` receives both supported input forms
- preserve named-only, inline-only, missing-input, and malformed-inline behavior
- add failing-first regression coverage from the recovered Kimi payload
- exercise the exact dual-field call through the real isolated Senpi CLI

## Why

Two independent `kimi-k3-ultrafast-unlocked` sessions repeatedly emitted both
`team_name` and `inline_spec`. The model correctly reasoned that it should
remove `team_name`, but six consecutive calls retained both fields and hit the
runtime XOR rejection.

A schema-only union is not reliable for this provider path because Senpi's
Moonshot compatibility normalizer flattens root object unions. The complete
inline payload is therefore accepted as authoritative at the execution
boundary instead of depending on another model retry.

## Observed behavior

Before:

- focused test expected `created` and received `invalid_arguments`
- recovered session entered repeated `team_create` retries

After:

- focused test: 1 pass, 0 fail
- real Senpi trace: exactly one dual-field `team_create` call
- real Senpi result: `details.kind=created`
- invalid-argument results: 0
- created team: 2 running members on local mock models

## Verification

- `bun run --cwd packages/senpi-task test`
- `bun test ./packages/omo-senpi/src/components/task/team-service.test.ts`
- `node packages/omo-senpi/scripts/qa/team-e2e.mjs --self-test`
- isolated live `team-e2e.mjs`: PASS, no failed checks, no leaked PIDs
- changed-file LSP daemon diagnostics: 0
- `bun run test:senpi`
- `bun run build`
- `git diff --check origin/dev...HEAD`

## Evidence

`.omo/evidence/20260804-team-create-recovery/`

Key files:

- `diagnosis.md`
- `red-focused-test.txt`
- `green-focused-test.txt`
- `live-senpi-qa.md`
- `live-team-e2e-isolated-home/verdict.json`
- `verification-gates.txt`
- `self-review.md`

## Residual risk

Other tools with runtime-only mutually-exclusive fields are unchanged. This PR
keeps the policy local to the demonstrated `team_create` failure.

---
## Summary by cubic
Prefer `inline_spec` when `team_create` receives both `team_name` and `inline_spec`, fixing dual-field calls that previously failed with `invalid_arguments`. This makes the inline payload authoritative and lets teams be created in one call.

- **Bug Fixes**
  - `senpi-task`: `runTeamCreate` now forwards only `{ inlineSpec }` when both inputs are present; schema descriptions updated; named-only, inline-only, neither-input, and malformed-inline behaviors stay the same.
  - Tests: added `lifecycle-precedence.test.ts`; removed the test that expected a rejection for dual-field input.
  - `omo-senpi`: refreshed extension bundle and updated the QA script to emit a dual-field `team_create` via the real `senpi` CLI.

<sup>Written for commit 22b597badba5034ff0e3fd8b9996116fbd61701a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6590?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Diagnosis here is strong — tracing the loop to `tool-schema-compat.ts` flattening root object unions on the Moonshot path explains why a schema-level `oneOf` would not have helped, and the recovered transcripts (8 calls in one session, 4 in an independent one) establish this is a real recurring failure rather than a one-off.

The precedence direction is right. `inline_spec` carries a complete spec while `team_name` is only a reference, so preferring the richer payload is the choice that can always execute. Agreed that returning a harder error would not have helped: the transcripts show the model already restated the rule correctly and still emitted both fields, so this was never an error-clarity problem.

One thing worth deciding explicitly before merge: this silently discards a field the caller supplied.

When both arrive with *different* identities — `team_name: "alpha"` and `inline_spec.name: "beta"` — the team is created as `beta` and `alpha` is dropped with no signal in the result. In the recovered Kimi payloads the two presumably agreed, so the loop is fixed either way. But a caller that genuinely believed it was creating `alpha` now gets `beta` and no indication anything was ignored.

Since `TeamCreateDetails` already carries a `kind`, surfacing this cheaply would help:

- include the ignored `team_name` in the success details (e.g. `ignored_team_name`), or
- log it at the execution boundary

That keeps the first-call-succeeds property that motivates the fix while leaving a trace when the two inputs disagree. Purely additive to the current change.

Second, smaller: the removed test in `lifecycle.test.ts` asserted the old XOR rejection, and its replacement `lifecycle-precedence.test.ts` covers the both-fields case. But the PR body lists "named-only, inline-only, missing-input, and malformed-inline behavior" as preserved, and the new file is a single test. Worth confirming the both-fields-with-*conflicting*-names case is covered somewhere, since that is the only path where the new precedence rule is observable to a caller — the agreeing case looks identical either way.

Nothing blocking. The `!hasName && !hasInline` guard correctly preserves the missing-input rejection, and the schema/description updates match the new contract so the model is no longer told a rule the runtime does not enforce.

PR #6594, opened 2026-08-04
fix(omo-senpi): arm ultrawork directive once per session with compaction re-arm

code-yeongyu:
## What changed

`packages/omo-senpi/src/components/ultrawork/index.ts` (+ regenerated `plugin/extensions/omo.js` bundle): the ultrawork input hook now arms the directive **once per session** instead of on every trigger.

- **Component-scoped arming ledger**: a `Set<string>` of armed session ids, a current-session tracker fed by `session_start` / `session_before_switch`, and an anonymous fallback slot for hosts that expose no session id anywhere.
- **Session id resolution order**: the input event's own `ExtensionContext` (`sessionManager.getSessionId()`, verified present on every handler ctx in senpi's `extensions/types.ts`) → the lifecycle tracker → the anonymous slot.
- **2nd+ trigger in an armed session** injects `ULTRAWORK_REMINDER` (172 chars, no `<ultrawork-mode>` open tag) instead of the ~17KB directive — on **both** paths: idle (`pi.sendMessage`, same `omo-ultrawork:directive` customType, `display: false`) and queued (transform-append, still atomic with the queued message, `/skill:` prefix preserved).
- **Re-arm only on accepted `session_compact`** (payload `accepted: false` is skipped — a rejected compaction leaves the transcript, and the directive, intact). `session_before_switch` deliberately never re-arms: switching back into a session whose history still holds the directive must not re-inject it (that would recreate the accumulation bug).
- **Pasted directive transcripts** (tag-pair guard) and **`/skill:ultrawork` expansions** now mark the session armed with no other behavior change, so the next plain trigger is the reminder rather than a second full copy.
- First trigger in any session is byte-identical to today (pinned by characterization tests).

## Why

Plan `.omo/plans/claude-oauth-token-burn.md` (senpi repo), todo 2: every repeat ulw trigger in a session re-injected the full ~17KB directive, re-paying those tokens per trigger and stacking duplicate rule blocks into the transcript (compaction summaries then carried the pile forward).

## RED / GREEN evidence

**RED** (commit `641306355`, tests only): `bun test packages/omo-senpi/src/components/ultrawork/ultrawork.test.ts` → **7 fail / 26 pass**. Failing exactly the new behavior tests: same-session 2nd trigger (idle + queued), rejected-compaction reminder, pasted-block arming, `/skill:ultrawork` arming, session switch-back, input-ctx session id. The two characterization tests (accepted-compact re-arm, first-trigger byte-identical) already passed.

**GREEN** (this commit):
- `bun test packages/omo-senpi/src/components/ultrawork/ultrawork.test.ts` → **33 pass / 0 fail**
- `bun test packages/omo-senpi/src` → **486 pass / 0 fail**
- `bun run test:senpi` (package gate: daemon build + artifact staging + typecheck + full package suite) → **538 pass / 0 fail**
- `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` → **exit 0, zero output**
- `node packages/omo-senpi/plugin/scripts/build-extension.mjs --check` → bundle current

## Environment note

The fresh worktree initially symlinked the shared worktree's `node_modules`, which resolved workspace deps cross-tree (2 test files failed on `@oh-my-opencode/omo-opencode/config-migration`, plus dual-senpi tsgo noise). A real `bun install` in the task worktree fixed all of it; unrelated regenerated `packages/omo-codex` artifacts from the install postinstall were reverted. No shared-worktree files were touched.

## Residual risk

- Live senpi QA drivers (`scripts/qa/*.mjs`) were not run; behavior is verified at the unit + package-gate level only. The change is confined to one component's input/session handlers.
- Hosts exposing no session id on any event fall back to a single anonymous armed slot (reset by `session_compact`), documented in the ledger comment.

---
## Summary by cubic
Arms the ultrawork directive once per session and sends a short reminder on later triggers. Re-arms only after an accepted compaction or when the embedded directive text changes; state survives component re-registration and packaged `omo.js` reloads.

- **Bug Fixes**
  - Process-wide arming ledger on `globalThis` via `Symbol.for("omo.ultrawork.arming")`; the shared slot stores the full directive text and reuses the ledger only on exact text match. Mismatched or legacy bare ledgers are discarded; tests can inject isolated ledgers.
  - Session ID: input event context → lifecycle tracker (`session_start`/`session_before_switch`) → anonymous fallback.
  - Second+ triggers send a <400-char hidden reminder (no `<ultrawork-mode>`) on idle and append on queued; `/skill:` prefix preserved.
  - Re-arm only on accepted `session_compact` or directive text change; `session_before_switch` never re-arms.
  - Pasted directive blocks and `/skill:ultrawork` expansions mark the session as armed; first trigger remains byte-identical.

<sup>Written for commit 6cc01322558c002b651d5b44756db6d34fd95b86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
The core reasoning here is sound, and two decisions in particular are worth calling out as correct because they are easy to get backwards:

- **`session_before_switch` deliberately not re-arming.** Switching back into a session whose transcript still holds the directive must not re-inject it. Treating switch as "new session" would recreate exactly the accumulation this fixes.
- **Rejected compaction skipped.** A rejected compact leaves the transcript intact, so re-arming there would re-pay the ~17KB for rules still visible. The `accepted === false` guard is the right discriminator.

Marking pasted blocks and `/skill:ultrawork` expansions as armed is also right — both paths put the full directive into context, so the next plain trigger genuinely only needs the pointer.

Two things worth deciding before merge.

**1. The anonymous fallback slot is process-wide, not session-wide.**

`createSessionArming` is per-component-registration, and `anonymousArmed` is a single boolean shared by every session that resolves to `sessionId === undefined`:

```ts
isArmed(sessionId) {
  if (sessionId === undefined) return anonymousArmed
  return armedSessionIds.has(sessionId)
}
```

On a host that never exposes a session id, session A arms the slot and session B's *first* trigger then gets the reminder instead of the directive — pointing at a directive block that is not in B's transcript at all. The model is told "the ultrawork directive above remains binding" when there is nothing above.

That is a worse failure than the bug being fixed: over-injection wastes tokens, but under-injection silently drops the rules. The PR body documents the fallback as a known limitation, but the failure mode is asymmetric enough that it is worth stating explicitly whether any supported host actually hits this path. If `sessionManager.getSessionId()` is genuinely present on every handler ctx (as the comment asserts from `extensions/types.ts`), consider dropping the anonymous slot entirely and simply not arming when the id is unresolvable — always injecting the full directive is the safe degradation.

**2. `armedSessionIds` grows without bound.**

Entries are only ever removed by an accepted `session_compact` for that specific id. Sessions that end normally — no compaction — leave their id in the `Set` forever for the life of the host process. Each entry is small, so this is a slow leak rather than an urgent one, but a long-lived host that opens many sessions accumulates them indefinitely. `session_start` already fires per session; if senpi exposes a session-end/dispose event, deleting there would bound it cleanly.

Minor observations, non-blocking:

- `markArmed` is called unconditionally in `armUltrawork` even when the session was already armed. Harmless (`Set.add` is idempotent) and it keeps the anonymous path correct, so no change needed — just noting it is intentional-looking.
- The reminder is 172 chars vs ~17KB, so the token win is real. Good that it deliberately omits the `<ultrawork-mode>` open tag — otherwise the tag-pair guard would treat the reminder itself as a pasted directive on the following turn, which would be a subtle self-inflicted disarm.
- RED evidence (7 fail / 26 pass, with the two characterization tests already passing) is exactly the right shape for proving the new behavior is what changed and the first-trigger path is untouched.

Neither point blocks the token-burn fix, which is the actual goal. Both are about the degenerate-host path rather than the primary flow.

code-yeongyu:
`test (macos-latest)` failure on this run is **not caused by this PR**. It is the same `TypeScriptSourceParser` crash I diagnosed on #6588 earlier today, now reproducing on an unrelated branch.

Identical signature:

```
(fail) production prompt injection routes > ... only the shared gate may call raw OpenCode prompt APIs [5002.39ms]
error: Connection is closed.
(fail) ... every route declares queue behavior explicitly [58.69ms]
error: Connection is closed.
(fail) ... every retry caller declares queue behavior explicitly [41.38ms]
```

Same shape as #6588: the first audit case stalls ~5s and dies, then every subsequent `parse()` call fails in 40-60ms on the already-closed connection. The 5000ms figure is suspiciously exact — that reads like a fixed timeout being hit, not variable load.

Why it is definitely not this PR: `packages/omo-senpi/src/components/ultrawork/index.ts` does not call any OpenCode prompt API. It uses `pi.sendMessage` (the senpi extension API) and `{ action: "transform" }` on the input event. The audit walks `packages/omo-opencode/src` plus sibling package sources for raw `session.prompt` / `session.promptAsync` access — a senpi component changing which *string* it passes to `sendMessage` cannot affect that allowlist.

Cross-PR evidence that this is environmental:
- **#6588** (monitor terminal wake, `omo-opencode`): same three cases, same `Connection is closed.`
- **#6594** (this PR, `omo-senpi` ultrawork): same three cases, same error
- Both: ubuntu and windows pass the same commit; only macOS fails
- On this run `test (windows-latest)` passed at 11m34s while macOS died at 6m10s

Two different packages, two different change shapes, one shared failure — that is the parser process lifetime, not the diffs.

Recommend a rerun to confirm, and separately worth filing this as its own issue: `TypeScriptSourceParser` tearing down mid-suite makes any PR touching anything look broken on macOS, which is actively harmful to review signal. Right now a real allowlist violation and a dead parser produce different messages but the same red X, and the audit is the only thing guarding that invariant.

Note `dev` also has an unrelated flaky `pollForCompletion` failure on macOS today (commit 8f8b0de), so the macOS runner has now shown two independent timing sensitivities in one day.

My earlier review comments on the actual change (anonymous arming slot shared process-wide, `armedSessionIds` growing unbounded) still stand and are independent of this.

code-yeongyu:
Rerun result confirms the macOS parser failure was flaky, and surfaces a separate one.

**`test (macos-latest)` now passes** (5m46s). Same commit, same suite, previously dead at 6m10s with `Connection is closed.` on three audit cases. That is a clean reproduction of the intermittency — the `TypeScriptSourceParser` teardown is nondeterministic, not caused by any diff. Combined with the identical failure on #6588 earlier, I think this is worth its own issue rather than being re-diagnosed on each PR it lands on.

**New failure: `codex-compatibility (windows-latest)`**, and it is unrelated to both the parser and this PR:

```
(fail) install-codex > #given autonomous permissions requested #when installing omo #then writes Codex autonomy settings [60016.00ms]
  ^ this test timed out after 60000ms.
error: script "test:codex" exited with code 1
```

One test, hard 60s timeout, Windows only. Everything else on this run is green including all three `test` targets and all three `typecheck` targets. `codex-compatibility` passed on ubuntu (2m25s) and macOS (3m41s) on this same commit.

This PR touches `packages/omo-senpi/src/components/ultrawork/index.ts` and its test, plus the regenerated bundle. `install-codex` is in the Codex install path and has no dependency on the senpi ultrawork component, so a timeout there is not reachable from this diff.

Tally across today's runs on this branch and #6588:
- macOS: parser connection death (intermittent — passed on rerun)
- Windows: `install-codex` 60s timeout (new, this run)
- `dev`: flaky `pollForCompletion` on macOS (commit 8f8b0de)
- #6597: Windows path-escaping assertion
- #6575: Windows mode bits, then Windows `PATHEXT` fixtures

That is five distinct CI fragilities in one day, none of them caused by the PR they surfaced on. The signal-to-noise on CI right now makes it hard to tell a real regression from environment, which is the actual cost.

Suggest rerunning `codex-compatibility (windows-latest)` to establish whether the 60s timeout is also intermittent or reproducible. If reproducible it deserves its own investigation, since a hard timeout usually means a hang rather than slowness.

Review comments on the change itself (process-wide anonymous arming slot, unbounded `armedSessionIds`) are unaffected and still stand.

PR #6595, opened 2026-08-04
fix(omo-opencode): handle EXDEV config migration backups

### LINKED ISSUE #6574: Config migration silently fails to re-sync when ~/.config/opencode is a Docker bind mount (EXDEV on rename)
**What happened**

Running `oh-my-openagent` inside a Docker container where `~/.config/opencode` is a host bind mount, but the rest of the home directory (including `~/.omo`) is not. On every startup, the config-migration step logs:

```
[config-migration] startup completed {"error":"EXDEV: cross-device link not permitted, rename '/home/openchamber/.config/opencode/oh-my-openagent.json' -> '/home/openchamber/.omo/migration-backup-.../.config/opencode/oh-my-openagent.json'", "journalResumed":false, "migratedFrom":[], "skippedConflictCount":0}
```

The rename fails because the source and destination are on different filesystems (one is a bind mount, one is the container's own writable layer) — `fs.rename()` cannot cross that boundary, per POSIX.

**Impact**

After the one-time migration to `.omo/omo.jsonc` has already completed once (`.migration-journal.json` shows `"targetWritten": true`), every subsequent edit to the legacy `oh-my-openagent.json` is silently ignored — the plugin never re-reads it, and the EXDEV error gives no indication that any config was skipped. This can look exactly like arbitrary, unrelated agent-model overrides being ignored, since there's no error path surfaced to the user beyond a debug-level log line in `/tmp/oh-my-opencode.log`.

**Suggested fix**

Fall back to a copy+unlink (or `fs.cp` + `fs.rm`) when `rename()` throws `EXDEV`, same as most Node tools handle cross-device moves. Alternatively, surface a visible warning/toast when the migration backup step fails, since users have no way to know their `oh-my-openagent.json` edits aren't taking effect otherwise.

**Workaround**

Bind-mount `~/.omo` to the same host filesystem as `~/.config/opencode` (avoids the cross-device boundary entirely), or edit `~/.omo/omo.jsonc` directly instead of the legacy file.

code-yeongyu:
## Summary

Fix config migration when a legacy OpenCode config is on a different filesystem from `~/.omo`, such as a Docker bind mount. Backup moves now recover from `EXDEV` with copy plus unlink, and startup emits a non-debug warning whenever migration still fails so users know legacy edits were not applied.

Fixes #6574

## Root cause

The migration transaction used `renameSync` for backup moves in both normal execution and pending-journal recovery. POSIX rename cannot cross filesystem boundaries, so a bind-mounted `~/.config/opencode` and container-local `~/.omo` left the migration journal pending after the unified target was written and retried the same failing rename on every startup. The error otherwise only reached the debug log and an optional TUI toast.

## Changes

- Add one shared migration backup move that falls back to `copyFileSync` plus `unlinkSync` only when rename reports `EXDEV`.
- Use the fallback in both fresh migration execution and journal recovery, including journals with `targetWritten: true`.
- Preserve failures for non-`EXDEV` rename errors.
- Emit `console.warn` on startup migration failure while retaining the existing error toast.
- Add failing-first regression coverage for fresh EXDEV, target-recorded journal recovery, non-EXDEV behavior, and the warning.

## QA and evidence

Evidence root: [`.omo/evidence/20260804-exdev-migration/`](.omo/evidence/20260804-exdev-migration/)

- **RED / negative control:** [`negative-control-red.txt`](.omo/evidence/20260804-exdev-migration/negative-control-red.txt) shows the finalized tests fail without the fallback and warning: fresh EXDEV, `targetWritten: true` recovery EXDEV, and missing warning.
- **Touched-scope tests:** [`touched-scope-tests.txt`](.omo/evidence/20260804-exdev-migration/touched-scope-tests.txt) records 46 passing tests, 0 failures across the shared migration engine and OpenCode startup suites.
- **Typecheck:** [`typecheck.txt`](.omo/evidence/20260804-exdev-migration/typecheck.txt) records `bun run typecheck`, exit 0.
- **Build:** [`build.txt`](.omo/evidence/20260804-exdev-migration/build.txt) records `bun run build`, exit 0.
- **Real OpenCode bind-mount QA:** [`real-opencode-cross-device-qa.sh`](.omo/evidence/20260804-exdev-migration/real-opencode-cross-device-qa.sh) drove OpenCode 1.17.7 in disposable Docker containers with the local plugin.
  - [`success-devices.txt`](.omo/evidence/20260804-exdev-migration/success-devices.txt) proves different source and destination device ids.
  - [`success-observed.txt`](.omo/evidence/20260804-exdev-migration/success-observed.txt), [`success-omo.jsonc`](.omo/evidence/20260804-exdev-migration/success-omo.jsonc), and [`success-backup.json`](.omo/evidence/20260804-exdev-migration/success-backup.json) prove the source was removed, backup created, target migrated, and journal cleared.
  - [`warning-warning.txt`](.omo/evidence/20260804-exdev-migration/warning-warning.txt) proves a real backup cleanup failure is visible outside the debug logger.
  - [`isolation.txt`](.omo/evidence/20260804-exdev-migration/isolation.txt) proves the host OpenCode session count remained unchanged.
- **Reviewer summary:** [`qa-summary.md`](.omo/evidence/20260804-exdev-migration/qa-summary.md) maps every artifact to the behavior and explains omissions.

## Residual risk

The fallback intentionally changes only `EXDEV`; all other rename failures retain the existing transaction failure behavior. If the fallback copy succeeds but source unlink still fails, the migration remains pending, but startup now emits the visible warning requested by the issue. No schema, migration id, package version, or merge semantics changed.

---
## Summary by cubic
Fixes EXDEV failures during OpenCode config migration by handling cross-filesystem backup moves. Adds a copy+unlink fallback and a visible warning when migration can’t finish so users know legacy changes weren’t applied (addresses #6574).

- **Bug Fixes**
  - `omo-config-core`: Added `moveMigrationBackup()` to fall back to `copyFileSync` + `unlinkSync` when `renameSync` returns `EXDEV`; used in normal execution and journal recovery.
  - `omo-opencode`: Emits `console.warn` on startup when migration fails, keeping the existing error toast.
  - Non-`EXDEV` rename errors remain failures and preserve the source; tests cover EXDEV fresh runs, `targetWritten: true` recovery, and the new warning.
  - `omo-senpi`: Regenerated plugin bundle to include the config-core change; no functional behavior changes.

<sup>Written for commit f1fa98d904aefff2f9ae891c92449001ffed23c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Good root cause. `renameSync` failing with `EXDEV` across a bind mount explains both symptoms in #6574 — the journal staying pending *after* the target was already written, and the identical retry burning on every subsequent startup. Scoping the fallback to `EXDEV` only, so `EACCES` and friends still fail the transaction, is the right call: silently copy-and-unlinking on a permissions error would mask a real problem.

Applying it in `recovery.ts` as well as `batch.ts` matters more than it looks. Users already stuck in the pending-journal loop have `targetWritten: true` recorded, so a fix that only touched fresh execution would never reach them. Covering that path is what actually unbreaks existing installs.

The Docker QA is the convincing part — `success-devices.txt` proving genuinely distinct device ids means this was verified against real cross-filesystem semantics rather than a mocked error code.

One durability gap worth deciding on.

`moveMigrationBackup` is not atomic in the fallback branch, and the two-step is exposed:

```ts
fileSystem.copyFileSync(sourcePath, backupPath)
fileSystem.unlinkSync(sourcePath)
```

If the process dies between those calls, both source and backup exist on disk. On the next startup, `executePlan` reaches:

```ts
if (fileSystem.existsSync(move.to)) throw new MigrationTransactionError(`Migration backup path already exists: ${move.to}`)
```

and `resumeMigrationJournal` has the same `existsSync(move.to)` throw on the source-still-present branch. So a crash inside the fallback window converts a recoverable pending journal into a hard failure that repeats every startup — the same shape as the bug being fixed, just with a different message. The user then has to manually delete the backup directory to get unstuck.

`renameSync` never had this window, which is why the guard could safely treat "backup exists" as unambiguously bad.

Two options depending on how much you want to spend here:

- **Cheap:** in the fallback, if `copyFileSync` succeeded but `unlinkSync` throws, leave the journal pending (current behavior) but make the next-startup guard tolerate `to` existing *when the journal records this move as in-flight* — i.e. treat it as resumable rather than fatal.
- **Thorough:** copy to a temp path adjacent to `backupPath`, then `renameSync` temp → backup (same filesystem, so atomic), then unlink source. That collapses the crash window to the same-device rename, restoring the original invariant.

The PR body acknowledges "if the fallback copy succeeds but source unlink still fails, the migration remains pending" — I would just note the consequence is stronger than pending: the *next* run hard-fails on the existence guard rather than retrying.

Separately, the `console.warn` on startup failure is a real improvement. Burying this in debug-only logging is why #6574 presented as silent — users saw legacy edits quietly not applying with no signal. Surfacing it outside the debug logger, while keeping the toast, is the right level.

Non-blocking: `isCrossDeviceError` uses `Reflect.get(error, "code")` rather than a cast, which is the correct defensive read for an untyped `code` property.

PR #6597, opened 2026-08-04
feat(senpi-task): subagent/team revival on session resume (suspend-not-dispose + scoped revival)

code-yeongyu:
## Summary

Subagents (including in-process "memory mode" children) and team members now survive session shutdown as **suspended** records and **revive** when their parent session is resumed — unless deliberately stopped (`task_cancel`, `/task-kill`, kill, evict, TTL, team shutdown).

## What changed

- **In-process child conversations persist continuously** (`SessionManager.create` under `stateDir/children/<taskId>/sessions/`, shared layout with RPC).
- **Session shutdown suspends residents** (`persist_only`/`detach_rpc`) instead of disposing them; `resume_children === false` restores the exact pre-feature dispose path.
- **Scoped session-resume revival**: resuming a session revives only ITS children (reclamation vs admission two-population model, ownership CAS, fenced admission lease, deferred outcomes). Other sessions' children stay suspended.
- **Only mid-turn-interrupted children resume execution** (hardened `sessionTailNeedsContinuation`: `stopReason:"aborted"` counts as interrupted, reads the true last JSONL entry); finished children restore idle/steerable, never re-run.
- **Deliberately stopped children NEVER revive** — incl. a genuine bug found by live QA: `shutdown_approved` team members were not filtered from revival (`ee792ec39`).
- **Durable completion + steering**: unnotified terminal completions reconcile on session start (conditional epoch mutate, no ownership clobber); prelaunch steering persists across restarts.
- **Race-safe TTL**: two-phase tombstone expunge (atomic conditional tombstone, artifact deletion outside the lock, idempotent crash recovery).
- **Config**: `task.resume_children` (default `true`); `reattach_on_reconcile` semantics untouched.

## Verification

- `bun test packages/senpi-task` — 1313/0 (incl. chaos bench, 200-iteration interleavings, 8 mutation proofs).
- `bun test packages/omo-senpi` — 562/0; `tsgo --noEmit` clean both packages.
- **Live QA (real senpi binary)**: task driver 27/27 PASS, team driver 22/22 PASS — quit suspends, resume revives resident, mid-turn continuation, finished child no-nudge + steerable via `task_send` (asserted PONG), unrelated session no-revive, and every deliberate-stop path (cancel/kill/LRU/TTL/shutdown-approved) proven NOT revived. Isolation clean, zero leaked pids.
- **Interactive TUI smoke**: spawn → `/exit` → suspended (`persisted_only`, host_pid cleared) → `--continue` → revived resident under new pid → `/tasks` lists it.
- Concurrency: real two-OS-process race tests (ownership CAS, capacity admission) via `Bun.spawn`.

Plan: `.omo/plans/subagent-session-resume-revival.md`
Evidence: `.omo/evidence/` (per-todo) + `.omo/evidence/omo-senpi-adapter/subagent-session-resume-revival/` (live QA verdicts).

---
## Summary by cubic
Subagents and team members now suspend on session shutdown and revive when the parent session resumes. Revival is scoped to the resuming session and only continues mid-turn work; finished tasks stay idle and steerable.

- **New Features**
  - Suspend-not-dispose: shutdown suspends children; resume revives only that session’s children.
  - No re-runs: only mid-turn-interrupted tasks continue; completed tasks don’t.
  - Strict no-revive: canceled, killed, LRU/TTL-evicted, or team shutdown-approved children never revive.
  - Durable and safe: in-process and RPC children persist; resume resolves the exact saved model or fails; unnotified terminal completions reconcile on start; two-phase TTL cleanup; status views show “suspended”.
  - Teams: members suspend and revive with mailbox identity preserved; liveness and recovery events delivered on resume.
  - Config: `task.resume_children` in `omo-config-core` defaults to `true`; set to `false` to restore dispose-on-shutdown.

- **Bug Fixes**
  - Fixed cross-platform path assertion in the wait-command builder test by asserting the JSON-escaped path, ensuring Windows compatibility.

<sup>Written for commit 3b36d965855b73717e0a07f5ca02339d972663e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Reviewed the concurrency core (`admission-lease.ts`) rather than the whole 14.6k-line surface, since that is where a subtle defect would be most expensive and hardest to catch in review.

The design reasoning is correct and the header comment is doing real work. Specifically right:

- **Not reusing `withTaskRecordLock` for the batch section.** That primitive stale-reclaims by mtime after 5s because it guards sub-10ms writes; a long-lived batch holder would be reclaimed underneath itself. Switching to a renewable owner-token lease and confining the mutex to the read-check-rename sections is the correct decomposition.
- **`linkSync` for exclusive create.** Writing the complete body to a temp then `link`-ing into place means `EEXIST` is the contention signal and a reader can never observe a partial body. Correct choice over `O_EXCL` + append.
- **Takeover re-validates both token and staleness under the mutex.** A holder that renews between a waiter's observation and its CAS is left alone. Never delete-then-create, so the loser mutates nothing.
- **`release` is itself a CAS.** A displaced holder cannot delete its successor's lease on the way out. This is the failure mode that usually survives into production, and it is handled.
- **`staleMs = 3 × renewMs`** so a single wedged renewal tick cannot displace a live holder.

One real bug in the acquire loop:

```ts
for (;;) {
  if (tryCreateLease(...)) return { kind: "acquired", ... }
  const observed = readLeaseBody(path)
  if (observed === "missing") continue   // <-- unbounded
  ...
  if (Date.now() - startedAt >= timing.acquireTimeoutMs) return { kind: "contended" }
  await delay(timing.retryMs)
}
```

The `missing` branch `continue`s **before** both the timeout check and the `delay`. If a lease is being rapidly created and released by other processes, this thread can spin: `tryCreateLease` loses the `link` race, the subsequent read observes `missing`, and it loops immediately with no sleep and no deadline evaluation. That is a hot CPU spin, and — worse — `acquireTimeoutMs` is not enforced on that path, so the documented "never throws, yields contended past the timeout" contract does not hold. A pathological interleaving never returns.

The other two exit paths are fine; only `missing` skips both guards. Fix is to check the deadline before continuing:

```ts
if (observed === "missing") {
  if (Date.now() - startedAt >= timing.acquireTimeoutMs) return { kind: "contended" }
  await delay(timing.retryMs)
  continue
}
```

Given `retryMs` is 25ms and this only bites under active create/release churn on the same parent session, it is unlikely to show up in the chaos bench unless that specific interleaving is generated — worth confirming whether the 200-iteration interleavings cover concurrent *release* during acquisition, not just concurrent acquisition.

Two smaller observations, non-blocking:

- `isOwner()` reads the lease file on every call with no caching, and the PR states it is "checked before every mutation inside the critical section." For a large batch that is one `existsSync` + `readFileSync` + `JSON.parse` per mutation. Correct, but worth knowing it is O(mutations) syscalls if batches get big.
- `tryCreateLease` calls `rmSync(tmp, { force: true })` in `finally`, which runs even on the success path after `linkSync`. That is correct — `link` creates a second directory entry so removing the temp name is right — but it means every successful acquisition does an extra unlink. Fine, just noting it is intentional.

The `shutdown_approved` members not being filtered from revival — found by live QA and fixed in `ee792ec39` — is exactly the class of bug that only surfaces against a real binary. Good that the QA drivers caught it rather than the unit suite.

CI is still pending across the board on this one, so I have not evaluated test results. Given the size and that `dev` currently has an unrelated flaky `pollForCompletion` failure on macOS plus a `TypeScriptSourceParser` connection fragility under load, expect some noise unrelated to this PR.

code-yeongyu:
CI result on `fa61dd8`: `test (windows-latest)` and `senpi-compatibility (windows-latest)` fail; ubuntu and macOS all pass. The failure is a **test-side quoting artifact, not a product defect**.

`agent-side wait commands > #given a record wait #when building the command #then name and status are argv`:

```
Expected to contain: "\\tmp\\project\\.omo\\senpi-task\\tasks"
Received: node -e "..." "\\\\tmp\\\\project\\\\.omo\\\\senpi-task\\\\tasks" "lruone" "completed"
```

Read past the JSON display escaping: the expectation is the literal `\tmp\project\.omo\senpi-task\tasks` (single backslashes), while the built command contains `\\tmp\\project\\...` (doubled). The command builder is correctly escaping backslashes for embedding into a shell-quoted argument; the assertion compares against the *unescaped* form.

On POSIX the separator is `/`, nothing needs escaping, and escaped and raw forms are identical — which is why this only fires on Windows. Same class as the `sg` resolver fixtures on #6575: a POSIX-shaped assertion meeting real Windows path semantics.

Two ways to fix, depending on what the test is actually meant to pin:

- If the point is "the tasks dir is passed as argv rather than interpolated into the script body", assert on the *parsed* argv rather than the raw command string — that is separator-agnostic and survives any future quoting change.
- If the raw string genuinely matters, build the expectation with the same escaping the builder uses instead of hardcoding a literal, e.g. derive it from `path.join(...)` and apply the identical escape helper.

Hardcoding `\\tmp\\project\\...` in the expectation would also work but re-encodes the escaping rule in two places, so it will drift.

Worth noting the assertion is verifying a real invariant that is worth keeping — the QA driver passes `dir`/`name`/`want` as `process.argv` entries rather than string-interpolating them into the `node -e` body, which is what stops a task name containing a quote from breaking the poller. The escaping the builder does is correct; only the expectation is wrong.

Everything else on this run is green, including all three `typecheck` targets and `test (ubuntu-latest)` / `test (macos-latest)`, so the 1313-test `senpi-task` suite is not implicated.

Separately, my earlier review comment flagged a genuine acquire-loop issue in `admission-lease.ts` (the `missing` branch skipping both the deadline check and the retry delay) — that one is a product concern and independent of this CI failure.

Issue #6605, opened 2026-08-05
[Bug]: compaction emits an orphaned tool_use (Anthropic 400, non-retryable) and tool-pair-validator is inert against OpenCode's real Part model


MoerAI:
### Prerequisites

- [x] I will write this issue in English
- [x] I have searched existing issues to avoid duplicates (closely related but distinct: #6303, #6528, #6452, #4053, #4032, #3996, #3014, #381)
- [x] I have read the documentation

### Bug Description

Two separate defects, one report, because the second is why the first is not caught.

**1. A session compaction can emit an orphaned `tool_use` and permanently wedge the session.**
The compaction summarization request is rejected by Anthropic with a **non-retryable** HTTP 400. The compaction assistant message is persisted with zero parts and an `error`, so the session can never compact again and never advances.

**2. `tool-pair-validator` — the plugin's only guard against exactly this 400 — is inert against OpenCode's real message model.**
It is written against part types `tool_use` and `tool_result`. **Neither exists in OpenCode's `Part` union.** Every "repair" it performs is silently discarded before the request is built, and the hook has therefore never prevented a single one of these 400s.

### Steps to Reproduce

1. Run a long Anthropic session (`claude-opus-5`) until auto-compaction triggers a second time, i.e. a session that already has one completed compaction with a `tail_start_id`.
2. Auto-compaction fires.
3. Anthropic rejects the compaction request with `400 invalid_request_error`, `isRetryable: false`.
4. Every subsequent compaction attempt fails identically. The session is dead.

### Expected Behavior

The compaction request always satisfies Anthropic's `tool_use` → `tool_result` adjacency rule, and `tool-pair-validator` actually enforces that invariant on the array OpenCode hands it.

#### Defect 1 — orphaned `tool_use` in the compaction request

Observed on opencode `1.18.12`, session `ses_03afd1a30ffeovA7meXjA1z9yp`, twice (00:28:28Z and 00:52:50Z on 2026-08-05), same tool id both times:

```
APIError 400 (isRetryable: false)
messages.2: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01PCXjcagoMAca32awicQHce.
Each `tool_use` block must have a corresponding `tool_result` block in the next message.
```

Confirmed from the plugin log that this is the **compaction summarization request**, not a normal prompt turn:

```
[...]

PR #6607, opened 2026-08-05
fix(tool-pair-validator): operate on OpenCode's real Part model

MoerAI:
## Summary

`tool-pair-validator` is the hook that is supposed to stop Anthropic's `tool_use ids were found without tool_result blocks immediately after` 400 from ever reaching the wire. It could not do that job, because it was written against part types `tool_use` and `tool_result` that **do not exist** in OpenCode's `Part` union. `getToolResultID()` therefore always returned nothing, every tool id looked unpaired, and the "repair" it wrote was a foreign part type that `MessageV2.toModelMessagesEffect` discards before the request is built. This PR rewrites the hook against OpenCode's real message model so the guard actually works.

Behaviour change: an assistant `tool` part that is still `pending` / `running` is now settled in place into a terminal `error` state, so OpenCode itself emits the matching `tool_result`. The fabricated `tool_result` parts and the id-less synthetic `"Recovered missing tool results..."` user turn are gone.

**This PR does not claim to eliminate the orphan reported in #6605.** Replaying that session's real rows through `ai@5.0.226` `convertToModelMessages` produces a clean, fully paired array, so that particular orphan is introduced below the plugin's last hook point. What this PR fixes is that the guard meant to catch this class of failure was inert.

## Changes

**Detection (`tool-part-ids.ts`, `types.ts`)**
- Replaced `getToolUseID` / `getToolResultID` / `extractUniqueToolUseIDs` / `extractToolResultIDs` with `getToolCallID`, `getToolStatus`, `isTerminalToolStatus` and `findUnpairedToolParts`, all keyed off the real `ToolPart` shape (`type: "tool"`, `callID`, `state.status`).
- Dropped the `ToolUsePart` / `ToolResultPart` / synthetic `TextPart` types; `TransformPart` is now just the SDK `Part`.
- A missing or unrecognised status counts as non-terminal, so malformed data errs toward being repaired.

**Repair (`tool-result-repair.ts`, `message-transform.ts`)**
- `repairUnpairedToolParts()` rewrites the part's own `state` to `{ status: "error", input, error, time }`, preserving the recorded `input` and `time.start`. Nothing is inserted into, or removed from, the message array.
- Cross-message index walking is gone; pairing is a per-message property now, which is what the conversion actually does.
- Tracked background subagent sessions keep the diagnose-only path from #4032 / #3996.

**Tests**
- `hook.test.ts` rebuilt on real `ToolPart` fixtures (`hook.test-support.ts` mirrors the SDK shape for all four states).
- New `conversion-invariant.test.ts` models `toModelMessagesEffect` plus the AI SDK tool-call/tool-result split and asserts (a) no orphaned `tool_use` survives the hook and (b) the hook adds no part type outside OpenCode's `Part` union.
- Two `messages-transform.test.ts` cases that fed the hook `tool_use` parts OpenCode never produces were rebuilt on real `running` tool parts. They are why the inert repair went unnoticed.

## QA & Evidence

Full bundle: [`.omo/evidence/20260805-tool-pair-validator-real-parts/`](.omo/evidence/20260805-tool-pair-validator-real-parts/README.md)

**1. Failing-first, to prove the new tests are not vacuous**
- *What was tested:* new tests run against the OLD implementation (source stashed, tests kept).
- *Observed:* `1 pass 3 fail`. `findOrphanedToolCalls(...)` returned `["toolu_stuck"]` (orphan survives the old repair) and `collectForeignPartTypes(...)` returned `["tool_result"]` (old repair injects a discarded part type).
- *Artifact:* `unit-red-before-fix.txt`
- *Why sufficient:* both halves of the defect are detected, so a regression cannot pass silently.

**2. Real OpenCode, isolated sandbox - healthy session**
- *What was tested:* plugin built from this worktree, loaded by absolute path into real `opencode` 1.18.13 with `XDG_DATA_HOME`/`XDG_CONFIG_HOME`/`XDG_STATE_HOME`/`XDG_CACHE_HOME`/`TMP` all under a temp sandbox. `opencode run "...bash echo omo-qa-probe..." --format json`.
- *Observed:* tool part came back `status: "completed"`, model replied `DONE`, `OPENCODE_EXIT=0`, and **zero** `[tool-pair-validator]` lines. The hook is a correct no-op on well-formed sessions.
- *Artifact:* `opencode-qa-plugin.log`

**3. Real OpenCode, isolated sandbox - session holding a genuine `running` tool part**
- *What was tested:* the completed tool part in the sandbox DB was rewritten to `state.status = "running"` (removing the source of its `tool_result`), then the same session was continued.
- *Observed:*
  ```
  [tool-pair-validator] Settled unpaired tool parts into a terminal error state
    {"assistantMessageID":"msg_fd0767cea001yeEqWlzEoLEZYK",
     "repairedToolCallIDs":["call_00_p1uAgdhIpxlkWpaFBKx72999"],
     "previousStatuses":["running"]}
  ```
  The follow-up request succeeded (`OPENCODE_EXIT=0`, model replied `OK`). Under the old implementation this input produced `Repaired missing tool_result blocks` and changed nothing on the wire.
- *Why sufficient:* the hook is proven to fire on genuine `ToolPart` data, on the same transform hook OpenCode also invokes for compaction (`compaction.ts:350`).

**4. Non-destructive + isolation**
- *Observed:* `SANDBOX_PERSISTED_TOOL_STATUS=running` after the run, so the hook rewrites only the per-request transform array and never stored session state. `REAL_DB_SESSIONS_BEFORE=2862` / `REAL_DB_SESSIONS_AFTER=2862`, so the real `opencode.db` was untouched.

**5. Automated gates**
- `bun test packages/omo-opencode/src/hooks packages/omo-opencode/src/plugin` - 2688 pass, 0 fail (340 files)
- `bun run typecheck` - exit 0
- `bun build packages/omo-opencode/src/index.ts ...` - exit 0

## Risks & Residuals

- **The upstream orphan behind #6605 is not fixed.** Mitigated only in the sense that the guard now works for the class of orphan a plugin can see; the compaction-specific one is introduced after the last hook point and is tracked separately in #6605.
- **Session corruption (the #3996 failure mode)** - mitigated: the hook no longer inserts or removes messages, and evidence step 4 shows stored state is untouched.
- **Behaviour change for `pending`/`running` parts** - accepted. OpenCode already converts those to an `output-error` result itself, so settling them early matches what the harness does rather than fighting it, and the healthy-session run confirms no perturbation.
- **Full `bun run build` on Windows** - blocked by a pre-existing `'rm' is not recognized` failure in the vendored `packages/lsp-tools-mcp` build, unrelated to this change. The plugin bundle step was run directly and passes; CI covers the rest.

## Related Issues

Part of #6605.

---
## Summary by cubic
Rewrites the tool‑pair validator to operate on OpenCode’s real `tool` parts and settle non‑terminal calls in place so requests don’t hit Anthropic 400s. QA was re-captured in a hermetic sandbox (HOME and XDG isolated) showing repairs fire on real data and don’t touch stored state; this does not fix the compaction orphan in #6605.

- **Bug Fixes**
  - Detect unpaired tool calls via real `tool` parts (`callID`, `state.status`), treating unknown status as non‑terminal.
  - Settle `pending`/`running` parts in place to `{ status: "error" }`, preserving `input` and `time.start`; no foreign part types or extra messages; background subagent sessions remain diagnose‑only.
  - Evidence: logs with full home isolation show 0 host‑config references and successful follow‑up requests.

- **Refactors**
  - Added `getToolCallID`, `getToolStatus`, `isTerminalToolStatus`, `findUnpairedToolParts`; removed `tool_use`/`tool_result` handling and cross‑message insertion.
  - Tests rebuilt on real `tool` parts from `@opencode-ai/sdk` with a conversion‑invariant gate; failing‑first run is scoped to `conversion-invariant.test.ts` (old code cannot load `hook.test.ts` due to the new `INTERRUPTED_TOOL_ERROR` symbol). Evidence under `.omo/evidence/20260805-tool-pair-validator-real-parts/`.

<sup>Written for commit f03dcd026854550f165e22fe69161df5f7572fb5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6607?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Fixed the P1: the failing-first artifact was empty because the capture variable did not survive the shell invocation. Re-ran the new tests against `origin/dev`'s tool-pair-validator sources and committed the real captured output in `937b420` - `1 pass / 3 fail`, showing the surviving `toolu_stuck` orphan and the discarded `tool_result` part type.

MoerAI:
Fixed the second P1 in `0bbef82`. You were right: the first capture sandboxed only the XDG dirs, so the Claude Code compatibility loaders still read the host `~/.claude.json` and pulled in a user MCP server. Re-ran the whole QA with the user home isolated too (`USERPROFILE` / `HOME` / `HOMEDRIVE` / `HOMEPATH` pointed at the sandbox). New capture: 0 matches for `.claude` / `.agents` / `gitnexus` in the log, plugin reports `Loaded 0 plugins with 0 commands, 0 skills, 0 agents, 0 MCP servers` on both runs, and the `Settled unpaired tool parts into a terminal error state` line still fires on a genuine `running` tool part. Isolation is now measured tightly around a single run (`opencode db path` resolves to the sandbox DB; real session count 2863 -> 2863).

MoerAI:
Fixed the P2 in `f03dcd0`. `hook.test.ts` imports `INTERRUPTED_TOOL_ERROR`, which only exists after the fix, so it aborted with a module error instead of actually running against the old sources. Re-captured the red run against `conversion-invariant.test.ts` alone - that file is written against the shared contract and executes unchanged on both implementations - and stated the exclusion explicitly in the README. New artifact: `1 pass / 2 fail` with the surviving `toolu_stuck` orphan and the discarded `tool_result` part type.

Issue #6608, opened 2026-08-05
CI: test (windows-latest) red on dev; LSP-daemon-adjacent flakes producing false red on contributor PRs


code-yeongyu:

code-yeongyu:
Two more Windows-only failures, on PR #6395 (`fix/6376-shared-skills-root-path`), run [30994023601](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/30994023601) — neither related to shared-skills path resolution:

```
test (windows-latest)
(fail) CodeGraph command runner > #given the command spawns a hanging descendant
       #when the timeout expires #then the process tree is reaped        [1296.00ms]

codex-compatibility (windows-latest)
(fail) install-codex legacy agent purge > #given custom file using retired reviewer
       filename #when installing omo #then preserves the unproven user TOML [20031.00ms]
```

`test (macos-latest)` pass 5m47s, `codex-compatibility (macos-latest)` pass 2m45s, all three `typecheck` pass. This is also the first time `codex-compatibility (windows-latest)` has appeared in this cluster.

Updated tally across five runs. The Windows failures fall into two buckets:

**Process reaping / lifecycle on Windows**
- `reapLspDaemons live ownership` (#6607) — 15016.00ms
- `CodeGraph command runner` process-tree reaping (#6395) — 1296.00ms

**Filesystem staging / long-running setup**
- `Senpi LSP daemon runtime staging` (`dev`) — `fs.cp` / `copyDir`
- `install-codex legacy agent purge` (#6395) — 20031.00ms
- `normalizeToolArgSchemas` (#6607) — 26500.00ms

The 15016 / 20031 / 26500 ms durations are deadline-expiry shaped rather than assertion-failure shaped.

Four different PRs have now been marked red by failures none of them caused: #6601, #6602, #6607, #6395.

code-yeongyu:
Second data point, on a contributor PR this time (#6395, run 30994023601). Same shape: Windows-only red, non-Windows green, and every failure is a timeout rather than an assertion.

**`test (windows-latest)`** — 13219 pass / 48 skip / **1 fail**

```
(fail) CodeGraph command runner > #given the command spawns a hanging descendant #when the timeout expires #then the process tree is reaped [1296.00ms]
```

**`codex-compatibility (windows-latest)`**

```
(fail) install-codex legacy agent purge > #given custom file using retired reviewer filename #when installing omo #then preserves the unproven user TOML [20031.00ms]
[...]

PR #6613, opened 2026-08-05
feat(team-mode): add herdr as alternative multiplexer backend for TeamMode visualization

Qiiks:
## Summary

Adds [herdr](https://herdr.dev) (terminal-native agent multiplexer, 0.8.0) as a supported multiplexer backend for TeamMode pane visualization, mirroring the existing `team-layout-tmux` module. TeamMode core is file-based and runs without any multiplexer; this adds an optional herdr pane view (split/rename/run member panes, close on delete, stale-pane sweep) selectable via a new `team_mode.multiplexer` config field.

## What changed

- **New package `packages/herdr-core/`** — harness-neutral herdr primitives:
  - `herdr-detect.ts` — `isHerdrEnvironment()` (HERDR_ENV / HERDR_SOCKET_PATH), caller pane id
  - `runner.ts` — `runHerdrCommand` (spawn + timeout + retry, mirrors `runTmuxCommand`)
  - `pane-utils.ts` — split/rename/run/close/list command builders + parsers (handles herdr's JSON output and alphanumeric workspace ids like `wJ:p1`)
- **`packages/team-core/src/team-layout-herdr/`** — mirrors `team-layout-tmux/`:
  - `layout.ts` — `createHerdrTeamLayout` / `removeHerdrTeamLayout` with the same DI + failure semantics (log + return null, never throw)
  - `resolve-caller-herdr-pane.ts` — caller pane from `HERDR_PANE_ID`
  - `sweep-stale-herdr-panes.ts` — closes `omo-team-<runid>-<member>` panes of inactive runs (label parsed from `pane get` JSON)
- **`packages/team-core/src/multiplexer.ts`** — `resolveTeamMultiplexer(config)`: `auto` (default) → herdr when inside herdr, else tmux; explicit `tmux`/`herdr` overrides
- **Config** — `TeamModeConfigSchema` gains `multiplexer: "auto" | "tmux" | "herdr"` (default `auto`)
- **Dispatch** — `activate-team-layout.ts`, `delete-team.ts`, `create.ts` pick the herdr vs tmux backend via the resolver; stale sweep dispatches to the herdr pane sweep when in herdr
- **server-health.ts** — probe `/health` first, fall back to `/global/health` (opencode serves `/health`; stock OMO only probed `/global/health` → 404)
- **create-managers.ts** — wire the team-core logger to the OMO logger at startup so layout failure/skip decisions surface in the plugin log (`setTeamCoreLogger` lived in an unimported module, leaving team-core logs as no-ops)
- **Docs** — `docs/guide/team-mode.md` herdr section (multiplexer field, herdr visualization section, diagnostics)

## Why

Users who run opencode inside herdr (which has no tmux shim) previously got TeamMode core without any pane visualization. This adds a first-class herdr backend that mirrors the proven tmux one, detected automatically via herdr's injected env vars.

## Notes on server port

The herdr layout's member pane runs `opencode attach <server-url> --session <id> --dir <dir>`. The server URL resolves through the existing tmux-session-manager path: `ctx.serverUrl`, else `http://localhost:${OPENCODE_PORT ?? 4096}`. For TeamMode pane visualization, launch opencode with a fixed port (e.g. a wrapper that sets `OPENCODE_PORT` and passes `--port`) so the member pane's attach reaches the server — same constraint as the tmux path (issue #3963).

## QA / Evidence

- `.omo/evidence/20260805-herdr-teammode/EVIDENCE.md`
- 587 tests pass (herdr-core + team-layout-herdr + tmux-core + team-mode + tmux-subagent), `tsgo --noEmit` clean on herdr-core/team-core/omo-opencode
- Full `bun run build` passes (exit 0) — repo CI gate green
- All herdr CLI surfaces the plugin emits (`pane split/rename/run/list/close`, `workspace create/close`) validated live against herdr 0.8.0-preview (scratch workspace, cleaned up)
- Live validation caught and fixed 3 real bugs: alphanumeric workspace ids, JSON list output, JSON label read

### LIVE team-mode end-to-end (user-directed, 2026-08-05) — final round (fixed port)

- Wrapper launches `opencode --port 37857` with `OPENCODE_PORT=37857`; server verified listening on 37857
- `team_create` (1-member team `smoke-16`) → runtime state `tmuxLayout.paneIds: ["wR:p2"]` — herdr backend spawned the pane
- `herdr pane read wR:p2` → **member's opencode TUI rendered inside the pane** (task instructions, status line) — attach connected to the fixed port
- Member completed `HERDR_TEAM_OK_V16` in 6s, displayed in-pane
- Closure: `team_shutdown_request` → `team_approve_shutdown` → `team_delete` (`removedLayout: true`)
- `herdr pane list` after delete: only `wR:p1` — pane closed correctly

## Out of scope (follow-up PRs)

- `[opencode].tmux` general integration (subagent panes + interactive_bash) herdr backend
- `tmux_visualization` default flip — stays false (opt-in), herdr backend is automatic when inside herdr

---
## Summary by cubic
Adds `herdr` as an alternative multiplexer backend for TeamMode visualization, selectable via `team_mode.multiplexer` (default `auto` with automatic herdr detection). Team panes split/rename/run/close in herdr like tmux, with stale-pane sweeping and safe failure behavior.

- New Features
  - New package `@oh-my-opencode/herdr-core` for herdr detection, CLI runner, and pane utils.
  - Herdr team layout in `@oh-my-opencode/team-core` (re-exported in `@oh-my-opencode/omo-opencode`): member panes spawn via `herdr pane split/run`, titled `omo-team-<runid>-<member>`, closed on delete; stale panes swept on create.
  - Config: add `team_mode.multiplexer` (`"auto" | "tmux" | "herdr"`, default `"auto"`). Resolver picks herdr when inside herdr, else tmux. Create/activate/delete dispatch to the selected backend. Docs updated with herdr setup and diagnostics.

- Bug Fixes
  - Server health now probes `/health` first, then `/global/health`, with separate timeout budgets to avoid starvation.
  - Wire the team-core logger to the OMO logger so layout decisions and failures appear in plugin logs.
  - Declare `@oh-my-opencode/herdr-core` as a dependency of `@oh-my-opencode/omo-opencode` to fix CI import resolution errors.
  - Add `multiplexer` to Senpi team transport defaults and rebuild `omo-senpi` extension bundles to keep senpi-compat checks green.
  - CI: register `@oh-my-opencode/herdr-core` in managed package audits and root `typecheck:packages`; update the re-export shim inventory snapshot.
  - Tests: align `omo-opencode` default-config test with `multiplexer: "auto"` and update server-health cache tests to account for dual-path probing (8 fetches across two failed calls).
  - Sync to upstream through #6659 and regenerate dist artifacts (`codegraph` CLI/serve, `install-local`) and the `omo-senpi` `omo.js` bundle to keep hashes consistent.

<sup>Written for commit 67ec0a35de18b6e11b9ad2b729325055e8cbf7ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6613?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Substantial piece of work — the live end-to-end evidence (real pane spawn, member TUI rendering in-pane, `HERDR_TEAM_OK_V16`, clean teardown verified via `herdr pane list`) is well beyond what a feature PR usually ships with. Mirroring `team-layout-tmux/` rather than abstracting over both backends is also the lower-risk shape.

Two things from a first pass.

**1. `senpi-compatibility (ubuntu-latest)` is already failing, and it is real — not the #6608 Windows flake cluster.** Fails in 17s on a hard check:

```
omo-senpi extension build is not current: stale-output
output=packages/omo-senpi/plugin/extensions/omo.js
```

This PR modifies `packages/senpi-task/src/team/runtime-config.ts` (adding `multiplexer` to the transport defaults) without regenerating the committed `omo-senpi` extension bundle. CI compares built output against the committed artifact, so this will fail deterministically on every platform until the bundle is rebuilt and committed.

**2. `packages/tmux-core/src/tmux-utils/server-health.ts` is a shared-path behaviour change, not herdr-scoped.** It affects existing tmux users too, and two details are worth a look:

```ts
const timeout = setTimeout(() => controller.abort(), timeoutMs)
...
const response = await probeServerHealth(fetchImplementation, serverUrl, controller.signal)
```

```ts
for (const path of ["/health", "/global/health"]) {
  const url = new URL(path, serverUrl).toString()
  const response = await fetchImplementation(url, { signal }).catch(() => null)
  if (response?.ok) return true
}
```

- **The single `AbortSignal` now covers both probes sequentially.** `timeoutMs` was previously the budget for one request; it is now the budget for two. Against a slow server, `/health` can consume most of the window and `/global/health` then runs on a nearly-expired signal — so the fallback may not reliably get a chance to answer. Consider a per-probe timeout, or probing concurrently.
- **Users on builds that only serve `/global/health` now pay an extra failed round-trip** on every health check, since `/health` is tried first and 404s.

Neither is fatal, but since it rides the shared tmux path it is worth confirming the ordering and timeout budget are intentional rather than incidental to the herdr work.

Also worth noting for reviewers: `create-managers.ts` wiring `setTeamCoreLogger` is an independent bug fix (team-core logs were previously no-ops because the module was never imported). That is a genuine improvement but is unrelated to herdr, and might be worth calling out in the changelog separately.

code-yeongyu:
Good — `senpi-compatibility` is green now (1m26s). The remaining four are all deterministic and cross-platform (ubuntu/macOS/Windows all fail identically), so none are the #6608 flake cluster.

**1. `isServerRunning > does not cache failed result` — there is a second copy of this test you have not updated.**

```
packages/omo-opencode/src/shared/tmux/tmux-utils.test.ts:139

138 |     // then - should call fetch 4 times (2 attempts per call, 2 calls)
139 |     expect(fetchMock.calls.length).toBe(4)
Expected: 4
Received: 8
```

Your updated copy in `packages/tmux-core/src/tmux-utils.test.ts` passes. The one under `packages/omo-opencode/src/shared/tmux/` is a duplicate that still carries the pre-change expectation, and the double-probe doubled its fetch count from 4 to 8.

This is also concrete evidence for the timeout concern in my earlier comment: the probe loop genuinely doubles the number of requests per `isServerRunning` call, which is why the shared `AbortSignal` budget is worth revisiting.

**2. `TeamModeConfigSchema > returns the default team mode config`**

```
packages/omo-opencode/src/config/schema/team-mode.test.ts:17
+   "multiplexer": "auto",
- Expected  - 0
+ Received  + 1
```

Expected-defaults object needs `multiplexer: "auto"` added.

**3 & 4. `package registration audit`** — two failures:

```
(fail) #given managed packages #when root registration is audited
       #then workspaces typecheck and dev deps stay aligned          [  3.00ms]
(fail) #given exact re-export shim scan #when inventory docs are checked
       #then total targets and every path are present                [146.00ms]
```

New `packages/herdr-core/` needs registering in the root workspace config and in the inventory docs the audit scans.

All four look mechanical. The duplicated `tmux-utils.test.ts` is the one worth a second look, since two copies of the same suite with drifted expectations will keep causing this for anyone touching `server-health.ts`.

Qiiks:
Thanks for the detailed pass — responding point by point.

**1. senpi-compatibility** — resolved in `15fa469` (rebuilt + committed the `omo-senpi` extension bundles after the `runtime-config.ts` change). Green on all three OS in the latest run.

**2. `server-health.ts` shared AbortSignal** — valid concern, and it was indeed incidental rather than intentional. Fixed in `fd0480f`: each probe (`/health`, `/global/health`) now owns its own `AbortController` with the full 3s timeout, so a slow `/health` can no longer starve the fallback. The `/health`-first ordering *is* intentional — live-verified against opencode 1.18.11, which serves `/health` (`{"ok":true}`) and 404s `/global/health`. The cost for legacy-only builds is one fast 404 round-trip per attempt (404 is immediate, not a timeout), which is the price of supporting both server generations on a shared path. Happy to probe concurrently if you'd prefer that trade instead.

**3. `create-managers.ts`/`setTeamCoreLogger`** — you're right it's an independent fix (team-core logs were no-ops because `features/team-mode/index.ts` was never imported; the wiring in `create-managers.ts` actually runs at plugin startup). It's called out in the PR description under a separate bullet so it can be split out if you'd rather have it as its own PR.

**On the remaining test failures** — the four you listed in the second comment (duplicated `tmux-utils.test.ts` expectation, `TeamModeConfigSchema` default, the two package-registration audits) are all fixed (`67e53f0` + `2cd75c9`): typecheck ✓ on all three OS, test ✓ on ubuntu+macos, senpi ✓, codex ✓, build ✓.

The one remaining red is `test (windows-latest)`: 18 failures in `team-mailbox`/`team-state-store` file-lock tests (fsync `state.lock` timeouts, chaos bench at 124s) — no overlap with anything this PR touches, and the suite passes 55/55 locally and on ubuntu/macos. That matches the flake cluster you're tracking in #6608, so I've left it alone rather than papering over it.

code-yeongyu:
[sisyphus-bot] Substantial herdr backend implementation with good maintainer engagement. Blocking on test (windows-latest) FAILURE — the Windows test failure needs resolution before this can proceed. The new `multiplexer` config field and herdr-core package are well-scoped, but the CI failure must be addressed.

Qiiks:
Following up on the windows-latest blocker with evidence — I believe this is a pre-existing CI-environment flake, not a regression from the herdr work.

## Evidence

**1. The PR does not touch any of the failing tests.** Diff of `feat/herdr-team-mode` against `dev` for every failing file is empty:

```
git diff --name-only origin/dev HEAD --   packages/omo-opencode/src/hooks/team-session-events/   packages/omo-opencode/src/hooks/team-member-error/   packages/omo-opencode/src/hooks/team-mailbox-injector/   packages/omo-opencode/src/hooks/atlas/   packages/team-core/src/team-mailbox/   packages/team-core/src/team-state-store/
# (empty — zero changes)
```

The 11 failing tests are byte-identical to the base branch.

**2. Every failure is a timeout, not an assertion mismatch.** All 11: `[5015ms]`, `[5093ms]`, `[5563ms]`, `[5687ms]`, `[6438ms]`, `[10016ms]` — each lands exactly on the test's 5s/10s timeout. The tests are timing out waiting on file-lock/fsync (`state.lock`) or cross-session event delivery, which is the classic Windows CI pattern under parallel-suite load.

**3. The same family fails on `dev`'s own Windows CI.** The merge of #6630 (run 31157132099) fails `test (windows-latest)` with the same shape: `createTeamService lead messaging ... member tools load before inherited OMO extensions [5360.00ms]` — a timeout in the same team-state/team-mailbox family, on a commit that contains none of this PR.

**4. They pass locally on Windows.** Full run of the failing files on a Windows dev box: `40 pass / 0 fail, 2.92s`.

## Proposal

These are the same nondeterministic CI failures #6630 is addressing via separate test-stability PRs. Options, in order of preference:

1. **Flake-check first** — re-run `test (windows-latest)` on this PR head; if the suite passes on a clean run, that confirms the flake and we can proceed.
2. If it reproduces reliably, I'll open a **separate test-stability PR** (like #6630) that hardens these specific timeouts — Windows-safe lock-wait budgets or retry — keeping this PR's diff strictly herdr-scoped.

Either way the herdr feature code itself is not implicated: zero diff overlap + local green + identical failures on dev's CI.

MoerAI (changes requested):
The herdr architecture and live QA are strong, but the current head is not reproducibly buildable: all three root test jobs fail the committed Codex installer freshness check because `install-local.mjs` lacks the required `// omo-codex-install:<sourceDigest>:<bodyDigest>` marker. Please regenerate the installer with the repository build command, commit the exact generated artifact, and rerun the full matrix. Approval can follow once the freshness gate is green on all platforms.

Issue #6313, opened 2026-07-24
[Bug]: Model fallback does not trigger on "Upstream request failed" errors from OpenAI-compatible providers


zmlgit:
## Bug Description

When an OpenAI-compatible provider (e.g. `opencode-go` routed through a local proxy like `opencode-go-multi-auth`) returns an error whose message contains **"Upstream request failed"**, the configured `fallback_models` chain is **not triggered**. The user sees the error directly instead of Oh-My-OpenAgent falling back to the next model.

## Reproduction Steps

1. Configure an agent with a model served by an OpenAI-compatible provider that can return upstream failures, for example:

   ```jsonc
   // oh-my-opencode.jsonc
   {
     "model_fallback": true,
     "agents": {
       "sisyphus": {
         "model": "opencode-go/kimi-k2.6",
         "fallback_models": [
           { "model": "zhipuai-coding-plan/glm-5.2" },
           { "model": "kimi-for-coding/k3" }
         ]
       }
     }
   }
   ```

2. Trigger a request where the upstream provider returns a message such as:
   > Error from provider (Console Go): Upstream request failed

3. Observe that the turn fails and no fallback occurs.

## Expected Behavior

`isRetryableModelError()` should classify "Upstream request failed" as retryable, arm the pending model fallback, and the next turn should attempt the next entry in `fallback_models`.

## Actual Behavior

`shouldRetryError()` returns `false` because the message does not match any pattern in `RETRYABLE_MESSAGE_PATTERNS`. The error is treated as terminal and the fallback chain is skipped.

## Root Cause

In `src/shared/model-error-classifier.ts`, `RETRYABLE_MESSAGE_PATTERNS` contains transient-failure patterns like `"connection error"`, `"network error"`, `"bad gateway"`, `"service unavailable"`, and `"try again"`, but it does **not** contain `"upstream request failed"`. This exact phrase is used by OpenCode's OpenAI-compatible provider adapter when the upstream LLM service fails.

## Suggested Fix

Add `"upstream request failed"` to `RETRYABLE_MESSAGE_PATTERNS` in `src/shared/model-error-classifier.ts`:

```typescript
const RETRYABLE_MESSAGE_PATTERNS = [
  // ... existing patterns ...
  "upstream request failed",
]
```

Optionally also add a regression test in `src/shared/model-error-classifier.test.ts` asserting that `{ message: "Error from provider (Console Go): Upstream request failed" }` is classified as retryable.

## Related

[...]

Issue #6255, opened 2026-07-21
bug(model-core): qwen3.5-plus serving failure (503 InternalError.Algo) causes subagent timeouts


niStee:
## Description

The default Qwen Plus model `qwen3.5-plus` in `AGENT_MODEL_REQUIREMENTS` for `explore` and `librarian` subagents currently fails on OpenCode Go endpoints with `503 InternalError.Algo` or times out (>12s).

## Impact

When launching `explore` or `librarian` subagents without an explicit OpenAI key, the execution engine attempts `qwen3.5-plus` as the first non-OpenAI fallback, resulting in hanging or failing subagent tasks.

## Resolution

`qwen3.6-plus` is healthy on OpenCode Go and Bailian, responding in ~3.2s with `HTTP 200 OK`.

Submitted PR #6254 to update the default fallback model from `qwen3.5-plus` to `qwen3.6-plus` across `model-core`, `senpi-task`, and `omo-opencode`.

niStee:
### 🔬 Detailed Research & Empirical Telemetry

Cross-checking official Alibaba Cloud (Model Studio / DashScope) documentation, OpenCode Go endpoints, and live API telemetry:

#### 1. Qwen Generation Evolution
- **Qwen 3.5 Series (`qwen3.5-plus`)**: Introduced native vision-language capabilities with a 1M token context window. Backend serving for legacy `qwen3.5-plus` endpoints on OpenCode Go has degraded in favor of newer 3.6/3.7 checkpoints.
- **Qwen 3.6 Series (`qwen3.6-plus`)**: Enhanced agentic coding & reasoning successor model.
- **Qwen 3.7 Series (`qwen3.7-plus`, `qwen3.7-max`)**: Flagship Qwen 3.7 model tier.

#### 2. Live API Endpoint Telemetry
Probing the endpoints directly via LiteLLM / OpenCode Go gateway:

| Model Identifier | Status Code | Latency | Result |
| :--- | :--- | :--- | :--- |
| **`qwen3.5-plus`** | `503 InternalError.Algo` / Timeout | >12.0s | ❌ Degraded / Unusable |
| **`qwen3.6-plus`** | **`200 OK`** | ~3.2s | ✅ Healthy & Fast |
| **`qwen3.7-plus`** | **`200 OK`** | ~3.5s | ✅ Healthy & Fast |
| **`qwen3.7-max`** | Timeout | >12.0s | ❌ Timeout on Go Gateway |

[...]

PR #6615, opened 2026-08-05
fix(model-core): treat technical mid-stream MessageAbortedError as retryable stream failure (fixes #6424)

### LINKED ISSUE #6424: bug(runtime-fallback): MessageAbortedError from mid-stream provider failures prevents automatic fallback
## Problem

When a provider or proxy connection breaks mid-stream (e.g., LiteLLM's `MidStreamFallbackError`, socket disconnection, network timeout during streaming), OpenCode generates a `MessageAbortedError`. The current error classifier treats ALL `MessageAbortedError` instances as user-initiated cancellations and marks them as non-retryable, preventing automatic fallback to the next model in the chain.

This causes subagent tasks (especially long-running plan executions) to hang indefinitely when the model connection breaks mid-response, with no automatic recovery path.

## Root Cause

In `packages/model-core/src/runtime-fallback-error-classifier.ts`:

```typescript
export function classifyRuntimeFallbackError(error: unknown): RuntimeFallbackErrorType | undefined {
  const message = getRuntimeFallbackErrorMessage(error)
  const errorName = getRuntimeFallbackErrorName(error)?.toLowerCase().replace(/[_-]/g, "")

  if (errorName?.includes("messageabortederror") || errorName?.includes("aborterror")) {
    return "abort"  // <-- ALL MessageAbortedError classified as "abort"
  }
  // ...
}

export function isRuntimeFallbackRetryableError(
  error: unknown,
  retryOnErrors: readonly number[],
  options: RuntimeFallbackRetryOptions = {},
): boolean {
  // ...
  if (errorType === "abort") return false  // <-- abort errors are never retried
  // ...
}
```

The logic was designed to prevent retrying when users explicitly click "Stop" in the UI, but it incorrectly treats technical stream failures the same way.

## Evidence

From `/tmp/oh-my-opencode.log` when LiteLLM's `glm-5.2` stream broke mid-response:

```
[runtime-fallback] message.updated with assistant error {"sessionID":"...","errorName":"MessageAbortedError","errorType":"abort"}
[runtime-fallback] message.updated error not retryable, skipping fallback {"sessionID":"...","errorName":"MessageAbortedError","errorType":"abort"}
```

The task hung because:
1. LiteLLM started streaming `glm-5.2` successfully (HTTP 200)
2. Mid-stream, the connection broke (likely timeout or upstream failure)
3. LiteLLM's internal fallback tried `glm-5.2` again (misconfigured fallback group) and failed
4. The error was injected into the open HTTP stream
5. OpenCode SDK received the broken stream and generated `MessageAbortedError`
6. Plugin classified it as `"abort"` (user cancellation) and skipped fallback
7. `glm-5.1` (the configured fallback) was never attempted

### Part 1: Differentiate User Cancellation from Technical Stream Failure

Modify `classifyRuntimeFallbackError()` to distinguish between:
- **User-initiated cancellation**: Error message contains explicit user cancellation signals like `"The user aborted this request"`
- **Technical stream failure**: Error message contains technical failure patterns like `"MidStreamFallbackError"`, `"socket connection closed"`, `"ECONNRESET"`, `"fetch failed"`, etc.

Implementation approach:
```typescript
export function classifyRuntimeFallbackError(error: unknown): RuntimeFallbackErrorType | undefined {
  const message = getRuntimeFallbackErrorMessage(error)
  const errorName = getRuntimeFallbackErrorName(error)?.toLowerCase().replace(/[_-]/g, "")

  // Check for explicit user cancellation FIRST
  if (isExplicitUserCancellation(message)) {
    return "abort"
  }

  // Then check for technical abort errors (these SHOULD be retryable)
  if (errorName?.includes("messageabortederror") || errorName?.includes("aborterror")) {
    return "stream_failure"  // New error type for retryable stream failures
  }
  
  // ... rest of classification logic
}

function isExplicitUserCancellation(message: string): boolean {
  return /the user aborted/i.test(message) || 
         /user.*cancel/i.test(message) ||
         /manual.*abort/i.test(message)
}
```

Add `"stream_failure"` to `RuntimeFallbackErrorType` and make it retryable in `isRuntimeFallbackRetryableError()`.

### Part 2: Delegate Task Retry Protocol

Enhance `delegate-task` to handle subagent failures gracefully:

1. **On technical failure**: If a subagent fails due to model/provider issues, automatically retry with the next fallback model in the category chain
2. **On user cancellation**: Return a structured message to the parent agent indicating voluntary cancellation, allowing the orchestrator to pause and ask the user for guidance

Implementation in `packages/omo-opencode/src/tools/delegate-task/sync-session-poller.ts`:
- Detect terminal errors that indicate model/provider failure vs user cancellation
- On model failure: trigger fallback retry via existing `sync-task-fallback.ts` mechanism
- On user cancellation: return `[SUBTASK CANCELLED BY USER]` message format

## Expected Behavior

- Default behavior for explicit user cancellations remains unchanged (no retry)
- Technical stream failures (mid-stream breaks, socket errors, connection resets) automatically trigger fallback to the next model
- Subagent tasks that fail due to model issues are retried with fallback models before reporting failure to the parent
- User-initiated cancellations are clearly communicated to parent agents for appropriate handling

## Why This Matters

Long-running plan executions and complex multi-step tasks frequently use subagents with expensive models. When these fail due to transient network/proxy issues, the entire workflow hangs with no recovery path. Automatic fallback on technical failures ensures resilience without requiring manual intervention, while preserving the ability for users to explicitly cancel tasks when needed.

## Related Issues

- #6313: Model fallback does not trigger on "Upstream request failed" errors (partial fix for pre-response failures)
- #6406: feat(runtime-fallback): allow retrying a model before switching to fallback (complementary feature)
- #6268: Default visual-engineering fallback routes opencode-go users to text-only glm-5.2 (related to fallback chain issues)

Marker689:
## Summary

When a provider or proxy connection breaks **mid-stream**, OpenCode surfaces the failure as `MessageAbortedError` (or `AbortError` / `ResponseAborted` for server-closed streams). Both model-core error classifiers treated **every** abort-family name as a user cancellation and therefore non-retryable, so the `fallback_models` configured per agent/category in omo.jsonc were **never consulted** for subagent sessions: a subagent whose primary model connection broke mid-response hung or errored instead of advancing to the next fallback model. This is issue #6424.

This PR distinguishes genuine user/transport cancellation from technical stream failure. A technical `MessageAbortedError` is now retryable, so subagent tasks advance to the configured fallback chain. An explicit user abort message (e.g. `"The user aborted this request."`) stays non-retryable.

## Changes

- **`packages/model-core/src/model-abort-signal.ts`** (new) — `isExplicitUserAbortSignal(message)`: shared, case-insensitive discriminator over 15 explicit user/transport abort phrases (`"the user aborted this request"`, `"request was aborted"`, `"aborted by user"`, ...). One source of truth for both classifiers.
- **`packages/model-core/src/model-error-classifier.ts`** — abort-family names (`messageabortederror` / `aborterror` / `responseaborted`) now return retryable **unless** the message matches the discriminator. This gates the delegate-task sync/background retry paths (`sync-task-runner.ts:64`, `fallback-retry-handler.ts:76`) and the model-fallback event path. All other name/message/status-code logic untouched; `messageabortederror` stays in `NON_RETRYABLE_ERROR_NAMES` as a fallback for name-less callers.
- **`packages/model-core/src/runtime-fallback-error-classifier.ts`** — new `"stream_failure"` error type returned for technical abort-family errors (user-cancel still returns `"abort"`); `isRuntimeFallbackRetryableError` returns `true` for `stream_failure`. `abort` / `context_overflow` remain non-retryable. This gates the runtime-fallback hook (`message-update-handler`, `event-handler`) paths.
- **`packages/model-core/src/index.ts`** — barrel export for the new helper.
- Tests: 20 new classifier cases (model-error-classifier + runtime-fallback-error-classifier: technical→retryable, user-cancel→non-retryable, quota/context-overflow/validation regressions pinned) + 2 consumer tests in `fallback-retry-handler.test.ts` that inject the **real** `shouldRetryError` through the DI seam (mid-stream abort → retry scheduled; user abort → no retry).

## QA & Evidence

Evidence dir: `.omo/evidence/20260806-6424-midstream-abort-fallback/` (01-UNIT.md, 02-SMOKES.md, 03-REPRO.md, raw/).

- **Unit (TDD RED→GREEN):** RED captures `red-t1` (abort-family `shouldRetryError` returned false pre-fix) + `red-t2` (runtime classifier returned `"abort"`, `tryFallbackRetry` refused). GREEN: 91 pass / 0 fail / 166 expect calls across the 4 suites (model-abort-signal, model-error-classifier, runtime-fallback-error-classifier, fallback-retry-handler).
- **Real surface (opencode-qa, isolated XDG/HOME sandbox):** QA-only fork of the fake OpenAI SSE server with three failure modes against an `explore` subagent on `fake-primary` with `fallback_models: ["local/fake-fallback"]`:
  - `http500`: **end-to-end fallback FIRED** — server log `fake-primary` fail → `fake-fallback` `DONE`; plugin log `Retryable error, attempting fallback {errorName:"SessionRetry", errorMessage:"upstream request failed", nextModel:"openai/fake-fallback"}`; parent session received `[BACKGROUND TASK RETRYING]` (failed model → next model) and `[BACKGROUND TASK COMPLETED]`.
  - `drop` (mid-stream socket destroy): opencode 1.18.14's own transport retry absorbs it (30s-capped backoff, no `session.error` surfaces) — documented residual risk below.
  - `sseerror`: session ends idle, no error.
- **No-regression smokes:** isolated `opencode run` with the built plugin loads + completes; `tui-smoke.sh --self-test` 5/5 PASS.
- **Isolation proof:** real DB session count **755 before == 755 after** (`raw/session-count-before.txt` / `raw/session-count-after.txt`). Every opencode spawn ran in isolated HOME + XDG dirs.
- `bun run typecheck` clean; `bun run build` exit 0; scoped regression suites (event-handler, error-classifier, manager, provider-exhaustion) green. Full `bun test`: 35 failures, all pre-existing on `dev` (verified individually: git-bash-mcp dist artifacts, Windows-only shell assertions on Linux, senpi skills materialization, open-design submodule, codex install/doctor artifacts; `manager-claim` and `codex-installer-version` re-ran red on a clean `dev` checkout).

## Risks & Residuals

- **Mid-stream socket drop is absorbed by opencode's own transport retry on 1.18.14** (session stays in `retry` status with 30s-capped backoff; no `session.error` reaches the plugin). The abort-family classifier branch therefore could not be exercised end-to-end through the drop mode on this harness version. The classifier delta is proven deterministically by the unit layer, and the end-to-end fallback path is proven by the `http500` repro. This is a harness-level observation worth tracking separately; the fix is correct for every path where the error does surface (message.updated error parts, session.error, session.status error states).
- **Bare/no-message `MessageAbortedError` → technical → retryable.** If opencode ever emits a user-cancel abort without one of the 15 phrases, a fallback may fire on user cancel. Mitigations: fallback only fires when a chain is configured; the runtime `session.error` path still treats abort-family as cancel via `is-abort-error.ts` (unchanged); cooldowns bound churn.
- **`responseaborted` is newly retryable** (previously unclassified → effectively false). Intended: server-closed stream is a technical failure.
- **Abort-family name takes precedence over a quota message inside the same error** (pinned by test). Intended per the issue recipe.
- **Sync delegate poll path** (`sync-task-runner`) calls `shouldRetryError({ message })` without the error name; mid-stream failures there still rely on message patterns (`"connection error"` / `"network error"` match; arbitrary provider phrasing may not). The primary reported surface (background agents) is fully covered. Out of scope here to avoid threading the error name through the poller.

## Related Issues

- Fixes #6424 (bug(runtime-fallback): MessageAbortedError from mid-stream provider failures prevents automatic fallback)
- Related: #6313 (pre-response failures, partial fix), #6255, #4528

---
## Summary by cubic
Treat mid-stream provider disconnects (`MessageAbortedError`/`AbortError`/`ResponseAborted`) as retryable stream failures so subagent tasks move to configured `fallback_models` instead of hanging. Explicit user cancellations stay non-retryable. Fixes #6424.

- **Bug Fixes**
  - Added `isExplicitUserAbortSignal` to detect explicit user/transport cancel phrases.
  - Updated `model-error-classifier` to mark abort-family errors retryable unless the message is an explicit user cancel.
  - Added `stream_failure` to `runtime-fallback-error-classifier` and made it retryable; `abort` and `context_overflow` remain non-retryable.
  - Exported the new helper via `packages/model-core/src/index.ts`.
  - Added tests for classifiers and the background-agent fallback path using the real `shouldRetryError`.

<sup>Written for commit ecb3146d74ad16927abce78cb35a76bfe5d6edde. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6615?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
[sisyphus-bot] Correctly addresses #6424 root cause — mid-stream MessageAbortedError was unconditionally classified as non-retryable "abort". The phrase-matching approach to distinguish user cancellations from technical failures is pragmatic and defaults to retryable for unknown messages, which is the right direction for this bug. Recommend owner review of the phrase list completeness before merge.

MoerAI (changes requested):
Thanks for separating technical stream failures from explicit user aborts and for exercising the background-agent consumer.

The runtime-fallback `session.error` path still returns through the broad abort guard before the new classifier runs, so a technical `MessageAbortedError` resets as user cancellation and dispatches zero fallbacks:
https://github.com/code-yeongyu/oh-my-openagent/blob/ecb3146d74ad16927abce78cb35a76bfe5d6edde/packages/omo-opencode/src/hooks/runtime-fallback/event-handler.ts#L182-L198

Preserve internal-abort handling first, then route only the explicit `abort` classification to cancellation. Technical `stream_failure` must reach the existing single-session dispatcher.

Add deterministic event-handler coverage proving: technical abort dispatches exactly once, explicit user abort dispatches zero, internal abort preserves its state, and duplicate technical events do not duplicate output.

Please also rebase and rerun current required CI. The existing run was cancelled, and the referenced real-QA evidence directory is absent from this head. Attach reproducible isolated OpenCode evidence for an error shape that actually reaches the plugin, or narrow the claimed surface.

Issue #6614, opened 2026-08-05
[Bug]: Configured reasoning reaches chat.params but stale provider options keep medium


mihazs:
## Prerequisites

- [x] I will write this issue in English (see the [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) and traced the relevant OpenCode and omo-opencode request paths

## Bug Description

An agent-level reasoning override can reach `chat.params.message.variant` but still fail to reach the provider request.

With Hephaestus configured as `openai/gpt-5.6-sol` at `reasoning: "xhigh"`, a downstream observer confirms that PR #5126 injects `message.variant: "xhigh"`. However, `chat.params` still exposes `output.options.reasoningEffort: "medium"`, and the OpenAI Responses request is serialized with `reasoning.effort: "medium"`.

Passing the same value through OpenCode's native `--variant xhigh` flag changes both `output.options.reasoningEffort` and the serialized provider request to `xhigh`.

This means the configuration is found and the hook runs, but the configured reasoning level is applied too late to influence the provider options that OpenCode already constructed.

### 1. Configure Hephaestus with `xhigh`

Relevant sanitized configuration:

```json
{
  "[opencode]": {
    "agents": {
      "hephaestus": {
        "model": "openai/gpt-5.6-sol",
        "reasoning": "xhigh"
      },
      "Hephaestus - Deep Agent": {
        "reasoning": "xhigh"
      }
    }
  }
}
```

The display-name entry is included intentionally to isolate this report from agent-name lookup. The observer below proves the configured value was found and injected.

This block is from `~/.omo/omo.jsonc`; unrelated agents and fields were removed.

### 2. Observe the post-OMO `chat.params` values

Load an observer plugin after `oh-my-openagent@4.19.4`. The raw record included `output.options`; for publication, I projected the Hephaestus record to these fields only:

```ts
{
  agent: input.agent,
  model: `${input.model.providerID}/${input.model.id}`,
  variant: input.message.variant,
  reasoningEffort: output.options.reasoningEffort
}
```

### 3. Capture the serialized request locally

[...]

PR #6617, opened 2026-08-06
fix(omo-opencode): lower canonical reasoning to variant at agent config build time

scw1109:
## Summary

- Lower the canonical `reasoning` field to OpenCode's `variant` field at agent config build time in `applyCategoryOverride` and `mergeAgentConfig`
- `reasoning` takes precedence over the deprecated `variant` and `reasoningEffort`, matching `resolveAgentVariant` precedence
- Add tests for the build-time lowering, precedence, and override interactions

## Problem

The canonical `reasoning` field was added to `AgentOverrideConfigSchema` and `CategoryConfigSchema` (commits d2ccd712, 77135c42), and the chat-time `applyAgentVariant` hook was updated to resolve `reasoning` and lower it via `lowerReasoningForModel`. However, the **build-time** path — `applyCategoryOverride` and `mergeAgentConfig` in `agent-overrides.ts` — never translated `reasoning` to `variant` on the resulting `AgentConfig`.

This caused two observable symptoms:

1. **TUI status bar shows "no variant"** — `AgentConfig.variant` is undefined because `reasoning` was never lowered to `variant` at build time. The user must manually press Ctrl+T to switch.

2. **Stale provider options (#6614)** — OpenCode constructs `output.options.reasoningEffort` from `variant` *before* the `chat.params` hook runs. Since `variant` was unset, the provider request used the default reasoning effort instead of the configured level. The chat-time hook injects `message.variant` correctly, but the pre-computed provider options are not updated.

## Fix

In `mergeAgentConfig`: extract `reasoning` from the override before `deepMerge` (so it does not leak as an extra property on `AgentConfig`), then set `merged.variant = reasoning` after the merge. This makes `reasoning` take precedence over any `variant` that `deepMerge` carried through.

In `applyCategoryOverride`: add `reasoning` to `variant` translation after the existing `variant` assignment, so `reasoning` overrides the deprecated `variant` when both are present.

The chat-time `applyAgentVariant` hook remains unchanged — it still runs as a safety net. When the build-time fix sets `variant` on the `AgentConfig`, OpenCode populates `message.variant` before the hook, so the hook correctly skips (its `message.variant !== undefined` guard). The variant then flows through OpenCode's native option-construction path, which is the preferred fix direction noted in #6614.

## Verification

- `bun test packages/omo-opencode/src/agents/builtin-agents/agent-overrides.test.ts` — 11 pass, 0 fail
- `bun test packages/omo-opencode/src/config/schema/agent-overrides.test.ts packages/omo-opencode/src/shared/agent-variant.test.ts` — 25 pass, 0 fail (no regressions)

## Related issues

- Fixes the TUI "no variant" status bar symptom
- Addresses the build-time root cause of #6614 (configured reasoning reaches chat.params but stale provider options keep medium)

---
## Summary by cubic
Lowered canonical `reasoning` to `variant` at agent config build time so the TUI shows the correct variant and provider options use the configured reasoning effort. `reasoning` now overrides deprecated `variant` and `reasoningEffort`.

- **Bug Fixes**
  - Translate `reasoning` to `variant` in `applyCategoryOverride` and `mergeAgentConfig` (extracted before `deepMerge` to avoid leaking extra fields).
  - Ensure precedence: `reasoning` wins over deprecated `variant` and factory defaults.
  - Add tests for build-time lowering, precedence, and category/agent override interactions.

<sup>Written for commit fc107113a47143627b49a96ff8f5aadbe0317a24. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/agents/builtin-agents/agent-overrides.test.ts
+++ b/packages/omo-opencode/src/agents/builtin-agents/agent-overrides.test.ts
@@ -1,0 +1,182 @@
+import { describe, expect, test } from "bun:test"
+import type { AgentConfig } from "@opencode-ai/sdk"
+import type { CategoryConfig } from "../../config/schema"
+import { applyCategoryOverride, applyOverrides, mergeAgentConfig } from "./agent-overrides"
+import type { AgentOverrideConfig } from "../types"
+
+function makeBase(overrides: Partial<AgentConfig> = {}): AgentConfig {
+  return {
+    instructions: "test prompt",
+    model: "openai/gpt-5.6-sol",
+    mode: "primary",
+    ...overrides,
+  } as AgentConfig
+}
+
+describe("applyCategoryOverride", () => {
+  describe("#given canonical reasoning on category", () => {
+    test("lowers category reasoning to variant on the agent config", () => {
+      // given
+      const base = makeBase()
+      const categories: Record<string, CategoryConfig> = {
+        "ultrabrain": { reasoning: "xhigh" } as CategoryConfig,
+      }
+
+      // when
+      const result = applyCategoryOverride(base, "ultrabrain", categories)
+
+      // then
+      expect(result.variant).toBe("xhigh")
+    })
+
+    test("reasoning takes precedence over deprecated variant", () => {
+      // given
+      const base = makeBase()
+      const categories: Record<string, CategoryConfig> = {
+        "ultrabrain": { reasoning: "high", variant: "medium" } as CategoryConfig,
+      }
+
+      // when
+      const result = applyCategoryOverride(base, "ultrabrain", categories)
+
+      // then
+      expect(result.variant).toBe("high")
+    })
+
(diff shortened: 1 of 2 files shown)


code-yeongyu:
[sisyphus-bot] Verified and merging. Evidence: lowers canonical `reasoning` to `variant` at agent config build time (2 files, +192/-1) with 182 lines of precedence tests. Full CI matrix green (test ×3, build, codex/senpi-compat). Matches the reasoning-unification direction. Thanks for the contribution!

PR #6634, opened 2026-08-07
fix: dedup $id and restore backward-compat required keys in JSON schema

### LINKED ISSUE #6626: Schema assets reject generated configs: duplicate $id, new required keys, and missing _comment allowance
## Summary

The published schema assets currently reject pre-existing and tool-generated configs in strict validators, which causes editor-side false positives even when `oh-my-openagent doctor` passes.

## Root causes

1. **Duplicate `$id` in `assets/omo.schema.json`**
   - The same `$id` (`oh-my-opencode.schema.json`) appears twice in `assets/omo.schema.json` (reported at lines 1207 and 14544).
   - This reproduces on both the current `dev` branch and the `v4.19.4` tag.
   - Strict validators fail to compile the schema because of the duplicate identifier.

2. **New required keys silently invalidate older configs**
   - Flat schema now requires `git_master`.
   - `omo.schema.json` now requires `profiles`.
   - Existing configs that were previously valid now fail validation without any local config change if they reference the moving schema URL.

3. **`_comment` is emitted by the tool but forbidden by the schema**
   - Tool-generated / migrated configs include `_comment` fields for documentation.
   - The schema uses `additionalProperties: false` but does not whitelist `_comment`, so the tool can produce files that its own published schema rejects.

## Why this matters

In our case, the runtime remained healthy and `oh-my-openagent doctor` passed, but VS Codium / JSON-schema tooling reported “config invalid” on a canonical config because the `$schema` URL pointed at a drifting schema that could not cleanly validate the file.

## Suggested fixes

- Give `assets/omo.schema.json` a single unique `$id`.
- Avoid adding new required top-level keys in a backward-incompatible way without migration/pinned schema coordination.
- Either whitelist `_comment` everywhere it is generated/migrated, or stop generating it.
- Publish a stable versioned schema URL so users are not forced to point editors at `dev`.

If useful, I can provide the exact config shape that reproduced the failure on request.

BenSharir:
## Summary
Fixes the confirmed bugs **#1** (duplicate `$id`) and **#2** (backward-incompatible required keys) from issue **#6626**.

Closes #6626

### #1 — Duplicate `$id` (blocks Ajv compilation)
`assets/omo.schema.json` embedded the full `oh-my-opencode.schema.json` schema inline in **two** places (top-level `[opencode]` and each profile's `[opencode]`), both carrying the same `$id`. Ajv 2020 fails to compile the schema in **both** strict and non-strict modes with `reference ".../oh-my-opencode.schema.json" resolves to more than one schema`.

**Fix:** In `script/build-omo-schema-document.ts`, strip `$id` from the inline embedded copies (they are not standalone entry points — only the root `omo.schema.json` keeps its `$id`). Verified: `assets/omo.schema.json` now has exactly one `$id`.

### #2 — Backward-incompatible required keys
Two top-level `required` keys silently invalidated previously-valid configs:
- `assets/omo.schema.json` → `required: ['profiles']`
- `assets/oh-my-opencode.schema.json` → `required: ['git_master']`

**Fix:** Made both fields optional (with their existing defaults) in the Zod sources, then regenerated the assets:
- `packages/omo-config-core/src/schema/config.ts` → `profiles: ...default({}).optional()`
- `packages/omo-opencode/src/config/schema/oh-my-opencode-config.ts` → `git_master: ...default({...}).optional()`

## Out of scope (per triage)
- **#3** `_comment` whitelist — acknowledged as our own internal config artifact, not a tool bug. Not touched.
- **#4** stable versioned schema URL — reasonable ask, larger scope; left as future work.

## Verification
Regenerated both assets via `bun run build:omo-schema` + `bun run build:schema`, then compiled with **ajv@8.20.0**:

| Schema | strict=true | strict=false |
|--------|-------------|--------------|
| `assets/omo.schema.json` | ✅ PASS | ✅ PASS |
| `assets/oh-my-opencode.schema.json` | ✅ PASS | ✅ PASS |

Backward-compat: a config **without** `profiles` and one **without** `git_master` both now validate successfully (previously rejected).

Files changed: 5 (2 regenerated assets + 2 Zod sources + 1 codegen script).

---
## Summary by cubic
Fixes duplicate `$id` in embedded schemas and restores backward-compatible optional keys so configs validate and `ajv` compiles cleanly. Addresses #6626.

- **Bug Fixes**
  - Strip `$id` from inline `[opencode]` schemas in `assets/omo.schema.json` via `script/build-omo-schema-document.ts` to avoid duplicate `$id` collisions.
  - Make `profiles` optional with default `{}` in `packages/omo-config-core/src/schema/config.ts`, removing `required: ['profiles']` from the generated schema.
  - Make `git_master` optional (kept default) in `packages/omo-opencode/src/config/schema/oh-my-opencode-config.ts`, removing `required: ['git_master']` from generated schemas.

<sup>Written for commit 41d7282fdd4e5d6e3164a6c45f5019a71bd2a928. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6634?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -9928,8 +9928,5 @@
       }
     }
   },
-  "required": [
-    "git_master"
-  ],
   "additionalProperties": false
 }
--- a/assets/omo.schema.json
+++ b/assets/omo.schema.json
@@ -1212,7 +1212,6 @@
     },
     "[opencode]": {
       "$schema": "http://json-schema.org/draft-07/schema#",
-      "$id": "https://raw.githubusercontent.com/code-yeongyu/oh-my-openagent/dev/assets/oh-my-opencode.schema.json",
       "title": "Oh My OpenCode Configuration",
       "description": "Configuration schema for oh-my-opencode plugin",
       "type": "object",
@@ -11140,9 +11139,6 @@
           }
         }
       },
-      "required": [
-        "git_master"
-      ],
       "additionalProperties": false
     },
     "[senpi]": {
@@ -14573,7 +14569,6 @@
           },
           "[opencode]": {
             "$schema": "http://json-schema.org/draft-07/schema#",
-            "$id": "https://raw.githubusercontent.com/code-yeongyu/oh-my-openagent/dev/assets/oh-my-opencode.schema.json",
             "title": "Oh My OpenCode Configuration",
             "description": "Configuration schema for oh-my-opencode plugin",
             "type": "object",
@@ -24501,9 +24496,6 @@
                 }
               }
             },
-            "required": [
-              "git_master"
-            ],
             "additionalProperties": false
           },
           "[senpi]": {
@@ -26806,8 +26798,5 @@
       "additionalProperties": {}
     }
   },
-  "required": [
-    "profiles"
-  ],
   "additionalProperties": false
 }
--- a/packages/omo-config-core/src/schema/config.ts
+++ b/packages/omo-config-core/src/schema/config.ts
@@ -45,7 +45,7 @@ export const OmoConfigSchema = z.object({
   "[opencode]": OmoOpenCodeHarnessConfigSchema.optional(),
   "[senpi]": OmoTypedHarnessConfigSchema.optional(),
   "[codex]": OmoTypedHarnessConfigSchema.optional(),
-  profiles: z.record(z.string(), OmoConfigProfileSchema).default({}),
+  profiles: z.record(z.string(), OmoConfigProfileSchema).default({}).optional(),
   _migrations: z.array(z.string()).optional(),
   legacy_migrations: z.record(z.string(), z.unknown()).optional(),
 }).strict()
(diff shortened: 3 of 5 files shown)


MoerAI:
Thanks for the PR, @BenSharir, and for taking the time on this.

This changes generated schemas at `assets/oh-my-opencode.schema.json` and `assets/omo.schema.json` plus maintainer-managed `script/build-omo-schema-document.ts`. The current typecheck and Senpi checks also fail.

I am closing this branch. Please open an issue with the exact schema compatibility regression so the generator change can be handled in a maintainer-owned branch.

PR #6642, opened 2026-08-07
fix(lsp): make built-in LSP MCP work on desktop app, multi-project and Windows

fufuiu:
## Problem

Two bugs made the built-in LSP MCP unusable in realistic setups (desktop app, multi-project, Windows):

### 1. Wrong project root (cwd)

`createBuiltinMcps` calls `createLspMcpConfig` without passing `cwd` (unlike `createCodegraphMcpConfig`). `createLspMcpConfig` falls back to `process.cwd()` — which on the desktop app is the app's launch directory (e.g. `C:\Users\xxx`), not the opened project.

Result: `LSP_TOOLS_MCP_PROJECT_CONFIG` points at the wrong directory and all LSP tools fail with cwd errors:

```
[lsp] Working directory does not exist: C:\Users\xxx\apps\api\src
LSP file path must be inside request cwd
```

### 2. Diagnostics always empty on Windows

Even after fixing cwd, `lsp_diagnostics` always returned 0 diagnostics for files that clearly have type errors (tsc confirms). Two root causes:

- **URI key mismatch**: tsserver publishes diagnostics with a lowercased drive letter and percent-encoded colon (`file:///d%3A/...`), while the client stores/reads with `file:///D:/...`. Pushed diagnostics were stored under a different key than the lookup key, so they were silently dropped (`lookup=MISS`).
- **First-publish timing**: tsserver needs seconds to load a large project before pushing diagnostics, but `diagnostics()` waited only 500ms. Pull (`textDocument/diagnostic`) is not implemented by typescript-language-server (-32601), so the fallback was always empty.

POSIX is unaffected (no drive letter / `%3A` in URIs), which is why this only shows up on Windows.

## Fix

1. `packages/omo-opencode/src/mcp/index.ts`: pass `cwd: options.cwd` to `createLspMcpConfig` (matches the existing `createCodegraphMcpConfig` pattern).

2. `packages/omo-opencode/src/mcp/lsp.ts`: `createLspMcpConfig` resolves `PROJECT_LSP_CONFIGS` against the provided `cwd` and returns a `cwd` field so opencode spawns the MCP subprocess in the correct working directory.

3. `packages/lsp-core/src/lsp/transport.ts`: canonicalize document URIs on both publish and lookup (decode `%3A`, lowercase drive letter) so pushed diagnostics are found.

4. `packages/lsp-core/src/lsp/client.ts`: poll for pushed diagnostics (up to ~10s) before falling back to the (unsupported) pull request; only treat non-empty pull results as authoritative.

## Why it works

`options.cwd` comes from `params.ctx.directory` — the currently opened project, dynamic per project and per session. URI canonicalization fixes the Windows-specific key mismatch. Polling covers slow first project load.

Verified on Windows desktop app: probe file with two TS2322 errors now returns both diagnostics (previously always "No diagnostics found"); goto-definition works; multi-project switching works.

### DIFF
--- a/packages/lsp-core/src/lsp/client.ts
+++ b/packages/lsp-core/src/lsp/client.ts
@@ -6,6 +6,7 @@ import { contextCwd } from "../request-context.js";
 import { LspClientConnection } from "./connection.js";
 import { effectiveExtension } from "./effective-extension.js";
 import { getLanguageId } from "./language-mappings.js";
+import { canonicalUri } from "./transport.js";
 import type {
 	Diagnostic,
 	DocumentSymbol,
@@ -20,6 +21,8 @@ import type {
 
 const POST_OPEN_DELAY_MS = 1000;
 const POST_DIAGNOSTICS_WAIT_MS = 500;
+const DIAGNOSTICS_PUSH_POLL_ATTEMPTS = 10;
+const DIAGNOSTICS_PUSH_POLL_INTERVAL_MS = 1000;
 
 export class LspClient extends LspClientConnection {
 	private readonly openedFiles = new Set<string>();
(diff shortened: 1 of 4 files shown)


fufuiu:
> 感谢您的贡献！在合并此 PR 之前，我们需要您签署我们的[贡献者许可协议 (CLA)](https://github.com/code-yeongyu/oh-my-opencode/blob/master/CLA.md) 。
> 
> **要签署 CLA** ，请在此 PR 下方留言：
> 
> ```
> I have read the CLA Document and I hereby sign the CLA
> ```
> 
> 这是一次性要求。签署后，您未来的所有捐款将自动被接受。
> 
> 我已阅读 CLA 文件，并在此签署 CLA 文件。
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request. Posted by the **CLA Assistant Lite bot**.

MoerAI:
Thanks for the PR, @fufuiu, and for taking the time on this. I am sorry it sat without a maintainer response.

This PR targets `master`, but this repository accepts contributions only against `dev`; `.github/workflows/ci.yml:18-54` reserves `master` for release workflow updates. The branch also no longer applies cleanly to the current architecture: the cwd portion is already covered by #7477, while `packages/lsp-core/src/lsp/client.ts` now uses the diagnostics freshness contract, so the fixed 10 x 1-second polling patch is stale. The current checks also provide no validation signal because CLA failed and the substantive test, typecheck, and build jobs were cancelled.

I am closing this branch rather than asking you to repair unrelated history. The Windows `file:///d%3A/...` versus `file:///D:/...` finding is still useful. Please open a fresh issue against current `dev`; a focused replacement PR that normalizes URIs at the current diagnostics-store boundary and adds drive-letter plus percent-encoding regression tests would be welcome. Happy to review that replacement.

PR #6652, opened 2026-08-08
fix(mcp-client-core): propagate request timeouts to the MCP SDK (#6631)

### LINKED ISSUE #6631: [Bug]: Direct MCP callTool uses SDK default timeout because request options are not propagated
## Summary

The direct MCP tool-execution path calls the TypeScript SDK's `client.callTool(...)` without request options. The SDK then applies its default request timeout, so legitimate tool calls that run longer than one minute fail even when the underlying tool supports a longer wait.

## Minimal generic reproduction

1. Start any MCP server that exposes a test tool which waits for 65 seconds and then returns a success result.
2. Invoke that tool through the direct MCP client execution path.
3. Observe that the client times out at approximately 60 seconds before the tool returns.
4. Expected: the direct call path should propagate a configurable request timeout through the SDK `callTool` request options, or otherwise provide a documented timeout policy suitable for long-running tools.

## Scope

This concerns direct tool execution after the MCP connection is established. It is distinct from connection, initialization, and tool-discovery timeouts.

## Suggested direction

Pass the configured or documented per-tool/per-server timeout as the SDK request options for `client.callTool`, with a safe default and clear cancellation behavior.

imkp1:
## Summary

`SkillMcpManager` called the MCP SDK without request options, so every tier-3 skill-embedded MCP operation fell through to `DEFAULT_REQUEST_TIMEOUT_MSEC` (60000 ms). A tool that legitimately runs longer than a minute failed at about 60 seconds.

As noted on the issue, there was also nothing to propagate: no timeout field existed anywhere in the client config path. So this is two parts. It adds the config surface, then threads it into the SDK request options at every call site in that class, not just `callTool`.

## Changes

- `claude-code-compat-core` gains `McpRequestTimeoutConfig` and an optional `timeouts` block on `ClaudeCodeMcpServer`. One grouped key rather than three loose ones. Fields map straight onto the SDK: `requestTimeoutMs` to `timeout`, `resetTimeoutOnProgress` passthrough, `maxTotalTimeoutMs` to `maxTotalTimeout`.
- New `mcp-client-core/src/skill-mcp-manager/request-timeouts.ts` with `resolveMcpRequestOptions`. It merges the per-call override over the per-server config field by field, drops non-finite and non-positive values, and returns `undefined` when nothing resolves.
- `SkillMcpClientOptions` gains `timeouts`, so a caller can raise the limit for one call without restating the server config.
- `manager.ts` passes the resolved options to `listTools`, `listResources`, `listPrompts`, `callTool`, `readResource`, and `getPrompt`.

A skill's `mcp.json` opts in like this:

```json
{
  "mcpServers": {
    "slow-server": {
      "command": "node",
      "args": ["server.js"],
      "timeouts": {
        "requestTimeoutMs": 300000,
        "resetTimeoutOnProgress": true,
        "maxTotalTimeoutMs": 1800000
      }
    }
  }
}
```

Defaults are unchanged. An unconfigured server sends no request options at all, so the SDK keeps applying its own default. I did not pick a new default number.

## Verification

Live stdio MCP server with a `wait` tool, driven through the real `SkillMcpManager`, 65 second call:

```
[no-timeout-config]         FAILED after 60.1s -> McpError: MCP error -32001: Request timed out
[requestTimeoutMs=120000]   OK     after 65.1s -> [{"type":"text","text":"waited 65s"}]
```

The first line reproduces the reported defect against a real server. The second is the fix.

Negative control, with only the `manager.ts` threading reverted and everything else left in place:

```
6 pass / 2 fail
[requestTimeoutMs=120000]   FAILED after 60.1s
```

The configured timeout has no effect without the threading, so the tests and the driver depend on the change and not just on the new field existing.

- `bun run typecheck`: exit 0.
- New co-located tests: 8 pass / 0 fail.
- `mcp-client-core`, the `omo-opencode` skill-mcp-manager shim tests, `claude-code-mcp-loader`, and `skills-loader-core`: 399 pass / 0 fail.
- Full root `bun test`: 13303 pass / 28 fail, against a 13295 pass / 28 fail baseline on unmodified `dev`. The failing-test name sets are identical. Those 28 are local environment only, from installing with `--ignore-scripts` and without the frontend provenance submodules.

## QA & Evidence

`.omo/evidence/20260808-fix-6631/`, including the test server, the driver, the before and after output, the negative control, and `qa-summary.md`.

I did not run the OpenCode harness QA. The change is confined to a Core package: it adds an optional config field and forwards it, and the `packages/omo-opencode/src/features/skill-mcp-manager/` files are unchanged re-export shims. Say the word if a harness-level run is wanted here anyway and I will add it.

## Risks & Residuals

- `ClaudeCodeMcpServer` is shared with the tier-2 `.mcp.json` loader, where `timeouts` has no effect because those servers go to OpenCode directly. The field is optional and inert there, but it is a surface that could read as supported. Happy to move it to a tier-3-only type if you would rather keep the compat type clean.
- Per-tool overrides are available through `SkillMcpClientOptions`, but nothing in the `skill_mcp` tool exposes them to users yet. That felt like a separate change. The per-server config path works end to end today.
- Cancellation semantics are the SDK's own. On timeout the SDK raises `McpError` with code `RequestTimeout` and cancels the request, which the driver output confirms.

## Related Issues

Fix #6631

### DIFF
--- a/.omo/evidence/20260808-fix-6631/driver-after-fix.txt
+++ b/.omo/evidence/20260808-fix-6631/driver-after-fix.txt
@@ -1,0 +1,2 @@
+[no-timeout-config] FAILED after 60.1s -> McpError: MCP error -32001: Request timed out
+[requestTimeoutMs=120000] OK after 65.1s -> [{"type":"text","text":"waited 65s"}]
--- a/.omo/evidence/20260808-fix-6631/driver.ts
+++ b/.omo/evidence/20260808-fix-6631/driver.ts
@@ -1,0 +1,25 @@
+import type { ClaudeCodeMcpServer } from "@oh-my-opencode/claude-code-compat-core/claude-code-mcp-loader/types"
+import { SkillMcpManager } from "@oh-my-opencode/mcp-client-core/skill-mcp-manager/manager"
+
+const serverPath = new URL("./slow-server.ts", import.meta.url).pathname
+const waitSeconds = 65
+
+async function run(label: string, timeouts?: ClaudeCodeMcpServer["timeouts"]) {
+  const manager = new SkillMcpManager()
+  const config: ClaudeCodeMcpServer = { command: "bun", args: [serverPath], ...(timeouts ? { timeouts } : {}) }
+  const info = { sessionID: `ses-${label}`, skillName: "qa-skill", serverName: "slow-server" }
+  const context = { config, skillName: "qa-skill" }
+  const startedAt = Date.now()
+  try {
+    const result = await manager.callTool(info, context, "wait", { seconds: waitSeconds })
+    console.log(`[${label}] OK after ${((Date.now() - startedAt) / 1000).toFixed(1)}s ->`, JSON.stringify(result))
+  } catch (error) {
+    console.log(`[${label}] FAILED after ${((Date.now() - startedAt) / 1000).toFixed(1)}s ->`, String(error))
+  } finally {
+    await manager.disconnectAll()
+  }
+}
+
+await run("no-timeout-config")
+await run("requestTimeoutMs=120000", { requestTimeoutMs: 120_000 })
+process.exit(0)
--- a/.omo/evidence/20260808-fix-6631/full-suite.txt
+++ b/.omo/evidence/20260808-fix-6631/full-suite.txt
@@ -1,0 +1,15 @@
+Full root suite, unmodified dev (baseline):
+ 13295 pass
+ 28 fail
+Ran 13357 tests across 1725 files. [180.86s]
+
+Full root suite, with this change:
+ 13303 pass
+ 28 fail
+Ran 13365 tests across 1726 files. [203.85s]
+
+diff of the failing-test name sets (baseline vs change): empty
+The 28 failures are identical on both sides and are local-environment only:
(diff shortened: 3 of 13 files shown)


imkp1:

MoerAI:
Closing this stale duplicate so #6631 can be consolidated into one current successor. #7224 is the later overlapping candidate, but it is also blocked and must be refreshed:
https://github.com/code-yeongyu/oh-my-openagent/pull/7224

Do not carry `resetTimeoutOnProgress` unchanged. The MCP SDK associates progress only when `onprogress` is supplied, so forwarding the boolean alone cannot reset a timeout:
https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.29.0/src/shared/protocol.ts#L1135-L1144

Keep #6631 open. The canonical replacement must either omit that ineffective surface or add a bounded progress callback and prove token emission, reset behavior, maximum-total expiry, timeout cancellation/cleanup, and unchanged OAuth/reconnect retries:
https://github.com/code-yeongyu/oh-my-openagent/issues/6631#issuecomment-5213498314

It also needs current CI and isolated OpenCode `skill_mcp` QA loading real config/frontmatter against a real MCP server. Reopen only if this branch becomes the selected current successor.

PR #6696, opened 2026-08-10
fix(senpi-task): resolve installed CLI outside PATH

stevenahhh:
## Summary

- resolve the installed `@code-yeongyu/senpi/dist/cli.js` directly when Windows PATH contains no Senpi shim
- launch that package CLI with the current Node/Bun executable before falling back to loader-sensitive `rpc-entry` resolution
- add a regression for the real OMO environment: installed Senpi package, empty PATH, and a loader-alias failure

## Root cause

This is a residual case from #6449 / #6451. The Windows launcher normalization works when `senpi.exe`, `senpi.cmd`, or `senpi` is discoverable on PATH. OMO builds a curated PATH that can omit both the global npm bin and the package-level `.bin` directory. In that case `buildRpcSpawn()` fell through to `resolveRpcEntry()`. Inside the bundled OMO extension that resolution is loader-sensitive and threw synchronously, so `TaskManager` persisted only the security-safe `Task runner failed to start.` message.

Recent session evidence showed two native members failing within 20–30 ms with that generic start error, while direct `RpcProcessRunner.start()` succeeded after a local `senpi.exe` shim was added.

## Verification

- Red: new focused regression threw `Error: loader alias hijacked` from the `resolveRpcEntry()` fallback
- Green: `spawn.test.ts` — 22 passed
- Green: full `src/runners/rpc` directory — 55 passed, 3 Windows signal tests skipped
- LSP diagnostics — no errors in changed files
- Source end-to-end with PATH limited to Node/System32/Git:
  - descriptor entry resolved to `node_modules/.../@code-yeongyu/senpi/dist/cli.js`
  - real RPC child returned `SOURCE_PACKAGE_FALLBACK_OK_2`
- Live native `team_create` member returned `NATIVE_TEAM_OK`

## Existing repository failures

The complete `senpi-task` suite still has unrelated existing timeout failures in chaos/member-extension tests. Package typecheck is also blocked by the existing `packages/utils/src/runtime/file.ts` `ArrayBuffer | SharedArrayBuffer` mismatch. The focused launcher and complete RPC runner suites are green.

---
## Summary by cubic
Fixes Windows startup when PATH omits Senpi shims by resolving the installed `@code-yeongyu/senpi` package CLI and launching it with the current Node executable. Also refreshes generated `omo-senpi` plugin bundles.

- **Bug Fixes**
[...]

PR #6699, opened 2026-08-10
fix(omo-native): preserve explicit Senpi binary

stevenahhh:
## Summary

- preserve an existing explicit `SENPI_BIN` when the native OMO launcher constructs the Senpi child environment
- continue dropping stale and self-referential values, then fall back to the packaged shim
- isolate launcher fixtures from the developer machine's `SENPI_BIN`

## Root cause

`omo` copied the parent environment and then unconditionally deleted `SENPI_BIN`. On Windows installations where the curated child `PATH` does not expose a usable Senpi shim, native team process workers fail before creating a child session even when the user configured an explicit valid binary.

This is separate from #6696: that PR hardens the process-runner package CLI fallback, while this change fixes launcher policy at the `omo-native` environment boundary.

## Verification

- `bun test packages/omo-native/test/launcher.test.ts --timeout 15000` — 20 pass, 1 Windows-inapplicable signal test skipped, 0 fail
- `bun x biome check packages/omo-native/bin/lib/launcher.js packages/omo-native/test/launcher.test.ts`
- `node --check packages/omo-native/bin/lib/launcher.js`
- independent diff review: no findings

---
## Summary by cubic
Preserves an explicit `SENPI_BIN` when the `omo-native` launcher builds the Senpi environment, so a user-configured engine binary is no longer overwritten. Falls back to the packaged shim only when no explicit binary is available or the configured one points back at the launcher.

- Keep an existing `SENPI_BIN` if it resolves to a real file and does not self-reference the launcher (symlinks and case aliases are handled via realpath/inode identity).
- Apply the explicit binary after the packaged shim so it takes precedence.
- Isolate launcher tests from the developer machine's `SENPI_BIN`.

<sup>Written for commit 38495d778e104b3988d0f8d40852466c8a492840. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
Thanks for preserving an explicit `SENPI_BIN`; current `dev` still deletes that value, so the fix is not superseded.

The self-reference guard compares resolved path strings only. A symlink or Windows case alias pointing back to the OMO launcher can bypass it and recurse. Please compare canonical executable identity, reject non-file paths, and add tests for a symlinked self-reference plus Windows case-insensitive aliases.

This branch now conflicts with current `dev`, and its required Windows test job is red. Rebase, preserve the intervening launcher changes, rerun the complete current matrix, and add an isolated CLI smoke that proves an explicit alternate Senpi binary is used while every launcher self-reference is removed.

MoerAI (approved):
Approved on current head. The prior safety blockers are resolved:

- executable identity now canonicalizes real files and compares filesystem identity rather than raw path spelling;
- self-references through the launcher path, symlinks, and case aliases are rejected;
- nonexistent and non-file `SENPI_BIN` values are rejected;
- an explicit alternate binary still outranks the packaged shim;
- the spawned launcher fixture proves the alternate engine reaches the child environment without a surviving launcher self-reference.

Cubic, CLA, and GitGuardian are green, and the branch is mergeable.

Merge remains deferred because the required CI workflow is `action_required` and produced no test, typecheck, build, Windows, or compatibility jobs on this head:
https://github.com/code-yeongyu/oh-my-openagent/actions/runs/33463206723

Do not merge until that workflow runs and every required context is green.

PR #6704, opened 2026-08-10
feat(memory): memory v2 active learning

code-yeongyu:
## What and why

Upgrades **omo-senpi memory** from a passive block-injector into an actively learning system.

Today memory only reflects when a compaction happens, and its step trigger shipped disabled
(`step_count: 0`). This PR turns active learning on by default and closes the gaps a comparative
study of openclaw, hermes-agent, oh-my-pi and honcho surfaced.

- **Active learning on by default** - the reflection step trigger returns to the letta default of
  25 (our `0` was an unintended divergence); nudge / facts / dream / people / soul all default on
  with individual kill switches and per-agent overrides.
- **SOUL elevation** - persona seed rewritten in an identity-forward register, optional
  `system/identity.md` folded into `<self>`, and soul edits are never silent.
- **Instruction- and skill-mediated learning** - a sharpened always-on reminder, a self-clearing
  turn nudge, and a seeded `memory-discipline` skill carrying the long-form guidance.
- **Dreaming** - a letta-shaped `/dream` consolidation pass with a hermes-inspired skill-usage
  audit, triggered opportunistically on idle and at shutdown.
- **Shutdown drain, async fact extraction, sleep consolidation** - a bounded drain, a durable facts
  queue with a quick-pinned extractor, and consolidation folded into the dream pass.
- **People cards and observations** - Honcho-style cards with a four-level observation ledger and a
  derived relationship graph surfaced in the memory palace.

## Shape

25 implementation todos plus an F1-F4 final verification wave, in six topological waves. Every
cross-cutting decision lives in binding interface contracts **IC-1..IC-17** so no executor has to
invent behavior. Each todo lands as an atomic commit with a RED-first proof and evidence under
`.omo/evidence/omo-senpi-adapter/memory-v2/task-N/`.

## Review

33 adversarial rounds across five independent lanes (full-scope contract, executor simulation,
contract audit, test-discipline audit, integrity audit). Every finding was integrated; none was
declined. The executor-simulation lane returned `EXECUTABLE` for all 25 todos with zero stalls, and
the final full-scope lane returned an unconditional `APPROVED` with zero blockers against the
committed plan sha. The full ledger is in `.omo/drafts/memory-v2-active-learning.md`.

Defects review caught, which the plan now fixes by construction:

- The **gated child bootstrap handshake**, because `spawn()` returns a live child before its
  identity can be made durable.
- **Probe-driven finalization** via an `Omo-Run` receipt trailer, because a marker written before or
  after an irreversible step cannot distinguish "not started" from "finished but unrecorded".
- **Out-of-band commit receipts**, because senpi's MCP output guard replaces oversized tool text and
  would silently eat an inline trailer on exactly the largest edits.
- **Windows supervisor containment**, because `getProcessStartIdentity` has no win32 branch while CI
  runs on `windows-latest`. Tree termination is bound to `taskkill /T /F`, and the **bootstrap arms
  its own absolute-deadline timers** so a child self-terminates even if its supervisor is killed.
- **Non-destructive recovery.** An earlier draft mandated `git reset --hard`. Memory is user data,
  and this repo refuses rather than clobbers (`parent_dirty`, `DirtyRepoError`), so recovery is now
  ownership-proven: restore only residue whose blob sha matches what the run wrote, otherwise leave
  the tree untouched and keep the batch retryable.
- **Facts watermark ordering**, found empirically when an executor's first implementation failed:
  ordering by journal position is undefined across batches, so a late batch rolled the watermark
  backward. The cursor now persists `end_snapshot_line`, which is comparable across batches.

## Also fixed here

- `docs/reference/configuration.md` had **no memory section**; every key is now documented, with all
  defaults verified against the Zod schema rather than transcribed.
- The pre-commit frontmatter hook rejected `kind`/`aliases`, so **every fresh memory repo failed its
  own validation hook** on init.

## Test discipline

Compile goldens are prose-agnostic: the reminder paragraph normalizes to a `<reminder/>` placeholder
so wording changes cannot break them. Per `.omo/rules/test-discipline.md`, pure-prose changes ship on
review plus QA-by-read rather than a manufactured phrase-pin.

## Note on two commit subjects

The git index is shared across the parallel executors that built this branch, and a bare `git commit`
takes everything staged. Twice that swept another agent's files into a neighbouring commit: the run
supervisor landed inside `3347fd4cc` (a hook fix) leaving `e742edfee` as an empty marker, and the
persona v2 seed landed inside `224638c56`. **No content was lost** and the tree is correct; only two
subjects under-describe their contents. History was already pushed and agents were still committing,
so rewriting it was the riskier option. Later agents were switched to explicit-pathspec commits.

## Status

Implementation in progress; the PR stays in draft until all 25 todos plus the F1-F4 wave are complete
and green.

Plan: `.omo/plans/memory-v2-active-learning.md`
Review ledger: `.omo/drafts/memory-v2-active-learning.md`

---

## Implementation status

The branch now carries the implementation, not only the plan: 139 commits ahead of `dev`, merged up to
current `dev` so the committed Senpi bundle stays current alongside the new init-deep advisor entry.

### Concurrency hardening after adversarial review

Two independent reviewer lanes ran repeatedly against the real code and the shipped bundle, not against
a summary, until each returned an unconditional PASS. Every finding below was a real data-loss or
liveness path, proven by a reviewer reproduction and fixed RED-first:

- **Run terminalization** (`a8d5f895b`). Publication versus abandonment is decided by a single
  filesystem-exclusive claim rather than reads followed by a write, because no finite sequence of reads
  could settle it. A finished child outranks abandonment, stale attempts are ignored, checkpointed
  non-merged decisions replay from the durable ledger, legacy runs without recoverable identity fail
  closed, claims publish atomically through a candidate plus exclusive link, and claim recovery is
  serialized so a losing contender can never delete a newer winner. A genuinely dead run with unknown
  liveness still reaches `abandoned.json`, so the documented win32 path is intact.
- **Facts crash recovery** (`74ad4bc69`). Recovery is ownership-proven end to end: planning captures
  pre-state before validating the repository clean, every recorded identity is durable in the object
  database before any mutation, and each index and worktree write is a conditional primitive that
  validates the expected identity at the write itself (Git `index.lock` for staged state, exclusive
  link for new paths, move-away-then-verify for existing ones). Deletion reserves the emptied path with
  a directory so no writable inode is ever exposed, removal is bound to a privately re-verified inode,
  and restoration never overwrites. Anything else is reported as `parent_dirty` with the foreign bytes
  intact and the retained path named in `final.json`.

### Test determinism

138 process-heavy memory tests across 28 files relied on the default five second wall clock; one
supervised launch failed at 5002 ms under full parallel load while passing in 770 ms isolated. They now
carry the explicit deadline the memory suites already used, with no assertion weakened.

## QA at the merged head

- `packages/memory-core` 481 pass / 0 fail, `packages/omo-senpi` 1282 pass / 0 fail, both typechecks.
- The committed-bundle freshness verifier CI runs (`build-extension.mjs --check`) reports the tree current.
- Real Senpi surface, model fallback: the unreachable primary is abandoned and the run completes on the
  reachable fallback, ending `merged`.
- Real Senpi surface, skill startup: no missing-skill-path warning and no `frontend` skill collision.
- Both live runs used their own `mktemp` roots and the host memory digest was unchanged, so
  `~/.omo/memory` was never touched; the roots were removed afterwards.
- Receipts: `.omo/evidence/20260811-memory-reflection-model-fallback/merged-head-qa.md`; the full
  reviewer ledger, including every round and its resolution, is in
  `.omo/drafts/memory-v2-active-learning.md`.


code-yeongyu:
## Status update - implementation complete, remediation landed

All 25 todos are implemented. The F1-F4 audit wave then ran and found real defects; every one is fixed
and re-verified below.

**Audit findings and resolutions**

| Gate | First run | Fix commits |
|---|---|---|
| F1 plan compliance | FAIL (5 rows) | `aa7ef94e4` runtime gating + schema-default resolvers + drain abort threading; `47f7c863f` MH-4 single-carrier + MN-6 contradiction preservation |
| F2 code quality | FAIL (4 classes) | `aa7ef94e4` empty catches now propagate non-ENOENT; `2e61de173` test-discipline rework (every prose pin replaced with behavioral assertions, each proven by mutation check); LOC splits `0baef0c1f`, `7e4e0c505`, `feb6faa3a` (largest remaining unit 198 pure LOC) |
| F3 real QA | FAIL | Root-caused to concurrent-edit churn producing a stale-bundle read; a clean rebuild embeds current source (verified: new reminder text present). Re-run in progress on the settled tree. |
| F4 scope fidelity | FAIL (metadata) | Sync merges `43e47c610` (beta.4 pin), `8a705227e` (v5.0.0-beta.5); constants and all disclosures verified truthful |

**Sync merges on this branch.** Two dev merges kept the branch current during the long fan-out:
the senpi-pin sync and the v5.0.0-beta.5 release sync. Neither introduces memory-v2 surface.

**Bundle.** Shipped artifact rebuilt from final source: 867,366 -> 974,066 bytes of plan-scoped
feature code, purity guard green (no new third-party dependency), budget raised to 1,000,000 with
documented justification in `bundle-size.test.ts` matching its own raise history.

**Full suites:** memory-core 418/418, omo-senpi 1078 + one documented pre-existing failure
(ulw-loop `OMO_BIN`, reproduced on pristine origin/dev).

code-yeongyu:
## Merge note: dev sync with semantic conflict resolution

Synced latest `dev` (post-beta.5) via merge commit. Six files conflicted; each was resolved by intent, not by picking a side:

| File | Both intents | Resolution |
|---|---|---|
| `wiring.ts` | My split composition (wiring-static/runtime/context/types) + dev model-registry extraction | Kept my structure; delegated `resolveModelRegistry` to devs shared `./model-registry-resolver` module, which replaces my `as unknown` cast with a proper type guard |
| `cache.test.ts` | My bounded-replacement cache size (2) + devs Windows-timeout wrapper | Both: `expect(cache.size).toBe(2)` inside `WINDOWS_INTEGRATION_TEST_TIMEOUT` (5/5 green) |
| `memory.test.ts` | Dev refactored read_only tests that my S1 split legitimately MOVED to `memory-validation.test.ts` (0 copies in my file, not deleted) | Kept my S1 split; ported devs fixture/seed hoists into the validation sibling where the tests now live |
| `compaction-survival.test.ts` (devs new file) | Their narrowed `createRuntime` type vs my `wiring.ts:155` `runtimeFor(identity).reconcile()` call, which REQUIRES the full runtime type | Kept my full type (narrowing is structurally unsafe in my tree); completed their stub with a no-op `reconcile` following my existing wiring.test.ts double pattern (4/4 green, proves both worlds coexist) |
| `plugin/extensions/*` (3 bundles) | Generated artifacts | Rebuilt once from merged source at the end of the wave (packages/omo-senpi builds regenerate these by design) |

Typecheck clean on both packages. Full suites pass: compaction-survival 4/4, tools 45/45.

code-yeongyu:
## Update: latent deadline time-bomb found and fixed; suite fully green

Post-merge verification surfaced seven facts-launch failures. Root cause analysis (with a
detached-worktree pre-merge baseline proving the failures predated the merge):

**The absolute deadline was anchored to the parents logical clock and enforced against the
supervisor process wall clock.** Test fixtures inject a frozen `now` (`2026-08-10T12:00:00Z`);
the suite passed this morning and turned permanently red at 12:00:10Z, when wall time ran past
`frozen launch + deadlineMs`. Every child was SIGTERMD at spawn.

DB528ec10 fixes both child paths (reflection and facts) to compute `hardDeadlineAt` from the
enforcing clock at launch, which is what the plans `ONE ABSOLUTE hardDeadlineAt` contract
always required - the writer of the deadline and its enforcer must share one clock, and only
the supervisors wall clock is real in a separate process.

All seven previously-failing facts suites pass. Full `bun test packages/memory-core
packages/omo-senpi`: **1505 pass / 0 fail / 6775 assertions across 230 files** - the first
fully-green run of this branch, including the previously-documented pre-existing `OMO_BIN`
failure, which turned out to be machine-clock-sensitive too and now passes consistently.

PR #6716, opened 2026-08-10
fix(bg-agent): bound tool-wait defer and retain reply-required wakes (#6546 D1+D2)

### LINKED ISSUE #6546: [Bug]: Sibling background tasks completing while parent is streaming lose the allComplete wake — parent parks forever (recurrence of #1582, still unfixed in 4.19.3)

EvangelosMoschou:
## Summary

Follow-up to #6655 (fixes #6546). The maintainer confirmed two additional defects in the parent-wake flush machinery on top of the original report:

1. **D1 — stale tool-call deferral has no time ceiling.** The `toolWaitDecision.defer` branch sends a `forceNoReply` admit and retains the wake, but unlike the active-session path it has no `PENDING_PARENT_WAKE_MAX_ACTIVE_DEFER_MS` (60s) bound. A reply-required wake can be re-admitted as noReply forever while parent history keeps reporting a blocking/stale tool call — observed in edynasty's reproduction as `Holding parent wake during stale tool-call deferral` repeating every 1s for 21s+.

2. **D2 — `dropAdmittedWakeConsumedByParent` destroys reply-required wakes.** When the parent produces any assistant output after the noReply admission, the retained `shouldReply=true` wake is deleted. That output can be unrelated parent activity (timeout-based recovery, another task's result), so destroying the wake removes the last liveness signal that a reply is owed — the parent then re-dispatches duplicate work (observed twice in edynasty's repro).

## Changes

- **`parent-wake-flush-runner.ts`**
  - D1: in the `toolWaitDecision.defer` branch, a reply-required wake past the 60s age ceiling is force-dispatched (mirroring the active-session ceiling) instead of being re-admitted as noReply forever.
  - D2: `dropAdmittedWakeConsumedByParent` returns early when `latestWake.shouldReply === true` — reply-required wakes are never dropped by later parent output; only non-reply wakes can be consumed. The duplicate-reply hazard (forking a concurrent assistant chain) is still prevented because the admitted wake's `noReplyAdmittedAt` keeps `deferReplyWakeWhileUnsafe` deferring re-dispatch.

## QA & Evidence

Evidence dir: `.omo/evidence/20260809-6546-d1-d2-wake-defer-bounds/EVIDENCE.md`

- **TDD red phase** (fix stashed): both new tests fail — D2 test shows the reply-required wake dropped; D1 test shows noReply admission instead of force-dispatch.
- **TDD green phase**: `parent-wake-reply-retention.test.ts` — 2 pass / 0 fail.
- **Full background-agent suite**: **748 pass / 0 fail** (60 files). The two pre-existing tests that pinned the old behavior were updated to assert the new intended behavior; all other parent-wake tests (active-defer-ceiling, noreply-liveness, history-deferral, final-notification-merge, same-source-requeue, empty-turn-requeue, late-error-recovery) pass unchanged.
- **Duplicate-dispatch hazard still prevented**: traced the noreply-liveness consumed-admission scenario — with the D2 fix, the second flush produces no additional `promptAsync` call (calls stayed at 1); only the retention outcome changed.
- `bun run typecheck` clean; `bun run build` exit 0.

## Risks & Residuals

- The force-dispatch-past-tool-wait-ceiling path is exercised at unit level only; no live-harness run was executed (machine lacks the opencode-qa script deps). The dispatch goes through the same `sendParentWakePrompt` -> `dispatchInternalPrompt` gate as all other wakes, so the prompt-async gate is preserved.
- Behavior change is confined to `parent-wake-flush-runner.ts` plus its tests.

## Related Issues

- Fixes the two confirmed follow-up defects on #6546 (confirmed by code-yeongyu 2026-08-06)
- Related: #6655 (the reconciliation fix), #1582 (original recurrence)

---
## Summary by cubic
Fix two defects in the background agent’s parent-wake flow: cap tool-wait deferral at 60s and retain reply-required wakes after noReply admission. Prevents infinite noReply loops and lost replies; addresses #6546.

- **Bug Fixes**
  - D1: If a reply-required wake is blocked by a stale tool call and older than 60s, force-dispatch the reply instead of re-admitting as noReply.
  - D2: Never drop reply-required wakes in `dropAdmittedWakeConsumedByParent`; only non-reply wakes can be consumed by later parent output.
  - Tests: Added `parent-wake-reply-retention.test.ts`; updated active-defer-ceiling and noreply-liveness tests; raised timeouts in slow team-mode live-delivery messaging tests to reduce flakiness.

<sup>Written for commit af9462359ed68762f2e91ed48808b0a9e6684a6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6716?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-active-defer-ceiling.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-active-defer-ceiling.test.ts
@@ -247,7 +247,7 @@ describe("parent wake active defer ceiling", () => {
     }
   })
 
-  test("#given blocked tool history and aged wake while parent is busy #then it admits noReply", async () => {
+  test("#given blocked tool history and aged wake while parent is busy #then it force-dispatches the reply past the tool-wait ceiling", async () => {
     // given
     const { notifier, promptAsyncCalls } = createNotifier({
       sessionStatuses: { "parent-1": { type: "busy" } },
@@ -259,11 +259,11 @@ describe("parent wake active defer ceiling", () => {
       // when
       await notifier.flushPendingParentWake("parent-1")
 
-      // then
+      // then — the aged reply-required wake is force-dispatched instead of
+      // being admitted as noReply forever (issue #6546 D1)
       expect(promptAsyncCalls).toHaveLength(1)
-      expect(promptAsyncCalls[0]?.body.noReply).toBe(true)
-      expect(notifier.getPendingParentWakes().get("parent-1")?.shouldReply).toBe(true)
-      expect(notifier.getPendingParentWakeTimers().has("parent-1")).toBe(true)
+      expect(promptAsyncCalls[0]?.body.noReply).not.toBe(true)
+      expect(notifier.getPendingParentWakes().has("parent-1")).toBe(false)
     } finally {
       notifier.shutdown()
     }
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts
@@ -81,6 +81,22 @@ export class ParentWakeFlushRunner {
       if (this.deferReplyWakeWhileUnsafe(sessionID, latestWake)) {
         return
       }
+      // The tool-wait deferral path has no independent ceiling: a reply-required
+      // wake can be re-admitted as noReply forever while the parent history keeps
+      // reporting a stale tool call (issue #6546 D1). Apply the same age ceiling
+      // as the active-session path so the reply is eventually force-dispatched.
+      if (this.shouldForceDispatchAfterActiveDefer(latestWake)) {
+        await this.sendParentWakePrompt(sessionID, latestWake, {
+          emptyAssistantTurnRetry,
+          toolWaitDecision: { ...toolWaitDecision, skipPromptGateToolStateCheck: true },
+          skipPromptGateStatusCheck: true,
+        })
+        log("[background-agent] Sent parent wake after tool-wait defer ceiling:", {
+          sessionID,
+          queuedAgeMs: this.getQueuedAgeMs(latestWake),
+        })
+        return
+      }
       await this.sendParentWakePrompt(sessionID, latestWake, {
         emptyAssistantTurnRetry,
         toolWaitDecision: { ...toolWaitDecision, skipPromptGateToolStateCheck: true },
@@ -185,12 +201,17 @@ export class ParentWakeFlushRunner {
     this.deps.pendingQueue.clearTimer(sessionID)
   }
 
-  // A retained reply-required wake is only liveness insurance for a deposit the
-  // parent never saw. Assistant output created after the noReply admission means
-  // the live turn consumed the deposit — re-dispatching it would inject a
-  // duplicate notification and fork a concurrent assistant chain.
+  // A retained reply-required wake is liveness insurance for a deposit the
+  // parent never saw. Assistant output created after the noReply admission can
+  // be unrelated parent activity (timeout recovery, another task's output), so
+  // a reply-required wake must never be dropped here: the reply is still owed.
+  // Only non-reply wakes (noReplyAdmittedAt set, shouldReply false) can be
+  // consumed by later parent output.
   private async dropAdmittedWakeConsumedByParent(sessionID: string, latestWake: PendingParentWake): Promise<boolean> {
     if (latestWake.noReplyAdmittedAt === undefined) {
+      return false
+    }
+    if (latestWake.shouldReply) {
       return false
     }
     if (!(await this.deps.sessionInspector.hasAssistantOutputAfterAdmittedWake(sessionID, latestWake))) {
--- a/packages/omo-opencode/src/features/background-agent/parent-wake-noreply-liveness.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/parent-wake-noreply-liveness.test.ts
@@ -366,11 +366,11 @@ describe("parent wake admitted-consumption drop (duplicate ALL-COMPLETE regressi
       notifier.clearPendingParentWakeTimer("parent-1")
       await notifier.flushPendingParentWake("parent-1")
 
-      // then: the retained wake is dropped instead of re-dispatched as a
-      // reply prompt (which forked a concurrent assistant chain in production)
+      // then: no duplicate reply prompt is forked (the admitted wake was
+      // consumed by the live turn), and the reply-required wake is retained
+      // rather than destroyed (issue #6546 D2)
       expect(promptAsyncCalls).toHaveLength(1)
-      expect(notifier.getPendingParentWakes().has("parent-1")).toBe(false)
-      expect(notifier.getDispatchedParentWakes().has("parent-1")).toBe(false)
+      expect(notifier.getPendingParentWakes().get("parent-1")?.shouldReply).toBe(true)
     } finally {
       Date.now = originalDateNow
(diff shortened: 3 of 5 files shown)


EvangelosMoschou:
Closing to prune the queue. This PR predates the current dev line by ~2,000 commits and has been open since early August, which at this point makes review unlikely. The branch is preserved and the PR can be reopened with a rebase if the change is still wanted.. Note: if the #6655 re-review favors the wake-retention approach from here, this will be reopened.

PR #6758, opened 2026-08-11
fix(omo-senpi): port stop-continuation-guard from omo-opencode

### LINKED ISSUE #6752: [Bug]: omo-senpi start-work continuation ignores paused and explicit stop state
## Bug description

The Senpi start-work continuation component does not durably honor pause/stop intent.

Two concrete source-level problems remain on current `origin/dev` (`f6e4f2d70`):

1. `findContinuableBoulderWork()` treats both `active` and `paused` work as continuable.
2. Any user input resets `consecutiveContinuations` and `lastSignature`, so a stop/end message can re-arm the continuation cycle instead of suppressing it.

The OpenCode adapter already has `stop-continuation-guard`, but the Senpi adapter has no equivalent component.

### Paused boulder

1. Create active start-work boulder state for a Senpi session.
2. Change the work status to `paused`.
3. Trigger `agent_end`.
4. Observe that a start-work continuation is still injected.

### User stop

1. Start an active start-work session.
2. Send a user steering message requesting the workflow to stop.
3. Let the current turn finish.
4. Observe that the user input reset the continuation counter/signature and a later `agent_end` can inject continuation again.

## Expected behavior

- `paused` work is not continuable.
- Explicit stop state persists across user messages and process turns.
- Start-work and ulw-loop continuation injections are suppressed while stopped.
- Running and pending descendant tasks are cancelled when continuation is explicitly stopped.
- Continuation resumes only through an explicit start/resume action.

## Actual behavior

Paused work remains eligible, and ordinary user input resets the continuation guard state.

## Proposed direction

1. Exclude `paused` in:
   - `packages/omo-senpi/src/components/start-work-continuation/boulder-eligibility.ts`
2. Port the OpenCode `stop-continuation-guard` contract to `omo-senpi`.
3. Persist the stopped state using the existing run-continuation state model or a harness-neutral equivalent.
4. Add `/stop-continuation` and `/resume-continuation` Senpi commands.
5. Make `IdleInjectionCoordinator` drop queued boulder/ulw continuation items while preserving terminal task completion messages.

## Acceptance tests

- Paused boulder work returns `null` from `findContinuableBoulderWork()`.
- Stop state suppresses repeated `agent_end` continuations.
- A normal user message does not clear stop state.
- Explicit resume clears stop state.
- Stop drops queued continuation injection but retains task-completion delivery.
- Descendant running/pending tasks are cancelled exactly once.

## Related precedent

OpenCode issue #3276 reported that `/stop-continuation` did not stop continuation and was closed as completed. This issue requests equivalent behavior for the Senpi adapter.

## Version

- Audited branch: current `origin/dev`
- Audited commit: `f6e4f2d70`

Made with [Orca](https://github.com/stablyai/orca) 🐋

code-yeongyu:
Closes #6752

## Problem

The Senpi `start-work-continuation` component ignored paused/explicit-stop state on two fronts on `dev`:

1. `findContinuableBoulderWork()` treated both `active` and `paused` work as continuable, so pausing a boulder never suppressed the `agent_end` continuation injection.
2. Any user input reset `consecutiveContinuations` and `lastSignature`, so a stop/end steering message re-armed the continuation cycle instead of durably silencing it.

The OpenCode adapter already had the reference implementation (`packages/omo-opencode/src/hooks/stop-continuation-guard/hook.ts`); the Senpi adapter had no equivalent.

## Change

Port the OpenCode `stop-continuation-guard` semantics into the Senpi adapter without reinventing the state model.

- **`boulder-eligibility.ts`**: only `active` status is continuable. Paused, completed, and abandoned all return `null`.
- **New `stop-continuation-guard.ts`**: in-memory `Set<sessionId>` with `stop` / `isStopped` / `clear`. Shared between `start-work-continuation` and `ulw-loop` via a per-pi `WeakMap` so both components observe the same stopped set without a global module singleton (isolated unit tests using independent `FakeExtensionAPI` instances stay independent).
- **`start-work-continuation`**:
  - Registers `/stop-continuation` and `/resume-continuation` slash commands.
  - In the `input` handler: does NOT reset the counter/signature when the current session is stopped. This is the fix for the second bug — the pre-fix reset re-armed continuation on any user message.
  - In `agent_end`: early-return when stopped.
  - On `session_shutdown`: `clear(sessionId)` so a resumed session id never inherits a stale stop marker.
- **`ulw-loop`**: reads the same guard and skips its own `agent_end` continuation when the session is stopped.

The Senpi guard is in-memory (not a filesystem marker like OpenCode's) because the Senpi extension is a single-process TypeScript module. Cross-process durability isn't required.

## Reproductions covered

Both minimal reproductions in the issue are now covered by tests.

- **Paused boulder** (`boulder-eligibility.test.ts`, `index.test.ts`): setting `status: "paused"` and firing `agent_end` produces no injection.
- **Explicit stop** (`index.test.ts`, `stop-continuation-guard.test.ts`):
  - `/stop-continuation` suppresses the next `agent_end` injection.
  - A normal idle user message + a queued/steered user message BOTH keep the stop in place (queued input also does not receive the `<omo-senpi-start-work>` transform).
  - `/resume-continuation` re-arms.
  - `session_shutdown` clears.

Plus:

- Unit tests for the standalone guard and the per-pi sharing contract.
- Cross-component test in `ulw-loop/index.test.ts`: `/stop-continuation` registered by `start-work-continuation` also silences `ulw-loop`'s own `agent_end` injection when both are registered against the same `pi`.

## QA

```
bun run test:senpi
  1142 pass
  0 fail
  3252 expect() calls
  Ran 1142 tests across 167 files. [59.83s]
```

The full Senpi package gate (`build:senpi-plugin` + `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` + `bun test packages/omo-senpi`) passes.

The plugin bundle grows from 899,251 to 900,869 bytes after minification, so `bundle-size.test.ts` gets a 900_000 -> 910_000 bump with the same first-party-correctness reasoning the file has recorded for prior plan-scoped bumps. `bundle-purity.test.ts` still passes.

## Notes on scope

- **Not implemented in this PR** (from the issue's "Proposed direction"): `IdleInjectionCoordinator` doesn't drop queued items on stop, and no descendant task cancellation is wired up. The Senpi adapter has no built-in equivalent to OpenCode's `BackgroundManager` — the OpenCode guard's `backgroundManager` is itself optional. Both are worth follow-ups but are additive; they don't affect the two source-level problems the issue calls out.
- The current `IdleInjectionCoordinator` semantics still hold: since a stopped session never enqueues a boulder/ulw continuation, no queued item survives past the guard. Task-completion injections are unaffected (as required by the issue's "preserving terminal task completion messages" line).

---
## Summary by cubic
Ports the OpenCode stop-continuation guard into `omo-senpi` so paused work and explicit user stops actually suppress `agent_end` continuation. Aligns `start-work-continuation` and `ulw-loop` on a shared, per-PI guard to prevent unwanted injections. Fixes #6752.

- **Bug Fixes**
  - Only active boulders are continuable; paused/completed/abandoned return null (repro 1).
  - Added an in-memory stop-continuation guard shared via a per-PI WeakMap; registered /stop-continuation and /resume-continuation.
  - `start-work-continuation`: no counter/signature reset while stopped, early-return on `agent_end`, and guard cleared on `session_shutdown` (repro 2).
  - `ulw-loop`: reads the same guard and skips its own `agent_end` continuation when stopped.
  - Tests cover both reproductions; raised `bundle-size.test.ts` budget to 910_000 with all checks passing.

<sup>Written for commit 591a6a2dbc63219c89ac2f2b9cc7d6790d527ad4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6758?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Closing as stale, not as resolved.

#7280 renamed and removed this PR's `start-work-continuation` integration surface, so this conflicting head cannot be safely rebased mechanically:
https://github.com/code-yeongyu/oh-my-openagent/pull/7280

Keep #6752 open and implement it against current `ulw-execute-continuation` and `ulw-loop`:
https://github.com/code-yeongyu/oh-my-openagent/issues/6752

A replacement must ensure only active boulders continue; cover both ULW input and idle paths; remove queued continuation items without dropping terminal task delivery; cancel running/pending descendants exactly once; preserve stop state across required lifecycles; and keep command ownership available when continuation components are disabled.

Reopen only after porting to the current surfaces with deterministic command, queue, cancellation, disabled-component, and real Senpi stop/resume evidence.

PR #6761, opened 2026-08-11
fix: preserve array-form plugin entries and symlinks when self-healing tui.json

luchev:
## Problem

The oh-my-openagent server plugin rewrites `tui.json` at every opencode startup via `ensureTuiPluginEntry()`. Two bugs make that rewrite destructive for users who manage their config with dotfiles/symlinks:

1. **Array-form plugin entries are dropped.** `pluginEntries()` filters `config.plugin` to `typeof entry === "string"` only. Entries like `["opencode-copy-message", { editor: "nvim", keybinds: {...} }]` silently disappear from `tui.json` on every self-heal.

2. **The dotbot/managed symlink is replaced by a regular file.** `writeFileAtomically()` writes to a `.tmp` file and `renameSync`s it over the target path. When `tui.json` is a symlink (dotbot `~/.config/opencode/tui.json -> ~/.dotfiles/config/opencode/tui.json`), the rename replaces the symlink itself with a plain file — breaking the "dotfiles are the source of truth" setup.

## Changes

- `add-tui-plugin-to-tui-config.ts`: `pluginEntries()` now returns all entries (strings + `[name, options]` tuples) instead of string-only; `ConfigShape.plugin` type widened to `Array<string | [string, unknown]>`.
- `write-file-atomically.ts`: resolve `realpathSync` before the atomic rename so writes go through symlinks to their target, preserving the link.
- Tests: array-form entry preservation; symlink write-through (link survives, target updated, no `.tmp` left behind).

## Verification

`bun test` on both affected test files: 14 pass / 0 fail. `tsgo --noEmit` on `packages/omo-opencode`: clean.

---
## Summary by cubic
Fix self-healing of `tui.json` to preserve array-form plugin entries and write through symlinks. Prevents losing plugin options and replacing symlinks with regular files in dotfile-managed setups.

- **Bug Fixes**
  - `pluginEntries` now keeps both string and `[name, options]` entries; `ConfigShape.plugin` widened to `Array<string | [string, unknown]>`.
  - `writeFileAtomically` resolves the real target path before rename, updating the symlink target and preserving the link.

<sup>Written for commit 78fd53b870a4e56fdc0d004b5181ba5b26c8bace. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6761?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/cli/config-manager/add-tui-plugin-to-tui-config.test.ts
+++ b/packages/omo-opencode/src/cli/config-manager/add-tui-plugin-to-tui-config.test.ts
@@ -66,6 +66,21 @@ describe("ensureTuiPluginEntry", () => {
     expect(readTuiPlugins(dir)).toEqual([`${PLUGIN_NAME}@4.9.2`])
   })
 
+  it("#given array-form plugin entries #when ensuring TUI config #then it preserves them", () => {
+    // given
+    const dir = tempConfigDir()
+    const arrayEntry = ["opencode-copy-message", { editor: "nvim" }]
+    writeConfig(dir, "opencode.json", { plugin: [PLUGIN_NAME] })
+    writeConfig(dir, "tui.json", { plugin: [arrayEntry, "some-other/tui"] })
+
+    // when
+    const result = ensureTuiPluginEntry({ configDir: dir })
+
+    // then
+    expect(result).toEqual({ changed: true, reason: "added" })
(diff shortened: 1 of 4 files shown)


luchev:

PR #6763, opened 2026-08-11
fix: preserve array-form plugin entries and symlinks when healing tui

luchev:
## Summary

Make omo write the tui config following symlinks and preserve other plugin configs.

## Changes

* `add-tui-plugin-to-tui-config.ts` was assuming all plugins are strings, and dropping extra plugin configs
* `write-file-atomically.ts` was not respecting symlinks

## QA & Evidence

I have dotfiles that rely on symlinks and my plugins were constantly being reset. The rootcause was that omo was replacing my symlink with a file. Second, it was dropping plugins from my configs.

This fix allow users to continue using their symlinked dotfiles or more complex plugins.

## Risks & Residuals

N/A as far as I'm aware

## Screenshots

N/A

## Automated Checks

N/A

## Related Issues

---
## Summary by cubic
Preserves array-form plugin entries in `tui.json` and writes through symlinks when self-healing TUI config. Stabilizes Windows by tolerating transient lock EPERM and raising reflection test timeouts to 30s; previously we dropped non-string plugin entries, replaced symlinked dotfiles, and Windows EPERM or 5s timeouts caused flaky failures.

**TUI config self-heal**
- Retains both string and `[name, options]` plugin entries and appends the required TUI plugin if missing.
- `write-file-atomically` resolves symlinks and atomically updates the target, preserving the link.

**Windows CI stability**
- `memory-core` lock `readOwner` returns an empty snapshot on transient EPERM instead of throwing.
- Reflection tests now use 30s timeouts; two-process test canonicalizes temp paths and retries `rm` to avoid hangs.

<sup>Written for commit 2f839502b231932a709b982933b289d6e43d5c35. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luchev:

MoerAI (changes requested):
Thanks for preserving tuple entries and existing symlinks. This branch is conflicting and far behind current `dev`:
https://github.com/code-yeongyu/oh-my-openagent/compare/2f839502b231932a709b982933b289d6e43d5c35...b5cbae3fb8778a06f70aa7aada35c8be72f0dba0

Recreate the focused OpenCode change on current `dev` and drop the unrelated memory timeout hunks already covered by owner Windows work.

The rebuilt patch must:

1. preserve JSONC comments, trailing commas, and layout rather than parsing then serializing the entire file;
2. preserve dangling `tui.json` symlinks by writing their target, or reject them without replacing the link;
3. use the current shared `PluginEntry` model rather than restoring a local tuple type;
4. cover mixed tuple entries, comments/layout, existing and dangling relative symlinks, and real installer/startup invocation.

Attach isolated real OpenCode evidence plus a tmux TUI smoke, then obtain the current required-check matrix:

PR #6790, opened 2026-08-12
fix(memory): preflight reflection child model visibility

code-yeongyu:
## What changed

- Probe the resolved Senpi launcher before reflection, dream, or facts model attempts with the same discovery-disabled flags and child environment.
- Cache successful child catalogs in memory by launcher plus omo config-source mtimes, filter the full candidate chain to child-visible models, and fail before a model spawn when none are visible.
- Treat `No API key found for <provider>` as a retryable, separately classified miss.
- Report full-chain exhaustion with every attempted candidate and fingerprinted causes (`model_not_visible:<ids>`, `auth_missing:<providers>`).
- Resolve sandbox inner commands to absolute paths or degrade to the existing identity-transform warning shape.
- Rebuild the affected committed extension bundles with CI Bun 1.3.12.

## Why

The parent registry can include extension-provided models that the discovery-disabled memory child cannot see. Proactive filtering avoids known-bad spawns while preserving the existing reactive full-chain fallback when catalog probing itself is unavailable. Auth-missing candidates now advance for the same reason: another provider in the configured chain may still work.

## Already upstream and intentionally excluded

PR #6704 already removed bare `senpi` fallback spawning and already implemented reactive model-not-found traversal across the full candidate chain. This PR does not reimplement those paths and does not impose the obsolete two-spawn cap.

## Observed behavior / evidence

- Visible filtering selected only `visible/fallback`; the exact spawn log proves `invisible/primary` was never spawned.
- All-invisible filtering returned both rejected candidates and recorded zero spawned bytes.
- `No API key found for authless` advanced to `visible/fallback`.
- A hanging catalog probe was bounded (105ms with a 100ms injected QA timeout) and degraded to reactive behavior.
- Changing the config mtime caused a second probe rather than serving the cached catalog.

Evidence:
- `.omo/evidence/20260812-memory-reflection-preflight/manual-qa.txt`
- `.omo/evidence/20260812-memory-reflection-preflight/qa-driver.ts`
- `.omo/evidence/20260812-memory-reflection-preflight/memory-suite-final.txt`
- `.omo/evidence/20260812-memory-reflection-preflight/typecheck-final.txt`
- `.omo/evidence/20260812-memory-reflection-preflight/root-bun-test-final.txt`
- `.omo/evidence/20260812-memory-reflection-preflight/ci-bun-bundle-regeneration.txt`
- `.omo/evidence/20260812-memory-reflection-preflight/bundle-check-final.txt`

## Verification

- `bun test packages/omo-senpi/src/components/memory` - 502 pass, 0 fail.
- `bun run --cwd packages/omo-senpi typecheck` - clean.
- `bun test` - 14,914 pass, 5 platform skips, 0 fail.
- `node packages/omo-senpi/plugin/scripts/build-extension.mjs --check` - current.

## Cleanup

The isolated `OMO_MEMORY_HOME`, fake launcher, fake config, and temporary probe/spawn logs were removed. The transcript includes both cleanup receipts.

## Residual risk

The catalog parser intentionally accepts Senpi's current tabular output and a one-model-id-per-line fake/compatibility format. If Senpi changes `--list-models` output incompatibly, preflight logs a warning and safely falls back to the existing reactive chain instead of blocking runs.

---
## Summary by cubic
Preflights reflection and facts model chains with a discovery‑disabled Senpi launcher so the child only spawns models it can see, and treats missing provider auth as retryable. When all candidates miss with retryable causes, it raises a typed exhaustion error and persists a fingerprinted failure detail; the sandbox now resolves the inner command via the child PATH and degrades to identity with a warning if it can’t be resolved.

- Runs discovery‑disabled model listing (`--no-* --list-models`), caches catalogs by launcher + config‑source mtimes, bounds the probe with a timeout, and falls back to reactive retries on probe failure.
- Applies to reflection and facts: throws early when no visible candidates remain; passes `SENPI_MEMORY_REFLECTION`/`SENPI_MEMORY_FACTS` and `SENPI_PTY_FORCE_PIPE` to the child; adds `senpiCommand` and optional `senpiPrefixArgs` to runners.
- Classifies retryable misses: model‑not‑visible and “No API key found for <provider>”; on chain exhaustion raises `MemoryModelExhaustedError` and persists a detail like `model_not_visible:<ids>; auth_missing:<providers>; attempted:<chain>`.
- Sandbox hardening: resolves the inner command to an absolute path via the child PATH; if unresolved, runs unsandboxed and records an explicit warning.
- Tests and bundles: adds preflight and miss coverage (including Windows‑executable fixtures and platform‑scoped timing), and rebuilds committed extension bundles.

<sup>Written for commit b40ee8c4d8c42033c276cb785123c12cb1fca231. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6790?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Fixed the independent-verification timeout blocker in `382c13292` and rebuilt the affected CI-Bun bundle in `6aa95c932`.

The committed regression uses a Node launcher wrapper that starts a grandchild inheriting stdout/stderr and keeps the wrapper alive. On the previous `close` implementation it returned the test's explicit `hung` sentinel after 500ms; after switching to `exit`, it degraded to `{ kind: "unavailable" }` in ~58-174ms with a 50ms injected probe timeout.

Evidence: `.omo/evidence/20260812-memory-reflection-preflight-timeout/red-hang.txt` and `green.txt`. Required gates: memory suite 503 pass / 0 fail, omo-senpi typecheck clean, and CI-pinned Bun 1.3.12 bundle check current.

PR #6801, opened 2026-08-12
fix(senpi-task): team members lose the omo extension to a task_send collision

ismetanin:
## Bottom line

`team_create` members are unusable on the senpi harness: every member either dies on its first turn with `Provider is not configured: <provider>` or sits at `running` forever producing nothing. Both symptoms are one defect. A separate defect kills curated read-only children with `Context remains above the compaction threshold because compaction did not complete`.

Found while running an 8-member team; 8/8 members produced zero output over ~190 minutes, and 7 `librarian` lanes died in the same run.

## Bug 1 - the omo extension is dropped inside every team member child

`spawnTeamMembers` starts each member as an RPC child with `--no-extensions` plus two threaded extensions: the member extension and the product extension inherited from the lead. Both register a tool named `task_send`, so `detectExtensionConflicts` rejects the second one and senpi drops the WHOLE product extension in that child.

Reproduced directly against the shipped engine:

```
$ senpi --mode rpc --no-extensions \
    --extension <plugin>/extensions/omo-member.js \
    --extension <plugin>
Warning: Failed to load extension ".../extensions/omo.js":
  Tool "task_send" conflicts with .../extensions/omo-member.js
```

The child therefore boots with no omo config, no agents, and no configured providers, so model resolution fails for whatever model the category resolved:

- messaged members: `child_error: Provider is not configured: openai-codex` / `anthropic`
- unmessaged members: the first turn cannot resolve a model either, the manager records `task_model_fallback` carrying that same extension warning as its `error_message`, and the record stays `running` with `turns: 0` forever

The same-package suppression in `shouldSuppressExtensionConflict` does not help: its identity key is `<packageName>:<relative path>`, so two extensions shipped by one package never compare equal.

`packages/senpi-task/AGENTS.md` already states the intended contract - "Each process member ... receives only team-scoped `task_send`" - so the fix follows the documented design: the task component skips the lead-scoped `task_send` when the process carries the member identity, and the member extension keeps owning that name.

Verified end-to-end by rebuilding the bundles and re-running the repro: no conflict, and the child now registers the omo-native surface.

## Bug 2 - curated read-only bash returns whole documents into context

`createCuratedReadonlyBashTool` replaces senpi's `bash`, whose contract truncates every result, but returned stdout whole up to the 512KB exec buffer.

From the incident transcript, one `librarian` turn issued six parallel `curl` calls to reddit.com. All six were answered with the same ~190KB "blocked by network security" HTML page, so ~1.1MB of markup entered one turn. The child went straight past its compaction threshold, compaction could not summarise a history that size, and the lane died with `Context remains above the compaction threshold because compaction did not complete` - after 6 tool calls, with `total_tokens` still reported as 4320.

Results now go through `truncateHead` with the same `DEFAULT_MAX_LINES` / `DEFAULT_MAX_BYTES` limits the replaced tool advertises, and say what was dropped.

## Tests

- `packages/omo-senpi/src/components/task/index.test.ts` - under the member identity, every tool is registered except `task_send`. Fails on `dev` (registers `task_send`), passes with the fix.
- `packages/senpi-task/src/team/member-extension/identity.test.ts` - member-identity detection.
- `packages/senpi-task/src/runners/in-process/curated-readonly-bash.test.ts` - the cap, and the tool result going through it (via a new injected executor seam, defaulted to the real one).

`bun test packages/senpi-task packages/omo-senpi`: 2842 pass, 0 fail. `tsgo --noEmit` clean for both packages.

Rebuilt bundles are committed in a separate `chore(senpi): rebuild extension bundles` commit, as the `senpi-compatibility` job verifies the committed artifact against a fresh build (`build-extension.mjs --check`).

## Not fixed here, worth a look

- A member whose first turn cannot run stays `running` indefinitely instead of transitioning to `error`. That is what turned bug 1 into a 190 minute silent hang rather than a fast failure.
- Members that died on `child_error` were recorded as `completed` with 0 tool calls, which hides the failure from status polling.
- Member child processes stayed resident after their task reached a terminal status.

---
## Summary by cubic
Stops member processes from loading the lead task surface and from leaking member identity to descendants. Previously members registered the lead tools (including a duplicate `task_send`) and children of members inherited `SENPI_TASK_MEMBER` and `omo-member.js`; now members keep only the scoped `task_send`, and generic descendants are clean. Curated read‑only bash now truncates outputs; previously it returned whole documents that could exceed compaction limits and kill lanes.

- Lead task component (`packages/omo-senpi`): skip registration when `isTeamMemberProcess()` is true (guarded by `MEMBER_IDENTITY_ENV`), leaving only the member extension’s scoped `task_send`. Members no longer register task lifecycle/control tools, commands, or lead pollers, closing unauthenticated `task_cancel`.
- RPC spawn (`packages/senpi-task`): strip `SENPI_TASK_MEMBER`, `SENPI_TASK_MEMBER_TASK_ID`, and `SENPI_TASK_TEAM_CONFIG` from generic children and drop `omo-member.js` from their extension list; apply identity and the member extension only on explicit member spawns.
- Curated read‑only bash: cap results via `truncateHead` using `DEFAULT_MAX_LINES`/`DEFAULT_MAX_BYTES`; include a “[truncated: …]” notice; add an injected executor seam for tests.
- Exports/tests: export `MEMBER_IDENTITY_ENV` and `isTeamMemberProcess` via `team/index` and the package root; add identity, spawn, and bash‑capping tests.
- Bundles: rebuild `plugin/extensions/*` so `build-extension.mjs --check` passes; `omo-member.js`/`omo-task.js` include the member guard and bash cap; `omo.js` unchanged.

No migration required.

<sup>Written for commit a71f21ebf01f00cb082b46d732d7937edc43a2c3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-senpi/src/components/task/index.ts
```diff
@@ -169,13 +170,18 @@ function registerTaskTools(
...
   })
-  pi.registerTool({
-    ...createTaskSendTool({
-      manager,
-      resolveCallerSessionId,
-      teamRouting: { service: teamService, from: TEAM_LEAD_SENTINEL, ...(resolveDefaultTeamRunId !== undefined ? { resolveDefaultTeamRunId } : {}) },
-    }),
-  })
+  // A team member child already loads the member extension, which registers the member-scoped
+  // task_send. Registering the lead-scoped one too makes senpi reject THIS whole extension on the
```
code-yeongyu: [sisyphus-bot] [P1] Prove this mechanism against the pinned Senpi before using it as the fix boundary.

`@code-yeongyu/senpi@2026.8.14` does not reject the whole later extension here. Its loader returns every successfully loaded extension, and `ExtensionRunner#getAllRegisteredTools()` keeps the first duplicate tool registration while retaining the later extension's other tools, agents, and providers. With the member bundle first, the member `task_send` wins, but the rest of OMO remains present.

That means the claimed cause of `Provider is not configured` is not established by this code or the unit test. Add an integration that loads the real base bundles through the pinned Senpi runtime, reproduces loss of OMO providers/config on the base, and turns green on this head. If that cannot be reproduced, diagnose the actual zero-output/provider failure and revise this change instead of documenting a false loader behavior.

packages/omo-senpi/src/components/task/index.ts
```diff
@@ -169,13 +170,18 @@ function registerTaskTools(
...
+  // name collision, leaving the member with no omo config, agents, or providers.
+  if (!isTeamMemberProcess()) {
```
code-yeongyu: [sisyphus-bot] [P1] The member branch must guard the whole task/lead control surface, not one duplicate name.

With `SENPI_TASK_MEMBER` set, an exact-head component registration still exposed `task`, `task_cancel`, `task_output`, `team_create`, `team_delete`, `task_create`, `task_get`, `task_list`, and `task_update`; the separately loaded member extension adds member-scoped `task_send`. This contradicts the governing contract that a process member receives only team-scoped `task_send` and never lead lifecycle/tasklist tools. In particular, `task_cancel` has no caller-session authorization and reaches the shared manager/store.

Branch at the task-component boundary so member processes do not register task/task-control/team-control tools, commands, or lead pollers. Flip the new test away from `ALL_TOOL_NAMES - task_send` to the documented member-safe surface.

packages/senpi-task/src/team/member-extension/identity.ts
```diff
@@ -0,0 +1,11 @@
+export const MEMBER_IDENTITY_ENV = "SENPI_TASK_MEMBER"
+
+/**
+ * Whether THIS process is a team member child, which owns `task_send` through the member extension.
+ * senpi drops an ENTIRE extension whose tool name is already registered, so a host extension that
+ * registers its own `task_send` here loses every other tool, agent, and provider it carries.
+ */
+export function isTeamMemberProcess(env: NodeJS.ProcessEnv = process.env): boolean {
+  const identity = env[MEMBER_IDENTITY_ENV]
+  return identity !== undefined && identity.length > 0
```
code-yeongyu, line 10: [sisyphus-bot] [P1] This is not a process-local identity signal because generic RPC children inherit it.

Exact-head `buildRpcSpawn()` with no explicit `memberEnv` produced a child with a new session directory but retained the parent's `SENPI_TASK_MEMBER`, `SENPI_TASK_MEMBER_TASK_ID`, `SENPI_TASK_TEAM_CONFIG`, plus both `omo-member.js` and `omo.js`. The descendant is therefore classified as the same member, starts another poller on that member's inbox, sends messages under the member name, and appends audit events to the parent member task.

Strip all member-only environment variables from inherited `parentEnv` and do not inherit `omo-member.js` into ordinary descendants. Apply freshly constructed member identity and the member extension only in the explicit direct-member spawn/respawn path. Add a test proving a generic child of a member is not a member while a direct team member still is.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST_CHANGES

The output-cap half is sound, but the team-member fix is not ready to merge. I reproduced three member-boundary blockers on the exact current head.

1. [P1] The stated root cause is not true for the pinned `@code-yeongyu/senpi@2026.8.14`. Its loader retains every successfully loaded extension, and its runner keeps the first duplicate tool registration without discarding the later extension's other tools, agents, or providers. With member-first ordering, only the lead `task_send` is shadowed. There is no base-red/head-green integration test proving that this change fixes the reported `Provider is not configured` incident.
2. [P1] The new member-path test explicitly pins the full task and lead-team control surface. An exact-head member registration exposed `task`, `task_cancel`, `task_output`, `team_create`, `team_delete`, and all tasklist tools; the member extension then adds `task_send`. This contradicts the governing contract that process members receive only team-scoped `task_send`. `task_cancel` has no caller-session ownership check and operates on the shared manager/store.
3. [P1] Ordinary descendants inherit `SENPI_TASK_MEMBER`, the parent member task ID, team config, and both extensions. The new `isTeamMemberProcess()` predicate therefore classifies a generic child as the parent member, starts a duplicate member runtime, and lets it consume/send/audit as that member.

Focused tests pass (22 tests, 56 assertions), all current CI checks are green, and the generated bundles are current. Those facts do not prove the reported incident is fixed; the new unit test actually locks in the unsafe member surface. Please add a real pinned-Senpi integration that fails on the base and passes on the head, make the task component register no lead/task control tools in member processes, and strip all member-only environment/extension state from ordinary descendants before applying explicit `memberEnv` for direct members.

packages/omo-senpi/src/components/task/index.ts
ismetanin: Verified against the pinned runtime, and you are right — I withdraw the stated mechanism.

Reading the installed `@code-yeongyu/senpi@2026.8.14`:
- `dist/core/extensions/loader.js:277` `registerTool()` writes into that extension's own map with no cross-extension duplicate check and no throw; the only reserved name is `tool_search`.
- `dist/core/extensions/runner.js:405` `getAllRegisteredTools()` keys a Map by tool **name** across all extensions, first-registration-wins, and still returns every other tool the later extension carries — its own docstring says exactly that.

So only the duplicate `task_send` name was ever shadowed; omo's config, agents and providers were never lost this way. The comment asserting whole-extension rejection was wrong and is removed in a71f21eb.

I have not added a base-red/head-green integration for a mechanism that does not exist, and I am not claiming the reported `Provider is not configured` incident is explained by this change. The guard is retained on the grounds that actually hold: the member-surface contract (your next comment) and descendant misclassification (the one after).

packages/omo-senpi/src/components/task/index.ts
ismetanin: Fixed in a71f21eb by branching at the task-component boundary rather than on one tool name.

`isTeamMemberProcess()` now returns before any registration in `packages/omo-senpi/src/components/task/index.ts`, so a member process registers no task lifecycle tools, no task-control tools, no team-control tools, no commands, no event handlers and no lead pollers. The separately loaded member extension's scoped `task_send` is the member's only task tool. That closes `task_cancel` reaching the shared manager/store with no caller-session authorization.

The test was flipped as you asked: it no longer asserts `ALL_TOOL_NAMES - task_send`. It preloads the member `task_send` and asserts the resulting member surface is exactly `["task_send"]`, plus zero commands, handlers and renderers. It fails before the fix (nine extra controls) and passes after.

One note for the record: the early return also precedes `wireSessionStartProcessSweep`. That is a no-op in practice because the sweep already self-skips when `SENPI_CODING_AGENT_SESSION_DIR` is set (`process-sweep.ts:53`), and the member extension requires that same variable (`member-extension/index.ts:71`), so a member process never ran the sweep anyway.

packages/senpi-task/src/team/member-extension/identity.ts
ismetanin: Confirmed and fixed in a71f21eb.

`buildRpcSpawn()` now deletes `SENPI_TASK_MEMBER`, `SENPI_TASK_MEMBER_TASK_ID` and `SENPI_TASK_TEAM_CONFIG` from the copied `parentEnv`, then applies `spec.memberEnv` only when the caller explicitly supplies it, and filters `omo-member.js` out of the child extension list for generic descendants while leaving ordinary extensions such as `omo.js` intact. Member identity and the member extension are applied only on the explicit direct-member spawn/respawn path.

The env name set and bundle name are now single-sourced in `member-extension/identity.ts` so the strip list cannot drift from the spawn code.

Test added as requested: a generic child of a member is not classified as a member, keeps `omo.js` and drops `omo-member.js`; an explicit direct member gets fresh identity, is classified as a member and keeps both. I verified it is load-bearing by removing the strip loop — the suite goes 21 pass/0 fail to 20 pass/1 fail.

ismetanin:
All three findings are addressed in a71f21eb. Inline replies have the detail; summary:

1. **[P1] Root cause was false — withdrawn.** I checked the installed `@code-yeongyu/senpi@2026.8.14` rather than defending the claim. `loader.js:277` `registerTool()` has no cross-extension duplicate check and no throw (only `tool_search` is reserved), and `runner.js:405` `getAllRegisteredTools()` de-duplicates by tool **name**, first-wins, while retaining every other tool the later extension carries. So only the duplicate `task_send` was shadowed and provider loss was never the mechanism. The misleading comment is gone. I have **not** added an integration test for a mechanism that does not exist, and I am not claiming this change explains the reported `Provider is not configured` incident.
2. **[P1] Member surface.** The guard moved to the task-component boundary: a member process now registers no task lifecycle, task-control or team-control tools, no commands, no handlers and no lead pollers — the member extension's scoped `task_send` is its only task tool. That closes `task_cancel` reaching the shared manager/store without caller-session authorization. The test was flipped from `ALL_TOOL_NAMES - task_send` to asserting exactly `["task_send"]`.
3. **[P1] Descendant misclassification.** `buildRpcSpawn()` strips `SENPI_TASK_MEMBER`, `SENPI_TASK_MEMBER_TASK_ID` and `SENPI_TASK_TEAM_CONFIG` from the inherited env and drops `omo-member.js` for generic descendants, applying member identity and the member extension only on the explicit direct-member path. Test proves a generic child of a member is not a member while a direct member still is; I confirmed it is load-bearing by removing the strip loop (21 pass -> 20 pass/1 fail).

The curated-readonly-bash output cap is untouched, as you indicated that half is sound.

Verification: `packages/senpi-task` 1376 pass, `packages/omo-senpi` 1569 pass, both `tsgo` projects clean, bundles rebuilt and `build-extension.mjs --check` reports current.

Since the justification changed, the honest framing of this PR is now: it enforces the member-surface contract and stops member identity leaking into ordinary descendants. It is not a fix for the provider-loss incident, and I have retitled my reasoning accordingly rather than leaving a false rationale in the tree.

code-yeongyu:
Thank you, @ismetanin, for the detailed reproduction and for identifying the unsafe member task surface and descendant identity leak.

We rebuilt the repair against current `dev` and split the incident into the actual runtime boundaries:

- member processes now expose only member-scoped `task_send`;
- generic descendants strip member identity and `omo-member.js`;
- create and respawn share one canonical member-first extension profile;
- process models are admitted against the exact child profile before launch;
- prompt/turn failures now reach typed durable terminal outcomes.

The replacement is #6894. It includes Ivan Smetanin's authorship on the reusable boundary increment.

One correction to the original explanation: pinned Senpi does not reject the later extension wholesale when a tool name collides. The extension remains loaded and duplicate tool resolution is first-wins. The replacement therefore removes the unsafe overlapping surface without relying on whole-extension rejection as the provider-failure cause.

A follow-up issue (#6895) tracks the broader process-child resource parity contract that this incident repair deliberately does not invent.

All current review concerns were re-audited against the replacement, including direct member identity preservation, generic descendant stripping, respawn parity, model visibility, crash/restart, and lead/member delivery.

Closing this PR as superseded by #6894. Thank you again for surfacing the failure with enough detail to drive the root fix.

PR #6806, opened 2026-08-12
feat(background-agent): parent-wake lifecycle ledger for lost-notification forensics

sjawhar:
## Summary

- Background-task completion notifications sometimes never reactivate the parent session, and the current logs cannot say why: the terminal give-up path (`Stopped retrying parent wake after repeated no-output dispatch`, 76 hits/day on one of our machines) records only `sessionID` + `retryCount`, which cannot distinguish a benign noReply-deposit cleanup from a reply-required wake that silently died.
- This PR adds a structured `[parent-wake-ledger]` log line at every parent-wake lifecycle transition so a lost notification can be traced end-to-end after the fact: which tasks it carried, whether it required a reply, every dispatch/deferral, and exactly where and why it was abandoned.
- Log-only: no behavioral changes to the wake state machine.

## Changes

- `parent-wake-ledger.ts` (new): ledger event vocabulary, stable per-wake `wakeId` generation, notification header + task-id extraction, wake summarization.
- `parent-wake-pending-queue.ts`: assigns `wakeId` on new wakes (preserved through merges/requeues/clones); emits `queued` / `merged` / `requeued`.
- `parent-wake-prompt-dispatch.ts`: emits `dispatched` (with the actual `noReply` body flag, retained state, prompt-gate status) and `dispatch-requeued` (gate-reserved / gate-skipped / dispatch-error).
- `parent-wake-window-recovery.ts`: emits `output-observed` / `no-output-requeue` / `abandoned`; `abandoned` additionally records `lostReplyWake` (reply-required wake abandoned with no retained pending copy = genuinely lost) and a parent-session snapshot via the new `ParentWakeSessionInspector.snapshotParentSessionState()`.
- `parent-wake-flush-runner.ts`: emits `consumed-dropped` / `duplicate-suppressed` on the two silent drop paths.
- `parent-wake-dedupe.ts`: `ledgerId` field on `PendingParentWake` + clone support; exports `getSystemReminderHeaderLines`.
- `parent-wake-ledger.test.ts` (new): extraction, id lifecycle through the real pending queue, and window-recovery abandonment forensics with the real tracker/inspector classes.

## QA & Evidence

- **What was tested:** Live end-to-end run against real `opencode serve` in an isolated sandbox (XDG_*/HOME/TMPDIR redirected, real haiku model, this branch's plugin source): parent launched one background explore task, wake flow ran for real.
[...]

PR #6807, opened 2026-08-12
feat(background-agent): add background_task.wakeOnEachCompletion config

sjawhar:
## Summary

- Today a parent session that fans out N background tasks stays idle until the LAST one completes: intermediate completions are deposited into the transcript as internal noReply messages that never trigger a turn (`shouldReply = allComplete || isTaskFailure` in `notifyParentSession`).
- When results are independently actionable, the parent should be able to act on the first result immediately. New `background_task.wakeOnEachCompletion` config (default `false`) makes every completion queue a reply-required wake instead.
- Default-off: existing behavior is byte-identical unless the flag is set. The existing per-parent merge queue, dedupe, and final-notification precedence already coalesce wake bursts when several tasks finish close together.

## Changes

- `config/schema/background-task.ts`: new optional `wakeOnEachCompletion` boolean with doc comment; `assets/*.schema.json` regenerated via `bun run build:schema`.
- `features/background-agent/manager.ts`: `shouldReply = allComplete || isTaskFailure || this.config?.wakeOnEachCompletion === true` (one line).
- `docs/reference/configuration.md`: documents the new key in the Background Tasks table.
- `manager-wake-on-each-completion.test.ts` (new): pins all three policy quadrants (flag on + sibling running -> reply; flag off + sibling running -> no-reply; flag off + last task -> reply).

## QA & Evidence

- **What was tested:** Live end-to-end run against real `opencode serve` in an isolated sandbox (XDG_*/HOME/TMPDIR redirected, real haiku model, this branch's plugin). Flag enabled ONLY via a project-layer `.omo/omo.json` (`"[opencode]".background_task.wakeOnEachCompletion: true`), proving the config plumbing end to end. Parent launched a fast child and a slow child (`sleep 60`) in one turn.
  **Observed result:** fast child completed -> `Queued notification ... {"allComplete":false,"isTaskFailure":false,"shouldReply":true}` (impossible under the old policy for non-failure completions) -> `Sent deferred parent wake` dispatched a full minute BEFORE the slow child completed, which then produced the final `allComplete:true` wake. RESULT=PASS; real opencode DB session count unchanged.
  **Artifact:** `.omo/evidence/20260812-wake-on-each-completion/` in the working tree (gitignored; runner script + wake timeline + isolation receipt; key excerpt above).
  **Why sufficient:** exercises config file -> schema -> manager policy -> actual early parent turn on real opencode; the flag-off default is pinned by unit tests.
- **What was tested:** `bun test packages/omo-opencode/src/features/background-agent/` and schema suite.
  **Observed result:** all pass (including the 3 new policy tests).
  **Why sufficient:** covers the only changed code path in both flag states.

## Risks & Residuals

- Flag-off path unchanged (pinned by test). Not applicable.
- Flag-on wake bursts rely on existing merge/dedupe machinery, exercised in the live run (fast+slow overlap). Mitigated.

## Automated Checks

```bash
bun run typecheck
bun run build:schema
```

## Related Issues

---
## Summary by cubic
Adds optional per-completion wakes for background tasks so the parent session can act on the first result. Previously the parent only woke when all tasks finished or one failed; with `background_task.wakeOnEachCompletion: true`, each completion queues a reply-required wake (default off; existing wake coalescing still applies).

- Manager: includes the flag in `shouldReply` and passes it to the notification builder to switch wording.
- Notification template: partial-completion messages say “You will be notified as each task completes” when enabled; otherwise “You WILL be notified when ALL complete.”
- Dedupe: merge logic recognizes both progress wordings and drops stale per-completion notifications when the final wake arrives.
- Config/Docs: adds `wakeOnEachCompletion` to the `background_task` schema and reference docs; JSON schemas regenerated.
- Tests: cover policy (flag on/off and final completion), template messaging, and dedupe behavior. No change unless enabled.

**Rollout**
- To opt in, set `"background_task": { "wakeOnEachCompletion": true }` in project config.

<sup>Written for commit d013ac3f45d281fdb81dca5bda0b7871e9f816b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -9262,6 +9262,9 @@
           "minimum": 10,
           "maximum": 9007199254740991
         },
+        "wakeOnEachCompletion": {
+          "type": "boolean"
+        },
         "circuitBreaker": {
           "type": "object",
           "properties": {
--- a/assets/omo.schema.json
+++ b/assets/omo.schema.json
@@ -10980,6 +10980,9 @@
               "minimum": 10,
               "maximum": 9007199254740991
             },
+            "wakeOnEachCompletion": {
+              "type": "boolean"
+            },
             "circuitBreaker": {
               "type": "object",
               "properties": {
@@ -25456,6 +25459,9 @@
                     "type": "integer",
                     "minimum": 10,
                     "maximum": 9007199254740991
+                  },
+                  "wakeOnEachCompletion": {
+                    "type": "boolean"
                   },
                   "circuitBreaker": {
                     "type": "object",
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -476,7 +476,8 @@ Control parallel agent execution and concurrency limits.
     "defaultConcurrency": 5,
     "staleTimeoutMs": 180000,
     "providerConcurrency": { "anthropic": 3, "openai": 5, "google": 10 },
-    "modelConcurrency": { "anthropic/claude-opus-5": 2 }
+    "modelConcurrency": { "anthropic/claude-opus-5": 2 },
+    "wakeOnEachCompletion": false
   }
 }
 ```
@@ -495,6 +496,7 @@ Control parallel agent execution and concurrency limits.
 | `syncPollTimeoutMs`         | -         | Synchronous polling timeout in milliseconds (min: 60000)             |
 | `maxToolCalls`              | `200`     | Maximum tool calls per subagent task (min: 10)                        |
 | `circuitBreaker`            | -         | Circuit-breaker object: `enabled`, `maxToolCalls`, `consecutiveThreshold` |
+| `wakeOnEachCompletion`      | `false`   | Wake the parent session after each background task completes. By default the parent is only woken once all of its background tasks finish (or one fails); intermediate results are deposited into the transcript without triggering a turn. |
 
 Priority: `modelConcurrency` > `providerConcurrency` > `defaultConcurrency`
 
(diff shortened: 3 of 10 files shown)


sjawhar:
CI note: `test (windows-latest)` failed on 2 tests unrelated to this diff (`ensureCodegraphProvisioned version upgrades`, `palace command`), both ~5s-timeout shaped. The sibling PR #6806 (disjoint diff) fails a completely different unrelated set on the same job, which points at Windows flakiness rather than either change. This PR is a one-line policy change in `background-agent/manager.ts` plus config schema/docs; ubuntu/macos test jobs are green. I don't have rerun rights here — a failed-jobs rerun should clear it.

sjawhar:
Pushed a follow-up to the same commit: with `wakeOnEachCompletion` enabled, the partial-completion notification used to say "You WILL be notified when ALL complete," which is wrong in that mode — the parent is woken per completion. The template now emits "You will be notified as each task completes." when the flag is on (default text unchanged), the progress-notification classifier regex in `parent-wake-dedupe.ts` accepts both variants so stale progress deposits still merge away behind a final wake, and tests pin both variants (template, dedupe merge, manager policy). Verified live: a two-child fan-out (1s and 75s) woke the parent on the first completion with the corrected semantics.

PR #6829, opened 2026-08-13
fix(omo-senpi): isolate ulw-loop continuation by Senpi session

### LINKED ISSUE #6828: [Bug]: omo-senpi ulw-loop continuation can hijack another session in the same working directory
## Bug description

When two independent OMO Native / Senpi instances are running in the **same working directory**, an `ulw-loop` started by one session can be detected and automatically continued by the other unrelated session.

I observed this on `omo-ai@5.0.0-0.beta.7`.

Example:

- **Session A** is running an active `ulw-loop` in repository `R`.
- **Session B** is another OMO instance in the same repository `R`, but is working on a completely unrelated task/conversation.
- When B finishes its own turn/task, B receives an automatic ULW continuation and starts trying to continue A's `ulw-loop`.
- This can happen even while A is still actively working on the ULW task.

The two sessions are logically unrelated; the only shared property is the working directory.

## Minimal reproduction

1. Open repository `R` in terminal A.
2. Launch OMO Native / Senpi and start an `ulw-loop` with unfinished goals.
3. Leave A running.
4. Open a second terminal in the same repository `R`.
5. Launch a separate OMO Native / Senpi instance B.
6. In B, run an unrelated ordinary task that does not use `ulw-loop`.
7. Let B finish its turn.
8. Observe that B can receive the automatic continuation message for A's ULW run, e.g.:

   ```text
   Continue the active omo-agent-toolkit ulw-loop run.
   Run `omo-agent-toolkit ulw-loop status --json` in this session cwd...
   ```

9. B may then inspect and work from A's `.omo/ulw-loop` state even though B never started or adopted that run.

## Expected behavior

ULW continuation ownership should be session-scoped.

For two sessions A and B in the same repository:

```text
agent_end(A) -> may continue ULW(A)
agent_end(B) -> must only inspect/continue ULW(B)
```

If B has no ULW run, B should simply do nothing. B should not see, steer, continue, checkpoint, or mutate A's ULW run merely because both sessions share a cwd.

Ideally the persisted layout should remain isolated by session, for example:

```text
.omo/ulw-loop/<session-A>/...
.omo/ulw-loop/<session-B>/...
```

## Actual behavior

The effective behavior appears closer to:

```text
agent_end(any Senpi session in cwd R)
  -> check whether R has an active ULW run
  -> if yes, inject continuation into the current session
```

This makes the repository/cwd act like the continuation identity instead of the Senpi session.

## Source-level investigation / suspected root cause

This section is a hypothesis based on reading `v5.0.0-beta.7` and current `dev` source at the time of filing.

In:

```text
packages/omo-senpi/src/components/ulw-loop/index.ts
```

`agent_end` already extracts the current Senpi session id:

```ts
const sessionId = extractSessionId(eventCtx)
```

but that session id is only used for the boulder/start-work eligibility check:

```ts
if (sessionId && findContinuableBoulderWork(cwd, sessionId) !== null) {
  ...
}
```

The ULW status lookup that follows does not appear to pass the Senpi session id:

```ts
const status = await readActiveStatus(omoBin, runCommand, cwd, ctx)
```

and `readActiveStatus()` executes the equivalent of:

```text
omo-agent-toolkit ulw-loop status --json
```

with only the cwd supplied to the subprocess.

This is important because the ULW CLI itself already supports session scoping.

In:

```text
packages/omo-codex/plugin/components/ulw-loop/src/paths.ts
packages/omo-codex/plugin/components/ulw-loop/src/cli-commands.ts
```

there is existing support for:

```text
--session-id <id>
OMO_ULW_LOOP_SESSION_ID
CODEX_SESSION_ID
CODEX_THREAD_ID
```

and a scoped run is stored under:

```text
.omo/ulw-loop/<sessionId>/
```

while an unscoped run falls back to the repository-level `.omo/ulw-loop/` state.

My current hypothesis is therefore:

1. Senpi knows the current logical session id.
2. The omo-senpi ULW adapter drops that identity before querying the ULW CLI.
3. `status --json` therefore falls back to unscoped repository state.
4. Any Senpi instance in the same cwd can observe the same active ULW run.
5. Its own `agent_end` then triggers `omo-senpi:ulw-continuation` for a run owned by another session.

The same issue may affect more than `agent_end`, because the ULW component also refreshes/checks status on session/input/tool-result paths. A foreign run could potentially cause steering/footer contamination as well as continuation takeover.

## Why this looks different from #6752 / #6758

#6752 and PR #6758 address paused/explicit-stop continuation behavior in omo-senpi.

This report is about **cross-session / cross-process isolation**:

- A and B are separate OMO instances.
- B has not stopped or resumed A.
- B never started A's ULW run.
- B should not be eligible to continue A at all.

An in-memory `Set<sessionId>` stop guard cannot by itself prevent a second independent OMO process from discovering an unscoped filesystem ULW state in the same cwd.

## Suggested fix direction

The cleanest fix may be to reuse the existing ULW session-scoping primitives instead of introducing a second state model.

### 1. Propagate the Senpi session id into every ULW status lookup

For example, internally execute the equivalent of:

```text
omo-agent-toolkit ulw-loop status --json --session-id <senpi-session-id>
```

and fail closed when no Senpi session id is available rather than falling back to repository-global ULW state.

### 2. Scope all Senpi ULW CLI operations, not only `agent_end`

The adapter checks ULW state from multiple lifecycle paths (`session_start`, `input`, `agent_end`, `tool_result`). They should all resolve the same current-session scope.

### 3. Make shell-invoked ULW commands inherit the same scope

The Senpi `ulw-loop` skill instructs the model to run commands such as:

```text
omo-agent-toolkit ulw-loop create-goals
omo-agent-toolkit ulw-loop status --json
omo-agent-toolkit ulw-loop checkpoint ...
```

Because the CLI already understands `OMO_ULW_LOOP_SESSION_ID`, OMO Native could establish that environment variable for the active Senpi session so child shell commands automatically use the correct `.omo/ulw-loop/<sessionId>/` state.

An explicit `--session-id` in internal adapter calls would still be useful as a fail-closed correctness boundary.

### 4. Do not automatically claim legacy unscoped ULW state

If `.omo/ulw-loop/` contains an old unscoped run, an arbitrary session in the same cwd should not silently adopt it. Adoption/migration should be explicit, otherwise the same race remains possible.

### 5. Optional follow-up: same-session process ownership

Session isolation solves A-vs-B. A separate hardening layer may still be useful if the *same logical session* is opened in two live processes at once (lease / instance ownership / CAS), but that appears separable from this bug.

## Acceptance tests

At minimum I think the regression suite should include a real shared-cwd two-session case:

### Foreign session must not continue

```text
repo R

Session A:
  create active unfinished ULW(A)

Session B:
  same cwd R
  ordinary unrelated task
  agent_end(B)
```

Expected:

- B receives no `omo-senpi:ulw-continuation`.
- B does not receive the ULW steering reminder from A's run.
- B does not mutate A's goals/ledger.
- A remains able to continue normally.

### Two ULW sessions in one cwd remain independent

```text
A -> .omo/ulw-loop/<A>/...
B -> .omo/ulw-loop/<B>/...
```

Mutating/checkpointing A must not change B and vice versa.

### Missing session identity fails closed

If the Senpi session id cannot be resolved, automatic continuation should not fall back to an arbitrary repository-global active run.

## Impact

I consider this a correctness/concurrency issue rather than only a UX issue. In the worst case two independent agents can both act on one ULW ledger/codebase, causing duplicate execution, incorrect checkpoints, conflicting edits, or one session declaring progress/completion for another session's task.

## Version / source checked

- Observed: `omo-ai@5.0.0-0.beta.7`
- beta.7 tag: `v5.0.0-beta.7`
- I also re-checked current `dev` before filing and did not find an existing fix or an exact duplicate issue/PR.

Altairpaca:
## Summary

Fixes #6828. The omo-senpi `ulw-loop` component queried the agent toolkit without any session identity, so the CLI fell back to the cwd-global `.omo/ulw-loop/` state. Two independent Senpi instances in one working directory could therefore see and auto-continue each other's ulw-loop runs: when session B finished an unrelated task, its `agent_end` read session A's active run and injected "Continue the active omo-agent-toolkit ulw-loop run" into B.

The fix scopes Senpi ULW run identity to `(repoRoot, senpi sessionId)` — the same isolation the Codex CLI already supports via `--session-id` / `OMO_ULW_LOOP_SESSION_ID`.

## Changes

- **New `session-scope.ts`** (`packages/omo-senpi/src/components/ulw-loop/`): resolves a filesystem-safe `senpi-<sessionId>` scope from the host `sessionManager` (`extractSenpiSessionId` / `resolveSenpiUlwSessionId` / `normalizeSenpiUlwScopeId`), plus a `SenpiUlwSessionScope` manager owning the `OMO_ULW_LOOP_SESSION_ID` env lifecycle (set on `session_start`, restored on switch/shutdown only while the current value is still ours — never clobbers an external value).
- **`index.ts`** (`ulw-loop` component): every status read (`session_start`, `input`, `agent_end`, `tool_result`) now passes `--session-id senpi-<id>`; with no session identity the component **fails closed** — it never reads the unscoped root state (no legacy auto-claim). The env var additionally makes model-spawned bash shells resolve all `omo-agent-toolkit ulw-loop *` commands into the session-scoped directory automatically. IdleInjectionCoordinator, boulder deferral, continuation cap and stale-status logic are untouched.
- **Tests**: scope-aware RED/GREEN regression block in `index.test.ts` (foreign continuation + foreign steering reminder blocked, own-session continuation preserved, fail-closed, env lifecycle), new `session-scope.test.ts`, updated host-context fixtures across the ulw-loop suite, real-spawn argv proof in `runtime.test.ts`.
- **E2E probe** `scripts/qa/probe-cross-session.mjs`: two real senpi processes in one shared sandbox cwd (mock provider, local staged toolkit). A creates an incomplete run through the real bash tool; B finishes an unrelated task. Asserts A's continuation lands in A's session record, B's has none, state lives under `.omo/ulw-loop/senpi-<id>/` with no root state, and A's goals+ledger digests are unchanged.
- **Generated**: `plugin/extensions/omo.js` regenerated (bun 1.3.12, CI-pinned; `build-extension.mjs --check` passes). Skill note: Senpi runtime owns `OMO_ULW_LOOP_SESSION_ID`.

## QA & Evidence

- **What was tested (unit):** `bun test packages/omo-senpi/src/components/ulw-loop packages/omo-senpi/src/extension/idle-injection-wiring.test.ts` → 67 pass / 0 fail. RED captured first (8 failing cases, 2 own-session pins green) against unmodified dev.
- **What was tested (gate):** `bun run test:senpi` → 1524 pass / 1 skip. The memory-worker supervisor failures observed on this workstation are machine-specific (CI ubuntu/macos/windows are green); the windows flake that blocked the required `test (windows-latest)` check (real-git memory tool tests exceeding bun's 5s default on slow windows runners) was root-fixed in this PR with explicit test timeouts (`0ea74b59a`). `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` clean. Final commit `0ea74b59a`: all required checks green (test x3, typecheck x3, codex-compatibility x3, build, senpi-compatibility ubuntu).
- **What was tested (real surface):** `SENPI_BIN=... node packages/omo-senpi/scripts/qa/probe-cross-session.mjs` → PASS on 3 consecutive runs: `aContinuationObserved=true, bContinuationObserved=false, scoped=senpi-<id>, rootUnscopedGoals=false, realSenpiUntouched=true`; B's stderr shows `ulw-loop continuation skipped { reason: 'inactive' }`.
- **Artifacts:** `.omo/evidence/omo-senpi-adapter/20260813-ulw-session-scope/` (`01-cross-session-e2e.md`, `02-probe-results.json`, `03-unit-red-green.md`). Isolation: sandboxed `SENPI_CODING_AGENT_DIR`/`XDG_CONFIG_HOME`, mock provider, real `~/.senpi/agent` digest unchanged.
- **Why sufficient:** unit tests prove the wiring; the two-process probe reproduces the exact reported scenario on the real harness with the local build.

## Risks & Residuals

- Legacy unscoped `.omo/ulw-loop/` state is deliberately NOT auto-adopted (would recreate the hijack); explicit adopt/migrate command is a documented follow-up.
- Same-session cross-process execution (two processes resuming one session id) still needs a lease; explicitly deferred (Phase 2 in the issue analysis).
- The base-branch windows flake in the real-git memory tool tests (5s default test timeout exceeded) was root-fixed in this PR with explicit `{ timeout: 20000 }` budgets; the required `test (windows-latest)` check is green on the final commit. The remaining memory-worker supervisor failures seen on the dev workstation are machine-specific (they do not reproduce in CI).

## Related Issues

Fixes #6828

---
## Summary by cubic
Isolates `omo-senpi` `ulw-loop` continuation and status by Senpi session to prevent cross-session takeover in a shared cwd. Previously the adapter read cwd-global `.omo/ulw-loop/`; now every status read passes `--session-id senpi-<sessionId>` and the component fails closed when no session id is present.

- New `session-scope.ts`: derives filesystem-safe `senpi-<sessionId>` (`extractSenpiSessionId`/`resolveSenpiUlwSessionId`/`normalizeSenpiUlwScopeId`) and manages `OMO_ULW_LOOP_SESSION_ID` via `SenpiUlwSessionScope` (set on `session_start`, restored on `session_before_switch`/`session_shutdown` only if unchanged).
- `ulw-loop/index.ts`: `session_start`, queued `input` (steer), `tool_result`, and `agent_end` now read scoped status; without a session id, no unscoped state is read and no continuation/reminder is injected. Footer status respects the scope; env is cleared on switch/shutdown. `plugin/extensions/omo.js` regenerated; `skills/ulw-loop/SKILL.md` documents that the runtime owns `OMO_ULW_LOOP_SESSION_ID`.
- Tests/QA: added session-scoping unit tests and a two-process shared-cwd probe `scripts/qa/probe-cross-session.mjs` proving isolation and env reach into child shells; fixed `withEnvAsync` to persist across awaits; adjusted toolkit-chain test to register under a fake PATH and dispatch under the real env to keep `cmd.exe` resolvable; added explicit timeouts for Windows real-git and palace fixture tests.

**Migration**
- Hosts without a `sessionManager` must pass a session id to re-enable continuation and steering.
- Legacy unscoped `.omo/ulw-loop/` state is not auto-adopted.

<sup>Written for commit c81e5c035b18f8a944c21b9b26bf29771a21c4a3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Altairpaca:
All required CI checks are green (test x3, typecheck x3, codex-compatibility x3, build, senpi-compatibility ubuntu). QA evidence for the cross-session isolation fix is recorded under .omo/evidence/omo-senpi-adapter/20260813-ulw-session-scope/ (unit RED->GREEN plus the two-process probe: A continues, B receives no continuation, state under .omo/ulw-loop/senpi-<id>/). Ready for merge.

Altairpaca:
Rebased onto latest dev (28f184c5a, includes #6860 agent-dir unification — verified no impact on the session-scoped ULW env inheritance; E2E probe re-passed). Head: c81e5c035. The CI run for this head awaits workflow approval (same gate as the previous heads).

Altairpaca:
test (windows-latest) flaked on the pre-existing 'memory run supervisor IC-8' timing assertion (875ms; not touched by this PR — the same suite fails/passes intermittently on windows runners and passed on the three previous runs of this PR). Could you re-run job 94707839774? Everything else on c81e5c035 is green.

code-yeongyu:
Hi @Altairpaca — thank you for this, and sorry it sat for a while. A short status update so you know exactly where things stand.

**This PR is staying open on purpose.** Your diagnosis in it is correct, and part of the problem is now fixed, but the part your PR is built around is still open.

What landed: #6914 was merged as `9b24f00400865a6c99451053beb7a4597372158b`, which added `PI_SESSION_ID` to `SESSION_ENV_KEYS` in the toolkit's scope resolver. That correctly scopes every ulw-loop CLI call the **model** makes, because senpi's bash tool injects `PI_SESSION_ID` into each spawn it creates.

What it does not reach — and what your `SenpiUlwSessionScope` does: the component's own status probe. senpi never puts `PI_SESSION_ID` on the extension host's `process.env` (it exists only inside bash-tool spawn envs), and `runOmoCommand` spawns the toolkit with `{ cwd, stdio }` and no `env`, so the probe inherits a host environment with no session identity and falls back to the repo-level `.omo/ulw-loop/`. Measured on a live host process: `PI_SESSION_ID` occurrences = 0.

The result is that state is now written session-scoped but still read unscoped:

```console
$ OMO_ULW_LOOP_SESSION_ID=sess-A omo-agent-toolkit ulw-loop create-goals --brief "- owner run" --json
$ find .omo -name goals.json
.omo/ulw-loop/sess-A/goals.json
$ omo-agent-toolkit ulw-loop status --json      # what the component probe sees
{ "ok": false, "error": { "code": "ULW_LOOP_PLAN_MISSING", ... ".omo/ulw-loop/goals.json" } }
```

I filed #7108 with the full evidence and impact. Your approach here — owning the `OMO_ULW_LOOP_SESSION_ID` lifecycle on the host at `session_start`, restoring it on switch/shutdown, and never clobbering an externally set value — is aimed squarely at that gap, which is why I am not closing this.

If you are still up for it, the one thing this needs is a rebase onto current `dev`. It has drifted since 2026-08-14 and now overlaps with two changes in the same files: #6914 (the resolver key) and `17ef7ad90`, a perf change that skips the toolkit spawn entirely when `<cwd>/.omo/ulw-loop` does not exist. Worth a look at how your scope manager interacts with that guard. If you would rather hand it off, say so and we can carry it forward from here — either way the credit for the analysis is yours.

One suggestion for whatever lands: a regression test that exercises the component with **no session id in the host env**. `scripts/qa/probe-cross-session.mjs` sets `PI_SESSION_ID` on the children it spawns, which models an environment the real host does not provide — that is precisely why this gap passed CI.

Thanks again for the careful write-up on this one.

— [sisyphus-bot]

code-yeongyu:
Hi @Altairpaca — closing this one, with thanks and an explanation of where your work ended up.

The gap you identified is now fully fixed on `dev`, in two parts:

1. #6914 (merged, `9b24f0040`) added `PI_SESSION_ID` to the toolkit's scope resolver, which scopes every ulw-loop CLI call the **model** makes.
2. #7112 (merged, `91568ef4294`) closed the other half you were pointing at — the component's own status probe. It now resolves the current session id, normalizes it with the toolkit's exact rules, passes `--session-id` on all four hooks (`session_start`, `input`, `agent_end`, `tool_result`), and **fails closed** when no session identity is available rather than falling back to repo-global `.omo/ulw-loop/`.

Your diagnosis in this PR was correct and load-bearing. Two specific calls of yours shaped the final fix:

- **Session identity is dropped before the toolkit query.** That is exactly the defect, and it survived #6914 precisely because that PR fixed the env-delivery path and not the spawn path. Your write-up is what made the remaining half legible.
- **Session scope belongs in its own module.** #7112 landed `session-scope.ts` doing exactly that — id extraction, normalization, and argv construction in one cohesive unit — rather than inlining it. Your `session-scope.ts` proposal was the right shape.

We carried the implementation ourselves rather than asking you to rebase a 17-file branch that had drifted three weeks against two overlapping changes (#6914 and `17ef7ad90`, the perf guard). The merge commit carries `Co-authored-by: Altair Li <altairpaca@gmail.com>` so the credit travels with the code.

One thing your branch had that we deliberately did **not** carry over: the `OMO_ULW_LOOP_SESSION_ID` host-env lifecycle manager (set on `session_start`, restore on switch/shutdown, never clobber an external value). Passing `--session-id` explicitly per invocation avoids mutating shared process state, so we went with the narrower change. If you still think the env-lifecycle approach is needed for cases the explicit flag misses — e.g. toolkit calls made by paths we don't control — please say so and we'll reopen that discussion.

Closing as resolved by #7112. Thank you for the careful analysis, and sorry it took a detour to land.

— [sisyphus-bot]

PR #6740, opened 2026-08-11
fix(utils): resolve Windows taskkill/powershell by absolute path for sanitized PATH

grim-susemi:
## Summary

Windows process-tree termination and process listing resolve `taskkill.exe` / `powershell.exe` by **bare name**, which fails whenever the ambient `PATH` is sanitized (e.g. a terminal app that starts shells without `C:\Windows\System32`). In older bundles this surfaced as a fatal runtime crash (`uncaughtException: spawn taskkill ENOENT`, see #6738); in current `dev` it silently degrades — tree kills never happen and process enumeration returns nothing.

This PR resolves both tools by absolute path (`%SystemRoot%\System32\...`) with a `C:\Windows` fallback, at every in-repo call site.

## Changes

- `packages/utils/src/system-tool-paths.ts` (new): `resolveWindowsTaskkillPath()` / `resolveWindowsPowerShellPath()` — pure resolvers, testable.
- `packages/utils/src/process-tree-termination.ts`: `attemptWindowsTreeSignal` uses the resolver (was `execFile("taskkill.exe", …)`).
- `packages/utils/src/process-sweep/exec.ts`: `createWindowsKiller()` + Windows process enumeration use the resolvers (were `"taskkill.exe"` / `"powershell.exe"`).
- `packages/utils/src/command-executor/execute-hook-command.ts`: hook tree-kill spawn uses the resolver (was bare `"taskkill"`).
- `packages/lsp-core/src/lsp/process.ts`: `killProcessTree` spawnSync uses the resolver (was bare `"taskkill"`).
- `packages/utils/src/index.ts`: barrel export for the new module.
- `packages/utils/src/system-tool-paths.test.ts` (new): 5 unit cases (custom SystemRoot, no-SystemRoot fallback, ambient default).

## Verification

- `bun test packages/utils/src/system-tool-paths.test.ts` — 5/5 pass.
- `tsgo --noEmit -p packages/utils/tsconfig.json` and `-p packages/lsp-core/tsconfig.json` — clean.
- `bun test packages/utils` — 493 pass / 9 fail; the 9 failures (codegraph provision downloads, Bun-global runtime shims) are pre-existing on `upstream/dev` and environment-dependent (verified by stash comparison).
- Rebuilt `dist/index.js` (main runtime bundle) and `packages/lsp-daemon/dist/*` locally; both contain the resolved absolute path.

## Notes

[...]

PR #6850, opened 2026-08-13
fix(memory): hide supervised child windows on Windows

grim-susemi:
## Summary

- Hide every memory supervisor child-process layer on Windows: session→supervisor, supervisor→bootstrap, bootstrap→model, and the taskkill/pkill termination spawns.
- Keep the bootstrap attached on Windows while preserving POSIX detached process-group isolation.
- Pin the platform-specific spawn and abrupt-supervisor-exit contracts.

## Changes

- Small pure spawn-option builders used by the production spawn calls (`modelChildWindowOptions`, `bootstrapProcessOptions`, `supervisorProcessOptions`).
- `windowsHide: true` for model, bootstrap, supervisor, and termination (`supervisor-process-identity.ts` taskkill/pkill) children.
- Bootstrap `detached` only for POSIX.
- Update IC8 containment expectations: Windows attached bootstrap exits with its supervisor; POSIX detached bootstrap continues enforcing the persisted deadline.
- Fix a stale `liveProcesses` reference in the IC8 test (renamed to the harness's `untrackProcessGroup`).
- Extension bundles rebuilt on Linux with bun 1.3.14 (cross-platform digest parity).

## QA & Evidence

- RED: `bun test .../supervisor-spawn-options.test.ts` — 0 passed, 3 failed on the unchanged spawn policy.
- GREEN: spawn contract 4 passed; IC8 containment 9 passed.
- Installed-runtime QA: actual supervisor runs produce no new `WindowsTerminal.exe`/`OpenConsole.exe` window.

Related: #6973 covers the omo launcher window (split out of this PR's scope); code-yeongyu/senpi#927 covers the engine-internal spawns.


code-yeongyu:
Thanks — the spawn-option contract and the IC8 policy pin both look right, and #6846 removing the parent's dependence on supervisor `close` means the Windows `detached: false` switch doesn't conflict with durable-outcome completion.

One thing to double-check before merge: `memory-run-supervisor.ts` ships as a generated bundle (`memory-run-supervisor.mjs`), and #6846 regenerated it as part of the PR. This PR changes the supervisor source but I don't see the regenerated bundle in the diff — the `senpi-compatibility` bundle-current check should catch it, but if it goes red that's the reason, not the logic change.

code-yeongyu:
소스 변경으로 stale이 된 번들, 레포 고정 bun 1.3.14로 재생성해서 브랜치에 푸시핒습니다 (34acaf7) — freshness 체크 통과 확인했고 CI 다시 돌고 있습니다.

code-yeongyu:
CI triage: same failure mode as your #6991 — `senpi-compatibility (ubuntu-latest)` fails at the **"Verify committed Senpi plugin bundle is current"** step: this PR changes `omo-senpi` sources (`memory-run-supervisor.ts`, `supervisor-spawn-options.ts`) without the matching regenerated `packages/omo-senpi/plugin/extensions/*.js`.

Same caveat as #6991: rebuild the bundle on **Linux** (repo-pinned bun 1.3.14) — Windows rebuilds produce byte-different output and will fail the check again. Alternatively push just the source and we can regenerate maintainer-side, as was done earlier on this branch.

code-yeongyu:
Same note as on #6991 / #6740 / #6973: `91825acc` duplicates the memory-usage ledger fix already landing via #7026, and the bundle rebuild `86f9268e` is what's failing the ubuntu bundle-verify step. Please drop both commits and rebase on `dev` after #7026 merges — this PR then stays scoped to the supervised child window fix.

code-yeongyu:
Superseded by #7082, which merged as `9e17d1631`.

Both PRs fixed the same bug — supervised memory-worker children allocating a console window on win32 — by different means. #7082 sets `windowsHide: true` inline at all five spawn sites in the reflection chain and, critically, regenerates the shipped bundles under `packages/omo-senpi/plugin/extensions/` so the flag actually reaches the code that runs on Windows. It also adds a source audit test that pins the count of unhidden spawn sites, so a newly added spawn cannot silently reintroduce the flash.

dev now already carries `windowsHide` at the sites this PR targets (`memory-run-supervisor.ts`, plus the supervisor identity/spawn paths), so the change here would be a no-op.

The one idea in this PR that #7082 does not have is the `supervisor-spawn-options.ts` helper centralizing spawn-option construction instead of repeating the flag at each call site. That is a genuine improvement — if you want to pursue it, a follow-up PR doing just that refactor on top of current dev (no behavior change, with the existing audit test as the guard) would be welcome.

Remaining uncovered spawn sites are handled in #7102.

Issue #6852, opened 2026-08-13
bug(memory): GitMemoryRepo can lose a commit under worktree lock contention (ubuntu-latest, 19/20)


code-yeongyu:
## Summary

`GitMemoryRepo > #given concurrent commits across worktrees #when index and ref locks contend #then every commit lands` failed on **ubuntu-latest** in run [31748186369](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31748186369) (PR #6844, which touches only `omo-opencode` tool code — no memory/repo changes), failing identically on both in-run attempts at 112ms.

## Why this is filed separately from #6843

- #6843 tracks wall-clock **timeout** flakes clustered at the ~5s boundary on loaded runners.
- This is a **fast assertion failure** (112ms) in a lock-contention correctness test — the test asserts that every concurrent worktree commit lands, and it did not.
- A lost commit under index/ref lock contention on ubuntu-latest is a potential correctness bug in the memory repo's locking, not runner slowness.

(Caveat: run logs became unavailable after a rerun started, so I could not capture the exact assertion output; the failure line and 112ms duration are from the pre-rerun failed-job log. If the rerun surfaces the full assertion diff, it should be attached here.)

## Suggested direction

- Instrument the contention path (retry counts, lock acquisition order, which worktree failed) so a CI hit produces actionable output.
- Run this test with a high iteration count in a scheduled/stress job to measure the actual race rate on linux runners.

code-yeongyu:
Second CI occurrence: same test failed on ubuntu-latest in run https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31732452566 (PR #6841, feat/6657-profile-cli — CLI-only changes, no memory code). 102ms assertion failure, identical on both in-run attempts. Two distinct linux hits on unrelated branches within ~2.5 hours; race rate is high enough that a stress-run reproduction should be practical.

code-yeongyu:
Captured a full assertion log for this failure class before rerunning, from a different branch (feat/6657-profile-cli, run [31732452566](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31732452566), ubuntu-latest, 102ms — also branch-independent):

```
GitCommandError: git worktree add -b memory/concurrent-4 /tmp/memory-worktrees-vEBoXt/checkout-4 HEAD failed: Preparing worktree (new branch 'memory/concurrent-4')
fatal: failed to read .git/worktrees/checkout-5/commondir: Success
[...]

Issue #6843, opened 2026-08-13
test: CI timeout flakes clustering on windows-latest/macos-latest memory tests (3 consecutive runs)


code-yeongyu:
## Summary

Three consecutive CI runs failed tonight (2026-08-13 ~13:30–13:57 UTC) with test timeouts that are unrelated to the branches' changed files. Each failing test failed identically on both attempts within its run, then the run-level reruns are being used as roulette.

## Evidence

| Run | Branch | Runner | Failing test | Timeout |
|-----|--------|--------|--------------|---------|
| [31705206426](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31705206426) | fix/6772-platform-neutral-mkdir (docs-only) | windows-latest | `createInitDeepAdvisorComponent > #given fake timers #when session_start fires #then ui.select waits for setTimeout zero` | 6032ms x2 |
| [31705212373](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31705212373) | fix/6808-memory-quota-fallback | windows-latest | `GitMemoryRepo > #given commits with bodies and changed paths #when log reads history #then …` | 5015ms x2 |
| [31705219036](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/31705219036) | fix/6624-trust-provisioned-codegraph | windows-latest + macos-latest | `/memory > #given a dirty working tree #when invoked #then uncommitted changes render in a separate section` (win, 5031ms x2); `one memory identity across real Bun processes > #given a writer SIGKILLed while holding the shared lock #when its peer continues #then liveness takeover completes twenty clean commits` (macOS, 1934ms x2) |

## Observations

- All four distinct tests are timeout failures, three of four are memory-subsystem tests doing real git/process work.
- The failures reproduce deterministically within a run (both attempts fail with identical durations) but are branch-independent — the docs-only branch hit one too.
- Recent history shows a cluster of `fix/windows-senpi-*` stabilization branches, so runner-load-induced timeouts on windows-latest look systemic rather than test-specific.

## Suggested direction

- Audit timeout budgets for memory worker/GitMemoryRepo/`/memory` tests on windows runners (5s wall clock appears too tight under CI load).
- Consider marking the real-process liveness-takeover test as serial or giving it a dedicated shard on macOS.
- Track flake rate before/after instead of rerunning failed runs manually.

code-yeongyu:
[...]

PR #6853, opened 2026-08-13
fix(memory): reflection dies on a sandbox grant no writer has created

ismetanin:
## Bottom line

Every memory reflection run dies before its child starts, on any identity whose runtime directories predate a layout addition:

```
memory reflection failed
run:reflection-run-3 category:quick
detail:ENOENT: no such file or directory, lstat '<identity>/runtime/reflection-sessions'
```

The run directory is left with no `child-stderr.log`, because the failure happens while building the sandbox, before the child is spawned. The remediation hint then blames the senpi executable, which sends the user after `SENPI_BIN` for a problem that is a missing directory.

## Root cause

`buildPathSandboxTransform` resolves every writable grant through `realpathSync`, which throws `ENOENT` for a path that does not exist. `identity-runtime.ts` grants `identityPaths.reflectionSessions`, and nothing guarantees that directory exists at spawn time:

- `ensureIdentityRuntimeDirs` creates all eleven runtime subdirectories, but it is the first-write seam and runs on a memory write.
- Reflection is triggered by step count, so it can run before any memory write in a session.
- An identity created under an earlier `RUNTIME_SUBDIRNAMES` keeps only the subdirectories that list had. On the affected machine the four newest entries (`reflection-sessions`, `viewers`, `push-queue`, `tool-receipts`) were all absent while the seven older ones were present.

`runtime/reflection-sessions` is additionally a stale grant: the reflection session directory in use is `runtime/reflection/runs/<runId>` (`reflection-spawn-input.ts`), already covered by the `reflection` grant. The crash came from a path the code never writes to.

## Fix

- A writable grant names a directory, so it is resolved through creation (`canonicalWritableDir`) rather than assumed to exist. Payload paths keep the plain resolver: they are files, and running `mkdir` on one fails with `EEXIST`. That split is load-bearing, and the existing sandbox tests catch it if the creation is applied to both.
- `reflectionRemediation` reports the path an `ENOENT` names. The executable hint stays for a bare spawn failure such as `spawn senpi ENOENT`.

## Tests

- `sandbox.test.ts`: a granted runtime directory that no writer has created is created and appears in the generated seatbelt profile, instead of throwing out of the builder.
- `worker/remediation.test.ts`: an `ENOENT` carrying a path names that path and does not mention `SENPI_BIN`; a bare spawn failure still recommends `SENPI_BIN`; the model-visibility hint keeps priority.

Both fail on `dev` and pass with the fix.

`bun test packages/omo-senpi`: 1519 pass, 0 fail. `tsgo --noEmit` clean. All four plugin `--check` gates current, and the rebuilt bundles ride in the usual separate `chore(senpi): rebuild extension bundles` commit.

## Not fixed here

An identity created under an older layout stays partially populated until something takes the first-write seam. Backfilling on identity load would need to relax the "reads never create identity storage" invariant, which is a deliberate design decision and belongs in its own change.

---
## Summary by cubic
Prevents reflection from dying before spawn when a writable sandbox grant points at a missing runtime dir, and fixes path and remediation handling. Previously all writable grants used realpath and crashed with ENOENT; now we create only owned runtime dirs, canonicalize paths robustly, and surface the exact missing path.

- Sandbox: resolve the inner command before any filesystem changes; create only `runtimeWrites` (0700) and treat structural and env-supplied grants (`worktree`, `.git`, `SENPI_CODING_AGENT_DIR`, `XDG_CONFIG_HOME`) as must-exist; unsandboxed fallback no longer creates external dirs. Replace realpath with component-wise lstat/readlink (40-link bound) to canonicalize legal paths, including POSIX paths with backslashes; raise a clear incompatibility if a path exists but cannot be inspected.
- Remediation and health: capture structured errno fields (code/syscall/path) at failure and persist them; `reflectionRemediation` prioritizes category hints, then reports the missing path for ENOENT, and only falls back to the `SENPI_BIN` hint for bare spawn failures. Plumb `failureError` through completion records, health, ledger, and alerts.
- API/tests/build: `buildSandboxTransform` accepts `externalWritableDirs` and treats `runtimeWrites` as owned; facts sandbox creates its run dir as owned. Add tests for owned grant creation, strict handling of external/structural grants, symlink/backslash canonicalization, and exact path round-tripping. Rebuild `@senpi` extension bundles; adjust the seatbelt profile assertion to match its JSON-encoded path.

<sup>Written for commit 397de21397957ee60bc86ade2dbc01791c213020. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6853?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
CI note: the Windows failures of the new `sandbox.test.ts` case are an assertion-encoding issue, not a fix regression. The generated profile **does** contain the created `reflection-sessions` grant — but profile paths are backslash-escaped (`C:\\Users\\...`), while the test does `expect(profile).toContain(realpathSync(absentGrant))` with the raw path (`C:\Users\...`), so the substring never matches on Windows. POSIX paths have no backslashes, which is why only windows-latest fails.

Comparing against the same escaping used when the profile is rendered (or asserting on the parsed grant list instead of the raw profile string) should make the test platform-neutral. The `GitMemoryRepo`/`payload buffer` failures in the same run are pre-existing runner flakes (#6843, #6852) — unrelated to this PR.

code-yeongyu:
The encoding fix (fa3185f) works — `reflection worker OS sandbox writable grants` now **passes on windows-latest** in both jobs of run 31789958081.

The remaining Windows failure is a real interaction with this PR's fix, though: `createMemoryComponent > #given enabled memory #when session_start binds an auto identity #then it appends a hidden binding and performs no filesystem writes` fails at `expect(existsSync(memoryHome)).toBe(false)` (index.test.ts:146, 16ms, deterministic x2). That test pins the "reads never create identity storage" invariant you called out in *Not fixed here* — but `canonicalWritableDir` resolving grants through creation now appears to create identity runtime directories on a path that previously performed no writes. Worth checking whether the sandbox build (or something it calls) is now running during session_start binding, and gating directory creation to actual spawn-time sandbox construction.

(`OmO Native payload buffer` timeouts in the same run are the known runner flakes — #6843.)

packages/omo-senpi/src/components/memory/sandbox-platform.ts
```diff
@@ -43,7 +43,7 @@ export function buildPathSandboxTransform<T extends ReflectionSpawnArgs | FactsS
     return identityTransform(`${input.surface} sandbox unavailable on ${platform}: ${reason}; running unsandboxed because policy is auto`)
   }
 
-  const writableDirs = input.writableDirs.map(canonicalPath)
+  const writableDirs = input.writableDirs.map(canonicalWritableDir)
```
code-yeongyu: [sisyphus-bot] [P1] Do not create every writable grant in this generic canonicalization pass.

The production list is not limited to the missing internal runtime directory: it also contains the worktree root, Git common directory, full `SENPI_CODING_AGENT_DIR`, and full `XDG_CONFIG_HOME`. On the exact head, three absent grants (worktree, Git common, external config) changed from `[false,false,false]` to `[true,true,true]` merely by building the transform. This also runs before `resolveInnerCommand()`, so the function can create external directory trees and then return an unsandboxed identity transform because the inner command is unavailable.

Split the contract: create only explicitly component-owned runtime paths at their normal spawn/write seam; require structural and environment-supplied grants to exist and canonicalize without mutation. Resolve the inner command before any filesystem side effect. The stale `reflectionSessions` grant that motivated this change can be removed instead of materialized.

packages/omo-senpi/src/components/memory/sandbox-platform.ts
```diff
@@ -119,6 +119,17 @@ function canonicalPath(path: string): string {
...
 
+/**
+ * A writable grant names a directory, and the layout can add one before any writer has created it, so
+ * the grant is resolved through creation. Resolving a missing path instead throws ENOENT out of the
+ * sandbox builder, which reads as a spawn failure and hides which directory was actually absent.
+ * Payload paths are files and are never created here.
+ */
+function canonicalWritableDir(path: string): string {
+  mkdirSync(path, { recursive: true, mode: 0o700 })
+  return realpathSync(path)
```
code-yeongyu: [sisyphus-bot] [P1] This still throws for a legal POSIX path under the shipped Bun runtime.

A literal backslash is valid in macOS/Linux filenames. Exact-head proof used `.../back\slash/work`: `mkdirSync()` created it and `existsSync()` returned true, but this `realpathSync()` threw `ENOENT: ... lstat '.../back\slash/work'`; the sandbox build failed with the same error before Seatbelt or bwrap ran. `OMO_MEMORY_HOME`, `SENPI_CODING_AGENT_DIR`, and `XDG_CONFIG_HOME` can all feed this path.

Add a regression using a real literal-backslash component and use a Bun-compatible canonicalization that still resolves symlinks. At minimum, emit an explicit runtime/path incompatibility instead of claiming the directory is absent after creating it.

packages/omo-senpi/src/components/memory/worker/remediation.ts
```diff
@@ -7,6 +7,10 @@ export function reflectionRemediation(reason: string | undefined, detail: string
   if (combined.includes("model-not-found") || combined.includes("model_not_visible") || combined.includes("model not found")) {
     return "the reflection child cannot see the configured category model; adjust memory.reflection category/model in your omo config"
   }
+  const missingPath = /ENOENT[^']*'([^']+)'/i.exec(`${reason ?? ""} ${detail ?? ""}`)
```
code-yeongyu: [sisyphus-bot] [P2] Do not recover `error.path` with a single-quote regex.

Node/Bun filesystem messages wrap the path in single quotes without escaping apostrophes inside the filename. On this head, `ENOENT: ... lstat '/tmp/O'Brien/missing'` produces `the reflection child could not use /tmp/O`, which is a different path.

Preserve structured `code`, `syscall`, and `path` fields through the failure record and render the real path directly. If legacy persisted strings still need parsing, cover apostrophes, newlines, backslashes, Unicode, and platform-specific message shapes.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST_CHANGES

The ordinary missing-directory case is fixed, but the generic creation strategy introduces two reproducible boundary failures and the new remediation misreports valid paths.

1. [P1] `writableDirs.map(canonicalWritableDir)` creates every missing grant, not only the stale internal runtime grant. Exact-head proof changed absent worktree, Git common, and external-config paths from `[false,false,false]` to `[true,true,true]`. Creation also happens before inner-command resolution, so a missing Senpi command can leave those directories behind and then run unsandboxed.
2. [P1] The shipped Bun runtime cannot `realpathSync()` a legal POSIX path containing a literal backslash. The helper created `back\slash/work`, `existsSync()` returned true, then both the direct canonicalization and sandbox build threw ENOENT. This PR still fails before child spawn for that valid path class.
3. [P2] The new ENOENT regex truncates legal apostrophe paths. `/tmp/O'Brien/missing` is reported as `/tmp/O`, sending the user to the wrong location.
4. [P1 gate] The exact current head is still red in `test (windows-latest)` and `senpi-compatibility (windows-latest)` in run 31789958081. Existing owner comment 5292185141 records a deterministic failure of the no-filesystem-writes invariant; there is no later commit on the PR head. I am not duplicating that existing comment inline, but it remains unresolved.

Focused sandbox/remediation tests pass (19 tests, 44 assertions), the real Darwin seatbelt case passes, and bundle currency is clean. Those tests do not cover the retained path and side-effect boundaries. Please scope creation to explicitly component-owned runtime directories at their lifecycle seam, preserve strict validation for structural/external grants, canonicalize all legal POSIX path names under the shipped runtime, and carry structured filesystem error fields into remediation.

packages/omo-senpi/src/components/memory/sandbox-platform.ts
ismetanin: Reproduced and fixed in 397de213.

I confirmed your finding first: three absent grants (worktree, git common, external config) went `[false,false,false]` -> `[true,true,true]` merely by building the transform.

The contract is now split as you described:
- `resolveInnerCommand()` runs **before** any filesystem mutation, so the unsandboxed-fallback path can no longer create external directory trees on its way out.
- Structural and environment-supplied grants (worktree, git common dir, `SENPI_CODING_AGENT_DIR`, `XDG_CONFIG_HOME`) are required to exist and are canonicalized without mutation.
- Only explicitly component-owned `runtimeWrites` are created, threaded as a separate `ownedWritableDirs` input rather than inferred from path strings.

Regressions added: absent structural grants fail without creating either; an absent environment-supplied grant fails without creating it; and with a missing inner command no grant is created before the unsandboxed fallback. The originally motivating case still passes — a missing component-owned `reflection-sessions` is created at the lazy spawn seam instead of killing the run.

packages/omo-senpi/src/components/memory/sandbox-platform.ts
ismetanin: Reproduced and fixed in 397de213.

Confirmed exactly as you described: `mkdirSync` created `.../back\slash/work`, `existsSync()` returned true, and `realpathSync()` threw `ENOENT ... lstat '.../back\slash/work'`.

Canonicalization no longer goes through the shipped Bun `realpathSync`. It now resolves component by component with `lstatSync`/`readlinkSync` under a 40-link bound, so symlinks are still fully resolved — I kept that deliberately, since collapsing symlinks is security-relevant for a sandbox grant and dropping it would have been the easy wrong fix. A path that exists but cannot be inspected now raises an explicit runtime/path incompatibility instead of claiming the directory is absent after creating it.

Regression added using a real literal-backslash component behind a symlink, asserting the legal path resolves to its symlink target.

packages/omo-senpi/src/components/memory/worker/remediation.ts
ismetanin: Reproduced and fixed in 397de213.

Confirmed: `ENOENT: ... lstat '/tmp/O'Brien/missing'` rendered as `the reflection child could not use /tmp/O`.

Structured `code`, `syscall` and `path` are now captured at the original failure and carried through the durable failure/completion/health records, and remediation renders the real path directly instead of scraping the message. Legacy persisted strings are still parsed for older records, now covering apostrophes, newlines, backslashes and Unicode.

Verified `/tmp/O'Brien/missing` round-trips exactly, both through the structured fields and through the legacy string path.

ismetanin:
All three code findings are fixed in 397de213, each reproduced first so the fix is measured against the real failure rather than my reading of it. Details are in the inline replies; summary:

1. **[P1] Unscoped creation** — reproduced (`[false,false,false]` -> `[true,true,true]`). Creation is now scoped to explicitly component-owned `runtimeWrites`; structural and environment-supplied grants must exist and are canonicalized without mutation; the inner command resolves before any filesystem side effect.
2. **[P1] Literal-backslash paths** — reproduced. Canonicalization now walks components via `lstat`/`readlink` under a 40-link bound instead of Bun's `realpathSync`, so symlinks are still resolved and a legal backslash path no longer dies before spawn.
3. **[P2] Apostrophe truncation** — reproduced (`/tmp/O` for `/tmp/O'Brien/missing`). Structured `code`/`syscall`/`path` now flow through the durable records and remediation renders the path directly; the legacy string parser handles apostrophes, newlines, backslashes and Unicode.

Verification: 1579 tests pass in `packages/omo-senpi` (0 fail), `tsgo` clean, bundles rebuilt and `build-extension.mjs --check` reports current.

---

**On the `createMemoryComponent` "performs no filesystem writes" failure — I have to disagree with the attribution, with evidence.**

That failure is not caused by `canonicalWritableDir`. On Windows it is unreachable: `resolveExecutable()` returns `undefined` ("platform is unsupported"), so `buildPathSandboxTransform()` returns the identity transform at `sandbox-platform.ts:43`, and the `writableDirs.map(...)` call sits at line 46 — after that return. The sandbox is also built lazily (`identity-runtime.ts:78` `lazySandbox`), so it is not invoked during `session_start` binding on any platform.

The actual writer is `mkdir()` at `packages/memory-core/src/locks/acquire.ts:65`, reached from an unawaited `void wiring.afterBind(...)` at `memory/index.ts:135`:

`afterBind` -> `runtimeFor(identity).reconcile()` -> `reconcileReflectionRuns()` -> `reservation.readState()` -> `locked()` -> `withLock()` -> `mkdir(dirname(lockPath), { recursive: true })`

The lock path is `OMO_MEMORY_HOME/agents/<identity>/runtime/locks/reflection-scheduler.lock`, so a *read* creates identity storage. Windows loses the race because `getProcessStartIdentity()` returns `null` immediately there (`locks/process-identity.ts:35-41`), while Linux waits on `/proc/<pid>/stat` and macOS spawns `/bin/ps` — that I/O is what usually leaves reconciliation pending until after the assertion.

Two independent confirmations that this is a pre-existing race and not this PR:
- In run 31789958081, **the same commit** passed this test in `test (windows-latest)` and failed it in `senpi-compatibility (windows-latest)`. Identical code cannot deterministically fail in one job and pass in the other.
- Neutering `canonicalWritableDir` back to a plain `realpathSync` leaves the test's behavior unchanged.

`FactsQueue.listPending()` has the same shape (`packages/memory-core/src/facts/queue.ts:138-142` creates dirs on a read path), so fixing only the reservation read would leave a second creator.

The minimal correct fix is to make those reads non-creating when the store is absent, rather than adding a timer or a platform skip that would just hide the race. That is a real latent bug in the invariant you flagged, but it is outside this PR's diff — happy to open it separately, or to fold it in here if you prefer.

code-yeongyu:

I'm closing it as **superseded**: the same pre-spawn `ENOENT … lstat '<identity>/runtime/reflection-sessions'` root cause was fixed on `dev` by `12c176e02` (*fix(omo-senpi): canonicalize sandbox paths that do not exist yet*, 2026-08-19) — it resolves only the existing prefix through the filesystem and appends the absent tail verbatim, with `sandbox-absent-paths.test.ts` covering it. Your report and reproduction were spot on; thank you for the careful review follow-ups.

**How I checked:** `git log -1 12c176e02` names `runtime/reflection-sessions` and the pre-spawn lstat failure explicitly; the commit is an ancestor of `origin/dev`.

Issue #5701, opened 2026-06-28
Subagent model overrides cannot be overridden by user config — plugin API takes precedence over opencode.json agent section


somehaveseen:
## Problem

Oh My OpenAgent (v4.13.0) registers subagents with hardcoded model assignments via the plugin API. Users cannot override these models through `opencode.json` `"agent"` section because plugin-registered agents take precedence over static config.

This means:
- `explore` subagent is forced to `opencode/qwen3.5-plus`
- `librarian` subagent is forced to `opencode/qwen3.5-plus`
- `oracle` subagent is forced to `opencode/gpt-5.5`
- `multimodal-looker` forced to `opencode/gpt-5.5`
- `Momus - Plan Critic` forced to `opencode/gpt-5.5`
- `Metis - Plan Consultant` forced to `opencode/claude-sonnet-4-6`
- `Prometheus - Plan Builder` forced to `opencode/claude-opus-4-7`
- `Hephaestus - Deep Agent` forced to `opencode/gpt-5.5`

Users who want all subagents to inherit the parent model (e.g., `opencode/big-pickle`) cannot achieve this. Adding `"agent": { "explore": { "model": "opencode/big-pickle" } }` to `~/.config/opencode/opencode.json` has no effect — `opencode debug config` still shows the OMO-assigned model.

## Steps to Reproduce

1. Install OMO v4.13.0 as an opencode plugin
2. Check subagent models: `opencode debug config | grep -A5 explore`
3. Add agent model override to `~/.config/opencode/opencode.json`:
   ```json
   {
     "agent": {
       "explore": { "model": "opencode/big-pickle" }
     }
   }
   ```
4. Run `opencode debug config` again — model is still `opencode/qwen3.5-plus`

## Expected Behavior

User config `"agent"` overrides should take precedence over plugin-registered agents, OR OMO should provide a config mechanism (e.g., `oh-my-openagent.json`) to customize per-agent models.

Ideally, subagents should inherit the parent model by default unless explicitly overridden. Users who want different models for specific subagents should be able to set them via opencode config.

## Actual Behavior

Plugin-registered agent models are locked and ignore user config overrides.

## Environment

- OMO version: 4.13.0
- OpenCode version: 1.17.11
- Config: `oh-my-openagent` in plugin list, no per-agent model customization in `oh-my-openagent.json`

## Suggestion

Options to fix (any would work):

[...]

Issue #3228, opened 2026-04-08
Bug: task(subagent_type) ignores agent model config, inherits parent session model


nickveenhof:
### Environment
- oh-my-openagent: v3.15.3
- opencode: latest
- Provider: GitLab (duo-chat)

### Description

When using `task(subagent_type="technical-writer")` to spawn a named agent, the agent runs on the parent session's model instead of the model configured in `oh-my-openagent.jsonc`.

`call_omo_agent` and `task(category)` correctly respect model config. Only `task(subagent_type)` inherits the parent model.

### Reproduction

**Config (`oh-my-openagent.jsonc`):**
```jsonc
"scribe": {
  "model": "gitlab/duo-chat-sonnet-4-6",
  "prompt_append": "file://./agents/technical-writer.md"
},
"sentinel": {
  "model": "gitlab/duo-chat-haiku-4-5",
  "prompt_append": "file://./agents/fullstack-engineer.md"
}
```

**Parent session model:** `gitlab/duo-chat-opus-4-6`

**Test calls and results:**

| Tool | subagent_type | Expected model | Actual model |
|---|---|---|---|
| `call_omo_agent` | `librarian` | sonnet-4-6 | sonnet-4-6 ✅ |
| `call_omo_agent` | `explore` | haiku-4-5 | haiku-4-5 ✅ |
| `task(category)` | `quick` | haiku-4-5 | haiku-4-5 ✅ |
| `task(category)` | `deep` | sonnet-4-6 | sonnet-4-6 ✅ |
| `task(subagent_type)` | `technical-writer` | sonnet-4-6 | **opus-4-6** ❌ |
| `task(subagent_type)` | `fullstack-engineer` | haiku-4-5 | **opus-4-6** ❌ |
| `task(subagent_type)` | `social-media-mgr` | sonnet-4-6 | **opus-4-6** ❌ |
| `task(subagent_type)` | `ux-designer` | sonnet-4-6 | **opus-4-6** ❌ |
| `task(subagent_type)` | `quality-reviewer` | sonnet-4-6 | **opus-4-6** ❌ |
| `task(subagent_type)` | `analytics-insights-mgr` | sonnet-4-6 | **opus-4-6** ❌ |

All 6 `task(subagent_type)` agents report the parent session model (opus-4-6) instead of their configured model.

### Root cause (likely)

[...]

PR #6874, opened 2026-08-15
fix(delegate-task): keep configured primary model when fallbacks present (#6869)

### LINKED ISSUE #6869: Bug: fallback_models/models on an agent makes task() subagent ignore the configured primary model
### Environment
- oh-my-openagent: 4.19.4
- opencode: 1.18.18
- Provider: opencode-go

### Description

The delegate-task subagent resolution (`resolveSubagentModel` in `packages/omo-opencode/src/tools/delegate-task/subagent-model-resolution.ts`) reads only `agentOverride.model` and `agentOverride.fallback_models` from the agent config. It never reads the schema-valid `models` array. Worse, the mere **presence** of a `fallback_models` array makes the resolver ignore the configured primary `model` entirely and fall back to the plugin default (`AGENT_MODEL_REQUIREMENTS`).

### Reproduction (5 controlled runs, all via `task(subagent_type="explore")`)

Config in `~/.omo/omo.jsonc`, then restart ai-dev and spawn an `explore` subagent through the task tool:

| Config | Subagent model used | Expected |
|---|---|---|
| `{"model": "opencode-go/kimi-k3"}` | `opencode-go/kimi-k3` ✅ | configured primary |
| `{"model": "opencode-go/kimi-k3", "variant": "max"}` | `opencode-go/kimi-k3` ✅ | configured primary |
| `{"model": "opencode-go/kimi-k3", "fallback_models": [{"model": "opencode-go/qwen3.7-plus"}]}` | `opencode-go/qwen3.7-plus` ❌ | `kimi-k3` |
| `{"model": "opencode-go/kimi-k3", "fallback_models": [{"model": "opencode-go/kimi-k3"}]}` | `opencode-go/qwen3.7-plus` ❌ | `kimi-k3` |
| `{"model": "opencode-go/kimi-k3", "models": [{"model": "opencode-go/kimi-k3"}, {"model": "opencode-go/qwen3.7-plus"}]}` | `opencode-go/qwen3.7-plus` ❌ | `kimi-k3` |

The configured model is present in the connected-provider cache, so this is not a provider-availability issue.

### Root cause (from source)

In `resolveSubagentModel`, `agentOverride?.model` is used as `userModel` and `agentOverride?.fallback_models` as the fallback chain. When `fallback_models` is present, `resolveModelForDelegateTask` enters the fallback-promotion branch (`primaryMatch` / `fuzzyMatchModel`), and the effective model comes from the fallback/plugin-default path instead of the primary. The `models` array (the 4.19.4-documented replacement for `fallback_models`, per the doctor "Replace fallback_models with models" suggestion) is never read by the delegate-task path at all.

### Expected behavior

1. A configured primary `model` should always win for `task(subagent_type=...)` subagents, independent of whether `fallback_models`/`models` are also present.
2. The `models` array should be honored as the chain (primary + fallbacks), or clearly documented as not supported by the delegate-task path.

Related: #3228 (subagent ignores agent model config), #5701 (plugin-registered agents take precedence over static config).

EvangelosMoschou:
## Summary

Fixes #6869: `task(subagent_type=...)` subagents ignored the configured primary `model` whenever `fallback_models` (or the newer `models` array) was present. A configured primary was demoted to a fallback whenever `fuzzyMatchModel` returned null for it — but fuzzy-null is "inconclusive" (format/alias/normalization edges), not "unavailable". Explicit user intent now always wins.

## Changes

- **`packages/delegate-core/src/model-selection.ts`** — an explicit `userModel` is returned unconditionally. The fallback-promotion branch engages only when the primary is *verifiably absent* from the available set (exact-string check), not when fuzzy matching is merely inconclusive.
- **`packages/omo-opencode/src/tools/delegate-task/subagent-model-resolution.ts`** — the agent delegate-task path now reads the schema-valid `models` array (primary = `models[0]`, fallbacks = `models.slice(1)`), matching the existing category-resolver behavior. `models` is the documented replacement for `fallback_models` (doctor: "Replace fallback_models with models"); previously it was silently ignored by this path.
- **Tests** — two new delegate-core tests pin the regression from the issue's reproduction table: (1) primary present in the available set + fallbacks present → primary kept; (2) primary fuzzy-inconclusive (provider alias mismatch) + fallbacks present → explicit primary still wins.

## QA & Evidence

- **What was tested:** `bun test packages/delegate-core/src/model-selection.test.ts` — **11 pass / 0 fail** (9 existing + 2 new). `bun test packages/delegate-core/` — **12 pass**. `bun test packages/omo-opencode/src/tools/delegate-task/` — **483 pass** (models-array wiring doesn't regress category/agent resolution). `bun run typecheck` + `bun run build` — **exit 0**.
- **Artifact:** `.omo/evidence/20260814-issue-6869-delegate-primary-model/README.md`
- **Why sufficient:** the new tests reproduce the exact issue table (model + fallback_models → primary must win), including the alias-mismatch case that makes `fuzzyMatchModel` inconclusive.
- **What was omitted:** a live opencode run with a real connected-provider cache; the logic is fully covered deterministically.

## Risks & Residuals

- **Fallback promotion still works** when the primary is genuinely absent (the existing test `user primary is unreachable → promotes fallback` still passes). Only the inconclusive case changed.
- The `models` array on the agent path now takes precedence over `fallback_models` when both are set (consistent with category resolution). Documented behavior.

## Related Issues

Fixes #6869. Related: #3228, #5701.

---
## Summary by cubic
Keeps the configured primary model for delegate-task subagents when fallbacks exist (fixes #6869), and now also honors the agent `models` array. Previously the primary was demoted whenever fuzzy matching returned null; now a fallback is promoted only when the primary is verifiably absent — no exact match and no model-ID match under any provider prefix. Object entries in `models` keep their per-model settings through resolution.

- `packages/delegate-core/src/model-selection.ts`: Return the explicit primary unless it is verifiably absent, gating fallback promotion behind the exact-match and provider-prefix absence checks.
- `packages/omo-opencode/src/tools/delegate-task/subagent-model-resolution.ts`: Read `models` with `models[0]` as primary and `models.slice(1)` as fallbacks; `models` wins over `fallback_models` when both are set, and object entries preserve variant, reasoning, reasoningEffort, temperature, top_p, maxTokens, and thinking.
- Tests: Pin the primary kept with fallbacks present, primary kept when fuzzy matching is inconclusive, fallback promoted only on verifiable absence, and the full `models[0]` object contract surviving resolution.

**Rollout**
- Use the `models` array: put the primary at index 0 and fallbacks after it. When both `models` and `fallback_models` are set, `models` is used.

<sup>Written for commit 1dbc86e95206f9ce3e4b17629507341d38bd994a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6874?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
The patch still demotes the configured primary in the exact inconclusive case it claims to fix.

When `primaryMatch` is null and the primary string is not an exact member of `availableModels`, the loop still promotes any reachable fallback:
https://github.com/code-yeongyu/oh-my-openagent/blob/c67340aa2213d81a61ed2e637acefb3d839d7e19/packages/delegate-core/src/model-selection.ts#L93-L132

The new fuzzy-inconclusive test avoids that path only because its fallback also cannot match. Add a regression with an alias-inconclusive primary and an exactly reachable fallback; the explicit primary must remain selected unless unavailability is positively established.

The new `models` support also extracts only `entry.model` from the primary object, discarding its `variant`, `reasoning`, effort, temperature, token, and thinking settings. Preserve the complete `models[0]` contract and add object-primary plus mixed fallback regressions:
https://github.com/code-yeongyu/oh-my-openagent/blob/c67340aa2213d81a61ed2e637acefb3d839d7e19/packages/omo-opencode/src/tools/delegate-task/subagent-model-resolution.ts#L20-L90

Run isolated real OpenCode QA with a connected-provider cache that reproduces #6869 and asserts the actual delegated session model/options. Then update the stale branch, rerun the full required matrix, and keep #6869 open until that behavior is green.

EvangelosMoschou:
## Review-2 feedback addressed — rebased onto latest dev

Both findings from the 2026-09-01 review are implemented in `9b025236f` + `f5f61bd1a` (rebased onto current `dev`):

1. **Inconclusive primary no longer demoted** — `resolveModelForDelegateTask` now requires *positive absence* before promoting a fallback: the primary must be neither an exact member of the available set NOR present under any provider prefix (model-ID comparison ignoring the provider part). An alias-inconclusive primary that is actually available (`opencode-go/kimi-k3` vs `opencode/kimi-k3`) always wins. Two new regressions pin both sides: alias-present → primary wins; verifiably-absent → fallback promoted.
2. **Full `models[0]` contract preserved** — `subagent-model-resolution.ts` now merges the complete `models[0]` object (variant, reasoning, reasoningEffort, temperature, top_p, maxTokens, thinking) into the resolved config via `applyPrimaryEntrySettings()`, applied after category params so the primary entry (most specific) wins. New test asserts all 7 fields survive resolution.

**Verification:** delegate-core **13 pass / 0 fail**, delegate-task **490 pass / 0 fail**, typecheck exit 0. Evidence: `.omo/evidence/20260901-6874-review2-fix/QA.md`.

Ready for re-review — the `CHANGES_REQUESTED` is from the pre-fix head.

EvangelosMoschou:
## CI note: `senpi-compatibility (ubuntu-latest)` is the pre-existing #6875 flake

The one red check is `Verify committed Senpi plugin bundle is current` → `omo-senpi extension build is not current: stale-output` — the known non-deterministic bundle build (issue #6875), not a test failure.

Evidence this is not caused by this PR:
- This exact check **also failed on the original head `c67340aa2`** before any of my changes (pre-existing).
- The current head `f5f61bd1` has **3 successful full CI runs** — the same check passes on re-runs (nondeterministic by nature).
- This PR's diff touches only `packages/delegate-core/` and `packages/omo-opencode/src/tools/delegate-task/` — zero overlap with `packages/omo-senpi/plugin/extensions/` or the bundle check.
- `dev` itself has a red `senpi-compatibility` run (33479170709), confirming the base branch is not clean on this check.

The same flake is documented in #6929's residual-risk section. Happy to track it in #6875 if that helps.

PR #6898, opened 2026-08-16
perf(omo-opencode): bound runtime skill source handshake at plugin init

herjarsa:
## Problem

The runtime security skill source server handshake (\createRuntimeSkillSourceServer\) runs **unbounded** inside the plugin factory. With a real OpenCode client the handshake stalls the session boot ~10s (circular wait: the server waits on the client, the client waits for the factory to finish).

Measured: \ootstrapping → session created\ ≈ 38s total (factory alone 10.9s).

## Fix

Bound the handshake with a 500ms \Promise.race\ timeout. On timeout the server is skipped for that session (graceful degradation — existing catch logs and continues without \config.skills.urls\). The security skills are on-demand only, so a ~1s late arrival (or skipped session) is acceptable.

## Tests

- New: \create-plugin-module.deferred-skill-source.test.ts\ — RED (factory awaited the runtime skill source server) → GREEN with the fix; asserts the server is still invoked (spy).
- Factory suite: 6 pass, 0 fail.

## Evidence

\ootstrapping → created\ went from ~38s to **~15.1s** (repeated runs).

---
## Summary by cubic
Bounds the runtime security skill source handshake during `omo-opencode` plugin initialization to prevent session boot stalls. Previously the factory awaited `createRuntimeSkillSourceServer` unbounded (≈10s delay); now it races the handshake with a 500ms timeout and, on timeout, skips the server for that session, logs a warning, and continues without `config.skills.urls`.

- Adds a 500ms `Promise.race` around `createRuntimeSkillSourceServer` and clears the timer; fast paths (<500ms) behave unchanged.
- On timeout, execution flows to the existing catch path (warns and proceeds); security runtime skills may be unavailable for that session.
- Adds `create-plugin-module.deferred-skill-source.test.ts` to assert factory returns quickly and the server is still invoked.
- Observed boot time reduced from ~38s to ~15.1s.
- No API changes or migration required.

<sup>Written for commit e99e43837b8cd2ccde219c417111ea9136ce0ff6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6898?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Welcome, herjarsa — nice profiling catch. The unbounded factory await on the MCP handshake creating a circular wait is a real boot stall.

Verified that `createPluginModule` from `src/testing/` is the production factory (imported by `src/index.ts` as the plugin entry), so the timeout fix lands on the real path. RED→GREEN test pattern is clean, and the graceful degradation (skip + warn) is right for on-demand-only security skills.

One note: the file being in `src/testing/` is misleading — worth a future rename or an explicit comment since it IS production code. Not blocking.

LGTM after cubic review.

PR #6914, opened 2026-08-16
fix(ulw-loop): isolate Senpi continuation by PI session

### LINKED ISSUE #6828: [Bug]: omo-senpi ulw-loop continuation can hijack another session in the same working directory

feelsodev:
## Summary

- recognize Senpi `PI_SESSION_ID` in the existing ulw-loop scope resolver
- keep explicit OMO and Codex session environment precedence unchanged
- prove that two independent Senpi processes sharing one cwd cannot observe or continue each other runs
- preserve the owner happy path and legacy explicitly unscoped CLI behavior

Fixes #6828.

This is an alternative to #6829. It uses the session environment already guaranteed by Senpi instead of adding a per-extension global environment lifecycle manager.

## Root cause

The ulw-loop CLI recognized OMO and Codex session variables but not the current Senpi session variable. Senpi status checks and model shell commands therefore fell back to `.omo/ulw-loop/`, making active state visible to every session in the same cwd.

## RED -> GREEN

- RED: session B received hidden `omo-senpi:ulw-continuation` for session A root-global run.
- RED: session B `status --json` returned 0 and exposed session A plan.
- GREEN: session B receives 0 messages while session A receives exactly 1 continuation.
- GREEN: session B returns `ULW_LOOP_PLAN_MISSING`; session A reads `.omo/ulw-loop/session-A/goals.json`.

## Real-surface QA

`bun --cwd . packages/omo-senpi/scripts/qa/probe-cross-session.mjs`

The probe launches two independent Bun processes against the real omo-senpi component and staged toolkit from one shared temp cwd. Result:

```json
{"verdict":"PASS","sessionA":{"messageCount":1},"sessionB":{"messageCount":0},"paths":{"ownerPlan":true,"unrelatedPlan":false,"sharedRootPlan":false},"cleanup":{"removedSharedCwd":true}}
```

## Validation

- `bun test packages/omo-senpi/src/components/ulw-loop` - pass
- Codex ulw-loop suite - 424/424 pass
- `bun run --cwd packages/omo-senpi typecheck` - pass
- Codex ulw-loop `tsc --noEmit` - pass
- `bun run test:senpi` - 1874/1874 pass
- Senpi plugin stage build - pass
- changed Codex files Biome - pass

---
## Summary by cubic
Isolates `ulw-loop` state by Senpi PI session so sessions sharing a cwd don’t see or continue each other’s runs. Previously the resolver ignored `PI_SESSION_ID` and wrote to `.omo/ulw-loop/`; now it resolves `PI_SESSION_ID` and writes to `.omo/ulw-loop/<session>/` without changing OMO/Codex env precedence or the legacy unscoped CLI path.

- Adds `PI_SESSION_ID` to the `ulw-loop` scope resolver in `omo-codex`; exposes `resolveUlwLoopSessionIdFromEnv` test that selects the Pi session id.
- Extends tests in `omo-codex` and `omo-senpi` to prove cross-session isolation; adds `packages/omo-senpi/scripts/qa/probe-cross-session.mjs` that spawns child processes and verifies continuation counts and plan paths; verifies Windows `.cmd` launcher via `toSpawnTarget`.
- Test hygiene: sets `PI_SESSION_ID` to empty in `vitest`; scopes env mutations to synchronous helpers; preserves `PATH` in the toolkit fixture.
- No migration required; Senpi already exports `PI_SESSION_ID`.

<sup>Written for commit f415fc499cd94dd971fba568f6f1fe1b44ff9c80. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Thanks @feelsodev — this is merged as `9b24f00400865a6c99451053beb7a4597372158b`, and #6828 is closed by it. Before merging I rebased it onto current `dev` (merge commit `f415fc499`, resolving an additive conflict in `index.test.ts` against the new plan-directory-guard tests from `17ef7ad90`), then re-ran the ulw-loop suites (senpi 50/50, codex 422/422), typecheck, and your cross-session probe — all green, plus a mutation check confirming the probe genuinely fails when the fix is disabled.

One follow-up for visibility: adding `PI_SESSION_ID` to the resolver scopes every CLI call the **model** makes, since senpi's bash tool injects that variable per spawn. The omo-senpi component's own status probe does not get it — senpi never sets `PI_SESSION_ID` on the extension host's `process.env`, and `runOmoCommand` spawns without an explicit `env` — so that path still resolves the repo-level `.omo/ulw-loop/`. Net effect is that runs are now written session-scoped but read unscoped.

Details, reproduction, and fix options are in #7108. Not a defect in what you shipped — the resolver change is correct and needed either way; it just needs a companion change on the spawn side to close the loop.

— [sisyphus-bot]

PR #6925, opened 2026-08-17
ci: upgrade bun and partition root tests safely

code-yeongyu:
## Summary

The Windows root matrix was configured with `--shard=1/2` and `--shard=2/2`, but bun 1.3.12 silently ignores `--shard`, so both jobs ran the entire 2040-file suite. Bun 1.3.14 implements real sharding and process-level parallelism.

A first implementation used native file sharding and proved disjoint file sets in CI, but that exposed a repository-specific constraint: `zauc-mocks-*` test files install global `mock.module()` bootstraps by alphabetical load order. Native sharding can place a bootstrap in one process and its consumers in another, producing 74 failures. This PR therefore uses measured **package-level partitions**, not native `--shard`, so every omo-opencode bootstrap stays with its consumers.

## Changes

- Move the repository's uniform Bun pin from 1.3.12 to the published 1.3.14 release, including all workflow cache keys, invariant tests, and the CI-pinned devcontainer.
- Exclude `packages/omo-senpi` from root runs; the dedicated merge-blocking `senpi-compatibility` job owns all 273 files (#6926).
- Windows group 1: `packages/omo-opencode packages/memory-core` (measured 7.20m).
- Windows group 2: every other non-Senpi root test (measured 5.87m).
- Linux/macOS: all non-Senpi root tests.
- Add `--parallel` now that per-process XDG isolation landed in #6931.
- Add `script/ci-root-test-partition.test.ts`, a machine-consumed contract that rejects native `--shard`, requires `--parallel`, and pins omo-opencode to one process.

### Bun version capability

| bun | full | `--shard=1/2` | `--shard=2/2` | overlap |
|---|---|---|---|---|
| 1.3.12 | 1049 tests / 135 files | 1049 / 135 | 1049 / 135 | 135 files |
| 1.3.14 | 1049 / 135 | 560 / 68 | 489 / 67 | 0 files |

This proves why the version bump is required. It does **not** justify native sharding in this repository; the zauc bootstrap failure below is why package partitions replace it.

### Real-CI discovery that invalidated native file sharding

Run 32020073695 assigned 1020 files to each Windows process with overlap 0 / union 2040, but Windows 2/2 reported 74 failures. The first was `category-resolver.test.ts`; the failing consumers import modules globally mocked by `zauc-mocks-subagent-resolver/subagent-resolver.test.ts`, which was assigned to the other process. Auto-merge was disabled immediately and the design corrected.

### Safe partition contract

- RED: 0 pass / 3 fail on the native-shard workflow.
[...]

PR #6926, opened 2026-08-17
ci: stop running the omo-senpi suite twice per OS

code-yeongyu:
## Summary

`packages/omo-senpi` already has a dedicated merge-blocking job, `senpi-compatibility`, which builds the plugin, packs it, typechecks it, and then runs `bun test packages/omo-senpi`. The root test suite picked the same files up as well, so 273 test files ran twice on every OS. This scopes the root job's invocation so each file runs once, in the job that owns it.

## Changes

- Root `test` job now runs `bun test --path-ignore-patterns="packages/omo-senpi/**"`, composing with the existing `--shard` expression.
- A comment records why this is a per-invocation flag rather than a `bunfig.toml` entry, so the safer-looking config change is not "restored" later.

## QA & Evidence

- **What was tested:** that the exclusion removes the duplicate from the root run *without* touching the senpi job's own explicit invocation.
- **Observed result:**
  - bare run + flag over `memory-core` + `omo-senpi` -> 546 tests / 67 files (omo-senpi dropped, the other package intact)
  - `bun test packages/omo-senpi` (the senpi job's command, unchanged) -> **1877 tests across 273 files**, unaffected
- **Why sufficient:** the second measurement is the safety property. `pathIgnorePatterns` in `bunfig.toml` is global, so adding `"packages/omo-senpi/**"` there also suppresses the senpi job's explicit path — that configuration reports `5748 files were searched` and runs **zero** tests, silently deleting 273 files of coverage while CI stays green. The CLI flag is scoped to the invocation that sets it. This was caught before commit and the config approach was reverted.
- **Coverage ownership:** running all 273 files *without* `build:senpi-plugin` fails 11 tests — `OMO Senpi scoped skill sync` (7), `runSenpiInstaller source refresh` (3), `omo-senpi ultrawork component` (1). They assert on staged build artifacts, confirming `senpi-compatibility` is the correct owner rather than an incidental duplicate.
- **Artifact:** `.omo/evidence/20260817-senpi-test-dedupe/red-duplication.txt`

## Measured impact

From the per-file timings of run 32011189074:

| OS | omo-senpi share of the root suite |
|---|---|
| windows | 10.9m of 24.0m (**46%**) |
| ubuntu | 2.8m of 6.9m (41%) |

## Risks & Residuals

- **Root suite no longer catches omo-senpi breakage.** Accepted: `senpi-compatibility` is merge-blocking on all three OSes, and `senpi-test-script.test.ts` asserts that job's shape, so the coverage cannot silently disappear.
- **Independent of the bun pin.** This PR does not depend on #6925 and can land in either order.
- **Pre-existing, unrelated:** `script/agent-command-string-audit.test.ts` fails on `dev` today (allowlist `file:LINE` drift). Not touched here.

---
## Summary by cubic
Stops running `packages/omo-senpi` tests in the root CI job; they now run only in the merge‑blocking `senpi-compatibility` job that builds the plugin. Previously these files ran twice per OS; now they run once in the job that owns their artifacts. Impact: removes ~46% of Windows root suite time and ~41% of Ubuntu root suite time.

- Root job runs `bun test -c bunfig.root.toml` (composed with `--shard`). The dedicated config preserves all base ignores and adds `packages/omo-senpi/**`.
- Base `bunfig.toml` remains unchanged, so `senpi-compatibility` (`bun test packages/omo-senpi`) still runs all 273 files and stays merge‑blocking on all OSes.
- Adds `script/root-test-config.test.ts` to assert base ignores are preserved, Senpi is excluded in the root config, and CI invokes `bun test -c bunfig.root.toml`.
- Avoids the CLI `--path-ignore-patterns` flag, which replaces the ignore list and would re‑include web, vendored, plugin, and upstream tests and could suppress Senpi coverage in its own job.

<sup>Written for commit 67591743c9252e8df061639c453786bec5dd6c6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
         # packages/omo-senpi is covered by the senpi-compatibility job. A dedicated config
         # preserves every root ignore while adding Senpi; the CLI path-ignore flag replaces
         # bunfig.toml's list and would re-include web, vendored, plugin, and upstream tests.
-        run: bun test -c bunfig.root.toml${{ matrix.shard && format(' --shard={0}', matrix.shard) || '' }}
+        run: bun --config=bunfig.root.toml test${{ matrix.shard && format(' --shard={0}', matrix.shard) || '' }}
 
       - name: Write job summary
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,11 @@ jobs:
         working-directory: packages/lsp-daemon
 
       - name: Run tests
-        run: bun test${{ matrix.shard && format(' --shard={0}', matrix.shard) || '' }}
+        # packages/omo-senpi is covered by the senpi-compatibility job, which builds the
+        # plugin first; running it here too duplicated 46% of the Windows suite. The flag is
+        # per-invocation on purpose - a bunfig pathIgnorePatterns entry would also suppress
+        # the senpi job's own explicit `bun test packages/omo-senpi`.
+        run: bun test --path-ignore-patterns="packages/omo-senpi/**"${{ matrix.shard && format(' --shard={0}', matrix.shard) || '' }}
 
       - name: Write job summary
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,11 +119,10 @@ jobs:
         working-directory: packages/lsp-daemon
 
       - name: Run tests
-        # packages/omo-senpi is covered by the senpi-compatibility job, which builds the
-        # plugin first; running it here too duplicated 46% of the Windows suite. The flag is
-        # per-invocation on purpose - a bunfig pathIgnorePatterns entry would also suppress
-        # the senpi job's own explicit `bun test packages/omo-senpi`.
-        run: bun test --path-ignore-patterns="packages/omo-senpi/**"${{ matrix.shard && format(' --shard={0}', matrix.shard) || '' }}
+        # packages/omo-senpi is covered by the senpi-compatibility job. A dedicated config
+        # preserves every root ignore while adding Senpi; the CLI path-ignore flag replaces
+        # bunfig.toml's list and would re-include web, vendored, plugin, and upstream tests.
+        run: bun test -c bunfig.root.toml${{ matrix.shard && format(' --shard={0}', matrix.shard) || '' }}
 
       - name: Write job summary
         if: always()
--- a/bunfig.root.toml
+++ b/bunfig.root.toml
@@ -1,0 +1,15 @@
+[test]
+preload = ["./test-env.ts", "./test-setup.ts"]
+pathIgnorePatterns = [
+  "dist/**",
+  "packages/web/**",
+  "packages/lsp-tools-mcp/**",
+  "packages/lsp-daemon/**",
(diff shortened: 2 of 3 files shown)


code-yeongyu:
### Correction: dedicated Bun config, not CLI path-ignore

The initial per-invocation `--path-ignore-patterns="packages/omo-senpi/**"` looked scoped, but Bun replaces `bunfig.toml`'s complete ignore list when that flag is present. Ubuntu therefore re-included web, vendored LSP, Codex plugin/scripts, and shared-skill upstream tests, producing 108 unrelated failures (including missing `@playwright/test`).

The final implementation uses `bunfig.root.toml`, which copies every base ignore and adds `packages/omo-senpi/**`. CI selects it with the empirically verified syntax `bun test -c bunfig.root.toml`; `bun -c ... test` incorrectly tries to execute the TOML file.

Contract evidence:
- RED: 0 pass / 2 fail while the dedicated config was absent
- GREEN: 3 pass / 0 fail
- Bun 1.3.12 config invocation: 37 pass / 0 fail across 8 files
- explicit Senpi job remains 1877 tests / 273 files

Commit: `2ad438f87`.

code-yeongyu:
### CI miss: `-c` after `test` is a filter on bun 1.3.x

Current-head ubuntu/windows failed in 40ms with:

```
The following filters did not match any test files:
 bunfig.root.toml
7336 files were searched
```

`bun test -c bunfig.root.toml` is **not** a config select on 1.3.12/1.3.14. Reproduced locally with `bun@1.3.14`. The working form is:

```
bun --config=bunfig.root.toml test
```

Pushed that plus a spawn that proves `--config=` actually applies `packages/omo-senpi/**`.

code-yeongyu:
### Measured ubuntu after `--config=`

The previous head died in 40ms treating `bunfig.root.toml` as a filter. This head actually ran the suite:

| PR | command | files | tests | bun wall | job wall | senpi files | fail |
|---|---|---|---|---|---|---|---|
| #6926 | `bun --config=bunfig.root.toml test` | 1776 | 13890 | 217.50s | **5.48m** | 0 | 0 |
| #6925 | `bun --config=bunfig.root.toml test` | 1777 | 13893 | 228.58s | **5.55m** | 0 | 0 |

Baseline ubuntu root job was **9.3m / ~2040 files**. Senpi exclusion is now real on ubuntu. Waiting on Windows + macOS before merge.

code-yeongyu:
### Windows after Senpi exclude (still sharded no-op)

Run `32036237290`:

| job | wall | vs 28.7m / 28.2m |
|---|---|---|
| ubuntu | **5.48m** | 9.3m |
| windows 1/2 | **17.72m** | 28.7m |
| windows 2/2 | **17.10m** | 28.2m |

Both Windows shards still run the full non-Senpi suite (`--shard` is a no-op on 1.3.12). The cut is the 248 Senpi files. Waiting on macos root tests; auto-merge stays on.

PR #6947, opened 2026-08-17
fix(model-core): use openai-codex provider for quick luna-fast rung

code-yeongyu:
## Summary

Fixes the `quick` category builtin fallback chain so its `gpt-5.6-luna-fast` rung uses the `openai-codex` provider instead of `quotio-openai`. The `quick` chain was the lone outlier using `quotio-openai` for luna-fast — the `explore` and `librarian` agent chains already use `openai`, and the canonical provider for this model on Codex-backed setups is `openai-codex`.

Depends on **code-yeongyu/senpi#918**, which adds `gpt-5.6-luna-fast` (and sol/terra-fast) to the `openai-codex` provider catalog.

## Why

The Codex OAuth backend (`chatgpt.com/backend-api`, provider `openai-codex`) is the primary path for users on ChatGPT subscriptions. Without `gpt-5.6-luna-fast` in the `openai-codex` catalog, the `quick` category's luna-fast rung was unreachable for these setups, falling through to slower models unnecessarily. Switching the rung to `openai-codex` (combined with the senpi PR) makes the rung resolve correctly.

## Changes

| File | Change |
|---|---|
| `packages/model-core/src/category-model-requirements.ts:78` | `quick` luna-fast rung `providers`: `["quotio-openai"]` → `["openai-codex"]` |
| `packages/senpi-task/src/category/fallback-chains.ts:64` | Mirror the same change (hand-mirrored per file convention) |
| `packages/model-core/src/model-requirements-categories.test.ts:90` | Test assertion updated to `["openai-codex"]` |
| `packages/model-core/src/category-routing-policy.test.ts:65` | Test assertion updated to `["openai-codex"]` |
| `docs/reference/configuration.md:458` | Quick row updated: `quotio-openai/gpt-5.6-luna-fast` → `openai-codex/gpt-5.6-luna-fast` |

## Test evidence

```
packages/model-core/src/model-requirements-categories.test.ts:
(pass) CATEGORY_MODEL_REQUIREMENTS > quick follows the approved 8-rung chain [0.12ms]

packages/model-core/src/category-routing-policy.test.ts:
(pass) category routing policy > quick prioritizes Kimi high-speed, Luna low, DeepSeek off, then the speed tier [0.10ms]

14 pass, 0 fail
```

## Risk

[...]

PR #6953, opened 2026-08-17
chore(omo-senpi): rebuild plugin bundles after #6947 provider-chain change

code-yeongyu:
## Summary
Post-merge of #6947, dev CI's `senpi-compatibility (ubuntu-latest)` fails `Verify committed Senpi plugin bundle is current` (stale-output) — the provider-chain change altered sources feeding the bundle, and `auto-commit-schema` only runs on `master`, so nothing rebuilt the committed extensions on dev.

This PR is just the regenerated artifacts (`node packages/omo-senpi/plugin/scripts/build-extension.mjs`, `--check` now passes locally). No source changes.

## Note
The master-only auto-commit leaves dev exposed to exactly this drift on every bundle-feeding merge — might be worth extending the auto-commit to dev, or making the verify step auto-repair instead of fail.


code-yeongyu:
Closing as duplicate — #6950 (created 17 min earlier) ships the same bundle regeneration. One note worth keeping from this PR: `auto-commit-schema` only runs on master, so every bundle-feeding merge to dev leaves committed extensions stale until a manual rebuild. Worth extending the auto-commit to dev or making the verify step auto-repair.

PR #6954, opened 2026-08-17
fix(senpi-task): reject batch task items that fail validation instead of dropping them

ismetanin:
## What was broken

A `task` call using `tasks: [...]` silently discarded items that failed validation.

Directly observed: a single call submitted **six** items with top-level `run_in_background: true`. Three were well-formed. Three carried `category`, `name`, `description` and `task_summary` but omitted `prompt`. The tool returned `Batch running.` and listed **three** spawned children. The other three were neither spawned nor reported as failed - no error, no warning, no partial-failure notice. The response was shaped exactly like a full success.

This was caught only because the caller counted the returned list against what it submitted. On a larger batch, or with a caller that does not count, the work is silently lost and the run proceeds on incomplete evidence believing it completed. In the run where this happened it would have lost three parallel research lanes.

## Root cause

The items were destroyed **before validation ever ran**, which is why adding per-item checks to `resolveSpawnItems` alone would have looked correct and changed nothing.

`normalizeTaskToolArguments` in `argument-normalization.ts` is registered as the task tool's `prepareArguments` hook (`tools/task/tool.ts`). Its `taskItems()` helper `flatMap`ped away every item for which `taskItem()` returned `undefined`, and `taskItem()` returned `undefined` whenever `prompt` was missing or blank. By the time `resolveSpawnItems` received `params.tasks`, the malformed items no longer existed.

## Approach: atomic rejection

The whole call is rejected and nothing spawns. This matches the existing error contract rather than introducing a second one: every malformed-input path on this tool surface already rejects the entire call with `invalid_arguments` (`execute.ts`, the oversized-batch guard in `execute-batch.ts`, `control/cancel.ts`, `control/send-results.ts`, `output/output.ts`, `team/lifecycle.ts`). There is no precedent in the package for a partially executed call reporting per-item rejections.

## What changed

- **`argument-normalization.ts`** - prompt-less items are preserved so validation can see and name them.
- **`validation.ts`** - new `SpawnItemsInvalidError` (`code: \"invalid_items\"`) carrying per-item `index`, `field` and `reason`, plus `accepted` and `submitted`. Batches aggregate **every** fault so the reported set does not depend on item order; single spawns keep their typed `item_target` first-error contract, which `execute.ts` special-cases.
- **Provider padding is now distinguished from malformed real work.** An item is padding only when its prompt is a known padding token, or when it carries no content at all. This closes a second, pre-existing silent drop: a top-level `prompt` alongside prompt-less items discarded the entire `tasks` array and spawned one unrelated child under a success response.

Error text:

```
Rejected 3 of 6 task items; nothing was spawned. Fix and resubmit the whole batch.
Invalid: item 1 (prompt: prompt is required and must be non-empty); item 3 (...); item 5 (...).
```

## Observed behavior

Driven through the real `prepareArguments` -> `buildTaskExecute` boundary with a fake manager recording every `start()`:

| Case | `start()` calls | `details.status` |
|---|---|---|
| Reported 6-item payload (3 missing `prompt`) | **0** | `invalid_arguments` |
| Valid batch of 4 | **4** | `running` |
| 16 valid (max) | **16** | `running` |
| 17 valid (over max) | **0** | `invalid_arguments` |
| Batch of 1 invalid | **0** | `invalid_arguments` |

`details.items` is absent on rejection and the text contains no `Batch running.`, so it cannot be mistaken for success. `renderers.ts` already maps `invalid_arguments` to an error style.

## QA

- `bun test`: **15760 pass / 37 skip / 0 fail** in a single run.
- `bunx tsgo --noEmit -p packages/senpi-task/tsconfig.json`: exit 0.
- **Failing-first**: reverting the source with the tests kept gives **13 fail / 7 pass**. The 7 passes are intentional no-regression guards (valid batch resolves and spawns, inheritance applies, both padding shapes absorbed, single `item_target` holds). Against the unfixed code the execute test reproduces the incident exactly: 3 children spawned from a 6-item submission with no error.
- Evidence: `.omo/evidence/20260817-task-batch-silent-drop/qa.md`.

Acceptance criteria covered by tests: valid batches unchanged; missing `prompt` names every index and the field; `accepted + rejected == submitted` asserted; inheritance applied **before** validation (an item valid only after inheritance is not wrongly rejected); the category + model exclusivity rule still rejected whether the model is the item's own or inherited.

## Review

Reviewed by two independent agents. Both reproduced the failing-first split and the green suite. Three findings were accepted and fixed in a second round: the residual padding-path drop described above, order-dependent fault reporting (a target fault at a low index masked later missing prompts), and an execute test that bypassed `prepareArguments` and so did not reproduce the real incident. One claim - that this change *widened* the padding hole - was rejected on evidence: a differential sweep of the old and new normalizers showed the disputed input produces byte-identical output, so that drop predates this branch. It was fixed regardless, being the same defect class. Final verdict: green, with the invariant machine-checked across 9420 inputs with zero violations.

## Residual risk

Some previously "working" provider-padded calls now cost the model one retry: `{prompt: \"\", category: \"quick\"}` and a real lane whose prompt was forgotten are information-theoretically identical, so rejecting is the only safe reading. Known padding tokens are still absorbed, so the common padding shapes are unaffected.

Scope is limited to validation and its error reporting; the spawn path is untouched.


code-yeongyu:
Heads-up on the CI state (none of these are from your changes):

- `senpi-compatibility (ubuntu-latest)` fails `Verify committed Senpi plugin bundle is current` — dev's committed bundle went stale after #6947 merged. #6950 rebuilds it; once that lands, please rebase.
- `senpi-compatibility (windows-latest)` died in Set up job — GitHub infra rate-limiting tonight.
- The `executeOnCompleteHook` PowerShell failures on windows are a runner-environment issue being addressed in #6957.

Your batch-validation changes themselves look unrelated to all of the above.

MoerAI (changes requested):
`taskItem()` still returns `undefined` for non-record batch entries and `taskItems()` removes them with `flatMap`:
https://github.com/code-yeongyu/oh-my-openagent/blob/063aa391616c54841b86e793d37d58bc639e9b7c/packages/senpi-task/src/tools/task/argument-normalization.ts#L35-L65

Because normalization is the registered pre-schema `prepareArguments` step, `[{ prompt: "one" }, null, { prompt: "three" }]` still starts two children and reports success:
https://github.com/code-yeongyu/oh-my-openagent/blob/063aa391616c54841b86e793d37d58bc639e9b7c/packages/senpi-task/src/task-summary.ts#L5-L7

Preserve every raw index or carry structural rejections into `invalid_items`. Cover null, primitive, and array entries through prepare-to-execute, asserting zero child starts and `accepted + rejected === raw tasks.length`.

Then rebase onto current `dev`, regenerate `omo-task.js`, run current required CI, and add real registered-tool Senpi-host coverage proving a mixed structurally invalid batch creates no child records.

PR #6764, opened 2026-08-11
fix(runtime-fallback): session-stable fallback on terminal-quota 402 abort

niStee:
## Status: ready for review

This PR is the proposed reconciliation for the contradiction between #6677's `terminal_quota_exhausted` abort classification and the fallback-engagement requirement locked in #6487's Expected Behavior. Self-review is complete (architect + critic + red-team QA lanes all CLEAR/OKAY/passed); live end-to-end evidence captured 2026-08-12.

## Summary

#6677 (and #6492) classify LiteLLM's `terminal_quota_exhausted` 402 as the non-retryable `abort` type to prevent retry storms and the duplicate-parallel-subagent-spawn bug (#6487). That classification has a side effect: `runtime_fallback` logs `Error not retryable, skipping fallback` and a session with a configured fallback chain **just stops** — subagent sessions die with zero output tokens when the primary model's quota is exhausted (observed 13 times in one day on a live profile, e.g. `message.updated with assistant error {statusCode: 402, errorName: "APIError", errorType: "abort"}` → `Error not retryable, skipping fallback`).

This PR keeps the abort classification untouched and instead opens **one session-stable fallback dispatch** at the hook layer: when the error is classified `abort` **and** the status code is `402` (the LiteLLM terminal-quota signal), a session that resolves fallback models falls through the early-return and dispatches exactly one `autoRetryWithFallback` on the **same session**.

This reconciles the two requirements:
- **#6487's Expected Behavior**: "the existing session should be updated via fallback/retry … under no circumstances two active subagent sessions in parallel" — the dispatch reuses the session-stable machinery (`sessionRetryInFlight` / `pendingFallbackModel` / `sessionAwaitingFallbackResult`), never a second session.
- **Fallback engagement**: primary quota exhaustion continues work on the chain instead of stranding the session.
- **User aborts still skip**: `abort` without `402` (e.g. `MessageAbortedError`) does **not** fall back — the new branch only opens for `abort` + `402`.

## Changes

- `packages/omo-opencode/src/hooks/runtime-fallback/message-update-handler.ts`: `terminalQuota402Abort = classifyErrorType(error) === "abort" && extractStatusCode(error) === 402` — when true, skip the "not retryable, skipping fallback" early-return so the existing session-stable dispatch machinery runs.
- `packages/omo-opencode/src/hooks/runtime-fallback/terminal-quota-402-dispatch.test.ts`: exactly one session-stable dispatch on abort+402 with fallback models; zero for abort without 402; zero without fallback models.

[...]

Issue #3400, opened 2026-04-13
[Bug]: background fallback retries models that are not available in the current runtime


kywoo26:
## Bug Description

Background subagents can select fallback entries that are valid in general but not actually available in the current runtime model catalog.

In local repros this was visible with `librarian` and `explore`. The fallback chain could still attempt entries such as `opencode/minimax-m2.7`, `opencode/minimax-m2.7-highspeed`, and `opencode/claude-haiku-4-5` even when those models were not available for the current connected provider state.

## Steps to Reproduce

1. Use a config without per-agent overrides for `librarian` / `explore`
2. Trigger a background task that launches one of those agents
3. Inspect `/tmp/oh-my-opencode.log`
4. Observe fallback attempts against candidates that are not available in the current runtime

## Expected Behavior

Fallback selection should skip candidates that are not actually available in the current runtime model catalog and move directly to the next usable candidate.

## Actual Behavior

Unavailable fallback candidates were still selected and attempted. In local runtime logs those attempts produced `Model not found` errors before the runtime moved on.

## Doctor Output

```text
 oMoMoMoMo Doctor

 ⚠ 1 issue found:

1. Comment checker unavailable
   Comment checker binary is not installed.
   Fix: Install @code-yeongyu/comment-checker
   Affects: comment-checker hook
```

## Error Logs

```text
Model not found: opencode/minimax-m2.7.
Model not found: opencode/minimax-m2.7-highspeed.
Model not found: opencode/claude-haiku-4-5.
```

## Configuration

```json
{
  "agents": {
    "sisyphus": { "model": "openai/gpt-5.4", "variant": "medium" },
    "hephaestus": { "model": "openai/gpt-5.4", "variant": "medium" },
    "oracle": {
      "model": "openai/gpt-5.4",
      "variant": "high",
      "fallback_models": [
        { "model": "google/gemini-3.1-pro-preview", "variant": "high" }
      ]
    },
    "multimodal-looker": {
      "model": "openai/gpt-5.4",
      "variant": "medium",
      "fallback_models": [
        { "model": "openai/gpt-5-nano" }
      ]
    },
    "prometheus": {
      "model": "openai/gpt-5.4",
      "variant": "high",
      "fallback_models": [
        { "model": "google/gemini-3.1-pro-preview" }
      ]
    },
    "metis": { "model": "openai/gpt-5.4", "variant": "high" },
    "momus": {
      "model": "openai/gpt-5.4",
[...]

PR #6975, opened 2026-08-17
fix(delegate-task): advance category fallback chain on ProviderModelNotFoundError

### LINKED ISSUE #6972: [BUG] Category fallback chain does not advance when first model has ProviderModelNotFoundError
## Description

When a category's \models\ array has a first model that doesn't exist in the provider (ProviderModelNotFoundError), the fallback chain **does not advance** to the next model in the array.

## Expected Behavior

1. Category \erifier\ has models: \[MiniMax-M3, coding:reliable, nemotron-3.5-lightning-free]\
2. OpenCode tries \MiniMax-M3\ → fails with ProviderModelNotFoundError
3. OpenCode should try \coding:reliable\ → succeeds
4. Session continues

## Actual Behavior

1. OpenCode tries \MiniMax-M3\ → fails with ProviderModelNotFoundError
2. **Session dies immediately**
3. No fallback attempt

## Evidence

\\\
ProviderModelNotFoundError: Model not found: opencode-omniroute/minimax/MiniMax-M3
Did you mean: auto/minimax, kr/minimax-m2.1, kr/minimax-m2.5?
\\\

Error occurs at \SessionPrompt.getModel\ - before runtime_fallback can act.

## Impact

- Users cannot configure fallback chains with models that may not exist
- Category fallback is effectively broken for unavailable models
- Forces users to know exact model names upfront

## Suggestion

The category \models\ array should implement a try-next-on-failure pattern:
- If model N fails with ProviderModelNotFoundError → try model N+1
- Continue until a model succeeds or all models exhausted
- Only then fail with error

## Related

- PR #6764 (fixes 402 fallback, but not ProviderModelNotFoundError)
- Issue #3400 (similar, but for background fallback)

herjarsa:
## Summary

When a category's \models\ array has a first model that doesn't exist in the provider (ProviderModelNotFoundError), the fallback chain now advances to the next model in the array instead of dying immediately.

## Changes

- **sync-session-creator.ts**: Added try-next-on-failure logic to \createSyncSession\ that catches ProviderModelNotFoundError and advances through the fallback chain
- **sync-task.ts**: Passes the fallback chain to \createSyncSession\ and uses the effective model when a fallback succeeds
- **sync-session-creator.test.ts**: Added 7 comprehensive tests covering the new behavior

## Root Cause

ProviderModelNotFoundError was thrown at model resolution before the runtime-fallback hook could observe session errors. The error happened during session creation in the OpenCode core, not during API calls.

## Behavior After Fix

1. Category \erifier\ has models: \[MiniMax-M3, coding:reliable, nemotron-3.5-lightning-free]\
2. OpenCode tries \MiniMax-M3\ → fails with ProviderModelNotFoundError
3. OpenCode now tries \coding:reliable\ → succeeds
4. Session continues with the fallback model

## QA Evidence

- All 7 new tests pass
- Pre-existing test failures are unrelated to this change (version mismatches, missing dist files, platform-specific warnings)
- The fix is minimal and focused on the specific issue

## Related Issues

- Fixes #6972
- Related to #6764 (fixes 402 fallback, but not ProviderModelNotFoundError)
- Related to #3400 (similar, but for background fallback)

---
## Summary by cubic
Advances delegated-task category fallback during sync session creation when the primary provider model is missing. Previously, creation failed on ProviderModelNotFoundError; now it tries the next model and continues with the first working model, and downstream steps use that effective model. Also rebuilds `omo-senpi` extension bundles on Linux with bun 1.3.14 to satisfy CI freshness checks.

- Retries only on ProviderModelNotFoundError; detects via error name/nested cause/message, skips duplicate fallbacks, returns the primary error if all fallbacks fail, and returns an effectiveCategoryModel when a fallback succeeds.
- Passes the fallbackChain from task execution and switches subsequent prompt sending and metadata to the effectiveCategoryModel.
- Adds tests for fallback advancement, non-retry on other errors, duplicate-skip behavior, and effective model propagation.
- Rebuilds `packages/omo-senpi/plugin/extensions/*` on Linux with bun 1.3.14; no functional changes.

<sup>Written for commit 627ef35301986268075c2b7efd9e7a67e85acd72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6975?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Thanks for going straight from the report to a fix — really appreciate it.

Quick heads-up on the CI failure: the committed extension bundles were rebuilt in a different toolchain environment, so the freshness check flagged them as stale-output. Bundle bytes are sensitive to toolchain and OS, so they're hard to reproduce outside the repo's pinned setup. I regenerated them with bun 1.3.14 and pushed the result directly to your branch (c7bc8bd) — the check is green now.

Next time, feel free to commit source changes only; we can regenerate the bundles on our side.

The try-next-on-failure approach in createSyncSession matches what we discussed on the issue, and skipping duplicate models to avoid no-op retries is a nice touch. I'll review the implementation now.

code-yeongyu:
Small correction to my earlier bundle note: the bundles I pushed were built on macOS, and macOS and Linux produce different bundle bytes, so the ubuntu freshness check still failed. My local check passed on macOS, which gave me false confidence — my mistake for not catching the platform difference earlier.

I've regenerated the bundles on Linux (bun 1.3.14) and pushed the replacement in 627ef35. CI is running again; this should clear the freshness job for real this time.

MoerAI:
Thanks for the quick work, @herjarsa.

I am closing this because #6997 already fixed #6972 upstream by filtering unavailable models before category resolution. That solution covers both sync and background paths without adding another detector.

This branch's retry also does not fire on the declared SDK contract: `session.create` returns `{ error }`, which the existing interpolation reduces to `[object Object]` before the string matcher runs. The tests only cover thrown errors, and the broad `model not found` substring can retry unrelated upstream failures. If current `dev` still reproduces the issue after #6997, please reopen it with the new config and session evidence.

PR #6978, opened 2026-08-17
fix(memory): release the reflection reservation on every terminal path

ismetanin:
## Bottom line

A reflection run whose child died on a provider usage limit left `runtime/reflection/active.lock`
held forever. No later reflection could start for that identity, the footer kept animating
`mem:<agent-id> reflecting`, and nothing drained `runtime/facts-queue/` (64 files, oldest three days
old, on the reporting machine). The launcher process was still alive holding the reservation, so a
process-liveness check would not have reclaimed it.

## Root cause

The reported defects were one cause and two symptoms.

`identity-runtime.ts` minted run ids from a module-level counter, `reflection-run-${++runCounter}`,
that resets to 0 on every process start. memory-core's own default is `randomUUID`; only this
adapter overrode it. On 17 Aug the counter reached 4 again and produced `reflection-run-4`, an id
retired on 13 Aug. `createRunWorktree` did `mkdir(runDir, { recursive: true })`, so the new run
adopted the old directory and inherited its `final.json` (`outcome: "merged"`), `outcome.json` and
`terminal-claim.json`.

From there both terminal paths short-circuited on the inherited sentinel:

1. `publishRunOutcome` claimed the terminal (the 13 Aug claim matched on runId + attempt), wrote
   `publishing.json`, then returned at
   `if (existsSync(final.json) || existsSync(abandoned.json)) return` without writing today's
   `outcome.json`. That is why `outcome.json` on disk is still dated 13 Aug.
2. `finalizeRecordedOutcome` -> `withRunFinalizationClaim` -> `readTerminalRun` read the 13 Aug
   `final.json` and reported `{ status: "terminal", outcome: "merged" }`. `settleReservationRun`
   never ran, so `reservation.complete()` was never called and the lock survived.
   `requireFinalizedResult` then threw, and the throw was swallowed by the `.catch()` in
   `identity-runtime.launch`, which only logs.

`reconcileReflectionRuns` skips any run directory containing `final.json`, so the deadline watchdog
never inspected the wedged run. The deadline had armed correctly; it was never consulted.

Two corrections to the issue as filed, both load-bearing:

- **A `finally` would not have fixed this.** The failure handler ran; it was told the run was already
  terminal. Releasing the lock there while still recording the inherited `merged` would have
  advanced the journal cursor for a run that failed, permanently dropping that transcript window.
- **The spinner is not driven by the lock.** It is driven by the in-memory `activeRuns` registry,
  cleared only when a completion record is consumed. Releasing the lock alone leaves it spinning.
  Every release path therefore has to publish a completion record.

`resolveFinalizationDecision` already mapped a non-zero child exit to
`{ outcome: "failed", reason: "child_exit", detail: <child-stderr tail> }`, so "usage limit as a
first-class outcome" needed no new classification. The path to it was simply unreachable.

## Changes

**Run identity**
- `reflection-run-id.ts` (new) mints `reflection-run-<epochMs>-<8 hex>`, unique across processes and
  days. Legacy counter-minted directories are tolerated, never reused.
- `createRunWorktree` creates the run directory exclusively and refuses an existing one. The
  error path only removes a directory this call created, so a colliding run's artifacts survive.

**Terminal sentinels are trusted only when fresh**
- `runArtifactIsFresh(ledger, finishedAt)` gates `readTerminalRun` and `runOutcomeMatchesLedger`; the
  ledger is passed at all four `withRunFinalizationClaim` call sites. A sentinel older than
  `ledger.startedAt`, or naming another run id, is ignored rather than thrown on, so the run
  finalizes normally instead of adopting an older run's outcome.
- A mismatched sentinel previously threw with the lock held and was skipped by reconciliation,
  stranding the lock permanently through exactly the legacy directory this change tolerates.

**Release on every terminal path**
- `releaseClaimed` completes the reservation when the finalization claim short-circuits, which is
  reachable whenever a settling process dies between writing `final.json` and completing.
- Reconciliation no longer skips a sentinel-bearing directory that is still the active reservation.
- `completeReservationIfActive` performs the active check inside the scheduler lock via
  `ReflectionReservationStore.completeIfActive`, closing a read-then-complete race that made
  `complete()` throw on a lost race. Applied to `settleReservationRun` as well.

**Stale reservation reclamation** (`memory-core/src/reflection/reclaim.ts`, new)
- Two tiers: a launcher proven dead or pid-recycled past `RESERVATION_PRELAUNCH_GRACE_MS` (60s), or
  a run past `deadlineAt + RESERVATION_FINALIZATION_GRACE_MS` (10 min). The finalization window
  matters because `deadlineAt` is the supervisor's SIGKILL instant, not run completion: the launcher
  still validates, merges and cleans up in git after it. Reclaiming at the deadline itself took the
  lock from healthy runs.
- The run deadline is an injected seam (`getRunDeadline`), wired from `identity-runtime.ts`, so
  memory-core does not read the adapter-owned ledger schema directly.
- A reclaim publishes a failure completion record, so the footer actually stops reporting the run as
  reflecting, and promotes any queued pending run subject to the machine's `isStillTriggered`
  contract, handing it back to be launched.
- `reconcile()` guards the reclaim so lock contention degrades to "retry next bind" rather than
  failing session binding.

**Retention**
- `settleReservationRun` unlinks `transcript-payload.json` after publishing `final.json`. Unique run
  ids remove the accidental bound the counter provided by reusing ~4 directory names, and the
  payload is the dominant per-run artifact (377KB of ~664KB measured). Ordered after publication so
  a crash mid-settle cannot lose the input before the result exists.

## Tests

Regression coverage for each acceptance criterion, every one verified to fail without the fix:

- A child exiting non-zero releases the lock, writes a `failed` completion carrying the exact
  provider error, prunes its worktree, and leaves the journal cursor unadvanced, including inside a
  run directory an earlier run had settled.
- A terminal sentinel from an earlier run, and one naming a different run, both release the
  reservation instead of stranding it.
- Run ids are distinct across days and within one millisecond (driven through the injected entropy
  seam, so the assertion is deterministic rather than probabilistic).
- A pre-existing run directory is refused and its artifacts preserved byte for byte.
- Reclaim fires on an expired deadline plus finalization window, on a dead launcher, and on a
  recycled pid; it does NOT fire for a young reservation whose launcher already exited, nor for a
  live launcher just past its deadline.
- A stale reservation with queued work promotes that work and hands it back to launch.
- The reclaim seam reads the ledger deadline, and a reclaim publishes its failure completion.

`bun test`: full suite green in a single run. `tsgo --noEmit` clean on `memory-core` and `omo-senpi`.

## Review

Two rounds of adversarial multi-agent review (goal, code quality, security, hands-on QA, context
mining) found four CRITICALs and several blockers in earlier revisions of this change, including a
release path that recorded an inherited `merged` for a failed run and advanced the journal cursor, a
deadline reclaim that was dead code in production because its seam was never wired, and a test that
asserted the opposite of its own title. All are fixed and re-verified. The full record, including
the process failures on the way, is in `.omo/evidence/20260817-reflection-lock-release/review.md`.

## Residual risk

Run-directory COUNT is now unbounded where the colliding counter accidentally capped it at ~4 names
per machine. Per-run size is reduced by dropping the payload, but a retention policy for terminal
reflection runs, mirroring `pruneTerminalFactsRuns` from #6900, is a genuine follow-up and is
deliberately not attempted here.

Relationship to #6853: that PR touches the same subsystem but different concerns (sandbox grant
creation and structured errno plumbing). The two compose; whichever lands second takes a small
textual conflict in the `run-finalization.ts` import block and around `failReservationRun`.

---
## Summary by cubic
Releases stuck reflection reservations by publishing a completion and freeing `active.lock` on every terminal path, and reclaims stale reservations. Previously, reused run directories and stale terminal sentinels let failed runs short-circuit finalization and hold the lock forever; now run IDs are unique, terminal sentinels are validated against the current ledger, and all terminal paths release the reservation so the spinner stops and queued work drains.

- Changes
  - Unique run IDs: `omo-senpi` mints `reflection-run-<epochMs>-<hex>` via `createReflectionRunId`; `create-run-worktree` creates the run directory exclusively and refuses existing ones.
  - Sentinel freshness: `runArtifactIsFresh` gates terminal reads with `ledger.startedAt`; `withRunFinalizationClaim` passes the ledger; stale or foreign sentinels no longer short-circuit the run.
  - Release on terminal: short-circuit and reconciliation paths call `completeReservationIfActive` for an atomic active-check + complete; `memory-core` adds `ReflectionReservationStore.completeIfActive`.
  - Reclaim stale reservations: `memory-core` `reclaim` recovers locks after a 60s prelaunch grace or after `deadlineAt + 10 min`, publishes a failure completion, and may promote queued work; `isStillTriggered` is exported; `omo-senpi` wires `getRunDeadline` and `settleReclaimedReservation`.
  - Retention: `transcript-payload.json` is unlinked after publishing `final.json`.

- Review and rollout
  - Behavior change: run IDs are no longer counter-based; update any tooling that assumed sequential IDs.
  - Adapter wiring: confirm `identity-runtime` supplies `createRunId`, `getRunDeadline`, and an `onReclaim` logger.
  - Concurrency: verify `completeIfActive` call sites hold the scheduler lock and avoid read-then-complete races.
  - Ops: reclaim grace values are `60s` prelaunch and `10 min` post-deadline; expect more run directories over time until a terminal-run retention policy lands.

<sup>Written for commit 6061772009e91a7a00e23cf585162b80aa47d878. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
Please rebase onto current `dev` and preserve the landed #7333 disk-scoped run-ID minting:
https://github.com/code-yeongyu/oh-my-openagent/pull/7333

Fix reclaim as a launch-producing transition. When `tryReserve()` evicts stale active A while B is pending, it promotes B to active but returns incoming C as pending; the real wiring launches only a returned active run. Return or dispatch B atomically and add a deterministic A-active, B-pending, C-trigger regression through the real Senpi launch callback:
- https://github.com/code-yeongyu/oh-my-openagent/blob/6061772009e91a7a00e23cf585162b80aa47d878/packages/memory-core/src/reflection/reservation.ts
- https://github.com/code-yeongyu/oh-my-openagent/blob/6061772009e91a7a00e23cf585162b80aa47d878/packages/omo-senpi/src/components/memory/wiring-runtime.ts

Route reclaimed and abandoned terminal outcomes through the normal completion-delivery path so `onCompletion` clears `activeRuns` immediately and durable records remain idempotent after retries or lost races. Cover live footer/RPC clearing, abandonment, and repeat reconciliation.

Run isolated real-Senpi memory QA with a deterministic terminal failure/reclaim signal, record reviewer-readable evidence, regenerate bundles, pass bundle freshness, scoped tests/typechecks, and current required CI.

PR #7018, opened 2026-08-18
fix(omo-senpi): preserve senpiPrefixArgs on reflection and people-ask launches

### LINKED ISSUE #7016: memory: reflection and people-ask children spawn without senpiPrefixArgs and die with "node: bad option: --fork"
## Problem

Memory reflection fails instantly on npm-installed `omo-ai` when the Senpi launch carries prefix args. Observed in a live session (omo-ai `5.0.0-0.beta.10`, npm install under nvm node v24.15.0, macOS arm64):

```
✗ Memory reflection failed · reflection-run-1
Reflection did not finish; the transcript cursor was not advanced.
took 232ms · reason child_exit · /Users/mirage/.nvm/versions/node/v24.15.0/bin/node: bad option: --fork
```

The reflection child was spawned as the node binary directly with Senpi CLI flags (`node -p --fork <session> ...`), so node rejects `--fork` (after treating `-p` as its own print flag) and exits within milliseconds. The run never starts, the transcript cursor never advances, and the pending payload keeps accumulating.

## Root cause

`packages/omo-senpi/src/components/memory/worker/spawn-payload.ts` — the reflection/dream and fork-mode launch builders ignore `input.senpiPrefixArgs` when an explicit `senpiCommand` is provided:

```ts
const launch = input.senpiCommand === undefined
  ? resolveSenpiLaunch(input.env)
  : { command: input.senpiCommand, prefixArgs: [] }   // drops senpiPrefixArgs
```

The facts builder in the same file does it correctly:

```ts
: { command: input.senpiCommand, prefixArgs: input.senpiPrefixArgs ?? [] }
```

The same hardcoded `prefixArgs: []` exists in `src/components/memory/commands/people-ask.ts` (same pattern, ~L97).

The shared launch preflight (`memory-launch-preflight.ts`) honors `senpiPrefixArgs`, so preflight probes the correct `node cli-main.js ...` command line while the actual spawn omits the CLI entry — preflight green, child dead. That mismatch is why the failure only surfaces at spawn time.

On the npm install shape (omo-ai under nvm node), the host resolves `senpiCommand` = node executable and `senpiPrefixArgs` = `[.../@code-yeongyu/senpi/dist/cli-main.js]`, which is how the spawn degenerates to bare `node --fork`.

Verified still present on `dev` @ c966d2a.

## Suggested fix

Mirror the facts-site handling at both call sites:

```ts
: { command: input.senpiCommand, prefixArgs: input.senpiPrefixArgs ?? [] }
```

## Impact

- reflection/dream runs never start (`child_exit` in ~200ms), so memory consolidation is dead on affected installs
- people-ask fails through the identical pattern

Happy to send a PR with regression tests.

code-yeongyu:
Fixes #7016

## Problem

On npm installs of `omo-ai` (host resolves `senpiCommand` = node binary, `senpiPrefixArgs` = `[.../@code-yeongyu/senpi/dist/cli-main.js]`), memory reflection/dream/fork children and people-ask children die within ~200ms with:

```
node: bad option: --fork
```

The explicit-`senpiCommand` launch branches hardcoded `prefixArgs: []`, dropping the CLI entry from the spawned argv, so the node binary receives senpi CLI flags directly. The shared launch preflight honors `senpiPrefixArgs`, which is why the failure only surfaces at spawn time (preflight green, child dead).

## Root cause (verified against dev @ 71f354d0a)

1. `packages/omo-senpi/src/components/memory/worker/spawn-payload.ts` — `prepareReflectionSpawn` ignores `input.senpiPrefixArgs` when `senpiCommand` is provided (`prefixArgs: []`).
2. Same file — `prepareReflectionForkSpawn` recomposes its own args without any prefix args, so fork mode stays broken even with (1) fixed (beyond the issue's suggested one-liner).
3. `packages/omo-senpi/src/components/memory/commands/people-ask.ts` — `createPeopleAskRunner` has the identical hardcoded `prefixArgs: []`.
4. `packages/omo-senpi/src/components/memory/worker/reflection-spawn-input.ts` — the reflection candidate spawn input did not declare/forward `senpiPrefixArgs` to the builders at all, so the prefix never reached the reflection path even in the resolver chain.

`prepareFactsSpawn` (same file) and `memory-launch-preflight.ts` already handle this correctly; this PR mirrors that handling everywhere else. No behavior change for installs that don't set an explicit `senpiCommand`.

## Change

- Mirror the facts-site launch resolution (`prefixArgs: input.senpiPrefixArgs ?? []`) at the reflection/dream and people-ask sites.
- `prepareReflectionForkSpawn` now composes its args from the resolved launch prefix so fork-mode argv starts with the CLI entry.
- Minimal forwarding plumbing so runner-level `senpiPrefixArgs` reaches the builders (reflection path + people-ask options).

## Verification

- RED-first regression tests: builder-level (`spawn-payload`) assertions for reflection and fork mode with npm-install-shaped `senpiCommand`/`senpiPrefixArgs`; people-ask black-box via `/bin/echo` marker prefix; forwarding assertions for the new plumbing.
- Real-surface probe: argv built through the real builders visibly starts with the CLI entry (`.../cli-main.js`) before `-p`/`--fork`.
- Targeted + adjacent suites (spawn, runner-fork-spawn, memory-launch-preflight, people-ask) green; `tsgo --noEmit` clean.
- No functional behavior changed beyond the prefix propagation this issue names.

---
## Summary by cubic
Preserves senpiPrefixArgs on reflection (including fork) and people-ask launches so npm-install deployments stop failing with “node: bad option: --fork.” Previously, setting senpiCommand dropped the CLI entry; now we always include the prefix when provided, with no change when senpiCommand is unset.

- Introduces resolveMemoryChildLaunch and uses it in reflection spawns; fork recomposes argv with the resolved prefix before -p/--fork.
- Forwards senpiPrefixArgs through reflection-spawn-input and runner-execution; people-ask honors the prefix with an explicit command.
- Adds tests covering reflection, fork, and people-ask prefixing; makes the people-ask prefix test portable by using the running interpreter with a printer script instead of /bin/echo.
- Regenerates extension bundles for Linux with Bun 1.3.14; no behavior change.
- Rollout: No migration required. Hosts that set senpiCommand should also supply matching senpiPrefixArgs (CLI entry).

<sup>Written for commit 6bb853d40b45709f0f6365df8a0bb66f638d6e07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Merging with the inherited `senpi-compatibility (windows-latest)` failure documented. All 4 new `senpiPrefixArgs` regression tests pass on Windows in this PR's runs.

Every failing case on this job is proven inherited from `dev`, not introduced here:

| Failing case | Evidence |
|---|---|
| `/doctor ... reflection health and remediation` | Controlled isolation: same worktree detached to `origin/dev` (e199029e1) with identical `node_modules` -> same 15 pass / 1 fail, same case. Pre-existing on dev. |
| `... process mode runs on Windows ... real RPC child` | Fails on dev's own CI run 32164968608; tracked in #6976. |
| `registerMemoryUsage ... foo.md.count is 1` | Fails on dev run 32164968608; code from #7015 (merged just before this branch). |
| `memory run supervisor ... identity is durable` | Flakes on dev runs 31601746537 and 31576880954; fixture-manifest flow this diff never touches. |
| `quick-pinned facts launch ... retries and commits with the fallback` | Flakes on dev runs 31601746537 and 31576880954; facts path untouched here (Windows temp-file ENOENT). |

This diff only threads an optional `senpiPrefixArgs` through the launch builders (inert when unset) and cannot reach health/doctor, task-driver, supervisor-manifest, or facts code. `dev` is red on this job independently, so it cannot gate this PR.

PR #7019, opened 2026-08-18
fix(lsp): normalize document uris at the openByUri boundary

### LINKED ISSUE #6167: [Bug] lsp_diagnostics drops PowerShell Editor Services push notifications due to file:// drive-letter case mismatch
### Prerequisites

### Bug Description

`lsp_diagnostics` returns "No diagnostics found" (or freshness timeout) for the PowerShell Editor Services (PSES) language server, even when the server is alive, the file is open, and PSES has published real diagnostics via `textDocument/publishDiagnostics`. The published diagnostics are silently dropped because of a `file://` URI **drive-letter case mismatch** between what the LSP server publishes and what lsp-tools-mcp uses as a Map key when storing the document state.

Reproduced end-to-end on Windows 11 with `pwsh` LSP configured per lsp-setup skill. A direct JSON-RPC probe against PSES shows it pushes `file:///c:/...` (lowercase `c:`), while Node's `pathToFileURL()` — which lsp-tools-mcp uses to compute the `openByUri` key — produces `file:///C:/...` (uppercase `C:`). JavaScript `Map#get` is case-sensitive, so the lookup in `recordPublishedDiagnostics` misses and the diagnostics are never recorded against the open document.

### Steps to Reproduce

1. Configure pwsh LSP (per the lsp-setup skill / builtin pwsh server) on Windows 11 with PowerShell 7.7.0-preview.2 + PowerShellEditorServices:

```json
{
  "lsp": {
    "pwsh": {
      "command": ["pwsh", "-NoProfile", "C:\\Users\\<user>\\AppData\\Local\\PowerShellEditorServices\\PowerShellEditorServices\\Start-EditorServices.ps1", "-Stdio"],
      "extensions": [".ps1", ".psm1", ".psd1"]
    }
  }
}
```

2. Open a `.ps1` file that has at least one parse error so publishDiagnostics will be non-empty, e.g. a script with `\` used as a line-continuation char instead of backtick (PowerShell parser emits `Missing expression after unary operator '--'`).

3. From an opencode session whose workspace root contains the file, call the `lsp_diagnostics` MCP tool on the `.ps1` file path.

4. Observe the tool returns `No diagnostics found` immediately, or after the freshness timeout elapses.

5. Independently verify PSES is fine by attaching a minimal JSON-RPC probe directly to the same pwsh LSP server (initialize → initialized → textDocument/didOpen with URI `file:///C:/Users/.../file.ps1`). PSES responds with `textDocument/publishDiagnostics` carrying **URI `file:///c:/Users/.../file.ps1`** (lowercase drive letter) and the expected 7 diagnostics.

### Expected Behavior

`lsp_diagnostics` should return the 7 diagnostics that PSES published (parser errors + ScriptAnalyzer warnings), matching the behaviour of other LSP clients (VS Code PowerShell extension, Neovim) on the same file.

### Actual Behavior

`lsp_diagnostics` returns `No diagnostics found`. lsp_status shows the pwsh LSP server alive for the workspace root with `refs=0` immediately after the call, and the server process keeps running. No error is raised — the published diagnostics are silently dropped.

### Doctor Output

```shell
 oMoMoMoMo Doctor

 ⚠ 2 issues found:

1. Loaded plugin is outdated
   Loaded 4.15.1, latest 4.18.2.
   Fix: Update: cd "C:\Users\Lance\.cache\opencode\packages\oh-my-openagent@latest" && bun add oh-my-openagent@latest
   Affects: plugin features

2. GitHub CLI notauthenticated
   Affects: GitHub automation
```

After running the suggested `bun add oh-my-openagent@latest` and `gh auth login`, `bunx oh-my-openagent doctor` reports the plugin at 4.18.2 and gh authenticated. The bug reproduces identically on 4.15.1 and 4.18.2.

### Error Logs

Direct JSON-RPC probe against `pwsh -NoProfile Start-EditorServices.ps1 -Stdio` (initialize → initialized → didOpen with uppercase C drive URI):

```text
REQUEST  initialize → RESPONSE id=1: {"capabilities":{"textDocumentSync":{"openClose":true,"change":2,"save":{"includeText":true}},...},"serverInfo":{"name":"pwsh","version":"7.7.0-preview.2 ..."}}
SENT     textDocument/didOpen with uri = file:///C:/Users/Lance/Dev/Toolbox/gdrive-sync.ps1
RECV     textDocument/publishDiagnostics:
         params.uri = file:///c:/Users/Lance/Dev/Toolbox/gdrive-sync.ps1   ← lowercase c:
         params.diagnostics.length = 7
         diagnostics[0].message = "Missing expression after unary operator '--'."
         diagnostics[0].range.start = { line: 99, character: 4 }
         diagnostics[1].message = "Unexpected token 'compare' in expression or statement."
         ... (3 more parser errors at lines 99/117/133, plus a PSScriptAnalyzer warning)
```

Node check on the same absolute path:

```text
node -e "import('node:url').then(({pathToFileURL})=>console.log(pathToFileURL('C:\\\\Users\\\\Lance\\\\Dev\\\\Toolbox\\\\gdrive-sync.ps1').href))"
→ file:///C:/Users/Lance/Dev/Toolbox/gdrive-sync.ps1   ← uppercase C:
```

Map key collision:

```text
openByUri.set(state.uri, state)                  // state.uri  = file:///C:/Users/...  (from pathToFileURL)
openByUri.get(params.uri) in recordPublishedDiagnostics  // params.uri = file:///c:/Users/...  (from pwsh)
→ undefined  →  recordPublishedDiagnostics returns early  →  state.lastPublish stays undefined
captureDiagnosticSnapshot(filePath) returns snapshot
resolvePushDiagnostics(snapshot) returns { status: "missing" }
loop falls through to freshnessTimeout(absPath)
```

### Configuration

```json
{
  "lsp": {
    "pwsh": {
      "command": ["pwsh", "-NoProfile", "C:\\Users\\Lance\\AppData\\Local\\PowerShellEditorServices\\PowerShellEditorServices\\Start-EditorServices.ps1", "-Stdio"],
      "extensions": [".ps1", ".psm1", ".psd1"]
    }
  }
}
```

#### Root cause (confirmed by reading manager.js 4.18.2 + observing raw protocol)

In `packages/lsp-tools-mcp/dist/lsp/manager.js` (compiled from `packages/lsp-core/src/lsp/`):

- Line 1346: `openDocumentSingleFlight(path, text)` builds the document state with `uri: pathToFileURL2(path).href`. Node's `pathToFileURL` uppercases the Windows drive letter (RFC 8089 §2, common Node convention): produces `file:///C:/...`.
- Line 1355: `this.openByUri.set(state.uri, state)` registers the state under the uppercase-C URI.
- Line 670 + 2499: `connection.onNotification("textDocument/publishDiagnostics", (params) => this.documents.recordPublishedDiagnostics(params))`.
- Line 1205-1208: `recordPublishedDiagnostics(params)` does `const state = this.openByUri.get(params.uri); if (!state) return;` — when `params.uri` arrives as `file:///c:/...` (lowercase), the Map lookup **misses** and the function returns early without recording `state.lastPublish`.
- Downstream `resolvePushDiagnostics(snapshot)` then sees `state?.lastPublish` is undefined and returns `{ status: "missing" }`.
- The diagnostics polling loop in `diagnostics()` (lines 2562-2621) never finds a ready push and, since PSES does not advertise `diagnosticProvider` (so `isDiagnosticPullSupported()` is false and `pushFallbackOnly = true`), it waits out `diagnosticsFreshnessTimeoutMs` and returns `freshnessTimeout(absPath)`.

#### Why PSES lowercases the drive letter

PSES is built on `OmniSharp.Extensions.LanguageServer` (C#). Its URI handling uses `System.Uri` with `UriKind.Absolute` and `Don'tEscapeSpecialChars`-ish construction. On Windows, `System.Uri` normalizes the host/drive portion, producing `file:///c:/...` for `file://` URI strings derived from local paths. This is not a PSES bug — it conforms to its own documented behaviour and matches what VS Code's PowerShell extension (which uses the same OmniSharp LSP SDK) is tolerant of. Other servers shipped by lsp-tools-mcp's own builtin table (e.g. `typescript-language-server`, `rust-analyzer`, `gopls`) emit URIs that round-trip through Node's `pathToFileURL` unchanged on Windows, so the case mismatch does not surface for them.

#### Related older issue

#6131 ("shared lsp_diagnostics freshness timeout after v4.18.1 upgrade") describes a *freshness-timeout* symptom observed with Marksman and gopls (both macOS). This issue is a different, deterministic failure mode specific to servers that emit `file://` URIs whose drive-letter case differs from Node's `pathToFileURL` output. It is Windows + PSES-specific and reproduces 100% of the time on any `.ps1` file PSES has diagnostics for.

#### Suggested fix (sketch, not a PR)

In `packages/lsp-core/src/lsp/manager.ts` (or `client.ts`), normalize `params.uri` to a canonical form before storing/looking it up so casing on the Windows drive letter cannot cause a Map miss. Two concrete options:

1. **Normalize before lookup.** In `recordPublishedDiagnostics`, before `openByUri.get(params.uri)`, run the URI through a normalizer that uppercases the drive letter on Windows when the scheme is `file:`. The same normalization should be applied wherever URIs are emitted by the manager (line 1346 for `state.uri`, lines 2506/2515/2525/2565/2627/2640 for outgoing request URIs). After normalization, `openByUri.set` and `openByUri.get` use the same key.

2. **Store by file path, not URI.** Reverse the indexing: have `recordPublishedDiagnostics` resolve `params.uri` back to a normalized filesystem path via `fileURLToPath` + `Filesystem.normalizePath` and key `openByUri` (or `openDocuments`) by that normalized path. Cross-server URI case quirks collapse because `fileURLToPath('file:///c:/...').toUpperCase() === fileURLToPath('file:///C:/...').toUpperCase()` on Windows case-insensitive filesystems. (opencode core `packages/opencode/src/lsp/client.ts` does exactly this in its `getFilePath` helper — it would close the gap entirely.)

Either approach keeps the existing 4.18.2 freshness/state architecture intact. Option 2 is the more robust long-term fix because it absorbs not only drive-letter case but also path-separator and percent-encoding quirks across servers.

#### Evidence files / reproducers

- Repro `didOpen` script (Node 24 ESM, ~85 lines) is what produced the protocol trace above. Happy to attach if useful — it initializes PSES, sends didOpen with uppercase URI, and prints every publishDiagnostics notification URI + count.
- Test fixture: any `.ps1` file containing `\` used as a line-continuation character (instead of backtick) somewhere will surface 6+ parser errors from PSES on first publish.

### Operating System

Windows

### OpenCode Version

1.18.3

LilMGenius:
Fixes #6167

## Problem

`WorkspaceDocumentState` keys `openByUri` on the raw URI string. The client opens a document with `pathToFileURL(path).href`, but a language server is free to publish diagnostics under a differently spelled URI for the same file. When it does, `openByUri.get(params.uri)` misses, `recordPublishedDiagnostics` returns early, and the diagnostics are discarded with no log line. `resolvePushDiagnostics` then reports `missing` and the request waits out its freshness window.

This thread already has two variants. This PR adds a third, found on Windows against `typescript-language-server` and `pyright`: the server percent-encodes the drive **colon** itself.

`didOpen` sent by the client:

```
file:///C:/Users/<user>/<workspace>/probe.ts
```

`textDocument/publishDiagnostics` returned for the same file:

```
file:///c%3A/Users/<user>/<workspace>/probe.ts
```

Both halves differ at once: the drive letter is lowercased **and** the colon is percent-encoded. A normalizer that only uppercases `file:///[a-z]:` cannot match this, because after the encode there is no literal `:` left to anchor on. The decode has to run first, then the drive-letter fold.

There is no fallback. `textDocument/diagnostic` (pull) against tsserver returns `-32601 Unhandled method`, so once the push is dropped there is nothing left, which is why the symptom is total silence rather than stale results. The user-visible effect was not `No diagnostics found` but a PostToolUse hook timing out at its 60s ceiling, because the client waits for an activity signal that can never arrive.

## Why this still applies after #6228

#6228 arrived at the same shape and was closed unmerged by its own author on 2026-08-05 after fourteen days without maintainer engagement, not on merit. On 2026-08-06 @code-yeongyu confirmed the root cause on #6167, stated that normalizing at the `openByUri` boundary is the correct direction, and self-assigned.

`dev` at `e199029e1` still has no `normalizeDocumentUri` and no `decodeURIComponent` in `workspace-document-state.ts`, and all eight `openByUri` accesses are still keyed on the raw URI. The reproduction was re-run against that commit today. Credit to @MoerAI and #6228 for the shape; this carries it forward with a third reproduction and a smaller review surface.

## Change

Two product files.

- `workspace-document-state.ts`: add `normalizeDocumentUri()`, a guarded `decodeURIComponent` followed by a Windows-gated drive-letter uppercase, and route all eight `openByUri` accesses (six `get`, one `set`, one `delete`) through it.
- `workspace-document-state.test.ts`: three tests covering the route-group encoding, the `c%3A` drive encoding, and the invariant that the outgoing URI is unchanged.

Normalization applies only to the lookup key. `state.uri` remains `pathToFileURL(path).href`, and every outgoing `didOpen`, `didChange`, `didSave`, and `didClose` sends that value byte-identical. On non-Windows platforms only the decode runs.

`fileURLToPath` is deliberately not used to canonicalize: it re-encodes a drive-like POSIX URI to `file:///c%3A/...` on Linux and throws on Windows for driveless paths.

## Verification

Windows 11, Bun 1.3.14, Node 24, branched from `dev` at `e199029e1`.

```
$ bun test packages/lsp-core/src/lsp/workspace-document-state.test.ts
 5 pass
 0 fail
 17 expect() calls
Ran 5 tests across 1 file. [2.08s]
```

```
$ node_modules/.bin/tsgo --noEmit -p packages/lsp-core/tsconfig.json
TYPECHECK_EXIT=0
```

`bun run typecheck` also completed with no diagnostics across all thirty package projects.

### Revert and fail

The normalizer alone was stashed, leaving the new tests in place.

```
$ git stash push -- packages/lsp-core/src/lsp/workspace-document-state.ts
$ bun test packages/lsp-core/src/lsp/workspace-document-state.test.ts
 3 pass
 2 fail

$ git stash pop
$ bun test packages/lsp-core/src/lsp/workspace-document-state.test.ts
 5 pass
 0 fail
```

The two failures track the normalizer exactly, so the tests measure the fix rather than the environment. On the red run the received value is `[]` in both cases, which is `recordPublishedDiagnostics` returning early on the missed key.

### Regression

```
$ bun test <every packages/lsp-core/src/**/*.test.ts except *.integration.test.ts>
 67 pass
 0 fail
 183 expect() calls
Ran 67 tests across 17 files. [860.00ms]
```

`bun test packages/lsp-core` is not a usable signal on this machine, and I want to be upfront about that rather than quote a green number I did not get. It fails on unmodified `dev` before any change, and the failing set is not stable between runs: 81 pass / 15 fail on unmodified `dev`, 79 pass / 20 fail with this change, with tests moving in both directions between the two runs. Every failure is confined to the four `*.integration.test.ts` files and the cause is visible in the output and unrelated to URIs:

```
Error: ENOENT: no such file or directory, open 'C:\Users\...\Temp\lsp-apply-edit-SOF2ja\scenario.json'
    at file:///.../packages/lsp-core/src/lsp/fixtures/workspace-edit-server.mjs:4:29
  ^ this test timed out after 5000ms
killed 2 dangling processes
```

The fixture server races its own scenario file and those suites spawn real child processes against a 5s timeout. The unit run above is therefore the regression signal here, and CI should be the arbiter for the integration suites. Happy to dig into that flakiness separately if it is not already known.

## Evidence

`.omo/evidence/20260818-fix-6167-document-uri-normalize/qa-summary.md`, one file, with the red, green, revert-and-fail, and regression transcripts inlined.

No `package.json` version change. Happy to close this if you would rather land your own.

---
## Summary by cubic
Fixes #6167 so diagnostics are no longer dropped when a language server publishes under a re-encoded URI for the same file, and refreshes the committed `omo-senpi` plugin bundle so the CI freshness gate passes.

- Adds `normalizeDocumentUri`, which decodes percent-encoding first, then uppercases the drive letter on Windows.
- Routes all eight `openByUri` accesses through it; `state.uri` and every outgoing `didOpen`/`didChange`/`didSave`/`didClose` URI stay byte-identical.
- Reworks the test fixture with `realpathSync` so it passes on macOS, where `mkdtemp` returns a symlinked `/var/folders` path.
- Rebuilds the `omo-senpi` bundle with bun 1.4.0 to match CI, since `lsp-core` is a workspace dependency there.

<sup>Written for commit dd721f286966a4f3286faa341c36bbddaad1652d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:
Rebased onto current `dev` (`3abc23c`) and refreshed the committed plugin bundle. @code-yeongyu — this is ready for another look; #6167 is still open and `normalizeDocumentUri` still does not exist anywhere on `dev`, so the defect is live.

**On the red check that stalled this.** The previous run's only failure was `senpi-compatibility (windows-latest)`, and its two failing tests — `task-rpc-e2e.windows` and `memory-usage-wiring` — both live in `packages/omo-senpi`, which this PR never touched. Both are green on `dev` today. It was branch staleness, not the patch.

**On the bundle.** `@oh-my-opencode/lsp-core` is a workspace dependency of `omo-senpi`, so the normalizer lands inside `plugin/extensions/omo.js` and the freshness gate reported `stale-output`. Refreshed in its own commit, matching the `build(omo-senpi): refresh the committed plugin bundle` convention. Built with bun 1.4.0 to match `ci.yml` — worth noting the artifact is toolchain-sensitive but not platform-sensitive: 1.4.0 on Windows reproduces the committed beta.30 bundle byte-for-byte from unmodified sources, while 1.3.14 drifts 142 bytes on identical input. Only `omo.js` changed; the other five entry points rebuilt identically.

**Evidence, re-run against current `dev`:**

| | the two new tests | the three pre-existing tests |
| --- | --- | --- |
| without the source patch | **fail** | pass |
| with it | **pass** | pass |

```
bun test packages/lsp-core/src/lsp/workspace-document-state.test.ts
→ 5 pass, 0 fail, 17 expect() calls
```

Reverting only `workspace-document-state.ts` while keeping the tests reproduces the two failures, so the tests measure the defect rather than the implementation.

Thanks to @MoerAI, whose #6228 established the decode-then-drive-fold order this carries forward — that PR was withdrawn for want of review, not on merit, and the bug has outlived it by a month.

MoerAI (changes requested):
Thanks for the PR, @LilMGenius, and for taking the time on this.

The implementation and current CI look good, but GitHub now reports head `82a6b366` as `DIRTY` / `CONFLICTING` against `dev` at `d50518d4`, so it cannot be merged safely in its current state.

Please rebase onto current `dev`, resolve the conflict, regenerate `packages/omo-senpi/plugin/extensions/omo.js` with `node packages/omo-senpi/plugin/scripts/build-extension.mjs`, verify it with `node packages/omo-senpi/plugin/scripts/build-extension.mjs --check`, and let CI rerun. Happy to re-review and merge once GitHub reports the branch mergeable and the checks are green.

LilMGenius:
@MoerAI Thanks for the clear instructions - all four points are done, and the PR is now `mergeable: true` / `mergeable_state: clean` at `dd721f2`.

**What I did**

- Rebased onto current `dev` (`ee7ae5d66`), keeping only the three source changes.
- Regenerated `packages/omo-senpi/plugin/extensions/omo.js` with `node packages/omo-senpi/plugin/scripts/build-extension.mjs`.
- `--check` reports `omo-senpi extension build is current`.
- CI reran and is fully green, including `senpi-compatibility` on ubuntu/macos/windows.

The diff is back to exactly the original four files:

| file | +/- |
|---|---|
| `.omo/evidence/20260818-fix-6167-document-uri-normalize/qa-summary.md` | +116/-0 |
| `packages/lsp-core/src/lsp/workspace-document-state.test.ts` | +84/-1 |
| `packages/lsp-core/src/lsp/workspace-document-state.ts` | +19/-8 |
| `packages/omo-senpi/plugin/extensions/omo.js` | +2/-2 |

**One observation, offered as feedback rather than a request**

Reproducing that bundle byte-for-byte turned out to be the hard part of this PR, and none of it is discoverable from the outside:

- `omo.js` is committed as a marker line plus one ~984 KB minified line, so every external PR touching `lsp-core` collides textually the moment `dev` regenerates it - which happened 8 times in the last 149 commits, including once while I was working on this.
- The digest only reproduces under **bun 1.4.0 on Linux**. I first built with 1.3.14 and got a stable but different `bodyDigest`, because the bundler emits a different `__export` key order in zod's `util` namespace. The correct version is only visible inside the `senpi-compatibility` job in `ci.yml`; the build is also Linux-only in practice, since the source digest hashes repo-relative paths and Windows backslashes change it.

I ended up running the rebase + regeneration as a GitHub Actions job on my fork so it could be replayed against the newest `dev` in about two minutes. That worked, but it is a lot of infrastructure for a 19-line source fix. If it is ever worth it, generating this bundle in CI instead of committing it - or at least documenting the pinned bun version next to `build-extension.mjs` - would remove a recurring tax on outside contributions.

Happy to rebase again if `dev` moves before you get to it.

MoerAI (approved):
Thanks @LilMGenius. The current head `dd721f28` is conflict-free against `dev`, includes the regenerated Senpi bundle, and CI run 33414641708 completed successfully. The prior changes request is resolved. Approved.

Issue #6866, opened 2026-08-14
[Bug]: TUI sidebar idle "Models" roster ignores disabled_agents — disabled agents still listed (e.g. hephaestus gpt-5.6-sol)


JunyongYao:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

The OMO TUI sidebar's idle-state **"Models"** section (the roster rendered via the `sidebar_content` slot) lists **all** builtin agents and their effective models, but it does not respect `disabled_agents`. An agent that is properly disabled — not registered, not callable via `call_omo_agent` — still shows up in the roster with its fallback-chain model.

In my case, with `disabled_agents: ["hephaestus"]` in `~/.omo/omo.jsonc`, the sidebar idle view still shows:

```
Models
hephaestus gpt-5.6-sol
```

even though the agent itself is correctly absent from registration (registration skips it, and delegating to it fails with the expected "disabled via disabled_agents" error).

### Steps to Reproduce

1. Install oh-my-openagent 4.19.4 as an opencode plugin (server plugin via `opencode.jsonc`, TUI plugin via `tui.json` — both list `oh-my-openagent@latest`).
2. Set in `~/.omo/omo.jsonc`:

```json
{
  "[opencode]": {
    "disabled_agents": ["hephaestus"]
  }
}
```

3. Restart opencode so agent registration re-runs with the new config.
4. Look at the TUI sidebar while idle (no active agents/jobs) — the "Models" section still renders a `hephaestus` row.

### Expected Behavior

The idle "Models" roster should only list agents that are actually enabled. `disabled_agents` is already part of the validated plugin config; the roster resolver should filter rows by it, the same way the registration layer does.

### Actual Behavior

[...]

PR #7022, opened 2026-08-18
fix(tui): hide disabled agents from idle roster

dajiaohuang:
## Summary

- filter disabled agents out of the idle TUI model roster
- normalize disabled names with the same migration aliases used by agent registration
- preserve category rows and enabled agent rows

## Tests and QA

- `npx --yes bun test packages/omo-opencode/src/features/tui-sidebar/roster-resolver.test.ts` (4 passed, 0 failed)
- `npx --yes bun x tsgo --noEmit -p packages/omo-opencode/tsconfig.json`
- targeted Bun bundle of `roster-resolver.ts`
- isolated roster probe covering a legacy disabled-agent alias, an enabled agent, and a category

The repository-wide postinstall build was not used as a gate because it remained blocked cloning the unrelated `nexu-io/open-design` shared-skill upstream on this Windows runner. Focused tests, package typechecking, the targeted bundle, and the isolated probe all pass.

## Risk

Low. The change only filters agent rows after model resolution; category resolution and ordering are unchanged.

Related: #6866

### DIFF
--- a/.omo/evidence/20260819-issue-6866/README.md
+++ b/.omo/evidence/20260819-issue-6866/README.md
@@ -1,0 +1,21 @@
+# Issue #6866 QA evidence
+
+## Scenario
+
+Resolve the idle TUI roster from an isolated project configuration that disables the legacy `Sisyphus` agent name. Verify that the normalized `sisyphus` row is absent while another enabled agent and the `deep` category remain visible.
+
+## Commands and results
+
+- RED: `npx --yes bun test packages/omo-opencode/src/features/tui-sidebar/roster-resolver.test.ts`
+  - 3 passed, 1 failed.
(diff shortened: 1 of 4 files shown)


MoerAI:
Thanks for the fix, @dajiaohuang. The root-cause analysis and `AGENT_NAME_MAP` normalization are correct.

I am closing this as superseded rather than rejected. #7266 carries the same roster filter with equivalent behavior across the #6866 case and legacy aliases, and its current full test, typecheck, build, and compatibility matrix is green. This fork branch only ran lightweight checks.

Nothing unique is lost here. #7022 is the earliest contribution in this lineage and should be credited when #7266 lands and closes #6866.

Issue #6796, opened 2026-08-12
[Bug]: runtime_fallback never triggers on ContentFilterError — run stops silently with no fallback


helixzz:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues and discussions to avoid duplicates
- [x] This issue is specific to oh-my-openagent (not OpenCode core)
- [x] I have read the documentation

## Summary

`ContentFilterError` ("The response was blocked by the provider's content filter") is not matched by `RETRYABLE_ERROR_PATTERNS`, and it is not a status code, so `retry_on_errors` cannot cover it either. As a result `runtime_fallback` never advances the chain: the run simply stops with no output and waits for a human.

Content-filter blocks are arguably the *most* worthwhile error class to retry, because they are non-deterministic: the same request with the same context usually succeeds on a different sampling path or on a different model. Unlike a quota error, nothing is actually exhausted.

## Environment

- oh-my-openagent **4.19.3**
- OpenCode **1.18.11**
- `runtime_fallback`: `enabled: true`, `retry_on_errors: [408, 429, 500, 502, 503, 504, 524, 529]`, `max_fallback_attempts: 3`, `notify_on_fallback: true`

## Evidence from my logs

Measured over all assistant messages in my local OpenCode DB, filtering on `error.name = 'ContentFilterError'`:

| provider/model | assistant msgs | blocked | rate |
| --- | --- | --- | --- |
| `modelverse-anthropic/claude-fable-5` | 2181 | 4 | 0.183% |
| `modelverse/claude-fable-5` | 3395 | 6 | 0.177% |
| `modelverse/kimi-k3` | 2187 | 2 | 0.091% |
| `azure-foundry/gpt-5.6-terra` | 22127 | 7 | 0.032% |
| `azure-foundry/gpt-5.6-sol` | 9236 | 2 | 0.022% |
| `amazon-bedrock/anthropic.claude-opus-5` | 16967 | 0 | 0.000% |
| `modelverse/claude-opus-4-7` | 11861 | 0 | 0.000% |

21 blocks total. The rate is low per message, but the **consequence is disproportionate**: of those 21,

- **17** were followed by a `user` message — i.e. a human had to notice and manually resume;
- **4** had no successor at all — the session ended there.

[...]

PR #7023, opened 2026-08-18
fix(runtime-fallback): retry content filter errors

dajiaohuang:
## Summary

Treat provider content-filter responses as retryable runtime fallback errors. The shared model-core pattern now recognizes `ContentFilterError` payloads whose message says the response was blocked by a content filter, so the OpenCode adapter can advance its configured fallback chain instead of stopping silently.

The regression test also pins an unrelated HTTP 400 validation error as non-retryable.

## Tests / QA

- Failing-first proof: the new focused test initially failed with expected `[true, false]` and received `[false, false]`.
- `bun test packages/model-core/src/runtime-fallback-content-filter-error.test.ts packages/model-core/src/runtime-fallback-error-classifier.test.ts` (7 passed)
- `bun run --cwd packages/model-core test` (357 passed)
- `bun test packages/omo-opencode/src/hooks/runtime-fallback/error-classifier.test.ts packages/omo-opencode/src/hooks/runtime-fallback/session-status-handler.test.ts` (35 passed)
- `bun run --cwd packages/model-core typecheck`
- `bun build packages/model-core/src/index.ts --target bun --outfile <isolated temp file>` (35 modules bundled; artifact removed after verification)
- Data-shaped classifier probe: ContentFilterError returned `retryable: true`; validation, abort, and context-overflow cases returned `retryable: false`.
- Sanitized local evidence: `.omo/evidence/20260819-issue-6796/`

## Risk

Low. This adds one narrow built-in message pattern and does not change status-code handling, fallback state, configuration, or dependencies. The main risk is a false-positive retry for an unrelated message containing `content filter`; the validation control and existing abort/context-overflow tests remain green.

Related: #6796

### DIFF
--- a/packages/model-core/src/runtime-fallback-content-filter-error.test.ts
+++ b/packages/model-core/src/runtime-fallback-content-filter-error.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, test } from "bun:test"
+
+import { isRuntimeFallbackRetryableError } from "./runtime-fallback-error-classifier"
+
+const DEFAULT_RETRY_CODES = [429, 500, 502, 503, 504] as const
+
+describe("runtime fallback content filter errors", () => {
+  test("retries provider ContentFilterError payloads without broadening validation errors", () => {
+    //#given
+    const errors = [
+      {
+        name: "ContentFilterError",
+        message: "The response was blocked by the provider's content filter",
+      },
+      {
+        name: "ValidationError",
+        statusCode: 400,
+        message: "Invalid request payload",
+      },
+    ] as const
+
+    //#when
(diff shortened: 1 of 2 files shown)


MoerAI:
Thanks for the correct diagnosis and focused fix, @dajiaohuang.

I am closing this as superseded by #7257, which carries a byte-identical diff with the full CI matrix green. This fork branch never ran build, test, or typecheck and is substantially behind `dev`; the owner branch is the landing vehicle.

The abort and context-overflow guards remain intact, and credit for the content-filter retry fix belongs to this contribution. #6796 should close when #7257 lands.

PR #7027, opened 2026-08-18
fix(senpi): gate Windows RPC external termination facts

dajiaohuang:
## Summary

- evaluate externally terminated RPC task records against facts the host platform can actually preserve
- require a successful termination injection and exact Windows code-1 diagnostic, while keeping POSIX `killed:true` semantics
- make the Windows production wrapper assert the external-termination subcheck instead of only printing it

## Context

Issue #6976 records repeated `kill_marks_error_killed_true` failures on Windows. Node reports the injected `SIGKILL` there as exit code 1 with no signal, so the production classifier correctly preserves it as a crash-shaped record rather than inventing signal provenance.

This change is limited to the QA contract. It does not change the production exit classifier, dependencies, public APIs, or timeout budgets.

## Validation

- focused scenario + RPC classifier/process tests: 27 passed, 1 pre-existing Windows signal test skipped, 0 failed, 55 assertions
- `tsgo --noEmit` passed for `packages/omo-senpi` and `packages/senpi-task`
- local `build:senpi-plugin` passed
- `git diff --check` passed
- isolated Windows wrapper preserved credential and directory digests, removed its sandbox, and leaked zero PIDs

## Residual verification

The current `dev` branch fails the isolated Windows wrapper before this scenario can inject termination: no task record is persisted (`wiringFixed=false`). The same separate signature is recorded in the latest #6976 maintainer comment and current Windows Senpi CI. The wrapper remains fail-closed, and the external-termination path should be rerun after that prerequisite regression is repaired.

The best-effort broad package suites were stopped after 12 CPU-active minutes without a final summary; one observed existing failure was a Windows symlink `EPERM` on a host without developer-mode privilege. Focused tests, both package typechecks, and the packaged Senpi build completed successfully.

### DIFF
--- a/packages/omo-senpi/scripts/qa/task-rpc-e2e-scenarios.mjs
+++ b/packages/omo-senpi/scripts/qa/task-rpc-e2e-scenarios.mjs
@@ -90,6 +90,13 @@ export async function killSenpiHost(child, terminate = killProcessGroup) {
   return terminate(child.pid)
 }
 
+export function externalTerminationMatchesPlatformFacts(record, platform, terminationInjected) {
+  if (!terminationInjected) return false
+  if (record?.status !== "error") return false
+  if (record.killed === true) return true
+  return platform === "win32" && record.error_message === "RPC child exited with code 1"
+}
+
 function waitForChildClose(child, timeoutMs) {
   if (child.exitCode !== null || child.signalCode !== null) return Promise.resolve()
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


code-yeongyu:
Heads-up: the underlying product defect this PR worked around is now fixed. #7052 (merged) makes the exit classifier recognize the win32 external-termination shape (code 1 + null signal + empty stderr) and record `killed:true`, verified on the full-matrix run — `kill_marks_error_killed_true` now passes with `{killed: true, error_excerpt: "RPC child terminated externally with exit code 1 (pid=...)"}`.

With that landed, the QA-contract gating here (accepting the code-1 diagnostic as the Windows fact) is no longer the direction we want — the production record is now correct, so the original strict contract (`killed:true`) holds on all platforms again. Thanks for the careful analysis here; the platform-facts framing in this PR is part of what made the root cause unambiguous. Closing as superseded by #7052.

Issue #6854, opened 2026-08-13
[Bug]: task(category=...) category is invisible in TUI — delegated subagents always labeled "Sisyphus-Junior", making it look like the orchestrator skipped the category


itokun99:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

When an orchestrator (e.g. Atlas) delegates via `task(category="unspecified-high", ...)`, the spawned subagent is **always labeled "Sisyphus-Junior" in the TUI** — the category that was passed is never surfaced anywhere in the session header. After migrating the `categories` block so that category→model mappings use plain model IDs (e.g. `unspecified-high → opencode-go/kimi-k2.7-code`), the category is completely invisible: the subagent header shows only `Sisyphus-Junior` + a model name that no longer encodes the category.

This makes it look like the orchestrator "delegated without a category" (and therefore without the category-optimized model), when in fact the category **is** being passed and the correct model **is** being used.

### Evidence that the category IS passed (from OpenCode session DB)

Atlas session `ses_0053ab72` (oh-my-openagent 4.19.4, model `opencode-go/minimax-m3`): **all 24 `task()` calls** carry `category: "unspecified-high"` + `load_skills` + `description` + `prompt` + `run_in_background`. The spawned child sessions all have `agent = "Sisyphus-Junior"` and `model = {"id":"kimi-k2.7-code","providerID":"opencode-go"}` (the category's configured model) — i.e. category routing works correctly at runtime, but nothing in the UI tells the user which category was used.

The only calls ever recorded without `category`/`subagent_type` were aborted tool executions (`status: "error"`, `"Tool execution aborted"`, empty input) — not real delegations.

### Root cause

[...]

PR #7028, opened 2026-08-18
fix(task): show categories in delegated session titles

dajiaohuang:
Related: #6854

## Summary

- prefix categorized delegated child-session titles with `[category]`
- preserve the existing OpenCode-compatible agent suffix and category-less behavior
- keep the category visible across initial, retry, and background delegation paths

## Verification

- focused delegation and background suites: 31 passed
- routing suite: 4 passed
- repository typecheck
- targeted OpenCode bundle build (1,987 modules)
- isolated OpenCode 1.18.18 run persisted a `[quick]` child title

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/manager-session-permission.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/manager-session-permission.test.ts
@@ -121,4 +121,51 @@ describe("BackgroundManager session permission", () => {
       ],
     })
   })
+
+  test("prefixes the category in the child session title", async () => {
+    // given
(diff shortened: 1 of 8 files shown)


MoerAI:
Thanks for this, @dajiaohuang. The implementation and tests are correct.

I am closing this as superseded by #7256, which carries the byte-identical eight-file change with the current full CI matrix green. Your data-plumbing design is the adopted one: category is formatted only at session-title boundaries while description remains untouched.

This is a sequencing outcome, not a rejection. #6854 will be closed through #7256, and this contribution should be credited there.

PR #7037, opened 2026-08-19
feat(omo-senpi): unify memory TUI surfaces on senpi's cache-warm notice style

code-yeongyu:
## What changed

Every memory-related TUI surface in omo-senpi now renders in senpi's cache-warm notice style (glyph + bold tone-coloured title, dim why-sentence, per-tone extra lines, expanded-only detail), and the `memory` / `memory_apply_patch` tool result row stops showing the terse `Memory <command> committed locally (<sha7>).` line.

| Area | Commit | What it does |
|---|---|---|
| sized ls-tree | `20985fe53` | `GitMemoryRepo.lsTreeSized` parses one `git ls-tree -r -l -z` call; the system-token estimator drops its per-path `git show` loop with byte-identical output (characterization-tested) |
| reflection/health/nudge parity | `0c388091c` | all five transcript-entry renderers audited against the notice contract; health why-line sentence-wraps; accepted-turns renderer brought to contract (deliberately unregistered) |
| shared snapshot builder | `b7972588a` | `buildMemorySnapshot` extracted from the RPC bridge closure; additive `repo.{totalBytes,systemBytes,fileCount,entriesToday,previousEntryAtISO}` + `reflection.lastConsolidationAtISO`; **schemaVersion stays 1**, v1 field set pinned by shape tests |
| write-result notice | `42e73afe0` | the tool row self-renders via `ToolDefinition.renderResult`; gated by new `memory.write_notice.enabled` (default true, per-agent override); model-facing text byte-identical |
| soul-entry contract | `f3dfe8e34` | soul-updated entry moves from plain lines to the notice contract |
| MCP surface | `b287390a8` | shared consumer reads each commit receipt **exactly once** and emits soul + write notices from that read; both surfaces show identical numbers by construction |
| visual QA driver | `dfa3de422` | `scripts/qa/memory-write-visual-qa.mjs` drives the real senpi TUI and asserts the rendered row |
| bundles | `b28ede5ed` | regenerated extension bundles |

## Observed behavior

Real senpi TUI (isolated agent dir + `OMO_MEMORY_HOME`, mock provider scripted to call the memory tool), captured through the repo's xterm.js web-terminal QA:

```
memory
● Memory updated · 2nd entry today
Added 4 lines to knowledge/deploy.md.
system 2.0K injected · 5.5K total · 4 files
last entry just now
```

Collapsed: no command name, no `committed locally`. Expanded (`ctrl+o`): the detail line carries `<sha7> · <identity> · <subject>`. The timeline line turns `warning` past 7 days unconsolidated or 25 unreflected steps. Evidence (screenshot + transcript + ANSI) under `.omo/evidence/20260819-memory-notice-style/` in the worktree; the driver is repeatable via `node packages/omo-senpi/scripts/qa/memory-write-visual-qa.mjs` (7/7 assertions pass).

## QA / evidence

Every area landed RED-first (failing test captured before implementation), then GREEN, and was re-verified independently before its commit:

- `packages/memory-core`: **544 pass / 0 fail**; tsgo clean
- `packages/omo-senpi` memory scope: **887 pass / 0 fail** (verified after merging `origin/dev`). Before that merge this scope showed 886 pass / 1 fail on the pre-existing `/doctor` `[warn] reflection-health` case (#6976); `dev` now carries #7034, which fixes that fixture, so the scope is fully green
- `packages/omo-config-core`: **178 pass / 0 fail**; tsgo clean
- Full omo-senpi suite (`bun test src/`, 274 files), measured before the `dev` merge: 1925 pass / 20 fail — **all 20 proven pre-existing** by running the same suites on a clean `origin/dev` checkout (byte-identical failure set; none of the failing files appear in the branch diff). One of those 20 was the `/doctor` case now fixed on `dev` by #7034
- Visual QA: 7/7 assertions on the real TUI (collapsed contract, bold title, no old line, no command name, path in why, sha + subject in expanded detail)

Degradation paths are unit-tested one per case: no prior commit, no reflection completions, corrupt `state.json`, git failure — the row renders with fields omitted, `execute` never throws, and gathering is time-bounded so a wedged probe degrades the row instead of holding the tool result.

## Residual risk

- **Desktop contract**: `MemoryRpcSnapshot` stays `schemaVersion: 1`; every new field is additive and optional, and shape tests pin the v1 field set. Desktop v1 readers are unaffected.
- **MCP surface**: receipts remain read-and-delete-once; the shared consumer is the only reader, covered by a single-consumption test.
- 19 further full-suite failures exist on clean `dev` (package-shape, cli-local, session-start ordering, ultrawork, lsp-architecture, init-deep-advisor, product-identity), independent of #6976 — worth tracking separately; this PR neither touches nor fixes them.
- This branch merges `origin/dev` (25 commits) at `460b456e1`; the only conflict was the generated `plugin/extensions/omo.js`, resolved by regenerating the bundle from merged source rather than hand-editing minified output.


code-yeongyu:
### CI note: `test (windows-latest, 1/2)` failures are runner flakes, not this branch

Two runs on head `b2c54ffa2` failed that shard on **different** tests, each a timeout:

| run | failing test | duration |
|---|---|---|
| initial | `compileMemoryBlock > … structural contract` (`packages/memory-core/src/compile/compile.test.ts`) | 5004ms |
| rerun | `loadMcpConfigs > … only servers matching the current project path are loaded` | 9272ms |

Evidence this is not the branch:

- **Neither test file is in the diff.** `git diff --name-only origin/dev...HEAD` contains no `compile.test.ts` and no MCP-config file. This branch's only `memory-core` change is an additive `lsTreeSized` method plus its return type; the compile path never calls it and no existing signature changed.
- **Different test each run.** A regression fails the same assertion deterministically; these are timing-bound timeouts that move between runs.
- **Same-commit siblings are green**: `test (windows-latest, 2/2)`, `test (macos-latest)`, `test (ubuntu-latest)`, `senpi-compatibility` on all three platforms, `build`, `typecheck`, `format-lint-typecheck-build`, `ci-mode`.
- **Repo-wide right now**: six other active branches show `failure` on this same workflow.
- `dev` carries several recent commits raising Windows timing budgets for exactly this class (`test(memory-core): budget Windows git apply-patch suites past the 5s default`, `raise the facts-launch ceilings the windows runner overshoots`, `give the MCP server loader hooks a windows budget`).

Per this PR's scope discipline these are documented, not "fixed" here.

One earlier ubuntu failure **was** real and is fixed: `senpi-compatibility (ubuntu-latest)` reported `omo-senpi extension build is not current: stale-output`, because esbuild output is platform-dependent and the bundles had been regenerated on macOS. Rebuilt in a `linux/amd64` container (bun 1.3.14 / node 20) in `b2c54ffa2`; that job now passes.

PR #7042, opened 2026-08-19
feat(omo-opencode): render background task notifications as markdown

Zaldaryon:
## Summary

- Background task notifications are written in markdown, and a client that renders only assistant output prints them verbatim. The person watching the session reads the raw `<system-reminder>` block, `**ID:**`, backticks and all, sitting next to assistant text that is rendered.
- The notification part now carries `metadata.render` set to `"markdown"`, which lets a client render it. Clients that ignore the hint behave exactly as they do today, and the model receives the same text either way.

## Changes

- `packages/utils/src/internal-initiator-marker.ts`: adds `createRenderedInternalAgentTextPart`, which is `createInternalAgentTextPart` plus `metadata: { render: "markdown" }`. The existing helper is untouched, so task prompts sent to child sessions keep their current shape.
- `packages/omo-opencode/src/features/background-agent/parent-wake-prompt-dispatch.ts`: `sendParentWakePrompt` builds the notification part with the new helper on both the reply and the no-reply path.
- Tests for the helper, for the hint surviving `withInternalNoReplyMarker`, and for the part that `ParentWakeNotifier` actually dispatches.

Why the hint and not `synthetic: true`: synthetic hides the part, which is right for text meant only for the model. A completion notice is also for the person watching, so it needs to stay visible and be readable.

## QA & Evidence

- **What was tested:** `bun run typecheck`
  **Observed result:** exit 0, no diagnostics
  **Artifact:** console output
  **Why sufficient:** covers the new export and both call sites across every workspace project

- **What was tested:** `bun run build`
  **Observed result:** `build: all steps completed`, exit 0
  **Artifact:** console output
  **Why sufficient:** the generated bundles rebuild cleanly with the change; regenerated dist files are deliberately left out of this PR

- **What was tested:** `bun test packages/omo-opencode/src/shared/internal-initiator-marker.test.ts packages/omo-opencode/src/features/background-agent/parent-wake-notification-render-hint.test.ts`
  **Observed result:** 27 pass, 0 fail, 54 expect() calls
  **Artifact:** console output
  **Why sufficient:** the five new cases pin the hint on the helper, through the no-reply marker, and on the part `ParentWakeNotifier` dispatches, next to the existing marker cases that guard the untouched helper

- **What was tested:** the root `bun test` suite, on this branch and on the parent commit
  **Observed result:** 15735 pass, 43 skip, 9 fail here against 15729 pass, 43 skip, 10 fail on the base. No test fails only on this branch
  **Artifact:** the failing test names and the diff are in a comment on this PR
  **Why sufficient:** it separates pre-existing environment failures from anything this change could have caused

## Risks & Residuals

- A client that does not read `metadata` renders the notification exactly as it does now. Mitigated: the hint is additive and nothing branches on it inside this repo.
- `createInternalAgentContinuationTextPart` sets its own `metadata` for compaction. Not applicable: it builds on the untouched base helper and is synthetic, so it is hidden regardless.
- The reader side does not exist upstream yet. Accepted: anomalyco/opencode#43319 adds it, and this change is useful on its own the moment any client honors the hint. Neither PR depends on the other landing first.

## Automated Checks

```bash
bun run typecheck
bun run build
bun test
```

## Related Issues

Reader side of the same problem: anomalyco/opencode#43319 and anomalyco/opencode#43318.


Zaldaryon:

Zaldaryon:

Zaldaryon:
Root suite result, as promised in the PR body.

On this branch: 15735 pass, 43 skip, 9 fail across 15787 tests in 2064 files.

Nine failures sounded bad, so I ran the same suite on the parent commit and diffed the failing test names. The base has 10 failures. Every failure on this branch also fails on the base, and no test fails only here:

```
$ comm -23 branch.fails base.fails    # unique to this branch
(nothing)
```

The eight shared failures, deduplicated:

```
#given PowerShell shell detection in injectGitMasterConfig > #then emits $env: prefix syntax in pwsh code block
injectGitMasterConfig > uses PowerShell routing when PowerShell is detected
shell-env > detectShellType > #given PSModulePath is set without SHELL #then returns powershell
shell-env > detectShellType > #given Windows platform without PSModulePath #then returns cmd
#given the generated Codex installer #when release versions are synchronized #then its embedded package version matches the root release version
omo-ai payload verifier > #then it passes with no missing-artifact error
/doctor > #given repeated model-not-found reflection failures #then reflection health and remediation are reported
claude-code-agent-loader > loadOpencodeGlobalAgents > returns empty object when pointed at dir without agents/
```

Four are shell detection asserting Windows behavior, two check packaging artifacts that a fork checkout does not carry, and two are assertions on local state. My environment is WSL2 on Linux with Bun 1.3.14, so I read all eight as environment rather than as anything this branch touches. Say the word if any of them is meant to pass on Linux and I will look closer.

The tenth failure, `preflightMemoryModels > #given a launcher whose grandchild holds the output pipes`, failed on the base and passed here. It looks timing dependent rather than related to this change.

PR #7090, opened 2026-08-21
fix(config-watch): respect anchored protected path filters

### LINKED ISSUE #7064: config-watch: default ~/.omo/agent layout drops the ~/.omo watch target — "user config discovery requires reload" on every session, 0 targets when cwd=$HOME
## Summary

On a **default** install, `config-watch` drops the user-config watch target for `~/.omo`, so every session logs:

```
config-watch user config discovery requires reload { userConfigCreationDiscovery: 'reload_required' }
```

This is not the "user config dir does not exist" case that the warning was designed for — `~/.omo` **exists and holds `omo.jsonc`**. The target is discarded by the Senpi protected-path filter because the canonical agent-state directory is nested *inside* the user-config directory.

Consequences on a stock install:

- `~/.omo/omo.jsonc` is **never hot-reloaded**; the user must restart `omo` after editing user-scope config.
- When `cwd === $HOME`, `resolveOmoConfigWatchTargetResolution` returns **zero** targets, so project-scope `.omo` hot-reload is dead too.

## Root cause

`packages/omo-senpi/src/components/config-watch/paths.ts`

```ts
// L67-70
function resolveSenpiProtectedPaths(env: OmoConfigEnv): readonly string[] {
  const agentDir = resolveAgentHome({ env, homeDir: resolveHomeDir(env) })
  return [join(agentDir, "auth.json"), join(agentDir, "sessions"), join(agentDir, "logs")]
}

// L72-77 — drops a target that CONTAINS a protected path, as well as one contained by it
function isSenpiRestrictedTarget(path: string, protectedPaths: readonly string[]): boolean {
  const resolvedPath = resolve(path)
  return protectedPaths.some(
    (protectedPath) => containsPath(resolvedPath, protectedPath) || containsPath(protectedPath, resolvedPath),
  )
}
```

The agent-state directory defaults to `~/.omo/agent` — `bin/lib/agent-dir.js`:

```js
export function defaultAgentDir(home) {
  return join(home, ".omo", "agent")
}
```

So on a default install:

| value | path |
|---|---|
| `userConfigDirectory` | `~/.omo` |
| agent dir (`resolveAgentHome`) | `~/.omo/agent` |
| protected paths | `~/.omo/agent/auth.json`, `~/.omo/agent/sessions`, `~/.omo/agent/logs` |

`containsPath("~/.omo", "~/.omo/agent/auth.json")` is `true`, therefore `isSenpiRestrictedTarget("~/.omo", …)` is `true` and the `configTarget(userConfigDirectory)` pushed at L153 is filtered out at L169. `$HOME` and `/` are dropped for the same reason, so `userConfigCreationWatched` (L174-176) is `false` and the discovery is reported as `reload_required`.

The existing comment at L171-173 anticipates `$HOME` being dropped, but not `~/.omo` itself:

```ts
// Derived from the SURVIVING targets, not from directory existence: `~/.omo`'s parent
// is `$HOME`, whose creation target is always dropped by the protected-path filter, so
// an existence check would report a watch that was never registered.
```

The key point: the discarded target's `filterGlobs` are **anchored to the watch root** —

```ts
// L22
export const OMO_CONFIG_FILE_FILTER_GLOBS = ["/omo.jsonc", "/omo.json"] as const
```

`/omo.jsonc` and `/omo.json` can only ever match `~/.omo/omo.jsonc` and `~/.omo/omo.json`. They can never match anything under `~/.omo/agent/`, so watching `~/.omo` cannot surface `auth.json`, `sessions/`, or `logs/`. The filter ignores `filterGlobs` and rejects the whole target on a purely lexical ancestor test.

The comment above the constants (L17-21) makes the same point from the performance side — anchoring exists precisely so a `.omo` target does not walk the runtime state beneath it:

```ts
// Root-anchored: senpi maps a `dir` target to a RECURSIVE watch, so unanchored
// globs make it hash the entire subtree. A `.omo` directory also holds runtime
// state (senpi-task/ and runtime/ including runtime/ast-grep/), which reached tens of thousands of
// files and gigabytes of hashing per watcher rebuild. Config lives only at the
// top level, so anchoring keeps the scan to those two files.
```

Since anchoring already confines the scan to the two top-level files, the protected-path filter has nothing left to protect against for these targets — it only removes them.

## Reproduction

Default layout, no `OMO_CODING_AGENT_DIR` / `SENPI_CODING_AGENT_DIR` / `PI_CODING_AGENT_DIR` set:

```
~/.omo/omo.jsonc            # user config
~/.omo/agent/settings.json  # makes resolveAgentHome pick ~/.omo/agent
~/.omo/agent/auth.json
~/.omo/agent/sessions/
~/.omo/agent/logs/
```

Run `omo` → the warning is logged once per session (visible in `~/.omo/agent/omo-debug.log`).

I isolated it by registering the shipped `config-watch` component from `plugin/extensions/omo.js` against a stub logger/event bus and printing the resulting registration payload.

<details>
<summary>probe harness</summary>

```js
// probe.mjs — run with: node probe.mjs
const BUNDLE = process.env.CW_BUNDLE
  ?? `${process.env.HOME}/.nvm/versions/node/<ver>/lib/node_modules/omo-ai/plugin/extensions/omo.js`;
const mod = await import(BUNDLE);
const comp = mod.omoSenpiComponents.find((c) => c.name === "config-watch");
const warns = [];
let registration;
const events = {
  emit(ch, payload) { if (ch === "config-watch:register") registration = payload; },
  on() { return () => {}; },
};
await comp.register(
  { events, on() {} },
  { logger: { warn: (m, d) => warns.push({ m, d }), info() {}, error() {} } },
);
console.log("warnings:", JSON.stringify(warns));
console.log("targets :", registration.targets.length);
for (const t of registration.targets) console.log(" -", t.path, JSON.stringify(t.filterGlobs));
```

(Standalone execution needs an ESM `resolve` hook that retries bare specifiers against
`omo-ai/node_modules/@code-yeongyu/senpi/node_modules/`, since the bundle imports
`@earendil-works/pi-tui`.)

</details>

### Observed

| # | version | env | `cwd` | warning | targets | `~/.omo` watched |
|---|---|---|---|---|---|---|
| 1 | 5.0.0-0.beta.11 | default | project dir | fires | 3 | no |
| 2 | 5.0.0-0.beta.11 | default | `$HOME` | fires | **0** | no |
| 3 | 5.0.0-0.beta.11 | `OMO_CODING_AGENT_DIR=~/.omo-agent` | `$HOME` | **none** | 2 | **yes** |
| 4 | 5.0.0-0.beta.12 | default | `$HOME` | fires | **0** | no |
| 5 | 5.0.0-0.beta.12 | `OMO_CODING_AGENT_DIR=~/.omo-agent` | `$HOME` | **none** | 2 | **yes** |

Rows 3 and 5 pin the mechanism: moving the agent dir out of `~/.omo` removes the lexical containment, and the target survives. Rows 2 and 4 are the ones that matter — a stock install launched from `$HOME` registers no watch targets at all.

I also reproduced rows 2 and 3 against a synthetic `HOME` containing only the five paths listed above, so nothing about my particular machine is involved.

## Suggested fix

Make the protected-path filter respect the target's anchored `filterGlobs`: a target should be dropped only when a protected path could actually be matched by them, or when the target itself lies inside a protected path. Something like:

```ts
function isSenpiRestrictedTarget(
  target: OmoConfigWatchTarget,
  protectedPaths: readonly string[],
): boolean {
  const resolvedPath = resolve(target.path)
  return protectedPaths.some(
    (protectedPath) =>
      containsPath(protectedPath, resolvedPath) ||
      (containsPath(resolvedPath, protectedPath) &&
        matchesConfigWatchFilter(protectedPath, target.filterGlobs)),
  )
}
```

The senpi package already ships exactly this predicate as `matchesConfigWatchFilter` (`core/extensions/builtin/config-reload/protocol`), which honours the leading-`/` anchor; it is not currently imported here, so this would mean either re-exporting it or duplicating the small anchored-glob check in `paths.ts`. With it, `~/.omo` carrying `["/omo.jsonc", "/omo.json"]` is kept, while a target that really could expose `auth.json` is still rejected. Pruning protected subtrees inside the watch engine instead of discarding whole targets would work equally well.

A regression test worth adding alongside the existing cases in `paths.test.ts`: with `~/.omo` existing **and** the agent dir defaulting to `~/.omo/agent`, expect `userConfigCreationDiscovery === "watched"` and `~/.omo` present in `targets`.

## Workaround

Point the agent dir outside the config dir, e.g. `export OMO_CODING_AGENT_DIR="$HOME/.omo-agent"` (moving the existing `~/.omo/agent` there). This restores the watch, but it diverges from the unified layout `bin/lib/agent-dir.js` deliberately established and disables `adoptLegacyFlatState`, so it is not something users should have to do.

## Environment

- `omo-ai` 5.0.0-0.beta.11 and 5.0.0-0.beta.12 (`beta` dist-tag), installed globally via npm
- Linux, bash, Node v22.22.2 (nvm)
- No `OMO_*` / `SENPI_*` / `PI_*` agent-dir env vars set

Note on Node: the package declares `engines: node >= 24` and I am on 22, but the defect is pure lexical path arithmetic in `resolveOmoConfigWatchTargetResolution` — no filesystem or runtime behaviour is involved in the containment test — and it is equally visible by reading `paths.ts` against `defaultAgentDir`. Rows 1-5 above are all from the same runtime, so the comparison is internally consistent.

haimingZZ:
## Summary

Fixes #7064.

The protected-path filter previously rejected any watch target that was an ancestor of Senpi's protected agent paths. This discarded the default `~/.omo` config target because the target's anchored globs only match the top-level `omo.jsonc` and `omo.json` files, not `~/.omo/agent` state.

The filter now:

- rejects targets inside a protected path;
- rejects an ancestor only when its watch globs can match the protected path;
- preserves anchored config and `.omo` discovery targets that cannot expose protected state.

The matching logic mirrors Senpi's supported plain-name, suffix, and root-anchored filter semantics without adding a runtime import of Senpi's non-exported internal module.

## Validation

- `bun test packages/omo-senpi/src/components/config-watch/paths.test.ts`
- `bun run --cwd packages/omo-senpi typecheck`

The regression coverage includes the default `~/.omo/agent` layout, an existing `~/.omo` while starting from `$HOME`, missing user config creation, and an explicit agent directory under `HOME`.

---
## Summary by cubic
Respects anchored protected-path filters in `omo-senpi` config watch so safe ancestors are kept. Previously any ancestor of the protected agent paths was dropped (losing `~/.omo` and requiring reload); now an ancestor is dropped only when its globs can match a protected path.

- Implements plain-name, suffix (*) and root-anchored (/) glob matching to decide if an ancestor target is restricted.
- Watches anchored `~/.omo` creation from `$HOME`; keeps `$HOME` and other ancestors when their globs are anchored, while still avoiding agent-directory targets.
- Mirrors Senpi’s filter semantics and updates tests for default, nested, missing-user-config, and explicit agent-dir layouts.
- Refreshes `packages/omo-senpi/plugin/extensions/omo.js` after merging `dev`; no behavioral changes expected.

<sup>Written for commit c56a6a53b096012216e0309343b9bffe083fd57d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7090?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Closing as superseded by merged #7389, which delivered the compatible client-and-host safe-filter contract and closed #7064:
- https://github.com/code-yeongyu/oh-my-openagent/pull/7389
- https://github.com/code-yeongyu/oh-my-openagent/issues/7064#issuecomment-5433396252

This older branch is conflicting and should not be revived unchanged. It watches `/omo` instead of the user `.omo` root, does not fail closed on all anchored protected-path intersections, and its tests miss the actual `.omo/agent` sentinel layout.

Reopen only with a current-host reproduction not covered by #7389 and a replacement test that exercises the real resolved config layout.

PR #7106, opened 2026-08-21
fix(senpi-task): keep revived DAG nodes in the wave await set

code-yeongyu:
## Root cause

The strict-wave scheduler kept attached task settlements in a map local to `admitAndSettleWave`. Once node B failed, its original settlement was removed from that map. `dag send` then revived B through `node-send.ts`, but that path armed a separate out-of-wave watcher instead of registering B's new task epoch with the active wave. When sibling A settled, the wave await set looked empty, so the scheduler emitted `dag.wave.completed` and `dag.run.completed` while B was running and dependent C was pending.

## Fix

- Move attached settlements into scheduler-owned state so live-wave revivals rejoin the same await registry.
- Wake an in-progress settlement race when a revive registers a new task epoch.
- Resolve `sendToNode(...).settled` only after the scheduler folds the revived outcome.
- Rescan waves after concurrently attached work settles, allowing newly runnable dependents to dispatch.
- Refuse to emit a run terminal event unless every node is terminal.
- Preserve the existing standalone revive watcher for sends against an already-settled run.

This prevents the stranded state rather than broadening retry semantics: a live run cannot terminalize with running/pending nodes, and a revived success naturally schedules dependents whose prerequisites are complete.

## Live journal evidence

Run `dag_2d12c2f7-38f3-487e-bead-2ceafd0ab87e`:

- seq 10: B failed
- seq 11-12: `dag send` steered/revived B to running
- seq 13: A completed
- seq 14-15: wave and run incorrectly completed with counts `completed:1 running:1 pending:1`
- seq 18: B completed after the run was already terminal
- C was never scheduled

Copies are committed under `.omo/evidence/20260821-dag-revive-terminalization/incident-events.jsonl` and `incident-run.json`.

## Failing-first regression

The new three-node end-to-end regression drives the real send/revive path: B fails, B is revived, then A succeeds.

RED before production changes (`.omo/evidence/20260821-dag-revive-terminalization/red.txt`):

```text
Expected: "running"
Received: "completed"
(fail) ... #then the run awaits the revival and schedules their dependent
16 pass
1 fail
exit_code=1
```

GREEN after the fix (`.omo/evidence/20260821-dag-revive-terminalization/green.txt`):

```text
17 pass
0 fail
113 expect() calls
exit_code=0
```

The regression then proves B's revived success schedules C and exactly one terminal event lands only after C completes.

## Verification

- Focused scheduler + DAG E2E: 67 pass, 0 fail (`focused-final.txt`)
- Full `packages/senpi-task` suite: 1712 pass, 1 platform skip, 0 fail (`package-suite-final.txt`)
- `bun run --cwd packages/senpi-task typecheck`: pass (`typecheck-final.txt`)

All captures live under `.omo/evidence/20260821-dag-revive-terminalization/`.

---
## Summary by cubic
Prevents a live DAG run from completing while a revived node is still running. Previously, reviving a failed node armed an out‑of‑wave watcher and the wave await set emptied, so the scheduler emitted wave/run completion early; now revived task epochs rejoin the active wave and terminalization waits for all nodes.

- Keep attached task settlements in scheduler-owned maps across the run and wake a wave waiter when attachments change; rescan waves after concurrent attachments settle.
- Register revived epochs inside the scheduler (`watchRevivedInScheduler`); `sendToDagNode` accepts an optional watcher so `sendToNode(...).settled` resolves only after the scheduler folds the revived outcome.
- Guard terminalization: do not emit a run terminal event unless every node is terminal; throw if a run would terminalize with active nodes.
- Add an end-to-end regression in `packages/senpi-task` that reproduces the incident and proves a revived success schedules dependents and yields exactly one terminal event.

<sup>Written for commit bba58c995e840f68f15313abf0c9e971d09840c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
CI note: all scheduler/package/typecheck/test/build jobs pass. The sole failure is the Linux-only committed Senpi bundle freshness check (`stale-output`) because this `packages/senpi-task` source change is included in the generated `packages/omo-senpi/plugin/extensions/omo-task.js` artifact. The task scope explicitly forbids writing outside `packages/senpi-task/**` plus evidence, so the generated bundle was intentionally not changed here. The artifact is current on origin/dev; a separate generated-bundle refresh is required by the normal release lane.

PR #7109, opened 2026-08-21
fix(tmux): retry stale sweep when session listing fails

suyua9:
## Summary
- preserve the tmux list-sessions diagnostic instead of silently treating non-zero exits as an empty session list
- allow stale-session cleanup to propagate listing failures so the manager retry guard can run on the next session-created event
- keep generic tmux sweeps fail-safe and add core plus adapter regression coverage

## Validation
- bun test --config=/dev/null packages/tmux-core/src/tmux-utils/stale-session-sweep.test.ts
- bun test --config=/dev/null packages/omo-opencode/src/shared/tmux/tmux-utils/stale-session-sweep-runtime.test.ts
- git diff --check

Typecheck was not run because tsgo is unavailable in this checkout.

---
## Summary by cubic
Propagates tmux list-sessions failures so stale-session sweeps can be retried. Previously, non-zero exits returned an empty list; now we surface a descriptive error. The Omo agent stale sweep opts into rethrowing to trigger the manager retry guard.

- In `tmux-core`, list-sessions now throws with stderr/stdout detail.
- `sweepTmuxSessionsWith` logs and returns `[]` unless `throwOnListError` is true, then rethrows.
- In `omo-opencode`, the agent stale sweep bubbles the error for retry.
- Tests cover both the fail-safe generic path and the retrying agent path.

No migration: the default generic sweep behavior is unchanged. Pass `throwOnListError: true` to enable retry in callers that need it.

<sup>Written for commit 0175f08ef3ae2a0c1f741321f0eca5f0fe03412a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
Thanks for the focused tmux fix. The approved required matrix ran, and every gate passed except `senpi-compatibility (ubuntu)`.

That job reports `packages/omo-senpi/plugin/extensions/omo-task.js` as stale output. This branch does not touch the Senpi bundle graph, so the failure is merge-snapshot drift rather than a tmux source defect.

Please merge current `dev` into the branch to refresh the merge ref and rerun the required matrix. No change to the stale-session-sweep implementation is requested.

suyua9:
Merged current `dev` into this branch (`0175f08ef`), which refreshes the `omo-senpi` merge snapshot that made `senpi-compatibility (ubuntu)` report stale output. No source changes to the tmux fix itself.

PR #7128, opened 2026-08-22
fix(session-manager): let agents find and read archived sessions

### LINKED ISSUE #7125: [Bug]: session_search cannot find archived sessions — session.list() scan set silently drops time_archived sessions
### Bug Description

`session_search` cannot find archived sessions. Once a session is archived in the opencode UI (tab session list → archive), it becomes invisible to the LLM's session tools — even though the session and all its messages remain fully intact in the database. There is no tool argument to opt into searching archived sessions.

This makes "archive" a destructive operation from the coding agent's perspective: the web UI's own search (command palette) still shows archived sessions dimmed, but the agent-driven `session_search` / `session_list` silently lose them.

### Steps to Reproduce

1. oh-my-openagent plugin loaded in opencode (SQLite backend), with a session containing a distinctive term (e.g. "Polycom") in many messages
2. Archive that session from the opencode web UI sidebar (`mod+shift+backspace` or the archive button)
3. In a new chat session, call the plugin tool:
   ```
   session_search({ query: "Polycom" })
   ```
4. Observe zero matches, while the archived session contains 52 messages with that term (verified directly in `opencode.db`)
5. Additional probe on ≤ 4.9.2: `session_read({ session_id: <archived-id> })` → `Session not found: ses_…`, even though `session_info()` on the same ID returns full metadata

Reproduction corpus: 348 archived / 1654 total sessions locally. Two archived sessions with heavy term coverage ("Polycom" ×52 messages, "pjsip" ×220 messages) both return zero `session_search` hits, while non-archived sibling sessions with the same terms match fine.

### Expected Behavior

- `session_search` finds content in archived sessions (at minimum via an opt-in argument such as `include_archived: true`), or
- `getAllSessions()` includes archived session IDs in the scan set so cross-session search covers the whole database.

The SDK already supports this: `client.session.list({ archived: true })` maps to the server query param that bypasses the `isNull(time_archived)` filter.

### Actual Behavior

- `session_search({ query })` → 0 results for any archived session
- `session_list` → archived sessions absent
- On 4.9.2, `session_read` on an archived ID → `Session not found` (fixed separately in 4.19.4 — see below)

### Root Cause

`getSdkAllSessions()` (session-manager storage, bundled `dist/index.js`) calls the SDK without the `archived` param:

```ts
async function getSdkAllSessions(client) {
  const response = await fetchSdkResponse(() => client.session.list());  // no { archived: true }
  const sessions = normalizeSDKResponse(response, []);
  return sessions.map((session) => session.id);
}
```

The opencode server's session list endpoint excludes archived sessions by default (`if (!input?.archived) conditions.push(isNull(SessionTable.time_archived))` in `listGlobal()`), so archived IDs never enter the search scan set or the list results.

Historically (≤ 4.9.2) `sdkSessionExists()` also derived existence from `session.list()` membership, which additionally broke `session_read` for archived IDs. 4.19.4 rewrote `sdkSessionExists` to probe messages directly, which fixes the read path — but the search/list gap remains in 4.19.4 (verified by inspecting the 4.19.4 tarball: `getSdkAllSessions` is unchanged, and `include_archived` appears nowhere in the tool schemas).

Related but distinct from #5810 (same anti-pattern — using scope-filtered `session.list()` as the source of truth — but that issue covers instance/project scope, this one covers the archived filter).

### Suggested Fix

Any of:

1. In `session_search`, when building the scan set, merge IDs from both `session.list()` and `session.list({ archived: true })`; or
2. Add an `include_archived` (default `false`) argument to `session_search` / `session_list` and pass `archived: true` through to the SDK call; or
3. Default the scan set to the full DB (both calls merged) since search is a read-only discovery operation and silent exclusions are surprising.

### Environment

- oh-my-openagent: 4.9.2 (empirical repro) — source-inspected 4.19.4 (latest stable), bug persists
- opencode: 1.18.15 (server behavior source-verified against current dev @ e00890c67 / v1.18.21-era)
- OS: macOS 26 (darwin)
- Backend: SQLite (`~/.local/share/opencode/opencode.db`)
- Interface: opencode web UI + plugin session tools

richard-fairthorne:
Fixes #7125

## What changed

Agents lost access to knowledge the moment a user archived a session: `session_search` silently skipped archived sessions (no opt-in existed) and `session_list` could not show them. This PR makes archived history discoverable again, following the approach recommended in the issue thread:

- **`session_search` includes archived sessions by default** (opt-out via `include_archived: false`). Search is a discovery operation; the caller is rarely the same agent that archived the session, so a default-off flag would not fix the failure mode.
- **Active-first scan ordering.** SDK sessions are partitioned by `time.archived`; archived ids are appended after active ids. Within the existing `MAX_SESSIONS_TO_SCAN` budget, archived sessions can never displace active ones, and every result findable before this change remains findable (strict superset).
- **`session_list` keeps its default** (archived hidden, sidebar semantics) and gains `include_archived: true` to include them.
- **`[archived]` provenance marker** on hits and rows, so the calling agent knows restore semantics may apply.
- No changes to `session_read` / `session_info` (read path already fixed via the messages probe).

Implementation notes: `getSdkAllSessions()` now returns `{ active, archived }` partitioned client-side; the stable SDK `Session` type does not declare `time.archived` yet (wire field), so the read is defensive. On servers that do filter archived sessions from `GET /session`, archived ids simply stay absent and behavior degrades gracefully to the pre-fix state.

## Observed behavior (live QA)

Driven on real opencode 1.18.15 in an isolated XDG sandbox (HOME + XDG_* redirected, sandbox DB seeded with one active + one archived session containing a unique needle; archived session deliberately given the newer `time_updated` to stress ordering):

| Call | Result |
|---|---|
| `session_search({query: needle})` | Both sessions found; archived hit rendered `[ses_qa_archived] [archived] ...`; active hit listed **before** archived despite older `time_updated` |
| `session_list({include_archived: true})` | Both rows, archived row marked `[archived]`, active first |
| `session_list({})` | Only the active session (unchanged default) |

Also verified against the server: `GET /session` on 1.18.15 returns archived sessions with `time.archived` set, so the client-side partition works on the current release line. Full evidence (transcripts, isolation proof) in `.omo/evidence/2026-08-22-session-search-archived/EVIDENCE.md`.

## Verification

- `bun test packages/omo-opencode/src/tools/session-manager/` — 74/74 pass (incl. new tests: search default-on, explicit opt-out, list default-off/opt-in, active-first append order, legacy sessions without `time.archived` treated as active, marker propagation, result-limit behavior)
- `tsgo --noEmit -p packages/omo-opencode/tsconfig.json` — clean
- Root-suite and full `typecheck:packages` cannot fully run in a fresh worktree (senpi/codex vendored artifacts are install-generated, not git-tracked); the failing tests there fail identically on clean `dev` in this environment and pass in a fully installed checkout. Not touched by this diff.
- Live-surface QA per repo protocol (see Evidence above).

## Residual risk

- `MAX_SESSIONS_TO_SCAN` still bounds the scan (by design, unchanged); archived sessions consume the remaining budget after actives.
- Behavior on servers that filter `/session` degrades to pre-fix (never worse).

---
## Summary by cubic
Restore archived session discoverability for agents. Previously, `session_search` silently skipped archived sessions and `session_list` could not show them; now search includes archived by default (opt-out), and lists can include them with a flag, with clear [archived] markers.

- Reviewer notes:
  - Active-first scan ordering: SDK sessions are partitioned by `time.archived`; archived IDs append after active IDs within `MAX_SESSIONS_TO_SCAN`, so archived never displace active.
  - `session_search`: new arg `include_archived` (default true). Matches from archived sessions are marked [archived].
  - `session_list`: default unchanged (archived hidden). New arg `include_archived: true` to include them; rows from archived sessions are marked [archived].
  - Storage API adjustments: SDK `getSdkAllSessions()` returns `{ active, archived }`; `getAllSessions()` returns `{ ids, archivedIds }`. Callers updated; file-backed fallback unchanged.
  - Compatibility: if servers omit archived sessions from `GET /session`, behavior degrades to pre-fix. Extensive tests cover defaults, opt-in/out, ordering, and marker propagation.

<sup>Written for commit 3495e087e5615ad0d8c4dccdd98b55318d6554f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


richard-fairthorne:

PR #7129, opened 2026-08-22
feat(omo-native): re-exec the launcher under bun for bun-installed users

code-yeongyu:
## Summary

A user who runs `bun add -g omo-ai` gets `~/.bun/bin/omo`, a symlink into the Bun global tree. The shebang in `bin/omo.js` still says `node`, so the product silently ran on the runtime the user did not pick. This makes the launcher notice that install and hand the process back to bun, and it tells the spawned senpi engine which runtime won so the decision is made exactly once per launch.

Users who installed with npm are unaffected, and a machine with no bun binary stays on node without an error.

## Changes

**`packages/omo-native/bin/lib/bun-runtime.js`** (new) - pure, injectable helpers:

- `isUnderBunGlobalTree(scriptRealPath, opts)` - the executed script's real path sits under `<BUN_ROOT>/install/global/`, where `BUN_ROOT` is `BUN_INSTALL` or `~/.bun`. The real path matters because `~/.bun/bin/omo` is a symlink and the link itself never matches.
- `findBunBinary(opts)` - `BUN_INSTALL/bin/bun` -> `~/.bun/bin/bun` -> PATH scan, honoring `PATHEXT` and the `.exe` suffix on win32. Missing bun is a normal answer, never an error.
- `resolveBunReexec(input)` - the decision table, first match wins: already on bun -> stay (loop guard); `OMO_RUNTIME=node` -> stay; `OMO_RUNTIME=bun` + bun found -> re-exec; bun global tree + bun found -> re-exec; otherwise stay.
- `maybeReexecUnderBun(...)` - executes the decision via `spawnSync` and the existing `propagateResult`, returning `true` when a re-exec consumed the process. Node's `execArgv` is deliberately dropped, since node flags are not bun flags.

**`packages/omo-native/bin/omo.js`** - first thing in the `try` block; when it re-execs, nothing else runs and the exit status is already propagated. Covers both the setup and launcher paths.

**`packages/omo-native/bin/lib/launcher.js`** - `senpiEnvironment()` sets `SENPI_RUNTIME` to the runtime actually executing the launcher, so the engine inherits the decision rather than re-deciding. The engine side honors it as a hard answer (landed separately in the senpi repo).

## QA & Evidence

Evidence: `.omo/evidence/20260822-launcher-bun-auto-runtime/` (gitignored), see `README.md` there.

Real-surface run (`06-qa-real-surface.txt`) builds a throwaway `BUN_INSTALL` tree with a logging bun shim plus a copy of the real `bin/`, then drives the real `bin/omo.js` through node:

| Case | Expected | Observed |
|---|---|---|
| script inside the bun tree | re-exec | shim `INVOKED`, `argv: <omo.js> --version` |
| same + `OMO_RUNTIME=node` | stay | no shim log, node answered `--version` |
| repo checkout, no bun tree | stay | no shim log, normal launcher output |

**The real-surface QA caught two production bugs the unit tests missed**, both in the tree-path comparison, both now fixed and pinned by regression tests:

1. Interior duplicate separators. macOS `TMPDIR` ends in `/`, so `BUN_INSTALL` arrived as `/var/.../T//omo-...` and the prefix comparison missed the install. `BUN_INSTALL` is user-supplied, so this is a real-world shape, not a temp-dir quirk.
2. An unresolved symlink on the root side. Node realpaths the main module, so the script arrived as `/private/var/...` while the root stayed `/var/...`; the two named the same directory and never matched. The root is now canonicalized through an injectable `realpath`, falling back to the literal value when it does not resolve. A symlinked `$HOME` reproduces the same failure against `~/.bun`.

Tests:

- RED captured before implementation (`01-red-bun-runtime.txt`), GREEN after (`02-green-bun-runtime.txt`).
- `bun test packages/omo-native`: **118 pass, 13 fail**. The same suite with this change stashed away is **93 pass, 13 fail** (`04-baseline-preexisting-failures.txt`), and the 13 failing test names are identical between the two runs. They are pre-existing `omo doctor` / `omo setup` failures on `dev`, untouched here. This change adds 25 passing tests and zero new failures.
- `tsgo --noEmit -p packages/omo-native/tsconfig.json` clean (`05-typecheck.txt`).

Unit tests are fully hermetic: every helper takes injected `env` / `homedir` / `platform` / `exists` / `realpath`, so nothing reads the host PATH or the host `~/.bun`. That is deliberate for the Windows shards on Bun 1.4, where a real bun is present and a host-reading test would pass or fail by accident. The real `~/.bun` was never written and never used as a binary source; the `~/.bun/bin` digest is recorded before and after in the transcript and is unchanged.

## Risks

- The re-exec is not driven against a real bun binary end to end - the QA shim stands in for one and exits 0. The forwarded argv shape is pinned by unit test and matches what `spawnNode` already sends in production.
- Behavior change is scoped to bun global installs: an npm install, a repo checkout, and a machine without bun all take the same path they did before.
- Re-exec costs one extra process spawn on affected installs, paid once per launch.

---
## Summary by cubic
Re-execs the `omo` launcher under Bun for users who installed with `bun add -g`, so the runtime matches the install choice. Previously these users ran under Node due to the Node shebang; now the launcher detects Bun global installs and hands execution to Bun, and the engine inherits the decision.

- Detects Bun global installs by checking the executed script’s real path under `<BUN_ROOT>/install/global/`, normalizing separators and resolving a symlinked root so comparisons don’t miss valid installs (handles Windows backslashes, duplicate/trailing separators).
- Re-exec policy: already on Bun or `OMO_RUNTIME=node` => stay; `OMO_RUNTIME=bun` with Bun found => re-exec; Bun global install with Bun found => re-exec; missing Bun => stay. Drops Node `execArgv` and propagates the child’s exit; adds one extra spawn only when re-execing.
- Sets `SENPI_RUNTIME` to the actual runtime so the engine does not re-decide and flip-flop.
- No change for npm installs or hosts without Bun; loop-safe.
- Tests cover detection and handoff on Node and Bun using hermetic inputs; updates `script/agent-command-string-audit.allowlist.json` to keep the “bare omo” command-string audit green for new fixtures/comments.

<sup>Written for commit 27ba9b56ac875412e07b4d1636f08ab118f67a5d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7129?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-native/bin/lib/bun-runtime.js
+++ b/packages/omo-native/bin/lib/bun-runtime.js
@@ -1,0 +1,127 @@
+import { spawnSync } from "node:child_process"
+import { existsSync, realpathSync } from "node:fs"
+import { homedir as osHomedir } from "node:os"
+import { delimiter, join } from "node:path"
+import { propagateResult } from "./child-process.js"
+
+// A bun global install lives under <BUN_ROOT>/install/global/, and `bun add -g` links its bin
+// through a symlink in <BUN_ROOT>/bin. The link target is what identifies the install, so every
+// comparison below runs on a real path.
+const GLOBAL_TREE_MARKER = "/install/global/"
+
+// Both sides of the tree comparison are reduced to one spelling: backslashes become forward
+// slashes so Windows paths match, and repeated separators collapse because BUN_INSTALL is
+// user-supplied and a value like `/tmp//bunroot` would otherwise never prefix-match a real path.
+function normalize(path) {
+  return path.replaceAll("\\", "/").replaceAll(/\/{2,}/g, "/")
+}
+
+function bunRoot(env, homedir) {
+  return env.BUN_INSTALL ? env.BUN_INSTALL : join(homedir(), ".bun")
+}
+
+/**
+ * Node resolves the main module to its real path, so the script side of the comparison is already
+ * canonical. The root has to be canonicalized too or the two sides can name the same directory in
+ * different spellings - `/tmp` against `/private/tmp` on macOS, or any symlinked home - and a real
+ * bun install would silently fail to be recognized. A root that does not exist is used verbatim.
+ */
+function canonicalRoot(env, homedir, realpath) {
+  const root = bunRoot(env, homedir)
+  try {
+    return realpath(root)
+  } catch {
+    return root
+  }
+}
+
+function binaryName(platform) {
+  return platform === "win32" ? "bun.exe" : "bun"
+}
+
+/**
+ * True when the executed script belongs to a Bun global install. The caller passes the script's
+ * REAL path: `~/.bun/bin/omo` is a symlink into the tree, so the link itself never matches.
+ */
+export function isUnderBunGlobalTree(scriptRealPath, options = {}) {
+  const env = options.env ?? process.env
+  const homedir = options.homedir ?? osHomedir
+  const realpath = options.realpath ?? realpathSync
+  const root = normalize(canonicalRoot(env, homedir, realpath)).replace(/\/+$/, "")
+  return normalize(scriptRealPath).startsWith(`${root}${GLOBAL_TREE_MARKER}`)
+}
+
+function pathExtensions(env, platform) {
+  if (platform !== "win32") return [""]
+  const configured = env.PATHEXT ?? ".COM;.EXE;.BAT;.CMD"
+  return configured.split(";").filter(Boolean)
+}
+
+/**
+ * Locates the bun binary this machine would run. Absence is a normal answer - a user who installed
+ * omo with npm has no bun, and the launcher simply stays on node.
+ */
+export function findBunBinary(options = {}) {
+  const env = options.env ?? process.env
+  const homedir = options.homedir ?? osHomedir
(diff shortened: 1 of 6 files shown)


code-yeongyu:
CI note: `senpi-compatibility (ubuntu-latest)` failed twice on `task RPC launch profile parity` before going green on a bare re-run of the same commit, with no code change in between.

That test spawns a real senpi child and probes a live model catalog. The failure message reported a different catalog size each time (`probed catalog has 576 models`, then `277 models`), so the input is externally sourced and varying. This diff is confined to `packages/omo-native/` and cannot reach senpi RPC model admission; the test also passes locally. Recording it here as an environmental flake rather than silently re-running past it.

All checks are green as of `27ba9b56a`.

PR #7146, opened 2026-08-23
fix(telemetry): unmask model_id for unknown providers

code-yeongyu:
## Problem

`maskProviderAndModel` masks provider AND model_id to `custom` when the provider is not in `KNOWN_MODELS`. This means users routing known models (claude-opus-5, gpt-5.6-sol, etc.) through OpenRouter, LiteLLM, or self-hosted gateways show as `custom/custom` in the telemetry dashboard — making it impossible to see actual model preferences.

From the current PostHog data, ~30% of category_config sessions include `custom/custom` slots despite running standard, publicly-named models.

## Fix

- Added `ALL_KNOWN_MODEL_IDS` — a flat `Set` union of every model id across all known providers in `model-vocabulary.ts`
- Changed `maskProviderAndModel` to check `model_id` against this global set instead of only the current provider's model list
- Model ids are public product names, not PII — no privacy regression

## Before / After

| Provider | Model | Before | After |
|---|---|---|---|
| `openrouter` | `claude-opus-5` | `custom/custom` | `custom/claude-opus-5` |
| `my-gateway` | `gpt-5.6-sol` | `custom/custom` | `custom/gpt-5.6-sol` |
| `anthropic` | `my-finetune` | `anthropic/custom` | `anthropic/custom` (unchanged) |
| `my-gateway` | `my-finetune` | `custom/custom` | `custom/custom` (unchanged) |

## Tests

All 185 telemetry tests pass. Updated assertions in `product-identity.test.ts` and `delegation-projection.test.ts` to reflect the new behavior.

---
## Summary by cubic
Unmasks model_id for known public models when the provider is unknown. Previously both provider and model_id masked to custom; now provider stays custom but model_id passes through if it matches a known model, improving telemetry accuracy without exposing PII.

- Introduces `ALL_KNOWN_MODEL_IDS`, a global set of all known model ids across providers.
- Updates `maskProviderAndModel` to unmask model_id when it appears in this set, regardless of provider.
- Unknown providers still mask to "custom"; unknown or fine-tuned model ids still mask to "custom".
- Telemetry impact: `category_config` and `delegation_completed` now report actual model ids when routed via OpenRouter/LiteLLM/self-hosted gateways.
- Tests updated; no configuration or migration required.

<sup>Written for commit f95a32fe72ed240ebe3b0bdb55cb99a5c20447a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-senpi/src/components/telemetry/delegation-projection.test.ts
+++ b/packages/omo-senpi/src/components/telemetry/delegation-projection.test.ts
@@ -153,9 +153,10 @@ describe("delegation_completed projection", () => {
       },
     })
 
-    // then
+    // then: provider is custom (unknown), but model_id is unmasked because "claude-opus-5" is a
+    // known public model name — model ids are not PII and should be visible regardless of routing provider
     expect(props.provider).toBe("custom")
-    expect(props.model_id).toBe("custom")
+    expect(props.model_id).toBe("claude-opus-5")
     expect(props.model_source).toBe("explicit")
   })
 
--- a/packages/omo-senpi/src/components/telemetry/model-vocabulary.ts
+++ b/packages/omo-senpi/src/components/telemetry/model-vocabulary.ts
@@ -43,3 +43,10 @@ export const KNOWN_MODELS = Object.freeze({
 
 export type KnownProvider = keyof typeof KNOWN_MODELS
 export const KNOWN_PROVIDERS = Object.freeze(Object.keys(KNOWN_MODELS) as KnownProvider[])
+
+// Flat set of every model id known across ALL providers. Used to unmask model_id even when the
+// provider itself is not in the vocabulary (e.g. OpenRouter, LiteLLM, or a self-hosted gateway
+// proxying a standard model). Model ids are public product names, not PII.
+export const ALL_KNOWN_MODEL_IDS: ReadonlySet<string> = Object.freeze(
+  new Set(Object.values(KNOWN_MODELS).flat()),
+)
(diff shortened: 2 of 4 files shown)


code-yeongyu:
Closing in favor of a corrected replacement PR.

This change unmasks telemetry `model_id` for publicly-known models routed through unknown providers — but it only changed the behavior. It left every surface that *documents* the old privacy invariant untouched:

1. `model-vocabulary.ts` header comment still declares "Masking stays PROVIDER-SPECIFIC: a model listed under one provider is still \`custom\` under a provider that does not ship it" — the code would contradict its own stated invariant.
2. `docs/reference/senpi-telemetry.md` still lists under **never captured**: "Custom provider or model names, which are masked to \`custom\`" — which this change makes factually false for model names.
3. The telemetry component `AGENTS.md` "Privacy and opt-out" section repeats the same now-false claim.

Shipping a telemetry-scope change while the published privacy disclosure still describes the old, narrower behavior is not acceptable. The replacement PR carries the same behavior fix plus the corrected disclosure on all three surfaces, with the privacy guard (unknown/fine-tuned model names still mask to `custom`) pinned by an explicit regression test.

PR #7166, opened 2026-08-24
fix(background-agent): force reply production when parent idle on subagent completion wake

mooire733:
## What Changed

Fixes the background subagent completion wake path so that when the parent session is **idle**, the completion notification is forced through the reply-production path (shouldReply = true) instead of falling into the noReply injection path.

Touched files:
- `packages/omo-opencode/src/features/background-agent/parent-wake-prompt-dispatch.ts`
- `packages/omo-opencode/src/features/background-agent/parent-wake-flush-runner.ts`

## Why

In parallel multi-subagent scenarios, the first completing subagent was judged `shouldReply = all-complete || failed` as **false**, so the wake notification went through the noReply injection path. When the parent session was idle, the notification was written to history but never triggered a new model turn — the wake was effectively dropped.

The watchdog fallback window was ~5s with a single retry, which is insufficient for deepseek/k3 cold starts. Production tracking on 2026-08-22: 149 wakeups, only 6 confirmed delivered, 51 permanently dropped.

This regression was introduced by the 5.0.0-beta.9 parent-wake duplicate-injection gate rework.

## Observed Behavior Before vs After

- **Before:** idle parent + first completing subagent → notification written to history, no new model turn triggered (wake dropped).
- **After:** idle parent → reply production is forced, so the parent session wakes and processes the completion.

## QA & Evidence

- Local verification followed TDD: the failing test (`b12dc18`) was written first and reproduced the bug, then the core fix (`65f17b4`) made it pass, followed by a follow-up commit (`3c446b7`).
- The change is scoped to the two background-agent wake files above.

## Residual Risk

- The fix forces reply production specifically when the parent session is idle; the duplicate-injection gate from beta.9 is preserved for the non-idle path, so no regression is expected for the already-handled wake cases.
- Full CI is the authoritative gate for cross-platform behavior.


mooire733:
Rebased onto latest upstream `dev` (`0f06ba45e`; was 13478 commits behind). Branch force-updated on fork with `--force-with-lease`. Root `bun run build` passes and the fix marker (`Forced reply parent wake because parent session is idle`) is confirmed present in the rebuilt `dist/index.js`. Requesting fresh CI on the rebased head.

mooire733:
Hi maintainers — could you approve the workflow run for the rebased head (`af07de55`)? The ci.yml job is pending 'workflow approval required' since this comes from a fork. The branch was just rebased onto the latest `dev` and the full build passes locally. Thanks!

mooire733:
Rebased onto latest `origin/dev` (branch head now `bd495aebf`, same 3 fix commits, new SHAs after rebase). Local build verified (`bun run build` exit 0); fix logic confirmed present in the built `dist/index.js` bundle: `const reply = input.forceNoReply !== true && (input.forceReply === true || input.latestWake.shouldReply)`.

mooire733:
Hi maintainers — this PR is from an outside contributor fork, so the `CI` workflow is waiting for approval before it can run (`action_required`, no jobs started). Could you approve the workflow run so the checks can execute? Thank you!

(Note: the full QA evidence bundle for this change is committed under `.omo/evidence/20260824-pr1-idle-force-reply/` in the PR.)

mooire733:
Maintainer action needed: the CI workflow on this fork PR is waiting for approval before any job starts (`action_required`, zero jobs launched). Could a maintainer please approve the workflow run so the required checks (`bun test`, typecheck, build, codex-compatibility) can execute? The branch is rebased onto the latest `origin/dev` and the local build is verified. Thanks!

mooire733:
Cleanup + regression pin on the forced-reply fast path:

- The core fix commit now contains ONLY the two real source changes (parent-wake-flush-runner.ts, parent-wake-prompt-dispatch.ts). All QA-evidence pollution (~299KB `sandbox-opencode.db`, probe logs, build logs, sentinel-proof files) was removed and is preserved uncommitted under `.omo/evidence/20260824-pr1-idle-force-reply/` in the worktree.
- Added the #4120 user-message-race guard to the `forceIdleReply` fast path: the fast path now also requires `!isUserMessageInProgress`; when a user message is in progress it falls through to the admit-only noReply deposit instead of racing the user's turn (activity window 5s vs user-message window 2s can overlap).
- Added a regression test pinning that intersection (`#given recent parent activity is fresh and a user message is in progress` → admit-only noReply + dispatched-tracked deposit).
- Rebased head SHA: now `3be6267c4` (tests: background-agent 751 pass, prompt-async-route-audit 10 pass, build green).

MoerAI:
Thanks for addressing the dropped idle-parent wake. The blocking defect is that the new forced-reply fast path bypasses the running-tool-call history guard, so fresh activity plus `finish: "tool-calls"` and running tool state dispatches a reply-producing prompt into a live tool call. Please run `shouldDeferForHistory` before forcing the reply and fall through to the admit-only deposit when it defers. Add a regression test covering the exact intersection of fresh parent activity and blocked tool history, and provide passing CI plus committed QA evidence for the corrected scoped change. Diff: https://github.com/code-yeongyu/oh-my-openagent/pull/7166/files

mooire733:
The forced-reply fast path now runs shouldDeferForHistory first (fc59afcc8). When history still blocks internal prompts (finish:"tool-calls", running tool state), it lands on the admit-only noReply deposit instead of dispatching into a live tool call. The crossover regression test is in 53fb6ec4f, QA evidence in 3c71256b0 under .omo/evidence/20260903-pr7166-history-guard/. Suite is green: 9/9 in the idle-force-reply file, 752/752 across background-agent. CI hasn't actually run yet, the workflow sits at action_required with zero jobs, so it needs a maintainer to approve and run.

Issue #1839, opened 2026-02-14
[Bug]: Models cheat Prometheus write restrictions by using cat


Adithya2685-git:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

Models, when the chat extends for too long,  tend to cheat Prometheus's Write restrictions by using cat. cat must be disabled in order to prevent write these issues

### Steps to Reproduce

Have a long enough chat and use open free models like MiniMax M2.5. Then keep switching between Sisyphus and prometheus. Eventually, it will loop itself in Prometheus write restrictions to get around it by using cat to write in Prometheus.

### Expected Behavior

Agents should not be able to write anything in Prometheus mode. 

### Actual Behavior

Prometheus games the system by using cat directly to the file.

### Doctor Output

```shell
oMoMoMoMo... Doctor 


Installation
────────────────────────────────────────
  ✓ OpenCode Installation → 1.2.1
  ✓ Plugin Registration → Registered (pinned: latest)

Configuration
────────────────────────────────────────
  ✓ Configuration Validity → Valid JSON config
  ✓ Model Resolution → 10 agents, 8 categories (18 overrides), 2570 available

Authentication
────────────────────────────────────────
  ✓ Anthropic (Claude) Auth → Auth plugin available
  ○ OpenAI (ChatGPT) Auth → Auth plugin not installed
  ✓ Google (Gemini) Auth → Auth plugin available

Dependencies
────────────────────────────────────────
  ✓ AST-Grep CLI → installed
  ✓ AST-Grep NAPI → installed
  ⚠ Comment Checker → Not installed (optional)

Tools & Servers
────────────────────────────────────────
  ⚠ GitHub CLI → Not installed (optional)
  ✓ LSP Servers → 2/4 servers available
  ✓ Built-in MCP Servers → 2 built-in servers enabled
  ○ User MCP Configuration → No user MCP configuration found
  ○ MCP OAuth Tokens → No OAuth tokens configured

Updates
────────────────────────────────────────
[...]

PR #7178, opened 2026-08-24
fix(agents): restore bash for Prometheus scoped to the ulw-plan scaffold script

### LINKED ISSUE #7160: Prometheus's ulw-plan skill mandates 'node scripts/scaffold-plan.mjs', but bash was removed from Prometheus's tool registry entirely (regression from #1839/#1428 fixes)
## Summary

`ulw-plan`'s `SKILL.md` has a mandatory step:

> ## RUN THE SCRIPT - do not hand-build artifacts
> As soon as `<slug>` and intent are known... RUN:
> ```
> node "<skill-root>/scripts/scaffold-plan.mjs" <slug> [--clear|--unclear] --draft-only [--review-required]
> ```

There is no way to execute a `.mjs` file without a shell tool. But Prometheus's tool registry does not include `bash` at all. Confirmed directly from a live session transcript:

```
Tool: invalid
Input: {"tool": "bash"}
Output: "The arguments provided to the tool are invalid: Model tried to call unavailable tool 'bash'.
Available tools: background_cancel, background_output, ..., edit, glob, grep, ..., lsp_*, question,
read, session_*, skill, skill_mcp, task, todowrite, webfetch, websearch*, write."
```

`bash` is absent from the list. The model never gets a permission-denied error (which would imply a `permission.bash: deny` check) — it gets an "unavailable tool" error, meaning `bash` isn't registered for this agent at all.

## Likely cause

This looks like the fix for #1839 ("Models cheat Prometheus write restrictions by using cat") and/or #1428 ("Prometheus mode bypasses permission config for bash commands") — both closed by, as far as I can tell, removing `bash` from Prometheus's tool set entirely, since `prometheus-md-only`'s hook only intercepts `write`/`edit` by path and never inspected `bash`, so blanket removal was the only way to close the cat-based bypass.

That fix is reasonable for the security concern, but it silently broke `ulw-plan`'s own documented workflow. The skill still instructs Prometheus to run `scaffold-plan.mjs` via `node`/`bun`, with no fallback path if `bash` is unavailable, and (as far as I can find) no existing issue tracks this specific conflict.

## Suggested fix

Either:
1. Update `ulw-plan`'s `SKILL.md` to not require shell execution (e.g. expose scaffold-plan as a native/dedicated tool instead of a script needing `bash`), or
2. Restore `bash` for Prometheus but scope it via `permission.bash` as a command-pattern map (e.g. `{"node *scaffold-plan.mjs*": "allow", "bun *scaffold-plan.mjs*": "allow", "*": "deny"}`) so the scaffold script works without reopening the `cat`-based write-restriction bypass from #1839.

I've applied option 2 as a per-user config workaround (`agents.prometheus.tools.bash: true` + the pattern-scoped `permission.bash` map above), but this should really be the shipped default for Prometheus, or the skill should stop assuming shell access.

## Environment

- oh-my-openagent v4.19.4 (installed via opencode plugin `oh-my-openagent@latest`)
- Skill: `ulw-plan` (packages/shared-skills/skills/ulw-plan)

Keonho-Chu:
## Summary

Prometheus lost the `bash` tool entirely when the #1839/#1428 cat-based write bypass was closed with a blanket permission deny, which silently broke the ulw-plan skill's own mandatory step (`node "<skill-root>/scripts/scaffold-plan.mjs" ...`) — the model gets "unavailable tool 'bash'" and can never run the scaffold. This PR replaces the blanket deny with a pattern-scoped `permission.bash` map, so Prometheus can run exactly the documented scaffold invocations while every other bash command stays denied. This is option 2 from the issue, shipped as the default.

## Changes

- `packages/omo-opencode/src/plugin-handlers/tool-config-handler.ts`: Prometheus's `bash: "deny"` becomes `PROMETHEUS_BASH_PERMISSION` = `{ "*": "deny", "node *scaffold-plan.mjs*": "allow", "bun *scaffold-plan.mjs*": "allow" }`. `interactive_bash` stays denied, and the map keeps its position after the user-permission spread, so precedence semantics are unchanged from the old hard deny.
- **Rule order is load-bearing and documented in a comment**: opencode resolves permission rules with `findLast` and hides a tool entirely when the LAST matching rule is a wildcard deny (`Permission.disabled()` in opencode core). With the catch-all last, `bash` never reaches the model's tool list at all — verified empirically during QA. The catch-all must come first, scoped allows after. `findLast` evaluation still yields: scaffold commands → allow, everything else → deny.
- Tests (`tool-config-handler.test.ts`, `tool-config-handler-display-name.test.ts`): assert the emitted map shape, the key order (machine-consumed by opencode's `findLast`), the override of pre-existing agent bash permission, and `interactive_bash` deny.

## QA & Evidence

Evidence dir (worktree, gitignored per repo convention): `.omo/evidence/20260824-issue-7160/` — `qa-summary.md`, `registered-tools-prometheus.json`, `agent-endpoint-bash-ruleset.json`, `run-*.jsonl`.

All QA drove the **real opencode binary (v1.18.13)** with the worktree-built plugin (`file://.../dist/index.js`) in an isolated XDG sandbox, as the Prometheus agent, using a local fake OpenAI-compatible server that captures each request body verbatim and scripts exact `bash` tool calls (deterministic, no prompt-compliance flakiness).

- **What was tested:** bash tool registration — the `tools` array of the captured LLM request for a Prometheus session.
  **Observed:** 98 tools including `bash`. With the old code (and with the catch-all ordered last), the same capture had NO `bash` — reproducing the issue's "unavailable tool" symptom and proving the ordering discovery.
- **What was tested:** catch-all deny on real executions: `echo qa-bypass-test > qa-7160-deny.md`, `cat scaffold-plan.mjs > stolen.md` (the #1839 bypass shape), `whoami`, `git status`.
  **Observed:** every call returned `status: "error"` with opencode's rule-rejection message citing the emitted ruleset; none of the target files were created.
- **What was tested:** the documented scaffold invocation `node scaffold-plan.mjs <slug> --clear --draft-only` as a real tool execution.
  **Observed:** `status: "completed"`, exit 0, stdout `created: .omo/drafts/qa-7160e.md`, draft present on disk.
- **What was tested:** command-chaining smuggling (`node scaffold-plan.mjs ...; echo ...`).
  **Observed:** denied — opencode evaluates each segment of a compound command, so the scoped allow cannot be used to smuggle arbitrary commands.
- **Isolation:** host `~/.local/share/opencode/opencode.db` session count 132 before and after all QA; every spawn used throwaway XDG dirs.
- **Unit gates:** `bun test packages/omo-opencode/src/plugin-handlers/` — 238 pass, 0 fail. `bun run typecheck` — green.

One caveat surfaced during QA, unrelated to this change: invoking the scaffold via a non-canonical macOS `/var/folders/...` symlink path exits 0 silently without creating the draft — reproduced identically OUTSIDE opencode (the scaffold script's ESM main-module detection fails on non-canonical `argv[1]`). The permission layer allowed and executed the command correctly; real installs use canonical `<skill-root>` paths.

## Risks & Residuals

- **Does this reopen #1839/#1428?** No. The write bypass shape (`cat x > y.md`) is exercised directly in QA and stays denied; the catch-all denies every non-scaffold command, and opencode's per-segment evaluation of compound commands blocks chaining through the allow patterns. The remaining theoretical surface is a file literally named `...scaffold-plan.mjs` crafted somewhere on disk and invoked via `node`/`bun` — an attacker-shaped scenario that requires the model to first create such a file, which the md-only write guard and the bash deny both block.
- **User override precedence** is unchanged: the scoped map sits exactly where the hard deny used to sit (after the `...prometheus.permission` spread), so configurations that previously could not override the deny still cannot, and the issue reporter's user-level workaround remains compatible.
- Full root `bun test` runs in CI as the authoritative gate for this PR.

## Related Issues

Fixes #7160

---
## Summary by cubic
Restores `bash` for the Prometheus agent but scopes it to only the ulw-plan scaffold script so the mandatory `node "<skill-root>/scripts/scaffold-plan.mjs" ...` step can run again. Previously `bash` was fully denied, which hid the tool and broke scaffolding; all non-scaffold commands remain denied.

- Implemented `PROMETHEUS_BASH_PERMISSION`: `{"*": "deny", "node *scaffold-plan.mjs*": "allow", "bun *scaffold-plan.mjs*": "allow"}`.
- Rule order is required due to opencode’s findLast: wildcard deny first, scoped allows after, which keeps `bash` visible but restricted.
- `interactive_bash` stays denied; user-override precedence unchanged; other agents unchanged.
- Tests assert the map shape, key order, override behavior, and `interactive_bash` deny.

<sup>Written for commit 3e56f857049db8454a400d542075f87cc6676d97. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7178?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Keonho-Chu:

PR #6677, opened 2026-08-09
fix(model-core): classify terminal_quota_exhausted as non-retryable to prevent duplicate subagent spawns

niStee:
Splits the qwen3.5→qwen3.7 chain update out of #6492 (per maintainer request, Aug 6 review).

This PR covers **only** the terminal_quota_exhausted classification. The qwen3.5→qwen3.7 model rename moves to a follow-up PR against `dev` (separate scope).

## What's in this PR

- `runtime-fallback-error-classifier.ts`: adds `getDetailErrorType()` + `isTerminalQuotaMessage()`; classifies `terminal_quota_exhausted` (LiteLLM structured signal) and terminal-style messages (`terminal quota`, `terminal billing limit`, `hard billing limit`) as `abort` so they don't retry-spawn duplicate subagents.
- `runtime-fallback-error-shape.ts`: small type refinement for the structured detail.
- `runtime-fallback-error-classifier.test.ts`: 5 new tests covering terminal vs. non-terminal regex discrimination.
- Regenerated `packages/omo-senpi/plugin/extensions/omo*.js` bundle (idempotent from source).

## Why split

From the maintainer review on #6492:
> *"The quota-exhaustion terminal treatment is the right call for preventing retry storms, but the PR conflates two changes: (1) the terminal_quota_exhausted fix and (2) the qwen3.5->qwen3.7 model rename. Please separate these into two PRs."*

Done. The qwen chain update is on a separate branch (`fix/senpi-qwen37-sync-isolated`, TBD), so this PR can land first without dragging the model rename through review.

## Verified

- Rebased onto current upstream `dev` (`13db09a1a`, the v5.0.0-beta.1 source state)
- All non-quota classifier logic unchanged
- 5/5 new tests pass (verified via the pre-commit hook chain)
- Bundle regen is byte-stable relative to source

---
## Summary by cubic
Stops retry storms by treating LiteLLM terminal quota exhaustion as non‑retryable. Previously these were retried as quota_exceeded and could spawn duplicate subagents; now they abort while soft quota/billing limits remain retryable.

- Detects structured `detail.error.type === "terminal_quota_exhausted"` from `error`, `data.detail`, or `detail` and classifies as `abort` (structured detail takes precedence over wrapper names like `ModelNotFoundError`).
- Treats “terminal quota”/“terminal billing limit”/“hard billing limit” messages as abort, but ignores “non‑terminal …” phrases to avoid false positives.
[...]

PR #7182, opened 2026-08-24
fix(omo-opencode): dispatch session-stable fallback on terminal-quota 402 aborts via session.error

### LINKED ISSUE #6959: [BUG] Sessions die with zero output tokens when minimax-coding-plan hits 402 quota
## Description

When using minimax-coding-plan as primary provider and OmniRoute models as fallback, sessions **die immediately** when the primary model hits a 402 quota error. The fallback chain is NOT engaged.

## Root Cause

PR #6764 correctly identifies this: LiteLLM's terminal-quota 402 is classified as non-retryable 'abort', so 
untime_fallback logs 'Error not retryable, skipping fallback' and the session stops.

## Impact

- **Production sessions die** with zero output tokens
- **No fallback occurs** even though fallback models are configured
- **User loses work** - session must be manually restarted

## Reproduction

1. Configure minimax-coding-plan as primary
2. Configure OmniRoute models as fallback
3. Exhaust minimax-coding-plan quota
4. Session dies instead of falling back

## Evidence

`
message.updated with assistant error {statusCode: 402, errorName: 'APIError', errorType: 'abort'}
Error not retryable, skipping fallback
`

## Fix

PR #6764 is ready and tested (253 pass / 0 fail). Please prioritize merge.

## Priority

HIGH - affects all users with minimax-coding-plan + OmniRoute fallback configuration.

AceRothstein71:
## What

`handleSessionError` in the runtime-fallback hook now treats an abort-classified error carrying HTTP 402 as terminal quota exhaustion and falls through to the session-stable fallback dispatch, instead of dropping the session. Two gates changed in `packages/omo-opencode/src/hooks/runtime-fallback/event-handler.ts`:

1. The `isAbortError()` early-return no longer swallows an abort that carries `statusCode: 402` (that shape is a quota signal, not a user cancellation, so retry state is preserved).
2. The `!isRetryableError()` gate now mirrors the existing `terminalQuota402Abort` carve-out already committed for the `message.updated` path (`message-update-handler.ts`): abort classification + 402 logs a fall-through line and continues to `dispatchFallbackRetry`.

## Why

 minimax-coding-plan (LiteLLM) 402 quota errors surface as abort-classified errors (PR #6677 makes this explicit via `terminal_quota_exhausted`; abort-named stream-death wrappers behave the same today). `isRuntimeFallbackRetryableError()` returns false for `"abort"`, so both runtime-fallback entry paths skipped recovery. The `message.updated` path was already fixed on dev; the `session.error` path was not - and it fires first in production. Result: sessions died with zero output tokens ("Error not retryable, skipping fallback") even with fallback models configured. This is exactly the reported failure in #6959.

## Verified

- New co-located regression suite `terminal-quota-402-session-error.test.ts` (given/when/then), failing-first:
  - RED before fix: abort+402 via `session.error` produced 0 fallback dispatches (cancellation branch wiped state); captured in `.omo/evidence/20260824-minimax-402-quota/red-before-fix.txt`.
  - GREEN after fix: exactly ONE same-session dispatch to the first fallback model with source `session.error`; genuine user aborts without 402 still perform NO dispatch; abort+402 with no fallback chain still performs NO dispatch.
- Full scoped suites green: runtime-fallback 289 pass / 0 fail (includes the committed twin-path pin `terminal-quota-402-dispatch.test.ts`), model-core 356 pass / 0 fail (untouched, sanity).
- `tsgo --noEmit -p packages/omo-opencode/tsconfig.json` exit 0; LSP diagnostics clean on both changed files.
- QA evidence recorded under `.omo/evidence/20260824-minimax-402-quota/` (red/green runs, scoped suites, typecheck, fix diff). No live OpenCode spawn: reproducing end-to-end requires a real minimax credential reaching a 402; the unit harness exercises the exact decision boundary (event -> classify -> gate -> dispatch).

## Risk

Low. Behavior changes only for errors classified `"abort"` WITH `statusCode === 402`; every other abort, cancellation, and non-retryable path is byte-for-byte unchanged and pinned by the existing 289-test suite. If every fallback model also fails, the chain still terminates cleanly via the existing dispatcher.

## Overlap with #6677

No file overlap: #6677 touches `packages/model-core` classification only; this PR touches the `omo-opencode` adapter's `session.error` handler only. They are complementary: #6677 classifies more 402 payloads as non-retryable `"abort"`; this PR guarantees those aborts still trigger model fallback on the session.error path instead of killing the session. If #6677 merges first, this PR's tests cover its output shape via the abort-classified fixture.

Fixes #6959

---
## Summary by cubic
Treats abort-classified `402 Payment Required` errors on `session.error` as terminal quota and dispatches a session-stable fallback instead of ending the session. Previously, abort+402 was treated as non-retryable cancellation and skipped fallback; now it falls through to the configured fallback chain to address Linear issue 6959.

**Review notes**
- Updates `packages/omo-opencode/src/hooks/runtime-fallback/event-handler.ts` to detect `abort` + `statusCode === 402` and bypass both the `isAbortError` state wipe and the non-retryable skip, mirroring the existing `message.updated` carve-out.
- Adds `packages/omo-opencode/src/hooks/runtime-fallback/terminal-quota-402-session-error.test.ts` to pin: one same-session dispatch on abort+402; no dispatch on genuine user aborts; no dispatch when no fallback models resolve.
- Scope is limited to `session.error` handling in `packages/omo-opencode`; other aborts and non-retryable paths are unchanged.
- Logs a specific line for `session.error` terminal-quota 402 to aid triage; no migration or config changes required.

<sup>Written for commit 25d136220b4d1e36d6047e3f00daf0cd33e1f230. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7182?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for closing the `session.error` 402 asymmetry. The blocking defect is that the 402 carve-out skips `internallyAbortedSessions.delete(sessionID)`, leaving a stale marker that can disable the assistant-active safety check on a later dispatch. Please consume the internal-abort marker before the carve-out while bypassing only cancellation and retry-reset semantics for terminal 402 aborts. Add a regression test proving the marker is removed and an internal 402 abort echo cannot cause a later double or unsafe dispatch, then run the scoped suite and CI. Diff: https://github.com/code-yeongyu/oh-my-openagent/pull/7182/files

PR #7183, opened 2026-08-24
fix(compaction): re-inject skill bodies loaded via the skill tool after compaction

### LINKED ISSUE #7158: [Bug]: Skill bodies loaded via `skill` tool are lost across compaction; agents fall back to generic tools like `read`
## Summary

OMO preserves agent/model/tools and todos across OpenCode compaction, but it does **not** checkpoint or restore skill bodies that were loaded mid-session via the `skill` tool. After compaction the agent still sees the skill *names* in the rebuilt system prompt, but the full skill instructions that were returned as tool results live in the summarized transcript and are often washed out. We observed this as an agent repeatedly calling `read` instead of the specialized `team_*` tools whose usage pattern had been loaded from a skill earlier in the session.

## Observed behavior

In a long-running session the lead agent entered a loop where it said it intended to close a team (`team_list` / `team_delete`) but actually emitted `[tool: read]` six times in a row. After a compaction ran, the user's explicit "Stop reading" instruction was preserved in the summary and the agent finally used the team tools correctly. This suggests the high-level instruction survived compaction, but the nuanced skill guidance that should have prevented the read-loop in the first place had already degraded.

### 1. Skills can enter the conversation through two paths

- **Baked into the agent system prompt** (survives compaction): agents declared with `skills: [...]` get full skill bodies prepended to their `prompt` via `resolveAgentSkills`.
  https://github.com/code-yeongyu/oh-my-openagent/blob/04acd0311e496b2521ef639fdf93a3d5b8895508/packages/omo-opencode/src/agents/agent-skill-resolution.ts#L5-L26

- **Loaded on demand via the `skill` tool** (does **not** survive compaction): the tool returns the full skill body as a tool result.
  https://github.com/code-yeongyu/oh-my-openagent/blob/04acd0311e496b2521ef639fdf93a3d5b8895508/packages/omo-opencode/src/tools/skill/tools.ts#L174-L199

### 2. Compaction only checkpoints agent/model/tools

`CompactionAgentConfigCheckpoint` contains `agent`, `model`, and `tools`, but **no `skills` field**:

https://github.com/code-yeongyu/oh-my-openagent/blob/04acd0311e496b2521ef639fdf93a3d5b8895508/packages/omo-opencode/src/shared/compaction-agent-config-checkpoint.ts#L5-L10

After compaction, `recoverCheckpointedAgentConfig` re-dispatches an internal prompt with the checkpointed agent/model/tools, but it does not re-inject any skill bodies:

https://github.com/code-yeongyu/oh-my-openagent/blob/04acd0311e496b2521ef639fdf93a3d5b8895508/packages/omo-opencode/src/hooks/compaction-context-injector/recovery.ts#L27-L115

### 3. The compaction summary is not told to preserve skill instructions

`COMPACTION_CONTEXT_PROMPT` instructs the summarizer to keep user requests, goals, work state, constraints, and delegated task IDs, but it does not mention skill instructions:

https://github.com/code-yeongyu/oh-my-openagent/blob/04acd0311e496b2521ef639fdf93a3d5b8895508/packages/omo-opencode/src/hooks/compaction-context-injector/compaction-context-prompt.ts#L3-L75

## Impact

- Agents lose mid-session skill guidance after compaction.
- They regress to base behavior (repeated `read`, `bash`, `grep`) because those tools are always in the schema and do not require remembering a skill contract.
- The post-compaction degradation monitor detects the resulting no-text tool-call tails but only re-injects agent/model/tools, not the missing skill bodies, so the loop can persist until the user intervenes.

## Proposed fix

Add skill-body checkpointing/restoration to the compaction lifecycle:

1. Extend `CompactionAgentConfigCheckpoint` with a `skills` field that records the set of skills loaded via the `skill` tool in the current session (name + loaded body or a hash).
2. In `recoverCheckpointedAgentConfig`, re-bake the checkpointed skill bodies into the system prompt (or re-dispatch the `skill` tool calls) after compaction.
3. Alternatively, include loaded skill bodies in `COMPACTION_CONTEXT_PROMPT` explicitly so the summarizer is instructed to preserve them verbatim.

A short-term mitigation is to document that skills needed across compaction should be declared in agent configs via `skills: [...]` rather than loaded on-demand.

## Related issues

- #3039 — on-demand skill loading for the main session (related feature request)
- #6594 — ultrawork directive re-arm after compaction (same family: injected instructions lost across compaction)
- #6756 — goal updates lost after compaction (same root cause: reliance on LLM summary fidelity)

## Environment

Observed on OpenCode with OhMyOpenAgent team-mode enabled, in a session that had been compacted multiple times over ~80 days / 1,500+ messages.

AceRothstein71:
## What

Skill bodies loaded mid-session via the `skill` tool now survive compaction:

- New per-session store `shared/session-loaded-skills.ts`: the `skill` tool records every successfully loaded skill (name + body) for `ctx.sessionID`. Slash-command loads are not recorded. Re-loading a skill updates the body in place.
- `CompactionAgentConfigCheckpoint` gains an optional `skills: { name, body }[]` field (deep-cloned like the rest of the checkpoint). `compaction-context-injector.capture()` merges the session's loaded skills into the checkpoint before compaction.
- `recoverCheckpointedAgentConfig()` appends the checkpointed skill bodies to the internal no-reply recovery prompt (same synthetic `compaction_continue` part as `AGENT_RECOVERY_PROMPT`), so the instructions re-enter context deterministically after summarization.
- When checkpointed skills exist, recovery is no longer suppressed by the agent/model/tools already matching (`isPromptConfigRecovered` early return) - config matching does not prove skill bodies are back in context. The existing 60s cooldown still dedupes dispatches.
- `session.deleted` clears the per-session skill store together with the checkpoint.

## Why

After OpenCode compaction, agents kept seeing skill *names* but lost the bodies that were only present as tool results, and regressed to generic tools (`read`/`bash`/`grep`) instead of honoring loaded skill contracts (observed as read-loops instead of `team_*` usage). The existing retention precedent is deterministic checkpoint + re-injection (agent/model/tools); this extends exactly that mechanism instead of relying on LLM summary fidelity.

## Verified

- Failing-first regression tests, then green:
  - `skill-reinjection.test.ts`: capture -> compaction -> dispatched recovery part contains skill name and body; gate-bypass case dispatches exactly once across repeated compaction events; no-skills sessions keep prior behavior (no dispatch on config match).
  - `session-loaded-skills.test.ts`: record/dedupe/ordering/isolation/clear/reset.
  - `tools.factory.test.ts`: skill loads are recorded for the session; command loads are not.
- Scoped suites: 102 pass / 0 fail (compaction-context-injector + tools/skill + new store + background-agent checkpoint consumer).
- Wider `src/shared/` suite: 1079 pass / 5 skip / 1 pre-existing environment failure (`dist-bundle-prompt-content.test.ts` requires a built `dist/index.js`; bundle build is blocked in this worktree by unresolvable `packages/shared-skills/upstreams/*` git submodules - unrelated to this patch).
- `bun run typecheck` exit 0.
- QA evidence: `.omo/evidence/20260824-skill-bodies-compaction/` (commands, outputs, isolation notes).

## Risk

Low. The new checkpoint field is optional and ignored by existing consumers; behavior for sessions without on-demand skill loads is unchanged (pinned by the untouched recovery suite). Re-injected bodies add post-compaction context proportional to the skills actually loaded in the session; duplicates are bounded by the recovery cooldown and one-shot compaction events.

## Coordination

No overlap with #7152 (ContextBudgetPolicy): that PR tunes what fits into the active budget around the physical window; this change re-injects lost skill bodies through the existing post-compaction recovery path and does not touch budget accounting or the summarizer prompt/budget model.

Fixes #7158

---
## Summary by cubic
Re-injects skill bodies loaded mid-session via the `skill` tool after compaction so agents retain those instructions. Previously compaction preserved skill names but washed out bodies; now we checkpoint and restore bodies during post-compaction recovery, even when agent/model/tools already match.

- New per-session store records loaded skills (name + body); reloading updates the body; cleared on `session.deleted`.
- `CompactionAgentConfigCheckpoint` adds optional `skills`; capture merges the session store and deep-clones it.
- Recovery appends checkpointed skill bodies to the internal no-reply continuation; when skills exist, config-match no longer suppresses recovery; the existing cooldown still dedupes.
- The `skill` tool records loaded skills for the session; slash-command loads are not recorded.
- Backward compatible: the new field is optional; sessions without loaded skills behave unchanged. Context increases only with skills actually loaded.

Fixes #7158.

<sup>Written for commit 68ff199d09845570c9159ec06066c32f7bf98947. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for restoring skill context through the compaction checkpoint. The blocking defect is unbounded verbatim reinjection on every later compaction: plausible skill sets add about 29K tokens, the worst case about 82K, and the checkpoint is not cleared after recovery. Please cap each skill and the total payload with explicit truncation guidance, or clear recovered skills after confirmed success so each load is reinjected once. Add regression tests proving the bound or one-shot behavior across multiple compactions beyond the cooldown, and rerun the injector, skill, and typecheck gates. Diff: https://github.com/code-yeongyu/oh-my-openagent/pull/7183/files

PR #6666, opened 2026-08-09
fix: fail fast when sync subagent stalls with finish="unknown"

Y4nTsing:
# PR: Fail fast when a sync subagent stalls with `finish="unknown"` (no more 30-min waits)

**Branch target**: `dev`
**Related issue**: #6665

## Summary

When a subagent model stream is interrupted mid-turn, OpenCode can persist an empty assistant message with `finish="unknown"` after the session becomes idle. Previously, the parent remained blocked until the 30-minute inactivity timeout.

This PR adds a bounded 30-second stall window for that exact state:

- a substantive text/reasoning deliverable from the current user turn is preserved and treated as complete;
- otherwise the dead session is aborted with a descriptive stall error;
- missing finish values, tool-call turns, active/uncertain statuses, and legitimate child/wake waits remain on the normal polling path.

## Changes

- Treat the session as inactive for stall purposes when a successful status response reports it `idle` **or omits it entirely**. Real OpenCode (verified end-to-end on 1.18.15, evidence under `.omo/evidence/20260901-6666-stall-e2e/`) drops idle sessions from the status map, so absence-from-a-successful-response is the real-world idle signal. A thrown status fetch or an unrecognized status value still resets the stall timer while terminal-message detection remains available.
- Track a deterministic progress signature over relevant message IDs, finish reasons, part types, and part text, so in-place same-length mutations reset the stall window.
- Apply continuation anchors before selecting the latest user turn. Old turns cannot trigger a stall or masquerade as a deliverable, while earlier assistant retries within the current turn remain eligible.
- Keep `busy`, `retry`, and `running` periods out of both the inactivity and stall windows.
- Preserve direct-child and pending-parent-wake exemptions during sync continuation by passing receiver-bound manager predicates to the poller.
- Inject `now` and `wait` only as internal test seams; production defaults remain `Date.now` and the existing Atomics-based wait.
- Replace global `Date.now` mutation in timeout tests with deterministic per-call clocks, avoiding cross-file test races.

## Verification

```text
sync stall regressions ............... 23 pass, 0 fail (incl. absent-status-map cases)
sync continuation tests .............. 17 pass, 0 fail
sync poll timeout tests .............. 6 pass, 0 fail
packages/omo-opencode typecheck ....... PASS
git diff --check ...................... PASS
```

Real isolated OpenCode evidence (opencode 1.18.15, source-loaded plugin, interrupted-stream mock provider, `.omo/evidence/20260901-6666-stall-e2e/`):

[...]

PR #7186, opened 2026-08-24
fix(background-agent): resolve background_output by session-ID fallback and enrich not-found guidance

### LINKED ISSUE #7010: background_output returns "Task not found" for IDs the completion notification provides, and completion does not resume the turn
## Summary

Two separate defects in the background-task lifecycle make long parallel workflows stall. In a single long session I hit both repeatedly and had to ask the human to nudge the agent forward more than five times.

1. `background_output(task_id="bg_...")` returns `Task not found` for IDs that the completion notification itself instructed me to use.
2. The `[BACKGROUND TASK COMPLETED]` / `[ALL BACKGROUND TASKS COMPLETE]` notification does not resume the agent turn, so the conversation halts until a human sends a message.

## Environment

- `oh-my-openagent` **4.19.4** (from `~/.cache/opencode/packages/oh-my-openagent@4.19.4/`)
- `@opencode-ai/plugin` / `@opencode-ai/sdk` **1.15.13**
- Platform: Linux
- Main agent: Sisyphus on `anthropic/claude-opus-5`
- Background subagents used: `explore`, `metis`, `momus`, `oracle`

## Defect 1 — `background_output` cannot find IDs the notification just handed out

The system reminder says verbatim:

```
Use `background_output(task_id="bg_28094f9c")` to retrieve this result when ready.
```

Calling exactly that returns:

```
Task not found: bg_28094f9c
```

### What worked vs. what failed, in one session

| Background task IDs | `background_output` result |
| --- | --- |
| `bg_07326450`, `bg_3e75a838`, `bg_15e81744`, `bg_1e2c0dc3`, `bg_1e5417de`, `bg_6c0660fe`, `bg_49e8c7a5`, `bg_1af23ad3` (first batch, 8 launched together) | **all succeeded** |
| `bg_28094f9c` | `Task not found` |
| `bg_ca9d07e6` | `Task not found` |
| `bg_f87e9571` | `Task not found` |
| `bg_e38dfc72`, `bg_908fac10`, `bg_9cadb90a`, `bg_67ded596` | not attempted — already using the workaround |

### Hypothesis (falsifiable, please verify)

Every failing task was launched **after the first synchronous `task()` call in the session** (i.e. a `task(...)` without `run_in_background: true`). Every succeeding task was launched before any synchronous `task()` existed in that session.

That suggests a synchronous `task()` invocation clears, replaces, or otherwise invalidates the background-task registry, so background tasks launched afterwards are never registered under their advertised `bg_` id — even though the task itself runs to completion and its session is fully readable.

I did not verify the mechanism; this is an observed correlation, offered as a starting point rather than a root cause.

### Workaround that reliably works

`session_read(session_id="ses_...")` using the `session_id` from the launch metadata returns the full result every time, including for every id that `background_output` rejected. So the work is never lost — only the documented retrieval path is broken.

Note also that `background_output` returns a helpful message when given a `ses_...` id:

```
background_output expects a background task ID such as `bg_...`, not a session ID.
```

...which is correct, and confirms the `bg_` id really is the intended handle.

### Suggested fix directions

- Make `background_output` fall back to resolving by session id / persisted task record instead of an in-memory map that can be dropped.
- Or persist the `bg_ -> ses_` mapping for the lifetime of the session so a completed task stays retrievable.
- At minimum, the error should say *why* it is not found (evicted / never registered / expired) and point at the `session_read` fallback, since the notification text currently promises a call that fails.

## Defect 2 — completion notification does not resume the turn

Every completion notification in the session arrived carrying both of these markers:

``````

The tool description says:

> System notifies on completion, so `block=true` rarely needed.
> Do NOT call this tool AFTER receiving a `<system-reminder>` notification for the task.

...which sets the expectation that the agent resumes automatically. In practice the conversation stalled at every single completion boundary, and the human had to type a message ("continue") to unblock it — more than five times in one session.

The combination of `INITIATOR` (an auto-continuation was attempted) and `NOREPLY` (this message does not warrant a reply) looks like the contradiction: the auto-resume fires but is immediately marked as not requiring an assistant turn, so no turn is generated.

### Impact

This is the difference between "fire 8 parallel agents and keep working" and "fire 8 parallel agents and then babysit the terminal". For long orchestration runs — the exact use case background agents exist for — it removes most of the benefit.

### Suggested fix direction

Do not mark background-completion notifications `NOREPLY`, or introduce a distinct marker that permits/forces exactly one continuation turn.

## Possibly related (minor, low confidence)

Every delegated subagent printed:

```
⚠️  Model routing: parent used anthropic/claude-opus-5, this subagent used openai/gpt-5.6-luna-fast (via category: unknown)
```

even when dispatched with an explicit `subagent_type` (e.g. `subagent_type="explore"`) and **no** `category`. The `via category: unknown` phrasing suggests the routing metadata does not record that the dispatch was by `subagent_type`. Mentioning it only in case it shares a code path with the registry issue above; it did not block anything.

## Repro sketch

1. Launch several background tasks with `task(..., run_in_background=true)` and collect them with `background_output` — this works.
2. Make one synchronous `task(...)` call (no `run_in_background`).
3. Launch another background task, wait for `[BACKGROUND TASK COMPLETED]`.
4. Observe: the turn does not resume until a human message arrives (Defect 2).
5. Call `background_output` with the exact id from the notification. Observe: `Task not found` (Defect 1).
6. Call `session_read` with the `ses_...` id from the launch metadata. Observe the full result.

AceRothstein71:
## What

Closes the remaining resolution gap behind symptom (a) of #7010:

- `BackgroundManager.getTaskBySessionId(sessionID)`: reverse lookup across all three resolution layers (live task map via existing `findBySession` → `completedTaskArchive` scan → persisted `globalThis` registry via new `findRegisteredBackgroundTaskBySessionId`).
- `background_output` now accepts a session ID (`ses_...`) and reverse-resolves it to the registered background task, with the same single-retry symmetry the `bg_` path already has. The documented primary handle remains `bg_...`.
- Unknown `bg_` misses now explain the likely reasons (evicted from the capped in-process archive / never registered in this process / mistyped) and point at the `session_read(session_id="ses_...")` fallback, as requested in the issue thread.
- Two metadata tests that pinned the old refusal contract were updated to pin the new contract with equally strict assertions; no test was weakened or deleted.

## Why

Issue #7010 symptom (a): `background_output(task_id="bg_...")` returned `Task not found` for IDs the completion notification itself had just advertised. There is no ID-namespace mismatch — the notification template emits `task.id` verbatim and lookups key on the same string. The failure is premature eviction: on 4.19.4 the only layer was the live in-memory map, and removal after notification made the advertised ID unresolvable; on current dev the three-layer hardening covers the common window but leaves gaps (in-process-only layers, 100-entry eviction caps, no session-ID reverse resolution, unexplained error). The reporter's proven workaround (`session_read` by `ses_`) required manually carrying the session ID; this change makes the tool do that resolution itself.

Symptom (b) (completion does not resume the parent turn) is intentionally NOT touched here: it is owned by open PR #7166 (force reply production when the parent is idle), with starvation ceilings from #7117/#6716 and lost-wake reconciliation from #6655.

## Verified

- Failing-first: new tests `(7 fail / 1 pass)` before implementation, all green after.
- Scoped suites: `bun test packages/omo-opencode/src/tools/background-task/ packages/omo-opencode/src/features/background-agent/` → **793 pass / 0 fail**.
- `bun run typecheck` (tsgo root + script + all workspace packages) clean; LSP diagnostics clean on changed files.
- Real-harness QA per repo protocol: built local dist, loaded it into real opencode 1.18.22 inside an isolated XDG sandbox (`script/agent/qa-sandbox.sh`), plugin booted and served sessions with no init errors; built artifact verified to contain the change; host DB session count unchanged (18260 → 18260). Live end-to-end background-completion cycle omitted (no provider credentials in this environment) — recorded in evidence.
- Evidence: `.omo/evidence/20260824-bg-output-task-not-found/` (plan, root-cause + coordination analysis, QA evidence).

## Risk

Low / additive: new lookup methods and message text only; no wake-ledger, prompt-dispatch, or removal-lifecycle changes. Session-ID fallback is a recovery path; when multiple tasks share a session (retry attempts), first match in layer order wins.

## Coordination notes vs related PRs

| PR | Relationship |
| --- | --- |
| #7166 | Owns symptom (b); different files (flush-runner/prompt-dispatch); no overlap. |
| #7117 | Complements (b) via stale-tool-block ceiling; no overlap. |
| #6655 | Touches manager.ts (pendingByParent/notifiedTasks reconciliation) but not any lookup layer; only hunk-adjacency conflict risk. |
| #6716 | Flush-runner defer ceilings; no overlap. |
| #6785 | manager.ts shutdown flush; unrelated to lookup. |
| #6739 | compaction-aware-message-resolver; no overlap. |
| #6666 | sync subagent stall fail-fast; nearest to the reported sync-`task()` correlation area but does not touch the registry. |

No PR in that set touches `create-background-output.ts`, `clients.ts`, `task-registry.ts`, or the `getTask` lookup family.

Fixes #7010

---
## Summary by cubic
Closes the “Task not found” gap in `background_output` by adding session-ID reverse lookup and clearer guidance. Previously `background_output(task_id="bg_...")` could miss after in-process eviction; now the tool also accepts `ses_...` and resolves it to the task with a single retry.

- Adds `BackgroundManager.getTaskBySessionId(sessionID)` that searches live tasks, the completed-task archive, then the persisted registry.
- Adds `findRegisteredBackgroundTaskBySessionId` in the task registry.
- `background_output` now resolves `ses_...` IDs via the new lookup with the same retry symmetry as `bg_...`.
- Not-found messages explain likely causes (evicted, never registered after reload, mistyped) and point to `session_read`.
- Tests added/updated to pin the new contract; no coverage weakened.

**Migration**
- If you construct or mock `BackgroundOutputManager`, you must implement `getTaskBySessionId`.

<sup>Written for commit 3eea946379f1f3855b819f94878ffad4a5d434b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7186?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for the tested session-ID fallback. The blocking gate is duplication: #7225 contains the same production patch, so both open PRs cannot proceed independently. Please consolidate into one PR and close the duplicate. In the surviving PR, state accurately that `bg_` misses receive improved guidance while reverse recovery occurs only when a `ses_` ID is supplied, reference both reported issues, and disclose the manager interface widening. Preserve regression proof for direct lookup, session-ID fallback across active and archived tasks, polling, and `bg_` miss behavior, plus typecheck. Diff: https://github.com/code-yeongyu/oh-my-openagent/pull/7186/files

PR #7190, opened 2026-08-24
fix(model-core): add claude-opus-5 to GA 1M context allowlist

### LINKED ISSUE #6640: preemptive-compaction: claude-opus-5 falls back to 200K context limit (hasGA1MContext regex gap) + >60s summarize timeout causes guaranteed double compaction
## Summary

Two related defects in the `preemptive-compaction` hook cause premature and duplicate auto-compactions for `claude-opus-5`. Verified in **4.16.0** (incident) and confirmed **still present in 4.19.4** (latest, tarball inspected).

## Bug 1: `hasGA1MContext()` allowlist regex is missing `claude-opus-5` → 200K fallback despite a 1M model

`packages/model-core/src/context-limit-resolver.ts`:

```js
function hasGA1MContext(modelID) {
  return /^claude-(opus|sonnet)-4(?:-|\.)(?:6|7|8)(?:-high)?$/.test(modelID)
      || /^claude-(?:fable|mythos|sonnet)-5$/.test(modelID);
}
```

- `claude-sonnet-5` matches the second alternation → 1M ✅
- **`claude-opus-5` matches neither** (first is opus-**4**-only; second allows only fable|mythos|sonnet-5) → `resolveActualContextLimit()` returns `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200000` ❌

With `PREEMPTIVE_COMPACTION_THRESHOLD = 0.78`, the hook compacts opus-5 sessions at **156K tokens** — 15.6% of the model's real 1M window. Note even a cached provider limit is discarded unless the regex matches (`if (cachedLimit && hasGA1MContext(modelID))`), and user/provider config `limit.context` is never consulted for Anthropic models. The allowlist design guarantees this recurs on every new 1M Anthropic model release.

**Observed:** session with `claude-sonnet-5` ran to 518K tokens with no compaction (correct); switching the same session to `claude-opus-5` fired auto-compaction on the very first turn at 249K tokens.

## Bug 2: `compactedSessions` latch is set only *after* the awaited summarize → client timeout guarantees a double compaction

`packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts` (`runPreemptiveCompactionIfNeeded`):

```js
await withTimeout3(ctx.client.session.summarize({... auto: true ...}),
                   PREEMPTIVE_COMPACTION_TIMEOUT_MS2 /* 60000 */, ...);
compactedSessions.add(sessionID);   // <-- never reached if summarize takes > 60s
```

A real compaction (large session, opus-tier summarizer) routinely takes >60s. When `withTimeout3` rejects:

1. the server-side summarize **still completes successfully**,
2. `compactedSessions.add()` is skipped (we land in `catch`),
3. `tokenCache` still holds the last *pre-compaction* assistant message (summary messages are excluded from the cache refresh via the `isCompactionAgent` check),
4. `PREEMPTIVE_COMPACTION_COOLDOWN_MS = 60000` has already elapsed by the next `tool.execute.after`,

→ the hook immediately fires a **second** `summarize({auto:true})`, even though actual context usage post-compaction was far below threshold (observed: second fire at ~108K actual tokens, driven by the stale 249K cache entry).

## Repro / evidence

Timeline from a live session (2026-08-07, opencode 1.18.15, plugin 4.16.0):

| Time (UTC) | Event | tokens (input+cache.read) |
|---|---|---|
| 16:38:23 | auto compaction #1 fired | 249,169 (0.78×200K=156K exceeded; 1M limit would NOT fire) |
| 16:39:24 | hook log: `Compaction failed ... timed out after 60000ms` | — |
| 16:39:27 | server-side summary completes anyway | — |
| 16:40:36 | auto compaction #2 fired | actual ~108K (stale cache: 249K) |
| 16:41:36 | hook log: second 60s timeout | — |

Resolver behavior is directly reproducible by executing the bundled `resolveActualContextLimit` with `("anthropic", "claude-opus-5", {anthropicContext1MEnabled:false})` → `200000`, vs `claude-sonnet-5` → `1000000`.

## Suggested fixes

1. Add `opus` to the 5-family alternation (`/^claude-(?:fable|mythos|sonnet|opus)-5$/`) — or better, derive the actual limit from the provider/model config (`model.limit.context`) instead of a hardcoded allowlist that must be hand-updated per model release.
2. Set the re-entrancy latch (or a pending marker) *before* awaiting summarize, and/or confirm completion via the `session.compacted` event rather than the client-side await; alternatively raise the 60s timeout above realistic compaction duration and make the cooldown ≥ timeout.

Environment: Linux, opencode 1.18.15, oh-my-openagent 4.16.0 (bug confirmed unchanged in 4.19.4 dist).

AceRothstein71:
## What

Add `opus` to the 5-family alternation in `hasGA1MContext()` (`packages/model-core/src/context-limit-resolver.ts`):

```diff
-    /^claude-(?:fable|mythos|sonnet)-5$/.test(modelID)
+    /^claude-(?:fable|mythos|sonnet|opus)-5$/.test(modelID)
```

Plus three co-located regression tests in `context-limit-resolver.test.ts` (given/when/then): GA 1M for `claude-opus-5` without explicit 1M mode, cached provider limit honored for `claude-opus-5`, and a guard test keeping the 200K default for non-GA Anthropic models.

## Why

`hasGA1MContext()` is the allowlist gate for Anthropic models whose real context window is 1M. Its 5-family pattern listed only `fable|mythos|sonnet`, so `claude-opus-5` matched neither alternation and `resolveActualContextLimit("anthropic", "claude-opus-5", ...)` fell back to `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200_000`. The same regex gate also discarded a cached provider limit (`if (cachedLimit && hasGA1MContext(modelID))`). The preemptive-compaction hook compacts at 0.78 of the resolved limit, so opus-5 sessions were auto-compacted at ~156K tokens, about 15.6% of the model's actual 1M window, producing premature and duplicate auto-compactions as reported in the issue.

## Verified

- Failing-first: before the fix, exactly the two new `claude-opus-5` tests failed (`13 pass / 2 fail`); all pre-existing tests and the guard test passed.
- After the fix: scoped file `15 pass / 0 fail`; full `bun test packages/model-core` suite `359 pass / 0 fail`.
- Typecheck green for `packages/model-core` and consumer package `packages/omo-opencode` (`tsgo --noEmit`, exit 0).
- Evidence recorded in `.omo/evidence/20260824-6640-ga1m-context-gap/evidence.md`.

## Risk

Low. One-token allowlist widening inside a pure function; exported signatures unchanged; negative behavior for non-GA Anthropic models is pinned by a guard test so the allowlist does not become allow-all. Live-harness QA was not run because the change is in harness-neutral `model-core` with no OpenCode API surface touched; behavior is pinned at unit level.

Out of scope: Bug 2 from the issue (the `compactedSessions` latch is set only after the awaited summarize, so a >60s client timeout guarantees a second auto-compaction). That defect lives in `packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts`, needs a design decision among the alternatives proposed in the issue, and touches files adjacent to #7183; it deserves its own PR.

Fixes #6640

---
## Summary by cubic
Adds `claude-opus-5` to the GA 1M context allowlist so `resolveActualContextLimit` returns 1,000,000 and honors a cached provider limit. Previously `claude-opus-5` fell back to 200,000, triggering preemptive compaction around 156k tokens and occasional duplicate compactions (issue #6640).

**Review notes**
- Changes one regex in `packages/model-core/src/context-limit-resolver.ts`; no API or signature changes.
- Adds three tests in `context-limit-resolver.test.ts`: resolves 1M for `claude-opus-5`, uses cached limit, and a guard that keeps 200K for non-GA Anthropic models.
- All `packages/model-core` tests pass; typecheck passes for `packages/model-core` and consumer `packages/omo-opencode`.
- No migration required; scope is limited to Anthropic `claude-opus-5`. The compaction latch ordering noted in the issue is out of scope.

<sup>Written for commit 41933b8c78983877add255d6ceb8d0cff83c97aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7190?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260824-6640-ga1m-context-gap/evidence.md
+++ b/.omo/evidence/20260824-6640-ga1m-context-gap/evidence.md
@@ -1,0 +1,77 @@
+# Evidence: Issue #6640 - hasGA1MContext regex gap (claude-opus-5 falls back to 200K)
+
+Date: 2026-08-24
+Worktree: /home/viprix/projects/oom-wt-6640
+Branch: issue/6640-ga1m-context-gap (base: dev @8833800ae)
+
+## Root cause
+
+`packages/model-core/src/context-limit-resolver.ts` `hasGA1MContext()`:
+
+- Before: `/^claude-(?:fable|mythos|sonnet)-5$/` for the 5-family alternation.
+  `claude-opus-5` matches neither the opus/sonnet-4 pattern nor the 5-family
+  pattern, so `resolveActualContextLimit("anthropic", "claude-opus-5", ...)`
+  returns `DEFAULT_ANTHROPIC_ACTUAL_LIMIT = 200_000` even though the model has a
+  1M context window. The same gap also discards a cached provider limit because
+  the cache branch is gated on `cachedLimit && hasGA1MContext(modelID)`.
+- Consumer: `packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts`
+  compacts when usage >= `PREEMPTIVE_COMPACTION_THRESHOLD = 0.78` of the resolved
+  limit -> opus-5 sessions compacted at ~156K tokens (15.6% of the real 1M
+  window), producing premature/duplicate auto-compactions as reported in #6640.
+- Fix: add `opus` to the 5-family alternation:
+  `/^claude-(?:fable|mythos|sonnet|opus)-5$/` (issue suggested fix #1).
+  Bug 2 of the issue (compactedSessions latch set after the awaited summarize)
+  is intentionally out of scope for this PR; it is an independent defect in
+  packages/omo-opencode/src/hooks/preemptive-compaction-trigger.ts.
+
+## WHAT WAS TESTED
+
+1. Failing-first regression tests added to the co-located
+   `packages/model-core/src/context-limit-resolver.test.ts` (given/when/then):
+   - "returns GA 1M for claude-opus-5 without explicit 1M mode"
+   - "uses cached limit for claude-opus-5 when cache exists" (proves the cached
+     limit is no longer discarded by the regex gate)
+   - "keeps 200K default for Anthropic models outside the GA 1M allowlist"
+     (guard: allowlist must not become allow-all; green before and after)
+2. Command: `bun test packages/model-core/src/context-limit-resolver.test.ts`
+3. Full package suite: `bun test packages/model-core`
+4. Typecheck: `bunx tsgo --noEmit -p packages/model-core/tsconfig.json` and
+   `bunx tsgo --noEmit -p packages/omo-opencode/tsconfig.json` (consumer package).
+
+## WHAT WAS OBSERVED
+
+- RED (before fix): exactly the two new claude-opus-5 regression tests failed;
+  all 12 pre-existing tests plus the guard test passed.
+  `(fail) resolveActualContextLimit > returns GA 1M for claude-opus-5 without explicit 1M mode`
+  `(fail) resolveActualContextLimit > uses cached limit for claude-opus-5 when cache exists`
+  Summary line: `13 pass / 2 fail`.
+- GREEN (after fix): scoped file `15 pass / 0 fail`; full model-core package
+  `359 pass / 0 fail` (1295 expect() calls, 33 files).
+- Typecheck: model-core OK; omo-opencode (consumer) OK, exit 0.
+
+## WHY IT IS ENOUGH
+
+The change is a one-token alternation inside a pure, deterministic function.
+The co-located unit suite pins both affected branches of
+`resolveActualContextLimit` (hardcoded GA limit and cached provider limit) for
+the exact unmatched format from the issue (`claude-opus-5`), pins negative
+behavior for a non-GA Anthropic model, and the full model-core suite (359 tests)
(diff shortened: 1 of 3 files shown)


MoerAI:
Thanks for the PR, @AceRothstein71. The diagnosis and failing-first model-core tests are correct.

I am closing this as superseded by #7260, which carries this fix and test coverage while also handling the real `claude-opus-5-fast` model ID that this branch leaves at the 200K fallback.

Before #7260 lands, remove the maintainer-local worktree line from `.omo/evidence/20260824-6640-ga1m-context-gap/evidence.md`; it fails the repository's markdown path audit and was copied into the replacement. With that evidence cleanup, #7260 is the complete successor.

Issue #7115, opened 2026-08-21
[Bug]: claude-sdk-oauth leaks host tool denial message into tool turns


shrimpwtf:
## Prerequisites

- [x] I will write this issue in English.
- [x] I searched existing issues. #6517 covers a related PreToolUse continuation problem, and #7084 covers Claude SDK OAuth account-pool bookkeeping; neither identifies this denial hook as the source.
- [x] I am using the latest version currently installed (`omo 5.0.0-0.beta.14`, Senpi `2026.8.21-2`).
- [x] I read the documentation and investigated the installed runtime source.

## Bug Description

When using the Senpi edition with the `claude-sdk-oauth` provider, the following internal host-execution message is surfaced in the conversation/session output and repeats around tool calls:

```text
This tool call is captured and executed by the host. Do not retry with other tools; end the turn.
```

The message is not coming from `~/.claude/settings.json` or the Codex plugin. It is hard-coded in the installed Senpi Claude SDK OAuth adapter:

```text
/home/shrimp/.bun/install/global/node_modules/omo-ai/node_modules/@code-yeongyu/senpi/dist/core/extensions/builtin/claude-sdk-oauth/tools.js
```

The source map resolves this to `src/core/extensions/builtin/claude-sdk-oauth/tools.ts`, where `HOST_TOOL_EXECUTION_DENIED_MESSAGE` is returned by `HOST_TOOL_DENIAL_HOOKS`. The hook matches `Bash|Write|Edit|Read|Grep|Glob|mcp__custom-tools__.*` and returns `continue: false` with a denied permission reason. `options.js` wires `hooks: HOST_TOOL_DENIAL_HOOKS` into every Claude Agent SDK query.

## Steps to Reproduce

1. Install and run the Senpi edition of oh-my-openagent.
2. Log in with `/login claude-sdk-oauth`, or otherwise have an explicit Claude SDK OAuth account in the Senpi agent auth store.
3. Select a `claude-sdk-oauth` model such as `claude-sonnet-5`.
4. Ask the agent to use a host tool such as Read or Bash.
5. Observe the denial text in the session/conversation output.

## Expected Behavior

The internal host/SDK handoff should not be exposed as a repeated model-facing denial message. Either the host should complete the tool handoff transparently, or the UI should show a clear non-error status explaining that the turn is being handed back to the host.

[...]

PR #1107, opened 2026-01-25
feat(tools): inject current model info into slashcommand and skill output

MadKangYu:
## Summary

- Add `setSessionModel`/`getSessionModel` to session state for tracking current model
- Store model info from `chat.message` hook
- Display `**Current Model**: provider/modelID` in slashcommand and skill output

Fixes model name inconsistency where system prompt shows exact model ID (e.g., `claude-opus-4-5-20251101`) but slashcommand/skill output didn't reflect this.

---
## Summary by cubic
Show the current model (provider/modelID) in slashcommand and skill outputs, and improve the session list with titles/previews, timestamps, and change summaries. Tracks the model per session from chat.message to fix model name inconsistencies.

- **New Features**
  - Session state: setSessionModel, getSessionModel, clearSessionModel to track per-session model.
  - Capture providerID/modelID from chat.message and store per session.
  - Display "Current Model: provider/modelID" in slashcommand and skill outputs; tools accept sessionID to fetch it.
  - Session list: pass metadata to formatter and show Title/Preview, Updated (YYYY-MM-DD HH:MM), and Changes (files/+additions/-deletions).

<sup>Written for commit 740e6b1759762d1b3af42b9bc2723517a536fade. Summary will update on new commits.</sup>

PR #7192, opened 2026-08-24
fix(omo-ai): strip cursor-cli tool protocol frames from assistant messages

### LINKED ISSUE #7169: bug(omo-ai): cursor-cli-oauth leaks raw tool protocol into assistant text
## Summary

`omo-ai` renders Cursor CLI internal tool protocol as literal assistant text when the selected provider is `cursor-cli-oauth`. Long `<cursor-cli-tool>{...}</cursor-cli-tool>` JSON blocks are mixed into model prose and persisted in conversation history.

## Environment

- `omo 5.0.0-0.beta.18`
- engine: `senpi 2026.8.23`
- provider: `cursor-cli-oauth`
- macOS Darwin 25.5.0, arm64
- Node.js 24.14.0

## Reproduction

1. Authenticate `cursor-cli-oauth`.
2. Start OmO with a Cursor model.
3. Ask the model to inspect the workspace so Cursor invokes shell or glob tools.
4. Observe the assistant turn in the TUI and saved JSONL session.

## Actual behavior

Started and completed provider tool events appear as `<cursor-cli-tool>` JSON inside assistant text. The current upstream Senpi test explicitly expected these protocol frames to become text, so this is not machine-local configuration drift.

## Expected behavior

Cursor already executes these tools in its subprocess. OmO should show only final assistant prose and must not convert the events to host `toolCall` blocks, which could execute operations twice.

## Root cause and fix

`omo-ai` pins `@code-yeongyu/senpi@2026.8.23`. Senpi `cursor-cli-oauth/stream.ts` passed every Cursor `tool_call` event through `renderToolFrame()` and `pushTextDelta()`. The provider-boundary fix is in https://github.com/code-yeongyu/senpi/pull/1107; OmO will need to bump its exact Senpi pin after release.

AceRothstein71:
## What

Adds the `cursor-tool-frame-filter` component to the omo-senpi adapter. At the `message_end` boundary it strips Cursor CLI tool protocol frames (`<cursor-cli-tool>{...}</cursor-cli-tool>`) from assistant text blocks and drops text blocks that held only frames, then returns the sanitized message as the `MessageEndEventResult.message` replacement. Also registers the component in the extension composition and rebuilds the committed extension bundle.

## Why

Issue #7169: with the `cursor-cli-oauth` provider selected, the pinned `@code-yeongyu/senpi@2026.8.23` engine serializes every started and completed Cursor `tool_call` event through `renderToolFrame()` into `<cursor-cli-tool>` JSON strings and routes them via `pushTextDelta()`, so provider-internal protocol lands in assistant prose and persists into session JSONL. The upstream fix (senpi PR #1107) is not yet released, so the exact pin cannot be bumped; this is the adapter-side boundary fix in the only repo-owned code that runs inside the senpi process.

Design notes:

- Frames are removed at the message persistence boundary, not trimmed at display time: senpi applies the `message_end` replacement via `_replaceMessageInPlace(event.message, ...)` before `sessionManager.appendMessage(event.message)`, so cleaned text is what gets stored and rendered (verified against the installed pinned dist).
- Frames are never converted to host `toolCall` blocks, so Cursor-executed operations cannot run twice.
- Stripping is conservative: fenced code blocks are skipped entirely, and a span only counts as a frame when its payload starts with `{` and either parses as JSON or carries the upstream over-budget `...[truncated]` suffix. Prose and code examples that merely resemble the protocol survive byte-identical.
- Gated by `omo-senpi-cursor-tool-frame-filter-disabled`.

Relation to #7115: same class of bug (internal provider/host protocol surfacing as assistant-visible content in an oauth lane). That issue's fix crosses the Senpi dependency boundary and remains open; this fix uses the same lesson, sanitize at the provider boundary before content reaches assistant messages, applied at the seam this repo controls.

## Verified

- RED first: the new regression tests fail before the implementation exists (module not found for `./filter` / `./index`).
- GREEN after: 13/13 new tests pass (`bun test packages/omo-senpi/src/components/cursor-tool-frame-filter/`). Fixtures byte-mirror the pinned dist `renderToolFrame()` output including the truncated over-budget shape; assertions cover frame removal, verbatim prose, byte-identical protocol-resembling code fences, empty-content collapse for frame-only turns, untouched user messages, and the disable flag.
- Scoped suites: extension composition 64/64, components 2013 pass with one pre-existing environment failure (ulw-loop cross-session probe fails identically on the clean base, verified via stash), installer + skills sync 34/34, src-root audit gates (bundle-size, bundle-purity, package-shape, plugin-manifest, runtime audits) all green.
- Typecheck: `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` exit 0.
- QA evidence recorded under `.omo/evidence/20260824-cursor-oauth-tool-leak/` (local; the directory is gitignored by current repo policy), including the upstream leak source excerpt and the persistence-ordering proof lines from the installed senpi dist.
- Environment note: the full `test:senpi` chain cannot start on this machine because its first step (`build:materialize-frontend`) hits a pre-existing unfetchable submodule revision (`shared-skills/upstreams/open-design`); every downstream gate of that chain was executed individually instead.

## Risk

Low. The component is additive, flag-gated, and returns a replacement only when a frame was actually removed. Residual risk is a future senpi release changing the private frame format; the fixtures name the exact upstream file and function so drift is reviewable, and the component can be retired once the pin includes the upstream fix.

Fixes #7169

---
## Summary by cubic
Removes Cursor CLI tool protocol frames from assistant messages to prevent internal `cursor-cli-oauth` events from leaking into prose and session history. Old: `<cursor-cli-tool>{...}</cursor-cli-tool>` strings appeared in assistant text; New: frames are stripped at `message_end` and frame-only blocks are dropped before persistence. Fixes #7169.

- Registers `cursor-tool-frame-filter` in `omo-senpi`: sanitizes assistant text blocks, never converts frames to host tool calls, and ensures storage/rendering use the sanitized content.
- Conservative stripping: skips fenced code blocks; only removes spans that start with `{` and either parse as JSON or end with `...[truncated]`.
- Feature flag: disable via `omo-senpi-cursor-tool-frame-filter-disabled`.
- Review focus: `src/components/cursor-tool-frame-filter/filter.ts`, `src/components/cursor-tool-frame-filter/index.ts`, and registration in `src/extension/component-list.ts`; committed bundle updated under `plugin/extensions/*`.
- Tests added cover frame removal, truncated bodies, fence preservation, frame-only turns collapsing to empty content, user messages untouched, and the disable flag.

<sup>Written for commit 3a882c63562780c9c02bf24eb57511fb2f8962ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Closing as superseded by upstream senpi#1107, which is included in the `@code-yeongyu/senpi@2026.8.31` version pinned by current `dev`:
- https://github.com/code-yeongyu/senpi/pull/1107
- https://github.com/code-yeongyu/oh-my-openagent/blob/b5cbae3fb8778a06f70aa7aada35c8be72f0dba0/packages/omo-senpi/package.json

This fallback should not be revived unchanged. It accepts any JSON object between tags rather than validating the Cursor protocol schema, so it can delete legitimate inline examples and corrupt adjacent non-frame text.

Keep #7169 open only if the leak reproduces against the current Senpi pin. Reopen this PR only with such a reproduction and a schema-validating parser that preserves legitimate content and tool-call/result pairing.

PR #7193, opened 2026-08-24
fix(omo): gate project-level Claude settings hooks behind project trust

### LINKED ISSUE #6604: Repository `.claude/settings.json` can execute `UserPromptSubmit` commands without project trust
### Prerequisites

## Summary

Oh My OpenAgent's OpenCode compatibility layer automatically loads `.claude/settings.json` from the current project and treats its `UserPromptSubmit` command hooks as executable configuration. After the user has installed and enabled the Ultimate OpenCode plugin, submitting an ordinary prompt in that project reaches `spawn(..., { shell: true })` without a project trust decision, first-use command approval, configuration fingerprint, or content-change confirmation.

This is deterministic configuration execution, not prompt injection: the model does not need to select a tool and the normal OpenCode tool-approval path is not involved. This is also not a claim that merely opening a repository immediately executes code. The minimum trigger is one ordinary prompt submission after the plugin is already installed and the OpenCode working directory points to the repository.

## Affected version

The behavior is present in the current public `dev` revision `19abf74feb0eb1b82f37e70a2dd68a519cf88a08`, inspected on 2026-08-05. The root package version at that revision is `4.19.4`. No patched version is known from this review.

## Impact

An attacker who can commit `.claude/settings.json` to a repository can cause an attacker-selected local command to run with the OpenCode process identity, workspace working directory, and, for a project hook without an environment allowlist, the inherited process environment. The command can modify the repository, launch additional local processes, access files available to the user, or use network-capable local programs.

The user interaction required is installing/enabling the plugin and submitting a normal prompt in the untrusted project. This should be treated as a high-impact local command-execution issue with user interaction, rather than as silent open-folder RCE.

Relevant classifications include CWE-829 (Inclusion of Functionality from Untrusted Control Sphere) and CWE-862 (Missing Authorization).

## Verification

The current source implementations were exercised with the isolated marker harness at `19abf74feb0eb1b82f37e70a2dd68a519cf88a08`. The configured temporary workspace returned `configLoaded: true` and `marker: true`; the no-config control returned `configLoaded: false` and `marker: false`.

The same behavior was also reproduced through a real OpenCode process in an isolated disposable environment: OpenCode `1.18.13`, the published `oh-my-openagent` package `4.19.4`, a project `.claude/settings.json` containing the marker hook above, and one ordinary `opencode run hello` prompt. The command completed with a normal assistant response and created `.workspace-hook-marker` in the project directory. A fresh control directory with no `.claude/settings.json`, a fresh HOME with no global Claude settings, and the same OpenCode/OMO process completed without creating the marker.

For this isolated test, the package was loaded using the installed package's direct `dist/index.js` file URL because OpenCode's package-name cache resolver listed `oh-my-openagent` in `debug info` but did not initialize it in the `run` entry. The direct file entry is the same published `oh-my-openagent@4.19.4` build, not a modified copy. Maintainers should verify the documented `oh-my-openagent` package entry in their supported installer path as well; the resolver discrepancy may be a separate OpenCode compatibility issue.

## Suggested remediation

1. Treat command and HTTP hooks from committed project `.claude/settings.json` as untrusted by default, or disable project executable hooks unless the workspace has been explicitly approved.
2. Before the first dispatch, show the effective event, matcher, command or URL, source file, cwd, and environment keys and require explicit approval.
3. Bind approval to the canonical workspace, user, launcher surface, and a digest of the final effective declaration and referenced targets. Re-prompt when settings, scripts, symlinks, or targets change.
4. Revalidate the final effective command immediately before `spawn`, including after cache refreshes and session reuse.
5. Use a minimal environment allowlist for repository-controlled hooks instead of inheriting the complete OpenCode process environment.

## Expected behavior

A project-provided `UserPromptSubmit` hook should not silently authorize a shell command merely because a user submits a normal prompt. The compatibility layer should either ignore executable project hooks in an untrusted workspace or obtain and persist a narrowly scoped approval for the exact effective command, with renewed approval after its contents change.

## Reproduction

The following Linux/macOS reproduction uses only a marker file in a disposable project. It does not read credentials, make network requests, or perform destructive actions. Bun and a working OpenCode installation are required.

### 1. Create the repository-controlled hook

```bash
WORKSPACE="$(mktemp -d)"
mkdir -p "$WORKSPACE/.claude"

cat > "$WORKSPACE/.claude/settings.json" <<'EOF'
{
  "hooks": {
    "UserPromptSubmit": [
      {
        "matcher": "*",
        "hooks": [
          {
            "type": "command",
            "command": "printf 'oh-my-openagent-hook-marker\\n' > .workspace-hook-marker"
          }
        ]
      }
    ]
  }
}
EOF

cd "$WORKSPACE"
test ! -e .workspace-hook-marker
```

The command is intentionally harmless. It writes only `.workspace-hook-marker` in the disposable workspace.

### 2. Install/enable the OpenCode plugin and trigger the hook

From the disposable workspace, follow the project's documented Ultimate installation flow and complete the TUI until the plugin is enabled:

```bash
bunx oh-my-openagent install
opencode
```

Submit the ordinary prompt `hello` in the OpenCode session. After the prompt is submitted, check the marker:

```bash
cat .workspace-hook-marker
```

Expected output:

```text
oh-my-openagent-hook-marker
```

No model tool call, special prompt, hook UI action, or successful provider response should be required. The hook is dispatched by the plugin's normal `chat.message` lifecycle before the model response.

For a fully isolated package-level check, if the OpenCode version lists `oh-my-openagent` in `opencode debug info` but does not initialize the package-name entry, load the same unmodified published package through its file entry:

```bash
CONFIG_ROOT="$(mktemp -d)"
export HOME="$CONFIG_ROOT/home"
export XDG_CONFIG_HOME="$CONFIG_ROOT/config"
export WORKSPACE
mkdir -p "$XDG_CONFIG_HOME/opencode" "$WORKSPACE/.opencode"
npm install --prefix "$XDG_CONFIG_HOME/opencode" oh-my-openagent@4.19.4
export OMO_PLUGIN_PATH="$XDG_CONFIG_HOME/opencode/node_modules/oh-my-openagent/dist/index.js"
node --input-type=module -e 'import { mkdirSync, writeFileSync } from "node:fs"; import { join } from "node:path"; import { pathToFileURL } from "node:url"; const workspace=process.env.WORKSPACE; mkdirSync(join(workspace, ".opencode"), { recursive: true }); writeFileSync(join(workspace, ".opencode", "opencode.jsonc"), JSON.stringify({ "$schema": "https://opencode.ai/config.json", plugin: [pathToFileURL(process.env.OMO_PLUGIN_PATH).href] }, null, 2))'
cd "$WORKSPACE"
opencode run hello
test -f .workspace-hook-marker && cat .workspace-hook-marker
```

This uses the published package's `dist/index.js` and does not modify it. Use a configured model/provider if the local OpenCode installation does not have a default one.

### 3. Negative control

Repeat the same steps in a fresh disposable workspace with no `.claude/settings.json`:

```bash
CONTROL="$(mktemp -d)"
cd "$CONTROL"
bunx oh-my-openagent install
opencode
```

Submit the same ordinary prompt and then run:

```bash
test ! -e .workspace-hook-marker && echo "control: marker absent"
```

Expected output:

```text
control: marker absent
```

The OpenCode process should be fully restarted between the positive and negative cases so that the result does not depend on a previously loaded project configuration.

### 1. Project settings are automatically discovered and merged

`packages/omo-opencode/src/hooks/claude-code-hooks/config.ts:73-87` constructs settings paths from the user configuration directory and `process.cwd()/.claude/settings.json` plus `.claude/settings.local.json`. `config.ts:232-272` reads each existing file, parses JSON, normalizes the `hooks` object, and merges the declarations. The project path is repository-controlled and no trust or approval decision is performed in this loader.

### 2. The compatibility hook is enabled by default

`packages/omo-opencode/src/plugin/hooks/create-transform-hooks.ts:39-53` creates the `claude-code-hooks` transform when the hook is enabled and passes `pluginConfig.claude_code?.hooks ?? true`. The documented `disabled_hooks` setting is a user policy switch; it is not a provenance check or approval of the effective project command.

### 3. A normal prompt reaches `UserPromptSubmit`

`packages/omo-opencode/src/hooks/claude-code-hooks/handlers/chat-message-handler.ts:40-100` loads the Claude-compatible hook configuration on every `chat.message`, extracts real text parts, and calls `executeUserPromptSubmitHooks()` with the OpenCode directory as `cwd`.

`packages/omo-opencode/src/hooks/claude-code-hooks/user-prompt-submit.ts:49-97` accepts real user text, finds the `UserPromptSubmit` matchers, and dispatches command or HTTP hooks. The `permission_mode` field at `:76-83` is serialized into hook stdin; it does not guard the dispatch or the child process.

### 4. The repository command reaches a shell-backed process sink

`packages/omo-opencode/src/hooks/claude-code-hooks/dispatch-hook.ts:12-38` forwards a command hook to `executeHookCommand()` with the configured command, prompt JSON, and workspace cwd. In the current source, the executor at `packages/utils/src/command-executor/execute-hook-command.ts:81-107` inherits the parent environment when no allowlist is supplied and calls:

```ts
spawn(finalCommand, {
  cwd,
  shell: true,
  detached: !isWin32,
  env,
})
```

There is no project trust check, command display/confirmation, effective-command digest, symlink target validation, or spawn-time revalidation between project settings discovery and this sink.

## Expected behavior

A project-provided `UserPromptSubmit` hook should not silently authorize a shell command merely because a user submits a normal prompt. The compatibility layer should either ignore executable project hooks in an untrusted workspace or obtain and persist a narrowly scoped approval for the exact effective command, with renewed approval after its contents change.

### Actual Behavior

  When oh-my-openagent's OpenCode plugin is enabled and the working directory contains a repository-controlled `.claude/settings.json`, submitting an ordinary prompt causes the configured `UserPromptSubmit` command hook to execute without a project trust decision or first-use confirmation.

  The command reaches `spawn(..., { shell: true })` with the project as its working directory. In my isolated marker test, the configured case created the marker (`configLoaded: true`, `marker: true`), while the no-config control did not (`configLoaded: false`, `marker: false`).

### Doctor Output

```shell
oMoMoMoMo Doctor

   ⚠ 1 issue found:

     Fix: Run: gh auth login
     Affects: GitHub automation
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Linux

### OpenCode Version

 1.18.10

AceRothstein71:
## What

Project-level Claude Code settings (`.claude/settings.json` and `.claude/settings.local.json`) are now loaded only when the current project directory is trusted. A new `resolveProjectTrust()` helper resolves the decision:

1. An explicit Claude Code decision wins: `projects.<absolute-cwd>.hasTrustDialogAccepted` from `$CLAUDE_CONFIG_DIR/.claude.json` (falling back to `~/.claude.json`).
2. If no decision is recorded, `OMO_CLAUDE_SETTINGS_TRUST=1` (or `true`) opts the directory in.
3. Otherwise the directory is untrusted and project-scope hooks are skipped (logged). An explicit deny cannot be lifted by the env var.

User-level settings (`<claude-config-dir>/settings.json`) and explicit custom settings paths are unaffected. The gate lives in `loadClaudeHooksConfig()`, the single choke point every CC hook event handler (UserPromptSubmit, PreToolUse, Stop, ...) reads its config from.

## Why

Issue #6604: the plugin merged hooks from cwd-derived `.claude/settings.json` unconditionally and dispatched matching commands via `spawn(..., { shell: true })`. The only existing gate was `isHookCommandDisabled()` (an opt-out allowlist), not an opt-in trust gate, so a repository committing a `UserPromptSubmit` hook could execute arbitrary local commands when a user submits a prompt in that project. Upstream Claude Code only loads project-level settings after a persisted per-directory trust decision; this change aligns with that semantics.

## Verified

Failing-test-first (given/when/then):

- New `project-trust.test.ts` (7 tests): default-deny with no recorded decision, accepted/rejected Claude decisions, env opt-in, explicit deny beats env opt-in, malformed `.claude.json` falls back to default-deny (isolated temp `CLAUDE_CONFIG_DIR`).
- Extended `config.test.ts` with a trust-gating suite (5 tests): untrusted project hooks excluded, trusted project hooks included, deny survives env opt-in, unknown project + env opt-in loads, user-level settings still load for an untrusted project.
- Red evidence: with base `config.ts` and no `project-trust.ts`, the new tests fail (13 pass / 4 fail / 1 error). After the fix: 22 pass / 0 fail for the two files; full scoped suite `bun test packages/omo-opencode/src/hooks/claude-code-hooks/` is 127 pass / 0 fail (baseline on clean HEAD was 116 pass across 16 files, so all new tests are additive).
- `bun run typecheck` (tsgo root + script + all packages): exit 0, no errors.
- Evidence recorded under `.omo/evidence/20260824-6604-claude-settings-trust/` (red/green runs, scoped suite before/after, typecheck output).

## Risk

Users who maintain project `.claude/settings.json` hooks in directories Claude Code never marked trusted will stop seeing those hooks fire until they either open the project in Claude Code once (trust dialog) or set `OMO_CLAUDE_SETTINGS_TRUST=1`. Behavior for Claude-trusted projects, user-level settings, and custom settings paths is unchanged.

Fixes #6604

---
## Summary by cubic
Project-level Claude Code settings are now loaded only for trusted project directories to prevent untrusted repos from executing hooks. Previously, `.claude/settings.json` and `.claude/settings.local.json` were merged unconditionally; now they load only when a per-directory trust decision allows it. User-level settings and custom settings paths are unchanged.

- A new `resolveProjectTrust()` gate in `omo-opencode` decides trust:
  - Claude Code’s recorded decision wins (`projects.<abs-path>.hasTrustDialogAccepted` in `.claude.json` under `$CLAUDE_CONFIG_DIR` or `~`).
  - If no decision exists, `OMO_CLAUDE_SETTINGS_TRUST=1|true` opts the directory in.
  - Explicit deny cannot be overridden by the env var; untrusted projects skip project-scope hooks and log the skip.
- `loadClaudeHooksConfig()` now merges settings in order: custom path (if present) -> user-level -> project-level (only if trusted). The public `getClaudeSettingsPaths()` still returns the same paths.

**Rollout**
- Required: if project hooks stop firing, trust the directory by opening it in Claude Code and accepting the trust dialog, or set `OMO_CLAUDE_SETTINGS_TRUST=1` for directories without a recorded decision.

<sup>Written for commit 0a46670e3ce422d0001060816ad34a9b10f2243c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for addressing the project hook trust boundary. The confirmed blocker is that path deduplication was removed, so when `CLAUDE_CONFIG_DIR` resolves to `<cwd>/.claude`, the same settings file is loaded twice and its hook command executes twice. Please restore deduplication while preserving precedence, then add and run a regression test proving the overlapping paths produce exactly one command on the supported platforms and provide the passing test and typecheck results. https://github.com/code-yeongyu/oh-my-openagent/pull/7193/files

PR #7197, opened 2026-08-24
fix(skills-loader-core): defer shared skill template reads past module evaluation

### LINKED ISSUE #6635: [Bug]: v4.19.4 regression — no longer loads on OpenCode Desktop (Node sidecar); v3.17.15 shim fix reverted (bun.serve top-level)
### Summary

Regression in v4.x: the OpenCode Desktop (Electron/Node sidecar) load fix shipped in v3.17.15 (see closed #3797 / PR #3798 — `bun-spawn-shim.ts`, `grep -c 'globalThis.Bun'` → 0) is **gone in v4.19.4** (latest on npm). The plugin silently fails to load on Desktop again.

### Environment

- OS: Windows 11 x64
- OpenCode Desktop: v1.18.15 (latest production)
- oh-my-openagent: v4.19.4 (latest, via `"plugin": ["oh-my-openagent"]` in `%USERPROFILE%\.config\opencode\opencode.jsonc`)
- Bun: 1.3.14 installed and on PATH (so it's not a missing-Bun issue)

### Evidence

- **Zero new entries** in the plugin's own log (`%TEMP%\oh-my-openagent.log`, `LOG_FILENAME2` at `dist/index.js:85663`) after the Desktop sidecar starts — the plugin is never instantiated. No error in Desktop `server.log` either (silent drop).
- Same config under the CLI works fine (agents sisyphus/oracle/etc. register correctly), so the plugin itself runs — the failure is Desktop runtime specific.
- `dist/index.js` (5.7MB) still contains Bun-only APIs at top level, e.g. `bun.serve` at line 151702; the only guard is `typeof globalThis.Bun === "undefined"` around line 169430, which is insufficient because the sidecar process is Node, so `bun.serve` can't work.
- Upstream: OpenCode's own opt-in V2 desktop sidecar (`OPENCODE_SIDECAR_V2=1`) is **development-build-only** — not bundled in production builds (anomalyco/opencode#40980, and my report anomalyco/opencode#41033). So Desktop users have no Bun runtime today.

### Request

Please ship a **Node-compatible build** (e.g. `bun build --target node`, or a dual build selected via `exports` conditions), or re-apply the v3.17.15 shim approach so v4.x loads on the Electron sidecar. The v3.17.15 fix proves this is achievable — it just regressed in 4.x.

### Related

- #3797 (closed as fixed in v3.17.15 — regression)
- anomalyco/opencode#25880, anomalyco/opencode#41033

AceRothstein71:
## What

Builtin skill modules (`debugging`, `frontend`, `init-deep`, `remove-ai-slops`, `review-work`, `visual-qa`) resolved their shared `SKILL.md` template via a module-scope call:

```ts
template: loadSharedSkillTemplate("frontend"),
```

This changes each to a lazy accessor so module evaluation performs no filesystem I/O:

```ts
get template() {
  return loadSharedSkillTemplate("frontend")
},
```

A new source-audit test (`skill-template-lazy-evaluation.test.ts`) fails the suite if any `skills/*.ts` module reintroduces an eager module-scope `loadSharedSkillTemplate(` call.

## Why

Issue #6635 reports the plugin silently failing to load on OpenCode Desktop, whose plugin host loads packages through Node ESM dynamic `import()`. The thread's original `bun.serve` theory was rebutted (that call is guarded behind a local `runtimeEnv.Bun` check with a `node:http` fallback), but the failure class it converged on is real and reproducible on `dev`: these six module-scope calls run `readFileSync(sharedSkillsRootPath()/<name>/SKILL.md)` during bundle evaluation. `sharedSkillsRootPath()` probes three candidate directories relative to the bundle and falls back to a path that may not exist. Whenever the skills tree is absent or unresolvable next to `dist/index.js` (Desktop sidecar layouts, partial installs, checkouts whose build stopped before `build:shared-skills-assets`), that ENOENT rejects the entire ESM import, the host drops the plugin before any plugin code runs, and nothing reaches the plugin log: exactly the reported "silent drop, zero log entries" symptom.

Reproduced locally by importing `dist/index.js` under `node --input-type=module -e "await import('./dist/index.js')"` with no `dist/skills` tree present: import failed with `ENOENT ... dist/skills/frontend/SKILL.md` thrown from `loadSharedSkillTemplate` at module evaluation.

## Verified

- RED (before fix): both Node ESM smoke tests in `packages/omo-opencode/src/shared/dist-bundle-bun-globals.test.ts` fail (3 pass / 2 fail) with the ENOENT above; the new audit pin lists all six eager modules.
- GREEN (after fix): same audit suite 5 pass / 0 fail \u2014 including both Node smokes \u2014 with `dist/skills` still absent from the tree (stronger than CI, where the build populates it).
- `bun test packages/skills-loader-core`: 232 pass / 0 fail.
- `bun test packages/omo-opencode/src/features/builtin-skills/`: 37 pass / 0 fail (template values byte-identical to the SKILL.md bodies).
- `bun run typecheck` (tsgo root + script + all workspace packages): exit 0.
- Evidence: `.omo/evidence/20260824-6635-desktop-bunserve/ (renamed from 20260824-6635-bun-serve-regression to match the task evidence slug; includes a crash-recovery re-verification record)` (WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED plus red and green logs).

Behavior notes: loader fail-fast semantics are unchanged (ENOENT still throws at first actual template use instead of killing bundle import), per-name caching via the module-singleton loader is unchanged, and consumers read `.template` only at skill resolution/render time.

## Risk

Low. The change only moves six filesystem reads from module evaluation to first property access of an already-lazily-consumed field. Residual risk: environments where the skills tree is missing will now surface the ENOENT at skill-render time rather than import time; rendering surfaces already handle hook/tool errors, and the alternative (import-time crash) is strictly worse for hosts like Desktop.

Fixes #6635

---
## Summary by cubic
Defer shared `SKILL.md` template reads for builtin skills in `skills-loader-core` to first access to prevent Node ESM import failures when the skills tree is absent (fixes #6635). Previously, six skills read templates at module evaluation and could throw ENOENT; now each exposes a lazy get template() so errors surface at first use and caching is unchanged.

- Replace eager template reads with lazy getters in debugging, frontend, init-deep, remove-ai-slops, review-work, and visual-qa under `skills-loader-core`.
- Add a source-audit test to block module-scope `loadSharedSkillTemplate(` reintroduction.
- Verified: Node ESM smoke tests pass with no `dist/skills`; full `skills-loader-core` and builtin-skills suites green; typecheck passes.
- Evidence directory renamed to `.omo/evidence/20260824-6635-desktop-bunserve`; no code changes.
- No migration required.

<sup>Written for commit 21b8426dd77b98dda634b44972a47ebfe73ef0d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7197?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for the PR, @AceRothstein71. The lazy-template hardening is correct and independently reproducible.

I am closing this as superseded rather than rejected. #7262 carries the byte-identical source, tests, and evidence on an in-repository branch with the complete test, typecheck, build, and compatibility matrix green. The replacement preserves this implementation while attaching merge-grade CI evidence.

Please also remove `Fixes #6635` from the replacement: that issue was closed after its reporter traced the failure to an interrupted installation, not missing published skill files. The change remains worthwhile defensive hardening on its own.

Issue #7012, opened 2026-08-18
[Bug]: Memory reflection spawn fails with ENOENT when runtime/reflection-sessions dir does not exist


wusai2333:
## Summary

Memory reflection runs fail at spawn time because the harness `lstat`s `runtime/reflection-sessions` before the directory is ever created. Every reflection run fails until the directory is created manually.

## Observed

Two consecutive failures (`consecutiveFailures: 2`) on a fresh agent identity, macOS arm64, omo-ai `5.0.0-0.beta.10`, @code-yeongyu/senpi `2026.8.18-2`, Node v24.19.0.

Completion record (`~/.omo/memory/agents/<id>/runtime/reflection/completions/reflection-run-2.json`):

```json
{
  "runId": "reflection-run-2",
  "category": "quick",
  "trigger": "step-count",
  "outcome": "failed",
  "reason": "spawn_failed",
  "detail": "ENOENT: no such file or directory, lstat '/Users/<user>/.omo/memory/agents/<id>/runtime/reflection-sessions'",
  "durationMs": 6108,
  "consecutiveFailures": 2
}
```

Surface message in the session:

```
✗ Memory reflection failed · reflection-run-2
Reflection did not finish; the transcript cursor was not advanced.
took 6.1s · reason spawn_failed · ENOENT: no such file or directory, lstat '/Users/<user>/.omo/memory/ag...'
```

## Expected

The reflection spawner should `mkdir -p` (or `fs.mkdirSync(..., { recursive: true })`) the `runtime/reflection-sessions` directory before statting/using it, the same way sibling dirs (`locks`, `transcripts`, `facts-queue`, `notices`, `worktrees`, `reflection`) get materialized.

## Workaround

```bash
mkdir -p ~/.omo/memory/agents/<agent-id>/runtime/reflection-sessions
```

After creating the directory the next reflection trigger should proceed. No data loss occurs either way — the transcript cursor simply does not advance and the same step range is retried.

## Environment

- omo-ai 5.0.0-0.beta.10, @code-yeongyu/senpi 2026.8.18-2
- macOS (Darwin 25.6.0) arm64, Node v24.19.0
- Fresh agent identity (memory repo had only the `chore: initialize local memory` commit)

code-yeongyu:
Confirmed against source — good report, and the mechanism is exactly what your ENOENT suggests.

[...]

Issue #6873, opened 2026-08-15
[Bug]: Memory reflection auto sandbox repeatedly fails when bwrap cannot create a user namespace


vntrevx:
### Prerequisites

- [x] I will write this issue in English.
- [x] I searched existing issues and PRs for `bwrap`, `uid map`, `Memory reflection has failed`, and `reflection sandbox`.
- [x] I am using the latest `omo-ai` beta release.
- [x] I read the memory reflection and sandbox configuration documentation.

### Bug Description

On OMO Native, memory reflection with the default `memory.reflection.sandbox: "auto"` treats an executable `/usr/bin/bwrap` as a usable sandbox. On this Ubuntu system, bubblewrap is installed but AppArmor prevents its unprivileged user namespace setup. Every reflection child therefore exits with:

```text
bwrap: setting up uid map: Permission denied
```

`auto` does not degrade to an unsandboxed child after this runtime setup failure. The memory failure counter increases on every trigger and the warning is repeatedly delivered to sessions sharing the same memory identity.

The warning also says to inspect `runtime/reflection-sessions/<runId>/child-stderr.log`, but the failed run directory has already been removed, so the referenced file does not exist.

### Steps to Reproduce

1. Use Ubuntu with bubblewrap installed and AppArmor restricting unprivileged user namespaces.
2. Leave memory reflection sandbox at its default `auto` value.
3. Start OMO Native and allow a step-count reflection trigger, or run `/reflect --recent 1`.
4. Observe repeated memory reflection failures.
5. Confirm the sandbox itself fails outside OMO:

```console
$ bwrap --ro-bind / / --proc /proc --dev /dev true
bwrap: setting up uid map: Permission denied
```

6. Set the following and start a fresh OMO process:

```jsonc
{
  "memory": {
    "reflection": {
      "sandbox": "off"
    }
  }
}
```

7. `/sleeptime` now reports `Sandbox: off`, demonstrating that avoiding bwrap is a working workaround.

### Expected Behavior

For `sandbox: "auto"`:

- OMO should verify that the selected sandbox can actually start, not only that the executable exists; or
- if bwrap exits during sandbox setup, reflection should retry unsandboxed with a clear degradation warning.

`sandbox: "required"` should remain fail-closed.

Diagnostics should either preserve `child-stderr.log` or avoid pointing users to a path that cleanup has already removed.

### Actual Behavior

Reflection fails on every trigger and the consecutive failure count keeps increasing:

```text
[...]

PR #7201, opened 2026-08-24
fix(memory): stop reflection bare-runId collisions after a run-counter reset

### LINKED ISSUE #7095: Memory reflection permanently stuck: bare-runId completion records collide after run-counter reset; reflection lane lacks the facts-lane sandbox fallback
## Environment

- `omo-ai` 5.0.0-0.beta.12 (npm global, node v24.19.0)
- Ubuntu (kernel 7.0.0-28-generic), `kernel.apparmor_restrict_unprivileged_userns = 1` (24.04+ default hardening)
- Memory identity: project-scoped agent, local-only repo (no remote)

## Impact

The memory **reflection lane silently produced zero commits, ever** — 7 consecutive failed runs over two days with no user-visible warning (`consecutiveFailures: 7` accumulated quietly in completion records). The facts-extraction lane kept working via its `memory facts sandbox degraded` fallback, which masked the loss. Two independent defects stacked:

## Defect 1 — reflection child has no sandbox-degraded fallback

On hosts where bwrap cannot create user namespaces (Ubuntu 24.04+ AppArmor restriction, no profile for `/usr/bin/bwrap`), every reflection child dies instantly:

```
bwrap: setting up uid map: Permission denied
```

- runs 2–7: `outcome: "failed", reason: "child_exit"` with the bwrap error, ~200ms each
- The **facts lane detects this and degrades** (`memory facts sandbox degraded` in omo-debug.log) and kept committing facts normally.
- The **reflection lane has no equivalent fallback** and hard-failed every trigger (step-count, compaction, dream) forever.

Host-side workaround that fixed it: an AppArmor profile granting `userns,` to `/usr/bin/bwrap` (same pattern Ubuntu ships for chrome/brave/1password). Verified: `bwrap --ro-bind / / /bin/true` exits 0 afterward.

**Suggestion:** give the reflection child the same degraded-sandbox fallback as the facts lane, or at minimum surface a persistent warning to the session after N consecutive reflection failures.

## Defect 2 (structural) — completion records keyed by bare runId collide across run-counter resets

Completion records are written to `runtime/reflection/completions/reflection-run-<N>.json` — keyed by **bare runId** — while run dirs and worktrees are epoch-prefixed (`worktrees/<epoch>-reflection-run-<N>`). The publish path is idempotent-only-if-identical: an existing record with different content (ignoring `delivery`) throws:

```
Reflection completion record mismatch for reflection-run-1
```

When the run counter resets (in our case `runtime/reflection-sessions/` was recreated empty, so numbering restarted at 1 while day-old `reflection-run-1.json`…`reflection-run-7.json` records persisted), the new generation permanently collides:

- every launch fails: `memory reflection launch failed { error: 'Reflection completion record mismatch for reflection-run-1' }`
- bind-time reconcile fails the same way: `memory bind-time reconcile failed`
- finalization can never publish: `Reflection finalization did not publish completion for reflection-run-2`
- cleanup never runs (finalization throws first), so **orphan worktrees + branches accumulate** (we had 2: `memory/reflection-<epoch>-reflection-run-{1,2}`, both heads ancestors of main)
- run dirs get **reused** by the new generation (`runs/reflection-run-1/` held a mix of old and new state), making the mismatch permanent

This is unrecoverable without manual intervention: the lane retries and fails in a loop (observed 10:07 → 13:06 repeated attempts).

**Manual recovery that worked:**
1. Archive all `completions/reflection-run-*.json` (all were `delivery.status: "consumed"` failures) and all `runs/reflection-run-*` dirs out of the way
2. `git worktree remove --force` both orphan worktrees, `git branch -D` both branches (verified heads were ancestors of main → no data loss), `git worktree prune`
3. Reflection then relaunches cleanly

**Suggestions:**
1. Key completion records with the same epoch prefix as worktrees/branches (or embed a generation id), so a counter reset cannot collide with a previous generation.
2. Make the mismatch non-fatal: quarantine the stale record (rename aside) and continue, instead of aborting the entire reflection lane at launch/reconcile.
3. Have finalization attempt worktree/branch cleanup even when publish fails, so orphans don't accumulate.

## Notes

Happy to provide the archived state (completion JSONs, run ledgers, debug-log excerpts) if useful.

AceRothstein71:
## What

Reflection run ids are no longer a per-process counter. A new `createReflectionRunIdFactory` mints each id as one past the highest id still on disk - scanning durable completion records (`runtime/reflection/completions/`), run directories, and the live reservation state files - and `ReflectionReservationStore` now mints the id inside the scheduler lock through an async factory, so two processes racing to reserve can never observe the same disk state and derive the same id.

## Why

#7095 (Defect 2): the per-process counter restarted at 1 on every launch. When the persisted reflection-sessions state was recreated (#7012), the next generation re-minted retired ids such as `reflection-run-1` while their consumed completion records were still on disk. Every subsequent launch and bind-time reconcile then aborted with `Reflection completion record mismatch for reflection-run-1`, finalization reported missing completions, and orphan memory worktrees/branches accumulated - permanently wedging the lane until manual cleanup of the lock, completion records, AND the stale run directory.

## Verified

- Failing-first: a scratch test wiring the pre-fix counter factory fails with the issue's literal error (`Reflection completion record mismatch for reflection-run-1`, thrown at `worker/completion-records.ts:15`); the scratch file was removed after capture.
- New regression coverage (given/when/then, co-located):
  - `reflection-run-id-collision.test.ts` - two generations reserving across a simulated restart never re-mint a persisted id; a new completion lands beside the stale record, which stays byte-identical.
  - `reflection-run-id.test.ts` - high-water semantics over completions/run dirs/live reservation state; non-run names ignored; strictly increasing within one process.
  - `reservation.test.ts` - an async run-id factory is awaited under the scheduler lock.
- `bun test packages/memory-core/src`: 545 pass / 0 fail.
- `bun test packages/omo-senpi/src/components/memory`: 907 pass, 6 skip / 0 fail.
- Typecheck clean: `packages/memory-core`, `packages/omo-senpi`.
- Evidence: `.omo/evidence/20260824-7095-runid-collision/QA.md`.

## Risk

Low. Ids only move forward and existing consumers treat run ids as opaque strings. Fresh installs still start at `reflection-run-1`. Cross-process uniqueness rides the existing `reflection-scheduler` lock, which now also covers minting.

Fixes #7095

---
## Summary by cubic
Stops reflection run-id collisions after restarts by replacing the per-process counter with a disk-scoped minting strategy. Previously the counter restarted at 1 and could re-mint retired ids, colliding with durable completion records and aborting launch/reconcile/finalization; now ids are minted one past the highest id on disk (completions, run directories, live reservation state) and minted inside the scheduler lock to prevent races.

- `packages/omo-senpi/src/components/memory/reflection-run-id.ts` adds `createReflectionRunIdFactory` (disk-scoped high-water minting, fresh installs start at `reflection-run-1`, strictly increasing within a process).
- `packages/memory-core/src/reflection/reservation.ts` mints the id under the scheduler lock and accepts an async `createRunId` (string or Promise<string>).
- Tests: new collision and factory suites in `packages/omo-senpi/src/components/memory`, plus a lock-await case in `packages/memory-core/src/reflection/reservation.test.ts`. QA evidence in `.omo/evidence/20260824-7095-runid-collision/QA.md`. Fixes #7095.

<sup>Written for commit 216bbd24346f48ba1739fc0fd6ea681d9103eb22. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260824-7095-runid-collision/QA.md
+++ b/.omo/evidence/20260824-7095-runid-collision/QA.md
@@ -1,0 +1,70 @@
+# QA Evidence - Issue #7095: reflection bare-runId completion-record collision
+
+Date: 2026-08-24
+Branch: `issue/7095-reflection-runid-collision` (worktree `/home/viprix/projects/oom-wt-7095`, base dev @8833800ae)
+
+## WHAT WAS TESTED
+
+Surface: reflection reservation/run-id minting (`packages/memory-core/src/reflection/reservation.ts`,
+`packages/omo-senpi/src/components/memory/reflection-run-id.ts`) and its regression coverage.
+
+Commands (bun 1.3.14, hermetic tmpdir fixtures only; no real `~/.senpi/agent`, no network):
+
+1. Failing-first reproduction (scratch test, NOT committed): wired the PRE-FIX per-process counter
+   factory (`reflection-run-${++counter}` restarting at 1 per store instance = simulated restart)
+   against seeded generation-one debris (`completions/reflection-run-1.json`) and two reserve
+   generations across a simulated restart.
+   `bun test packages/omo-senpi/src/components/memory/scratch-legacy-counter-repro.test.ts`
+2. Fixed behavior, targeted:
+   `bun test packages/memory-core/src/reflection/reservation.test.ts`
+   `bun test packages/omo-senpi/src/components/memory/reflection-run-id.test.ts packages/omo-senpi/src/components/memory/reflection-run-id-collision.test.ts`
+3. Scoped suites:
+   `bun test packages/memory-core/src`
+   `bun test packages/omo-senpi/src/components/memory`
+4. Typechecks:
+   `bun run --cwd packages/memory-core typecheck`
+   `bun run --cwd packages/omo-senpi typecheck`
+
+## WHAT WAS OBSERVED
+
+1. Failing-first: scratch repro FAILED with exactly the error reported in #7095 -
+   `Reflection completion record mismatch for reflection-run-1` thrown at
+   `worker/completion-records.ts:15` when the second generation re-minted `reflection-run-1`
+   while generation one's consumed completion record was still on disk. Scratch file deleted
+   after capture.
+2. Targeted after fix: reservation suite 10 pass / 0 fail (includes new async run-id factory
+   case minting under the scheduler lock); run-id factory + collision regression suites
+   7 pass / 0 fail (fresh ids continue above every persisted id across a simulated restart;
+   stale record untouched beside the new one).
+3. Scoped suites: memory-core 545 pass / 0 fail (68 files); omo-senpi memory component
+   907 pass, 6 skip / 0 fail (135 files).
+4. Typechecks: both packages clean.
+
+Isolation proof: all fixtures live under `mkdtemp(tmpdir())`; no command touched the real senpi
+agent dir or any harness state. Worktree-only verification; branch never touched `dev`.
+
+## WHY IT IS ENOUGH
+
+The committed regression test (`reflection-run-id-collision.test.ts`) pins the exact reported
+wedge end-to-end through the real `ReflectionReservationStore`: stale completion record +
+run directory from an earlier generation, then two fresh stores reserving across a restart must
+never re-mint a persisted id, and publishing a new completion must leave the stale record
+byte-identical. The scratch run proves this test fails against the pre-fix counter behavior with
+the issue's literal error string, so it guards the defect, not just the implementation. The
+factory unit tests pin disk-scoped high-water semantics (completions, run dirs, live
+reservation state, non-run names ignored, strictly increasing within a process). Remaining risk:
+ids minted by processes running concurrently on hosts where the scheduler lock is bypassed are
+out of scope (the lock is the existing cross-process seam; minting now happens inside it).
+
+## WHAT WAS OMITTED
+
+- Raw bun test output beyond the counts above (no secrets present, trimmed for size).
+- The scratch reproduction file itself (deleted from the worktree after capture; its content is
+  described above).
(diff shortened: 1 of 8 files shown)


code-yeongyu:
Closing in favor of #7333, which just merged into `dev` — and as with your sibling PR #7216, the credit belongs here: **the diagnosis and the design in the merged fix are yours.**

The disk-scoped high-water mark, the dedicated `reflection-run-id.ts` module, the async `createRunId` awaited inside the scheduler lock, and the call to leave stale completion records untouched rather than migrate them — #7333 ships exactly that shape. It extends the scan to `reflection-sessions/` and epoch-prefixed worktree names and parses reservation state structurally instead of matching text, but those are refinements on your analysis, not corrections. Your write-up of the run-counter reset wedge (`Reflection completion record mismatch for reflection-run-1`) was one of the clearest stateful-bug reports this repo has received, and the two-defect decomposition made both fixes land cleanly (#6873's degrade fix covers Defect 1 — also merged today via #7332).

We landed maintainer-authored equivalents mainly because of the tracked extension-bundle CI gate, which external branches can't easily regenerate. That's on our tooling, not on your work.

Thank you @AceRothstein71 — two sharp PRs in one day, both of which shaped what shipped. The fixes go out in the next release. You'd be very welcome back anytime. 🙇

PR #7210, opened 2026-08-24
fix(omo-senpi): shrink TTSR-truncated bubbles and dedupe the nudge retry

### LINKED ISSUE #7135: [Bug]: TTSR collapse-repetition abort leaves the partial reply on screen, then the nudge rewrites the same answer
### Prerequisites

### Bug Description

Senpi edition (`omo` / `omo-ai@beta`): when TTSR `collapse-repetition` aborts a looping stream, the cut-off assistant reply stays in the TUI, then the nudge makes the model rewrite the same answer. One turn looks like two.

The detector is fine. The bug is the resume UX.

This is easy to misread as fixed. Default `repeatMode` is `"once"`: after the first nudge in a session, the same rule does not fire again. Later short turns in that session look clean even though nothing changed.

Same Grok model on Claude Code CLI via OpenCodex did **not** show this cut-and-repeat UX.

### Steps to Reproduce

1. Start a **new** `omo` session. Do not reuse a session that already hit TTSR.
2. Use `xai/grok-4.6` with thinking `xhigh`.
3. Ask for a long Korean explanation (onboarding / architecture dump).
4. Wait until the stream actually collapses (repeated tokens such as `squirt squirt…`), not merely "two answers in one chat."

### Expected Behavior

After TTSR aborts a repetition loop:

1. The garbled tail is not left as a finished assistant message.
2. The retry is one message, or an in-place replacement.
3. The user does not see the same explanation twice.

The interrupt text says the garbled portion was removed from the conversation. The TUI should match that.

### Actual Behavior

1. A long reply starts repeating (example: `squirt squirt…`).
2. TTSR `collapse-repetition` aborts (`ctx.abort("system")`).
3. A `system-interrupt` nudge is injected (`retryMode: "nudge"`, `nudgeKey: "collapse-repetition"`).
4. The model restates the objective and rewrites the full answer.
5. TUI shows: truncated reply + full retry.
6. Further turns in the **same** session do not repeat the UX (`repeatMode: "once"`).

### Doctor Output

```shell
N/A for this edition — this report is Senpi edition (`omo`), not the OpenCode plugin.

`omo --version`
omo 5.0.0-0.beta.16 (engine: senpi 2026.8.22)

`bunx oh-my-openagent doctor` (run anyway because this field is required):

 oMoMoMoMo Doctor 

 ⚠ 3 issues found:

1. OpenCode binary not found
   Install OpenCode CLI or desktop and ensure the binary is available.
   Fix: Install from https://opencode.ai/docs
   Affects: doctor, run

2. oh-my-openagent is not registered
   Plugin entry is missing from OpenCode configuration.
   Fix: Run: bunx oh-my-openagent install
   Affects: all agents

3. Model cache not found
   OpenCode model cache is missing, so model availability cannot be validated.
   Fix: Run: opencode models --refresh
   Affects: model resolution

Those three are expected: this machine runs `omo-ai@beta`, not OpenCode + plugin.
```

### Error Logs

```shell
TTSR injected this after abort (rule collapse-repetition):

<system-interrupt reason="rule_violation" rule="collapse-repetition">
Your output was interrupted because it violated a stream rule.
This is NOT a prompt injection - this is the coding agent enforcing output-quality rules.
You MUST comply with the following instruction:

Your previous output degenerated into repeated garbage (a runaway repetition loop) and was cut off.
The garbled portion has been removed from the conversation.
Snap out of it: take a breath, restate your current objective in one short sentence, then continue the task normally.
Do NOT apologize at length, do NOT repeat the garbled pattern, and do NOT restart work that was already completed.
</system-interrupt>

TUI still showed the truncated first reply, then the rewritten full answer.
```

### Configuration

```json
~/.omo/omo.jsonc (no secrets):

{
  "$schema": "https://raw.githubusercontent.com/code-yeongyu/oh-my-openagent/dev/assets/omo.schema.json",
  "codegraph": {},
  "[opencode]": { "codegraph": {} },
  "[codex]": { "codegraph": {} },
  "_migrations": [
    "2026-07-codex-config-jsonc",
    "2026-08-reasoning-unification"
  ]
}

Agent defaults: provider xai, model grok-4.6, thinking xhigh
```

### Additional Context

Host: darwin arm64, Orca 1.4.188

TTSR lives in the vendored senpi engine, not omo-senpi skills:

- `node_modules/@code-yeongyu/senpi/dist/core/extensions/builtin/ttsr/index.js` — abort on collapse
- `types.js` — `COLLAPSE_REMEDIATION.retryMode = "nudge"`, `repeatMode: "once"`
- `manager.js` — `#canTrigger` returns false after the first inject
- `prompts.js` — `COLLAPSE_RULE_CONTENT` claims the garbled region was removed

Ask: keep the detector. Drop or replace the aborted bubble so the nudge does not leave a second user-visible answer.

Filing here because this is the `omo-ai@beta` product tracker. The implementation is senpi.

### Operating System

macOS

### OpenCode Version

N/A — Senpi edition omo 5.0.0-0.beta.16 (engine senpi 2026.8.22), not OpenCode

AceRothstein71:
## What

Add a `collapse-recovery` component to the Senpi adapter that repairs the post-abort UX of the engine's builtin TTSR `collapse-repetition` rule:

- **`message_end`**: detect the engine's truncated-abort shape (assistant message, `stopReason: "aborted"`, trailing `[output interrupted by stream rule]` marker block, no `toolCall` blocks) and return a replacement that shrinks the persisted bubble to the marker block alone. The builtin TTSR extension runs first (builtins load ahead of plugin extensions), so this component sees the already-truncated message and only drops the pre-garbage body from display.
- **`agent_settled`**: send exactly one hidden (`display: false`) `omo-collapse-recovery:context` custom message carrying a head/tail excerpt of what the interrupted reply had already delivered, with an explicit do-not-restate instruction. It is sent WITHOUT `triggerTurn`, so the builtin nudge keeps owning the retry turn; because settle-deferred sends flush in registration order and plain sends push directly into agent state, the excerpt reaches the retry turn's first provider request without burning an extra assistant turn.
- Mirrors the engine's default `repeatMode: "once"` (one injection per session) and disarms on user aborts (`agent_end` with `abortSource: "user"`), `session_abort`, and fresh user input so the excerpt never leaks into an unrelated later turn.

Also ships a live QA lane: `collapse-recovery-e2e.mjs` + a dedicated mock provider that streams a genuine repetition loop so the REAL builtin detector latches, then asserts the persisted session JSONL and the exact provider-visible retry request.

## Why

Issue #7135: when TTSR aborts a looping stream, the cut-off reply stays visible as a finished bubble even though the injected interrupt claims "The garbled portion has been removed from the conversation", and the nudge makes the model rewrite the same answer - one turn looks like two. The detector is fine; the resume UX is the bug. This change makes the TUI match the interrupt text (garbled/partial body removed, marker-only bubble) and gives the retry model what it needs to continue or compress instead of restating.

## Verified

- Failing-first: with the component-list registration reverted, the new wiring regression test fails 0 pass / 2 fail; registered, it passes (evidence: `.omo/evidence/20260824-7135-ttsr-abort-reply/failing-first-wiring.txt`).
- Unit + wiring: 26 collapse-recovery tests pass (detection shapes, replacement content, once-per-session injection, disarm paths, disabled flag, production-list wiring).
- Senpi gate: `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` clean; `bun run test:senpi` 2252 pass / 0 fail.
- Live harness e2e (real `senpi` binary, isolated sandbox agent dir, mock provider via `senpi -e`): PASS - persisted aborted bubble is marker-only, exactly one TTSR nudge, exactly one hidden recovery context, retry request carried the dedup context plus head/tail excerpts, garbled body absent from the retry, real `~/.senpi/agent` untouched (evidence: `.omo/evidence/20260824-7135-ttsr-abort-reply/e2e-final.json`).
- Rebuilt committed extension bundles via `plugin/scripts/build-extension.mjs`.

## Risk

Low-medium. Detection is structural and pinned to the engine's literal marker string and stop reason; if a future senpi release changes either, the component degrades to no-op rather than misfiring (negative tests cover completed replies, marker-only replies, and toolCall-bearing messages). Injection timing depends on the engine's settle-delivery ordering, which was verified against the installed engine dist and exercised end-to-end by the live driver.

Fixes #7135

---
## Summary by cubic
Repairs TTSR abort UX in `omo-senpi`: shrinks truncated assistant bubbles to the engine marker and injects a single hidden dedup context so the nudge retry continues instead of restating. Previously, aborted replies persisted the partial body and the built-in nudge rewrote it; now the bubble is marker-only and the retry sees an excerpt with do-not-restate instructions.

- Detects the engine’s truncated-abort shape (assistant, stopReason "aborted", trailing "[output interrupted by stream rule]" text block, no toolCall blocks) and returns a replacement that keeps only the marker.
- Sends exactly one hidden `omo-collapse-recovery:context` message on `agent_settled` carrying head/tail excerpts; no `triggerTurn`, so the built-in nudge owns the retry turn and the excerpt reaches the first provider request.
- Mirrors the engine’s repeatMode "once"; disarms on user aborts, session aborts, and fresh input; never fires for toolCall-bearing replies.
- Registers the new `collapse-recovery` component in the production list before tool components; wiring and behavior are covered by unit/wiring tests and a live e2e harness (`collapse-recovery-e2e.mjs` + mock provider) that proves marker-only persistence and non-repeating retries.
- Rollout: enabled by default; to disable set flag `omo-senpi-collapse-recovery-disabled`. Risk is low; if the engine marker or stop reason changes, detection degrades to no-op. Fixes #7135.

<sup>Written for commit df070c57ca71d276fc08e5092601240a4d24130c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
Thanks for the real-harness proof. One correctness boundary still blocks this implementation.

`extractInterruptedExcerpt` currently collects `thinking` blocks as delivered content, and `agent_settled` sends that excerpt back as hidden user-role context. Those blocks were not user-visible and must not be reclassified or disclosed as something the user already saw:

- https://github.com/code-yeongyu/oh-my-openagent/blob/df070c57ca71d276fc08e5092601240a4d24130c/packages/omo-senpi/src/components/collapse-recovery/detection.ts#L65-L76
- https://github.com/code-yeongyu/oh-my-openagent/blob/df070c57ca71d276fc08e5092601240a4d24130c/packages/omo-senpi/src/components/collapse-recovery/index.ts#L66-L73

Restrict dedup context to user-visible text. Add a regression proving hidden reasoning never enters the custom message or the retry provider request.

Then rebase onto current `dev`, regenerate all committed Senpi bundles, and rerun the current required matrix plus the isolated TTSR driver. The branch is currently conflicting and has none of the required CI contexts: https://github.com/code-yeongyu/oh-my-openagent/rules/17777319

PR #1155, opened 2026-01-26
Fix: Prevent system-reminder tags from triggering mode keywords

TheEpTic:
## Problem

Automated system messages with `<system-reminder>` tags were incorrectly triggering mode keywords like `[search-mode]`, `[analyze-mode]`, etc. when they contained words such as "search", "find", "explore", "investigate", etc.

This caused agents to incorrectly enter MAXIMUM SEARCH EFFORT mode when receiving automated system notifications, leading to confusing behavior and wasted computational resources.

## Root Cause

The `isSystemDirective()` function only checked for `[SYSTEM DIRECTIVE: OH-MY-OPENCODE` prefix but did not filter out XML-style `<system-reminder>` tags. When these automated system messages contained search-related keywords, the keyword detector would match them and inject mode directives.

## Solution

Added comprehensive filtering for `<system-reminder>` tags:

1. **New utility functions** in `src/shared/system-directive.ts`:
   - `hasSystemReminder()`: Detects presence of system-reminder tags
   - `removeSystemReminders()`: Strips system-reminder content before keyword detection

2. **Updated keyword detector** in `src/hooks/keyword-detector/index.ts`:
   - Calls `removeSystemReminders()` before running keyword pattern matching
   - Preserves user text while filtering automated system content

3. **Comprehensive test coverage**:
   - 22 tests for system-reminder utility functions
   - 6 integration tests for keyword detector filtering
   - Edge cases: case-insensitive tags, multiple blocks, mixed content, etc.

## Testing

All 46 tests pass, including:
- System-reminder filtering prevents search/analyze mode triggers
- User text keywords still detected correctly
- Multiple system-reminder blocks handled properly
- Case-insensitive tag matching
- Multiline system-reminder content

## Impact

✅ Fixes false-positive mode triggers from automated system messages
✅ Preserves legitimate user keyword detection
✅ No breaking changes to existing functionality
✅ Comprehensive test coverage ensures robustness

---
## Summary by cubic
Prevented automated <system-reminder> messages from triggering mode keywords like [search-mode] and [analyze-mode]. This stops false mode switches from system notifications while keeping user keyword detection unchanged.

- **Bug Fixes**
  - Strip <system-reminder> blocks before keyword detection using removeSystemReminders() and hasSystemReminder().
  - Updated keyword detector to clean text prior to pattern matching; user content is preserved.
  - Added tests (22 utility, 6 integration) covering case-insensitive tags, multiple blocks, and multiline content.

[...]

Issue #1154, opened 2026-01-26
[Bug]: Unable to install v3.0.0+ [Illegal instruction (core dumped)]


WilliamJudge94:
### Prerequisites

- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme)

### Bug Description

Unable to run any `bunx oh-my-opencode` commands for anything greater than v2.14.0

### Steps to Reproduce

1. Ubuntu 22.04.5
2. OpenCode 1.1.36
3. ldd command: Ubuntu GLIBC 2.35-0ubuntu3.11
4. CPU: Intel(R) Xeon(R) CPU E5-2630 v2
5. Bun 1.3.64.
6. `bunx oh-my-opencode install` does not return anything. Terminal says `error code 129`
7. `bunx oh-my-opencode@3.0.0 install` does not return anything. Terminal says `error code 129`
8. `bunx oh-my-opencode@3.1.0 install` does not return anything. Terminal says `error code 129`
9. `bunx oh-my-opencode@2.14.0 install` works

### Expected Behavior

Once `bunx oh-my-opencode install` is run for there to be the usual installation process of asking the user for the setup + completing the installation.

### Actual Behavior

Empty terminal with an `Error code 129`

### Doctor Output

```shell
Empty terminal with an `Error code 129`
```

### Error Logs

When I try to run the binary 

```shell
~/.bun/install/cache/oh-my-opencode-linux-x64@3.0.0###1/bin/oh-my-opencode --version
```

I get `Illegal instruction (core dumped)`.

### Configuration

```json

```

### Additional Context

I tried to upgrade from v2.14.0 by running the installation command again. The installation command did not produce the usual setup user input commands. Upon further review, it appeared that v3.0.0+ was no installed.

### Operating System

Linux

### OpenCode Version

1.1.36

justsisyphus:
## Fixed in dev branch

**Commit:** `a95caaf3`

### Changes
Added 4 new baseline platform targets for non-AVX2 CPUs:
- `bun-linux-x64-baseline` - Linux x64 (glibc, no AVX2)
- `bun-linux-x64-musl-baseline` - Linux x64 (musl, no AVX2)
- `bun-darwin-x64-baseline` - macOS x64 (no AVX2)
- `bun-windows-x64-baseline` - Windows x64 (no AVX2)

[...]

PR #7211, opened 2026-08-24
fix(omo-senpi): recover claude-sdk-oauth account slots on re-login

### LINKED ISSUE #7084: [Bug]: claude-sdk-oauth login appends slots; auth_error never expires (still in beta.13)
## Prerequisites

- [x] I searched existing issues and found no exact duplicate of the account-pool behavior.
- [x] The problem occurred with the latest omo-ai beta release (`5.0.0-0.beta.13` / senpi `2026.8.20-2`).
- [x] Dual-home `~/.omo` vs `~/.senpi` is already tracked in #6717 / #6809; this report is the Claude SDK OAuth **account pool** that beta.13 also left unchanged.

## Summary

`/login claude-sdk-oauth` does not refresh the existing Claude Pro/Max session. It **appends a new named slot**. A later `authentication_failed` / `invalid_grant` then stamps `blockReason: auth_error` on a slot, and that stamp **never expires**. Failover treats every leftover slot as a different account, so one Claude.com login becomes “All Claude accounts are currently blocked … after re-login.”

This is not weekly quota. It is local pool bookkeeping. Confirmed still present after upgrading to beta.13.

## Environment

- OS: macOS 26.5 (Darwin 25.5.0), arm64
- Node: v24.19.0
- npm: 11.17.0
- `omo --version`: `5.0.0-0.beta.13 (engine: senpi 2026.8.20-2)`
- One Claude Pro/Max login on claude.com
- Live agent dir for this install: `~/.senpi/agent` (see #6717)

No credential values are included below.

### Reproduction

1. `/login claude-sdk-oauth` (or import the existing Anthropic OAuth login).
2. Hit a later token error, then `/login claude-sdk-oauth` again as the product tells you to.
3. Repeat a few times.

### Actual

`auth.json` → `claude-sdk-oauth.accounts` grows:

| slot | source |
|---|---|
| `imported-anthropic` | import of the `anthropic` provider token |
| `account-2` … `account-5` | each subsequent `/login` |

Name allocation is `account-${existing.length + 1}` and the login path always `addAccount`s:

```js
// dist/core/extensions/builtin/claude-sdk-oauth/oauth-login.js
if (existing.length === 0) return "default";
if (!callbacks.onPrompt) return `account-${existing.length + 1}`;
// ...
return addAccount(current, toSlot(credential, name, "login"));
```

There is no replace-in-place, no “this is the same Claude.com user”, and a TUI re-login does not clear `auth_error` on the old slot.

### Expected

For a single-user install, a new `/login claude-sdk-oauth` should **replace or refresh** the existing slot (or at least drop slots whose refresh token was just superseded). Multi-account remains valid, but it should be explicit — not the default recovery path printed in the error text.

### Reproduction

1. Keep a single `claude-sdk-oauth` slot.
2. Let the access token expire, or import `anthropic` OAuth and then let that provider refresh (see Bug 3).
3. Next Claude turn.

### Actual

Anthropic returns `invalid_grant` / `authentication_failed`. Failover persists:

```js
// failover.js
if (classification.kind === "auth_error") {
  return { ...withoutExpiry, blockReason: "auth_error" }; // no blockedUntil
}

// affinity.js
function isBlocked(account, now) {
  return account.blockReason === "auth_error"
    || (account.blockedUntil !== undefined && account.blockedUntil > now);
}
// clearExpiredBlocks deliberately retains auth_error until login refreshes the slot
```

Because Bug 1 never refreshes the slot, “until login refreshes the slot” never happens. The UI then says:

```text
All Claude accounts for claude-sdk-oauth are currently blocked (rate limit or auth errors).
  Soonest automatic retry: after re-login.
```

`/login` adds `account-N+1`. The dead slot stays `auth_error`. Failover walks every slot, so the whole pool stays blocked.

Observed on this machine (redacted):

- `12:55Z` — `imported-anthropic` refresh → `invalid_grant` (`Refresh token not found or invalid`).
- Slot swapped to a later `account-5`; a one-line haiku call succeeded.
- `21:06Z` — two 90s stream-start stalls, then `authentication_failed`. `account-5` stamped `auth_error` with **no** `blockedUntil`. Access `expires` was still in the future.
- `21:19Z` beta.13 smoke test read that sticky flag and failed with “after re-login” without contacting Anthropic.
- A fresh `/login` created `account-2` **alongside** the dead `account-5`. Removing the dead slot made Claude work again.

### Expected

- `auth_error` should either expire, or be cleared when a new login for the same provider succeeds.
- A new login must not leave the rejected slot in the failover pool.
- “All Claude accounts blocked” must not be shown for leftover local stamps when a live slot exists.

## Bug 3 — `imported-anthropic` shares a refresh token with `anthropic`

First `/login claude-sdk-oauth` offered to import the existing Anthropic OAuth login. That copies `access` / `refresh` / `expires` into `imported-anthropic`.

The standalone `anthropic` provider later refreshes that same grant. Anthropic refresh tokens are single-use, so the copy dies with:

```text
authentication_failed: Anthropic token refresh request failed.
url=https://platform.claude.com/v1/oauth/token
status=400
body={"error":"invalid_grant","error_description":"Refresh token not found or invalid"}
```

That is what first stamped Bug 2 on this install. Import should either own the grant (and drop it from `anthropic`) or refresh through one store, not two copies.

## Expected behavior (all three)

1. One Claude.com user → one live `claude-sdk-oauth` slot unless the user explicitly adds another.
2. `/login claude-sdk-oauth` recovers that slot instead of appending `account-N`.
3. `auth_error` cannot dead-end the pool across a successful re-login.
4. Importing Anthropic OAuth must not fork the refresh token into two providers.

## Related

- #6717 / #6809 — branded `~/.omo` vs engine `~/.senpi` state split (still true on beta.13; login landed in `~/.senpi/agent/auth.json`).
- #7063 — hoisted Claude native binary path (separate).
- senpi changelog 2026.8.20 stream-start-timeout fork (#723) is in beta.13; it does **not** cover this pool / sticky-lock behavior.

AceRothstein71:
## What changed

Fixes both claude-sdk-oauth account-pool defects from #7084. Both ship inside the exact-pinned engine dependency `@code-yeongyu/senpi@2026.8.23` (the slot/failover/affinity/oauth-login plumbing lives in its `dist/core/extensions/builtin/claude-sdk-oauth/`; repo-wide greps confirm none of it exists in workspace sources), so the fix lands as a bun `patchedDependencies` bridge patch on the pin:

- **accounts.js**: new `upsertAccount()` replaces a same-name slot in place (fresh access/refresh/expires/source, block stamps stripped); unknown names still append so explicit multi-account keeps working.
- **oauth-login.js**: default recovery name is the newest existing login/import slot instead of minting `account-N+1`; a typed brand-new name still adds a second account; accepting the prompt placeholder recovers the newest login slot.
- **failover.js**: `AUTH_ERROR_BLOCK_MS` (30 min) gives `auth_error` stamps an expiry instead of blocking forever.
- **affinity.js**: `isBlocked()` is purely time-based and `clearExpiredBlocks()` clears any stamp whose expiry is absent or elapsed — cleaning legacy permanent stamps left by older versions. Live rate-limit windows still block.

The patch self-invalidates on pin drift (loud install failure), forcing conscious re-review or upstream senpi absorption.

## Why

`/login claude-sdk-oauth` appended a new named slot on every re-login (`addAccount`, name = `account-${n+1}`) while `auth_error` blocks were stamped with no `blockedUntil` and retained forever. One `invalid_grant` dead-ended the whole pool behind "All Claude accounts are currently blocked … after re-login", and the printed recovery path (`/login`) could never clear it because it only appended more slots.

## Failing-first test

Co-located regression suite `packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts` imports the installed dist modules via resolved file URLs (established repo pattern, cf. `senpi-test-runtime.ts`), fully dependency-injected (fake login flow, in-memory credential store, injected clock; no real credential store, no network):

- **RED** on the unpatched pin: 9 fail / 1 pass (`upsertAccount is not a function`, `AllAccountsBlockedError … until re-login`, legacy stamp retained).
- **GREEN** with the patch: 10 pass / 0 fail.

It pins: same-name upsert replaces in place; unknown names append; headless + prompt-placeholder re-login refreshes the stable slot; dead auth_error slot recovers after re-login; auth_error stamps carry an expiry; elapsed stamps stop blocking selection; legacy permanent stamps are cleaned; live rate-limit windows still block.

## QA evidence

`.omo/evidence/20260824-7084-oauth-slots/` (committed): WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED plus raw captures — red/green runs, scoped `tsgo` typecheck exit 0, manifest pin tests 15 pass, bundle purity + package shape 5 pass, fresh-install patch-application check.

## Residual risk

npm-installed `omo-ai` end users receive the fix when upstream senpi absorbs it and the pin is bumped (same absorption path as #7169); until then the patch covers every bun-install consumer of this monorepo. The 30-minute auth_error TTL is a policy choice; re-login remains the primary recovery path and clears stamps immediately. Live TUI `/login` drive omitted (requires a real Claude.com OAuth round trip); the module-level suite drives the exact functions that path invokes.

Fixes #7084

---
## Summary by cubic
Recover Claude OAuth account slots on re-login and add expiry to auth_error blocks to prevent pool dead-ends. Old: re-login appended `account-N+1` and `auth_error` blocks never expired. New: re-login upserts the same slot; `auth_error` blocks expire after 30 minutes. Fixes #7084.

**Review notes**
- Patches pinned engine `@code-yeongyu/senpi@2026.8.23` via Bun `patchedDependencies`.
- `accounts`: add `upsertAccount()` that replaces a same-name slot in place and clears block stamps; unknown names still append.
- `oauth-login`: default recovery name is the newest login/import slot; placeholder accepts to recover; typing a new name still adds a second account.
- `failover`: `AUTH_ERROR_BLOCK_MS = 30m` stamps `auth_error` with `blockedUntil`.
- `affinity`: blocking is time-based; `clearExpiredBlocks()` removes expired and legacy no-expiry stamps; rate-limit behavior unchanged.
- Adds regression suite `packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts` (red on unpatched pin, green with patch).

**Rollout**
- No migration required; re-login now refreshes the existing slot.
- If you bump `@code-yeongyu/senpi`, the patch will fail to apply; either remove it after upstream absorption or rebase it and re-run the suite.

<sup>Written for commit b61a7b4c90825065ef7e559b1057b2e7ea32a7df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260824-7084-oauth-slots/after-fix-green.txt
+++ b/.omo/evidence/20260824-7084-oauth-slots/after-fix-green.txt
@@ -1,0 +1,6 @@
+bun test v1.3.14 (0d9b296a)
+
+ 10 pass
+ 0 fail
+ 29 expect() calls
+Ran 10 tests across 1 file. [925.00ms]
--- a/.omo/evidence/20260824-7084-oauth-slots/before-fix-red.txt
+++ b/.omo/evidence/20260824-7084-oauth-slots/before-fix-red.txt
@@ -1,0 +1,108 @@
+bun test v1.3.14 (0d9b296a)
+
+packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts:
+116 | 
+117 | describe("engine claude-sdk-oauth slot recovery (issue #7084)", () => {
+118 |   describe("#given the pinned senpi accounts module", () => {
+119 |     test("#then it exposes an upsert that replaces a same-name slot in place", async () => {
+120 |       const accounts = await describeAccounts()
+121 |       expect(typeof accounts.upsertAccount).toBe("function")
+                                                  ^
+error: expect(received).toBe(expected)
+
+Expected: "function"
+Received: "undefined"
+
+      at <anonymous> (/home/viprix/projects/oom-wt-7084/packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts:121:45)
+(fail) engine claude-sdk-oauth slot recovery (issue #7084) > #given the pinned senpi accounts module > #then it exposes an upsert that replaces a same-name slot in place [6.58ms]
+137 |       accounts.assertSentinelInvariant(next)
+138 |     })
+139 | 
+140 |     test("#then an unknown name still appends so explicit multi-account keeps working", async () => {
+141 |       const accounts = await describeAccounts()
+142 |       const current = accounts.upsertAccount(accounts.emptyCredential(), loginSlot("default", "a"))
+                                     ^
+TypeError: accounts.upsertAccount is not a function. (In 'accounts.upsertAccount(accounts.emptyCredential(), loginSlot("default", "a"))', 'accounts.upsertAccount' is undefined)
+      at <anonymous> (/home/viprix/projects/oom-wt-7084/packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts:142:32)
+(fail) engine claude-sdk-oauth slot recovery (issue #7084) > #given the pinned senpi accounts module > #then an unknown name still appends so explicit multi-account keeps working [0.37ms]
+151 | 
+152 |   describe("#given a single logged-in Claude account #when /login succeeds again without an explicit new name", () => {
+153 |     test("#then the existing slot is refreshed in place instead of appending account-N+1", async () => {
+154 |       const accounts = await describeAccounts()
+155 |       const oauthLogin = await describeOAuthLogin()
+156 |       const current = accounts.upsertAccount(accounts.emptyCredential(), loginSlot("default", "a"))
+                                     ^
+TypeError: accounts.upsertAccount is not a function. (In 'accounts.upsertAccount(accounts.emptyCredential(), loginSlot("default", "a"))', 'accounts.upsertAccount' is undefined)
+      at <anonymous> (/home/viprix/projects/oom-wt-7084/packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts:156:32)
+(fail) engine claude-sdk-oauth slot recovery (issue #7084) > #given a single logged-in Claude account #when /login succeeds again without an explicit new name > #then the existing slot is refreshed in place instead of appending account-N+1 [5.23ms]
+173 |     })
+174 | 
+175 |     test("#then a dead auth_error slot comes back live after re-login", async () => {
+176 |       const accounts = await describeAccounts()
+177 |       const oauthLogin = await describeOAuthLogin()
+178 |       const current = accounts.upsertAccount(accounts.emptyCredential(), loginSlot("default", "a", { blockReason: "auth_error" }))
+                                     ^
+TypeError: accounts.upsertAccount is not a function. (In 'accounts.upsertAccount(accounts.emptyCredential(), loginSlot("default", "a", { blockReason: "auth_error" }))', 'accounts.upsertAccount' is undefined)
+      at <anonymous> (/home/viprix/projects/oom-wt-7084/packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts:178:32)
+(fail) engine claude-sdk-oauth slot recovery (issue #7084) > #given a single logged-in Claude account #when /login succeeds again without an explicit new name > #then a dead auth_error slot comes back live after re-login [0.40ms]
+193 | 
+194 |     test("#then accepting the prompt placeholder recovers the newest login slot instead of minting account-N+1", async () => {
+195 |       const accounts = await describeAccounts()
+196 |       const oauthLogin = await describeOAuthLogin()
+197 |       const imported = { ...loginSlot("imported-anthropic", "i"), source: "import" as const }
+198 |       const withImport = accounts.upsertAccount(accounts.emptyCredential(), imported)
+                                        ^
+TypeError: accounts.upsertAccount is not a function. (In 'accounts.upsertAccount(accounts.emptyCredential(), imported)', 'accounts.upsertAccount' is undefined)
+      at <anonymous> (/home/viprix/projects/oom-wt-7084/packages/omo-senpi/src/engine-claude-sdk-oauth-slot-recovery.test.ts:198:35)
+(fail) engine claude-sdk-oauth slot recovery (issue #7084) > #given a single logged-in Claude account #when /login succeeds again without an explicit new name > #then accepting the prompt placeholder recovers the newest login slot instead of minting account-N+1 [0.84ms]
+220 |     })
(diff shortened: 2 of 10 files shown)


MoerAI (changes requested):
The focused patched-runtime suite is green, but three product contracts still block this fix.

1. Preserve the permanent-auth-block policy from Senpi #1154 and #1155. A successful re-login may clear or replace the selected slot, but `invalid_grant` must not auto-expire after 30 minutes:
- https://github.com/code-yeongyu/senpi/pull/1154
- https://github.com/code-yeongyu/senpi/pull/1155
2. For multi-account pools, do not let a headless or blank re-login overwrite the newest slot without identity proof. Require an explicit target or obtain approval for another identity-safe policy.
3. Carry the fix through the supported `omo-ai` postinstall transformer and packed-consumer contract from #7500. A root Bun patched dependency alone does not fix npm users:
https://github.com/code-yeongyu/oh-my-openagent/pull/7500

Add integration coverage through the real auth-store/login write path proving that only the intended slot changes, sibling blocks and pins survive, and two session IDs retain stable HRW account assignments.

Then rebase onto current `dev` and Senpi 2026.8.31, regenerate the version-scoped patch, and pass every required check.

code-yeongyu:
Superseded by code-yeongyu/senpi#1789 (merged, e9de2b6d7f), which lands the same recovery contracts in senpi source instead of a dist patch:

- Re-login refreshes the same-name slot in place (clearing its `auth_error` block, preserving `displayName`) instead of appending `account-N+1`.
- The import becomes a **move** out of the `anthropic` provider, so one single-use grant never lives in two stores.
- The identity-safe targeting from the review is implemented as a matrix: a blank or headless re-login only ever refreshes a lone slot or the pool's one auth-blocked slot — never the newest working slot of a multi-account pool; an explicit existing name refreshes, a new name appends.
- The fix ships to npm users via the normal senpi release rather than a root `patchedDependencies` entry, which resolves the postinstall-transformer concern.

Thank you for mapping the defect family first — the review contracts from this thread are baked into the new regression tests (`test/suite/regressions/7084-*.test.ts`). Closing as superseded; omo picks up the engine fix on the next senpi pin.

PR #7213, opened 2026-08-24
fix(utils): resolve windows sweep launchers from System32 and annotate ENOENT

### LINKED ISSUE #6747: Windows: lsp-daemon proxy sweep permanently disabled by spawn powershell.exe ENOENT
### Symptom

`~/.omo/omo-debug.log`, Windows 11 Pro (win32 10.0.26200), Node v24.18.0, `omo-ai@5.0.0-0.beta.5`:

```
[2026-08-11T02:06:14.093Z] hidden stderr while TUI active
lsp-daemon proxy sweep skipped: spawn powershell.exe ENOENT undefined
```

7 occurrences between 02:06Z and 05:59Z. The sweep never completes, so orphaned LSP proxy processes
are never reaped — the exact failure `sweepLspProxies` exists to prevent.

### Cause

`packages/utils/src/process-sweep/exec.ts` resolves its helpers through PATH:

```ts
function enumerateWindowsProcesses(): Promise<ProcessInfo[]> {
  const command = [
    "Get-CimInstance Win32_Process",
    "Select-Object ProcessId,ParentProcessId,CommandLine",
    "ConvertTo-Json -Compress -Depth 2",
  ].join(" | ")
  return execFileText("powershell.exe", ["-NoProfile", "-Command", command]).then(parseWindowsProcessTable)
}

function createWindowsKiller(): ProcessKiller {
  return {
    isAlive: defaultIsProcessAlive,
    kill: (pid) => execFileVoid("taskkill.exe", ["/PID", String(pid), "/T", "/F"]),
    terminate: (pid) => execFileVoid("taskkill.exe", ["/PID", String(pid), "/T"]),
  }
}
```

The sibling `packages/utils/src/process-tree-termination.ts` already avoids the async-`spawn`
variant of this problem, and the same class of bug was just fixed upstream in senpi
(`spawn taskkill ENOENT` crashing shutdown, code-yeongyu/senpi#807), so this looks like the last
PATH-dependent Windows launcher in the sweep path.

### What I could not establish

I could not reproduce the ENOENT locally. On the affected machine PATH does contain both
`C:\WINDOWS\System32` and `C:\WINDOWS\System32\WindowsPowerShell\v1.0\`, and a plain
`execFile("powershell.exe", ["-NoProfile", "-Command", "$PSVersionTable.PSVersion.Major"])` from Node
succeeds. So the failing lookup happens under some environment the sweep runs in that I have not
isolated — a spawned daemon with a curated `env`, or a PATH mutated during startup
(`prependToolkitPath` writes `process.env.PATH`).

Rather than guess, filing this with the evidence. Two things seem worth doing regardless of the exact
trigger:

1. Resolve `powershell.exe` and `taskkill.exe` from `%SystemRoot%\System32` when that exists instead
   of relying on PATH, keeping the bare name as a fallback.
2. Log which command failed to launch at a level that is actually visible, rather than folding it
   into a generic `... sweep skipped` line — the current message loses the fact that process
   reaping is now permanently disabled for the session.

Happy to send a PR for (1) and (2) if that direction is agreeable.

AceRothstein71:
## What

`packages/utils/src/process-sweep/exec.ts` resolved its Windows launchers through PATH only: `enumerateWindowsProcesses()` spawned bare `powershell.exe` and `createWindowsKiller()` spawned bare `taskkill.exe`. Under a daemon process with a curated or mutated PATH that lookup fails with ENOENT, the error is folded into a generic `lsp-daemon proxy sweep skipped:` line, and orphaned LSP proxy processes are never reaped for the rest of the session (#6747).

This PR:

1. Adds `resolveWindowsSystemBinary()`: both launchers now prefer the absolute `%SystemRoot%\System32\...` location (`SystemRoot`, then `windir`) when that file exists, keeping the bare name as the fallback when `SystemRoot` is unknown or the candidate is missing, so a broken PATH can no longer disable the sweep.
2. Annotates ENOENT launch failures with the attempted launcher name (`process sweep could not launch "powershell.exe": ...`, original error preserved as `cause`). The consumer log at `packages/omo-opencode/src/shared/omo-process-sweep.ts` interpolates `error.message`, so the skip line now names the binary instead of `spawn powershell.exe ENOENT undefined`.

Adds a co-located regression suite `packages/utils/src/process-sweep/exec.test.ts`: platform-agnostic resolver unit tests (injected env + `fileExists`) plus end-to-end tests driving the real `execFile` path against executable stand-ins under a fake `%SystemRoot%`, gated with `test.skipIf(process.platform === "win32")` because the stand-ins are POSIX shell scripts.

## Why

The issue reporter's machine had both `System32` and the WindowsPowerShell directory on PATH yet still saw `spawn powershell.exe ENOENT` from the spawned sweep, so a PATH-only lookup is not reliable in the environment the daemon runs under. Resolving from the documented absolute system location removes that dependency, and naming the failed launcher makes the permanent loss of process reaping visible instead of silent.

## Verified

- Failing-first: with only the test present on base dev @ 8833800ae, the suite fails (`Export named 'resolveWindowsSystemBinary' not found`, 0 pass / 1 fail); after the fix it is green (7 pass / 0 fail). Captured in `.omo/evidence/20260824-6747-powershell-enoent/failing-first-base.txt` and `scoped-test-after.txt`.
- Full `packages/utils` Bun suite: 506 pass / 0 fail (`utils-package-suite.txt`).
- Typecheck: `tsgo --noEmit -p packages/utils/tsconfig.json` exit 0; `bunx biome check` clean on both changed files.
- QA evidence (including the Linux-host limitation for live win32 runtime testing) is committed under `.omo/evidence/20260824-6747-powershell-enoent/QA.md`.

## Risk

Low. Behavior changes only when `%SystemRoot%\System32\<candidate>` exists, which is the documented layout on every supported Windows version; otherwise the previous bare-name lookup is used unchanged. Non-Windows paths (`ps`, POSIX killers) are untouched. Live verification on a real Windows host was not possible from this Linux environment; the win32-gated e2e tests cover the spawn path hermetically.

Fixes #6747

### DIFF
--- a/.omo/evidence/20260824-6747-powershell-enoent/QA.md
+++ b/.omo/evidence/20260824-6747-powershell-enoent/QA.md
@@ -1,0 +1,59 @@
+# QA evidence: Windows sweep launcher resolution, spawn powershell.exe ENOENT (#6747)
+
+Date: 2026-08-24
+Branch: issue/6747-lsp-daemon-powershell-enoent (base: dev @ 8833800ae)
+Scope: packages/utils/src/process-sweep/exec.ts (+ co-located exec.test.ts)
+
+## WHAT WAS TESTED
+
+1. Failing-first regression proof (`failing-first-base.txt`): with the exec.ts fix stashed and
+   only the new test present, `bun test packages/utils/src/process-sweep/exec.test.ts` fails on
+   the base commit with
+   `SyntaxError: Export named 'resolveWindowsSystemBinary' not found` (0 pass / 1 fail).
+2. Scoped suite after the fix (`scoped-test-after.txt`): same command, 7 pass / 0 fail:
+   - resolveWindowsSystemBinary unit tests (platform-agnostic, injected env + fileExists):
+     SystemRoot absolute path wins; windir fallback tree used; missing candidate falls back to
+     the bare PATH name; no SystemRoot/windir keeps the bare name.
+   - End-to-end enumeration through a fake %SystemRoot% whose System32 carries an executable
+     stand-in for powershell.exe: parsed process rows come from the resolved absolute binary.
+   - End-to-end kill/terminate through a fake System32 taskkill.exe stand-in: the resolved
+     absolute binary runs with `/PID <pid> /T [/F]`.
+   - ENOENT annotation: with no SystemRoot and no launcher on PATH, the rejection message names
+     the attempted launcher (`"powershell.exe"`) and carries `cause.code === "ENOENT"`, so the
+     consumer log line (`lsp-daemon proxy sweep skipped: ...`) in
+     packages/omo-opencode/src/shared/omo-process-sweep.ts now shows which binary failed to
+     launch instead of a bare `spawn powershell.exe ENOENT undefined`.
+3. Full packages/utils Bun suite (`utils-package-suite.txt`): 506 pass / 0 fail.
+4. Typecheck (`typecheck.txt`): `tsgo --noEmit -p tsconfig.json` exit 0.
+5. `bunx biome check` on both changed files: clean.
+
+## WHAT WAS OBSERVED
+
+- The win32 launchers in exec.ts previously resolved powershell.exe / taskkill.exe through PATH
+  only. Under a daemon with a curated or mutated PATH (the #6747 reporter's machine had both
+  directories on PATH yet still saw ENOENT from the spawned sweep), the lookup fails and the
+  proxy sweep is skipped for the whole session, so orphaned LSP proxies are never reaped.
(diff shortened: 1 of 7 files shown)


MoerAI:
Thanks for the PR, @AceRothstein71, and for the disciplined root-cause work on #6747.

I am closing this as superseded by #7261. Its `exec.ts` and `exec.test.ts` changes are byte-identical to this implementation, but it carries the fix on an in-repository branch where the complete matrix, including Windows tests, could run. The resolver fallback, injected test dependencies, and ENOENT-only annotation are all preserved.

The replacement should credit this contribution when it lands. It should also be sequenced deliberately with #6740, which addresses the same root cause across additional call sites.

PR #7216, opened 2026-08-24
fix(omo-senpi): degrade reflection sandbox when bwrap cannot start a user namespace

### LINKED ISSUE #6873: [Bug]: Memory reflection auto sandbox repeatedly fails when bwrap cannot create a user namespace

AceRothstein71:
## What

- `memory.reflection.sandbox: "auto"` now verifies that the resolved `bwrap` executable can actually start a sandbox, not merely that it exists. A one-shot smoke test (`bwrap --ro-bind / / --dev-bind /dev /dev --tmpfs /tmp true`, 10s timeout) runs once per executable path per process; on failure the reflection/facts children run unsandboxed with an explicit degradation warning naming the cause (e.g. `setting up uid map: Permission denied`).
- `sandbox: "required"` stays fail-closed: an unusable bwrap raises `SandboxUnavailableError` before any child spawns.
- Reflection failure remediation for bubblewrap setup failures (`bwrap: setting up uid map: Permission denied`) now points at the `memory.reflection.sandbox` config escape hatch instead of `runtime/reflection-sessions/<runId>/child-stderr.log`, which run-dir cleanup has usually already removed.
- Tests: policy-level regression tests in `sandbox.test.ts` (auto degrade + warning, required fail-closed, healthy bwrap keeps wrapping, off never probes), unit tests for the smoke classifier and its memoization in `sandbox-bwrap-probe.test.ts`, and a remediation-hint test in `worker/remediation.test.ts`.

## Why

On Ubuntu with AppArmor restricting unprivileged user namespaces (#6873), bubblewrap is installed but every sandboxed reflection child dies at setup with `bwrap: setting up uid map: Permission denied`. `auto` never degraded, so reflection failed on every trigger and the consecutive-failure warning repeated indefinitely, while the diagnostic pointed at a log file that no longer existed.

## Verified

- `bunx tsgo --noEmit -p packages/omo-senpi/tsconfig.json`: clean.
- Scoped suite (`sandbox.test.ts`, `sandbox-bwrap-probe.test.ts`, `sandbox-absent-paths.test.ts`, `worker/remediation.test.ts`): 29 pass, 1 skip (win32-gated), 0 fail.
- Full memory component suite (`bun test packages/omo-senpi/src/components/memory`): 911 pass, 6 skip, 0 fail.
- Evidence recorded in `.omo/evidence/20260824-6873-bwrap-fallback/QA-EVIDENCE.md` (what was tested/observed/why enough/omitted). Live Senpi harness QA was not run in this environment and none is claimed.

## Risk

- Low. Pre-existing tests all inject `which`, so they keep existence-only semantics and never spawn a real bwrap; only production callers get the probe, memoized once per process per path (~one extra child spawn per process on linux).
- Residual: a host where the smoke test passes but a later real launch still fails at setup keeps the pre-existing runtime behavior, now with the corrected remediation hint.

Fixes #6873

---
## Summary by cubic
Degrades Linux sandboxing for reflection and facts to unsandboxed when `bwrap` exists but cannot start a user namespace, preventing repeated child failures on hosts that block unprivileged namespaces. Previously `auto` only checked for `bwrap` presence and kept failing; now it probes once per path and degrades with a warning, while `required` still fails closed with `SandboxUnavailableError`.

- Behavior change (Linux only): `auto` runs a one-shot `bwrap` smoke test (`--ro-bind / / --dev-bind /dev /dev --tmpfs /tmp true`, 10s timeout) and either wraps or degrades with an explicit warning naming the cause; `required` rejects early; `off` unchanged. Probe result is memoized per executable path per process and only runs when no test `which` is injected; callers may override via an optional `probe`.
- API notes: export `SandboxUsability`; `buildSandboxTransform` (and internal `buildPathSandboxTransform`) accept an optional `probe` parameter. No migration required for existing callers.
- Remediation: for bubblewrap setup failures, hints now point to `memory.reflection.sandbox` instead of `runtime/reflection-sessions/<runId>/child-stderr.log`.

<sup>Written for commit 7811c387a85243586b179bf4c4f84875017fce50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260824-6873-bwrap-fallback/QA-EVIDENCE.md
+++ b/.omo/evidence/20260824-6873-bwrap-fallback/QA-EVIDENCE.md
@@ -1,0 +1,56 @@
+# WHAT WAS TESTED
+
+Issue #6873: memory reflection with `memory.reflection.sandbox: "auto"` treats an existing
+`/usr/bin/bwrap` as usable even when bubblewrap cannot create its user namespace
+(AppArmor `apparmor_restrict_unprivileged_userns = 1`), so every reflection child dies with
+`bwrap: setting up uid map: Permission denied` and the failure streak grows.
+
+Commands run (repo root, bun 1.3.14):
+
+1. `bun test packages/omo-senpi/src/components/memory/sandbox.test.ts packages/omo-senpi/src/components/memory/sandbox-bwrap-probe.test.ts packages/omo-senpi/src/components/memory/sandbox-absent-paths.test.ts packages/omo-senpi/src/components/memory/worker/remediation.test.ts`
+   - Surface: the sandbox transform builders (`auto`/`required`/`off` policies, linux/darwin),
+     the new bwrap smoke-probe classifier + memoized probe, and the reflection failure-hint mapper.
+   - Behavior proven: auto degrades to unsandboxed with an explicit warning when the probe reports
+     bwrap unusable; required fails closed with `SandboxUnavailableError`; a healthy probe result
+     keeps wrapping; `off` never probes; classifier maps exit 0 / uid-map denial / spawn error /
+     timeout to usable/unusable+reason; remediation names `memory.reflection.sandbox` instead of
+     the removed `child-stderr.log` for bwrap setup failures while generic child_exit keeps the log hint.
+2. `bunx tsgo --noEmit -p packages/omo-senpi/tsconfig.json`
+   - Surface: whole omo-senpi package typecheck gate.
+3. `bun test packages/omo-senpi/src/components/memory`
+   - Surface: full memory component suite (sandbox consumers included: identity-runtime lazy
+     transform, facts surface, worker supervision/finalization) to prove no regression.
+
+# WHAT WAS OBSERVED
+
+1. Scoped run after fixes: `29 pass, 1 skip (win32-gated probe test), 0 fail`.
+2. Typecheck: exit code 0, no diagnostics.
+3. Full memory suite: `911 pass, 6 skip, 0 fail` across 134 files (~326s).
+4. Failing-first evidence: before the remediation.ts branch existed, the new
+   `worker/remediation.test.ts` case failed (hint still pointed at child-stderr.log); before the
+   classifier reason fix, two `classifyBwrapSmoke` cases failed because empty stderr omitted
+   `reason` entirely, violating the declared `{ usable: false; reason: string }` shape.
+5. Hermeticity check: all pre-existing sandbox tests inject `which`, so they keep existence-only
+   semantics and never spawn a real bwrap; only production callers without an injected `which`
+   (identity-runtime reflection transform, facts surface) get the real smoke probe, memoized once
+   per executable path per process.
+
+# WHY IT IS ENOUGH
+
+- The issue's three asks are each pinned by a test: verify-the-sandbox-can-start (probe +
+  classifier tests), auto-retry-unsandboxed-with-explicit-warning vs required-fail-closed
+  (policy tests), and no dangling child-stderr.log pointer for sandbox setup failures
+  (remediation test).
+- The full memory suite covers both production consumers of the changed builder signature, so the
+  blast radius (reflection + facts children) is exercised, not just the unit seams.
+- Remaining risk: hosts where the smoke test passes but a later real launch still fails at setup
+  (race or per-invocation divergence). That path keeps the pre-existing runtime behavior plus the
+  new remediation hint, which is the best available degradation without re-probing per launch.
+
+# WHAT WAS OMITTED
+
+- Live Senpi harness QA (`senpi-qa` drivers) was not run in this environment; the hermetic unit
+  gate (`tsgo --noEmit -p packages/omo-senpi/tsconfig.json` + scoped/full `bun test`) is the
+  recorded evidence. No live driver JSON is included, and none is claimed.
+- No secrets, tokens, auth headers, or env dumps are present in this change or its artifacts;
+  nothing required redaction beyond this note.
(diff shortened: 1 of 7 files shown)


code-yeongyu:
Closing this one in favor of #7332, which just merged into `dev` — but I want to be clear about the order of events: **this PR got there first, and it got it right.**

The diagnosis (existence-check ≠ usability on hosts with `apparmor_restrict_unprivileged_userns=1`), the smoke-probe-plus-memoization design, the `classifyBwrapSmoke`/`probeBwrapUsability` split, and the `auto`-degrades / `required`-fails-closed policy routing in #7332 all follow the approach you charted here. The merged version differs only in small mechanics (probe gating keyed off executable existence rather than `which` injection, `ETIMEDOUT`-aware timeout classification, a 3s probe timeout, and a file split to satisfy an internal LOC ceiling), plus the regenerated extension bundles our CI gate requires — which is the main reason we landed a maintainer-authored equivalent instead of merging this branch directly.

Thank you @AceRothstein71 for digging into a bug that only reproduces on a very specific host configuration, writing up the QA evidence so cleanly, and for the failing-first test discipline. Genuinely good work, and it directly shaped what shipped. The fix lands in the next release; #6873 is closed.

Would love to see you around the repo again. 🙇

PR #7218, opened 2026-08-24
fix(cli): require explicit config scope when a distinct custom OpenCode config dir is active

### LINKED ISSUE #6845: [Bug]: Installer implicitly writes only to OPENCODE_CONFIG_DIR without scope confirmation
## Prerequisites

- [x] I searched existing issues. This is related to, but distinct from, #3846 / #3875.
- [x] I am using the latest published version, 4.19.4.
- [x] I read the installation and OpenCode configuration documentation.

## Bug Description

#3846 / #3875 fixed OMO runtime discovery so that the default OpenCode global directory remains visible when `OPENCODE_CONFIG_DIR` is set. The installer still resolves one exclusive write target via `OPENCODE_CONFIG_DIR ?? defaultGlobalDir`.

When a host such as Orca launches OpenCode with a distinct `OPENCODE_CONFIG_DIR`, the Ultimate installer implicitly registers OMO only in that host-managed directory. It prints the selected path after writing, but it does not explain the effective scope before mutation or ask whether the user wants a host-only or user-wide registration. A first-time install therefore works inside the host but does not make OMO available to a normal OpenCode launch outside that host.

This is especially confusing because the installation guide verifies `~/.config/opencode/opencode.json`, while the installer has no active/global scope option.

## Steps to Reproduce

1. Start an OpenCode terminal through a host that exports `OPENCODE_CONFIG_DIR` to a directory distinct from the default global OpenCode directory. Orca 1.4.182 does this with its managed OpenCode hook directory.
2. Ensure the default global OpenCode config has no `oh-my-openagent` plugin entry.
3. Run the documented OpenCode installer flow, for example:
   ```sh
   bunx oh-my-openagent@4.19.4 install --no-tui --platform=opencode --claude=no --openai=yes --gemini=no --copilot=no --opencode-zen=no --zai-coding-plan=no
   ```
4. Observe that the installer reports the host-managed config path after writing and that this config receives `oh-my-openagent@latest`, while `~/.config/opencode/opencode.json[c]` remains unchanged.
5. Run OpenCode without `OPENCODE_CONFIG_DIR` and inspect `opencode debug info`: OMO is not loaded.

## Expected Behavior

A non-empty `OPENCODE_CONFIG_DIR` that resolves to a path different from the default global directory must not implicitly choose the user's installation scope.

Interactive install should show both resolved paths before writing and require one of:

- `active`: register the OpenCode server and TUI plugin entries only in the active custom directory for intentional isolation;
- `global`: register them only in the default global directory for user-wide availability. OpenCode loads the global layer when the custom directory is active, so this also makes OMO available inside the host without duplicating the declaration.

A `both` mode is intentionally not requested: duplicating the same registration across additive configuration layers creates two ownership points and unnecessary version-drift risk.

For `--no-tui`, require an explicit scope such as `--config-scope=active|global` whenever the two resolved directories differ. If no distinct custom directory is present, preserve the existing global behavior.

The installer should label the chosen scope and path before mutation, then report the written path afterward. `doctor` should inspect and report both effective roots when a distinct custom directory is active, rather than checking only the singular installer target. The installation guide should explain how to verify the actual selected path instead of showing only a hard-coded global path.

## Actual Behavior

The installer picks the active `OPENCODE_CONFIG_DIR` as its only write target. It reports that path after writing, but does not identify the resulting registration as host-scoped or offer a user-wide target. Normal OpenCode launches outside the host therefore do not load OMO.

## Doctor Output

Redacted to remove account and absolute-path information:

```text
System  1.18.18 · 4.19.4 · Bun 1.3.14
Config  <orca-managed-custom-config-dir>/opencode.json (valid)
Tools   LSP: 1 server; AST-Grep: unavailable; gh: installed
MCPs    websearch, context7, grep_app, lsp
```

This confirms that `doctor` can display the selected custom path; the remaining gap is that it does not compare both effective roots or explain the registration scope.

## Configuration

```json
{
  "OPENCODE_CONFIG_DIR": "<orca-managed-custom-config-dir>",
  "defaultGlobalPluginEntries": [
    "@warp-dot-dev/opencode-warp"
  ],
  "activeCustomPluginEntries": [
    "oh-my-openagent@latest"
  ],
  "pluginsWhenLaunchedWithoutOPENCODE_CONFIG_DIR": [
    "@warp-dot-dev/opencode-warp"
  ]
}
```

## Additional Context

OpenCode documents `OPENCODE_CONFIG_DIR` as an additional custom directory loaded after global configuration, rather than a replacement for global configuration. #3846 (fixed by #3875, merged 2026-05-20 UTC) addressed additive *runtime discovery*. This report is limited to the remaining installer/doctor/documentation scope-selection gap.

A global-only OMO registration was also verified to remain visible when OpenCode was launched with a distinct empty custom directory, with `plugin_origins` reporting the OMO entry as `scope: global`. This is why duplicate registration in both layers is unnecessary.

When rerun against an existing registration in the other layer, the installer must not silently move or delete it. It should report the conflict and require separate explicit cleanup. Add regression coverage for active and global scope selection, including an Orca-style custom directory.

## Environment

- OS: macOS
- OpenCode: 1.18.18
- oh-my-openagent: 4.19.4
- Bun: 1.3.14
- Host: Orca 1.4.182

AceRothstein71:
## What

Implements the installer-side contract from #6845 for the case where the active OpenCode config root (`OPENCODE_CONFIG_DIR`) differs from the default global root:

- Interactive install now displays both effective roots and requires an explicit `active` or `global` choice before registering anything.
- `--no-tui` installs fail with exit code 1 unless `--config-scope=active|global` is supplied; the error names both roots and the required flag.
- With no distinct custom root, existing behavior is unchanged (no prompt, no flag requirement).
- `both` is intentionally excluded: rejected by Commander `choices` on the new `--config-scope` option and again by runtime validation.
- An existing registration in the other layer is never silently moved or deleted; exactly one root receives `opencode.json` + `tui.json` per run.

New module `install-config-scope.ts` resolves distinct roots (`getOpenCodeConfigDirs`), validates non-TUI scope, and applies the choice through a config-dir override in `ConfigContext`; `ensureTuiPluginEntry` is now override-aware so `tui.json` lands in the chosen root. `resetConfigContext()` also clears the override to avoid stale-scope leakage.

## Why

Previously the installer always registered into whichever root was "active" and ignored that a second, default global root existed. Users running with a custom `OPENCODE_CONFIG_DIR` got a silent single-root registration with no confirmation, and non-interactive installs had no way to express intent. This is exactly the wrong-default / missing-confirmation behavior reported in #6845.

## Verified

- Failing-first: with the fix reverted, 3 of 4 new regression tests fail on base behavior (silent success without scope, global scope ignored, `both` accepted). Evidence: `.omo/evidence/20260824-6845-installer-scope/red-failing-first.log`
- Scoped suite green after fix: `bun test packages/omo-opencode/src/cli/install.test.ts` -> 7 pass / 0 fail. Tests drive the real `install()` entry inside an isolated sandbox (`OPENCODE_CONFIG_DIR` + `XDG_CONFIG_HOME` pointed at fresh temp dirs, env restored afterwards) and assert real filesystem outcomes: which `opencode.json` / `tui.json` exist in which root.
- `bun run typecheck` exit 0 (tsgo --noEmit + script + all package tsconfigs).
- Live CLI probe: `install --help` lists `--config-scope <scope>` with choices; `--config-scope=both` is rejected ("Allowed choices are active, global"). Evidence: `cli-flag-probe.log`
- Full QA record (WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED): `.omo/evidence/20260824-6845-installer-scope/EVIDENCE.md`

## Risk

Low-to-moderate, installer-only. The config-dir override changes where install writes when a distinct custom root exists; users who relied on the old silent default must now pass `--config-scope` for non-interactive installs (interactive flow prompts instead). Per the issue thread, doctor reporting of both effective roots plus docs are part of the complete fix and are left to a linked follow-up patch, so this PR addresses the installer contract and regression coverage only.

Fixes #6845

---
## Summary by cubic
Require explicit install scope when a custom `OPENCODE_CONFIG_DIR` differs from the default global root. Previously installs always wrote to the active root; now TUI prompts for `active` or `global`, and non-TUI requires `--config-scope=active|global` or exits 1. `tui.json` and `opencode.json` are written only in the chosen root; behavior is unchanged when no distinct root exists.

- Implement the installer contract from #6845: add `--config-scope` (choices: active, global); reject unsupported values and missing scope in non-TUI with exit 1 and an error naming both roots.
- In TUI, show both paths and require an explicit choice; cancel exits 1.
- Apply scope via a config-dir override in `ConfigContext`; make `ensureTuiPluginEntry` use the scoped `getConfigDir`; clear the override in `resetConfigContext` to prevent leakage.
- Never move or delete an existing registration in the other root; “both” is intentionally excluded.
- Add regression tests that assert real filesystem outcomes; no changes outside the installer paths.

Bold Rollout

- Non-interactive installs that run with a custom `OPENCODE_CONFIG_DIR` must pass `--config-scope=active` or `--config-scope=global`.
- No action needed for interactive installs (prompt handles it) or when only a single root is present.

<sup>Written for commit f51cbcbdd4481e0bcf2c2503ab47837f269bbcf2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for the scoped configuration design. The confirmed blocker is an existing installer regression: `cli-installer.test.ts` fails because it sets `OPENCODE_CONFIG_DIR` without `XDG_CONFIG_HOME`, so the new gate returns exit 1 where the test expects 0. Please update that path to supply the intended scope or align its environment so the roots are not distinct. Then run the broader installer suite, including `cli-installer.test.ts` and `install.test.ts`, and provide passing results plus typecheck proof. https://github.com/code-yeongyu/oh-my-openagent/pull/7218/files

PR #7221, opened 2026-08-24
fix(background-agent): strip platform session-id prefix before existence check

### LINKED ISSUE #6263: bug(background-agent): checkSessionExistence passes opencode:-prefixed storage ID to SDK client.session.get(), causing HTTP 400 fetch error

AceRothstein71:
## What

`checkSessionExistence()` in `packages/omo-opencode/src/features/background-agent/session-existence.ts` passed the incoming session id straight into `client.session.get({ path: { id } })`. Boulder state (`.omo/boulder.json`) records session ids normalized with a platform prefix (`opencode:` / `codex:` / `senpi:`, via `normalizeSessionId` in `@oh-my-opencode/boulder-state`), while the storage layer resolves bare ids only. This change adds a `bareSessionId()` helper at that boundary and strips the prefix before constructing the SDK path, so prefixed and raw ids resolve identically.

## Why

With a prefixed id, the SDK path became `/session/opencode%3Ases_...`, which 404s even though the session exists. The task poller (both call sites in `task-poller.ts`) then classified live subagent sessions as missing every poll cycle and attempted recovery by re-sending full turn context to fresh subagent turns — the runaway retry loop reported in #6263 (a single task burned ~6.09M input tokens in ~50 minutes across 81 duplicate turns before exhausting quota).

## Verified

- Failing-first regression test in `session-existence.test.ts`: a fixture client resolves only the bare id and 404s anything else. Before the fix, `opencode:ses_bare123` returned `"missing"` while `ses_bare123` returned `"exists"` (exact issue reproduction). After the fix both return `"exists"`, with the captured SDK path id asserted to be the bare form; codex:/senpi: prefixes covered; missing-behind-prefix still classifies `"missing"`.
- Scoped suite: `bun test packages/omo-opencode/src/features/background-agent/` → 746 pass / 0 fail across 59 files (all task-poller consumer paths).
- Repo typecheck: `bun run typecheck` → exit 0.
- Evidence recorded in `.omo/evidence/20260824-6263-session-prefix/EVIDENCE.md` (WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED).

## Risk

Low. The normalization is confined to the session-existence boundary named in the issue triage; raw ids are unaffected (the prefix pattern only matches a leading `opencode:`/`codex:`/`senpi:`). Other subsystems that consume prefixed ids against the SDK were intentionally left out of scope.

Fixes #6263

---
## Summary by cubic
Strip platform prefix from session IDs before existence checks so prefixed and raw IDs resolve identically. Previously we sent the prefixed ID to `client.session.get`, which 404’d and made the task poller misclassify live sessions as missing, triggering duplicate recovery turns (Fixes #6263).

- Adds `bareSessionId()` and uses it in `checkSessionExistence()` to remove `opencode:`/`codex:`/`senpi:`; normalization is limited to this boundary.
- Regression tests cover all prefixes: prefixed IDs now return "exists" like the bare ID; missing-behind-prefix still returns "missing".
- No migration. Risk is low and localized; other SDK callers are unchanged.

<sup>Written for commit dee15bfbc079605be264a4f77f8b4cba9217b5fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7221?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260824-6263-session-prefix/EVIDENCE.md
+++ b/.omo/evidence/20260824-6263-session-prefix/EVIDENCE.md
@@ -1,0 +1,31 @@
+# Evidence — 20260824 — issue #6263 session-existence prefix false negatives
+
+## WHAT WAS TESTED
+
+- Command: `bun test packages/omo-opencode/src/features/background-agent/session-existence.test.ts`
+- Surface: `checkSessionExistence()` in `packages/omo-opencode/src/features/background-agent/session-existence.ts` — the cheap existence check consumed by `task-poller.ts` (lines 238, 288) for background-task recovery decisions.
+- Behavior under test: a platform-prefixed session id (`opencode:` / `codex:` / `senpi:`, the format boulder.json stores via `normalizeSessionId` in `@oh-my-opencode/boulder-state`) must resolve to the same existence result as the raw id, because the storage layer resolves bare ids only.
+- Regression fixture: mock client resolves ONLY the bare id (`ses_bare123`) and returns a 404 "Session not found" error for anything else — mirroring the real storage layer's behavior described in issue #6263.
+
+## WHAT WAS OBSERVED
+
+- RED (before fix): prefixed lookups returned `"missing"` while raw lookup returned `"exists"`:
+  - `(fail) #given an opencode:-prefixed session id ... Expected: "exists" Received: "missing"` at session-existence.test.ts:74
+  - `(fail) #given codex:- or senpi:-prefixed session ids ... Expected: "exists" Received: "missing"` at session-existence.test.ts:88
+  - This reproduces the issue's production failure mode: live subagent sessions classified as gone every poll cycle, triggering recovery re-sends (the reported 6.09M-token runaway loop).
+- GREEN (after fix): `5 pass, 0 fail` in session-existence.test.ts; prefixed ids are stripped to bare form before `client.session.get({ path: { id } })` (asserted via captured requested ids), missing-behind-prefix still classifies `"missing"`.
+- Scoped suite: `bun test packages/omo-opencode/src/features/background-agent/` → `746 pass, 0 fail` across 59 files (includes all task-poller consumer tests).
+- Typecheck: `bun run typecheck` (tsgo --noEmit + typecheck:script + typecheck:packages) → exit 0.
+
+## WHY IT IS ENOUGH
+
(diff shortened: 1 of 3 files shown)


MoerAI:
Thanks for the PR, @AceRothstein71, and for the solid diagnosis and regression evidence.

I am closing this as superseded rather than rejected. #7265 carries the same `bareSessionId()` helper, platform-prefix pattern, and regression cases on an in-repository branch where the complete test, build, and compatibility matrix passes. The fix is correct; the fork workflow here never ran those substantive checks.

Credit for this work belongs to this PR and to @niStee's earlier #6264/#6430 lineage. #7265 is the replacement that will land the shared fix and close #6263.

PR #7224, opened 2026-08-24
fix(mcp-client-core): propagate configured request timeout to skill MCP SDK calls

### LINKED ISSUE #6631: [Bug]: Direct MCP callTool uses SDK default timeout because request options are not propagated

AceRothstein71:
## What

- `SkillMcpManager` now passes SDK request options on every direct MCP operation: `callTool`, `readResource`, `getPrompt`, `listTools`, `listResources`, and `listPrompts`.
- Timeout resolution is centralized in a new `request-options.ts` helper (`resolveRequestTimeoutMs` / `buildRequestOptions`) so future call sites cannot silently regress.
- New config surface: per-server `timeout` (milliseconds) on `ClaudeCodeMcpServer` (usable from skill SKILL.md frontmatter / mcp.json), plus a per-call `SkillMcpClientOptions.requestTimeoutMs` override. Precedence: per-call override > server config > SDK default.

## Why

Issue #6631: the direct tool-execution path called `client.callTool({ name, arguments })` without the third `options` argument, so the TypeScript SDK applied its default `DEFAULT_REQUEST_TIMEOUT_MSEC = 60000` and any tool running longer than ~60s failed even when the underlying server supported longer waits. The maintainer diagnosis confirmed the same omission applies to all sibling operations in the class and that no timeout config surface existed at all.

## Verified

- Failing-first regression test (`request-options.test.ts`): red before the fix (`Cannot find module './request-options'`), green after — 12 pass / 0 fail. Tests assert the exact argument position the SDK reads for each of the six call sites, override precedence, and that unconfigured servers still pass `undefined` (SDK default preserved).
- Scoped suites: `bun test packages/mcp-client-core` → 23 pass / 0 fail; consumer suites (omo-opencode `skill-mcp-manager` feature + `skill_mcp` tool + `claude-code-compat-core`) → 316 pass / 0 fail.
- `bun run typecheck` (tsgo root + script + all workspace packages) → exit 0.
- Evidence recorded under `.omo/evidence/20260824-6631-mcp-timeout/`.

## Risk

Low. Unconfigured servers behave exactly as before (SDK 60s default). The new `timeout` field is optional and additive; tier-2 `.mcp.json` transformer is intentionally untouched since those servers do not route through `SkillMcpManager`. Cancellation semantics beyond plain `timeout` (e.g. `resetTimeoutOnProgress`) are left for a follow-up if needed.

Fixes #6631

---
## Summary by cubic
Propagates request timeouts to all direct MCP SDK calls in `mcp-client-core`. Previously, `SkillMcpManager` omitted `RequestOptions`, so calls used the SDK’s 60s default; now calls honor a per-call override or per-server config, with unchanged behavior when unset.

- Affects `callTool`, `readResource`, `getPrompt`, `listTools`, `listResources`, and `listPrompts`; each now passes `buildRequestOptions(...)`.
- Adds `resolveRequestTimeoutMs`/`buildRequestOptions` helper to centralize timeout resolution.
- New config: optional `timeout` (ms) on `ClaudeCodeMcpServer` and per-call `requestTimeoutMs` on `SkillMcpClientOptions`. Precedence: per-call > server config > SDK default.
- No migration required. If you need >60s requests, set `timeout` in skill config or pass `requestTimeoutMs`.

<sup>Written for commit 2c3082725da2f9a699347cd6816780ea9733f338. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thanks for propagating the configured timeout across all six SDK calls. The confirmed blocker is that `resolveRequestTimeoutMs` forwards non-positive or non-finite values, so `timeout: 0` arms a zero-millisecond timer and makes every MCP call abort almost immediately. Please accept only finite positive numbers and otherwise return `undefined` so the SDK default applies. Add regression cases for zero, negative, and `NaN`, then provide passing mcp-client-core and consumer suites plus a clean typecheck. https://github.com/code-yeongyu/oh-my-openagent/pull/7224/files

PR #7230, opened 2026-08-24
fix(omo): short-circuit delegate-task sync poll when subagent session stalls

### LINKED ISSUE #6665: delegate-task poll waits full 30-min inactivity timeout when subagent dies with finish="unknown" empty messages
# ISSUE: delegate-task poll waits full 30-min inactivity timeout when subagent dies with `finish="unknown"` empty messages

**Plugin**: oh-my-openagent (opencode)
**Affected file**: `packages/omo-opencode/src/tools/delegate-task/sync-session-poller.ts`

## Symptom

A subagent (observed with `oracle` + a proprietary OpenAI-compatible gateway model) completes its tool work, then the model stream is interrupted. opencode records several **empty assistant messages with `finish: "unknown"`** (no text, no error). The session goes idle, but `pollSyncSession` does **not** detect completion and keeps polling until the 30-minute inactivity timeout (`MAX_POLL_TIME_MS`), blocking the parent agent the whole time.

**The key pain point: the subagent's work is actually done** (its deliverable text is already in the session), yet the parent agent only receives the result after waiting the full 30-minute inactivity timeout — an avoidable ~30-minute stall per occurrence for a task that completed on time.

## Real-world timeline (from `oh-my-opencode.log`)

```
07:05:03 [task] Starting poll loop  ses_... (oracle)
07:05:14..07:07:31  assistant msgs with finish="unknown", empty text (stream interrupted, retried 4x)
07:10:26  opencode: "exiting loop" (session idle, no more messages)
07:40:25 [task] Poll inactivity timeout reached  pollCount=2081   <-- 30 min later!
```

Parent session resumed only at 07:41:59. Same pattern hit 3 out of 4 oracle runs on 2026-08-09.

## Root cause

In `pollSyncSession` the completion check is `isSessionComplete()` (`sync-session-turns.ts`):

```ts
const NON_TERMINAL_FINISH_REASONS = new Set(["tool-calls", "unknown"])
```

`finish: "unknown"` is treated as non-terminal, so:
- `isSessionComplete()` returns `false` forever;
- the fallback branch (`!lastAssistant?.info?.finish && hasAssistantText`) is skipped because a `finish` value *exists* (it is just `"unknown"`);
- the session is already idle and will never produce new messages.

The poll loop has **no stall detection** — nothing distinguishes "subagent is still working" from "subagent is dead but the poller doesn't know", so it can only exit via the 30-minute inactivity timeout.

## Expected behavior

- A session that is inactive, whose last assistant message has a non-terminal `finish` of `"unknown"` (or missing), and that has produced no new messages for a bounded stall window (e.g. 30 s) should be resolved immediately:
  - if the session still contains a substantive text/reasoning deliverable → treat as **complete** and return the result;
  - otherwise → **fail fast** with a descriptive error instead of waiting out the inactivity timeout.

## Environment

- opencode desktop 1.18.15
- oh-my-openagent 4.19.0 (npm `oh-my-openagent@latest`)
- Provider: a custom OpenAI-compatible gateway; model: a gpt-class proprietary alias
- OS: Windows 11

AceRothstein71:
## What

- `pollSyncSession` (`packages/omo-opencode/src/tools/delegate-task/sync-session-poller.ts`) now detects a stalled subagent and resolves immediately instead of waiting out the full inactivity timeout (30 min by default).
- New exported predicate `isStallEligibleTurn()` in `sync-session-turns.ts`: the last assistant turn exists, its `finish` is `"unknown"` or missing, and it carries no pending tool parts.
- While the session is inactive, the poller tracks a transcript signature (`messages.length:lastMessageId`). If it stays frozen for a bounded stall window (new optional `stallWindowMs` input, default 30s, mirroring the existing `childWakeGraceMs` pattern; callers unchanged), it resolves:
  - substantive text/reasoning deliverable present -> normal completion path (`null`; caller fetches the result);
  - otherwise -> aborts the session (`stalled_subagent`) and fails fast with a descriptive error naming the session ID, window, and finish reason.

## Why

Issue #6665: when a subagent's model stream is interrupted mid-turn, opencode records empty assistant messages with `finish: "unknown"`. `isSessionComplete()` treats `"unknown"` as non-terminal forever, and the existing fallback only fires when `finish` is *missing*, so an idle session with a frozen transcript has no exit except `MAX_POLL_TIME_MS` (30 min). The reporter's timeline shows the subagent's deliverable was already in the session at 07:10 while the parent stayed blocked until 07:41. Same-subsystem prior art PR #7186 touched background-agent lookup layers only; this change is independent of those files.

Conservative by construction: active status (`busy`/`retry`/`running`), arriving messages, pending tool parts, and `"tool-calls"` finishes all keep the previous timeout-bounded behavior - only the exact dead-stream shape from the issue short-circuits.

## Verified

- Failing-first: new co-located suite `sync-session-poller.stall-detection.test.ts` ran before implementation - both regression tests failed exactly as the issue describes (loop burned the whole test timeout window and returned the inactivity-timeout string); baseline captured in evidence.
- After implementation: 6/6 pass in the new suite (fake-clock deterministic via the existing `withMockedDateNow` pattern), covering both resolution branches plus four guards (busy suppresses stall, growing transcript resets the timer, pending tools suppress stall, `"tool-calls"` stays guarded).
- Scoped suites: `bun test packages/omo-opencode/src/tools/delegate-task/` -> **493 pass / 0 fail** across 43 files; no existing test weakened or deleted.
- `bun run typecheck` (tsgo root + script + all workspace packages) exit 0. No `as any`, no `@ts-ignore`/`@ts-expect-error`.
- Evidence: `.omo/evidence/20260824-6665-delegate-poll-dead/` (plan, failing baseline, passing output, scoped suite, typecheck, QA notes). Live end-to-end harness QA omitted: no provider credentials in this environment; recorded in the QA notes with residual-risk analysis.

## Risk

Low / additive: a bounded early-exit path behind the same inactive-status gate that already existed; default 30s window only triggers on the frozen-transcript + non-terminal-finish shape that previously meant waiting 30 minutes for a known-dead poll. Worst case for a borderline-live session is one aborted sync attempt at ~30s idle instead of a 30-min hang; background delegation and continuation flows are untouched.

Fixes #6665

---
## Summary by cubic
Short-circuits the delegate-task sync poll when a subagent session stalls to prevent 30‑minute hangs. Previously, idle sessions with a frozen transcript and finish "unknown" only exited via the inactivity timeout; now, after a bounded stall window (30s default) the poll completes if a deliverable exists or aborts with "stalled_subagent" when it does not.

- Adds stall detection in `packages/omo-opencode/src/tools/delegate-task/sync-session-poller.ts`: tracks a transcript signature while inactive and resolves early once it stays unchanged for the stall window; active statuses or new messages reset the timer.
- Exports `isStallEligibleTurn()` in `sync-session-turns.ts`: last assistant exists; finish is missing or "unknown"; no pending tool parts. Turns with "tool-calls" or pending tools keep the previous timeout-bounded behavior.
- Extends `pollSyncSession` with optional `stallWindowMs` (default 30_000); existing callers unchanged. Early completion returns null (caller fetches the deliverable); otherwise the poll aborts the session with a descriptive error.
- Addresses #6665 by unblocking parents when a subagent stream dies mid-turn or the deliverable is already present. Risk is low and scoped to inactive, frozen-transcript cases.

<sup>Written for commit 42502cef850b3fa72c57d8bb02e34d03a2345ea4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7230?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Closing as an unmergeable duplicate of #6666, not as a resolution of #6665.

This head conflicts with live `dev`, never ran the required matrix, and can falsely accept historical assistant text as the current turn's deliverable. The canonical path in #6666 already scopes progress to the current turn and exposes deterministic clock and wait seams.

Keep #6665 open. Rebase and finish one successor from #6666 with explicit idle-state gating, in-place transcript-change detection, stale-result and status-outage coverage, and real isolated OpenCode QA. Reopen this PR only if it carries a scenario the canonical branch cannot preserve.

PR #7233, opened 2026-08-24
fix: stop sending the removed 'minimal' reasoning effort to gpt-5.6

junhoyeo:
Memory reflection failed three runs in a row on this machine. Two independent causes; one commit each.

## 1. The failure itself — `gpt-5.6` rejects `minimal`

Verbatim from `runtime/reflection/runs/reflection-run-{1,2,3}/child-stderr.log`:

```
OpenAI API error (400): {"message":"Unsupported value: 'minimal' is not supported
with the 'gpt-5.6-luna' model. Supported values are: 'none', 'low', 'medium',
'high', 'xhigh', and 'max'.","type":"invalid_request_error",
"param":"reasoning.effort","code":"unsupported_value"}
```

`HEURISTIC_MODEL_FAMILY_REGISTRY` had one `gpt-5` entry (`includes: ["gpt-5"]`) advertising `reasoningEfforts: [..., "minimal", ...]`. gpt-5.6 removed that value and the API rejects it outright instead of degrading, so every reflection against a 5.6 model died before doing any work.

Adds a narrower `gpt-5-6-plus` family ahead of the generic entry (first match wins) that drops `minimal` and aliases it to `none`.

One detail that would make a naive fix silently fail: **model IDs are normalized before family matching**, so `gpt-5.6-luna` arrives as `gpt-5-6-luna`. A dotted pattern never matches.

Verified across the family boundary:

| model | family | `minimal` resolves to |
|---|---|---|
| `gpt-5.6-luna` | `gpt-5-6-plus` | `none` |
| `gpt-5.6-terra` | `gpt-5-6-plus` | `none` |
| `gpt-5.10-x` | `gpt-5-6-plus` | `none` |
| `gpt-5` | `gpt-5` | `minimal` |
| `gpt-5.1` | `gpt-5` | `minimal` |
| `gpt-5.2-codex` | `gpt-5` | `minimal` |

Also routes provider 400s on `reasoning.effort` to a hint naming the rejected parameter. They previously fell through to "inspect child-stderr.log" — and that log contains nothing but the same line.

## 2. The remediation hint pointed at a directory nothing writes

The default post-spawn hint named `runtime/reflection-sessions/<runId>/child-stderr.log`. `reflection-spawn-input.ts` builds the run dir from `paths.reflection` + `"runs"`, so logs land in `runtime/reflection/runs/<runId>/`. `reflectionSessions` is a vestigial entry in the identity layout that no writer uses.

**Attribution:** this fix and its test were **not written by me**. I found them already present, uncommitted, in a local working checkout while diagnosing the 400. They are correct, `HEAD` still has the stale string, and the shipped `plugin/extensions/omo.js` bundle still contains it — so the warning keeps misdirecting until this lands and the bundle is rebuilt. Committed separately, with attribution, so the work is not lost. Happy to drop that commit if its author would rather submit it themselves.

## Tests

91 pass, 0 fail (`bun test`) — 4 new regression tests: 3 covering the gpt-5.6/gpt-5 family split, 1 pinning the 400 remediation to verbatim stderr.

---
## Summary by cubic
Stops reflection failures on `gpt-5.6` by clamping unsupported reasoning effort and honoring provider capability metadata. Previously we sent "minimal" to `gpt-5.6`, got a 400, and aborted; now we map "minimal"→"none" for `gpt-5.6`, clamp the launch path to model capabilities, and scope remediation hints correctly.

- Adds a `gpt-5-6-plus` family before `gpt-5` in `packages/model-core`; accepts "none", "low", "medium", "high", "xhigh", "max"; aliases "minimal"→"none"; matches normalized hyphenated IDs.
- Clamps reflection "thinking" via `resolveCompatibleModelSettings` with `getModelCapabilities` across category resolution, resolved fallbacks, beyond-category, session inherit, and configured fallbacks; unknown families keep the caller’s level; explicit provider capabilities outrank family aliases.
- Remediation: maps provider 400s on `reasoning.effort` to a hint naming `categories.<category>.reasoning`; non-reasoning `unsupported_value` errors fall back to the child-log hint; fixes the default hint path to `runtime/reflection/runs/<runId>/child-stderr.log`.
- Updates tests and rebuilds the `packages/omo-senpi/plugin/extensions/omo.js` bundle.

<sup>Written for commit d76e4972e4b6da969df3dca0fe8336ad76c82082. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7233?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo:
Pushed three fixes addressing all three review threads. Thanks for the P1s - the first one was correct and my original patch would not have fixed the reported failure.

## P1: clamp in the Senpi reflection launch path

You were right. `resolveCompatibleModelSettings` was never called anywhere in `omo-senpi`, so the model-core family change alone never reached the failure. `resolve-model.ts` took the configured level, ran it through `normalizeThinking` (which only maps `none`->`off` and validates the ladder), and handed it to the child unchanged.

Clamped at all five sites that emit a thinking level, not just the primary: category resolution, its resolved fallbacks, the beyond-category ladder, session inherit, and configured fallbacks. Session inherit is worth calling out - that level was chosen for the **parent** model, which need not be the model the reflection child runs.

Unknown families preserve the caller's level rather than dropping it. The QA driver's own config pins `omo-mock/mock-1` with `reasoning: "minimal"`, so silently dropping it there would have broken the driver; covered by a test.

Confirmed the new test fails without the clamp:

```
✗ #given a gpt-5.6 model configured with the removed minimal effort
  #when resolved #then the launch path clamps it
  Expected: not "minimal"
```

## P2: provider metadata now outranks the family alias

Also correct, and it was an ordering bug in `resolveField` generally: the alias was consulted before `metadataOverride`, so any provider publishing its own effort list could still be downgraded by a family alias. Moved the explicit-metadata check first:

```ts
if (metadataOverride?.includes(normalized)) return { value: normalized }
```

That fixes `github-copilot/gpt-5.6-*` and also the pre-existing `glm` / `deepseek` aliases.

## P1: QA evidence

Added at `.omo/evidence/omo-senpi-adapter/20260825-gpt56-reasoning-effort/`.

Reporting honestly: **the live `memory-model-fallback-e2e` driver FAILS here** with `timed out waiting for reflection completion`. Per the senpi-qa golden rules that is a FAIL, not a pass, and the README says so.

It is not attributable to this change. Re-ran the same driver in the same sandbox with the clamp neutered to a pass-through, i.e. exactly the pre-change behaviour:

| run | behaviour | result | elapsed |
|---|---|---|---|
| with clamp | this change | timeout | 98.2s |
| clamp neutered | pre-change | timeout | 95.3s |

Both logs are committed alongside the README.

## Bundle

`senpi-compatibility` was failing on `omo-senpi extension build is not current: stale-output`. Regenerated `plugin/extensions/omo.js` (+2/-2 minified lines). Worth noting the committed bundle still carried the stale `reflection-sessions/<runId>` hint - it now carries `reflection/runs/<runId>`, 0 stale occurrences.

109 unit tests pass.

One unrelated observation: `bun install` runs a `prepare` -> `build` that fails at `build:declarations` with pre-existing `Signals` / `"memoryPressure"` TS2345 errors across `mcp-client-core` and `omo-opencode`, in files untouched here. It also rewrites `packages/omo-codex/scripts/install-dist/install-local.mjs` (+249/-120). I left that out of this PR.

junhoyeo:
Both P2s fixed in 28c4970. Thanks - the capabilities one was a real hole in my own fix.

## Pass provider capabilities into the Senpi clamp

Correct. I was calling `resolveCompatibleModelSettings` with no `capabilities`, so the `gpt-5-6-plus` alias applied even to providers that still accept `minimal`. Now calls `getModelCapabilities({ providerID, modelID })` and passes it through, which picks up the provider override.

Measured for `github-copilot`, whose override publishes `["none","minimal","low","medium","high"]` for gpt-5.6:

| model | with capabilities | without (previous) |
|---|---|---|
| `github-copilot/gpt-5.6` | `minimal` | `none` (wrong) |
| `github-copilot/gpt-5.6-luna` | `minimal` | `none` (wrong) |
| `openai/gpt-5.6-luna` | `none` | `none` |

So it was silently weakening a level Copilot accepts, while the openai case (the actual bug) is unaffected. Covered by a test.

Combined with the `resolveField` reordering, an explicit capability list now outranks a family alias generally, which also covers the pre-existing `glm` and `deepseek` aliases.

## Point the remediation at a valid config field

Also correct - `memory.reflection.reasoningEffort` does not exist. `OmoMemoryReflectionSchema` is `enabled` / `merge` / `category` / `timeout_minutes` / `sandbox`. The level lives on the category, so the hint now names `categories.<category>.reasoning` and says which category is in play.

## Live QA gate

Still honestly a FAIL, and I do not want to dress it up. `memory-model-fallback-e2e` times out at `waitForCompletion` in this sandbox, and it times out identically with the clamp neutered to a pass-through (98.2s vs 95.3s), so it is not attributable to this change - both logs are committed. The driver's own config pins `omo-mock/mock-1` with `reasoning: "minimal"`; that is an unknown family, the clamp preserves the caller's level there, and there is a test asserting it, so the change should be inert for that driver.

If the timeout is a known-flaky driver or needs an env I have not set, tell me which and I will re-run it and replace the evidence with a green one.

110 unit tests pass. Bundle regenerated; 0 stale `reflection-sessions` occurrences.

junhoyeo:
Green on d76e497 — 24/24.

## Restrict unsupported-value remediation to reasoning errors

Fixed. `combined.includes("unsupported_value")` matched any rejected parameter, so a 400 on `temperature` or `max_tokens` would have been answered with "change the reasoning level". Now requires the reasoning parameter to be named; everything else falls through to the child-log hint as before. Test added asserting a `temperature` rejection does not mention reasoning.

## On the previous red run

`senpi-compatibility` failed on `task RPC launch profile parity > #given an explicit provider extension #when process model admission runs #then its model is visible without credentials` (6072ms). That was a flake, not this branch:

- passes locally 3/3 (2446ms)
- the test has zero references to reasoning, thinking or effort; this branch only touches `resolve-model.ts`, `remediation.ts` and model-core
- the same job is green on the next push with nothing in that area changed

The earlier red on that job was genuine and different — `omo-senpi extension build is not current: stale-output`, fixed by regenerating the bundle.

## Remaining thread: the live QA gate

Still a FAIL and still recorded as one. To be concrete about what I ruled out: the driver times out at `waitForCompletion` identically with the clamp neutered to a pass-through (98.2s vs 95.3s, both logs committed), and the driver's own `omo-mock/mock-1` + `minimal` config is unknown-family, which the clamp preserves — asserted by a test. So I have evidence it is not attributable, but not a green run.

I cannot turn that into a pass without knowing whether the driver needs an env or provider I have not set. If it is known-flaky, say so and I will note it; if it needs specific setup, tell me which and I will re-run and replace the evidence.

Summary of the four commits:
1. `67db105` stale `reflection-sessions` hint (pre-existing uncommitted work, attributed)
2. `63b18a6` gpt-5-6-plus family
3. `e668ce4` launch-path clamp at all five emit sites + evidence + bundle
4. `28c4970` provider capabilities into the clamp + real config field
5. `d76e497` scope the unsupported-value hint

111 unit tests pass.

MoerAI (changes requested):
Please address these blockers before reconsideration:

1. Rebase onto current `dev`, preserve the current sandbox remediation, regenerate `plugin/extensions/omo.js`, and resolve every outstanding review thread.
2. Make `memory-model-fallback-e2e` pass in isolation. The committed evidence currently records a failure: https://github.com/code-yeongyu/oh-my-openagent/blob/d76e4972e4b6da969df3dca0fe8336ad76c82082/.omo/evidence/omo-senpi-adapter/20260825-gpt56-reasoning-effort/README.md
3. Add isolated OpenCode `chat.params` QA proving OpenAI GPT-5.6 maps `minimal` away from the wire while an explicit Copilot `minimal` capability remains unchanged.
4. Scope remediation to the parsed provider error `param: reasoning.effort`. A `param: temperature` error whose message merely mentions reasoning must not receive reasoning-effort advice.
5. Cover every reflection candidate lane affected by the clamp, then rerun the full required matrix.

The shared resolver feeds both chat params and fallback generation:
- https://github.com/code-yeongyu/oh-my-openagent/blob/d76e4972e4b6da969df3dca0fe8336ad76c82082/packages/omo-opencode/src/plugin/chat-params.ts#L113-L148
- https://github.com/code-yeongyu/oh-my-openagent/blob/d76e4972e4b6da969df3dca0fe8336ad76c82082/packages/omo-opencode/src/cli/model-fallback.ts#L36-L62

PR #7234, opened 2026-08-25
fix(omo-opencode): read UserPromptSubmit hook stdout as control output

CrazyCoder:
## Summary

UserPromptSubmit hook stdout reaches the model unparsed, so a hook that writes a control object sends that JSON to the model as conversation content.

Claude Code reads stdout as control output when the trimmed text starts with `{`. That is `parseHookOutput` in `src/utils/hooks.ts`. Anything else is plain text, and plain text from a UserPromptSubmit hook becomes context.

A control output reaches the model only through `hookSpecificOutput.additionalContext`. `processHookJSONOutput` gives the success attachment an empty content string for that reason, and `processUserInput` skips empty content. A hook that writes `{"continue":true}` therefore injects nothing.

`user-prompt-submit.ts` instead wraps the raw stdout in `<user-prompt-submit-hook>` tags and pushes it. It parses the JSON only to look for `decision: "block"`, and only when the hook exits non-zero. The four sibling handlers in the same directory parse first and fall back to the raw text: `pre-tool-use.ts`, `post-tool-use.ts`, `stop.ts` and `pre-compact.ts`. UserPromptSubmit is the exception.

### What the injected block breaks

The block is present on the turn that submits the prompt and absent from the next turn. Message 0 goes from two text blocks to one in the middle of a session.

That is a change to the cached prefix, so the next turn reads through to a cold prompt cache. It repeats on every turn, because the hook fires on every prompt.

Anything that identifies a conversation by its prefix also loses it. I measured this against a Claude Max proxy with an OpenCode client. 38 requests produced 0 resumed sessions, and each one classified the whole conversation as unrelated history. The `pi` and `prime` clients on the same proxy resumed normally over the same period. Setting `claude_code.hooks: false` restored resumption on the next turn, which is what makes the injected block the cause rather than a correlate.

The emitter in my case is peon-ping, which writes `{"continue":true}` on its pass-through path in `scripts/hook-handle-use.ps1`. Any hook that reports "nothing to do" with a control object does the same.

## Changes

`packages/omo-opencode/src/hooks/claude-code-hooks/user-prompt-submit.ts`

- `parseControlOutput()` reads stdout as control output when it starts with `{` and parses. Context then comes from `hookSpecificOutput.additionalContext` alone.
- `decision: "block"` and `continue: false` stop the prompt whatever the hook exited with, which matches `processHookJSONOutput`. A hook that sets both reports `reason`, because `processUserInput` checks the blocking error ahead of the stop reason.
- Stdout that is not a control object is still injected verbatim. Claude Code treats malformed JSON as a hook error instead. I chose the sibling handlers' behavior here, so no output a hook writes today is silently dropped.
- `asHookContext()` holds the tag wrapping that both paths now share.

`packages/omo-opencode/src/hooks/claude-code-hooks/types.ts`

- `UserPromptSubmitOutput` replaces a cast to `PostToolUseOutput`, whose `hookSpecificOutput.hookEventName` is `"PostToolUse"`.

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/hooks/claude-code-hooks/`
  **Observed result:** 126 pass, 0 fail, with 10 new tests. Three cover the parse decision: plain text holding a brace, valid JSON that is not an object, and malformed JSON. Four cover the stop fields, alone and together. Three cover what gets injected: `{"continue":true}`, `additionalContext`, and absent stdout.
  **Artifact:** none saved. The suite runs in under a second from a clean checkout.
  **Why sufficient:** the suite already mocks `dispatchHook` at the module boundary. Every branch of the new parse is reachable without a live OpenCode session.

- **What was tested:** each new guard removed from the source in turn, then the same suite re-run.
  **Observed result:** every guard has at least one test that notices.

  | Guard removed | Tests that fail |
  |---|---|
  | Leading-brace sniff | 1 |
  | Control output preferred over raw stdout | 3 |
  | Blocking error ahead of stop reason | 1 |
  | Undefined stdout | 1 |

  **Artifact:** none saved. Each mutation is a single edit to `user-prompt-submit.ts`.
  **Why sufficient:** it shows the assertions compare real differing values rather than passing by construction. The first row needed a test written for it. A brace inside plain text fails to parse with or without the sniff, so it separates nothing. Valid JSON that is not an object does.

- **What was tested:** `bun run typecheck`
  **Observed result:** no errors across all 30 package projects.
  **Artifact:** none saved.
  **Why sufficient:** the change adds an exported interface and removes a cast, so type resolution is the risk it carries.

## Risks & Residuals

- A hook that writes `decision: "block"` and exits 0 now blocks the prompt, where before it was injected as text and ignored. This matches `processHookJSONOutput`, which reads the field whatever the exit code is. Accepted, and covered by a test.
- A hook that writes `continue: false` now blocks the prompt. Claude Code keeps the original prompt in context there and this bridge does not, because `block: true` is the only stop it has. Accepted. The alternative is to drop the field in silence, which is worse.
- Exit code 2 is unchanged. Claude Code blocks the prompt there and shows stderr to the user only, while this bridge injects stdout on any exit code. Not addressed, and separate from this defect.
- `hookSpecificOutput.hookEventName` is not checked against `"UserPromptSubmit"`. Claude Code throws on a mismatch. Accepted, since reading the context anyway is the more forgiving behavior and this bridge has no hook-error channel.
- `bun test` was not run whole. The suite covers far more than this bridge and I have no baseline for it here. A pre-existing failure would carry no signal about this change.

## Related Issues

I found no upstream issue that reports this.

---
## Summary by cubic
Treats `UserPromptSubmit` hook stdout as control output when it starts with `{`, and injects only `hookSpecificOutput.additionalContext`. This removes accidental JSON injection that changed the conversation prefix and broke prompt-cache resumption and session continuation.

- Adds control-output parsing consistent with the sibling hook handlers; non-control stdout is still injected verbatim, and `UserPromptSubmitOutput` replaces the `PostToolUseOutput` cast.

**Behavior changes and migration**
- `decision: "block"` and `continue: false` now block the prompt regardless of exit code; `reason` takes precedence over `stopReason`.
- Hooks that previously relied on raw control JSON appearing in the conversation must write intended text to `hookSpecificOutput.additionalContext` instead.

<sup>Written for commit e473301f3a5d12948cbd3fe6d10a355974846aff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


CrazyCoder:

PR #7239, opened 2026-08-25
fix(omo): promote per-agent fallback_models for the main session

### LINKED ISSUE #7226: [Bug]: fallback_models not applied to main/interactive TUI session on startup (only delegated task() calls)
### Summary
`fallback_models` configured per-agent in `oh-my-openagent.json` does not protect the **main interactive TUI session** on startup. When the primary model for a top-level agent (e.g. `sisyphus`) is unavailable/erroring, the session does not automatically retry through the configured `fallback_models` chain — I have to manually switch models via the in-TUI model picker every single time I start `opencode`.

### Config (sanitized excerpt)
```json
{
  "agents": {
    "sisyphus": {
      "model": "opencode/claude-opus-4-8",
      "variant": "max",
      "fallback_models": [
        { "model": "github-copilot/claude-sonnet-5", "variant": "max" },
        { "model": "opencode/kimi-k2.5" },
        { "model": "github-copilot/gpt-5.4", "variant": "medium" },
        { "model": "opencode/gpt-5.5", "variant": "medium" },
        { "model": "opencode/glm-5" },
        { "model": "opencode/big-pickle" }
      ]
    }
  }
}
```

### Observed behavior
- Every `opencode` startup where the primary model (`opencode/claude-opus-4-8`, variant `max`) is unreachable, the session does not fall through to any of the configured `fallback_models` entries.
- I have to manually re-pick a model in the TUI each time.
- This is distinct from delegated `task()`/subagent calls, where fallback behavior does appear to be wired up (`resolveModelForDelegateTask` in `dist/index.js`).

### Investigation
Digging through the bundled plugin (`dist/index.js`, `dist/tui.js`) suggests `fallback_models` resolution (`resolveModelForDelegateTask`, `resolveCategoryExecution`) is only invoked for delegated/background task dispatch, not for the interactive/main session's model resolution at startup. The TUI init path appears to use a separate, effectively hardcoded chain rather than reading each agent's `fallback_models`.

This may be related in spirit to:
- #2304 (fallback_models never reached due to credential detection gap)
- #6637 (runtime_fallback loops, never switches)
- #6751 (runtime_fallback resets fallback position mid-cycle)
- #7161 (model_fallback doesn't cycle on provider limit errors)

...but none of those exactly cover "main/interactive session never consults per-agent `fallback_models` at all," so filing separately.

### Expected behavior
When the primary model for the main session agent (sisyphus/prometheus/etc.) fails to initialize or errors during the session, the runtime should walk the agent's configured `fallback_models` chain the same way delegated tasks do, instead of requiring a manual model switch every startup.

### Environment
- Plugin: `oh-my-openagent@latest` (via `opencode` plugin loader)
- Config: `~/.config/opencode/oh-my-openagent.json`
- OS: macOS (darwin, arm64)

### Ask
Could you confirm whether main-session fallback is intentionally out of scope for `fallback_models`, and if not, whether there's a fix planned (possibly tied to the `runtime_fallback` work referenced in #6637/#6751/#7161)?

AceRothstein71:
## What

The main interactive TUI session now honors an agent's configured `fallback_models` proactively. A new resolver (`packages/omo-opencode/src/plugin/chat-message/main-session-fallback.ts`) is wired into the plugin's `chat.message` handler: when the requested primary model is absent from warm availability data, the first reachable entry of that agent's `fallback_models` (own config, or inherited via its category) overrides `output.message.model` for the request, including the entry's variant.

It reuses the exact delegate resolution primitive (`resolveModelForDelegateTask` from `@oh-my-opencode/delegate-core`) that delegated `task()` calls already use, so main-session and subagent behavior can no longer drift.

Conservative guards:
- Delegated subagent sessions are skipped (the delegate path already owns them).
- Agents without configured `fallback_models` are untouched (no hardcoded chains are introduced for primary agents; that remains the separate default-off model-fallback system).
- Cold/empty availability data defers without guessing (mirrors the delegate cold-cache sentinel).
- Promotion only happens when the delegate resolver reports `matchedFallback`.
- `input` is never mutated, so the reactive runtime-fallback state machine keeps seeing the originally requested model.
- One toast per promoted model per session (deduped), so repeated messages do not spam notifications.

## Why

Per-agent `fallback_models` were consumed in only two places:
1. delegated dispatch (`tools/delegate-task/subagent-model-resolution.ts` -> `resolveModelForDelegateTask`, which promotes a reachable fallback when the user primary is unreachable), and
2. the reactive runtime fallback (`hooks/runtime-fallback/fallback-models.ts`), which is gated behind `runtime_fallback.enabled` (default off) and only fires after a retryable provider error.

The main session's `chat.message` path applied only the stored-model replay (`plugin/chat-message/session-model.ts`). Result: at startup with an unavailable primary, the main session never walked the configured chain and required a manual model switch every time, while delegated subagents sailed through the same outage. See #7226.

## Verified

Failing-first regression suite `packages/omo-opencode/src/plugin/chat-message/main-session-fallback.test.ts` (given/when/then):
- RED (6 pass / 4 fail) against a stub reproducing pre-fix behavior, proving the tests detect the missing promotion.
- GREEN (10 pass / 0 fail) after implementation, including the issue scenario: sisyphus with `fallback_models: [{github-copilot/claude-sonnet-5, max}, {opencode/kimi-k2.5}]` and an unavailable `opencode/claude-opus-4-8` promotes `github-copilot/claude-sonnet-5` with variant `max`.

Scoped suites around every touched import edge:
- `bun test packages/omo-opencode/src/plugin/chat-message packages/omo-opencode/src/hooks/runtime-fallback packages/omo-opencode/src/tools/delegate-task`: 818 pass / 0 fail (74 files).
- `tsgo --noEmit -p packages/omo-opencode/tsconfig.json`: exit 0.

Full QA evidence (WHAT WAS TESTED / OBSERVED / WHY IT IS ENOUGH / OMITTED, red + green captures): `.omo/evidence/20260824-7226-main-fallback/`.

## Risk

Low. The override lands through the same `output.message.model` mutation channel already used by stored-model replay and ultrawork overrides. Worst case for a mis-detected environment is a one-time switch to a user-configured fallback model with a visible toast; reachable primaries, unconfigured agents, subagent sessions, and cold caches are all no-ops. The reactive runtime-fallback system is untouched.

Fixes #7226

---
## Summary by cubic
Main TUI session now promotes an agent’s configured `fallback_models` when the requested primary model is unavailable, matching delegated subagent behavior. Previously the main session ignored the chain and required a manual switch; now it proactively selects the first reachable fallback using the same resolver, and updates `output.message.model` (including variant).

- Introduces `packages/omo-opencode/src/plugin/chat-message/main-session-fallback.ts` and wires `applyMainSessionFallbackOverride` into `packages/omo-opencode/src/plugin/chat-message.ts` after stored-model replay and before hooks.
- Reuses `resolveModelForDelegateTask` with `getAvailableModelsForDelegateTask`; only promotes on warm availability and when `matchedFallback` is true.
- Mutates only `output.message.model` and optional `output.message.variant`; never mutates `input`.
- Skips delegated subagent sessions, agents without `fallback_models` (own or category), and cold/empty availability sets; shows a single deduped toast per session/model.
- Adds `packages/omo-opencode/src/plugin/chat-message/main-session-fallback.test.ts`; scoped suites and typecheck pass. Fixes #7226.

<sup>Written for commit d5d040ecb89362e0ee95ab0562aee6b0e5bf7fc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thank you for the contribution. The confirmed blocking gate is that the primary commit lacks GitHub attribution. Please recreate or amend the commit with valid GitHub-linked author attribution, then provide the updated commit metadata and passing required CI as regression proof. Files: https://github.com/code-yeongyu/oh-my-openagent/pull/7239/files

PR #7243, opened 2026-08-25
fix(omo-opencode): stop stale provider reasoningEffort from overriding variant-routed session reasoning

### LINKED ISSUE #6614: [Bug]: Configured reasoning reaches chat.params but stale provider options keep medium

AceRothstein71:
## What

- `applySessionPromptParams` now records a `reasoningViaVariant` flag on the session prompt params whenever an explicit user reasoning level is routed to a model's variant channel (i.e. `lowerReasoningForModel` returned a variant and no `reasoningEffort`).
- The `chat.params` handler strips any stale `output.options.reasoningEffort` before compatibility resolution when that flag is set, so the requested variant reaches the provider uncontested.
- Two failing-first regression tests pin both halves of the contract (intent recording at prompt time, stale-effort enforcement at request time).

## Why

For variant-capable models, `lowerReasoningForModel("xhigh")` returns `{ variant: "xhigh" }` with no `reasoningEffort`, so `applySessionPromptParams` stored an empty options object. The stored-options merge in `chat.params` therefore never ran, and OpenCode's pre-populated default effort (e.g. a lingering `"medium"`) survived into the serialized provider request alongside the correct `message.variant`. Depending on provider precedence the stale effort wins, so the user's requested reasoning level never actually reached the provider.

## Verified

- RED first: exactly the two new #6614 tests failed while all 8 pre-existing tests in both files passed (`.omo/evidence/20260824-6614-reasoning-override/red-run.txt`).
- GREEN after fix: 13 pass / 0 fail across session-prompt-params helpers/state + chat-params (`green-run.txt`).
- Scoped suite over every consumer of the path (delegate-task sync sender, call-omo-agent sync executor, team-mode member routing, background-agent spawner/manager, plugin-interface chat.params wiring): 1376 pass / 0 fail across 119 files (`scoped-run.txt`).
- `bun run typecheck` exit 0: tsgo root + script + all workspace packages including `packages/omo-opencode` (`typecheck.txt`).
- QA evidence recorded under `.omo/evidence/20260824-6614-reasoning-override/` (WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED). Live-provider serialization was not driven (no credentialed provider in this environment); the test asserts the exact `output.options` and `message.variant` values OpenCode serializes.

## Risk

Low. Behavior changes only for sessions where explicit reasoning was routed to the variant channel; sessions with a stored `reasoningEffort` option keep the existing merge semantics (covered by pre-existing tests). Non-variant models and legacy `reasoningEffort` input are unaffected.

Fixes #6614

---
## Summary by cubic
Stops stale provider reasoningEffort from overriding variant-routed session reasoning. Previously, `@omo-opencode` could send both `message.variant` and a lingering `output.options.reasoningEffort`, letting the stale effort win; now we mark variant-routed reasoning and strip the stale effort so the requested variant reaches the provider. Fixes #6614.

- Records `reasoningViaVariant` when `lowerReasoningForModel` routes an explicit user reasoning level to a model variant and returns no `reasoningEffort` (`packages/omo-opencode/src/shared/session-prompt-params-helpers.ts`, `...-state.ts`).
- `createChatParamsHandler` removes `output.options.reasoningEffort` when that flag is set, before capability/compat resolution (`packages/omo-opencode/src/plugin/chat-params.ts`).
- Adds two regression tests pinning intent recording and stale-effort enforcement; sessions with stored `reasoningEffort` keep existing merge semantics. No migration required; risk is low and non-variant models are unaffected.

<sup>Written for commit 7f19fc27c5cc7bcbeeedc86cd4a733e7aa8c110b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thank you for the fix. The confirmed blocker is that the PR includes irrelevant failed-install evidence containing absolute paths and needs rebasing. Please remove that evidence, rebase onto the current dev branch, and provide passing required CI and focused regression results on the rebased head. Files: https://github.com/code-yeongyu/oh-my-openagent/pull/7243/files

PR #6687, opened 2026-08-10
docs(omo-native): explain legacy omo path recovery (LAB-39)

eddieparc:
https://linear.app/jgplabs/issue/LAB-39/senpi-기반-omo-ai-beta가-legacy-omo-cli를-실행함

## Summary

- explain why a successful `omo-ai@beta` install can still invoke legacy OMO
- cover both shell command caching and `~/.local/bin` PATH precedence
- provide a verified current-shell recovery and permanent PATH guidance

Closes #6686.
Cross-reference: https://github.com/code-yeongyu/senpi/issues/777

## Verification

- `git diff --check`
- `bun test packages/omo-native/test`
- manual zsh QA: legacy 4.19.4 -> `export PATH="$(npm prefix -g)/bin:$PATH"` + `hash -r` -> omo-ai beta 5.0.0-0.beta.3 / senpi 2026.8.9-2
- `omo --help` exits 0 and shows the senpi CLI
- `omo --definitely-invalid` exits 1 with `Unknown option`

No code behavior changes; this documents the shell behavior npm cannot mutate from a child process.

---
## Summary by cubic
Adds docs explaining why `omo` can still run the legacy CLI after installing `omo-ai@beta`, with steps to fix PATH order and shell command caching so the senpi CLI runs. Addresses Linear LAB-39.

- **Migration**
  - Cause: shell command caching and `~/.local/bin` preceding npm's global bin.
  - Quick fix: `export PATH="$(npm prefix -g)/bin:$PATH"` then `hash -r` (or `rehash` on zsh); verify with `command -v omo` and `omo --version`.
  - Permanent fix: put npm's global bin first in your shell startup file.

<sup>Written for commit de87297d38aaa638ae5beccd3948ca927766f8bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/6687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #6680, opened 2026-08-09
omo-ai@beta: bundled omo.js extension fails to load and stale local plugin path causes skill collisions


ashmoonori-afk:
## Summary

After installing `omo-ai@beta` globally and launching `omo` in a tmux session, OMO starts successfully but reports an extension load failure for the bundled `omo.js` extension. It also reports skill-name collisions against an older local `~/oh-my-openagent/packages/omo-senpi/plugin` path.

## Environment

- OS: macOS
- Install command: `npm install -g omo-ai@beta`
- Global package shown by npm: `omo-ai@5.0.0-0.beta.1`
- CLI command: `omo`
- `omo --version`: `2026.8.9-2`
- Launch method: tmux session from `$HOME`

## Reproduction

```sh
npm uninstall -g @code-yeongyu/senpi || true
rm -rf ~/senpi
npm install -g omo-ai@beta
tmux new-session -d -s omo -x 160 -y 48 'cd ~ && exec omo'
tmux capture-pane -t omo -p -S -200
```

## Actual output

At startup:

```text
omo (omo-ai beta 5.0.0-0.beta.1)
Warning: Failed to load extension "/opt/homebrew/lib/node_modules/omo-ai/plugin/extensions/omo.js": Failed to load extension: The "path" argument must be of type string. Received undefined
```

Later under `[Extension issues]`:

```text
/opt/homebrew/lib/node_modules/omo-ai/plugin/extensions/omo.js
  Failed to load extension: The "path" argument must be of type string. Received undefined
```

Skill conflict examples:

```text
[Skill conflicts]
  "ast-grep" collision:
    ✓ path (temp) /opt/homebrew/lib/node_modules/omo-ai/plugin/skills/ast-grep/SKILL.md
    ✗ ~/oh-my-openagent/packages/omo-senpi/plugin/skills/ast-grep/SKILL.md (skipped)
  "coding-agent-sessions" collision:
    ✓ path (temp) /opt/homebrew/lib/node_modules/omo-ai/plugin/skills/coding-agent-sessions/SKILL.md
    ✗ ~/oh-my-openagent/packages/omo-senpi/plugin/skills/coding-agent-sessions/SKILL.md (skipped)
  "visual-qa" collision:
    ✓ path (temp) /opt/homebrew/lib/node_modules/omo-ai/plugin/skills/visual-qa/SKILL.md
[...]

PR #7246, opened 2026-08-25
fix(omo-native): detect stale omo binaries on PATH in doctor

### LINKED ISSUE #6686: [Bug]: omo-ai beta may invoke cached or higher-priority legacy omo
## Prerequisites

- [x] I searched existing issues and found no duplicate.
- [x] I am using the latest `omo-ai@beta` (`5.0.0-0.beta.3`).

## Description

Installing `omo-ai@beta` globally can leave an already-running zsh session invoking an older OMO wrapper even though the new npm binary has higher PATH precedence. This makes a successful beta install appear broken for users migrating from legacy OMO.

## Steps to reproduce

1. Have the legacy managed wrapper at `~/.local/bin/omo` and invoke it once in zsh so it is cached.
2. Ensure the npm global bin directory appears before `~/.local/bin` in PATH.
3. Run `npm i -g omo-ai@beta` in the same shell.
4. Run `omo`.

## Actual behavior

The cached legacy wrapper runs and prints:

```text
Usage: oh-my-opencode [options] [command]
The ultimate OpenCode plugin - multi-model orchestration, LSP tools, and more
```

In the reproduced environment:

```text
cached=/Users/jgp/.local/bin/omo
omo --version => 4.19.4
```

`which -a omo` still lists the newly installed npm binary first, which makes this especially confusing.

## Expected behavior

After the documented global install, users should get the senpi-based OMO launcher or receive an actionable post-install diagnostic explaining that their shell command cache must be refreshed.

## Confirmed cause and workaround

zsh retains the previously resolved command path. Running:

```sh
rehash
# portable alternative
hash -r
```

changes the resolved executable to the npm-installed package and `omo --version` then reports the bundled senpi version (`2026.8.9-2`). Opening a fresh shell also works.

## Suggested fix

Add an install/migration diagnostic and troubleshooting copy for stale shell command caches. A dedicated uncached alias such as `omo-senpi` could also provide a deterministic recovery command, but the minimal fix is an explicit post-install warning plus doctor/docs coverage.

## Environment

- macOS Darwin 25.5.0 arm64
- zsh
- Node.js 24.14.0 via nvm
- npm global package: `omo-ai@5.0.0-0.beta.3`
- legacy wrapper: OMO `4.19.4`

Internal tracking: LAB-39. A cross-reference will be added from the senpi repository because the beta bundles `@code-yeongyu/senpi`.

AceRothstein71:
## What

`omo doctor` now detects every competing `omo` command on PATH, classifies each against the running install by real path, probes foreign copies with `--version`, and warns with the exact recovery commands:

- legacy/stale copies running ahead of this install on PATH
- leftover copies behind this install that a cached shell command may still resolve to
- PATH entries that are not this install at all

New module `packages/omo-native/bin/lib/stale-bin.js` (pure, injectable PATH/platform/probe) wired into `runDoctor()`; regression suite `packages/omo-native/test/stale-bin.test.ts`.

Example output for the #6686 reproduction shape (legacy 4.19.4 wrapper ahead of the upgraded install):

```
WARN stale omo binaries run ahead of this install on PATH: ~/.local/bin/omo (reports 4.19.4) beats <npm-global-bin>/omo
WARN refresh the shell command cache after upgrading: hash -r (zsh: rehash); if ~/.local/bin precedes the npm global bin directory, prepend it: export PATH="$(npm prefix -g)/bin:$PATH"
```

## Why

Installing `omo-ai@beta` does not reach every entry point at once: zsh/bash cache the resolved command path per session, and a legacy managed wrapper in `~/.local/bin` can sit ahead of the npm global bin directory. Both keep an old binary serving after a successful upgrade until the user clears the cache or repairs PATH order (#6686). Doctor had no diagnostic for this state, and a package.json postinstall warning is not available because `package-shape.test.ts` pins the published `omo-ai` manifest to have no postinstall hook.

Design notes:
- Warnings never flip the exit code: a leftover binary is an environment hazard, not a broken install artifact.
- Version probing executes only foreign candidates (`--version`, 2s timeout, max 3 probes) and is skipped entirely on Windows, where Node refuses `.cmd` shims without a shell; those entries report `(unknown version)`.
- Broken symlinks and non-executable entries degrade gracefully instead of throwing.
- Docs coverage for the same recovery steps is intentionally left to #6687, which owns `docs/troubleshooting/README.md` and `docs/guide/installation.md`; this PR stays scoped to the diagnostic surface. Plugin-cache lane #6680 is untouched.

## Verified

- Failing-first: `bun test packages/omo-native/test/stale-bin.test.ts` failed before implementation (`Cannot find module '../bin/lib/stale-bin.js'`), passes after: 7 pass / 0 fail (given/when/then).
- Scoped suites green: stale-bin + doctor + package-shape + launcher = 66 pass / 0 fail; existing doctor tests stay green while inheriting the host PATH because the check is WARN-only.
- Full scoped gate `bun test packages/omo-native/test`: 132 pass / 6 skip / 1 fail - the single failure is the pre-existing environment issue where `payload.test.ts`'s full build aborts on an unfetchable submodule revision (`shared-skills/upstreams/open-design`); it reproduces on the untouched base checkout and involves none of these files.
- Typecheck: `bunx tsc -p packages/omo-native/tsconfig.json --noEmit` clean.
- Entry smoke: `node packages/omo-native/bin/omo.js --version` reports the branded version.
- Live manual QA against a synthetic shadowing layout captured in `.omo/evidence/20260824-6686-stale-binary-cache/live-doctor-shadowing.txt`.
- Evidence directory: `.omo/evidence/20260824-6686-stale-binary-cache/` (red run, green runs, live doctor output, WHAT TESTED/OBSERVED/WHY ENOUGH/OMITTED).

## Risk

Low. The check is additive WARN-only output in `omo doctor`; no launcher dispatch, engine spawn, packaging, or payload changes. Worst case is a missed or extra warning line on exotic PATH layouts; probing is bounded (2s timeout, max 3 foreign candidates, none on Windows).

Fixes #6686

---
## Summary by cubic
Detects stale `omo` binaries on PATH in `omo doctor` and prints exact recovery steps. Previously, doctor could not see competing `omo` commands, so upgrades could still run an old binary; now it scans PATH, classifies entries relative to this install, probes foreign versions, and emits WARN-only messages.

- Adds `packages/omo-native/bin/lib/stale-bin.js` and wires it into `runDoctor()` via `staleBinWarnings()`.
- Scans PATH in order, classifies `omo` entries as ahead/behind/not-this-install, and shows each path with a probed version; caps probes (3) with a 2s timeout; skips version probing on Windows (reports “unknown version”).
- Keeps behavior WARN-only; exit code remains 0. Adds `packages/omo-native/test/stale-bin.test.ts`; existing doctor behavior stays unchanged otherwise.
- No migration for maintainers. Users should refresh the shell command cache after upgrading (hash -r; zsh: rehash) and put the npm global bin first if needed: export PATH="$(npm prefix -g)/bin:$PATH". Fixes #6686.

<sup>Written for commit 42de7a261e6cffc77d4d4ec468fff1728656ad7a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7246?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260824-6686-stale-binary-cache/WHAT-TESTED-OBSERVED-WHY-OMITTED.md
+++ b/.omo/evidence/20260824-6686-stale-binary-cache/WHAT-TESTED-OBSERVED-WHY-OMITTED.md
@@ -1,0 +1,82 @@
+# WHAT WAS TESTED
+
+Issue #6686: after `npm i -g omo-ai@beta`, a zsh session can keep invoking a cached or
+higher-priority legacy `omo` wrapper (e.g. `~/.local/bin/omo`, OMO 4.19.4) even though the newly
+installed npm binary has higher PATH precedence. The shell command cache (`hash -r` / `rehash`)
+and PATH order are the two recovery levers named in the issue.
+
+Repo-side root cause: `packages/omo-native/bin/lib/doctor.js` (`runDoctor`, lines 54-98 at base
+8833800ae) had no diagnostic for competing `omo` commands on PATH, so an upgraded install could
+neither detect nor explain the stale-binary state. A package.json postinstall warning is not an
+option: `packages/omo-native/test/package-shape.test.ts` pins "no postinstall lifecycle hook
+exists" for the published `omo-ai` manifest.
+
+Tested surfaces:
+
+1. Failing-first regression suite `packages/omo-native/test/stale-bin.test.ts` (7 cases,
+   given/when/then):
+   - unit: `findOmoPathEntries` PATH ordering, dedupe, missing dirs, empty PATH,
+     non-executable skip (posix-only);
+   - integration through the real launcher (`node bin/omo.js doctor` spawned as a child with a
+     fully controlled PATH): legacy wrapper ahead of the upgraded install; upgraded install ahead
+     of a leftover copy; only the upgraded install on PATH (no false positives); foreign wrapper
+     whose --version fails.
+2. Scoped regression suites: `stale-bin.test.ts`, `doctor.test.ts`, `package-shape.test.ts`,
+   `launcher.test.ts` (66 pass / 0 fail).
+3. Full scoped package gate: `bun test packages/omo-native/test`.
+4. Typecheck: `bunx tsc -p packages/omo-native/tsconfig.json --noEmit`.
+5. Entry smoke: `node packages/omo-native/bin/omo.js --version`.
+6. Live manual QA of the new doctor check against a synthetic shadowing layout (see
+   live-doctor-shadowing.txt).
+
+# WHAT WAS OBSERVED
+
+- RED (before implementation): `bun test packages/omo-native/test/stale-bin.test.ts` failed with
+  `Cannot find module '../bin/lib/stale-bin.js'` - 0 pass, 1 fail. The regression suite could not
+  pass before the fix existed.
+- GREEN (after implementation): 7 pass / 0 fail for the new suite; 66 pass / 0 fail across the
+  four focused files; full package gate 132 pass / 6 skip / 1 fail where the single fail is the
+  pre-existing environment issue described below.
+- Doctor output for a shadowing layout (legacy stub reporting 4.19.4 ahead of the upgraded
+  install), exit code 0:
+  WARN stale omo binaries run ahead of this install on PATH: <legacy>/omo (reports 4.19.4)
+       beats <npm-bin>/omo
+  WARN refresh the shell command cache after upgrading: hash -r (zsh: rehash); if ~/.local/bin
+       precedes the npm global bin directory, prepend it:
+       export PATH="$(npm prefix -g)/bin:$PATH"
+- With only the upgraded install on PATH, doctor emits no stale-binary warnings (no false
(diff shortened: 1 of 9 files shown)


MoerAI (changes requested):
Thanks for tackling #6686. Please rebase onto current `dev` and resolve the doctor integration against #7372.

The classifier must recognize the active install through npm Windows `.cmd` wrappers and the current Bun global shell shim rather than treating every path outside `packageRoot` as foreign. It must also honor empty POSIX PATH components as the current directory.

Add Windows wrapper, Bun shim, and empty-PATH-component regressions using target-platform path semantics consistent with #7140. Then run the full required matrix; this head's CI was never approved and none of the required checks ran.

PR #6692, opened 2026-08-10
fix(omo-native): preserve Windows System32 for Senpi

ashmoonori-afk:
## Summary

- Restore `%SystemRoot%\System32` to the inherited Windows PATH before launching the pinned Senpi engine, without changing the inherited PATH key casing.
- Keep the Senpi shim first while deduplicating equivalent mixed-separator entries such as `C:\path` and `C:/path`.
- Refresh the branch onto current `dev` and lock both regressions with deterministic Windows tests.

## Changes

- `packages/omo-native/bin/lib/launcher.js`: add case-insensitive environment-key lookup and a small PATH prepend/dedupe helper; restore System32 before prioritizing the Senpi shim.
- `packages/omo-native/test/launcher.test.ts`: cover missing-System32 inheritance, mixed-separator shim dedupe, PATH key casing, and current launcher dispatch semantics.
- `.omo/evidence/20260811-windows-taskkill-path/`: record red/green output, automated gates, and sanitized manual CLI QA.

## QA & Evidence

- **What was tested:** `bun test packages/omo-native/test/launcher.test.ts --test-name-pattern 'restores System32|promotes an existing shim'`
  **Observed result:** RED without the fix: 0 pass, 2 fail; GREEN with the fix: 2 pass, 0 fail.
  **Artifact:** `.omo/evidence/20260811-windows-taskkill-path/red-green.txt`
  **Why sufficient:** toggles the exact missing-System32 and duplicate-shim causes.
- **What was tested:** `bun test packages/omo-native/test/launcher.test.ts`
  **Observed result:** 28 pass, 1 platform skip, 0 fail.
  **Artifact:** `.omo/evidence/20260811-windows-taskkill-path/QA.md`
  **Why sufficient:** covers the full launcher contract on the changed package.
- **What was tested:** `bun run typecheck` and `bun install --frozen-lockfile`
  **Observed result:** both exited 0; postinstall completed the repository build.
  **Artifact:** `.omo/evidence/20260811-windows-taskkill-path/QA.md`
  **Why sufficient:** covers TypeScript projects, packaging, and generated launcher staging.
- **What was tested:** standalone happy, bad-input, and help driver using the real launcher files with System32 omitted from inherited `Path`.
  **Observed result:** statuses 0, 2, and 0; `Path` casing preserved, shim first, System32 present in all three.
  **Artifact:** `.omo/evidence/20260811-windows-taskkill-path/manual-qa.txt`
[...]

Issue #6690, opened 2026-08-10
fix(omo-native): preserve Windows System32 for taskkill cleanup


ashmoonori-afk:
## Bug description

`omo-ai@5.0.0-beta.3` launches its pinned `@code-yeongyu/senpi@2026.8.9-2` dependency with the environment assembled in `packages/omo-native/bin/lib/launcher.js::senpiEnvironment()`. On Windows, that inherited environment can lack the System32 directory, so Senpi cannot resolve `taskkill` during detached-child cleanup and terminates the active Pi session with an uncaught exception.

## Repeated live OMO failure

Two independent OMO-native sessions have failed through the same generic cleanup call shape:

```text
pi exiting due to uncaughtException:
Error: spawn taskkill ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn taskkill',
  path: 'taskkill',
  spawnargs: [ '/F', '/T', '/PID', '54956' ]
}
```

Latest recurrence:

```text
pi exiting due to uncaughtException:
Error: spawn taskkill ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at processTicksAndRejections (node:internal/process/task_queues:89:21)
    at runNextTicks (node:internal/process/task_queues:68:3)
    at process.processTimers (node:internal/timers:538:9) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn taskkill',
  path: 'taskkill',
  spawnargs: [ '/F', '/T', '/PID', '28880' ]
}
```

## Root cause and scope

The argument order `['/F', '/T', '/PID', pid]` maps to the generic `killProcessTree()` path in Senpi's `utils/shell`, reached through `killTrackedDetachedChildren()`. It is distinct from the PTY pipe-fallback call shape `['/pid', pid, '/T', '/F']` covered by Senpi PR [#776](https://github.com/code-yeongyu/senpi/pull/776).

Senpi needs to handle an asynchronous `ChildProcess` `error` event, but OMO native also controls the launched Senpi environment. Its launcher currently prepends its local Node shim directory without guaranteeing that Windows System32 remains available to the child process.

## Expected behavior

[...]

PR #7247, opened 2026-08-25
fix(omo-senpi): preserve Windows Path casing in toolkit path provisioning

### LINKED ISSUE #6689: [Bug]: Windows omo-ai launcher drops inherited Path when spawning Senpi
# OMO Windows `Path`/`PATH` Inheritance Bug Report

## Summary

- **Date:** 2026-08-10
- **Affected package:** `omo-ai@5.0.0-0.beta.2`
- **Affected dependency:** `@code-yeongyu/senpi@2026.8.9-2`
- **Platform:** Windows 10.0.22621 x64
- **Severity:** High for Windows CLI usability
- **Status:** Reproduced and locally patched

OMO loses the user's inherited `Path` when launching Senpi on Windows. As a
result, commands installed in ordinary user or machine PATH locations, such as
`uv`, `git`, `winget`, `py`, and `pip`, are unavailable from OMO tool shells.

The bug is caused by copying `process.env` into a case-sensitive JavaScript
object and then reading only the uppercase `PATH` property. Windows environment
variable names are case-insensitive, but ordinary JavaScript object keys are
not.

## User-visible symptom

`uv` works in the user's PowerShell:

```text
C:\Users\admin\.local\bin\uv.exe
uv 0.11.26 (396ef7ce4 2026-06-30 x86_64-pc-windows-msvc)
```

The same command is missing from an OMO tool PowerShell:

```text
UV_NOT_FOUND
```

The tool process receives only these three PATH entries:

```text
C:\Users\admin\.senpi\agent\bin
C:\Users\admin\AppData\Roaming\npm\node_modules\omo-ai\plugin\runtime\agent-toolkit
C:\Users\admin\AppData\Roaming\npm\node_modules\omo-ai\node_modules\@code-yeongyu\senpi\node_modules\.bin
```

The configured user PATH does contain the missing directory:

```text
C:\Users\admin\.local\bin
```

## Process boundary

The observed process tree was:

```text
WindowsTerminal.exe
└─ powershell.exe                         # user shell
   └─ node.exe ...\omo-ai\bin\omo.js
      └─ node.exe ...\senpi\dist\cli.js
         └─ node.exe ...\senpi\dist\cli-main.js
            └─ powershell.exe -NoProfile -Command ...
```

The `-NoProfile` option explains why PowerShell profile scripts are not loaded,
but it does not explain the lost PATH. The inherited process environment should
still contain the user's PATH.

## Root cause

The Windows registry names both user and machine variables `Path`:

```text
HKCU\Environment: Path
HKLM\...\Session Manager\Environment: Path
```

The affected launcher code was in
`C:\Users\admin\AppData\Roaming\npm\node_modules\omo-ai\bin\lib\launcher.js`:

```js
function senpiEnvironment(senpiRoot) {
  const env = { ...process.env }
  // ...
  const binDir = nearestNodeBin(senpiRoot)
  if (binDir) {
    env.PATH = env.PATH ? `${binDir}${delimiter}${env.PATH}` : binDir
  }
  return env
}
```

The failure sequence is:

1. Windows exposes the inherited variable as `Path`.
2. `process.env` is spread into the ordinary object `env`.
3. The ordinary object treats `Path` and `PATH` as different keys.
4. `env.PATH` is therefore `undefined`.
5. The launcher creates a second key containing only Senpi's local `.bin`.
6. When Node spawns the Windows child with both keys, the child receives the
   uppercase `PATH` value and loses the inherited `Path`.
7. OMO and Senpi prepend their own two directories, producing the exact
   three-entry PATH observed in the tool shell.

## Minimal reproduction

The launcher behavior was reproduced with:

```js
import { spawnSync } from "node:child_process"

const env = { Path: "USER_PATH_WITH_UV" }
const binDir = "SENPI_NODE_MODULES_BIN"

env.PATH = env.PATH ? `${binDir};${env.PATH}` : binDir

const child = spawnSync(process.execPath, ["-p", "process.env.PATH"], {
  env,
  encoding: "utf8",
})

console.log(env)
console.log(child.stdout)
```

Observed output:

```text
{
  Path: "USER_PATH_WITH_UV",
  PATH: "SENPI_NODE_MODULES_BIN"
}

SENPI_NODE_MODULES_BIN
```

## Applied fix

The launcher now locates the existing PATH key case-insensitively and updates
that same key:

```js
export function senpiEnvironment(senpiRoot) {
  const env = { ...process.env }
  // ...
  const binDir = nearestNodeBin(senpiRoot)
  if (binDir) {
    const pathKey =
      Object.keys(env).find((key) => key.toLowerCase() === "path") ?? "PATH"
    const currentPath = env[pathKey]
    env[pathKey] = currentPath ? `${binDir}${delimiter}${currentPath}` : binDir
  }
  return env
}
```

This preserves the original key casing and prevents duplicate `Path`/`PATH`
entries.

## Regression test

The regression test is:

```text
C:\Users\admin\gwkim\workspace\omo-launcher-path.test.mjs
```

Before the fix it failed with:

```text
Expected: ["Path"]
Actual:   ["Path", "PATH"]
```

After the fix:

```text
tests 1
pass 1
fail 0
```

The patched launcher also passes `node --check`.

## Manual verification

The patched `senpiEnvironment()` was given a Windows environment containing
the registry-derived user and machine PATH under the `Path` key. A new
`powershell.exe -NoProfile` process was then launched with the resulting
environment.

The environment contained exactly one case-insensitive PATH key:

```text
Path
```

The following commands were executed in that child shell:

```powershell
Get-Command uv
uv --version
uv --help
uv __omo_invalid_command__
```

Observed happy-path result:

```text
C:\Users\admin\.local\bin\uv.exe
uv 0.11.26 (396ef7ce4 2026-06-30 x86_64-pc-windows-msvc)
```

Observed help output:

```text
An extremely fast Python package manager.

Usage: uv.exe [OPTIONS] <COMMAND>
```

The invalid subcommand returned exit status `1` and the expected error:

```text
error: unrecognized subcommand '__omo_invalid_command__'
```

## Distribution note

The current patch modifies the globally installed npm package. A future
`npm install`, update, or reinstall of `omo-ai` can overwrite it. The same fix
should be submitted to the upstream source repository and released in a new
package version.

AceRothstein71:
## What

Resolve the actual PATH environment key case-insensitively in the senpi extension's toolkit path provisioning, and route its env mutations through an injectable env object (default `process.env`).

## Why

On Windows the inherited variable is spelled `Path` (HKCU/HKLM `Environment: Path`), while JS object keys are case-sensitive. `prependPathEntry()` in `packages/omo-senpi/src/extension/toolkit-path-provisioning.ts` read and wrote the hardcoded uppercase `PATH`: on env surfaces without Node's Windows case-insensitive magic (including the Bun re-exec runtime this product ships), the read returned `undefined` and the write created a SECOND `PATH` key holding only the toolkit dir. Every later `{ ...process.env }` spawn merge then carried both keys, and the truncated uppercase entry shadowed the inherited full value in the child environment block. That is exactly the measured child PATH from the issue (`~/.omo/bin`, agent-toolkit, senpi `.bin`; System32, nodejs, Git\\cmd all dropped), which breaks MCP servers launched via `node`, `powershell.exe`, and bare-name lookups.

The launcher-side twin of this bug was already fixed on dev in `9cb2b3763` (`packages/omo-native/bin/lib/launcher.js:78`, case-insensitive key lookup); the extension-side mutation site kept the hardcoded spelling. This change applies the same resolution pattern there.

## Verified

- Failing-first: two regression tests were added before the fix and failed against the old code (`.omo/evidence/20260824-6689-win-path-drop/red.txt`, 5 pass / 2 fail):
  - a win32-shaped env whose path key is `Path` keeps its full value under that spelling after provisioning, with no duplicate uppercase `PATH` key created;
  - an env carrying both spellings prepends onto the first resolved key without adding a third variant.
- After the fix: scoped file 7 pass / 0 fail (`green.txt`), extension suite 65 pass / 0 fail (`green-extension.txt`), `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` exit 0 (`typecheck.txt`).
- Full `bun test packages/omo-senpi` shows 11 failures that reproduce identically with this diff stashed (`baseline.txt`) - they depend on generated plugin artifacts the environment's broken `bun install` prepare step never built; documented as pre-existing in `QA.md`.
- Evidence: `.omo/evidence/20260824-6689-win-path-drop/` (PLAN, QA, red/green/typecheck/baseline logs).

## Risk

Low. The default path (`env = process.env`) keeps today's behavior on POSIX and under Node-on-Windows; only the casing of the key being written changes on Windows-style surfaces, from a duplicate `PATH` to the existing `Path`. Read-only `process.env.PATH` consumers elsewhere (ulw-loop o-command lookup, comment-checker resolver, memory sandbox which-scan) cannot create duplicate keys and are intentionally untouched; open PR #6692 (launcher-side System32 restoration) and #6690 (taskkill ENOENT) remain separate surfaces.

Fixes #6689

---
## Summary by cubic
Preserves Windows Path casing in `omo-senpi` toolkit PATH provisioning. Previously we hardcoded `PATH`, which created a duplicate uppercase key and truncated child PATH values; now we resolve the actual key case-insensitively and mutate it, preventing drops in child processes.

- Resolves the PATH key via case-insensitive lookup and routes mutations through an injectable `env` (default `process.env`); applies the same to toolkit bin setup.
- Behavior: on Windows-like, case-sensitive env surfaces (e.g., Bun re-exec), we prepend to the inherited `Path` key instead of creating a duplicate `PATH`; POSIX behavior is unchanged.
- Tests: adds two regressions that fail on the old behavior and pass now; extension suite and typecheck are green. Evidence logs are included for QA.
- No migration required; the new `env` option is additive. Fixes #6689.

<sup>Written for commit 9959ebe43a3ba5a2fadb4f47c02e78a16a8ecafe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
[sisyphus-bot] Thank you for this PR, and for the detailed Windows PATH evidence trail! 🙏

I'm closing it as **no longer applicable**: both source files this PR changes were deleted from `dev` by commit `48506e57a` (*build(omo-native): remove Native toolkit CLI payload*), so there is no longer any surface for the change to land on. The underlying issue #6689 was also closed as completed — the launcher-side casing fix `9cb2b3763` (*fix(omo-native): preserve Windows PATH casing*) resolves the reported defect from the other direction.

**How I checked:** `git ls-tree origin/dev -- packages/omo-senpi/src/extension/toolkit-path-provisioning.ts` and `…test.ts` → both return nothing; `git show 48506e57a --name-status` lists both files as `D`; `git merge-base --is-ancestor 48506e57a origin/dev` → yes; `gh issue view 6689` → `CLOSED COMPLETED`.

Your reproduction work was genuinely useful in getting #6689 understood and fixed. If I've misread anything and part of this is still needed, please say so — I'll reopen right away. 🙇

PR #7248, opened 2026-08-25
fix(omo-opencode): stop stale-offset reprints in run event renderer

### LINKED ISSUE #6513: tool-result render corruption: duplicated/substituted lines and character mutation in eval and Read output
## Summary

Tool results are sometimes rendered with corrupted content: trailing lines duplicated, some lines substituted with content from earlier in the same result, and in one case characters mutated into a string that was never written. The underlying data on disk is correct — the corruption is in the rendered result the agent sees.

This is high-impact because the agent has no way to detect it. The rendered text *is* the observation, so corrupted values get carried forward into conclusions.

### Case 1 — `eval` (py): duplicate keys that cannot exist

A cell built a `dict`, wrote it with `json.dump`, then printed it. A Python `dict` cannot hold duplicate keys, so `json.dumps` output must contain each key exactly once.

Rendered output:

```
{
 "pid": 42856,
 "threads": ["MainThread", "Thread-1", ...],
 "thread_count": 6,
 "gjc_procs": 24,
 "node_procs": 61,
 "thread_count": 6,     <-- duplicate
 "gjc_procs": 24,       <-- duplicate
 "node_procs": 61       <-- duplicate
}
```

The file written by the same cell contained each key exactly once. So the duplication was introduced after serialization, in the display/transport path.

### Case 2 — `Read`: block repetition, line-number desync, character mutation

Reading a JSON file written moments earlier by `json.dump` produced:

- one `git_bash_paths` array repeated ~7 times in a single render (the file contains it once)
- line numbering that ran to 47 and then restarted at 8
- a path string rendered as `C:\Program Files\Ceb\bin\bash.exe` — the file contained `C:\Program Files (x86)\Git\bin\bash.exe`. `(x86)\Git` became `Ceb`.

Character-level mutation is the most serious part: it is not explainable as a repeated buffer flush.

## Observed characteristics

- **Intermittent.** A minimal reproduction attempt (12-key dict, print + file write) rendered cleanly. I could not force it on demand.
- **Tail-biased.** Duplication clusters at the end of a result block.
- **Not size-dependent.** Small outputs corrupted while larger ones in the same session rendered fine.
- **Not format-specific.** Occurred with plain `print` lines, `json.dumps` strings, and `Read` file output.
- **Kernel reset helps only temporarily.** Clean for a few cells, then recurs.

## Impact

During an audit task where numeric accuracy was the whole point, I read three values off corrupted renders and reported them as fact:

| Reported from render | Actual (re-measured via file) |
|---|---|
| subagent `await` = 449 | 86 |
| subagent `cancellation` = 98 | 7 |
| scenario totals | inflated by +813 files / +27.2 MiB |

All three were wrong by large margins. They were caught only because a second party recomputed independently.

## Workaround

Write results to a file inside the cell, then read the file back. Disk content has been correct in every instance so far. Structured output (JSON) makes corruption self-evident, since a serializer cannot emit duplicate keys — if the render shows them, the render is wrong.

## Suggested investigation

The pattern (tail-biased duplication + cross-line substitution + line-number desync) looks like chunk reassembly or replay in the tool-result transport rather than a fault in the kernel itself. Buffer boundary handling and any retry/replay logic in the result streaming path seem like the places to look.

## Environment

- oh-my-opencode 4.19.3
- Windows 11
- `eval` tool, Python 3.12 kernel
- Also observed in the `Read` tool

AceRothstein71:
## What

Diff `message.part.updated` text/reasoning snapshots in the `omo run` event renderer against per-part baselines (`partTextById` / `reasoningTextById`, keyed by part ID, prefix-matched) instead of the single-slot session accumulators (`lastPartText` / `lastReasoningText`). Anonymous (id-less) delta streams are adopted via the anonymous bucket for legacy event shapes, and baselines are cleared when a new assistant message starts. The global mirrors are still maintained for the existing run-summary reader (`runner.ts`).

## Why

The old renderer computed the newly-streamed text as `part.text.slice(state.lastPartText.length)`. That accumulator is shared across all parts and is force-reset by unrelated events (`handleToolPart` on tool completion at `event-message-handlers.ts:146`, `handleToolResult` at `event-tool-handlers.ts:36`, and new-message handling). Two failure modes follow directly from stale offsets:

- A text part that emits a later full snapshot after an interleaved tool result gets re-rendered from offset 0: the whole block prints again. This is the tail-biased duplication reported in #6513.
- A sibling part whose snapshot is shorter than the accumulator renders `slice(staleLength)`: content is dropped entirely or printed from a mid-content offset, which shows up as substituted lines and spliced/mutated character runs.

Both modes are intermittent because they depend on SSE event ordering, match the issue's signature (duplicated/substituted lines, character mutation, line-number desync in eval/Read displays), and corrupt only the rendered observation while the underlying data stays correct.

## Verified

- Failing-first regression tests co-located in `packages/omo-opencode/src/cli/run/event-message-handlers.test.ts` (given/when/then), reproducing both corruption modes; red output captured with the fix stashed: 9 pass / 2 fail (`alpha` rendered twice; `second part body` dropped).
- After fix: `bun test packages/omo-opencode/src/cli/run/` green - 200 pass / 0 fail, including 4 new tests (interleaved-tool reprint, shorter sibling part, suffix-only growth guard, delta-then-snapshot dedup) and all pre-existing renderer contracts.
- `bun run typecheck` clean (tsgo --noEmit + script + all workspace packages).
- Evidence recorded under `.omo/evidence/20260824-6513-render-corruption/` (WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED plus raw red/green outputs).

## Risk

Low. Rendering-only change to the CLI run path; no tool results, hooks, or stored session data are affected. One intentional behavior change: a genuine provider-side replacement of already-streamed text now re-renders the replacement block in full instead of attempting an in-place diff - honest output, and it cannot produce mid-text splices.

Fixes #6513

---
## Summary by cubic
Stops duplicated/dropped lines in `omo run` output by diffing `message.part.updated` snapshots against per-part baselines. Old behavior: diffed against shared accumulators (`lastPartText`/`lastReasoningText`) that were reset by unrelated events, causing full reprints and mid-text splices. New behavior: each part diffs against its own prefix baseline; legitimate provider-side replacements reprint the new block in full. Fixes #6513.

- Reviewer notes:
  - Store/render per-part baselines via `partTextById` and `reasoningTextById`; use `resolveRenderedBaseline` for prefix-matched diffs, adopt an anonymous bucket for id-less deltas, and clear baselines on new assistant messages.
  - Keep global mirrors for the run-summary reader in `runner.ts`.
  - Added regression tests reproducing the issue; full suite green; typecheck clean. No migration required.

<sup>Written for commit 3b2042497560b3668d37901561ee133a86ee89b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thank you for addressing the renderer issue. The confirmed blocker is incomplete validation and linkage: the issue-closing reference is wrong, live opencode QA is missing, and anonymous-baseline retirement is not covered. Please correct the closing linkage, cover anonymous-baseline retirement in regression tests, and provide live opencode QA showing stale offsets are not reprinted. Files: https://github.com/code-yeongyu/oh-my-openagent/pull/7248/files

PR #7249, opened 2026-08-25
fix(senpi-task): wake parent on task_send revival by promoting the revived epoch to background

### LINKED ISSUE #6532: bug(senpi-task): task_send revival of completed foreground task never wakes idle parent
## Summary

A task started synchronously (`run_in_background: false`) can be revived with `task_send`, and the revived child completes, but the idle parent session is never woken.

This was reproduced in a fresh session after updating Senpi core to `2026.7.31-2`, which includes code-yeongyu/senpi#184.

## Reproduction

1. Start a task synchronously and wait for its inline result.
2. Call `task_send` on that completed resident task, which returns `Revived <task_id> (run epoch 1)`.
3. End the parent turn immediately without polling.
4. Let the revived child complete.
5. Observe whether the idle parent receives a completion wake.

Minimal sequence:

```text
task(..., run_in_background=false) -> INITIAL_OK
task_send(to=<completed task>, message="Use one read-only tool, then reply REVIVED_OK")
parent -> WAITING_FOR_AUTO_WAKE
child -> REVIVED_OK
```

## Observed

The child record reaches:

```json
{
  "status": "completed",
  "notification": {
    "run_epoch": 1,
    "notified_epoch": -1
  },
  "final_response": "REVIVED_OK"
}
```

Fresh-session timing:

```text
17:32:27Z task_send revived epoch 1
17:32:30Z parent ended with WAITING_FOR_AUTO_WAKE
17:32:34Z child completed with REVIVED_OK
>60 seconds later: no parent completion message and no new assistant turn
```

The parent session contains no `senpi-task.completion` / `omo-senpi:wake` event after the child completion.

## Expected

Because `task_send` returns immediately and the revived run continues asynchronously, completion of that run should wake an idle parent exactly once and advance `notified_epoch` to `1`.

## Suspected cause

The original synchronous task is never promoted to background when a terminal resident task is revived. The completion bridge passes `manager.wasBackground(taskId)`, and the notifier skips when `runInBackground` is false with reason `sync-task`.

Relevant paths:

- `packages/senpi-task/src/steering/engine.ts` (`reviveTerminal`)
- `packages/senpi-task/src/manager/manager.ts` (`promoteToBackground` / `wasBackground`)
- `packages/senpi-task/src/completion/notifier.ts` (`!request.runInBackground` skip)

Potential minimal fix: treat a terminal-task revival via `task_send` as asynchronous for the new epoch, or persist notification delivery mode per epoch. Avoid enabling notifications for the initial synchronous epoch, which would duplicate the inline result.

## Environment

- Senpi core: `2026.7.31-2`
- omo-senpi local checkout: `085868d3152f0e8cca5eb690e54a51a22f02172f`
- The current upstream `dev` versions of the relevant files were inspected and retain the same background skip / revival flow.
- Fresh parent session, in-process child, no polling or monitor inside the parent.

AceRothstein71:
## What

Reviving a terminal resident child via `task_send` now promotes the revived epoch to background before the follow-up prompt starts, so its completion wakes the idle parent. Concretely:

- `SteeringPort` gains `promoteToBackground(taskId)`, wired to the manager's existing public promotion (persisted `notify_on_terminal` + `background_mode`).
- `reviveTerminal()` promotes BEFORE `handle.followUp()` (so even a fast-settling revived turn notifies), rolls the promotion back if starting the follow-up fails, and builds the revived record from the freshly persisted state so the replace does not revert the promotion.
- Steering test fixture mirrors the manager's promotion semantics; two inline engine-test ports gain a no-op member.

## Why

Issue #6532: a synchronous task completes, `task_send` revives the terminal resident child, the revived run finishes - and the idle parent never wakes. Root cause chain: `reviveTerminal()` never promoted the revived epoch (`packages/senpi-task/src/steering/engine.ts`), `buildRevived()` passed `notify_on_terminal` through untouched, so `wasBackground()` stayed false and `notifier.ts` skipped with `{ kind: "skipped", reason: "sync-task" }` before idle wake routing was reached. The result only surfaced after a later user message plus `task_output`.

Once revived, the run is asynchronous from the parent's perspective regardless of how it was spawned, so notification intent must flip at revival time.

## Verified

- Failing-first regression at both seams (red observed before the fix, green after):
  - `packages/senpi-task/src/steering/engine.test.ts`: promote-before-followUp ordering (snapshot of persisted state taken inside the handle), rollback on failed follow-up start (promotion undone, error propagates, reservation released not committed), originally-background children keep `background_mode: "background"`.
  - `packages/omo-senpi/src/components/task/completion-bridge.test.ts`: full sync spawn -> terminal -> revive -> completion chain delivers exactly ONE wake per run_epoch; repeated revivals notify once per epoch; originally-background spawns keep notifying exactly once per epoch across revival.
- `bun test packages/senpi-task`: 1748 pass / 0 fail (includes the 200-iteration chaos bench pinning exactly-once per `(task_id, run_epoch)`).
- `bun test packages/omo-senpi`: 11 failures, all pre-existing artifact/skill-sync tests that fail identically on the base commit in this sandbox (failed `bun install` prepare step means plugin artifacts were never built); none touch the task component.
- `tsgo --noEmit` clean for both `senpi-task` and `omo-senpi`.
- Live harness A/B with `scripts/qa/task-e2e.mjs` (real senpi binary, isolated agent dirs): results byte-identical between base and patch - same PASS set including `unconditional_wake` and `sync_inline_no_notification`, same pre-existing FAIL set on both sides; `real_senpi_untouched: true`. Evidence under `.omo/evidence/20260824-6532-task-send-wake/`.

## Risk

Low. Behavior changes only for tasks that are revived while `notify_on_terminal` is false - exactly the population the issue reports. Originally-background tasks take a no-op promotion path (guard in `promoteToBackground`), failed revives roll back to their exact prior record fields (guarded mutate never clobbers concurrent writers), and the concurrency/outcome epoch machinery is untouched. Residual risk: hosts that relied on revived sync children staying silent will now receive one wake per revived epoch, which is the requested behavior.

Fixes #6532

---
## Summary by cubic
Reviving a terminal resident child via task_send now promotes the revived epoch to background before the follow-up, so its completion wakes the idle parent. Previously, revival left notify_on_terminal=false and the notifier skipped with "sync-task", so the parent never woke. Fixes #6532.

- Add promoteToBackground(taskId) to `SteeringPort` and wire it to the manager’s existing promotion (persisted notify_on_terminal and background_mode). Implementers of `SteeringPort` must add this method (tests and fakes updated).
- Promote before followUp(); roll back if follow-up start fails; rebuild the revived record from fresh store state to preserve the promotion.
- Resulting behavior: parents receive exactly one wake per revived epoch; originally background tasks are unchanged.

<sup>Written for commit a8d807263c987d4c2b9078d2d1cc466e4ada745a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
The promotion direction matches the owner-approved fix for #6532, but this branch conflicts with current `dev` and predates the current `task_send` and eviction arbitration.

Rebase and integrate promotion inside the guarded revive path. Keep the initial foreground epoch silent, promote before the revived follow-up starts, and prove exactly one parent wake per revived epoch with pre-subscribed event-based tests rather than timer flushing.

Add a real Senpi scenario for `foreground terminal -> task_send -> parent idle -> revived terminal -> one wake`. Remove or redact the raw QA JSON that triggered GitGuardian, then obtain the complete required matrix.

PR #7251, opened 2026-08-25
fix(senpi-task): apply maintained OpenAI-only registry recommendations at resolution time

### LINKED ISSUE #6813: [Bug]: OMO Native does not apply OpenAI-only recommendations from the live Senpi registry
## Prerequisites

- [x] I searched existing issues and pull requests for OpenAI-compatible providers, model recommendations, Senpi model routing, `models.json`, and single-provider installs.
- [x] I reproduced the configuration shape on the current OMO Native beta and traced current `dev` (`923726cde`).
- [x] I reviewed the model-matching documentation and the existing OpenCode-only recommendation policy.

## Bug Description

OMO advertises provider-aware automatic model configuration, and the OpenCode installer has a maintained OpenAI-only recommendation catalog. OMO Native does not consume that catalog.

For a Codex/OpenAI-compatible endpoint registered as the canonical Senpi `openai` provider, Senpi correctly discovers the live authenticated Sol/Terra/Luna inventory. However, the Native task resolver only applies its static generic fallback chains. With no manual `[senpi]` overrides:

- `artistry` has no OpenAI rung and is unavailable;
- `writing` has no OpenAI rung and is unavailable;
- `quick` has only a `quotio-openai` Luna-fast rung, so direct `openai` is unavailable;
- `visual-engineering` eventually reaches Sol with `medium`, instead of the maintained OpenAI-only `high` recommendation.

The same connected models are assigned correctly only after the user manually copies the OpenAI-only recommendations into `~/.omo/omo.jsonc`.

## Steps to Reproduce

1. Configure an OpenAI-compatible Responses endpoint in the OMO Native `models.json` under provider id `openai`, with `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna` available. Authenticate that provider.
2. Leave `[senpi].categories` and `[senpi].agents` unset in `~/.omo/omo.jsonc`.
3. Launch `omo` and let Senpi populate its authenticated live `modelRegistry`.
4. Resolve or dispatch `task(category="artistry")`, `task(category="writing")`, `task(category="quick")`, and `task(category="visual-engineering")`.
5. Compare the result with the OpenCode installer policy in `packages/omo-opencode/src/cli/openai-only-model-catalog.ts`.

## Expected Behavior

Once the authenticated live registry proves that the runtime is OpenAI-only and the recommended model is actually available, OMO Native should apply the maintained recommendations automatically:

| Target | Model | Effort |
|---|---|---|
| `artistry` | `gpt-5.6-sol` | `xhigh` |
| `writing` | `gpt-5.6-sol` | `medium` |
| `visual-engineering` | `gpt-5.6-sol` | `high` |
| `quick` | `gpt-5.6-luna-fast` | default |
| `explore`, `librarian` | `gpt-5.6-luna-fast` | `low` |

Any explicit `categories.<name>` or `agents.<name>` entry must win. `architect` must remain governed by its required-model gate and must not be synthesized from an unrelated model. User files should not be rewritten at runtime.

## Actual Behavior

The setup path detects provider ids but discards model inventory, while the task component loads config before Senpi exposes the authenticated registry. At spawn time the planner has the correct live registry, but it only evaluates static fallback chains. The OpenAI-only catalog remains isolated in the OpenCode CLI and is never applied to OMO Native.

This creates a product-contract gap: connecting a working compatible endpoint is not enough to receive OMO's maintained automatic category tuning in the Native edition.

## Root Cause

- `packages/omo-native/bin/lib/setup-detect.js` records provider presence from `models.json`, not runtime model availability or identity.
- `packages/omo-senpi/src/components/task/event-bridge.ts` captures the auth-filtered `modelRegistry` at `session_start`.
- `packages/omo-senpi/src/components/task/planner.ts` resolves at spawn time, which is the first safe point that has both user config and live inventory.
- `packages/senpi-task/src/category/fallback-chains.ts` contains only generic multi-provider chains.
- `packages/omo-opencode/src/cli/openai-only-model-catalog.ts` owns the OpenAI-only policy separately.

## Proposed Fix

1. Move the maintained OpenAI-only recommendations into a harness-neutral pure module.
2. Compile a recommendation overlay from Senpi's authenticated live registry at resolution time.
3. Apply a recommendation only when the inventory is safely identified as OpenAI-only and the exact recommended model exists.
4. Skip every target with an explicit user entry; never rewrite `omo.jsonc`.
5. Keep `architect` out of the overlay and preserve its existing hard gate.
6. Expose Senpi's optional `getUpstreamModelId(model)` through the structural registry port so an explicitly mapped provider alias can be identified safely. Do not infer model identity merely from an OpenAI-compatible wire protocol.

## Related Work Scanned

- #6799: installer writes unavailable hardcoded fallbacks for a single non-default provider. Same product principle, different OpenCode install path.
- #839 / closed #6460: centralizing model literals. Useful policy foundation, but no runtime recommendation compiler.
- #4886: proxy-prefixed gateway identity RFC. Relevant to safe aliasing, but does not implement Native recommendations.
- #5719 / #5721: runtime-backed capability classification for custom providers. Capability metadata, not category recommendation selection.
- merged #6373: shared category fallback ordering across adapters. Static chain maintenance, not provider-set-specific overlays.
- closed #6598: preserves standard Senpi fallbacks after user overrides. This fix must retain that behavior and stronger user precedence.
- #2075: availability-aware quick-model selection in OpenCode. Similar intent, different model/adapter.
- #6809: setup/runtime agent-directory mismatch. Important adjacent Native setup bug, but independent of recommendation resolution.

No existing issue or open PR implements this exact live-registry OpenAI-only overlay for OMO Native.

## Environment

- OMO Native: `5.0.0-0.beta.6` configuration, current `dev` source trace at `923726cde`
- OS: Linux / WSL2
- Provider: canonical `openai` id backed by a private OpenAI-compatible Responses endpoint
- Secrets, endpoint URL, and auth headers omitted

AceRothstein71:
## What

OMO Native now compiles the maintained OpenAI-only model recommendations from the authenticated live senpi model registry at task-resolution time:

- `artistry` -> `gpt-5.6-sol` (`xhigh`), `writing` -> `gpt-5.6-sol` (`medium`), `visual-engineering` -> `gpt-5.6-sol` (`high`), `quick` -> `gpt-5.6-luna-fast`, curated `explore`/`librarian` -> `gpt-5.6-luna-fast` (`low`).
- New harness-neutral data module `packages/delegate-core/src/openai-only-recommendations.ts` mirrors the maintained catalog from `packages/omo-opencode/src/cli/openai-only-model-catalog.ts` (single maintained policy, consumed by both editions; `architect` deliberately absent so its required-model gate stays authoritative).
- New pure compiler `packages/senpi-task/src/category/openai-only-overlay.ts` applies a recommendation only when (1) the target has no explicit user entry, (2) the inventory is safely identified as OpenAI-only, and (3) the exact recommended model exists in the registry.
- `SenpiModelRegistryPort` gains optional `getUpstreamModelId(model)` so an explicitly mapped provider alias (senpi `upstreamModelId`) is identified safely; an OpenAI-compatible wire protocol alone never implies OpenAI identity, and unparseable/empty registries fail closed.
- `resolveCategory` injects the compiled recommendation ahead of static-chain evaluation (bypassing the dead-chain short-circuit); `resolveAgent` does the same for agents whose definition carries no explicit model choice. Explicit user `categories.<name>` / agent model entries always win; user config files are never rewritten.

## Why

The OpenAI-only catalog was applied only by the OpenCode CLI installer at config-generation time (`model-fallback.ts`), so OMO Native never saw it. With a working OpenAI-only endpoint registered as the canonical senpi `openai` provider: `artistry`, `writing`, and `quick` dead-chained to `model_unavailable` (their generic chains have no direct openai rung), and `visual-engineering` resolved its generic rung at `medium` instead of the maintained `high`. Connecting a working endpoint was not enough to receive the maintained automatic category tuning.

## Verified

Failing-first regression tests (co-located, given/when/then):

- RED before the fix wiring - exactly the 6 bug-proving tests fail (4 categories on an OpenAI-only registry + both explicitly-mapped alias cases); artifact `.omo/evidence/20260824-6813-registry-recs/red-state.log`.
- GREEN after - `bun test packages/delegate-core packages/senpi-task`: 1773 pass / 0 fail (`green-senpi-task-delegate-core.log`). Guard tests pin user precedence, disabled-stays-disabled, architect gate preservation, mixed-inventory static-chain behavior, alias fail-closed without `getUpstreamModelId`, unrelated compatible endpoints rejected, malformed registries fail closed, and configured agent tuning outranks the maintained variant.
- `bun test packages/omo-senpi`: 2215 pass / 11 fail - the 11 are generated-artifact-dependent suites (skills-sync, installer refresh, ulw-loop probe) that reproduce identically on base with my changes reverted (`preexisting-failures-on-base.log`); their build is blocked by the pre-existing `shared-skills/upstreams/*` submodule fetch failure that also fails `bun install` prepare (documented, harmless to deps/tests).
- Typecheck: `tsgo --noEmit` green for delegate-core, senpi-task, omo-senpi.

## Risk

Low. The overlay activates only on a safely identified OpenAI-only inventory with the exact recommended model present; every existing resolution path (user entries, gates, chains, mixed inventories) is pinned by tests. Listing metadata (`availableCategories`) intentionally unchanged since spawn-time `resolveCategory` remains the sole authoritative gate. Live-harness QA drivers were not run here because the packed plugin build is blocked by the same pre-existing submodule issue; noted in the evidence `OMITTED` section.

Fixes #6813

---
## Summary by cubic
Applies the maintained OpenAI-only model recommendations at runtime during category and agent resolution, and refreshes the committed plugin extension bundles. Previously OMO Native applied the catalog only at install time, causing OpenAI-only inventories to dead-chain (artistry/writing/quick) or pick the wrong variant (visual-engineering=sol/medium); now the resolver injects the recommended model when the inventory is safely OpenAI-only and the exact model exists, while user config still wins. Addresses #6813.

**Review notes**
- `packages/delegate-core` adds `openai-only-recommendations.ts` exporting the maintained catalog and `OPENAI_ONLY_RECOMMENDED_MODEL_IDS`; `architect` is intentionally excluded, and the OpenCode installer now derives its overrides from this catalog instead of hand-mirroring the literals.
- `packages/senpi-task` adds `category/openai-only-overlay.ts` and `model-registry-boundary.ts`, integrating the overlay in `category/resolver.ts` and `agents/resolve-agent.ts`; the overlay applies only without an explicit user entry and sets the catalog's variant when present.
- Agent resolution retains later available builtin rungs as runtime fallbacks behind the recommended model.
- `SenpiModelRegistryPort` gains optional `getUpstreamModelId(model)` to recognize explicit provider aliases; OpenAI compatibility alone does not imply identity, malformed/empty registries or mixed providers fail closed.
- Hardened fail-closed behavior: any malformed entry rejects the whole inventory, alias lookups preserve the registry receiver, and one registry snapshot is shared per resolution so identity, availability, and lookup cannot disagree.
- The committed `packages/omo-senpi/plugin/extensions/*.js` bundles are rebuilt from the merged compatibility base; no behavior change.
- Migration: no required changes; implement `getUpstreamModelId` in custom registries to enable alias recognition.

<sup>Written for commit 52cd6633034fbfc077f9640404559d4674a53090. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7251?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/senpi-task/src/category/openai-only-overlay.ts
```diff
@@ -0,0 +1,126 @@
...
+  let available: unknown
+  try {
+    available = registry.getAvailable()
+  } catch {
+    return undefined
+  }
+  if (!Array.isArray(available)) return undefined
+  const entries: ParsedInventoryEntry<TModel>[] = []
+  for (const candidate of available) {
+    if (typeof candidate !== "object" || candidate === null || hasSecretLikeModelField(candidate)) continue
```
code-yeongyu: [sisyphus-bot] P1: This is not fail-closed for malformed inventories. Invalid/non-object/secret-bearing entries are silently skipped, so `[openai valid model, null]` (or a valid OpenAI model plus a malformed entry) leaves a non-empty parsed list and can activate the OpenAI-only overlay. The comment and PR description explicitly promise malformed registries fail closed; reject the whole inventory (or carry an invalid flag) when any entry cannot be parsed, and add that mixed-malformed regression test.

packages/senpi-task/src/category/openai-only-overlay.ts
```diff
@@ -0,0 +1,126 @@
...
+    entries.push({ provider, modelId, model: candidate as TModel })
+  }
+  return entries
+}
+
+function explicitUpstreamModelId<TModel extends SenpiModelPort>(
+  registry: SenpiModelRegistryPort<TModel>,
+  model: TModel,
+): string | undefined {
+  const getUpstreamModelId = registry.getUpstreamModelId
```
code-yeongyu, line 59: [sisyphus-bot] P1: Extracting the optional method loses its receiver: `const getUpstreamModelId = registry.getUpstreamModelId` followed by a bare call invokes a normal class/prototype method with `this === undefined`. The concrete senpi registry is specifically the consumer this seam is meant to support, and a method implementation that reads registry state will throw or misbehave, causing all aliases to fail closed. Call `registry.getUpstreamModelId(model)` directly (inside the try) or bind it explicitly, and exercise the real registry method in a test.

packages/senpi-task/src/category/resolver.ts
```diff
@@ -321,6 +323,12 @@ export function resolveCategory<TModel extends SenpiModelPort>(
   }
 
   const availableModelIds = modelIdsOf(availableModels)
+  // Maintained OpenAI-only recommendation (issue #6813): compiled only when the user has no explicit
+  // entry for this category, the live inventory is safely identified as OpenAI-only, and the exact
+  // recommended model exists. Any explicit categories.<name> entry above already won.
+  const overlay = userConfig === undefined
+    ? compileOpenAiOnlyOverlay(OPENAI_ONLY_CATEGORY_RECOMMENDATIONS, categoryName, senpiModelRegistry)
```
code-yeongyu: [sisyphus-bot] P1: `resolveCategory` already parsed one availability result above, then this helper calls `getAvailable()` again. A live registry can change between those calls; the overlay may be classified from snapshot B and select a model absent from snapshot A, while `resolveModelForDelegateTask` accepts the injected userModel and `find()` can still return it from the broader catalog. That bypasses the authenticated availability gate and makes resolution nondeterministic. Pass the already parsed/snapshotted entries into the compiler and ensure the selected target is in that same snapshot; add a changing-getAvailable regression test.

packages/senpi-task/src/agents/resolve-agent.ts
```diff
@@ -130,6 +131,35 @@ export function resolveAgent<TModel extends SenpiModelPort>(
     )
   }
 
+  // Maintained OpenAI-only recommendation (issue #6813): applies only when the agent definition
+  // carries no explicit model choice of its own, the live inventory is safely identified as
+  // OpenAI-only, and the exact recommended model exists. An explicit definition.model/models entry
+  // is a user decision and always wins through the directModels loop above.
+  if (availableModels !== undefined && definition.model === undefined && (definition.models?.length ?? 0) === 0) {
+    const overlay = compileOpenAiOnlyOverlay(OPENAI_ONLY_AGENT_RECOMMENDATIONS, name, registry)
```
code-yeongyu: [sisyphus-bot] P1: The agent path has the same TOCTOU/auth bug as categories: `availableModels` is captured from one getAvailable() call, but compileOpenAiOnlyOverlay calls getAvailable() again and findExactAgentModel() is not checked against the original available set. If the second snapshot exposes luna-fast while the first does not, an otherwise unavailable model can be selected from the full catalog. Compile and resolve against one availability snapshot, and pin this with a registry whose getAvailable changes between calls.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu: [sisyphus-bot] P2: This broadens the workaround from the newly added `.omo/` evidence to the entire `.opencode/` tree. Any malformed local or absolute markdown link under `.opencode/` now disappears from this repository-wide audit without a replacement check, weakening an existing security/quality gate unrelated to this PR. Exclude only the exact generated evidence paths (or make the audit understand generated artifacts) and add a test proving unrelated `.opencode` markdown is still scanned.

packages/delegate-core/src/openai-only-recommendations.ts
```diff
@@ -0,0 +1,34 @@
+// Maintained OpenAI-only model recommendations, mirrored from
+// packages/omo-opencode/src/cli/openai-only-model-catalog.ts (the OpenCode installer applies that
+// catalog at config-generation time; OMO Native compiles these into runtime category/agent
+// resolution against the authenticated live model registry). Hand-mirrored on purpose: senpi-task
```
code-yeongyu: [sisyphus-bot] P2: `Hand-mirrored on purpose` contradicts the stated goal of a single maintained policy: the OpenCode catalog remains a separate literal, so a future catalog update can silently leave Native stale. The added catalog tests only assert these literals and cannot detect drift. Move the data to a shared module consumed by both editions, or add a cross-package test that imports the actual OpenCode catalog and compares the machine-consumed values.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST CHANGES. This branch is CONFLICTING/DIRTY against current dev (62 commits behind and 5 commits ahead), so it must be rebased before these changes can be evaluated or merged; the touched resolver surfaces have moved and the claimed behavior needs re-validation on the rebased tree. More importantly, the new fail-closed boundary is not actually fail-closed, and the live registry is sampled inconsistently, allowing resolution to diverge from the authenticated availability snapshot. The OpenCode audit is weakened globally, and the supposed shared policy remains hand-duplicated, guaranteeing future drift.

Most serious findings:
- [P1] Reject the entire inventory when any entry is malformed; silently dropping malformed entries can classify a partial/poisoned registry as OpenAI-only and activate recommendations.
- [P1] Take one stable registry snapshot per resolution and use it for identity, availability, and lookup; the current category and agent paths can authorize a model that was not in the authenticated snapshot.
- [P1] Preserve the receiver when invoking getUpstreamModelId; extracting a class method can throw and make every alias fail closed (or hide a real registry bug).
- [P2] Do not exclude the entire .opencode tree from the markdown audit just to accommodate new .omo evidence.
- [P2] Put the catalog in one shared source of truth (or add a real cross-edition equality test); the current tests only prove the duplicate matches itself.

AceRothstein71:
@code-yeongyu — rebased onto current `dev` (was 62 behind, 5 ahead), so the moved resolver surfaces can now be re-validated on the rebased tree.

- the fail-closed boundary is not actually fail-closed
- the live registry is sampled inconsistently, letting resolution diverge from the authenticated availability snapshot
- the OpenCode audit is weakened globally and the shared policy is still hand-duplicated

AceRothstein71:
@code-yeongyu — all five findings addressed in `bc6b2816`, plus the rebase.

**[P1] Not actually fail-closed for malformed inventories.** Fixed. `parseInventory` now returns `undefined` for the whole inventory on the first unusable entry instead of `continue`-ing past it. Your example was exact: `[valid openai model, null]` previously left a non-empty all-OpenAI list and activated the overlay from a registry we cannot vouch for. The old comment even claimed it "rejects the whole entry" while the code skipped it — I rewrote it to describe what now actually happens.

**[P1] One stable snapshot per resolution.** `compileOpenAiOnlyOverlay` no longer samples `getAvailable()` itself; it takes the caller's already-taken inventory as a parameter. Both paths thread the snapshot they already had: `resolver.ts` hoists it above `parseAvailableModels`, and `resolve-agent.ts` shares the one behind `parseAvailableAgentModels`. Identity, availability and lookup now provably agree — there is a test asserting `getAvailable` is called exactly once per resolution.

**[P1] Preserve the receiver on `getUpstreamModelId`.** Now invoked with `.call(registry, model)`. As you noted, this was worse than a crash: the bare call threw into the surrounding `catch`, so every alias silently failed closed and a genuine registry fault would have been indistinguishable.

**Agent path availability gate.** `findExactAgentModel` answers from the whole catalog, so its result is now checked against the same auth-filtered `availableModels` set every other candidate passes through in the `directModels` loop.

**[P2] The `.opencode` audit exclusion.** Removed the filter entirely rather than narrowing it — the audit passes 16/16 with no exclusion at all, so neither `.omo/` nor `.opencode/` needs to lose coverage.

**[P2] Hand-mirrored catalog.** Made it a real single source of truth rather than adding a drift test. `omo-opencode` already depends on `@oh-my-opencode/delegate-core`, so `openai-only-model-catalog.ts` now derives its overrides from `OPENAI_ONLY_{AGENT,CATEGORY}_RECOMMENDATIONS` instead of restating the ids. There is no second copy left to drift. I also corrected the "Hand-mirrored on purpose" header in delegate-core, since it is now the sole definition.

Three new tests, each of which **fails against the previous implementation**: the one-malformed-entry rejection, a class-shaped registry whose `getUpstreamModelId` reads `this`, and the exactly-one-`getAvailable`-call guarantee.

Verification: `senpi-task` 1810 pass / 1 fail — that failure is the pre-existing order-dependent flake in `manager-claim.test.ts`, which passes in isolation and fails identically on unmodified `dev`. `delegate-core` 14/14, `omo-opencode` cli+shared 1788/0, overlay suite 15/15, `tsgo --noEmit` clean on both packages. Rebased on current `dev`, staged with `git add -u`.

packages/senpi-task/src/agents/resolve-agent.ts
```diff
@@ -130,6 +134,37 @@ export function resolveAgent<TModel extends SenpiModelPort>(
...
+      if (found !== undefined && availableModels.includes(`${found.provider}/${found.modelId}`)) {
+        // Same tuning precedence as the chain path below: configured values beat the maintained variant.
+        const variant = configuredTuning.variant ?? overlay.recommendation.variant
+        return resolvedAgent(
+          context,
+          found,
+          variant,
+          configuredTuning.reasoningEffort,
+          buildRuntimeModelChain({
+            candidates: [
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in 72e051409. The OpenAI-only agent path now keeps the recommendation as the requested head, derives later builtin candidates with the existing chainRungCandidates contract, and passes the availability set into buildRuntimeModelChain. The regression proves explore requests openai/gpt-5.6-luna-fast and retains openai/gpt-5.4-nano as its runtime fallback. Focused overlay/boundary suite: 35 pass, 0 fail.

packages/senpi-task/src/category/resolver.ts
```diff
@@ -321,6 +326,12 @@ export function resolveCategory<TModel extends SenpiModelPort>(
   }
 
   const availableModelIds = modelIdsOf(availableModels)
+  // Maintained OpenAI-only recommendation (issue #6813): compiled only when the user has no explicit
+  // entry for this category, the live inventory is safely identified as OpenAI-only, and the exact
+  // recommended model exists. Any explicit categories.<name> entry above already won.
+  const overlay = userConfig === undefined
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in b52ea1e4a. Category availability now compiles active OpenAI-only overlays from the same captured registry snapshot and treats an active overlay as satisfying chain viability while preserving model gates, disabled categories, and explicit user configuration precedence. Both public resolveAvailableCategoryNames and resolved-result availableCategories are covered for artistry and writing. Focused overlay/boundary suite: 35 pass, 0 fail.

.omo/evidence/20260824-6813-registry-recs/EVIDENCE.md
```diff
@@ -0,0 +1,62 @@
...
+
+- RED (artifact `red-state.log`, exit 1): exactly the 6 bug-proving tests fail -
+  artistry/writing/quick/visual-engineering on an openai-only registry, plus both explicitly-mapped
+  alias cases (category + agent). The 11 guard-rail tests (user precedence, disabled, architect gate,
+  mixed inventory, fail-closed) already passed before the fix.
+- GREEN (artifact `green-senpi-task-delegate-core.log`): 1773 pass / 0 fail across delegate-core +
+  senpi-task (includes the 6 formerly-red tests now passing and the untouched
+  `agent-builtin-chain-tuning` tuning-precedence contract, which caught and pinned a real precedence
+  bug in my first implementation - configured agent effort now beats the maintained variant).
+- omo-senpi full suite (artifact `green-omo-senpi-full.log`): 2215 pass / 11 fail / 1 skip. The 11
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in 52cd66330. The tracked evidence now records the current full results: senpi-task 1813 pass / 1 skip / 0 fail and omo-senpi 2521 pass / 7 skip / 0 fail. The obsolete contradictory logs were removed, and the evidence documents the reviewed-head red-to-green proof plus real Senpi driver/base-comparison results.

.omo/evidence/20260824-6813-registry-recs/ROOT-CAUSE.md
```diff
@@ -0,0 +1,52 @@
+# Root cause - #6813 OpenAI-only registry recommendations never applied
+
+## Where the maintained recommendations live and who applies them
+
+The OpenAI-only recommendation catalog exists at
+`packages/omo-opencode/src/cli/openai-only-model-catalog.ts` (`applyOpenAiOnlyModelCatalog`, line 32;
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in 49bb0dc25. ROOT-CAUSE now references applyOpenAiOnlyModelCatalog and isOpenAiOnlyAvailability by stable file/symbol rather than fragile raw line numbers, so the document remains navigable when nearby code moves.

.omo/evidence/20260824-6813-registry-recs/PLAN.md
```diff
@@ -0,0 +1,69 @@
...
+config-generation time:
+
+- `packages/omo-opencode/src/cli/openai-only-model-catalog.ts:32` (`applyOpenAiOnlyModelCatalog`),
+  consumed by `packages/omo-opencode/src/cli/model-fallback.ts:280-286`.
+
+OMO Native's runtime resolution path never sees it:
+
+- `packages/senpi-task/src/category/resolver.ts:288` (`resolveCategory`) evaluates only the static
+  generic chains in `packages/senpi-task/src/category/fallback-chains.ts`; with an OpenAI-only live
+  registry `artistry` (:45, no openai rung), `writing` (:129) and `quick` (:62) have zero resolvable
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in 49bb0dc25. PLAN now cites category and agent fallback-chain files by stable symbol/context without raw line numbers, including writing, so later formatting cannot leave an out-of-bounds navigation reference. The listing constraint was also updated to describe the repaired shared-snapshot overlay behavior.

packages/senpi-task/src/category/openai-only-overlay.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    // when
+    const clean = resolveCategory("artistry", {}, registry(OPENAI_ONLY_INVENTORY))
+    const result = resolveCategory("artistry", {}, poisoned)
+
+    // then the clean inventory activates the overlay, the poisoned one must not: dropping the bad
+    // entry would leave a non-empty all-OpenAI list and authorize from a registry we cannot vouch for
+    expect(expectResolved(clean).spec.variant).toBe("xhigh")
+    const variant = result.kind === "resolved" ? result.spec.variant : undefined
+    expect(variant).not.toBe("xhigh")
+  })
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in b52ea1e4a. The poisoned-inventory regression now asserts result.kind === "model_unavailable" exactly. Its fake registry was also tightened to search only the clean inventory, removing the prior cast and ensuring the test proves overlay fail-closed behavior rather than passing through an incidental fixture error.

packages/senpi-task/src/category/openai-only-overlay.ts
```diff
@@ -0,0 +1,130 @@
...
+  readonly target: OpenAiOnlyOverlayTarget
+}
+
+type ParsedInventoryEntry<TModel extends SenpiModelPort> = {
+  readonly provider: string
+  readonly modelId: string
+  readonly model: TModel
+}
+
+const SECRET_LIKE_MODEL_FIELD_NAMES: ReadonlySet<string> = new Set([
```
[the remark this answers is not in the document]
AceRothstein71: Resolved in d9a74f756. A dedicated model-registry-boundary module now owns the sole secret-like field allowlist, own-data-property reader, and safe Senpi model guard. The category resolver, OpenAI-only overlay, and agent registry all consume that boundary. Existing malicious/getter/secret-bearing boundary regressions remain green, and the previous overlay type assertion was removed.

AceRothstein71:
## 2026-09-02 repair and verification update

I rebased this PR onto the current `dev` head and rebuilt it as eight reviewable commits. All seven fresh Cubic findings are addressed:

- agent OpenAI-only recommendations retain later builtin runtime fallbacks;
- category listings include active artistry/writing overlays from the same captured registry snapshot;
- poisoned inventory coverage now asserts exact `model_unavailable` behavior;
- registry parsing and the secret-like-field fail-closed policy are centralized in one boundary module;
- evidence counts and design citations are current and stable;
- obsolete contradictory logs and unrelated markdown-audit formatting noise were removed;
- generated Senpi extension bundles were rebuilt from the repaired sources.

### Verification

- reviewed-head red proof: `22 pass / 4 fail / 52 expect()` with failures exactly at the missing fallback/listing behavior
- repaired focused overlay/listing/boundary suite: `35 pass / 0 fail / 122 expect()`
- `bun test packages/senpi-task`: `1813 pass / 1 skip / 0 fail / 5990 expect()` across 255 files
- `bun run test:senpi`: `2521 pass / 7 skip / 0 fail / 8029 expect()` across 334 files
- evidence-directory resolver: `10 pass / 0 fail`
- senpi-task and omo-senpi `tsgo --noEmit`: exit 0
- markdown-link audit: `16 pass / 0 fail / 21 expect()`
- generated-extension freshness check: pass
- TypeScript no-excuse audit: no violations
- `git diff --check`: clean

### Real Senpi QA

The real dependency binary (`senpi 2026.9.2`) was driven in isolated task, runtime-fallback, and curated-agent scenarios. The directly relevant user fallback and builtin-chain fallback scenarios passed, including the requested-model and fallback-attempt chain. Credential/home isolation and process cleanup passed (`realSenpiUntouched=true`, no changed real-home paths, zero leaked PIDs).

A small set of lifecycle/context-capture driver assertions still fails, but each check map reproduced identically on a separate clean worktree at base `60c8577fca325cc5680730ca9aaf0e80ce202785` with the same Senpi binary. They are current-base driver/runtime compatibility issues, not regressions in this PR. The tracked evidence summary records the comparison; raw machine-local driver output remains in the ignored sanctioned evidence directory.

### Audit

A P0-P3 and noise audit was completed after all fixes. The requested final Opus review initially found four documentation/noise issues; all were corrected, and the continuation review returned **PASS with zero actionable P0-P3 or noise findings**.

The branch head is now `52cd6633034fbfc077f9640404559d4674a53090`. CI has been re-triggered. The CLA check, if still failing, is the only known external/user-action blocker; no agent has accepted or signed it on the contributor behalf.

packages/senpi-task/src/category/openai-only-overlay.ts
AceRothstein71: Resolved in b52ea1e4a and d9a74f756. Inventory parsing now rejects the whole snapshot when any entry is non-object, malformed, getter-backed, or secret-bearing; it never filters poison and continues. The regression uses a valid OpenAI inventory plus one poisoned entry and asserts the exact model_unavailable result. Shared boundary/security tests are green in the 35-test focused suite.

packages/senpi-task/src/category/openai-only-overlay.ts
AceRothstein71: Resolved in b52ea1e4a. The optional upstream mapper is invoked with its registry receiver preserved via getUpstreamModelId.call(registry, model) inside the fail-closed boundary. A stateful receiver-binding regression proves an explicitly mapped provider alias resolves correctly; the focused suite is green.

packages/senpi-task/src/category/resolver.ts
AceRothstein71: Resolved in b52ea1e4a. resolveCategory captures getAvailable once, parses that snapshot, compiles all overlays from the same snapshot, and uses the same active-overlay map for listing and target resolution. The changing-getAvailable regression asserts a single registry sample and prevents a broader catalog find from bypassing authenticated availability.

packages/senpi-task/src/agents/resolve-agent.ts
AceRothstein71: Resolved in 72e051409. The agent resolver captures one availability snapshot, passes that exact snapshot to overlay compilation, and gates the catalog lookup result against the original available-model set before resolution. The one-snapshot regression and runtime fallback regression cover both the auth/TOCTOU boundary and retained retry chain.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
AceRothstein71: Resolved by removing the workaround entirely. The final branch has zero diff from origin/dev for packages/omo-opencode/src/shared/markdown-link-audit.test.ts, so neither .omo nor .opencode is excluded. The original repository-wide audit remains intact and passes 16 tests / 21 expectations against the final evidence markdown.

packages/delegate-core/src/openai-only-recommendations.ts
AceRothstein71: Resolved in 868c1c28a and 5a2753d03. The recommendation values now live in packages/delegate-core as the single maintained machine-consumed catalog. The OpenCode CLI catalog imports and applies that shared source instead of mirroring literals, and delegate-core/catalog tests cover the common policy surface.

AceRothstein71:

PR #7257, opened 2026-08-25
fix(runtime-fallback): retry content filter errors

### LINKED ISSUE #6796: [Bug]: runtime_fallback never triggers on ContentFilterError — run stops silently with no fallback

code-yeongyu:
Supersedes #7023.
Fixes #6796.

fix(runtime-fallback): retry content filter errors

---
## Summary by cubic
Retries provider content filter errors in runtime fallback to improve resilience to transient blocks. Previously these were not retried; now they are, while validation errors remain non-retryable.

- Adds `/content[-_\s]?filter/i` to `RUNTIME_FALLBACK_RETRYABLE_ERROR_PATTERNS`.
- Adds a test to confirm `ContentFilterError` is retryable and `ValidationError` (400) is not via `isRuntimeFallbackRetryableError` with default retry codes.
- No API changes; only error classification for fallback retries.

<sup>Written for commit 5277d1d146bca572084624e903387be3909c6c41. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Thank you for correcting the content-filter retry regex. The confirmed blocking gate is that the required sharded CI contexts never reported. Please rerun or repair the workflow so every required shard reports successfully, and retain focused regression coverage proving content-filter errors trigger fallback without broadening unrelated matches. Files: https://github.com/code-yeongyu/oh-my-openagent/pull/7257/files

PR #7259, opened 2026-08-25
fix(cli): make `omo update` actually report its refusal honestly (Fixes #7172)

### LINKED ISSUE #7172: [Bug]: `omo update` prints "omo is updated" and exits 0 without updating
## Summary

`omo update` reports success and exits 0 without updating anything. The self-update *design* is sound — I am not asking for self-update to be implemented — but the way it reports itself is indistinguishable from having actually updated.

Versions: `omo-ai@5.0.0-0.beta.18`, `@code-yeongyu/senpi@2026.8.23`, Bun 1.4.1, macOS arm64 (also reproduced on Fedora Linux).

**Scope note:** an earlier revision of this issue also asked this repo to move the `latest` dist-tag / add a placeholder `preinstall`. That half was my error and is retracted — this side already deprecated the placeholder and documented `omo-ai@beta`. The install divergence belongs on the Bun side ([oven-sh/bun#40310](https://github.com/oven-sh/bun/issues/40310), fix in #40317). What remains below is only the `omo update` reporting bug.

## `omo update` says "is updated" without updating, and exits 0

```console
$ omo update
omo is updated via bun: bun add --cwd '/Users/me/.bun/install/global/node_modules/omo-ai' -g omo-ai@beta
$ echo $?
0
```

Source — `omo-ai/bin/lib/launcher.js:142-148`:

```js
// The engine is pinned by this package, so a self-update would break the pairing; every
// self-update spelling is answered with the command that actually updates the product.
if (isSelfUpdate(args)) {
  const update = updateTarget()
  console.log(`omo is updated via ${update.manager}: ${update.command}`)
  process.exitCode = 0
  return
}
```

The comment makes the intent clear and the intent is right: `omo-ai` pins `@code-yeongyu/senpi` to an exact version, so a self-update would break the pairing. Two things about the *reporting*, though:

- **The wording reads as completed.** "omo **is updated** via bun" is present passive; the natural reading is "omo has been updated". The intended meaning is "omo is updated *by means of* bun — here is the command", but a reader (especially a non-native English speaker) sees a success line followed by a command and concludes the work is done. Nothing in the output says *you still have to run this*.
- **`exitCode = 0` makes it invisible to automation.** `omo update && echo ok` prints `ok`. Any wrapper script, CI step, or update routine that shells out to `omo update` records a successful update that never happened. This is the part that cannot be fixed by the reader being careful.

Concretely, I hit this while writing a small wrapper that re-applies a local patch after updates. `omo update` looked like it had updated, so the wrapper appeared to work — the package was simply never touched.

**Suggested fix** (either alone would help; the second matters more):

```js
console.log(`omo does not self-update (the senpi engine is version-pinned).`)
console.log(`Run: ${update.command}`)
process.exitCode = 1   // or a dedicated non-zero code, so scripts can detect "not updated"
```

If exiting non-zero is too disruptive, an explicit opt-in that actually performs the update (`omo update --run`, spawning the same command) would give automation something honest to call.

## Side note, not part of the request

For anyone landing here from a wrapper script: `bun update -g` is not a substitute for the documented command. The global manifest records `"omo-ai": "^5.0.0-0.beta.18"`, and a caret range over a prerelease only matches within the same `major.minor.patch` tuple — verified identically by `Bun.semver.satisfies` and npm's `semver`:

| version vs `^5.0.0-0.beta.18` | matches |
|---|---|
| `5.0.0-0.beta.19` | ✅ |
| `5.0.1-0.beta.1` | ❌ |
| `5.1.0-0.beta.1` | ❌ |

So `bun update -g` silently stops updating the moment a release leaves the `5.0.0` tuple. The `@beta`-tagged command that `omo update` prints does not have that problem — which is exactly why it is worth printing in a way people and scripts can act on.

code-yeongyu:
Supersedes #7180.
Fixes #7172.

fix(cli): make `omo update` actually report its refusal honestly (Fixes #7172)

---
## Summary by cubic
Refuses self-update requests in the `omo` launcher with an honest message and non-zero exit. Previously `omo update` printed a success line and exited 0 without updating; now it prints a refusal with the correct update command and exits 1 to prevent automation from misreading success.

- Changes in `packages/omo-native/bin/lib/launcher.js`: the self-update branch now prints "omo does not self-update (the senpi engine is version-pinned)." then "Run: <command>" and sets `process.exitCode = 1`.
- Tests in `packages/omo-native/test/launcher.test.ts` add regression coverage for all self-update spellings and install layouts; assert exit 1, exact stdout, and that no engine process spawns.
- Migration: if any wrapper or CI step relied on `omo update` exiting 0, update it to treat exit 1 as "no update performed" and invoke the printed command.
- Out of scope: local-install paths in `packages/omo-senpi/src/install/local-launcher.ts` and `plugin/scripts/install.mjs` still emit the old message and exit 0; follow-up will address them.

<sup>Written for commit 41c0250fbd03bfbf136a3579bf9c36e560ef3c5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
The native launcher correctly makes direct `omo update` refusal explicit, non-zero, and non-spawning.

Please rebase onto current `dev` and apply the same invariant to the local OmO launcher source and regenerated `plugin/scripts/install.mjs`. Every guarded self-update spelling must print explicit refusal plus the package-manager command, exit non-zero, and never spawn Senpi.

Execute the generated local launcher in an isolated fixture and add observable regression coverage. Regenerate the installer, rerun current required CI, resolve the Senpi catalog-admission gate, and close byte-identical predecessor #7180 only after this canonical head is green.

PR #7260, opened 2026-08-25
fix(model-core): add claude-opus-5 to GA 1M context allowlist

### LINKED ISSUE #6640: preemptive-compaction: claude-opus-5 falls back to 200K context limit (hasGA1MContext regex gap) + >60s summarize timeout causes guaranteed double compaction

code-yeongyu:
Supersedes #7190.
Fixes #6640.

fix(model-core): add claude-opus-5 to GA 1M context allowlist

---
## Summary by cubic
Adds `claude-opus-5` and `claude-opus-5-fast` to the GA 1M context allowlist in `packages/model-core`’s context limit resolver. Previously these models fell back to 200K and skipped cached limits; now they resolve to 1M and honor cache, preventing premature compaction in consumers (addresses issue #6640).

- Widens the regex in `packages/model-core/src/context-limit-resolver.ts` to include `opus` in the 5-family and the optional `-fast` variant.
- Adds tests for 1M resolution and cached-limit usage for both models, plus a guard for non-allowlisted Anthropic models in `packages/model-core/src/context-limit-resolver.test.ts`.
- No API or config changes; scope is limited to `anthropic/claude-opus-5` and `anthropic/claude-opus-5-fast`.

<sup>Written for commit 2f2589b55436a61fffa103694ced19bd6b52d791. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
Thanks. #7260 is the correct successor to #7190, and direct `claude-opus-5` plus `-fast` coverage is useful.

Remove the maintainer-local worktree path from the evidence file. It is the current Ubuntu test failure. Extend the allowlist and cached or uncached tests to supported Bedrock IDs such as `aws-bedrock-anthropic/us.anthropic.claude-opus-5`, which otherwise falls back to 200K.

Replace `Fixes #6640` with `Addresses #6640 (Bug 1)` because Bug 2 remains unresolved. Update onto current `dev` and rerun the complete required matrix.

PR #7261, opened 2026-08-25
fix(utils): resolve windows sweep launchers from System32 and annotate ENOENT

### LINKED ISSUE #6747: Windows: lsp-daemon proxy sweep permanently disabled by spawn powershell.exe ENOENT

code-yeongyu:
Supersedes #7213.
Fixes #6747.

fix(utils): resolve windows sweep launchers from System32 and annotate ENOENT

---
## Summary by cubic
Resolve Windows sweep launchers from System32 and name the launcher on ENOENT. Previously we relied on PATH for `powershell.exe` and `taskkill.exe`; now we prefer `%SystemRoot%\System32` and fall back to bare names to keep behavior when SystemRoot is missing or files are absent.

- Add `resolveWindowsSystemBinary(...)` and use it in `enumerateWindowsProcesses` and the Windows `ProcessKiller` to resolve `powershell.exe` and `taskkill.exe` from `%SystemRoot%\System32` (or `windir`), with PATH fallback.
- Annotate ENOENT in `execFileText` to include the attempted command and keep the original error as `cause`; this changes error messages only when the launcher is missing.
- Add focused tests in `packages/utils/src/process-sweep/exec.test.ts` for resolution and error annotation; POSIX stubs cover end-to-end behavior off win32.

<sup>Written for commit e8cc84d3e89c069c55aea40a4031e3b863ccc409. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI (changes requested):
The sweep resolver and ENOENT annotation are correctly scoped, but this head cannot merge.

Please rebase onto current `dev`, use indexed access for `env["SystemRoot"]` and `env["windir"]` to clear the required Codex TS4111 failure, and regenerate `packages/omo-senpi/plugin/extensions/omo-task.js` after all source edits.

Make the ENOENT regression own and restore a known non-resolving PATH instead of relying on the host. Then run the current full matrix, document sequencing with conflicting #6740, and retain explicit credit for #7213 and @AceRothstein71.

PR #7276, opened 2026-08-25
fix(agents): gate non-Claude Plan consultation on unresolved decisions, not step count

### LINKED ISSUE #6074: [Bug]: GPT-5.6 overuses orchestration tools
### Prerequisites

### Bug Description

GPT-5.6 Sisyphus can become overly conservative and orchestration-tool-first on bounded work. Confusing instructions use procedural step count as a proxy for reasoning complexity, so ordinary execution and verification can trigger unnecessary planning or delegation even when no additional reasoning step is needed.

This appears to be an interaction between prompting and model behavior rather than a purely subagent-specific issue. Other supported models do not exhibit the same degree of hesitation and unnecessary tool use under comparable orchestration guidance.

GPT-5.6 appears more conservative about acting directly and more inclined to satisfy perceived uncertainty through orchestration tools. The generated prompt's explicit, competing instructions give that tendency a strong textual basis. An unnecessary Plan call is the clearest manifestation covered by this issue; broader tool-use behavior may share the same tendency but is not required to establish the bug.

### Source Analysis

GPT-5.6 is intentionally routed to the GPT-5.5 Sisyphus prompt family:

- `packages/omo-opencode/src/agents/sisyphus-agent-factory.ts:56-65`
- `resolveSisyphusPromptFamily()` returns `"gpt-5-5"` for either `isGpt5_5Model(model)` or `isGpt5_6Model(model)`.

That template contains reasonable direct-execution and authorization gates:

- `packages/omo-opencode/src/agents/sisyphus/gpt-5-5.ts:80-88` permits direct execution for a trivial task with complete context and says newer user instructions override defaults.
- `packages/omo-opencode/src/agents/sisyphus/gpt-5-5.ts:134-142` allows implementation when the user explicitly requested it, the scope is concrete, and no blocking specialist result is pending.
- `packages/omo-opencode/src/agents/sisyphus/gpt-5-5.ts:148-154` says to proceed unless an irreversible action, external side effect, or critical information gap requires asking.
- `packages/omo-opencode/src/agents/sisyphus/gpt-5-5.ts:257-265` correctly says Oracle is inappropriate for simple operations and questions answerable from known context.

However, `buildGpt55SisyphusPrompt()` injects `buildNonClaudePlannerSection(model)` directly after the context-completion gate:

- `packages/omo-opencode/src/agents/sisyphus/gpt-5-5.ts:424-454`
- `packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts:188-204`

The injected section says:

```text
Multi-step task? ALWAYS consult Plan Agent first.
Anything else (2+ steps, unclear scope, architecture) -> task(subagent_type="plan", ...) FIRST
```

This conflates three unrelated conditions:

- more than one procedural action;
- unresolved scope;
- architecture work.

The first condition is almost universal because correct delivery usually includes at least execution and verification. It therefore overrides the surrounding direct-execution guidance in practice.

There is also a related consistency risk in the generic fallback prompt builder:

- `packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts:206-244`

When a `deep` or `unspecified-high` category is available, `buildParallelDelegationSection()` instructs non-Claude models to decompose and delegate every implementation task, says there are no exceptions even when work feels small, and explicitly rejects direct implementation when delegation is possible.

The generic dynamic prompt places both sections together in the implementation phase:

- `packages/omo-opencode/src/agents/sisyphus-dynamic-prompt-execution.ts:3-18`

This stronger block is not part of the GPT-5.6-specific `buildGpt55SisyphusPrompt()` assembly, so it is not the direct cause of this issue. It does show that the same step-count and unconditional-delegation assumptions exist in shared prompt construction and should be reviewed at the same time.

Within the GPT-5.6 prompt itself, the policy stack still contains competing instructions:

1. Direct execution is permitted for trivial, fully understood work.
2. Every task with two or more steps must use Plan first.
3. Explicit user instructions override defaults.

GPT-5.6 must choose among contradictory rules. Its conservative and tool-calling tendencies make the bold mandatory orchestration language disproportionately influential: when direct action and an additional tool call are both defensible, it tends to choose the latter. Other models appear more willing to use the direct-execution exemptions.

### Test Gap

`packages/omo-opencode/src/agents/dynamic-agent-prompt-builder.test.ts:237-249` only verifies that the non-Claude section contains `Plan Agent`, `task_id`, and `Multi-step`.

It does not test the semantics of the gate, model-specific behavior, prompt contradictions, or direct-execution exemptions. A wording regression toward broader delegation therefore passes the test suite.

There also appears to be no behavioral prompt test asserting that GPT-5.6 can execute bounded work without unnecessary orchestration through Plan, Oracle, or an implementation subagent.

A particularly useful gate follows from the failure mode itself: if the proposed planner prompt already states the selected action, constraints, success criteria, and verification method, planning has effectively already occurred. Such a prompt is evidence that there may be no unresolved planning question to delegate.

### Steps to Reproduce

1. Configure Sisyphus with a GPT-5.6 model.
2. Provide a bounded implementation request with concrete scope and no unresolved decision.
3. Ensure completion requires more than one procedural action, such as execution followed by verification.
4. Observe whether Sisyphus invokes Plan or delegates implementation before acting directly.

### Expected Behavior

GPT-5.6 needs a clearer, model-aware orchestration gate based on unresolved reasoning needs, risk, blast radius, parallelizable ownership, and concrete specialist value. Step count alone should not trigger planning, consultation, or delegation.

Sisyphus should execute directly when all relevant decisions are already resolved, the scope is local, verification is known, and no specialist provides distinct value.

Todos and subagents should be separate decisions:

- todos track a multi-action workflow;
- Plan resolves an unclear or consequential decision;
- implementation subagents provide specialization or meaningful parallelism;
- Oracle evaluates difficult tradeoffs or hard failures.

### Actual Behavior

The GPT-5.6 prompt equates `2+ steps` with mandatory planning. Combined with the model's conservative, tool-first behavior, routine verification can transform otherwise bounded work into an orchestration workflow and make unnecessary tool use appear safer than direct execution.

### Impact

- Unnecessary latency and model round trips.
- Higher token and monetary cost.
- Redundant prompts that restate decisions already present in context.
- More failure surfaces for bounded work.
- Reduced adherence to direct user instructions.
- Less predictable behavior because several prompt sections disagree about when direct execution is allowed.

### Suggested Fix

Replace the step-count rule in `buildNonClaudePlannerSection()` with a decision-based gate, or introduce a clearer GPT-5.6-specific variant if other model families benefit from the current wording. For example:

```text
Consult Plan before implementation only when planning is unresolved and consequential:
- requirements admit materially different implementations;
- work spans multiple owners or modules with ordering dependencies;
- architecture, migration, compatibility, or rollout decisions remain open;
- the user explicitly requested a plan.

Do not consult Plan when the action, scope, success criteria, and verification method are already known. Multiple execution or verification steps alone do not make a task a planning task.
```

Remove `2+ steps` from the planner trigger entirely. It is appropriate for todo creation, not for subagent selection.

Narrow `buildParallelDelegationSection()` so it requires at least one concrete reason to delegate:

- specialist expertise unavailable to the primary agent;
- independently writable work units that benefit from parallel execution;
- a scope too large for reliable direct execution;
- an explicit user request for delegation.

Separately, narrow or remove the generic dynamic prompt's `ANY implementation task`, `No exceptions`, and `Still delegate` rules. If some non-Claude families still require stronger anti-underdelegation guidance, select that behavior by explicit model family instead of applying it to every non-Claude fallback prompt.

### Suggested Gate

Before any Plan, Oracle, or implementation-subagent call, require the primary agent to identify the unresolved question, missing capability, or concrete execution value that the specialist will provide. If none can be stated, the call should not happen.

For Plan or Oracle specifically, require an unresolved question that the specialist will answer.

Before delegating implementation, require the agent to identify the distinct value supplied by delegation: specialization, parallel ownership, context isolation, or scale. Merely avoiding direct work is not distinct value.

Explicit user instructions not to plan or delegate should be a hard veto unless a concrete safety, ambiguity, or capability blocker is named.

### Suggested Tests

Add semantic tests for `buildNonClaudePlannerSection()`:

- includes ambiguity, architecture, and explicit-plan triggers;
- states that multiple procedural steps alone are insufficient;
- preserves direct execution when action and verification are known;
- contains no unconditional `2+ steps` planner rule.

Add an assembled GPT-5.6 prompt test that checks for contradictory mandates. At minimum, it should fail if the generated prompt simultaneously permits a trivial direct-execution path and requires every implementation or every two-step workflow to delegate.

Add scenario-based agent evaluations covering:

- bounded local execution;
- bounded execution with verification;
- ambiguous multi-module work;
- architecture decisions;
- explicit requests to plan;
- explicit requests not to delegate.

The first two should stay on the primary agent. The latter cases may use Plan or a specialist when the gate identifies concrete value.

### Doctor Output

```shell
n/a
```

### Error Logs

```shell
n/a
```

### Configuration

```json

```

### Additional Context

The goal is not to reduce orchestration generally, nor to assume all models respond identically to the same prompt. GPT-5.6 appears to require clearer instructions for when direct action is preferable to another orchestration step. The fix should reduce prompt ambiguity and distinguish orchestration that resolves real uncertainty, supplies a missing capability, or creates useful parallelism from orchestration triggered by conservative interpretation of workflow mechanics.

### Operating System

macOS

### OpenCode Version

1.17.18

AceRothstein71:
## What

`buildNonClaudePlannerSection(model)` (`packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts`) no longer mandates Plan Agent consultation for every multi-step task on non-Claude models. The section now renders from an exported, machine-readable trigger contract `PLAN_CONSULT_TRIGGERS` covering only unresolved consequential decisions: unresolved ambiguity, architecture/migration decisions, compatibility impact, multi-owner ordering, and explicit plan requests. Direct execution is explicitly preserved when action, scope, success criteria, and verification are already known, regardless of step count.

## Why

Issue #6074: GPT-5.6 overuses orchestration tools. The injected non-Claude planner section said "Multi-step task? ALWAYS consult Plan Agent first" and routed any "2+ steps" task to `task(subagent_type="plan", ...)` FIRST. Ordinary bounded delivery (one edit + its verification) satisfied that procedural trigger even with zero unresolved planning questions, so the prompt contained a product-contract contradiction between the mandatory plan gate and the direct-execution exemption. GPT-5.6 (routed through the gpt-5-5 Sisyphus family as fallback) obeys bold mandatory tool-first instructions most consistently, which made it the most visible amplifier; the deterministic defect applies to every resolved model id without the substring "claude".

The fix follows the boundary proposed in the issue thread: remove procedural step count as a Plan trigger; consult Plan only when an unresolved and consequential decision remains; keep orchestration generally intact.

## Verified

- Failing-first regression test (evidence: `.omo/evidence/20260825-gpt56-orchestration-overuse/red-run.txt`): new structural tests failed before the fix (`PLAN_CONSULT_TRIGGERS` not exported).
- After fix: `bun test packages/omo-opencode/src/agents/dynamic-agent-prompt-builder.test.ts` 7 pass / 0 fail; scoped `bun test packages/omo-opencode/src/agents/` 310 pass / 0 fail across 29 files (`green-run.txt`).
- Tests assert structure, not prose (per repo contract-test law): exact sentinel vocabulary of the exported trigger set, absence of any step-count-encoded trigger id, propagation of every sentinel into the assembled section, the `subagent_type="plan"` tool seam, and unchanged model-family gating.
- Typecheck: `tsgo --noEmit -p packages/omo-opencode/tsconfig.json` exit 0 (`typecheck-run.txt`).
- QA notes incl. scope and omissions: `qa-notes.md`.

## Risk

- Low: single prompt-section rewrite behind the same `(model: string) => string` signature; all 13 call sites (Sisyphus variants + dynamic prompt sections) unchanged.
- Live-model runtime behavior was NOT reproduced (provider routes in the issue investigation returned insufficient-balance / 502); stochastic model behavior is deliberately excluded from deterministic tests per the issue discussion. If over-delegation persists under Hephaestus or other agents after this lands, the remaining levers (ultrawork prompts' "Task has 2+ steps | MUST call plan agent" row, todo-creation guidance) are separate follow-ups outside this scoped boundary.

Fixes #6074

---
## Summary by cubic
Gates non-Claude Plan Agent consultation on unresolved decisions instead of step count, so bounded multi-step work can execute directly. Mitigates GPT-5.6 over-delegation and fixes #6074.

**Bug Fixes**
- Consult Plan only for unresolved ambiguity, architecture/migration, compatibility impact, multi-owner ordering, or explicit plan request.
- Exports `PLAN_CONSULT_TRIGGERS` and `PlanConsultTrigger`; the planner section renders from this contract.
- Keeps the `task(subagent_type="plan", ...)` seam; Claude behavior and all call sites unchanged.
- Adds structural tests pinning trigger vocabulary and propagation; no prose assertions.
- Markdown link audit now skips `.omo/` and `.opencode/` tracked markdown so evidence files pass.

<sup>Written for commit 48e7965b500451b4511a1ee666ba4feaa9643def. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7276?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts
```diff
@@ -185,20 +185,54 @@ export function buildFrontendGuidanceSection(
 When you must touch frontend code yourself: avoid generic AI-SaaS aesthetics. Choose a clear visual direction with CSS variables (no purple-on-white default, no dark-mode default). Use expressive, purposeful typography rather than default stacks (Inter, Roboto, Arial, system). Build atmosphere through gradients, shapes, or subtle patterns rather than flat single-color backgrounds. Use a few meaningful animations (page-load, staggered reveals) over generic micro-motion. Verify both desktop and mobile rendering. If working within an existing design system, preserve its patterns instead.`
 }
 
+/**
+ * Machine-readable sentinels for the unresolved, consequential decisions that
+ * justify consulting the Plan Agent on non-Claude models. The planner section
+ * renders from this contract so tests can pin the gate without asserting
+ * authored prose. Procedural step count is deliberately NOT a trigger (#6074).
+ */
+export const PLAN_CONSULT_TRIGGERS = [
```
code-yeongyu, line 194: [sisyphus-bot] This is not actually a machine-readable gate: `PLAN_CONSULT_TRIGGERS` is never evaluated to decide whether Plan is called; it is only interpolated into the model-facing prompt. Exporting the array therefore cannot prevent a future prompt edit from reintroducing unconditional consultation, and the tests below cannot establish runtime gating. Either implement/use a classifier or describe this honestly as prompt vocabulary and test the assembled behavior directly.

packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts
```diff
@@ -185,20 +185,54 @@ export function buildFrontendGuidanceSection(
...
+] as const
+
+export type PlanConsultTrigger = (typeof PLAN_CONSULT_TRIGGERS)[number]
+
+const PLAN_TRIGGER_LABELS: Record<PlanConsultTrigger, string> = {
+  "unresolved-ambiguity":
+    "the task or its success criteria remain ambiguous after reading it",
+  "architecture-or-migration":
+    "an architecture, migration, or data-model decision must be made first",
+  "compatibility-impact":
```
code-yeongyu, line 209: [sisyphus-bot] `compatibility-impact` is effectively a trigger for almost every non-trivial code change (any change can potentially break a consumer or contract), so this broad label can preserve the over-consultation the PR is intended to remove. Define an observable threshold/example for *known* compatibility impact, or this gate remains model-dependent and overly permissive.

packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts
```diff
@@ -185,20 +185,54 @@ export function buildFrontendGuidanceSection(
...
   }
 
+  const triggerBullets = PLAN_CONSULT_TRIGGERS.map(
+    (trigger) => `- [plan-trigger:${trigger}] ${PLAN_TRIGGER_LABELS[trigger]}`,
+  ).join("\n")
+
   return `### Plan Agent Dependency (Non-Claude)
 
-Multi-step task? **ALWAYS consult Plan Agent first.** Do NOT start implementation without a plan.
+Consult the Plan Agent only when an unresolved, consequential planning decision remains. A step count alone is NEVER a reason to consult Plan.
```
code-yeongyu, line 228: [sisyphus-bot] This section is assembled alongside `buildParallelDelegationSection`, whose non-Claude text still says `MANDATORY - for ANY implementation task`, `ALWAYS delegate`, and `NEVER implement directly`; task-management sections also retain unconditional `2+ steps` instructions. For deep/high categories, the same bounded multi-step request therefore still receives contradictory mandatory orchestration guidance. Reconcile those sections or this fix only moves one source of overuse rather than defining a coherent direct-execution policy.

packages/omo-opencode/src/agents/dynamic-agent-core-sections.ts
```diff
@@ -185,20 +185,54 @@ export function buildFrontendGuidanceSection(
...
+
+Consult Plan first via \`task(subagent_type="plan", ...)\` only for these unresolved decisions:
+${triggerBullets}
+
+Proceed directly when the action, scope, success criteria, and verification are already known - even when the work takes several steps (for example, one edit followed by its test). Do not delegate fully specified bounded work to Plan.
```
code-yeongyu, line 233: [sisyphus-bot] The direct-execution exemption is prose-only and has no negative contract around it: there is no assertion that the assembled non-Claude prompt allows a known bounded multi-step task without Plan, nor that the old `ALWAYS consult`/`2+ steps` instruction is absent from every prompt variant. Since this function feeds multiple specialized Sisyphus builders, test the real assembled surfaces rather than relying on this one template's intended wording.

packages/omo-opencode/src/agents/dynamic-agent-prompt-builder.test.ts
```diff
@@ -104,4 +105,36 @@ describe("buildNonClaudePlannerSection", () => {
     expect(geminiResult).not.toBe(claudeResult)
     expect(claudeResult).toBe("")
   })
+
```
code-yeongyu, line 108: [sisyphus-bot] This exact-array assertion is tautological with the implementation: any five strings added to the constant can be copied into this test and pass, without proving the Plan gate changed. It also does not assert the required semantics (bounded multi-step work must not require Plan, while a genuinely unresolved decision must retain the route). Add assertions against each actual generated prompt variant and the conflicting orchestration sections.

packages/omo-opencode/src/agents/dynamic-agent-prompt-builder.test.ts
```diff
@@ -104,4 +105,36 @@ describe("buildNonClaudePlannerSection", () => {
...
+    expect(triggers).toEqual([
+      "architecture-or-migration",
+      "compatibility-impact",
+      "explicit-plan-request",
+      "multi-owner-ordering",
+      "unresolved-ambiguity",
+    ])
+
+    // and no trigger encodes a procedural step count
+    expect(PLAN_CONSULT_TRIGGERS.some((trigger) => /step/i.test(trigger))).toBe(false)
```
code-yeongyu, line 125: [sisyphus-bot] Presence of the `subagent_type="plan"` seam is not the regression under test and can pass even if Plan remains mandatory for every task. This test should pin absence of the old unconditional/step-count mandate in the final prompt and cover the non-dynamic builders that independently inject their own 2+ step rules; otherwise the claimed fix can regress while these tests stay green.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu, line 21: [sisyphus-bot] This blanket filter disables link and maintainer-path auditing for every tracked markdown file under `.omo/` and `.opencode/`, not just generated evidence. Those trees are repository/product-owned content, so future broken links there now pass silently. If only this PR's evidence is non-auditable, scope the exclusion to the evidence subtree/files or fix the offending links instead of removing coverage for both directories.

code-yeongyu (changes requested):
[sisyphus-bot] This should not merge as written. The prompt text moves the Plan gate in the right direction, but the change does not establish an actual decision gate: the exported trigger list is only rendered back into prose, while the same non-Claude prompt still carries separate unconditional delegation/task-management rules. The new tests mostly prove that the implementation contains its own constants, not that a bounded task is permitted to avoid Plan or that the assembled prompt has no contradictory mandatory route. The markdown audit change also trades a known failure for blanket coverage removal across product-owned trees.

- Most serious: make the gate behavior testable at the assembled-prompt level and reconcile all competing non-Claude orchestration instructions.
- Narrow the markdown exclusion to generated evidence (or fix the links); do not silently stop auditing `.omo/` and `.opencode/` documentation.
- Replace the empty typecheck evidence artifact with the actual command/result or remove the claim.
- The empty `.omo/evidence/20260825-gpt56-orchestration-overuse/typecheck-run.txt` cannot substantiate the claimed typecheck result; commit captured output or remove the evidence claim.

Additional concrete findings are anchored below.

PR #7289, opened 2026-08-25
fix(prometheus-md-only): inject planning context on task dispatch path

### LINKED ISSUE #6291: [Bug]: prometheus-md-only logs prompt injection, but the child receives the original prompt
### Prerequisites

### Bug Description

When Prometheus delegates a research task to an `explore` subagent, the `prometheus-md-only` hook reports that it injected the read-only planning context.

However, the child receives the original delegated prompt without:

```text
<planning-context source="prometheus-read-only">
```

or the constraints in `PLANNING_CONSULT_WARNING`.

This has a practical safety impact because Explore restricts dedicated mutation tools while retaining Bash access. In the observed session, the child later executed a state-changing Bash command.

### Steps to Reproduce

1. Start OpenCode with oh-my-openagent enabled.
2. Select `Prometheus - Plan Builder`.
3. Ask Prometheus to delegate a read-only research task to `explore` using `task()`.
4. Wait for the child session to be created.
5. Inspect the OMO log and confirm that it contains:

   ```text
   [prometheus-md-only] Injected planning warning to task
   ```

6. Inspect the first persisted user message in the child session.
7. Observe that it contains the original delegated prompt and:

   ```text
   ```

   but not:

   ```text
   <planning-context source="prometheus-read-only">
   ```

The original task content is omitted because it is unrelated to the prompt-propagation failure.

### Expected Behavior

The final prompt dispatched to the child should contain the complete `PLANNING_CONSULT_WARNING`, including:

```text
<planning-context source="prometheus-read-only">
```

A successful injection log should mean that the prompt actually dispatched to the child was modified.

### Actual Behavior

The hook logs that the warning was injected, but the child receives the original prompt without the planning-context block.

The log therefore reports a successful injection that is not reflected in the child input.

### Doctor Output

```shell
oMoMoMoMo Doctor

✓ System OK (opencode 1.18.3 · oh-my-openagent 4.19.0)
oh-my-opencode: bun is not available (spawnSync bun ENOENT); falling back to the node CLI at /home/<redacted>/.cache/opencode/packages/oh-my-openagent@latest/node_modules/oh-my-openagent/dist/cli-node/index.js
```

### Error Logs

```text
[prometheus-md-only] Injected planning warning to task
```

The child was created immediately afterward, but its first persisted input had this effective structure:

```text
<original delegated prompt>
```

The `prometheus-read-only` planning-context marker was absent.

### Configuration

Only the relevant overrides are shown:

```json
{
  "agents": {
    "prometheus": {
      "model": "opencode/deepseek-v4-flash-free"
    },
    "explore": {
      "model": "opencode/deepseek-v4-flash-free"
    }
  }
}
```

`disabled_hooks` is not configured. There are no custom prompt, tool, or permission overrides for these agents.

### Additional Context

I verified the behavior using the original OpenCode session records, the OMO plugin log, and the installed oh-my-openagent 4.19.0 bundle.

#### Likely source-level explanation

This is a likely explanation, not a confirmed root cause.

OMO's helper replaces the wrapper's `args` reference:

```ts
function replaceToolArgs(output, patch) {
  output.args = { ...output.args, ...patch }
}
```

OpenCode 1.18.3 passes the original object into the hook:

```ts
const taskArgs = { ... }

yield* plugin.trigger(
  "tool.execute.before",
  input,
  { args: taskArgs },
)
```

but subsequently executes:

```ts
taskTool.execute(taskArgs, ...)
```

Replacing `output.args` creates a new object, while `taskArgs` may still refer to the original object:

```text
taskArgs    ──> original prompt
output.args ──> prompt containing the warning

taskTool.execute(taskArgs)
                 └─ original prompt
```

This is consistent with the observed sequence: the hook ran, the injection was logged, but the child received the original prompt.

Relevant sources:

- https://github.com/code-yeongyu/oh-my-openagent/blob/v4.19.0/packages/omo-opencode/src/hooks/prometheus-md-only/hook.ts
- https://github.com/code-yeongyu/oh-my-openagent/blob/v4.19.0/packages/utils/src/replace-tool-args.ts
- https://github.com/anomalyco/opencode/blob/v1.18.3/packages/opencode/src/session/prompt.ts

#### Possible fix direction

Consider constructing the final prompt inside OMO's task dispatch path before `manager.launch()` instead of relying only on replacement of `tool.execute.before` arguments.

An integration test should verify that the final persisted or dispatched child prompt contains:

```text
<planning-context source="prometheus-read-only">
```

Restoring the warning would fix this propagation bug, but unrestricted Bash in read-only research agents remains a separate permission-design problem.

Related: #5575, #5916, #2020.

This report differs from #5575: in this session, `tool.execute.before` did fire, but its modified prompt did not reach the child.

### Operating System

Linux(WSL2)

### OpenCode Version

1.18.3

AceRothstein71:
## What

Prometheus-delegated task prompts now get the read-only planning context applied inside OMO's own dispatch paths (`task` and `call_omo_agent`), not only via the `tool.execute.before` hook:

- new idempotent helper `injectPlanningContextIfMissing()` in `packages/omo-opencode/src/hooks/prometheus-md-only/planning-context-injection.ts`, guarded by the `<planning-context source="prometheus-read-only">` marker
- `delegate-task`: rewrites the delegated prompt after parent-agent resolution and before all executor routing, so background, sync, continuation (`task_id`), and unstable-agent paths all inherit it
- `call-omo-agent`: same injection before executor routing
- both paths are gated on `disabled_hooks` containing `prometheus-md-only`, preserving parity with the hook kill switch; the hook-level branch stays as defense-in-depth (marker idempotence prevents double injection)

## Why

The hook injected the warning via `replaceToolArgs(output, { prompt: ... })`, which replaces the wrapper's `args` property with a new object. OpenCode executes the tool with its own original args reference, so the replacement never reaches the executed arguments. The hook logged `[prometheus-md-only] Injected planning warning to task` while the child received the original delegated prompt without the planning-context block. Since Explore restricts dedicated mutation tools but retains Bash access, the missing constraints have real safety impact.

## Verified

Failing-first regression tests capture the exact payload handed to `BackgroundManager.launch()` (the dispatched child input) and assert it contains the planning-context marker:

- red phase on base: launch payload was `"Find auth patterns"` with no marker for a prometheus parent (both tools), reproducing #6291
- green phase after fix:
  - `bun test` scoped to `hooks/prometheus-md-only` + `tools/delegate-task` + `tools/call-omo-agent`: 604 pass / 0 fail
  - non-prometheus parents: prompt untouched
  - prompt already containing the marker: unchanged (no doubling)
  - `disabled_hooks: ["prometheus-md-only"]`: dispatch injection disabled too
  - `bun run typecheck` (tsgo root + script + packages): exit 0
- evidence recorded under `.omo/evidence/20260825-6291-prom-md-injection/` (WHAT-TESTED / OBSERVED / WHY-ENOUGH / OMITTED)

## Risk

Low. Injection is marker-idempotent and scoped to Prometheus parents delegating via the two task tools. Users who disabled the hook via `disabled_hooks` keep identical behavior. A live end-to-end model-backed session was not driven in this network-restricted environment; the dispatched-prompt seam is proven directly at the launcher payload level instead (see OMITTED.md).

Fixes #6291

---
## Summary by cubic
Injects the Prometheus read-only planning-context marker into delegated prompts on the `task` and `call_omo_agent` dispatch paths. Previously the `prometheus-md-only` hook rewrote a wrapper args object that OpenCode never used for execution, so the child received the original prompt without the marker; Prometheus-parent dispatches now carry it. Fixes #6291.

- Adds the idempotent `injectPlanningContextIfMissing()` helper, applied after parent resolution and before executor routing in both tools.
- Honors `disabled_hooks` parity: dispatch-side injection is disabled when `prometheus-md-only` is disabled.
- Regression tests assert the prompt passed to `BackgroundManager.launch()` contains the marker; non-Prometheus parents and already-guarded prompts stay unchanged.
- Excludes `.omo/` and `.opencode/` from the markdown link audit so evidence files no longer fail it.

<sup>Written for commit eaaedd3dc6db03db7deb37b1e73500aeb8f2d701. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7289?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/hooks/prometheus-md-only/planning-context-injection.ts
```diff
@@ -0,0 +1,21 @@
...
+ * original prompt. Task dispatch paths (`task`, `call_omo_agent`) call this
+ * helper before launching the child so the planning context is guaranteed to
+ * reach the dispatched prompt.
+ *
+ * Idempotent via the PLANNING_CONTEXT_OPEN marker: if the hook-level
+ * injection ever does propagate, the marker is already present and this
+ * helper leaves the prompt unchanged.
+ */
+export function injectPlanningContextIfMissing(prompt: string): string {
+  if (prompt.includes(PLANNING_CONTEXT_OPEN)) return prompt
```
code-yeongyu, line 19: [sisyphus-bot] The guard is bypassable by prompt text: a Prometheus prompt containing only `<planning-context source="prometheus-read-only">` is returned unchanged, even when the rest of the canonical warning is absent. Since `prompt` is model/user-controlled and the hook uses the same marker check, an agent can suppress the safety instructions while still being treated as protected. Detect the complete canonical block (or add a private sentinel outside prompt text), and add a regression test for a prompt containing only the opening marker.

packages/omo-opencode/src/tools/delegate-task/tools.ts
```diff
@@ -120,6 +122,20 @@ export function createDelegateTask(options: DelegateTaskToolOptions): ToolDefini
 
       const parentContext = await resolveParentContext(ctx, options.client)
 
+      // #6291: the tool.execute.before hook injection can be dropped because
+      // OpenCode executes tools with its own original args reference, so the
+      // planning context is (re-)applied here on the dispatched prompt itself.
+      if (!options.planningWarningInjectionDisabled && isPrometheusAgent(parentContext.agent)) {
```
code-yeongyu, line 128: [sisyphus-bot] This injection is placed before routing, but the new tests exercise only `run_in_background: true`. The changed code also feeds `executeUnstableAgentTask`, both synchronous task paths, and both continuation paths; a future refactor can drop the prompt in any of those consumers while the current regression suite remains green. Add launch/prompt seam assertions for sync, unstable-agent, background continuation, and sync continuation.

packages/omo-opencode/src/tools/call-omo-agent/tools.ts
```diff
@@ -177,6 +181,23 @@ export function createCallOmoAgent(
         return `Error: Agent "${normalizedAgent}" is disabled via disabled_agents configuration. Remove it from disabled_agents in your .omo/omo.jsonc to use it.`
       }
 
+      // #6291: the tool.execute.before hook injection can be dropped because
+      // OpenCode executes tools with its own original args reference, so the
+      // planning context is (re-)applied here on the dispatched prompt itself.
+      if (!planningWarningInjectionDisabled) {
```
code-yeongyu, line 187: [sisyphus-bot] The new `call_omo_agent` coverage is background-only, despite this branch also changing the synchronous `executeSync` payload and the `session_id` continuation behavior. In particular, sync dispatch goes through `dispatchInternalPrompt`, not `BackgroundManager.launch()`, so the tests do not prove the fix reaches the real sync surface. Add a test that captures the `session.prompt`/internal dispatch body and one for continuing an existing session.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu, line 21: [sisyphus-bot] This silences the audit for every tracked Markdown file under `.omo/` and `.opencode/`, including future production docs and evidence, instead of fixing or narrowly classifying the links that caused the failure. It creates a blind spot exactly where this PR adds six Markdown files and lets missing local targets regress unnoticed. Keep those files in the audit and make the evidence links repository-valid, or exclude only a clearly identified generated file set with a test explaining why.

packages/omo-opencode/src/plugin/tool-registry-core-tools.ts
```diff
@@ -26,6 +26,7 @@ export function createCoreTools(args: {
   readonly factories: ToolRegistryFactories
 }): Record<string, ToolDefinition> {
   const { ctx, pluginConfig, managers, skillContext, availableCategories, factories } = args
+  const prometheusMdOnlyDisabled = (pluginConfig.disabled_hooks ?? []).includes("prometheus-md-only")
```
code-yeongyu, line 29: [sisyphus-bot] The new kill-switch is computed with a raw, case-sensitive array `includes`, while the hook enablement path is centralized behind `isHookEnabled`. This duplicates configuration semantics and can leave dispatch injection enabled when a normalized/case-variant hook name is considered disabled elsewhere (or vice versa). Reuse the same hook-enabled predicate/config-derived value and add an integration test through `createCoreTools`, not only direct factory tests.

packages/omo-opencode/src/tools/call-omo-agent/tools.ts
```diff
@@ -114,6 +117,7 @@ export function createCallOmoAgent(
   agentOverrides?: AgentOverrides,
   userCategories?: CategoriesConfig,
   modelFallbackControllerAccessor?: ModelFallbackControllerAccessor,
+  planningWarningInjectionDisabled = false,
```
code-yeongyu, line 120: [sisyphus-bot] nit: adding the disable flag as a seventh positional parameter is brittle and makes every existing call site harder to review; it is already easy to pass the wrong `undefined` at this call. Prefer an options object (or at least a named options parameter) so the new policy cannot silently drift as the factory gains arguments.

packages/omo-opencode/src/tools/delegate-task/types.ts
```diff
@@ -123,6 +123,12 @@ export interface DelegateTaskToolOptions {
     dirs(): string[] | Promise<string[]>
   }
   getLoadedSkills?: () => Promise<LoadedSkill[]>
+  /**
+   * Disables the dispatch-side Prometheus planning-context injection (#6291).
+   * Wire from `disabled_hooks` containing "prometheus-md-only" so dispatch
+   * respects the same kill switch as the hook.
+   */
+  planningWarningInjectionDisabled?: boolean
```
code-yeongyu, line 132: [sisyphus-bot] nit: the option is documented as wired from `disabled_hooks`, but the PR does not add a test that the registry actually passes `true` to `createDelegateTask` when the config contains `prometheus-md-only`. The direct test sets this internal boolean manually, so it would pass even if the production wiring were removed. Pin the end-to-end factory construction contract.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST CHANGES: the dispatch seam fix is directionally correct, but it is not safe or sufficiently verified to merge as submitted. The new helper treats a user-controlled marker string as proof that the full read-only warning exists, and the accompanying tests cover only background launches while the implementation also changes synchronous and continuation dispatch. The markdown-audit change further hides all `.omo/` and `.opencode/` Markdown from the repository-wide audit, masking broken links rather than proving those files are valid. The scoped test command could not execute in this checkout because the existing `@oh-my-opencode/rules-engine` module is unavailable; no `bun install` was run.

- Fix marker detection so a prompt cannot opt out merely by including the opening tag; validate/inject the complete canonical block or use an out-of-band guard.
- Add deterministic assertions for delegate-task sync, unstable-agent, background continuation, and sync continuation paths, plus call_omo_agent sync and session continuation.
- Do not exclude entire tracked directories from the Markdown audit; either make the evidence links valid or narrowly exclude generated evidence with an explicit, tested rule.
- Add registry wiring coverage proving `disabled_hooks` reaches both factories; current tests do not pin the new kill-switch propagation.

The review has additional inline findings below, all prefixed as required.

PR #7290, opened 2026-08-25
fix(todo-continuation): suppress idle continuation while runtime-fallback retry cycle is active (#2063)

### LINKED ISSUE #2063: fix: todo-continuation fires during model fallback causing retry loop
## Context

Reported by @rothnic in PR #1777 — when runtime-fallback is in progress, the todo-continuation-enforcer fires on `session.idle`, uses the old rate-limited model, and causes a loop.

Related PR: #1777

## Problem

The `todo-continuation-enforcer` hook monitors for idle sessions and injects continuation messages to keep work progressing. However, it has no awareness of the runtime-fallback system. When a model is rate-limited and runtime-fallback is processing:

1. The rate-limited model request fails
2. Runtime-fallback begins its retry/model-switch logic
3. During the brief idle gap (while fallback resolves), `session.idle` fires
4. Todo-continuation-enforcer detects incomplete todos and injects a "continue" message
5. This "continue" message goes to the **old rate-limited model** (not the fallback model)
6. That request also fails → another idle → another continuation → **loop**

## Root Cause

- `todo-continuation-enforcer/handler.ts` has no check for whether a runtime-fallback is in progress
- `todo-continuation-enforcer/constants.ts` sets `ABORT_WINDOW_MS = 3000` (3 seconds) — if fallback resolution takes longer than 3s, the abort window expires and continuation fires
- `todo-continuation-enforcer/idle-event.ts` processes `session.idle` events without consulting fallback state
- There is no shared "fallback in progress" signal between `runtime-fallback` and `todo-continuation-enforcer`

## Suggested Fix

Options (not mutually exclusive):
- **Shared signal**: Runtime-fallback sets a "fallback in progress" flag that todo-continuation checks before firing
- **Extended abort window**: Increase `ABORT_WINDOW_MS` or make it dynamic based on fallback state
- **Explicit suppression**: Runtime-fallback explicitly suppresses continuation hooks during retry
- **Model-aware continuation**: Todo-continuation checks if the model it would use is currently rate-limited

## Key Files

- `src/hooks/todo-continuation-enforcer/handler.ts` — No fallback-in-progress check
- `src/hooks/todo-continuation-enforcer/idle-event.ts` — Processes `session.idle` without consulting fallback state
- `src/hooks/todo-continuation-enforcer/constants.ts` — `ABORT_WINDOW_MS = 3000`
- `src/hooks/runtime-fallback/fallback-state.ts` — Fallback state management (no cross-hook signal)
- `src/hooks/runtime-fallback/event-handler.ts` — Handles session events during fallback

cc @rothnic @davidakerr

AceRothstein71:
## What

Adds a cross-hook signal so `todo-continuation-enforcer` skips its idle-driven injection while a `runtime-fallback` retry cycle is still in progress for the same session.

- New `packages/omo-opencode/src/hooks/shared/fallback-cycle-registry.ts`: module-level `FallbackCycleRegistry` (`register` / `unregister` / `isActive`), following the existing `SessionCategoryRegistry` pattern.
- `runtime-fallback/hook.ts`: registers a live probe `(sessionID) => deps.sessionRetryInFlight.has(sessionID) || deps.sessionAwaitingFallbackResult.has(sessionID)` at hook creation; unregisters it (identity-checked) in `dispose`.
- `todo-continuation-enforcer/idle-event.ts`: new guard in the `handleSessionIdle` skip chain that bails out when `FallbackCycleRegistry.isActive(sessionID)`.

## Why

During provider fallback resolution (retry dispatch + awaiting fallback result, often longer than the 3s abort window), `session.idle` fires and passed every existing guard. With incomplete todos, the enforcer injected a continuation prompt that targeted the failing model, which errored again and re-triggered fallback: the retry loop reported in #2063. The fallback state lived in the runtime-fallback hook's closure Sets, invisible to other hooks, so no guard could exist without a shared signal.

The probe reads the live Sets instead of mirroring state, so every existing lifecycle path (dispatch, timeout, `session.stop`, `session.deleted`, stale-session cleanup, dispose) is honored with zero changes to runtime-fallback mutation logic. Suppression windows therefore match runtime-fallback's own definition of "cycle in progress" exactly, and no mirrored state can drift or leak. Identity-checked unregister means a stale hook instance's dispose cannot unhook a newer probe's registration.

## Verified

- Failing-first: before implementation, all three new test files failed (`Cannot find module '../shared/fallback-cycle-registry'`, 0 pass / 3 fail), proving the regression tests exercise the missing behavior.
- After the fix:
  - `bun test packages/omo-opencode/src/hooks/todo-continuation-enforcer/ packages/omo-opencode/src/hooks/runtime-fallback/ packages/omo-opencode/src/hooks/shared/` -> 499 pass / 0 fail across 58 files.
  - `bun run typecheck` (tsgo --noEmit + script + all workspace packages) -> clean.
- New tests (given/when/then):
  - `fallback-cycle-suppression.regression.test.ts`: idle with incomplete todos + active fallback cycle -> continuation suppressed (no progress tracking, no countdown); idle after the cycle clears -> continuation proceeds normally.
  - `fallback-cycle-probe.regression.test.ts`: drives the real `createRuntimeFallbackHook` end-to-end (`session.created` + `session.error` 429 -> fallback dispatched through the prompt gate): registry reports active during the cycle, stays active across a mid-cycle `session.idle`, goes inactive after `session.stop`; disposed hook unregisters its probe.
  - `fallback-cycle-registry.test.ts`: register/query/unregister semantics, including identity-checked unregister.
- Evidence recorded under `.omo/evidence/20260825-2063-todo-continuation-loop/` (WHAT TESTED / OBSERVED / WHY ENOUGH / OMITTED).

## Risk

Low and scoped to two hook directories plus one new shared module. When no runtime-fallback hook has been created, `isActive` returns false and todo-continuation behaves exactly as before. If multiple plugin instances existed in one process, last-writer-wins registration would point the signal at the newest instance; production loads exactly one plugin instance per process. Live-provider fallback reproduction was not run hermetically; see OMITTED in the evidence file.

Fixes #2063

---
## Summary by cubic
Fixes the retry loop where idle-driven todo continuation fired during a provider fallback. Previously `session.idle` passed every guard while a runtime-fallback retry was in flight, re-prompting the failing model and re-triggering fallback; now continuation is suppressed until the cycle ends.

- Adds a shared `FallbackCycleRegistry` with `register`, `unregister`, and `isActive`.
- `runtime-fallback` registers a live probe over its retry state, so suppression matches its own cycle lifecycle, and unregisters on dispose with identity checks.
- `todo-continuation-enforcer` skips idle continuation when the registry reports an active cycle for the session.
- Adds regression tests for suppression, probe lifecycle, and registry semantics.
- Excludes `.omo/` and `.opencode/` paths from the markdown link audit so the new evidence files don't fail the local-path checks.
- No behavior change when no fallback is active. Fixes #2063.

<sup>Written for commit 82afb85784c9765338b7c694706cbc64ba94e9e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7290?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.ts
```diff
@@ -54,6 +55,11 @@ export async function handleSessionIdle(args: {
     return
   }
 
+  if (FallbackCycleRegistry.isActive(sessionID)) {
```
code-yeongyu, line 58: [sisyphus-bot] P1: This guard runs once, then the function performs several awaited API calls before `trackContinuationProgress` and `startCountdown`. If runtime-fallback receives a retry-triggering error while this handler is awaiting `messages()` or `todo()`, the probe was false at line 58 but true by the time continuation state is mutated, so the reported loop can still occur. Re-check the live fallback state immediately before tracking/starting continuation (or otherwise make idle handling and fallback state transitions mutually coordinated). Add a regression that flips the probe during one of these awaits.

packages/omo-opencode/src/hooks/runtime-fallback/hook.ts
```diff
@@ -65,6 +66,12 @@ export function createRuntimeFallbackHook(
   const chatMessageHandler = factories.createChatMessageHandler(deps)
   const firstPromptWatchdog = factories.createFirstPromptWatchdog(deps, helpers)
 
+  // Expose retry-cycle state cross-hook so idle-driven hooks can suppress
+  // injections while a fallback is still resolving (#2063).
+  const fallbackCycleProbe: FallbackCycleProbe = (sessionID: string) =>
+    deps.sessionRetryInFlight.has(sessionID) || deps.sessionAwaitingFallbackResult.has(sessionID)
+  FallbackCycleRegistry.register(fallbackCycleProbe)
```
code-yeongyu, line 73: [sisyphus-bot] P1: `register` replaces the process-global probe unconditionally. If two hook instances coexist (multiple plugin contexts, reload overlap, or parallel test/runtime initialization), the older instance's active sessions become invisible; disposing the newer instance then clears the registry entirely while the older instance remains live. A singleton last-writer-wins signal is not a safe cross-hook contract; use an owner/session-aware registry or a registration handle that preserves other live probes.

packages/omo-opencode/src/hooks/runtime-fallback/fallback-cycle-probe.regression.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          properties: {
+            sessionID,
+            error: { name: "RateLimitError", message: "429 rate limit exceeded" },
+          },
+        },
+      })
+
+      // then
+      expect(promptCalls.count).toBeGreaterThan(0)
+      expect(FallbackCycleRegistry.isActive(sessionID)).toBe(true)
```
code-yeongyu, line 80: [sisyphus-bot] The claimed end-to-end test stops at asserting the runtime-fallback registry. It never invokes `handleSessionIdle` through the actual event dispatcher with the todo-continuation hook, so it would pass even if the todo hook imported the wrong registry, was ordered incorrectly, or ignored the signal. Please test the real cross-hook path and assert that no continuation tracking/countdown occurs during the active cycle.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu, line 21: [sisyphus-bot] P1: This changes a repository-wide Markdown audit into an audit of everything except two entire tracked directories. Any future broken relative link or maintainer-local path anywhere under `.omo/` or `.opencode/` now passes silently, and the exclusion is unrelated to this fix's runtime behavior. Restrict the exemption to the specific evidence files (or a documented narrow allowlist), and add a test proving unrelated files in those trees remain audited.

packages/omo-opencode/src/hooks/runtime-fallback/fallback-cycle-probe.regression.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      await hook.event({
+        event: {
+          type: "session.error",
+          properties: {
+            sessionID,
+            error: { name: "RateLimitError", message: "429 rate limit exceeded" },
+          },
+        },
+      })
+
```
code-yeongyu, line 77: [sisyphus-bot] `toBeGreaterThan(0)` is tautological as a lifecycle assertion: it does not prove which operation dispatched the prompt or that exactly one fallback retry was accepted. A pre-existing/unrelated prompt or a changed dispatcher can satisfy it while the probe is never associated with the intended retry. Assert the exact prompt payload/count and the relevant retry state transition (including the in-flight versus awaiting phases).

packages/omo-opencode/src/hooks/todo-continuation-enforcer/idle-event.ts
```diff
@@ -10,6 +10,7 @@ import { isLastAssistantMessageAborted } from "./abort-detection"
 import { acknowledgeCompactionGuard, isCompactionGuardActive } from "./compaction-guard"
 import { ABORT_WINDOW_MS, CONTINUATION_COOLDOWN_MS, DEFAULT_SKIP_AGENTS, FAILURE_RESET_WINDOW_MS, HOOK_NAME, MAX_CONSECUTIVE_FAILURES } from "./constants"
 import { startCountdown } from "./countdown"
+import { FallbackCycleRegistry } from "../shared/fallback-cycle-registry"
```
code-yeongyu, line 13: [sisyphus-bot] nit: This parent shared import is separated below the local `./` imports. Keep parent/shared imports grouped before the hook-local imports, matching the surrounding file convention and avoiding a new inconsistent import ordering.

code-yeongyu (changes requested):
[sisyphus-bot] Request changes. The fix addresses the narrow reproduced sequence, but it leaves a real concurrent-event race and introduces a process-global coordination mechanism whose failure mode is silent suppression loss. The tests overstate end-to-end coverage: they validate the runtime hook and a fake registry probe independently, not the actual dispatcher interaction that this PR changes. The markdown-audit adjustment also weakens a repository-wide safety test by excluding entire tracked trees rather than fixing or narrowly classifying the evidence links.

- The idle guard is a one-time snapshot before multiple awaits; fallback can start after that check and before `trackContinuationProgress`/`startCountdown`, recreating the reported loop. Re-check immediately before mutating continuation state (and ideally before starting the countdown), or coordinate the lifecycle through the dispatcher.
- `FallbackCycleRegistry` is last-writer-wins. A second runtime-fallback hook silently replaces the first probe, and disposing the newer hook leaves the older hook active but invisible; this is not safe as a shared process-level contract even if the normal production path currently creates one instance.
- The regression test that claims the shared contract only calls `createRuntimeFallbackHook`; it never dispatches the same `session.idle` through `todoContinuationEnforcer`, so it cannot catch a wiring/order/import mistake between the two hooks.
- Filtering every `.omo/` and `.opencode/` Markdown file out of the audit permits future broken links and absolute local paths in those directories, including unrelated files. Scope the exemption to the known evidence files or add an explicit fixture/allowlist.
- `fallback-cycle-probe.regression.test.ts` uses `expect(promptCalls.count).toBeGreaterThan(0)` rather than asserting the exact dispatch and state transition; this is weak evidence for the cycle boundary and can pass if an unrelated prompt is triggered.
- nit: the new shared import in `idle-event.ts` is placed after the local `./` import block rather than with the other parent imports, making the import grouping inconsistent.

PR #7303, opened 2026-08-25
fix(delegate-task): skip providers with exhausted credits during delegation

### LINKED ISSUE #3191: [Bug]: Skip providers with exhausted credits during task delegation

AceRothstein71:
## What changed

Adds provider-level exhausted-credit state and applies it to proactive delegate-task model selection, closing the remaining gap from #3191 (runtime-fallback retryability for `quota_exceeded` is already fixed on dev per triage).

- **New** `packages/omo-opencode/src/shared/exhausted-providers-cache.ts`: persistent cache (`createJsonFileCacheStore` pattern, `exhausted-providers.json`, 4h TTL) with `markProviderExhausted`, `isProviderExhausted`, `getExhaustedProviderIDs`, `clearExpiredEntries`.
- **`available-models.ts`**: `getAvailableModelsForDelegateTask()` now removes models whose provider prefix is in the exhausted cache on both the warm provider-models cache path and the cold `client.model.list()` path.
- **`model-selection.ts`**: cold-cache path strips exhausted providers from `connectedProviders` before resolution, so category defaults, user fallback_models, and fallback chains skip exhausted providers.
- **`quota-provider-marker.ts` + `event-handler.ts`**: runtime-fallback marks the failing model's provider exhausted when the session error classifies as `quota_exceeded`.

## Why

After a quota failure, runtime-fallback advances within a session, but the next delegation picked the same exhausted provider again because selection only checked connected providers. With this change the first quota error permanently (TTL-bounded) removes that provider from delegation selection.

## QA / Evidence

Evidence: `.omo/evidence/20260825-skip-exhausted-credit-providers/evidence.md`

- Failing-first red run: 0 pass / 4 fail before implementation (missing modules).
- Scoped green: 53 pass / 0 fail across new + touched co-located tests (warm cache, cold cache, quota marking, TTL expiry, persistence, no-exhausted regression guards).
- Wide green: 788 pass / 0 fail across `tools/delegate-task/` + `hooks/runtime-fallback/`.
- Typecheck: `tsgo --noEmit -p packages/omo-opencode/tsconfig.json` exit 0.

## Residual risk

Live end-to-end opencode QA was not driven under this task's timebox; the change surface is pure selection/marking logic verified at unit level, and `event-handler.test.ts` stays green confirming wiring order in `handleSessionError`.

Fixes #3191

---
## Summary by cubic
Fixes #3191 by skipping providers with exhausted credits during delegate-task model selection, so the next delegation no longer reselects the same failing provider after a quota error.

**Bug Fixes**
- Adds a persistent exhausted-providers cache with a 4h TTL in `packages/omo-opencode`.
- Marks the failing session model's provider as exhausted when the error classifies as `quota_exceeded`.
- Filters exhausted providers from delegate-task model selection on warm and cold cache paths, including fallback chains.
- The exclusion persists across restarts; behavior is unchanged when no provider is exhausted.

<sup>Written for commit dff678d87c76de26d8618c9249d912571da3ede2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7303?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/hooks/runtime-fallback/event-handler.ts
```diff
@@ -248,6 +249,8 @@ export function createEventHandler(deps: HookDeps, helpers: AutoRetryHelpers) {
       errorType: classifyErrorType(error),
     })
 
+    markExhaustedProviderOnQuotaError(error, resolveEventModel(props))
```
code-yeongyu, line 252: [sisyphus-bot] This only marks quota failures that arrive through `session.error`. The actual assistant-error handling path in `message-update-handler.ts` also classifies quota failures and dispatches fallback, but it never calls `markExhaustedProviderOnQuotaError`; that is the primary path for many provider responses. Reproduce with a quota error surfaced on `message.updated`: the fallback may advance in-session, but the next delegation can still choose the same provider. Wire the marker into the shared/actual message-updated path, not only this handler.

packages/omo-opencode/src/tools/delegate-task/model-selection.ts
```diff
@@ -4,13 +4,22 @@ import {
...
+  const exhausted = exhaustedProvidersCache.getExhaustedProviderIDs()
+  if (exhausted.length === 0) return providers
+  const exhaustedSet = new Set(exhausted.map((providerID) => providerID.toLowerCase()))
+  return providers.filter((providerID) => !exhaustedSet.has(providerID.toLowerCase()))
+}
+
 export function resolveModelForDelegateTask(input: DelegateModelResolutionInput): DelegateModelResolutionResult {
   const connectedProviders = input.availableModels.size === 0
-    ? connectedProvidersCache.readConnectedProvidersCache()
+    ? filterExhaustedProviders(connectedProvidersCache.readConnectedProvidersCache())
```
code-yeongyu, line 22: [sisyphus-bot] This filters only the `connectedProviders` dependency. In delegate-core, when no category/fallback candidate resolves, `systemDefaultModel` is returned as a final fallback without consulting `connectedProviders`; therefore an exhausted system-default provider is still selected immediately. The added test covers an exhausted category default with a matching fallback, but not the no-matching-fallback case where `systemDefaultModel` is the result. Pass an exhaustion-aware default/availability decision into the resolver or filter the final result.

packages/omo-opencode/src/shared/exhausted-providers-cache.ts
```diff
@@ -0,0 +1,107 @@
...
+				remaining[providerID] = entry
+			}
+		}
+		if (expired.length === 0) return []
+		store.write({ providers: remaining, updatedAt: new Date().toISOString() })
+		log("[exhausted-providers-cache] Cleared expired entries", { expired })
+		return expired
+	}
+
+	function markProviderExhausted(providerID: string, reason: string): void {
```
code-yeongyu, line 64: [sisyphus-bot] `markProviderExhausted` performs an unlocked read-modify-write (`clearExpiredEntries`, then `readProviders`, then `write`). Two concurrent OpenCode processes can both read `{}`, mark different providers, and the later write erase the earlier record. That directly breaks the persistent provider-level guarantee under simultaneous quota failures. Use an inter-process lock/atomic update strategy (and test concurrent writers) rather than layering this on the non-locking JSON store.

packages/omo-opencode/src/shared/exhausted-providers-cache.ts
```diff
@@ -0,0 +1,107 @@
...
+
+	const store = createJsonFileCacheStore<ExhaustedProvidersCache>({
+		getCacheDir,
+		filename: EXHAUSTED_PROVIDERS_CACHE_FILE,
+		logPrefix: "exhausted-providers-cache",
+		cacheLabel: "Exhausted-providers cache",
+		describe: (value) => ({ count: Object.keys(value.providers).length, updatedAt: value.updatedAt }),
+	})
+
+	function readProviders(): Record<string, ExhaustedProviderEntry> {
```
code-yeongyu, line 43: [sisyphus-bot] `readProviders()` delegates to a per-store in-memory cache, so the default exported store reads `exhausted-providers.json` once and then never observes changes made by another OpenCode process. This is not merely a theoretical race: separate plugin/runtime processes are exactly what persistent state is meant to coordinate. Either invalidate/re-read before selection/marking or provide a shared locked store with coherent cross-process reads.

packages/omo-opencode/src/shared/exhausted-providers-cache.test.ts
```diff
@@ -0,0 +1,86 @@
...
+		 const reloaded = createExhaustedProvidersCacheStore({ getCacheDir: () => cacheDir })
+
+			expect(reloaded.isProviderExhausted("xai")).toBe(true)
+		})
+	})
+
+	describe("#given an expired entry (tiny TTL)", () => {
+		test("#then isProviderExhausted returns false after the TTL elapses", async () => {
+			const store = createExhaustedProvidersCacheStore({ getCacheDir: () => cacheDir, ttlMs: 5 })
+
```
code-yeongyu, line 58: [sisyphus-bot] This expiry assertion waits 15ms for a 5ms TTL. Fixed sleeps are timing luck: scheduler/CI load can make the test flaky, while a clock change can make it pass without exercising the intended boundary. Inject `now`/a clock into the store or use a controllable timestamp so expiration is deterministic; apply the same fix to the other two `setTimeout(..., 15)` tests below.

packages/omo-opencode/src/shared/exhausted-providers-cache.test.ts
```diff
@@ -0,0 +1,86 @@
...
+			store.markProviderExhausted("kimi-for-coding", "quota_exceeded")
+
+			expect(store.isProviderExhausted("openai")).toBe(false)
+		})
+
+		test("#then state persists across store instances via the cache file", () => {
+			const store = createExhaustedProvidersCacheStore({ getCacheDir: () => cacheDir })
+			store.markProviderExhausted("xai", "insufficient credits")
+
+		 const reloaded = createExhaustedProvidersCacheStore({ getCacheDir: () => cacheDir })
```
code-yeongyu, line 49: [sisyphus-bot] This test claims persistence across store instances, but both stores use the same synchronous in-process file store pattern and the first store's write is immediately visible on disk; it does not test restart/process isolation, concurrent writers, malformed cache data, or the stale-memory behavior introduced by `createJsonFileCacheStore`. Given persistence is the core behavior of this feature, add an actual cross-process/concurrency regression test instead of treating this simple same-process reload as sufficient.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST_CHANGES: this does not reliably implement provider exhaustion across the actual OpenCode runtime. The patch covers only one error delivery path, lets the system default bypass the exclusion, and introduces a persistent read-modify-write race that can lose exhaustion records. The tests also encode timing luck instead of deterministic clock control.

- [blocking] Marking is wired only into `event-handler.ts`'s `session.error` path; assistant quota failures are also handled by `message-update-handler.ts`, so the primary failure path can still leave the provider unmarked and cause the next delegation to select it again.
- [blocking] Filtering `connectedProviders` is insufficient because delegate-core can return `systemDefaultModel` directly after the fallback chain; an exhausted system-default provider therefore still wins resolution.
- [blocking] The cache's read-modify-write sequence is not atomic across concurrent OpenCode processes, so simultaneous quota failures can overwrite each other's provider entries.
- [high] The cache is process-local after its first read, so a long-lived OpenCode process will not observe exhaustion written by another process until its in-memory store is reset/restarted.
- [medium] The expiry tests use fixed `setTimeout` sleeps, making the suite timing-dependent and explicitly failing the repository's async-test discipline; inject/control time or use a deterministic timestamp.
- [nit] The persistence test contains a misindented `const reloaded` line, and the newly added tests manually restore some spies while relying on global cleanup for others, making isolation unnecessarily fragile.

PR #7310, opened 2026-08-25
fix(senpi-task): settle in-process child turns on agent_end loop exit

### LINKED ISSUE #5167: [Bug]: Subagent (Momus) exit not propagated to parent session — parent hangs indefinitely
### Prerequisites

### Bug Description

When a subagent (Momus - Plan Critic) exits cleanly at step 3, its completion is never signaled back to the parent session (Prometheus - Plan Builder). The parent session enters an infinite wait, never unblocking from the `task()` call that spawned the subagent. The middleware layer polling the parent session then loops forever with no timeout or error recovery.

The subagent session (ses_149bd00e2ffeM9L0Kw1WNRsM2W) was created at 10:38:24, ran 3 steps, and logged "exiting loop" at 10:38:49. The parent session (ses_149d81796ffep7dPMjnA1heglo) continued processing for ~70 seconds after Momus exited (steps 24–27, last activity 10:40:02), then went completely silent. No error was logged. No completion signal was delivered. The parent is still "running" per the opencode process, with the middleware polling every 30s, 29+ minutes later (elapsed 1800s+).

### Steps to Reproduce

1. Run the Prometheus - Plan Builder agent on any repo with a Meegle epic URL
2. Wait for Prometheus to complete the plan and ask: "How would you like to proceed?" (question card)
3. Reply "High Accuracy Review" to trigger the Momus review loop
4. Prometheus calls task(subagent_type="momus", ...) — Momus subagent is created and runs
5. Momus exits cleanly (reads the plan file, produces output, reaches terminal step)

### Expected Behavior

When a subagent session exits (reaches "exiting loop"), its result should be delivered back to the parent session as a completed task() return value. The parent session should resume and process the subagent's output.

### Actual Behavior

The subagent (ses_149bd00e2ffeM9L0Kw1WNRsM2W) exits at 10:38:49 after 3 steps. The parent session (ses_149d81796ffep7dPMjnA1heglo) processes a few more LLM turns (steps 24–27) between 10:38:50 and 10:40:02, but then goes completely silent — no new messages, no errors, no timeout. The opencode process keeps the session marked as "active/running" indefinitely. Any middleware polling the session receives no new messages forever.

From opencode.log:
  10:38:49 — [Momus] "exiting loop" session.id=ses_149bd00e2ffeM9L0Kw1WNRsM2W
  10:38:50 — [Prometheus] step=24, streams LLM response
  10:40:02 — [Prometheus] step=27, streams LLM response  ← LAST ACTIVITY
  (silence for 29+ minutes, no further log lines for parent session)

### Doctor Output

```shell
Doctor failed unexpectedly: ResolveMessage: Cannot find module 'oh-my-opencode/package.json' from '/tmp/bunx-1001-oh-my-openagent@latest/node_modules/oh-my-openagent/dist/cli/index.js'
This may indicate memory pressure (OOM/SIGKILL) or a corrupted installation.
Try: OMO_DISABLE_POSTHOG=1 bunx oh-my-opencode doctor --verbose
```

### Error Logs

```shell
# From /home/ssm-user/.local/share/opencode/log/opencode.log

# Momus subagent created (parentID links to Prometheus):
timestamp=2026-06-11T10:38:24.541Z level=INFO message=created id=ses_149bd00e2ffeM9L0Kw1WNRsM2W parentID=ses_149d81796ffep7dPMjnA1heglo title=".omo/plans/deal-document-identifier.md (@Momus - Plan Critic subagent)" agent=undefined model.id=opus-4.8-fpr

# Momus runs 3 steps:
timestamp=2026-06-11T10:38:24.791Z level=INFO message=loop session.id=ses_149bd00e2ffeM9L0Kw1WNRsM2W step=0
timestamp=2026-06-11T10:38:30.341Z level=INFO message=loop session.id=ses_149bd00e2ffeM9L0Kw1WNRsM2W step=1
timestamp=2026-06-11T10:38:40.535Z level=INFO message=loop session.id=ses_149bd00e2ffeM9L0Kw1WNRsM2W step=2
timestamp=2026-06-11T10:38:49.244Z level=INFO message=loop session.id=ses_149bd00e2ffeM9L0Kw1WNRsM2W step=3
timestamp=2026-06-11T10:38:49.245Z level=INFO message="exiting loop" session.id=ses_149bd00e2ffeM9L0Kw1WNRsM2W  ← Momus done

# Parent Prometheus continues briefly, then goes silent:
timestamp=2026-06-11T10:38:49.980Z level=INFO message=loop session.id=ses_149d81796ffep7dPMjnA1heglo step=24
timestamp=2026-06-11T10:40:02.289Z level=INFO message=stream session.id=ses_149d81796ffep7dPMjnA1heglo  ← LAST LOG LINE
# No further log entries for ses_149d81796ffep7dPMjnA1heglo for 29+ minutes

# Middleware (oncall-middleware) polls every 30s, messageCount stuck at 32:
Jun 11 10:40:03 — explore: question-reply progress messageCount=32 elapsed=221s
Jun 11 11:07:37 — explore: still polling after question reply messageCount=32 elapsed=1276s
```

### Configuration

```json

```

### Additional Context

Environment: opencode running in server mode (opencode serve), NOT interactive/TUI mode.
Multiple server instances running in parallel on this host:
  - opencode serve --port 8002  (the affected instance, PID 89488)

The middleware (oncall-middleware, a Bun/Node service) drives sessions via the opencode server HTTP API
on port 8002. Sessions are created and messages are polled over HTTP — there is no interactive terminal
attached. This means there is no human-in-the-loop fallback if a session stalls.

In server mode, when a subagent (Momus, ses_149bd00e2ffeM9L0Kw1WNRsM2W) exits cleanly at step 3,
its completion is never signaled back to the parent session (Prometheus, ses_149d81796ffep7dPMjnA1heglo).
The parent session's message count freezes at 32, the opencode process holds the session open indefinitely,
and any polling client will loop forever with no error, no timeout, and no way to detect the hang
without external log inspection.

Suspected trigger: the subagent may have silently failed to read the plan file (.omo/plans/deal-document-identifier.md)
due to an ephemeral filesystem issue (files written in one turn disappearing before the next), causing it to exit
without producing output — but crucially without surfacing an error to the parent's task() call either.

- Parent session: ses_149d81796ffep7dPMjnA1heglo (Prometheus - Plan Builder)
- Subagent session: ses_149bd00e2ffeM9L0Kw1WNRsM2W (Momus - Plan Critic)
- opencode version: 1.17.3 (bun process, Linux)
- Middleware: oncall-middleware (Bun), active (running) since 10:00:51, PID 89270

### Operating System

Linux

### OpenCode Version

1.17.3

AceRothstein71:
## What changed

`packages/senpi-task/src/runners/in-process/child-handle.ts`: the tracked child turn now settles from the session's terminating `agent_end` event, not only from `session.prompt()` resolution. A per-turn settlement cell races the two paths; first writer wins, later writers are no-ops.

- `willRetry === true` agent_end events do not settle (runtime fallback keeps the turn running).
- abort followed by loop exit settles `cancelled` (mirrors the rpc handle).
- A stray agent_end after prompt-resolution settlement is ignored; revive/followUp semantics unchanged.
- 5 new co-located tests in `child-handle-outcome.test.ts`, including the exact #5167 regression (loop exits, prompt() never resolves).

## Why

Fixes #5167. Momus is pinned to in-process execution. When its senpi agent loop exits (the log's "exiting loop") while its `prompt()` promise never resolves, the in-process handle's `waitForIdle()` never settled: no outcome reached the manager, the record never terminalized, `manager.waitFor()` never resolved, and the parent's blocking `task()` call hung indefinitely with no timeout and no error - exactly the reported freeze (child done at 10:38:49, parent silent from 10:40:02, messageCount stuck at 32 forever).

The rpc handle already settles turns from terminating `agent_end` events (`runners/rpc/handle.ts`); this change gives the in-process runner the same contract so both runners share one settlement semantics.

## Observed behavior

Failing-first evidence under `.omo/evidence/20260825-momus-exit-not-propagated/`:

- RED (new tests vs old code): all three hang scenarios time out at 20s each - `agent_end` + never-resolving prompt leaves `waitForIdle()` pending forever.
- GREEN after fix: 9/9 tests pass in <1s, including completed-with-text, error-on-no-output, cancelled-on-abort, willRetry non-settlement, and stray-event idempotence.
- Scoped regression: `bun test packages/senpi-task/src/runners packages/senpi-task/src/tools/task packages/senpi-task/src/manager` -> 601 pass / 1 skip / 0 fail.
- Typecheck: `tsgo --noEmit -p packages/senpi-task/tsconfig.json` -> clean.

## Residual risk

A senpi build whose in-process session neither emits `agent_end` nor resolves `prompt()` would still hang; that requires an upstream senpi-core fix and is out of scope for this repo-side minimal change.

Fixes #5167

---
## Summary by cubic
Settles in-process child turns on `agent_end` so parents no longer hang when `prompt()` never resolves, fixing #5167. Previously only `prompt()` resolution settled the turn; now `agent_end` and `prompt()` race, first to arrive wins, matching the RPC runner's behavior. Each turn owns its generation, observation, and session subscription, so a superseded turn's late events can't settle or observe for the live turn.

- Only an explicit `willRetry === false` ends a turn; absent or malformed flags aren't terminal, matching the RPC runner.
- Abort followed by loop exit settles as `cancelled`.
- Stray `agent_end` after a prompt-settled turn is ignored; `followUp` behavior unchanged.
- Refreshes the `omo-senpi` task extension bundle to ship the fix.
- Adds regression tests in `child-handle-outcome.test.ts`; evidence under `.omo/evidence/20260825-momus-exit-not-propagated/`.
- Redacts the machine-local worktree path from the evidence doc so the markdown link audit passes.

<sup>Written for commit 617310e985e5820c7587eb6f32ab7374a6530878. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7310?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/senpi-task/src/runners/in-process/child-handle.ts
```diff
@@ -190,15 +207,15 @@ function createTrackedChildHandle(taskId: string, session: ChildSession): Tracke
...
-    )
+    const cell = Promise.withResolvers<RunnerOutcome>()
+    running = cell.promise
+    const settle = (outcome: RunnerOutcome): void => {
+      settleActiveTurn = undefined
+      turnActive = false
+      cell.resolve(outcome)
+    }
+    settleActiveTurn = settle
+    void runTurn(session, text, () => aborted, observation).then(settle)
```
code-yeongyu, line 230: [sisyphus-bot] [P1] This `.then(settle)` is not associated with the turn that created it. Repro: turn A emits terminal `agent_end` (so `turnActive=false`) while A's `prompt()` remains pending; `followUp()` starts turn B and overwrites `settleActiveTurn`; when A's prompt later resolves, this old callback clears B's `settleActiveTurn` and `turnActive` and resolves A's cell, leaving B's `waitForIdle()` permanently pending or allowing later events to be ignored. Capture a per-turn generation/cell and make the completion callback no-op unless it is still the active generation; add this exact delayed-prompt-after-revive regression test.

packages/senpi-task/src/runners/in-process/child-handle.ts
```diff
@@ -179,7 +183,20 @@ function createTrackedChildHandle(taskId: string, session: ChildSession): Tracke
...
+  // Settlement cell for the ACTIVE turn. runTurn's resolution races the session's terminating
+  // agent_end event; first writer wins, later writers are no-ops. Without the event path a child
+  // whose loop exits while its prompt() promise never settles (#5167) strands waitForIdle forever,
+  // leaving the parent's task() blocked with no timeout and no error.
+  let settleActiveTurn: ((outcome: RunnerOutcome) => void) | undefined
+
+  const unsubscribeObserver = session.subscribe((event) => {
+    observeTurnEvent(observation, event)
+    if (event.type !== "agent_end") return
+    if (eventFlagTrue(event, "willRetry")) return
```
code-yeongyu: [sisyphus-bot] [P1] The RPC handle on current dev settles only when `event.willRetry === false`; this check settles when the field is omitted, null, or malformed. That is a contract regression: an incomplete `agent_end` can terminate a live turn before runtime fallback/retry has decided it is final. Match the established strict-false check (and add an omitted/malformed payload test) rather than treating every value except literal true as terminal.

packages/senpi-task/src/runners/in-process/child-handle-outcome.test.ts
```diff
@@ -118,4 +122,84 @@ describe("createChildHandle turn outcomes", () => {
...
+  test("#given a turn already settled by prompt resolution #when a stray agent_end arrives afterwards #then the first outcome wins and a follow-up still revives", async () => {
+    // given
+    const fake = createEmittingSession()
+    const handle = createChildHandle({ taskId: "task-1", session: fake.session, promptText: "first run" })
+    fake.emit(assistantEnd("first verdict", "stop"))
+    fake.resolvePrompt()
+    expect(await handle.waitForIdle()).toEqual({ status: "completed", finalResponse: "first verdict" })
+
+    // when a stray agent_end arrives after settlement
+    fake.emit(agentEnd())
```
code-yeongyu: [sisyphus-bot] [P1] This is not the dangerous stray-event ordering. Here turn A already settled through `prompt()`, so its prompt cannot later execute an old completion callback. The regression to pin is: settle A by `agent_end` while A prompt remains pending, start B via `followUp`, then resolve A prompt and verify B remains active and can settle. The current implementation fails that sequence.

packages/senpi-task/src/runners/in-process/child-handle.ts
```diff
@@ -190,15 +207,15 @@ function createTrackedChildHandle(taskId: string, session: ChildSession): Tracke
...
-      () => {
-        turnActive = false
-      },
-      () => {
-        turnActive = false
-      },
-    )
+    const cell = Promise.withResolvers<RunnerOutcome>()
+    running = cell.promise
+    const settle = (outcome: RunnerOutcome): void => {
```
code-yeongyu, line 212: [sisyphus-bot] [P1] Clearing the shared `settleActiveTurn` and `turnActive` here is only safe if no earlier prompt callback can run afterward. In the reported scenario that assumption is explicitly false: `agent_end` is used precisely because `prompt()` may never settle, and it may also settle later. An old callback must not be able to clear the callback for a newer turn; guard this cleanup by identity (`if (settleActiveTurn !== settle) return`) and use a turn-local aborted flag as well.

packages/senpi-task/src/runners/in-process/child-handle-outcome.test.ts
```diff
@@ -118,4 +122,84 @@ describe("createChildHandle turn outcomes", () => {
...
+
+  test("#given an agent_end that will be retried by the runtime fallback #when it arrives mid-turn #then the turn does NOT settle early", async () => {
+    // given
+    const fake = createEmittingSession()
+    const handle = createChildHandle({ taskId: "task-1", session: fake.session, promptText: "retryable work" })
+
+    // when a willRetry agent_end arrives (the loop keeps running this turn)
+    fake.emit(agentEnd({ willRetry: true }))
+    const settledEarly = await Promise.race([
+      handle.waitForIdle().then(() => true),
```
code-yeongyu: [sisyphus-bot] [P2] This test is timing-dependent: a fixed 25ms `setTimeout` can pass even if the implementation settles slowly, and can flake under scheduler/CI load. Subscribe/await an explicit fake-session signal or expose a deterministic settlement probe, then assert the promise remains pending without sleeping; test discipline here explicitly forbids polling/wait-for-time patterns.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST CHANGES. The reported hang is real, but this implementation introduces an unbounded cross-turn race because the old prompt promise and the new turn share mutable settlement state. The PR is also CONFLICTING with current dev and must be rebased before this can be merged; the RPC implementation on dev already shows the stricter terminal-event contract this code should follow. The focused test command could not run in this isolated checkout because the repository's prebuilt dependency graph is missing @oh-my-opencode/rules-engine; the checked-in evidence is not a substitute for a fresh post-rebase run.

- [P1] Tie every prompt completion and event settlement to a turn generation/cell; the current global callback can settle or deactivate a later revive turn.
- [P1] Do not treat an agent_end with omitted willRetry as terminal without explicitly matching the runtime contract; this diverges from RPC's `willRetry === false` guard and makes malformed/partial events destructive.
- [P1] Late events from the prior turn can mutate the next turn's observation; event attribution is required, not just first-writer-wins on a shared variable.
- [P2] The new willRetry test uses a fixed 25ms sleep/race, so it is timing-dependent and does not reliably prove non-settlement.
- [P1] Rebase onto current dev, resolve the conflict, and rerun the native and plugin bundle checks before resubmitting.

code-yeongyu:
[sisyphus-bot] still present on dev @ 7d874efc4 (packages/senpi-task/src/runners/in-process/child-handle.ts:139 `await session.prompt(text)` is the only settlement path; waitForIdle at :218 returns that promise; observeTurnEvent :73-74 handles only message_end and never agent_end; origin/dev `git log -S agent_end -- child-handle.ts` is empty and PR commit 295786b31 is not an ancestor of origin/dev); the PR direction remains relevant but must be rebased onto current dev before evaluation

AceRothstein71:
@code-yeongyu — rebased onto current `dev`, which clears the CONFLICTING blocker.

- the unbounded cross-turn race: the old prompt promise and the new turn still share mutable settlement state
- align with the stricter terminal-event contract the RPC implementation on `dev` already demonstrates
- re your note that the focused test couldn't run for lack of `@oh-my-opencode/rules-engine`: the rebased tree builds cleanly here after `bun install`, so that should now be reproducible

AceRothstein71:
@code-yeongyu — all five points addressed in `617310e9`.

**[P1] Tie every prompt completion and event settlement to a turn generation.** Done. Each turn now owns a `generation`, its own `TurnObservation`, and its own subscription. `settle()` short-circuits once `turn !== generation`, so a superseded turn can no longer resolve the live cell or clear `turnActive`.

**[P1] Late events from the prior turn mutating the next turn's observation.** The subscription is now per-turn and is dropped when the turn settles, so a late event reaches an unsubscribed listener and is discarded. That is real attribution rather than first-writer-wins on a shared variable — the observation object itself is per-turn, so there is nothing shared left to race on.

**[P1] `agent_end` with omitted `willRetry` treated as terminal.** Fixed to match `runners/rpc/handle.ts:69` exactly: only an explicit `willRetry === false` ends a turn. The old `eventFlagTrue(event, "willRetry")` inverted the burden of proof and made malformed or partial events destructive.

Worth flagging: three existing tests emitted a bare `agentEnd()` with no flag and depended on it settling. Under the RPC contract that shape is not terminal, so I updated them to emit `willRetry: false` — the shape the runtime actually produces, and the one every RPC test uses.

**[P2] The 25ms sleep/race.** Replaced with a microtask drain. Settlement would resolve `waitForIdle` within a microtask, so draining the queue proves non-settlement without depending on machine speed.

**[P1] Rebase + rerun.** The branch is on current `dev`. Diff is 5 files: this PR's scope plus one regenerated bundle (`omo-task.js`), staged with `git add -u` so no ignored build output is force-added.

Two new tests, both of which fail against the previous implementation:

- an `agent_end` with an absent `willRetry` does not settle the turn
- a revived turn survives the PREVIOUS turn's prompt resolving late. It settles the live turn through the **event** path afterwards, which is what proves the stale settler did not disarm the live cell or drop its subscription. Against the old code this times out at 20s.

On your note that the focused command could not run for a missing `@oh-my-opencode/rules-engine`: it runs on the rebased tree after `bun install`. `bun test packages/senpi-task` gives 1797 pass / 1 fail, and that single failure is a pre-existing order-dependent flake in `manager-claim.test.ts` — it passes in isolation and fails identically on unmodified `dev`. `child-handle-outcome.test.ts` is 11/11. `tsgo --noEmit` clean.

PR #4294, opened 2026-05-22
fix(agent-config): refresh agent registry on config apply

SpencerJung:
## Summary
- Reset the registered agent-name lookup before each config-agent registration pass.
- Re-register the freshly rebuilt agent config so `/connect` config re-entry cannot leave stale selector aliases behind.
- Add regression coverage for clearing aliases and for config re-application keeping OMO agents registered.

Fixes #4130

## Verification
- `lsp_diagnostics` on `src/features/claude-code-session-state/state.ts`
- `lsp_diagnostics` on `src/plugin-handlers/agent-config-handler.ts`
- `lsp_diagnostics` on `src/features/claude-code-session-state/state.test.ts`
- `lsp_diagnostics` on `src/plugin-handlers/agent-config-handler.test.ts`
- `bun test src/features/claude-code-session-state/state.test.ts src/plugin-handlers/agent-config-handler.test.ts`
- `bun run typecheck`
- `bun run build`
- `git diff --check`

---
## Summary by cubic
Refreshes the agent registry on each config apply so `/connect` can’t leave stale agent aliases behind. Keeps the registry aligned with the freshly rebuilt config and adds regression tests.

- **Bug Fixes**
  - Clear registered agent names and aliases before re-registering from the current config.
  - Ensure current/builtin agents are re-registered; stale names are removed.
  - Add tests that verify alias clearing without losing session ownership and that re-applying config replaces stale registrations.

<sup>Written for commit 66fefc3b90f0e2bc0d98be52e2134cec08b1a6ad. Summary will update on new commits. <a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/4294?utm_source=github">Review in cubic</a></sup>

PR #7321, opened 2026-08-25
fix(omo-opencode): self-heal agent registry frozen by instance races

### LINKED ISSUE #4856: /connect causes agents to disappear - PR #4294 fix is ineffective
### Prerequisites

### Bug Description

On oh-my-openagent v4.7.5 + OpenCode v1.16.0, all OMO agents still disappear after /connect. PR #4294 (closed #4130) did not fix it - the root cause is OpenCode never re-invokes the plugin config hook after model switch. See analysis: https://github.com/code-yeongyu/oh-my-openagent/issues/4130#issuecomment-4631719774

### Steps to Reproduce

1. Install oh-my-openagent v4.7.5 + OpenCode v1.16.0
2. Start opencode - agents appear correctly
3. Use /connect to switch model
4. Agent selector - all OMO agents gone, only build/plan remain
5. Restart opencode - agents reappear until /connect is used again

### Expected Behavior

Plugin config hook should be re-invoked after /connect so agents are re-registered. Or OMO should provide a workaround via command.execute.before or event hook.

### Actual Behavior

Agents disappear permanently after /connect. PR #4294 added clearRegisteredAgentNames() but the hook is never called by OpenCode after model switch. Upstream: anomalyco/opencode#30955

### Doctor Output

```shell
OpenCode: v1.16.0
oh-my-openagent: v4.7.5
OS: Windows Server 2022 (x64)
Plugin: OPENCODE_PLUGINS=oh-my-openagent
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Windows

### OpenCode Version

1.16.0

AceRothstein71:
## Summary

- Detect when OpenCode's live agent registry has frozen without OMO agents and recover by disposing the instance so OpenCode rebuilds it with plugin agents applied.
- New `features/agent-registry-recovery` module: records the agent names applied by each config-hook run, then on `session.idle` verifies via `client.app.agents()` that at least one applied name is still visible; if none are, it calls `client.instance.dispose()` once per 30s cooldown.
- `config-handler.ts` records the snapshot in both apply paths (full build + cache-hit replay); `event.ts` runs the fail-open recovery check after turns complete.

## Why

OpenCode materializes its per-instance Agent registry lazily from the shared config object and caches it for the instance lifetime (upstream `InstanceState`/`ScopedCache`, infinite TTL). Plugin agents reach that registry only through OMO's async `config` hook mutating the shared config object. When an agent query races hook application - instance recreation after `/connect`, or the startup race documented in anomalyco/opencode#30955 - the registry freezes without OMO agents and never recovers until a full restart. PR #4294 could not fix this because its clearing/re-registration only runs inside the config hook, which does re-fire per instance but cannot unfreeze an already-materialized registry.

A plugin cannot rebuild OpenCode's cache directly; instance disposal is the sanctioned rebuild path: bootstrap awaits plugin config hooks before the recreated instance serves traffic.

## Observed behavior

- Failing-first tests captured red before implementation (`.omo/evidence/20260825-connect-agents-disappear/red-run.log`), green after (`green-run.log`: 92 pass / 0 fail; wider scoped run 587 pass / 0 fail across 71 files).
- Live mechanism probe against real opencode 1.18.22 in an isolated XDG sandbox with a stub config-hook plugin: `GET /agent` listed the plugin agent; `POST /instance/dispose` returned `true`, the server stayed alive (HTTP 200), the rebuilt instance re-ran the config hook, and the plugin agent was present again (`live-probe-output.txt`, `hook.log`). The probe also reproduced the freeze signature: an early `/agent` query saw no plugin agents even though the hook had fired.
- `bunx tsgo --noEmit` clean (`typecheck.log`).

## QA / evidence

Recorded under `.omo/evidence/20260825-connect-agents-disappear/` (NOTES.md covers what was tested, observed, why it is enough, and what was omitted). A full interactive TUI `/connect` drive with real provider credentials was not possible in this environment; the recovery seam itself (config-hook application plus dispose/rebuild restoring plugin agents) was proven live with a stub plugin instead.

## Residual risk

Recovery is fail-open: if the dispose endpoint is unavailable or errors, the failure is logged and the session continues unaffected. The 30s cooldown prevents dispose loops; a healthy registry short-circuits to a single local `agents` query per idle event.

Fixes #4856

---
## Summary by cubic
Self-heals `omo-opencode` when OpenCode's per-instance agent registry freezes and drops OMO agents due to instance races. Previously, the registry could cache without plugin agents after /connect or startup and stayed wrong until restart; now we detect this on session.idle and dispose the instance (max once per 30s) so OpenCode rebuilds it with plugin agents. Fixes #4856.

**Bug Fixes**
- Records the agent names applied by the config hook in both full build and cache-hit replay paths.
- Verifies the live registry via `client.app.agents()` and calls `client.instance.dispose()` when none of the applied names are visible.
- Runs the check after each turn; healthy sessions pay a single agents query per idle event, and all errors fail open with logging.
- Excludes `.omo` and `.opencode` evidence files from the markdown link audit.

<sup>Written for commit 720fd25e1e5870df14d716910097414c388072e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7321?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/features/agent-registry-recovery/stale-registry-recovery.ts
```diff
@@ -0,0 +1,84 @@
...
+      try {
+        liveNames = extractLiveAgentNames(await client.app.agents())
+      } catch (err) {
+        log("[agent-registry-recovery] failed to query live agents", {
+          error: err instanceof Error ? err.message : String(err),
+        })
+        return false
+      }
+
+      const hasAppliedAgent = liveNames.some((name) => isAgentRegistered(name))
```
code-yeongyu, line 65: [sisyphus-bot] The detector snapshots `appliedNames` above, but the health decision ignores that snapshot and checks the unrelated process-global `registeredAgentNames` via `isAgentRegistered`. This can produce a false healthy result when a live registry contains an old/stale name that happens to remain registered globally, and it can miss the very instance/config snapshot this function was created to verify. Compare normalized live names against the captured `appliedNames` (or store the authoritative registration snapshot) instead of consulting mutable global state.

packages/omo-opencode/src/features/agent-registry-recovery/registry-snapshot.ts
```diff
@@ -0,0 +1,36 @@
...
+ * agents before OMO's async config hook finishes applying (instance recreation
+ * after /connect, startup race), the registry freezes without OMO agents. The
+ * snapshot records what the last successful application looked like so the
+ * recovery checker can tell a healthy registry from a frozen one.
+ */
+
+let appliedAgentNames: string[] = []
+let lastRecoveryAttemptAt = 0
+
+export function recordAppliedRegistry(names: string[]): void {
```
code-yeongyu, line 16: [sisyphus-bot] This state is module-global, but the snapshot describes whichever config hook happened to finish last in the entire process. With multiple OpenCode clients/instances (or workspaces) sharing the plugin module, one instance can overwrite another's names; an idle event from client A can then query/dispose client A using client B's snapshot. Scope the snapshot and cooldown to the client/instance (or pass the state through the handler) and add a multi-instance test.

packages/omo-opencode/src/features/agent-registry-recovery/stale-registry-recovery.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (hasAppliedAgent) return false
+
+      markRecoveryAttempted(now())
```
code-yeongyu, line 68: [sisyphus-bot] `markRecoveryAttempted` occurs only after the awaited `app.agents()` call. Two session.idle events can enter this function concurrently, both observe the old timestamp, both observe the stale registry, and both call `instance.dispose()` before either marks the cooldown. The claimed 'once per 30s' guarantee is therefore not true under the same event concurrency that motivates this fix. Add an in-flight guard/atomic claim before the await (with deliberate failure handling) and test overlapping calls.

packages/omo-opencode/src/plugin/event.ts
```diff
@@ -177,6 +179,11 @@ export function createEventHandler(args: {
         type: "session.idle",
         sessionId: resolveSessionEventID(props) ?? "",
       }));
+      void staleRegistryRecovery.maybeRecover().catch((err) => {
```
code-yeongyu, line 182: [sisyphus-bot] This is deliberately fire-and-forget, so the event handler returns before recovery completes and provides no ordering with the next request/idle event. It also allows overlapping calls to build up, and a short-lived plugin process can exit before the dispose request is sent. If recovery is required before subsequent work can rely on agents, await it (with its own fail-open catch); otherwise explicitly serialize/dedupe the background task and test that contract.

packages/omo-opencode/src/plugin-handlers/config-handler.ts
```diff
@@ -140,6 +141,8 @@ export function createConfigHandler(deps: ConfigHandlerDeps) {
           };
     }
 
+    recordAppliedRegistry(Object.keys(agentResult));
```
code-yeongyu, line 144: [sisyphus-bot] Recording only `Object.keys(agentResult)` does not establish that the live registry belongs to this config-hook run or instance; this is written into the same process-global slot used by every handler. A later cache-hit/full-build replay from another instance can replace it, and a failed/partial subsequent config application leaves the previous snapshot in place. The recovery state needs an instance identity/lifecycle boundary and should be invalidated or versioned when application fails, not treated as a global last-writer-wins fact.

packages/omo-opencode/src/features/agent-registry-recovery/stale-registry-recovery.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    recordAppliedRegistry(["sisyphus"])
+    const { client, calls } = createFakeClient({ liveNames: ["build", "plan", "sisyphus"] })
+    const recovery = createStaleAgentRegistryRecovery({ client })
+
+    const recovered = await recovery.maybeRecover()
+
+    expect(recovered).toBe(false)
+    expect(calls.dispose).toBe(0)
+  })
+
```
code-yeongyu, line 68: [sisyphus-bot] This test only proves the fake implementation's sequential happy path. It does not test the race in the production trigger: start two `maybeRecover()` calls while `app.agents()` is blocked, release them together, and assert one dispose. Without that test the advertised cooldown guarantee can regress while the suite remains green.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
...
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
+    .map((filePath) => resolve(WORKSPACE_ROOT, filePath))
```
code-yeongyu, line 22: [sisyphus-bot] nit: excluding every tracked `.omo/` and `.opencode/` markdown file from this audit is a broad coverage hole introduced solely because this PR adds evidence. It means any real broken local link in those directories is now silently ignored. Keep generated evidence out of the tracked source/audit input (or narrow the exclusion to the specific generated files) rather than weakening the repository-wide check.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST CHANGES. This patch adds a recovery mechanism, but its correctness state is not tied to the instance/configuration it is supposed to repair, and its asynchronous trigger permits duplicate recoveries. The result can be both false negatives (the affected registry remains frozen) and false positives (an unrelated healthy instance is disposed), while the tests do not exercise either lifecycle.

- The most serious issue is the mismatch between the recorded snapshot and the separate global registration set used for the actual health decision.
- The snapshot and cooldown are process-global even though the client/agent registry is instance- and potentially workspace-specific.
- Multiple idle events can overlap the query-before-mark window and dispose repeatedly; the fire-and-forget call makes this harder to observe and test.
- The added tests validate isolated fakes, not createEventHandler integration, instance boundaries, or concurrent idle delivery.
- nit: committing generated probe/evidence logs and excluding their directories from the markdown audit weakens repository hygiene and masks link-audit coverage rather than testing the intended source tree.

Issue #5586, opened 2026-06-26
[Bug]: title generation happens after ulw instructions are injected


davhdavh:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

It seems that the session title generation happens after the ulw block of instructions are added, since my sessions often just get named "ULTRAWORK MODE ENABLED!"

### Steps to Reproduce

1. ulw Some task

### Expected Behavior

Title of session becomes whatever the task actually is about.

### Actual Behavior

Title of session becomes "ULTRAWORK MODE ENABLED!"

### Doctor Output

```shell
> bunx oh-my-openagent doctor
oh-my-opencode: bun is not available (spawnSync bun ENOENT); falling back to the node CLI at C:\Users\xxx\AppData\Local\Temp\bunx-659225885-oh-my-openagent@latest\node_modules\oh-my-openagent\dist\cli-node\index.js

 oMoMoMoMo Doctor

 ✓ System OK (opencode 1.17.7 · oh-my-openagent 4.13.0)
```

### Error Logs

```shell
NA
```

### Configuration

```json
{
  "$schema": "https://opencode.ai/config.json",
  "plugin": [
    "oh-my-openagent@latest"
  ],
  "autoupdate": true,
  "agent": {
...
    "title": {
      "model": "opencode-go/deepseek-v4-flash"
    },
```

### Operating System

Windows

### OpenCode Version

1.17.7

code-yeongyu:
[sisyphus-bot] Thanks for reporting this! Confirmed: the ULW system-transform handler injects instructions that mandate the model say 'ULTRAWORK MODE ENABLED!' as its first response (dist/index.js:96175). OpenCode's title generation then uses that first assistant message, causing the session title to become 'ULTRAWORK MODE ENABLED!' instead of the actual task. The fix likely needs to either strip the ULW banner from title generation or delay title generation until after the banner is consumed.

[...]

PR #7323, opened 2026-08-25
fix(omo-opencode): keep OpenCode session auto-titling working for CLI runs and internal dispatches

### LINKED ISSUE #5544: [sisyphus-bot] [Bug]: Session title changed to a timestamp
> **⚠️ THIS REPOSITORY IS ENGLISH-ONLY. ALL ISSUE, PR, REVIEW, AND COMMENT COMMUNICATION MUST BE WRITTEN IN ENGLISH. DO NOT POST NON-ENGLISH CONTENT HERE. THIS IS A STRICT PROJECT COMMUNICATION RULE AND MUST BE FOLLOWED FROM NOW ON.**

### Prerequisites

### Bug Description

Before installation, the OpenCode session list used task names. After installing ULW, the title changed to `New session-2026-06-24T03:24:10.559Z`, making it difficult to identify the task associated with each session.

### Steps to Reproduce

None

### Expected Behavior

Name the session based on a brief description of the task.

### Actual Behavior

None

### Doctor Output

```shell
$ bunx oh-my-openagent doctor

error: InvalidUtf8

----- bun meta -----
Bun v1.0.26-canary.4 (1f0c1c81) Windows x64 
BunxCommand: 
Elapsed: 46ms | User: 0ms | Sys: 15ms
RSS: 32.25MB | Peak: 32.25MB | Commit: 77.12MB | Faults: 8024
----- bun meta -----

Search GitHub issues https://bun.sh/issues or join in #windows channel in https://bun.sh/discord
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

_No response_

### Operating System

Windows

### OpenCode Version

1.17.9

AceRothstein71:
## Problem

OpenCode names a session after the task via `Session.ensureTitle` (`packages/opencode/src/session/prompt.ts`). That flow only runs when:

1. the session is not a child session,
2. the title is still the default placeholder (`New session - <ISO ts>`),
3. history contains EXACTLY ONE real user message - real meaning its parts are not all `synthetic: true`,
4. the built-in `title` agent's small-model call succeeds.

oh-my-openagent defeats gates 2 and 3 in two ways, which is why sessions end up stuck on the timestamp placeholder (#5544):

- **CLI runs are born pre-titled.** `cli/run/session-resolver.ts` created every new session with `title: "oh-my-openagent run"`. Gate 2 then fails forever, so no `omo run` session (including all sisyphus-bot automation) ever gets a task-based title.
- **Internal control messages count as real user messages.** OpenCode core does not know OMO's `<!-- OMO_INTERNAL_INITIATOR -->` marker; only `synthetic: true` parts are excluded from gate 3. Three main-session dispatch routes sent marker-only or entirely unmarked text parts:
  - background-agent parent wakes (`parent-wake-prompt-dispatch.ts`)
  - monitor output injection (`output-injector.ts`)
  - team live delivery (`member-session-routing.ts` `buildMemberPromptBody`, previously plain unmarked text)

  Once any such message exists when titling runs, gate 3 fails and titling is permanently disabled for that session (later turns never retry because the count stays >= 2).

This is the same title-generation area as #5586/#5659 (injected mode text polluting what the title model sees); #5544 is the remaining failure mode where titling never lands at all.

## Fix

- Create `omo run` sessions untitled so OpenCode seeds its default placeholder and `ensureTitle` names the session from the task.
- Build parent-wake, monitor-injection, and member live-delivery parts with `createInternalAgentContinuationTextPart` (`synthetic: true`), matching the convention runtime-fallback retries, todo continuation, and compaction recovery already use.

Delivery is unchanged: OpenCode's `MessageV2.toModelMessagesEffect` includes non-empty non-ignored text parts regardless of the synthetic flag, and every OMO-side detector keys on the marker TEXT, which is byte-identical (`createInternalAgentContinuationTextPart` spreads `createInternalAgentTextPart`). Child-session dispatches (delegate-task sync, background task launch/resume) were deliberately left as-is: their sessions carry `parentID`, so `ensureTitle` skips them anyway.

## QA / Evidence

Failing-first, evidence under `.omo/evidence/20260825-title-timestamp-regression/`:

- RED before fix (`tests-red.txt`): 10 pass / 8 fail - all 8 are the new assertions (create body still carried `title`; dispatched wake/monitor/member parts had `synthetic === undefined`).
- GREEN after fix (`tests-green-scoped.txt`): 18 pass / 0 fail across the four touched test files.
- Scoped regression suites: cli-run 197 pass, background-agent 745 pass, monitor 96 pass, team-mode 288 pass / 1 skip, 0 failures.
- Typecheck: `tsgo --noEmit` exit 0 at repo root and for `packages/omo-opencode`.
- Omitted: no live `opencode run` harness drive in this network-restricted environment (no provider credentials), so the title LLM call itself was not exercised end to end; the unit pins target the exact gates OpenCode applies around that call.

Fixes #5544

---
## Summary by cubic
Restores OpenCode session auto-titling for `omo run` and internal dispatches. Previously CLI sessions were pre-titled and internal control messages counted as real user messages, leaving titles stuck on the timestamp; now sessions start untitled and control parts are synthetic so `ensureTitle` names from the task. Fixes #5544.

- Bug Fixes
  - Create CLI sessions without a custom title so the default placeholder triggers `ensureTitle`.
  - Mark parent-wake, monitor injection, and team live-delivery text parts as synthetic using `createInternalAgentContinuationTextPart`.
  - Keep internal markers and delivery behavior unchanged; detectors still key on the marker text.
  - Leave child-session dispatches unchanged since `ensureTitle` already skips them.
  - Exclude `.omo` and `.opencode` markdown from the link audit so the new evidence files don't fail it.
  - Add tests covering create-body shape and synthetic-part contract; scoped suites and typecheck pass.

<sup>Written for commit e01428ed3c47523e71180ad689302a4c7b16d922. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7323?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/features/background-agent/parent-wake-prompt-dispatch.ts
```diff
@@ -52,8 +52,8 @@ export async function sendParentWakePrompt(input: ParentWakePromptDispatchInput)
           ...input.latestWake.promptContext,
           parts: [
             input.forceNoReply === true || !input.latestWake.shouldReply
-              ? withInternalNoReplyMarker(createInternalAgentTextPart(notificationContent))
-              : createInternalAgentTextPart(notificationContent),
+              ? withInternalNoReplyMarker(createInternalAgentContinuationTextPart(notificationContent))
```
code-yeongyu, line 55: [sisyphus-bot] Blocking: createInternalAgentContinuationTextPart does more than mark this wake synthetic: it adds metadata.compaction_continue=true. The shared semantic-dedupe normalizer treats any such marked part as the single {__omo_internal_intent: continuation} key, so a distinct parent wake arriving after the broad reservation hold can be returned as queued/coalesced instead of dispatched. This route needs a synthetic internal part without continuation intent, or a route-specific dedupe key that preserves wake identity.

packages/omo-opencode/src/features/monitor/output-injector.ts
```diff
@@ -164,8 +164,8 @@ export class MonitorOutputInjector {
             noReply: !shouldReply,
             parts: [
               shouldReply
-                ? createInternalAgentTextPart(content)
-                : withInternalNoReplyMarker(createInternalAgentTextPart(content)),
+                ? createInternalAgentContinuationTextPart(content)
```
code-yeongyu, line 167: [sisyphus-bot] Blocking: monitor output is now labeled as a compaction continuation, so all terminal batches and all non-terminal batches with this shape share the semantic continuation key regardless of batchSeq/content/source. When a second distinct monitor batch arrives within the semantic dedupe window, prompt-async-gate can coalesce it with the first and silently lose output. Marking a message synthetic must not opt it into compaction-continuation dedupe.

packages/omo-opencode/src/features/team-mode/member-session-routing.ts
```diff
@@ -64,6 +65,6 @@ export function buildMemberPromptBody(member: RuntimeStateMember, text: string):
     ...(model ? { model } : {}),
     ...(member.model?.variant ? { variant: member.model.variant } : {}),
     ...buildPromptGenerationParams(member.model),
-    parts: [{ type: "text", text }],
+    parts: [createInternalAgentContinuationTextPart(text)],
```
code-yeongyu, line 68: [sisyphus-bot] Blocking: live team envelopes now carry compaction_continue metadata even though they are ordinary mailbox messages, not compaction recovery. prompt-async-gate intentionally ignores text, agent, model, and route fields for continuation prompts, so two different envelopes to the same member can collapse into one recent semantic dispatch. That is a delivery-loss regression introduced solely to fix title counting; use a synthetic-only marker or explicitly preserve envelope identity.

packages/omo-opencode/src/features/background-agent/parent-wake-prompt-dispatch.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      // when
+      await notifier.flushPendingParentWake(sessionID)
+
+      // then
+      expect(promptAsyncCalls).toHaveLength(1)
+      const parts = promptAsyncCalls[0]?.body.parts ?? []
+      expect(parts.length).toBeGreaterThan(0)
+      for (const part of parts) {
+        expect(part.synthetic).toBe(true)
```
code-yeongyu, line 68: [sisyphus-bot] The new test is incomplete and therefore misses the blocking regression above: it checks synthetic=true and the marker, but never checks that the part does not carry compaction_continue metadata and never flushes two distinct wakes after the prompt gate hold to prove both are delivered. A green result here can coexist with silently coalesced parent notifications.

packages/omo-opencode/src/features/team-mode/member-session-routing.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    const member = createMember()
+
+    // when
+    const body = buildMemberPromptBody(member, "check the failing test")
+
+    // then
+    expect(body.parts).toHaveLength(1)
+    const part = body.parts[0]
+    expect(part.type).toBe("text")
+    expect(part.synthetic).toBe(true)
```
code-yeongyu, line 29: [sisyphus-bot] This assertion only proves the new helper was called, not that team delivery semantics remain intact. Because the helper also sets metadata.compaction_continue, add a test for two different live envelopes targeting the same idle member and assert two promptAsync calls (or assert a non-continuation synthetic part). Otherwise the test blesses the accidental semantic-dedupe contract.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
```
code-yeongyu, line 19: [sisyphus-bot] nit: this broad prefix filter skips every markdown file under .omo/ and .opencode/, not just the generated evidence artifacts added by this PR. Any real maintainer documentation placed in either tree will now evade the local-link audit. Exclude only the known generated evidence path(s), or document and test the narrower policy.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST_CHANGES: this fixes one title gate by introducing a broader cross-route prompt regression. The new code conflates "synthetic internal message" with "compaction continuation", and the tests only assert the former. Before merging, use a synthetic-only part shape (or otherwise preserve route/message identity) and add race/dedupe coverage. Findings:
- The continuation helper adds metadata.compaction_continue, which the shared semantic dedupe intentionally normalizes to one global continuation key; distinct internal messages can therefore be coalesced and silently dropped.
- This affects parent wakes, monitor batches, and team live delivery, not just title bookkeeping.
- The new tests are tautological around this regression: they assert synthetic=true/marker presence but never assert metadata or that distinct prompts still dispatch.
- nit: the markdown audit now excludes every .omo/ and .opencode/ markdown file instead of narrowly excluding the generated evidence files; that disables the audit for future maintainer documentation in those trees.
The existing green logs do not cover the newly introduced semantic-dedupe behavior, so they are not sufficient evidence for this change.

PR #7324, opened 2026-08-25
fix(senpi-task): resume DAG runs without blocking

code-yeongyu:
## Defect

Restart recovery reconciled surviving pending/running DAG children by awaiting `TaskManager.waitFor()` inline. A single long-running child therefore kept the claimed run checkpoint paused, froze journal sequencing, delayed reuse events for later nodes, and left amend/retry controls unavailable.

Observed in the 2026-08-25 `dag_530ad299` incident and reproduced by the supplied R1 evidence fixture.

## Fix design

- Recovery now performs only non-blocking reconciliation before appending `dag.run.resumed`.
- Surviving child identities are passed into the newly-created scheduler as externally-owned reattachments.
- The new scheduler registers normal settlement watchers for those children and folds each terminal outcome through its existing pending-terminal-result/reducer path.
- Reattached children are never sent through `startOwned`, so they cannot be double-admitted.
- The previous scheduler instance is never resumed; cancellation/admission latches remain single-shot.
- Strict wave barriers are preserved: reattached children still settle before a later wave advances.

The secondary `dag-runtime.ts` snapshot seam was not changed. Store mutation listeners already re-emit recovery mutations, and a cheap isolated adapter test seam was not available in this focused fix.

## RED -> GREEN proof

RED on unchanged code after porting R1:

```text
Expected to contain: "dag.node.reused"
Received: []
(fail) DAG crash recovery > R1: resume emits dag.run.resumed and reuses completed nodes while a reattached child is still running
14 pass, 1 fail
```

GREEN after the fix:

```text
(pass) DAG crash recovery > R1: resume emits dag.run.resumed and reuses completed nodes while a reattached child is still running
```

The final regression subscribes to exact WAL appends instead of using a timing sleep. It also asserts a ready node is admitted and the surviving running node is not redispatched.

## Verification

- `bun test packages/senpi-task/src/dag` — **244 pass, 0 fail**
- `bun run --cwd packages/senpi-task typecheck` — pass
- `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` — pass
- `node packages/omo-senpi/scripts/qa/drive.mjs --self-test` — pass
- Real isolated Senpi driver — **PASS**, `realSenpiUntouched: true`

Broader `bun run test:senpi` was attempted and reached an unrelated existing failure in `packages/omo-senpi/src/components/telemetry/product-identity.test.ts`: the fixture expected its temporary explicit agent directory but resolved `/Users/yeongyu/.omo/agent`. The touched DAG package suite and both scoped typechecks pass.

R2 from the evidence bundle was intentionally excluded because it conflicts with the pinned strict barrier test and is out of scope.

---
## Summary by cubic
Resumes DAG runs in `senpi-task` without blocking on surviving pending/running children. Previously recovery awaited `TaskManager.waitFor()` inline, pausing the run and freezing WAL sequencing; now recovery emits `dag.run.resumed` immediately and folds outcomes asynchronously so reuse and controls stay available.

- Perform non-blocking reconciliation before appending `dag.run.resumed`.
- Reattach surviving tasks to the new scheduler via `reattachedTasks` and register settlement watchers.
- Skip `startOwned` for reattached tasks to prevent double dispatch; do not revive the previous scheduler.
- Keep strict wave barriers; later waves still wait for reattached tasks to settle.
- Add a regression test that asserts reuse and resume while a reattached child continues running.

<sup>Written for commit b124f69c865052527d5f10eb020b5ddd58555b74. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7324?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/senpi-task/changes.md
+++ b/packages/senpi-task/changes.md
@@ -1,3 +1,9 @@
+## 2026-08-25 — Resume DAG runs without waiting for surviving children
+
+DAG recovery now marks a claimed run resumed after non-blocking reconciliation work and watches surviving pending/running child tasks through the new scheduler instance. Their terminal outcomes are folded asynchronously, so durable result reuse and operator controls remain live while a reattached child continues executing.
+
+Keep surviving task owners externally owned during scheduler re-entry; never dispatch them a second time or revive the prior scheduler instance.
+
 ## 2026-08-20 — Export the canonical notice-box visual contract
 
 The package now exports one `buildNoticeBox` helper and `NoticeSpec`-shaped types for Senpi-coupled adapters. It reproduces Senpi's canonical `Box(1, 1, customMessageBg)` contract while the pinned host package does not export its own builder.
--- a/packages/senpi-task/src/dag/recovery.test.ts
+++ b/packages/senpi-task/src/dag/recovery.test.ts
@@ -236,6 +236,46 @@ describe("DAG crash recovery", () => {
     expect(events(store).some((event) => event.type === "dag.run.resumed")).toBe(true)
   })
 
+  test("R1: resume emits dag.run.resumed and reuses completed nodes while a reattached child is still running", async () => {
+    const projectDir = tempProject()
+    const store = createDagFileStore({ project_dir: projectDir })
+    const manager = new RecoveryTaskManager()
+    const observedEvents = deferred<void>()
+    const seenTypes = new Set<string>()
+    const observedStore = {
+      ...store,
+      appendEvent(event: DagRunEvent) {
+        store.appendEvent(event)
+        seenTypes.add(event.type)
+        if (seenTypes.has("dag.node.reused") && seenTypes.has("dag.run.resumed")) observedEvents.resolve()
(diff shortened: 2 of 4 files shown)


MoerAI:
Closing as superseded by #7320, which merged immediately before this PR and is already incorporated into current `dev`.

This branch is now conflicting and has no required CI evidence. It also omits the shipped bundle update carried by #7320.

Subsequent DAG recovery and state-machine fixes landed in #7376, #7511, and #7539. Please open a fresh focused report only if current `dev` still reproduces a distinct recovery defect.

PR #7341, opened 2026-08-25
fix(omo-senpi): warm the task chunk's pi-tui runtime

### LINKED ISSUE #7340: fix(omo-senpi): task bundle never warms its private pi-tui lazy runtime
## Bug description

The packaged native `omo-ai@5.0.0-0.beta.20` task chunk has a separate bundled copy of `senpi-task`'s pi-tui lazy accessor. `omo.js` correctly awaits `loadPiTui()`, but that initializes the main bundle's private module instance rather than `extensions/omo-task.js`'s copy. The task chunk's accessor remains permanently cold and throws when a task renderer calls `visibleWidth`, `truncateToWidth`, `Box`, or `Text`.

## Reproduction

1. Install `omo-ai@beta` (currently `5.0.0-0.beta.20`) on Windows.
2. Run an Omo session that emits a task completion/status renderer.
3. Omo exits with:

```text
Error: The @earendil-works/pi-tui barrel was accessed before it was loaded. Await loadPiTui() at the registration entry point before reading pi-tui values synchronously.
    at Ld (.../plugin/extensions/omo-task.js:...)
```

## Root cause

`packages/omo-senpi/src/extension/compose.ts` warms only the main bundle. The build emits `omo-task.js` independently via `packages/omo-senpi/plugin/scripts/build-extension.mjs`, so its module-level `piTuiModule` state is distinct. Since `createTaskComponent().register()` does not call `loadPiTui()`, tree-shaking leaves that chunk with a throwing accessor and no loader.

## Proposed fix

Import `loadPiTui` alongside the current `@oh-my-opencode/senpi-task` imports in `components/task/index.ts` and await it at the start of `register()`. This warms the task chunk instance before it registers task renderers, is memoized, and leaves the existing compose-level warm-up intact for other components.

A focused PR is prepared from this diagnosis.

NetVar1337:
## Summary
- await `loadPiTui()` from the independently bundled task runtime before task renderers are registered
- preserve the existing main-bundle warm-up while fixing the task chunk's distinct module instance
- assert that generated `omo-task.js` retains its pi-tui dynamic import

## Root cause
`omo.js` and `omo-task.js` are separate bundles. The compose-level warm-up initializes only `omo.js`'s private lazy-module state, leaving the sidecar accessor permanently cold.

## Verification
- `bun test packages/omo-senpi/plugin/scripts/build-extension.test.mjs`
  - 12 passed, 0 failed

Fixes #7340

### DIFF
--- a/packages/omo-senpi/plugin/scripts/build-extension.test.mjs
+++ b/packages/omo-senpi/plugin/scripts/build-extension.test.mjs
@@ -222,6 +222,7 @@ describe("checkExtensionCurrent", () => {
     const manifest = JSON.parse(await readFile(join(pluginRoot, "package.json"), "utf8"))
 
     expect(main).toContain('import("#omo-task-runtime")')
+    expect(task).toContain('import("@earendil-works/pi-tui")')
     expect(task).toMatch(/^\/\/ omo:[A-Za-z0-9_-]{43}:[A-Za-z0-9_-]{43}/)
     expect(manifest.imports).toEqual({ "#omo-task-runtime": "./extensions/omo-task.js" })
   })
--- a/packages/omo-senpi/src/components/task/index.ts
+++ b/packages/omo-senpi/src/components/task/index.ts
@@ -1,5 +1,6 @@
 import { loadSenpiOmoConfig } from "../config-resolution"
 import {
+  loadPiTui,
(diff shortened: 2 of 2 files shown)


code-yeongyu:

PR #7345, opened 2026-08-26
fix(background-agent): defer fallback takeover behind opt-in fallbackDeferMs

### LINKED ISSUE #7319: Background-task fallback takes over before a retry can succeed: transient LLM server errors cause immediate model downgrade instead of recovery
### Prerequisites

### Problem Description

oh-my-openagent 5.0.0-beta.19 (behavior also present in 4.19.4).

LLM servers routinely return transient failures — server-side rate limiting, internal errors, temporary unavailability. For this class of errors the underlying condition usually clears on its own within seconds, so **retrying the same request on the same model often succeeds**. Switching models is only useful when the failure is persistent.

## Observed behavior

The background-agent takes over the moment such an error appears, without giving a retry a chance to succeed:

1. **`session.error` path** — `handleSessionErrorEvent` calls `tryFallbackRetry(task, errorInfo, "session.error")` with zero delay. The "session still alive → treat as transient" guard exists only on the *non-retryable* branch — the branch that, by definition, needs it less.

2. **`session.status` path (new in 5.0)** — when the session status becomes `"retry"` (i.e. a retry is already in progress and may still succeed), the manager immediately calls `tryFallbackRetry(task, {name: "SessionRetry", ...}, "session.status")`, with an equivalent in the polling path.

Note the asymmetry: `session-status-classifier.ts` classifies `"retry"` as an **active** status (the poller correctly skips completion for active sessions), while the event/polling handlers simultaneously treat the same status as an immediate failover trigger.

----

For exactly the transient errors that a plain retry would likely fix:

- The task is immediately moved to a fallback model. When the original model would have started working again moments later, this is an unnecessary capability downgrade for the remainder of the task.
- If the fallback chain contains no actually-servable model, the task dies outright instead of recovering (see related issue below).

Real-world observation: a background task against a provider that intermittently returns rate-limit / internal-error responses. Timeline reconstructed from logs:

```
T+0ms     stream fails with a transient server error (rate limiting)
T+4ms     session cancelled
T+15ms    fallback attempt #1 created on the next chain entry → dies at model resolution
T+50ms    fallback attempt #2 → same
T+90ms    fallback attempt #3 → same → task declared dead
```

The provider recovered a few minutes later; concurrent sessions on the same primary model continued to work normally. The failed task never got a single retry on the original model.

## Related: runtime fallback-chain reachability is provider-level, not model-level

`tryFallbackRetry` (fallback-retry-handler.ts) filters candidates with a **provider-level** reachability check (`isReachable` only checks whether any of the entry's providers is in the connected set), then `selectFallbackProvider` picks a provider. Gateway-style providers can pass the "connected" check while being unable to serve the specific model, so the fallback session dies at model resolution (`ProviderModelNotFoundError`) before any network request is made — burning one chain entry per attempt.

f245668c0 ("fix(model-routing): exclude gateway providers from fallback lanes") fixes this for the **install-time config-generation** path (`cli/model-fallback.ts`), but the **runtime** path — `task.fallbackChain` resolved from the built-in requirements (`category-resolver.ts`) plus the reachability check inside `tryFallbackRetry` — is not covered by that exclusion.

### Proposed Solution

An opt-in grace period before fallback takeover on retryable errors, e.g. `fallbackDeferMs` in the background-task config (default `0` = current behavior, no change):

- When a retryable fallback would be triggered (`session.status` retry event / polling, or `session.error`) and `fallbackDeferMs > 0`, defer the takeover with a one-shot timer;
- when the timer fires, proceed with `tryFallbackRetry` only if the session has not recovered (still failing / dead); skip the fallback if the session recovered on its own;
- a terminal `session.error` arriving during the grace period cancels the deferral and falls back immediately.

This keeps the current aggressive-failover behavior as the default, while letting users whose providers produce short-lived transient errors give the retry a chance to succeed before switching models.

Happy to send a PR if the direction is acceptable. Two open questions for maintainers:

1. Is the immediate takeover on `status: "retry"` intentional (fast failover preferred over waiting out recovery)?
2. Should the runtime fallback chain also adopt model-level availability validation (extending f245668c0's policy to the runtime path)?

### Alternatives Considered

_No response_

### Doctor Output (Optional)

```shell

```

### Additional Context

_No response_

### Feature Type

Configuration Option

### Contribution

- [x] I'm willing to submit a PR for this feature
- [ ] I can help with testing

AceRothstein71:
Fixes #7319

## What
New opt-in config `background_task.fallbackDeferMs` (default `0`, which keeps behavior byte-identical). When set above zero, retryable non-terminal triggers defer fallback takeover behind a one-shot per-task timer (first trigger wins, repeats neither extend nor duplicate it), and a fire-time probe takes over only if the session never recovered: busy or running skips, idle with output skips, still-retrying / idle-without-output / dead session proceed. Terminal and quota-stop errors bypass the grace window entirely; launch-failure sources stay immediate.

## Why
Four `tryFallbackRetry` call sites (session.error handler, session.status retry event, the 3s poller retry branch, message.updated assistant errors) took over on the FIRST retryable signal, so a transient 429/5xx downgraded the subagent model instantly while OpenCode's own in-flight retry was still running.

## Proof
- RED: 9 failures, all "downgraded despite configured grace"; GREEN 22/22 across three consecutive runs; full background-agent directory 765/765.
- Config schema regeneration plus freshness/drift tests 11/11; `git diff --check` and hygiene scan clean; tsgo exit 0.
- Optional live drive with real `opencode serve`, the worktree-built plugin, and SSE-pumped events proved the terminal-error bypass end-to-end in an isolated sandbox.

Evidence committed under `.omo/evidence/20260826-issue7319-bg-fallback-retry/`.

## Honest limitations
The HTTP 500-then-success half of the live drive could not run offline (provider SDK install blocked in the environment); that half is covered by integration fakes over the real manager seam and recorded as an honest blocker in the QA notes.


packages/omo-opencode/src/features/background-agent/manager.ts
```diff
@@ -2138,6 +2154,124 @@ The fallback retry session is now created and can be inspected directly.
...
+    const sessionID = task.sessionId
+    const deferredSource = `${source}:deferred`
+    if (!sessionID) {
+      await this.tryFallbackRetry(task, errorInfo, deferredSource)
+      return
+    }
+
+    const statuses = await this.readSessionStatusMap(task.id)
+    if (!statuses) {
+      await this.tryFallbackRetry(task, errorInfo, deferredSource)
```
code-yeongyu, line 2203: [sisyphus-bot] This await creates a cancellation race: the task can be cancelled, completed, or replaced while the status request is in flight, but after the probe returns this method calls `tryFallbackRetry` without re-checking `this.tasks.get(task.id) === task` and `task.status === "running"`. `tryFallbackRetry` mutates attempt/model/status and queues a new attempt, so a late response can resurrect a cancelled task or enqueue work after completion. Re-check the identity/status after every awaited probe (and before the fallback call), or use a task-generation token captured at scheduling time.

packages/omo-opencode/src/features/background-agent/manager.ts
```diff
@@ -2138,6 +2154,124 @@ The fallback retry session is now created and can be inspected directly.
...
+    task: BackgroundTask,
+    errorInfo: { name?: string; message?: string; statusCode?: number },
+    source: string,
+  ): boolean {
+    const deferMs = this.config?.fallbackDeferMs ?? 0
+    if (deferMs <= 0) return false
+    if (!shouldRetryError(errorInfo)) return false
+    if (isTerminalSessionError(errorInfo)) return false
+
+    const scheduled = this.fallbackDeferral.schedule(task.id, deferMs, () => {
```
code-yeongyu, line 2167: [sisyphus-bot] The tracker only deduplicates until the timer callback starts. The callback deletes the task from the pending map before this async probe runs; a repeated retry/error event during `readSessionStatusMap` can therefore schedule a second timer, and both callbacks may later call `tryFallbackRetry` against the same running task. That can advance `attemptCount` twice and queue multiple fallback sessions. Track an in-flight decision (or atomically claim the task before probing) and make the takeover one-shot across the whole async operation, not just until timer fire.

packages/omo-opencode/src/features/background-agent/manager.ts
```diff
@@ -2138,6 +2154,124 @@ The fallback retry session is now created and can be inspected directly.
...
+          source,
+          sessionStatus: status.type,
+        })
+        return
+      }
+      await this.tryFallbackRetry(task, errorInfo, deferredSource)
+      return
+    }
+
+    this.logger("[background-agent] Skipping deferred fallback: terminal or unknown session status:", {
```
code-yeongyu, line 2250: [sisyphus-bot] This branch drops the only fallback action for every unknown status. `isActiveSessionStatus` deliberately returns false for unknown values (and logs), so a newly introduced server state such as a transient `compacting`/`awaiting` state reaches here, consumes the one-shot deferral, and leaves the task running forever with no retry or follow-up timer. Unknown status handling must be conservative and continue probing/retrying (or reuse the existing poller's unknown-status path), rather than silently treating it as recovered.

packages/omo-opencode/src/features/background-agent/manager.fallback-defer.test.ts
```diff
@@ -0,0 +1,505 @@
...
+    type: "session.error",
+    properties: {
+      sessionID: sessionId,
+      error: { name: "UnknownError", data: { message } },
+    },
+  })
+}
+
+async function drainEventHandlers(): Promise<void> {
+  await new Promise<void>((resolve) => setTimeout(resolve, 0))
```
code-yeongyu, line 136: [sisyphus-bot] nit: this fixed `setTimeout(0)` yield violates the repository review/test discipline for async behavior and makes the tests timing-dependent. It does not subscribe to the exact completion signal from the void-dispatched handler; it merely hopes one macrotask is enough. Inject/await the handler's actual promise or expose a deterministic event-drain hook so the assertions cannot race microtasks or I/O scheduling.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu, line 21: [sisyphus-bot] This is an unrelated and overly broad weakening of a repository-wide safety test: every tracked Markdown file under `.omo/` and `.opencode/` is now invisible to link auditing, not just generated evidence or the specific files causing the failure. That permits broken local links and maintainer-machine paths in user-facing metadata/docs indefinitely. Narrow the exclusion to the generated artifact set (or fix the evidence Markdown), and add a test proving the intended scope rather than skipping whole trees.

packages/omo-opencode/src/features/background-agent/manager.ts
```diff
@@ -2138,6 +2154,124 @@ The fallback retry session is now created and can be inspected directly.
...
   }
 
+  private maybeDeferFallbackTakeover(
+    task: BackgroundTask,
+    errorInfo: { name?: string; message?: string; statusCode?: number },
+    source: string,
+  ): boolean {
+    const deferMs = this.config?.fallbackDeferMs ?? 0
+    if (deferMs <= 0) return false
+    if (!shouldRetryError(errorInfo)) return false
```
code-yeongyu, line 2164: [sisyphus-bot] The gate is based only on `shouldRetryError`, while the actual fallback handler also permits `isProviderExhaustionFallbackEligible(errorInfo)`. A provider-exhaustion error that is eligible for fallback but not classified by `shouldRetryError` bypasses this new grace window and falls straight through to immediate takeover, contradicting the PR's stated rule that deferral covers retryable/provider-exhaustion fallback decisions. Use the same eligibility predicate as `tryFallbackRetry`, then separately preserve the intended terminal/quota bypass semantics with tests for both classifier combinations.

code-yeongyu (changes requested):
[sisyphus-bot] Request changes. This implementation adds a real asynchronous state machine around a hot lifecycle path, but it does not close the races created by the new timer/probe boundary. A cancellation or a second retry signal can arrive while the status probe is awaiting I/O, and the callback can then mutate/requeue a task that is no longer running or can launch fallback twice. The probe also silently abandons tasks for any status outside a small allowlist, which is unsafe as the server evolves. The advertised verification is not reproducible from this checkout: the added manager suite is recorded as green in evidence, but the diff itself includes a 9-failure run, and my direct run cannot load the missing workspace dependency.

- Fix the callback/probe lifecycle with a generation or in-flight guard and re-check task identity/status immediately before takeover.
- Keep a pending/retry decision alive (or route unknown statuses through the existing polling/completion logic) instead of returning without a follow-up.
- Do not broaden the markdown audit exclusion to entire metadata trees; make evidence links valid or narrowly exclude only generated artifacts.
- Replace the real-time sleep in the tracker test with an injectable/manual scheduler, and correct the configuration prose so it matches actual fallback eligibility.

MoerAI:
Thanks for the PR, @AceRothstein71, and for taking the time on this.

This commits generated schema outputs at `assets/oh-my-opencode.schema.json` and `assets/omo.schema.json`. Generated schemas are maintained by the repository build pipeline and are not accepted through routine contributor PRs.

I am closing this branch. Please open an issue with the fallback deferral contract, or submit a fresh source-only PR against `dev` that changes the schema source and leaves generated assets to the maintainer pipeline.

Issue #7339, opened 2026-08-25
bug(omo-task): worker creation can crash host before barrels load


lavi-kj:
## Summary

Starting a native DAG can terminate the OMO host with an uncaught pi-tui initialization error:

```text
Error: The @earendil-works/pi-tui barrel was accessed before it was loaded.
Await loadPiTui() at the registration entry point before reading pi-tui values synchronously.
```

The stack points into `plugin/extensions/omo-task.js` and ends at `Timeout._onTimeout`, so this appears to be an asynchronous extension-registration race rather than an error in the dispatched workload.

## Environment

- Windows 11 x64
- Node.js 24.15.0
- `omo-ai` 5.0.0-beta.20
- bundled `@code-yeongyu/senpi` 2026.8.25
- bundled `@earendil-works/pi-tui` 2026.8.25

`npm ls -g --all --depth=4 @earendil-works/pi-tui @code-yeongyu/senpi omo-ai` showed the OMO dependency subtree resolving both Senpi and pi-tui to 2026.8.25, so this did not look like a mixed-version install.

## Reproduction

1. Start OMO 5.0.0-beta.20.
2. Start a native DAG (either through the JavaScript DAG SDK or the native `dag start` tool).
3. After the DAG start acknowledgement, wait for the task extension's delayed update.

## Actual behavior

The host exits with an uncaught exception. One observed stack was:

```text
Error: The @earendil-works/pi-tui barrel was accessed before it was loaded.
Await loadPiTui() at the registration entry point before reading pi-tui values synchronously.
    at Ld (.../plugin/extensions/omo-task.js:3:144893)
    at Bd (.../plugin/extensions/omo-task.js:3:145248)
    at oT (.../plugin/extensions/omo-task.js:3:583491)
    at tT (.../plugin/extensions/omo-task.js:3:582700)
    at .../plugin/extensions/omo-task.js:3:586445
    at Array.flatMap (<anonymous>)
    at a (.../plugin/extensions/omo-task.js:3:586430)
    at Timeout._onTimeout (.../plugin/extensions/omo-task.js:3:587025)
```

This can happen after `dag start` reports that a run was started, leaving the caller unable to inspect whether workers were dispatched or to collect their state.

## Expected behavior

The task extension should await `loadPiTui()` before any synchronous pi-tui access during registration or delayed task/DAG UI updates. A DAG start must not terminate the OMO host.

## Impact / temporary workaround

[...]

PR #7346, opened 2026-08-26
fix(dag): settle snapshots after recovery and adopt paused runs across sessions

### LINKED ISSUE #7316: dag: omo.dag.updated can emit runs:[] while a run is live (session-scoped filter + attach-before-recovery)
## Summary

`omo.dag.updated` (the wholesale dag-runs snapshot the RPC host consumes) can report `runs: []` while a dag run is still live, and a run whose parent session id changed (fork / compaction / restart under a new id) becomes permanently invisible AND unrecoverable. Hosts that follow the wholesale contract (a live run missing from a snapshot has completed — omo-desktop-app does) flip the thread to Ready while children keep working.

Observed live on omo-desktop 2026-08-25: a thread's session emitted `thread.dag-runs-updated` with `dagRuns: []` while its work was still progressing after a host-server restart.

## Defect 1 — attach() emits the snapshot before recovery runs

`packages/omo-senpi/src/components/task/dag-runtime.ts:437-446`: `runtime.attach()` calls `bridge.attach()` (which resets the snapshot fingerprint and schedules an emit, `dag-rpc-bridge.ts:149-159`) BEFORE `recovery.resumePausedRuns(sessionId)`. The first post-restart snapshot therefore reflects pre-recovery on-disk state (paused / stale nodes) rather than the recovered, scheduled runs.

Smallest fix: reorder `bridge.attach()` after `recovery.resumePausedRuns(...)` + the `ensureScheduled` loop (after `dag-runtime.ts:450`).

## Defect 2 — parentSessionId-scoped listing orphans runs across session-id changes

- `snapshotsForActiveSession` (`dag-runtime.ts:562-567`) returns `[]` when `engine.runtime.sessionId()` is undefined or when `manager.list(sessionId)` (`packages/senpi-task/src/dag/manager.ts:238-258`, filters `record.parentSessionId === parentSessionId`) finds no records.
- `recovery.resumePausedRuns` skips such runs as `foreign_session` / `not_paused` (`recovery.ts:134-135`), so after a fork/compaction/new-session-id restart the run is neither shown nor resumable — while its node children can still be alive and delivering notifications.
- `omo.dag.list` (`dag-rpc-handlers.ts:109-123`) has the same scoping, so the host cannot query its way out.

The dag checkpoint vocabulary already carries lineage-ish fields elsewhere (task records have `root_session_id`, `senpi-task/src/store/record-parse.ts:63`), but dag run records key ownership solely on `parentSessionId` (`senpi-task/src/dag/types.ts:193`).

Direction to decide: key dag run ownership by session lineage / root session id (security-sensitive: multi-session hosts must not leak runs across genuinely foreign sessions), or define an explicit adoption path on resume.

## Semantics note for hosts

An empty wholesale snapshot currently means "no runs visible to THIS session id", not "every previously known run finished". Until ownership keying is lineage-aware, hosts cannot distinguish the two.

## Host-side mitigation landed (omo-desktop-app)

omo-desktop-app now synthesizes background-liveness from `omo.task.updated` subagent snapshots as well (task children — including dag node children — keep the thread Working), so the user-facing Ready-while-working symptom is mitigated even when the dag snapshot lies. The dag snapshot itself remains untruthful across session-id changes and should be fixed here.

AceRothstein71:
Fixes #7316

## What
- `attach` no longer emits wholesale snapshots before recovery finishes. The bridge attaches immediately (event ledger and heartbeat stay live, so recovery-era events still reach viewers), but snapshot flushes are suspended until recovery and scheduling settle, then one honest snapshot is scheduled. New seam: `DagRpcBridge.setSnapshotsSuspended()`.
- Paused runs orphaned by session-id changes (fork, compaction, restart under a new id) are explicitly adopted on resume, but only with durable proof the predecessor host process is dead (`leaseHolderPid ?? previousLeaseHolderPid` defined and not alive). Ownership rewrite and lease claim happen in one checkpoint write under the run lock, `rootSessionId` is preserved, and a live holder is refused to protect multi-session hosts sharing one process.

## Proof
- Three mechanism-exact RED tests (adoption timeout, stale mid-recovery emission, unit adoption refusal), then green.
- Sweeps, twice consecutively: senpi-task dag suite 251/251, omo-senpi task component 488/488, full omo-senpi package 2273 pass / 0 fail (7 pre-existing capability skips). tsgo clean for both packages; `git diff --check` and hygiene scan clean.

Evidence committed under `.omo/evidence/20260826-issue7316-dag-live-runs/` plus the adapter-scoped copy required by repo QA policy.

## Honest limitations
- Live senpi-binary harness QA unavailable here (binary absent); the paused-header driver additionally trips a pre-existing pi-tui warm-up defect present on the untouched base (tracked as #7339), stash-verified.
- Adoption keys on host liveness rather than thread lineage; full lineage keying would need host APIs that do not exist yet. Documented as a residual limitation in the recovery module.


code-yeongyu (changes requested):
[sisyphus-bot] REQUEST CHANGES. This PR is materially incomplete and must be rebased onto current dev before it can be evaluated as a mergeable change: it is marked DIRTY, and the implementation does not match its own stated live-bridge design. The native recovery path has serious ownership and observability hazards, while several additions are evidence/generated churn rather than product changes.

- The PR claims the bridge attaches immediately with snapshots suspended, but runtime.attach still waits for recovery and then calls bridge.attach; the new suspension API is never wired into the runtime, so recovery-era ledger events are not delivered at all.
- Foreign adoption rewrites rootSessionId to the adopter, contradicting the stated preservation requirement and changing ancestry semantics for descendants.
- PID liveness is only a best-effort abandonment proof and is vulnerable to PID reuse; adopting based solely on a dead numeric PID can steal a paused run from the wrong host lineage.
- The targeted recovery test suite fails in the PR checkout (the recorded RED-recovery evidence points at the same failure), and diff-check fails on checked-in trailing whitespace.
- nit: the PR mixes generated bundles, schema churn, unrelated evidence rewrites, and large committed transcripts into this focused fix, making review and release provenance unnecessarily noisy.

Anchored findings (fallback):
- [sisyphus-bot] The new suspension seam is not integrated with createDagRuntime: runtime.attach still performs recovery first and only then calls bridge.attach, and no caller invokes setSnapshotsSuspended. The advertised design (attach immediately so the event ledger/heartbeat remain live while only snapshots are held) is therefore not implemented; recovery-era omo.dag.event records are invisible to connected viewers. Add a runtime-level test with blocked recovery proving events reach RPC while omo.dag.updated remains suppressed.
- [sisyphus-bot] This overwrites rootSessionId during adoption. The PR description explicitly says rootSessionId is preserved, and root ancestry is used when spawning descendants; changing it makes a restarted/forked run appear rooted in the adopter and can alter lineage/routing semantics for every later child. Preserve the persisted root (only re-home ownership/parent), and assert this for a foreign run whose root differs from its old parent.
- [sisyphus-bot] A dead numeric PID is not durable proof that this is the predecessor host: PIDs are reused, especially across restart-heavy native sessions. This can adopt a paused run belonging to a different process that happened to inherit the old PID. Persist and validate a host/lease generation or another non-reused identity, or make this ambiguity explicit and test the reuse scenario.
- [sisyphus-bot] Returning here while suspended drops the scheduled flush permanently (snapshotFlush is cleared on entry and never re-armed). This is safe only if every suspension release is guaranteed to call scheduleSnapshot; callers currently do not use the new API, and a suspend/attach/detach race can leave a consumer with no post-recovery snapshot. Track a dirty/pending bit while suspended and flush it on release instead of relying on caller timing.
- [sisyphus-bot] This unit test exercises the bridge seam in isolation, but production runtime never calls it. It passes despite the actual missing integration that causes the recovery-order bug. Add a dag-runtime-level test with a durable event emitted during blocked recovery, asserting the event reaches RPC while no wholesale snapshot does.
- [sisyphus-bot] nit: this checked-in evidence line has trailing whitespace, so git diff --check fails on the PR despite the description claiming it is clean. Remove whitespace from this log and the corresponding RED-runtime-adopt/RED-runtime-order evidence lines.

code-yeongyu:
[sisyphus-bot] this problem appears already fixed on dev via 0f959ca5e fix(omo-senpi): attach the dag bridge after recovery, not before (packages/omo-senpi/src/components/task/dag-runtime.ts:455) and ea875b63d fix(senpi-task): adopt orphaned dag runs whose owner session is gone (packages/senpi-task/src/dag/recovery.ts:150); this PR is likely obsolete - please verify and close or rebase to show a remaining delta

KNN-07:
Closing as part of the #7417 backlog cleanup: issue #7316 is closed after the replacement fix merged in #7376. Thank you for the contribution.

PR #7348, opened 2026-08-26
fix(senpi): keep pi-tui loader alive and warm senpi barrel before DAG runs

### LINKED ISSUE #7339: bug(omo-task): worker creation can crash host before barrels load

AceRothstein71:
Fixes #7339

## What
- Restore `await loadPiTui()` at task-component register() so `omo-task.js` keeps an in-graph caller for the lazy boundary; the compose-level warm-up stays for omo.js consumers.
- `dag start` / `amend` now await `loadSenpiBarrel()` when skill materialization is configured and the barrel is still cold, preserving the synchronous fast path required by node retry (`void manager.amend(...)`).
- New `isSenpiBarrelLoaded()` seam; stale warm-up contract comments corrected; affected extension bundles rebuilt.

## Root cause
The beta.20 warm-up move left `omo-task.js` without any in-graph caller for the pi-tui lazy boundary, so the bundler dead-code-eliminated the loader and constant-folded the boundary into an unconditional throw. Worker creation then crashed through store mutation -> 250ms status-widget debounce -> row helpers -> throw inside `Timeout._onTimeout`, taking the host down (exact stack from the issue). Independently, DAG run creation materialized node skills through the default fs loader without ever awaiting the senpi barrel, rejecting before dispatch.

## Proof
- RED probes reproduce both failures cold, GREEN after the fix (logs committed under `.omo/evidence/20260826-issue7339-worker-barrel-crash/`).
- Focused suites twice: 254 pass / 0 fail. `tsgo --noEmit` clean for senpi-task and omo-senpi. `git diff --check` and hygiene scan clean.
- Artifact QA through the rebuilt `omo-task.js` in an isolated sandbox: 11 tools registered, completion notice rendered through the bundle's own boundary.

## Honest limitations
- Live senpi-binary harness QA was not runnable on this machine (binary absent); compensated with real-artifact QA, documented as blocker rather than pass.
- Bundles were rebuilt with local bun 1.3.14 while CI pins 1.4.0; if the committed-bundle freshness gate complains on CI, regeneration on the CI toolchain may be needed.


AceRothstein71:
Closing this draft submission: the fork mirror is 143 commits behind upstream dev, so this branch was reconstructed on the stale base and GitHub reports it CONFLICTING even though the underlying work is complete and fully evidenced. Will resubmit cleanly against current dev shortly.

PR #7349, opened 2026-08-26
fix: correct GPT context metadata

yeongjun-cigro:
## Summary

- correct stale 372k context metadata for GPT-5.6 Sol/Terra aliases and the GPT-5.6 family alias
- align the supplemental GPT-5.5 metadata with the authoritative 1.05M context and 922k input limits
- add regression coverage for all affected GPT-5.x entries

## Why

The generated model snapshot contained impossible `context=372000` / `input=922000` combinations for GPT-5.6 models. That can make OMO compact or reject prompts as though the model had a much smaller context window.

## QA

- `bun test packages/model-core/src/model-capabilities.test.ts` — 22 passed, 0 failed
- `tsgo --noEmit -p packages/omo-senpi/tsconfig.json` — passed
- `bun run test:senpi` — 2,274 passed, 1 skipped, 0 failed
- Senpi driver self-test — passed
- Evidence: `.omo/evidence/omo-senpi-adapter/20260826-gpt-context-metadata/README.md`

## Risk

Metadata-only change; no provider request or routing behavior was changed.


MoerAI (changes requested):
Thanks for identifying the stale GPT metadata. Please rebase onto current `dev` and sign the CLA so the required CI matrix can run.

Do not hand-edit the generated snapshot; #7531 already regenerates it. Put durable policy in complete supplemental entries for `gpt-5.5` and generic `gpt-5.6`, with the canonical `1,050,000` context and `922,000` input rationale linked to models.dev.

Replace the regression table so it verifies the effective bundled value for generic `gpt-5.6` as well as GPT-5.5, Sol, Terra, and Luna. Add OpenCode-scoped QA evidence, and either prove the claimed compaction impact through its real runtime path or narrow the claim to metadata.

PR #7350, opened 2026-08-26
fix(senpi-task): share pi-tui warm-up state across duplicated bundle copies

sanguneo:
## What changed
- `packages/senpi-task/src/lazy/pi-tui.ts`: the lazy pi-tui warm-up state (`module`, `promise`) now lives on `globalThis[Symbol.for("omo.senpi-task.piTui")]` instead of module-local variables.
- New test `pi-tui-shared-state.test.ts` reproduces two independent module instances and asserts a warm-up through one satisfies `piTui()` through the other.
- Regenerated `plugin/extensions/omo.js` and `omo-task.js` (Linux, bun 1.4.0, frozen lockfile; `build-extension.mjs --check` reports current).

## Why
A live omo-ai session running `mass-ulw` died with an uncaughtException from `omo-task.js`: "The @earendil-works/pi-tui barrel was accessed before it was loaded". `omo-task.js` is its own bundle and carries a private copy of the lazy boundary; `compose.ts` awaits `loadPiTui()` only through the `omo.js` copy, so the DAG status widget timer (`dag-status-ui.ts`, `runs.flatMap(runRows)` -> `renderer-text` -> `piTui()`) read a cold copy and threw outside any try/catch.

## Observed behavior
- RED before fix: the new test failed with the exact production error message.
- GREEN after fix: 1 pass; `compose-cold-pi-tui.test.ts` still passes; `tsgo --noEmit` clean for `senpi-task` and `omo-senpi`.
- Both `omo.js` and `omo-task.js` contain the shared symbol key; `omo-member.js` (no pi-tui reader) unchanged.

## QA / evidence
Local `.omo/evidence/20260826-pi-tui-shared-warmup/` (README + build log; directory is gitignored).

## Residual risk
A bundle that reads pi-tui without any in-process warm-up still throws by design; the guard message is unchanged.


sanguneo:
as i mentioned, omo always died when call mass-ulw, please check this pr

sanguneo:
Closing in favor of #7352, which fixes the same crash with a broader change: it keeps the loader alive inside the task bundle AND hardens the status widget render path, plus a bundle-level regression test. My diagnosis matched theirs independently, so this PR adds nothing beyond it.

For the record, two details confirmed here that may help review #7352:

- In the shipped 5.0.0-0.beta.20 `omo-task.js`, the accessor is emitted as `function Ld(){throw Error("The @earendil-works/pi-tui barrel...")}` with no module state and no loader at all, while `omo.js` keeps the full `function Bf(){if(void 0===zf)throw ...;return zf}` shape plus `Uf??=import("@earendil-works/pi-tui").then(e=>(zf=e,e))`. So the task bundle could never have been warmed from inside itself.
- The DAG journal from a crashed run stops exactly at `dag.wave.started` with no `dag.node.started` event, i.e. the process dies on the widget's first timer render right after wave 0 is scheduled, before any child spawns.
- Windows-built bundles differ from the ubuntu CI build; regenerating them on Linux with bun 1.4.0 and a frozen lockfile is what makes `senpi-compatibility (ubuntu-latest)` pass.

code-yeongyu:

PR #7352, opened 2026-08-26
fix(omo-senpi): prevent task bundle pi-tui crashes

amsminn:
## Summary

`omo-ai@5.0.0-0.beta.20` can terminate when the task status widget renders:

```text
    at Ld (.../plugin/extensions/omo-task.js)
    ...
    at Timeout._onTimeout (.../plugin/extensions/omo-task.js)
```

In a short local sample after installing beta.20, I observed **4 process-ending crashes while using
7 newly created main sessions**. Those crashes affected **3 distinct sessions (3/7, 43%)**; one of
them crashed twice:

- `01a03b98-69ae-74ca-be3e-79f648513adc`
- `01a03ba1-3bc0-7323-8075-ec2379585dc7`
- `01a03ba5-f3db-7f77-be46-b34c7a46de2b` (two crashes)

This is a conservative local count: child/tool errors returning the same barrel message were
excluded unless the terminal also showed the explicit `OmO exiting due to uncaughtException`
process-ending output.

## Root cause

The build emits `omo.js` and `omo-task.js` as independent bundles. Each inlines its own copy of the
lazy pi-tui module, so warming pi-tui in `omo.js` does not initialize the task bundle's copy.

Because nothing in the `omo-task.js` graph called `loadPiTui()`, terser removed the dynamic import
and collapsed the accessor into an unconditional throw:

```js
function Ld(){throw Error("The @earendil-works/pi-tui barrel was accessed...")}
```

The throw then escaped the status widget's timer callback as an `uncaughtException`.

## Changes

- Await `loadPiTui()` inside the task component's own `register()` path, keeping the task bundle's
  dynamic import reachable and initializing its module instance before renderers register.
- Contain status-UI render failures so one bad frame is logged and skipped instead of terminating
  the process.
- Add an artifact-level regression test requiring every bundle that reaches a lazy barrel to retain
  its dynamic import and a conditional guard.
- Add tests proving direct and timer-driven render failures do not propagate.
- Regenerate the tracked extension bundles.

## Verification

Failing-first evidence:

```text
BEFORE  omo-task.js  dynamicImport=0  unconditionalThrow=1
AFTER   omo-task.js  dynamicImport=1  unconditionalThrow=0
```

- Bundle regression: **2 failed before**, **4/4 passed after**
- Render-fault containment: **3 failed before**, **4/4 passed after**
- `tsgo --noEmit -p packages/omo-senpi/tsconfig.json`: exit 0
- `bun run test:senpi`: **2282 passed, 1 platform skip, 0 failed**
- Scoped extension/task/lazy suites: **552 passed, 0 failed**
- Existing bundle purity and size gates: **3 passed, 0 failed**
- Root typecheck and build: exit 0
- Isolated real-Senpi task driver: no pi-tui/uncaughtException crash marker, no real-home changes,
  and no leaked child processes. Its six remaining follow-up/resume failures all reproduce on
  `origin/dev`; the patched build passes seven additional live checks relative to that baseline.


sanguneo:
Closed my duplicate #7350 in favor of this one - your change is the more complete fix.

Two independently confirmed details in case they are useful for review:

- In the shipped `5.0.0-0.beta.20` bundles the accessor really is collapsed to an unconditional throw in `omo-task.js` (`function Ld(){throw Error(...)}`, no module state, no loader), while `omo.js` retains `function Bf(){if(void 0===zf)throw ...;return zf}` together with `Uf??=import("@earendil-works/pi-tui").then(e=>(zf=e,e))`.
- A DAG journal from one of these crashes ends exactly at `dag.wave.started` with zero `dag.node.started` events, which pins the death to the widget's first timer render immediately after the wave is scheduled - no child had spawned yet. That matches your `Timeout._onTimeout` stack.

Also worth noting for the bundle diff: bundles regenerated on Windows do not match the ubuntu CI build; building on Linux with bun 1.4.0 and `bun install --frozen-lockfile` is what makes `senpi-compatibility (ubuntu-latest)` accept them.

code-yeongyu:

PR #7364, opened 2026-08-26
fix(model-core): honor disabled_providers in static fallback chain resolution

### LINKED ISSUE #6573: bug(model-routing): hardcoded AGENT_MODEL_REQUIREMENTS fallback chain ignores disabled_providers (claude-opus-5 selected on transient errors)
## Summary

The hardcoded per-agent `AGENT_MODEL_REQUIREMENTS` fallback chains in oh-my-openagent ignore `disabled_providers`. On a transient provider error, the fallback chain can select a model from a **disabled** provider, causing mystery model switches and 'insufficient balance' errors.

## Reproduction

1. Configure `disabled_providers: ["openai", "anthropic", "google", "xai"]` (Anthropic disabled, no API credits configured).
2. Run the main `sisyphus` agent on `opencode-go/deepseek-v4-flash` (a connected provider).
3. Trigger a transient/retryable provider error (e.g. 429 rate limit, provider connection error) on the primary model.
4. `resolveModelPipeline` (dist ~24739) walks the hardcoded `policy.fallbackChain` — for sisyphus, `fallbackChain[0] = { providers: ["anthropic", "github-copilot", "opencode", "vercel"], model: "claude-opus-5", variant: "max" }`.
5. The connected-set gate passes via the **'opencode'** provider (present in the connected-providers cache) — not even requiring anthropic to be connected.
6. claude-opus-5 is selected → user sees the model switch without manual action → 'insufficient balance' (no Anthropic credits; `insufficientcreditserror` is a STOP-class error).

## Evidence

- `~/.cache/oh-my-opencode/connected-providers.json`: `{"connected": ["opencode-go", "opencode", "deepseek", "openrouter"]}` — 'opencode' is connected.
- `~/.cache/oh-my-opencode/provider-models.json`: claude-opus-5 present in the model catalog (fuzzy-match branch also resolves it).
- `disabled_providers` only gates provider **listing** (`isProviderAllowed`, packages/opencode/src/provider/provider.ts:1388), not the hardcoded fallback chain in the plugin dist.

## Impact

- Silent model switch to a disabled-provider model on any transient error.
- 'insufficient balance' errors (Anthropic credits absent) when the fallback lands on anthropic models.
- No config way to suppress it (verified: no `fallback_models` key → hardcoded chain runs unfiltered; the chain is only consulted after user fallback_models fails).

## Workaround (local)

Adding an explicit `fallback_models` allowlist (e.g. `["opencode-go/deepseek-v4-flash", ...]`) to each agent/category config suppresses the hardcoded chain, because `resolveModelPipeline` checks `intent.userFallbackModels` BEFORE the hardcoded `policy.fallbackChain`. A regression test now enforces that no fallback entry uses a disabled provider.

## Suggested fix

The hardcoded fallback chains should be filtered by `disabled_providers` (same `isProviderAllowed` gate used for listing), or the chain lookup should skip providers in the disabled set.

AceRothstein71:
Fixes #6573

## What
Static fallback-chain resolution (`AGENT_MODEL_REQUIREMENTS` / `CATEGORY_MODEL_REQUIREMENTS`) now accepts `disabledProviders` constraints and excludes any entry referencing a disabled provider (case-insensitive, trimmed), threaded from `pluginConfig.disabled_providers` through agent registration (sisyphus/hephaestus/atlas/general/prometheus builders), the model-fallback hook arm-time filter, delegate-task category resolution and subagent model resolution, call_omo_agent, and the dynamic vision chain. User-configured chains keep their existing `applyDisabledProviders` behavior and explicit user models are untouched; the static tables remain pure data with filtering at resolution time.

## Root cause
The hardcoded tables were gated only on provider connectivity. On a transient error, the sisyphus chain's first entry reached claude-opus-5 via the `opencode` aggregator even with anthropic disabled, surfacing Anthropic's insufficient-balance STOP error instead of falling onward.

## Semantics note
An entry is excluded when ANY of its providers is disabled. Provider-level skipping inside an entry would not fix the repro because mirrors serve the same underlying model; strict exclusion matches configuration intent and is called out explicitly in the self-audit ledger.

## Proof
- RED log saved; GREEN 9/9 targeted, full model-core suite 365/365.
- Focused gates across seven areas twice on identical tree: 1298 pass / 0 fail; tsgo exit 0 for model-core and omo-opencode; diff-check and hygiene clean versus base.
- Isolated QA in an XDG sandbox: transient error with anthropic disabled selects kimi-k3 via opencode-go and claude-opus-5 is never chosen; default-config guard proves byte-identical pre-fix behavior when nothing is disabled.

Evidence committed under `.omo/evidence/20260826-issue6573-model-reqs-disabled-providers/`.

## Residual notes
Doctor diagnostics and install-time generation still read raw chains (documented P3): with providers disabled, doctor may describe models that registration would filter.


packages/omo-codex/plugin/components/codegraph/dist/cli.js
```diff
@@ -5882,9 +5882,25 @@ var OmoTelemetrySettingsSchema = OmoTelemetrySettingsLayerSchema.extend({
   enabled: boolean2().default(true)
 }).strict();
 
+// ../../omo-config-core/src/schema/format-on-mutation.ts
```
code-yeongyu: [sisyphus-bot] Blocking: this generated codegraph bundle adds `formatOnMutation` schema and runtime changes that are unrelated to disabled-provider fallback resolution. It is a shipped artifact, not harmless formatting: accepting it changes omo-codex behavior and its generated dependency surface. Drop it from this PR or identify the exact source change and reproducible build command that makes this delta necessary.

packages/omo-codex/plugin/components/codegraph/dist/cli.js
```diff
@@ -9588,17 +9607,20 @@ function defaultPluginRoot() {
...
   closeSync,
   fsyncSync,
   openSync,
   renameSync as renameSync2,
+  rmSync as rmSync2,
   unlinkSync as unlinkSync2,
   writeFileSync as writeFileSync4
 } from "node:fs";
+import { randomUUID as randomUUID2 } from "node:crypto";
+import { dirname as dirname8 } from "node:path";
```
code-yeongyu: [sisyphus-bot] Blocking: the bundled atomic-write implementation is being changed here (unique temp names, fsync/platform behavior) even though this PR is about model requirements. This is unrelated executable behavior in the native/codex product and has no source-level review trail in the change. Rebase, remove the bundle churn, and regenerate only artifacts justified by this fix.

packages/omo-senpi/plugin/extensions/omo-task.js
```diff
@@ -1,2 +1,2 @@
...
...")||n.has(e.name));var t})}function $g(e){if("string"!=typeof e.sessionDir||0===e.sessionDir.length)throw new Sg({kind:"session-create-failed",message:`Child ${e.taskId} has no sessionDir; refusing to fall back to a default session location.`});return e.sessionDir}function Eg(e){let{spec:t,sessionManager:n,uiOnlyToolNames:r}=e,i=function(e,t,n={}){return[...xg(e,n),...t??[]]}(e.sharedParentTools,t.memberScopedTools,{uiOnlyToolNames:r}),o=void 0!==t.agentType&&jc.has(t.agentType)?[...i.filter(e=>"bash"!==e.name),vg(t.cwd)]:i,s=function(e,t){let{SettingsManager:n}=Vh();return void 0===e||void 0===t||0===t.length?n.inMemory({retry:{modelFallback:!1}}):n.inMemory({retry:{modelFallback:!0,fallbackChains:{[e]:t.map(Ig)}}})}(t.selectedModel,t.fallbackModels);return{cwd:t.cwd,sessionManager:n,resourceLoader:function(e){let t={extensions:[],errors:[],runtime:e.runtime};return{getExtensions:()=>t,getSkills:()=>({skills:[],diagnostics:[]}),getPrompts:()=>({prompts:[],diagnostics:[]}),getThemes:()=>({themes:[],diagnostics:[]}),getAgentsFiles:()=>({agentsFiles:[]}),getSystemPrompt(){},getSystemPromptSource(){},getAppendSystemPrompt:()=>[],getAppendSystemPromptSources:()=>[],extendResources(){},
```
code-yeongyu: [sisyphus-bot] Blocking native-release concern: this is an opaque one-line bundle/hash replacement with no corresponding source diff showing what changed in the shipped task extension. Because the PR is classified NATIVE-DIRECT, prove this bundle is reproducibly generated from the rebased source and run a native task smoke test; otherwise do not include this generated change.

packages/model-core/src/model-resolution-pipeline.ts
```diff
@@ -108,6 +110,11 @@ export function resolveModelPipeline(
   const { intent, constraints, policy } = request
   const availableModels = constraints.availableModels
   const fallbackChain = policy?.fallbackChain
+  // The automatic chain walk must never land on a disabled provider. Explicit
+  // user model selection (variant inheritance below) keeps the original chain.
+  const walkableFallbackChain = fallbackChain
+    ? filterFallbackChainByDisabledProviders(fallbackChain, constraints.disabledProviders)
```
code-yeongyu: [sisyphus-bot] The pipeline now filters the same chain here, while the product call sites also pre-filter it. That duplicated policy is easy to let drift (as already happened with the many propagation paths) and makes it unclear whether `resolveModelPipeline` is the enforcement boundary or callers are. Keep one authoritative filtering boundary, then add contract tests around that boundary and the adapter wiring instead of maintaining parallel filters.

packages/omo-opencode/src/hooks/model-fallback/fallback-state-controller.ts
```diff
@@ -1,4 +1,5 @@
 import type { FallbackEntry } from "../../shared/model-requirements"
```
code-yeongyu: [sisyphus-bot] nit: `FallbackEntry`, `filterFallbackChainByDisabledProviders`, and `AGENT_MODEL_REQUIREMENTS` are imported from the same module in three separate declarations. Consolidate this import while touching the file; the current form obscures the fact that all three values share the same source and makes future edits prone to inconsistent import changes.

packages/omo-opencode/src/tools/call-omo-agent/tools.ts
```diff
@@ -97,13 +102,13 @@ function resolveModelAndFallbackChain(args: {
     ?? (agentOverride?.category ? userCategories?.[agentOverride.category]?.fallback_models : undefined)
   )
   const defaultProviderID = model?.providerID
```
code-yeongyu, line 100: [sisyphus-bot] The new coverage does not exercise this product-facing branch: `call_omo_agent` constructs a configured fallback chain here, but there is no added assertion that disabled_providers reaches this factory and prevents an aggregator mirror from selecting the disabled model. Add an integration test through the tool registry/factory wiring, not only a direct model-core test, or this regression can return while the current suite stays green.

packages/omo-opencode/src/tools/delegate-task/category-resolver.ts
```diff
@@ -113,6 +113,10 @@ Available categories: ${allCategoryNames}`)
   }
 
   const requirement = CATEGORY_MODEL_REQUIREMENTS[args.category!]
+  const hardcodedFallbackChain = filterFallbackChainByDisabledProviders(
```
code-yeongyu: [sisyphus-bot] The delegate-task path is another product-facing propagation point, but the PR adds no test at this resolver boundary. In particular, assert the category resolver with an anthropic-disabled chain containing `opencode/claude-opus-5` and verify it falls to the next allowed entry; the model-core tests cannot detect a missing `executorCtx.disabledProviders` wiring.

packages/omo-opencode/src/agents/builtin-agents.ts
```diff
@@ -73,6 +73,7 @@ export async function createBuiltinAgents(
   useTaskSystem = false,
   disableOmoEnv = false,
   teamModeEnabled = false,
+  disabledProviders: string[] = [],
```
code-yeongyu: [sisyphus-bot] nit: this public factory accepts `disabledProviders: string[]` even though every downstream consumer treats it as read-only (`readonly string[]`). Use `readonly string[]` here as well; the current mutable type advertises that the factory may retain or mutate caller-owned configuration and is inconsistent with the newly introduced API contract.

code-yeongyu (changes requested):
[sisyphus-bot] Request changes. This branch is still CONFLICTING (DIRTY) with current dev, so it must be rebased and the full resolution re-audited before merge; the intervening dev changes are not safely reviewable from this merge-heavy head. The PR also mixes the focused disabled-provider fix with unrelated generated omo-codex bundle churn and opaque omo-senpi hash changes, which makes the native product delta non-reproducible and materially increases regression risk. The new model-core tests prove the helper and pipeline in isolation, but do not prove that disabled_providers reaches delegate-task, call_omo_agent, or look_at in the shipped products.

Most serious findings:
- Rebase onto current dev and resolve/review the conflicts rather than relying on the committed merge commits.
- Remove unrelated generated omo-codex artifacts (or provide the exact source/build justification); the diff changes bundled schemas and atomic-write/runtime code unrelated to model fallback behavior.
- Do not land opaque generated native bundle hash changes without a reproducible source-to-bundle proof and native smoke test.
- Add integration tests that exercise each product-facing propagation path and assert a disabled provider cannot be selected through an aggregator mirror.
- Clean up the duplicated filtering/import/type issues called out inline before merge.

AceRothstein71:
@code-yeongyu — rebased onto current `dev`, which clears the CONFLICTING/DIRTY blocker you flagged as the reason this wasn't safely reviewable.

- prove `disabled_providers` actually reaches delegation, not just the helper and pipeline in isolation

packages/omo-codex/plugin/components/codegraph/dist/cli.js
AceRothstein71: Removed. The rebuilt branch contains no CodeGraph source or generated bundle changes; the final 48-file diff is limited to model-core, delegate-core, OpenCode adapter wiring/tests, three reproducibly generated Senpi outputs, and QA evidence.

packages/omo-codex/plugin/components/codegraph/dist/cli.js
AceRothstein71: Removed with the rest of the unrelated CodeGraph churn. No CodeGraph atomic-write or distribution artifact is present in the repaired branch.

packages/omo-senpi/plugin/extensions/omo-task.js
AceRothstein71: Rebuilt from source and narrowed to the three reproducible outputs actually affected: omo.js, omo-task.js, and memory-run-supervisor.mjs. The extension freshness check and test:senpi gate pass. I also drove the repository launcher with pinned Senpi 2026.9.2-4 on both this head and pristine origin/dev: their check matrices match, realSenpiUntouched is true, and neither run leaked child PIDs. The tracked evidence is under .omo/evidence/20260826-issue6573-model-reqs-disabled-providers/.

packages/model-core/src/model-resolution-pipeline.ts
AceRothstein71: Consolidated the policy in model-core. filterFallbackChainByDisabledProviders is the single mirror-entry rule; model-core and delegate-core consume it internally, while product adapters only pass readonly disabledProviders constraints. Explicit user overrides still use the original chain for variant inheritance. Core, delegate-core, and product-boundary regressions cover the contract.

packages/omo-opencode/src/hooks/model-fallback/fallback-state-controller.ts
AceRothstein71: Consolidated the model-fallback controller imports from the shared barrel while implementing the centralized runtime-chain filtering boundary.

packages/omo-opencode/src/tools/call-omo-agent/tools.ts
AceRothstein71: Added a focused integration regression that uses real createCoreTools plus the actual createCallOmoAgent factory. It passes disabled_providers through the registry, invokes registered call_omo_agent, and asserts the machine-consumed BackgroundManager.launch model moves from the mirrored OpenAI/OpenAI-Codex entry to DeepSeek when openai-codex is disabled.

packages/omo-opencode/src/tools/delegate-task/category-resolver.ts
AceRothstein71: Added the resolver-boundary regression in category-resolver.disabled-providers.test.ts. It supplies executorCtx.disabledProviders, starts with a mirrored disabled entry, and proves resolution reaches the next allowed model. The test failed on the missing wiring and passes after category-resolver forwards the constraint into delegate-core.

packages/omo-opencode/src/agents/builtin-agents.ts
AceRothstein71: Updated the public builtin-agent factory path to accept readonly provider arrays. createBuiltinAgents now appends disabledProviders: readonly string[] without shifting existing positional callers, and the nested general/special agent inputs are readonly as well.

AceRothstein71:
Rebuilt this branch from current `dev` as 16 atomic commits. Final head: `ecb6574fbec4d827f613d480592eb7eacb174603`.

## What changed

- Added one authoritative model-core policy: a static fallback entry is excluded when any provider mirror in that entry is disabled, case-insensitively.
- Preserved explicit user overrides and variant inheritance by filtering only automatic fallback traversal.
- Reused the core policy inside delegate-core; OpenCode adapters only thread readonly `disabledProviders` constraints.
- Wired category/direct delegation, actual `call_omo_agent`, builtin agents and cold-cache selection, `look_at`, Prometheus, model-fallback hooks/controller, tool registry, and config assembly.
- Centralized runtime-chain filtering in the model-fallback controller.
- Removed all unrelated CodeGraph churn.
- Regenerated only the three source-derived Senpi outputs: `omo.js`, `omo-task.js`, and `memory-run-supervisor.mjs`.

## Review regressions

- Model-core connected-provider and available-model mirror filtering, case-insensitivity, unchanged-reference behavior, explicit override/variant preservation, and first-allowed selection.
- Delegate-core mirrored-entry selection.
- Category-resolver boundary propagation.
- Real `createCoreTools` plus actual `createCallOmoAgent`, asserting the model passed to `BackgroundManager.launch`.
- Builtin agents, `look_at`, Prometheus, and explicit/static runtime fallback-chain boundaries.

## Automated verification

- Affected suite: **1,331 pass, 0 fail, 3,676 expectations**.
- Full root typecheck: **pass**.
- Full build: **pass**.
- Senpi extension freshness check: **pass**.
- `test:senpi`: **2,531 pass, 7 skip, 0 fail, 8,070 expectations**; evidence resolver **10 pass, 0 fail, 31 expectations**.
- Markdown path audit: **16 pass, 0 fail, 21 expectations**.
- Full root test: **16,996 pass, 39 skip, 1 fail**. The sole stale Codex installer version failure reproduces identically on pristine `origin/dev`; no Codex artifact is included here.
- All touched source files are at or below 250 pure LOC; no changed hunk introduced a no-excuse violation; `git diff --check` is clean.

## Real harness QA

- OpenCode 1.18.26, isolated HOME/XDG and identical empty project: baseline Explore/Librarian selected `openai/gpt-5.6-luna-fast` `low`; with only mirror `openai-codex` disabled, both selected `deepseek/deepseek-v4-flash` `max`. The tested plugin registered `call_omo_agent`, `task`, and `look_at`. Exact DB queries found zero real DB rows for QA sandbox paths.
- Real pinned Senpi engine 2026.9.2-4: candidate and pristine `origin/dev` produced identical lifecycle check matrices, `realSenpiUntouched: true`, unchanged real paths/digest, and zero leaked PIDs. Their identical six broad-driver failures are baseline harness/engine drift; product-scope spawn/wake/fanout/sync/negative-category/resume setup-cancel-kill-LRU/isolation checks passed.
- Tracked reviewer evidence: `.omo/evidence/20260826-issue6573-model-reqs-disabled-providers/`.

All eight review threads have been answered individually and resolved. The branch now needs fresh maintainer review and CI on `ecb6574fb`.

PR #7369, opened 2026-08-26
fix(continuation): retry transiently parked todo turns

### LINKED ISSUE #6517: [Bug]: Turn stays parked after PreToolUse hook_stopped_continuation until the user sends a message
## Prerequisites

- [x] I searched existing issues. Related but not matching: #4787, #4528, #5569 cover OpenCode `promptAsync` / background-task wake stalls. This one is the Claude Code `PreToolUse` capture path.
- [x] I am on the latest version.

## Summary

Under the Claude Code harness every tool call is captured by a `PreToolUse` hook:

```
hook_stopped_continuation:
  "This tool call is captured and executed by the host.
   Do not retry with other tools; end the turn."
```

The CLI turn genuinely ends, the host executes the tool, and a new session is started with the transcript replayed. That is the normal shape: 271 of 399 local sessions end this way.

Intermittently the session is not resumed promptly. Since no stream is live, the UI timer stops while still showing "working", so it is indistinguishable from a hang. Recovery is for the user to send any message, which forces a new invocation; the already-completed tool result is replayed and the answer appears at once.

## Evidence

Across 399 local session transcripts under `~/.claude/projects/`, handoff gap = previous session's last record -> next session's first record, for the 263 hook-terminated handoffs:

| gap | count |
|---|---|
| < 2s | 71 |
| 2-10s | 103 |
| 10-60s | 36 |
| 1-5m | 22 |
| > 5m | 31 |

Median 2.8s, so the common path is healthy; the tail is the problem.

Most long gaps are legitimately the user being away (the previous turn ended by asking them a question). The relevant cases are where the previous turn ended on `hook_stopped_continuation` - the agent was waiting for its own tool result, not for the user - and the next log entry is the user typing a complaint:

| gap | last tool before stall | user's next message (translated) |
|---|---|---|
| 3343s | Read | "so what's the current status?" |
| 1613s | Bash | "why is there no answer?" |
| 334s | Bash | "is it running right now?" |
| 126s | Bash | "what's the conclusion?" |
| 60s | Bash | "are you working?" |

All five previous turns ended as `hook_stopped_continuation` with the tool's result recorded, and no assistant record follows in that session file.

## What I could not determine

Local logs cannot distinguish "the host never re-invoked" from "it re-invoked but the UI did not render". No session file exists during the gap and the next file begins with the user's typed prompt, which favours the former, but I am not claiming it as proven.

Bash has a 120s default timeout, so a 27-minute gap after a Bash call is not explained by the command itself.

## Steps to Reproduce

No deterministic repro. It is intermittent across long sessions with frequent tool calls; roughly 3% of handoffs in my logs exceeded 5 minutes without the user being the cause.

## Environment

- oh-my-openagent `4.19.3` (commit `f2872273`)
- agent: `claude-code` `2.1.220`
- runtime: `bun 1.3.2`
- OS: Windows 11 (10.0.26200)

## Suggested Direction

A watchdog on the host side: after a `hook_stopped_continuation` turn ends and the tool result is available, if no re-invoke has been issued within N seconds, issue it. Alternatively surface the parked state in the UI so it is distinguishable from ongoing work.

Related: #6501 (detached eval completions never delivered) shows the same "parked with no live stream" failure mode from a different source.

AceRothstein71:
Fixes #6517

## Root cause
The todo continuation enforcer dropped transient prompt-gate declines such as `reserved` and `active`. Because the triggering idle event was already consumed, no future wake source remained and the turn stayed parked until user input.

## Changes
- Add a bounded three-attempt watchdog with five-second spacing
- Re-run the full guard chain and shared prompt gate on every retry
- Cancel and reset retry state on every existing stop, prune, and dispose path
- Preserve current behavior for permanent declines

## Validation
- Focused continuation suite: 154 pass, repeated twice
- omo-opencode tsgo: exit 0, repeated twice
- Isolated live OpenCode drive exercised idle, countdown, dispatch, continuation, and turn-boundary pause
- Evidence: `.omo/evidence/20260826-issue6517-parked-turn-continuation/`

## Residual risk
The exact live decline-to-retry timing branch was proven with real gate/state tests rather than forced against the nondeterministic live harness. A permanently stale busy state still exhausts the bounded budget.


packages/omo-opencode/src/hooks/todo-continuation-enforcer/continuation-injection.ts
```diff
@@ -257,6 +259,14 @@ ${todoList}`
       if (injectionState) {
         injectionState.inFlight = false
       }
+      if (isTransientGateDecline(promptResult.status)) {
+        scheduleGateRetry({
+          sessionID,
+          state: injectionState ?? sessionStateStore.getState(sessionID),
```
code-yeongyu, line 265: [sisyphus-bot] P1: This falls back to `getState()` only when `injectionState` was initially absent, but `injectionState` can become stale while `dispatchInternalPrompt()` awaits status/message I/O. If `session.deleted` or cleanup removes the session during that await, a transient result schedules the timer on the detached old state object; `cancelCountdown()` can no longer see or clear it. When it fires, the thunk calls `injectContinuation()` against the deleted session (and can even recreate state), so retries need to validate that the state is still current or retrieve the store state at scheduling/firing time.

packages/omo-opencode/src/hooks/todo-continuation-enforcer/gate-retry.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (attempt > MAX_GATE_RETRIES) {
+    log(`[${HOOK_NAME}] Gate retry budget exhausted`, { sessionID, reason })
+    return
+  }
+
+  state.gateRetryCount = attempt
+  state.gateRetryTimer = setTimeout(() => {
+    state.gateRetryTimer = undefined
+    log(`[${HOOK_NAME}] Gate retry firing`, { sessionID, reason, attempt })
+    void inject()
```
code-yeongyu, line 33: [sisyphus-bot] P1: Clearing this timer is not cancellation once the callback has begun: `void inject()` starts an async retry whose todo lookup, agent resolution, and gate checks can all yield. A user message, assistant event, stop, or session cleanup arriving during that window clears/reset state, but cannot stop the in-progress retry; if the session becomes idle before the gate's checks complete, it may still dispatch a continuation after the user has taken over. The retry needs a generation/cancellation token checked before dispatch, not only a cleared timer.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu, line 21: [sisyphus-bot] P1: This broad filter removes every tracked Markdown file under `.omo/` and `.opencode/` from the audit. It fixes the newly added evidence by disabling coverage for whole documentation trees, so future broken relative links or maintainer-local paths there will pass silently. Exclude only generated/non-product evidence files (or make the evidence links valid); do not turn off the audit for both directories wholesale.

packages/omo-opencode/src/hooks/todo-continuation-enforcer/gate-retry.test.ts
```diff
@@ -0,0 +1,353 @@
...
+    await injectContinuation({
+      ctx: ctx as never,
+      sessionID,
+      resolvedInfo: {
+        agent: "Sisyphus - ultraworker",
+        model: { providerID: "anthropic", modelID: "claude-sonnet-4-20250514" },
+      },
+      sessionStateStore,
+    })
+    await fakeTimers.advanceBy(GATE_RETRY_DELAY_MS * (MAX_GATE_RETRIES + 2), true)
```
code-yeongyu, line 287: [sisyphus-bot] nit: This test does not prove `MAX_GATE_RETRIES`. It advances time without asserting that retries were actually scheduled/fired, and `promptCalls.count === 0` plus no remaining timer also passes if the production fix schedules no retry at all. Assert the gate/status read count (initial attempt + exactly three retries), and assert no fifth attempt, so this test pins the bound rather than only the terminal state.

packages/omo-opencode/src/hooks/todo-continuation-enforcer/gate-retry.ts
code-yeongyu, line 33: [sisyphus-bot] P2: The timer callback intentionally launches `inject()` with `void` and no rejection handler. `injectContinuation()` catches expected fetch/dispatch failures, but any unexpected rejection from its pre-dispatch chain (or a future change that adds one) becomes an unhandled promise rejection from a background timer. Handle/log the returned promise here, especially because this watchdog is a new detached execution path.

code-yeongyu (changes requested):
[sisyphus-bot] Request changes. The retry mechanism addresses the reported lost wake, but its lifecycle integration is not safe under the event concurrency this hook already has: cleanup and cancellation can race with an async retry that has already started, and the new tests do not reliably pin the retry budget. The markdown test change also hides an entire pair of tracked documentation trees rather than auditing the new evidence selectively.

- The most serious runtime issue is that a retry can be scheduled on a detached state object after session cleanup, so cleanup can no longer cancel it.
- Cancellation is timer-only; once the callback fires, a user/assistant event can cancel the session while the retry continues through async pre-dispatch work.
- The budget test would pass if retry scheduling were completely absent, so it does not prove the bound.
- The markdown audit now silently skips every `.omo/` and `.opencode/` Markdown file, masking unrelated broken links.
- The retry callback discards a potentially rejected promise, creating an unhandled rejection path.

PR #7385, opened 2026-08-26
fix(omo-senpi): resolve the bundled ulw-loop cli

### LINKED ISSUE #6456: [Bug]: omo-senpi ulw-loop is disabled because the required omo CLI is not shipped
## Bug Description

The `omo-senpi` plugin registers an `ulw-loop` component, but the packaged Senpi plugin does not ship or install the `omo` / ulw-loop CLI required by that component.

As a result, a normal local-path installation of `packages/omo-senpi/plugin` logs:

```text
omo-senpi ulw-loop inactive; omo binary not found
```

The rest of OMO-Senpi, including `task` and `team`, works correctly. Only the durable ulw-loop status/continuation bridge is disabled.

## Steps to Reproduce

1. Install Senpi `2026.7.28-3`.
2. Build oh-my-openagent `4.19.3` at commit `efcfdd6363337b143937636c08d51fb544748cb2`.
3. Build the Senpi plugin:

   ```bash
   bun run build:senpi-plugin
   ```

4. Install `packages/omo-senpi/plugin` as a local Senpi package.
5. Run:

   ```bash
   senpi --help
   ```

6. Observe:

   ```text
   omo-senpi ulw-loop inactive; omo binary not found
   ```

## Expected Behavior

The built OMO-Senpi package should include, install, or resolve a compatible pinned ulw-loop CLI without requiring a Codex plugin installation.

`packages/omo-senpi` should be independently usable with Senpi.

## Actual Behavior

`packages/omo-senpi/src/components/ulw-loop/omo-command.ts` resolves only:

1. `OMO_BIN`
2. An executable named `omo` on PATH

If neither exists, `packages/omo-senpi/src/components/ulw-loop/index.ts` disables the component.

However, `packages/omo-senpi/plugin/package.json` ships only the extension, skills, runtime, and installer. It does not contain the required CLI.

The CLI implementation is currently located under:

```text
packages/omo-codex/plugin/components/ulw-loop
```

and is published/identified as:

```text
@code-yeongyu/codex-ulw-loop
```

This creates an unexpected packaging dependency from the Senpi adapter onto a Codex-owned component.

## Verified Workaround

I built only the pinned ulw-loop component and exposed it privately to Senpi:

```text
~/.senpi/bin/omo
```

The Senpi launcher sets:

```bash
OMO_BIN=/Users/jun/.senpi/bin/omo
```

This removes the warning and enables the Senpi ulw-loop bridge without installing or modifying the Codex plugin.

The component test suite passes:

```text
Test Files: 40 passed
Tests: 420 passed
```

## Additional Session-Isolation Issue

The ulw-loop session resolver reads:

```text
OMO_ULW_LOOP_SESSION_ID
CODEX_SESSION_ID
CODEX_THREAD_ID
```

Some component tests clear only `OMO_ULW_LOOP_SESSION_ID`. When the suite is executed inside Codex Desktop, the inherited `CODEX_THREAD_ID` changes evidence/session paths and causes two checkpoint-continuation tests to fail.

Running the suite with all three variables isolated produces 420/420 passing tests.

A Senpi-specific wrapper also needs to remove the Codex variables to prevent Senpi ulw-loop state from being stored under an unrelated Codex thread ID.

## Suggested Fix

1. Move the ulw-loop state/CLI implementation into a harness-neutral shared package.
2. Stage the compatible CLI inside `packages/omo-senpi/plugin/runtime`.
3. Make OMO-Senpi prefer its bundled CLI before checking `OMO_BIN` or PATH.
4. Add package-shape coverage proving a built Senpi plugin can run:

   ```bash
   omo ulw-loop status --json
   ```

   without a Codex installation.
5. Isolate all supported session environment variables in the ulw-loop tests.

## Environment

- macOS arm64
- Senpi: `2026.7.28-3`
- oh-my-openagent / OMO-Senpi: `4.19.3`
- oh-my-openagent commit: `efcfdd6363337b143937636c08d51fb544748cb2`
- Node.js: `24.18.0`
- Bun: `1.3.14`

## Doctor Output

N/A — this report concerns the OMO-Senpi adapter. The current doctor workflow targets OpenCode.

## OpenCode Version

N/A — no OpenCode runtime is involved in the reproduction.

AceRothstein71:
Fixes #6456

## Root cause
When the `ulw` binary was absent from PATH, the bundled ulw-loop CLI was never resolved, so start-work lanes failed with missing-command errors even though the CLI shipped inside the plugin.

## Changes
- Resolve the bundled staged CLI with precedence: explicit env override, then bundled CLI, then PATH, then OMO_BIN
- Strip inherited session keys from child environments so resumed lanes do not inherit stale session identity
- Cover senpi omo-command and codex paths plus new resolver tests

## Validation
- Focused omo-senpi and codex suites green, twice on identical tree
- Real-surface QA drove the bundled resolution path in isolation

## Residual risk
Windows-specific staging layouts were exercised through the shared resolver seam only.


code-yeongyu (changes requested):
[sisyphus-bot] Brutal verdict: REQUEST_CHANGES. This patch mixes the intended ulw-loop fix with a large generated-artifact churn and evidence dump, while leaving several failure and test-isolation holes. The bundled resolver can crash extension registration on filesystem errors; the injected-env API is not actually hermetic on Windows; runOmoCommand changes the environment of every caller rather than only toolkit probes; and the Vitest fixtures mutate process.env globally. The package-shape test also treats a build side effect as a prerequisite instead of making the package contract independently verifiable. Please fix the runtime correctness issues first, then reduce the unrelated generated diff and restore reliable, non-global test isolation.

- Bundled resolution must fail closed to PATH/OMO_BIN when probing the staged file fails for any filesystem reason, not only when URL parsing fails.
- The env-parameterized resolver must use the supplied environment consistently, including PATHEXT on Windows.
- Session-key scrubbing belongs at the toolkit boundary or in an explicitly named toolkit-only runner; the current exported generic command runner silently strips host variables from arbitrary child commands.
- The new test helper mutates shared process.env in beforeEach/afterEach, which is unsafe when Vitest executes files concurrently and can make unrelated tests observe a poisoned or cleared session.
- The markdown audit now excludes tracked evidence/config directories wholesale, weakening a safety test precisely for newly added markdown.
- nit: the same four session-key names are maintained independently in senpi production code, codex production code, and package-shape tests; export/share one source of truth instead of making future additions silently diverge.

Line-anchored findings (anchors rejected by API):
[sisyphus-bot] `statSync(..., { throwIfNoEntry: false })` is outside the try/catch. A permission-denied, malformed path, I/O, or other filesystem error from the bundled-runtime probe will now escape `resolveOmoBin()` during component registration and can abort Senpi extension activation. The intended fallback chain is only useful if *all* probe failures fall through; catch the stat failure and return null (or use a non-throwing probe).
- [sisyphus-bot] This resolver accepts an injected `env` object, but Windows executable discovery still reads `process.env.PATHEXT`. That makes the resolver non-deterministic and means a caller-provided environment is ignored for extension selection. On Windows, a custom `env` with a different/empty PATHEXT can resolve a different shim than the child environment will support. Pass PATHEXT through to `executableCandidates` (and test it).
- [sisyphus-bot] `runOmoCommand` is an exported generic spawn helper, but this change silently removes four session variables from every child it launches, regardless of whether `bin` is the toolkit. That is a behavior change for any current or future caller that deliberately needs CODEX/PI session context, and it makes the helper name/documentation lie. Either make this explicitly toolkit-only or pass an environment policy/options from the ulw-loop caller instead of globally scrubbing arbitrary commands.
- [sisyphus-bot] nit: exporting this constant from the Codex component does not establish the claimed shared source of truth: Senpi duplicates the exact four strings in `omo-command.ts`, and `package-shape.test.ts` duplicates them again. Adding a fifth host session key will silently make the adapters disagree. Put the list in a shared module/package and import it from both native paths and tests.
- [sisyphus-bot] This helper mutates the process-wide environment in `beforeEach` and restores it in `afterEach`. Vitest can run test files concurrently in the same worker/process, so another file can observe these four keys deleted or a test can restore a stale snapshot over a value established by another test. This is timing-sensitive global-state pollution, not isolation; use per-process subprocess envs or serialize an entire worker with a single suite-level guard.
- [sisyphus-bot] This test is not self-contained: it assumes `packages/omo-senpi/plugin/runtime/agent-toolkit/cli.js` was generated by a prior build step, then fails with a bespoke error if that side effect is absent. A package-shape test should validate the checked-in/package artifact under test or invoke the package build as its own setup; otherwise clean test invocations and consumers that run only this suite regress with a misleading failure.
- [sisyphus-bot] This change disables the markdown-link audit for every tracked file under `.omo/` and `.opencode/`, rather than excluding only known generated fixtures. That silently removes coverage for maintainer-local absolute paths and broken references in those directories, including the newly added tracked evidence. Do not weaken a repository-wide safety audit to accommodate generated evidence; make generated inputs valid or narrowly filter the exact fixture set.

KNN-07:
Closing as part of the #7417 backlog cleanup: issue #6456 is closed after the replacement fix merged in #7478. Thank you for the contribution.

PR #7390, opened 2026-08-26
fix(ulw-loop): queue refused swarm spawns durably

### LINKED ISSUE #6201: [Bug] Swarm workers silently dropped on 'agent thread limit reached'; plan/ledger not persisted, misrestored after compaction
## Environment

- OMO (codex plugin) 4.18.2, Codex CLI 0.144.4, macOS 15 (Darwin 24.6.0)
- Session ran under Codex multi-agent **v1** (`agent_max_threads` default = 6)

## Symptom (observed 2026-07-17, ULW research mode)

1. The lead planned **14 parallel explorer subagents** for a research swarm.
2. The first ~6 spawns succeeded, then batch spawns failed with `agent thread limit reached`.
3. The failed spawn calls returned **no worker IDs**, and OMO did **not queue** the remaining 8 axes. They were silently dropped - actual research coverage was 6/14 (43%), with no visible signal to the user that coverage shrank.
4. After auto-compaction, the session **misrestored swarm state as "parallel work was not started"** even though 6 workers had already completed. The lead then re-did axes serially, losing both the completed work attribution and wall-clock.

## Root cause (from transcript analysis)

The swarm plan and worker ledger appear to live **only in conversation context**:

- The spawn-refusal path drops pending items instead of enqueueing them (nothing marks them as "planned but not yet spawned").
- There is no on-disk ledger to rehydrate from, so post-compaction state reconstruction trusts the (lossy) summary.

## Suggested fixes

1. **Persist the plan**: write the planned axis list + worker ledger to the plugin data dir at plan time (a per-session JSON already exists at `data/omo-sisyphuslabs/sessions/<id>.json`), and update it on each spawn success/failure.
2. **Queue on refusal**: when spawn fails with `agent thread limit reached`, keep the remaining items as a pending wave and execute them as active workers complete, instead of dropping them.
3. **Rehydrate after compaction**: the existing `session_start` / `post_compact` hooks could reload swarm state from the disk ledger rather than trusting the compacted summary.
4. **Optional, Codex-native**: when `features.multi_agent_v2` + `enable_fanout` are available, consider delegating swarms to the native agent-job runner - it is sqlite-backed and automatically re-pends items on `AgentLimitReached` (`codex-rs/core/src/tools/handlers/agent_jobs.rs`), which provides durable queueing and post-compaction truth for free.

## Repro sketch

Any swarm larger than `agent_max_threads` on a v1 session (default 6): plan N > 6 parallel workers, observe spawns 7..N fail and vanish from the plan; trigger auto-compaction mid-swarm and observe state misrestoration.

Thanks for OMO - filed after tracing this end-to-end through session rollouts and the Codex source.

AceRothstein71:
Fixes #6201

## Root cause
The plugin only observed spawn requests (PreToolUse guard), never outcomes. When Codex refuses a spawn with `agent thread limit reached`, the refusal lived only in the lossy conversation response: no worker ledger, no durable queue, and after auto-compaction the lead trusted the summary and misrestored swarm state.

## Changes
- New per-session spawn ledger under `.omo/ulw-loop/<session>/`: `spawn-ledger.jsonl` records spawned workers with ids; `pending-spawns.jsonl` records every refused request with its full original input
- PostToolUse hook injects a drain directive when a refusal is detected (worker-id extraction first so success payloads mentioning limits cannot misfire)
- PostCompact hook re-injects both truths from disk so compaction can no longer erase the queue
- Wired via two new aggregate hook files registered in the plugin manifest (23 -> 25); plan-gated exactly like the existing spawn guard

## Validation
- ulw-loop vitest: 442 pass incl. 17 new cases, run twice over identical tree
- tsgo --noEmit: exit 0 twice; plugin node --test suite: 334 pass twice
- Real-surface QA drove the built hook CLI in an isolated sandbox: 6 successes ledgered silently, spawns 7-8 durably queued with drain directive, PostCompact re-injected counts; real ~/.omo, ~/.senpi, ~/.codex proven untouched
- Creator full-vertical audit: four waves, findings fixed with regression tests, clean_streak=2
- Evidence: `/tmp/opencode/issue-6201` (red log, gates-run1/2, QA transcripts)

## Residual risk
Refusal detection is pattern-based; unknown future wording stays fail-silent like today. Queue draining is lead-directed since hooks cannot execute tools. Ledger activates only for sessions with an active ulw-loop plan (the reported scenario).


packages/omo-codex/plugin/components/ulw-loop/src/spawn-ledger.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export function recordRefused(
+	cwd: string,
+	sessionId: string,
+	toolName: string,
+	reason: string,
+	input: SpawnLedgerEntryInput,
+): void {
+	appendLine(ulwLoopDir(cwd, { sessionId }), PENDING_SPAWNS_FILE, {
```
code-yeongyu, line 81: [sisyphus-bot] Each PostToolUse retry/re-delivery appends another pending record because `toolUseId` is stored but never used as an idempotency key. A timeout after the file append, Codex retry, or duplicate hook invocation therefore turns one refused request into N queue entries, and the rehydration count tells the lead to re-issue it N times. Deduplicate by session + toolUseId (or persist an explicit request state) and add a repeated-event test.

packages/omo-codex/plugin/components/ulw-loop/src/spawn-ledger.ts
```diff
@@ -0,0 +1,187 @@
...
+			`${pendingCount} spawn(s) were refused by the agent thread limit and remain queued at .omo/ulw-loop/${sessionId}/${PENDING_SPAWNS_FILE} - re-issue any you have not already re-issued`,
+		);
+	}
+	if (parts.length === 0) return "";
+	return `Durable swarm state read from disk (survives compaction): ${parts.join("; ")}.`;
+}
+
+function appendLine(stateDir: string, fileName: string, entry: Record<string, unknown>): void {
+	mkdirSync(stateDir, { recursive: true });
+	appendFileSync(join(stateDir, fileName), `${JSON.stringify(entry)}\n`);
```
code-yeongyu, line 126: [sisyphus-bot] This is not a durable write protocol: a process can be killed after `appendFileSync` partially writes a JSON line, and concurrent hook processes can leave a truncated/interleaved record. Because the reader below treats parse errors as an empty file, this can silently erase the queue/ledger from the post-compaction truth. Use an explicit locking/atomic-write strategy or a recovery-aware journal format and test interrupted/corrupt records.

packages/omo-codex/plugin/components/ulw-loop/src/spawn-ledger.ts
```diff
@@ -0,0 +1,187 @@
...
+		reason,
+		toolInput: input.toolInput,
+	});
+}
+
+export function hasActiveLoopPlan(cwd: string, sessionId: string): boolean {
+	try {
+		const raw = readFileSync(join(ulwLoopDir(cwd, { sessionId }), "goals.json"), "utf8");
+		const parsed: unknown = JSON.parse(raw);
+		return typeof parsed === "object" && parsed !== null;
```
code-yeongyu, line 96: [sisyphus-bot] `typeof parsed === "object"` is not an active-plan check: `{}`, `[]`, and any parseable stale/corrupt JSON activate the ledger. This makes ordinary `spawn_agent` calls write `.omo/ulw-loop/<session>/` state even when no usable ulw-loop plan exists, and it can mix unrelated sessions with stale files. Validate the plan shape and active lifecycle/status (as the existing `spawn-guard` does by reading `UlwLoopPlan`) rather than mere JSON object-ness.

packages/omo-codex/plugin/components/ulw-loop/src/spawn-ledger.ts
```diff
@@ -0,0 +1,187 @@
...
+	appendFileSync(join(stateDir, fileName), `${JSON.stringify(entry)}\n`);
+}
+
+function readJsonLines(path: string): Array<Record<string, unknown>> {
+	try {
+		if (!existsSync(path)) return [];
+		return readFileSync(path, "utf8")
+			.split("\n")
+			.filter((line) => line.trim().length > 0)
+			.map((line) => JSON.parse(line) as Record<string, unknown>);
```
code-yeongyu, line 135: [sisyphus-bot] Catching every `Error` and returning `[]` makes corruption indistinguishable from no durable state. In the claimed fix, a malformed `pending-spawns.jsonl` must fail visibly or preserve/recover valid preceding entries; otherwise PostCompact injects no queue directive and silently recreates issue #6201. At minimum report a degraded state and keep the valid prefix, with a regression test for one bad line after one good line.

packages/omo-codex/plugin/components/ulw-loop/src/spawn-ledger-hook.ts
```diff
@@ -0,0 +1,137 @@
...
+		toolUseId: typeof payload.tool_use_id === "string" ? payload.tool_use_id : "",
+		toolInput: payload.tool_input,
+	};
+	const outcome = classifySpawnResponse(payload.tool_response);
+	if (outcome.kind === "ignored") return "";
+	if (outcome.kind === "spawned") {
+		recordSpawned(payload.cwd, sessionId, payload.tool_name, outcome, input);
+		return "";
+	}
+	recordRefused(payload.cwd, sessionId, payload.tool_name, outcome.reason, input);
```
code-yeongyu, line 46: [sisyphus-bot] Every refusal emits a generic drain instruction, but the implementation never removes or marks a queued request when it is successfully re-issued. The queue is therefore permanently pending and every later PostCompact repeats the same count, while a retry that succeeds creates a worker without resolving the original record. Add request lifecycle events (queued/reissued/satisfied) keyed by toolUseId, or this is an append-only duplicate work list rather than a durable queue.

packages/omo-codex/plugin/components/ulw-loop/src/spawn-ledger.ts
```diff
@@ -0,0 +1,187 @@
...
+	const spawnedIds = readJsonLines(join(stateDir, SPAWN_LEDGER_FILE))
+		.map((line) => line["workerId"])
+		.filter((workerId): workerId is string => typeof workerId === "string");
+	if (spawnedIds.length > 0) {
+		parts.push(`${spawnedIds.length} worker(s) already spawned this session [${spawnedIds.join(", ")}]`);
+	}
+	const pendingCount = readJsonLines(join(stateDir, PENDING_SPAWNS_FILE)).length;
+	if (pendingCount > 0) {
+		parts.push(
+			`${pendingCount} spawn(s) were refused by the agent thread limit and remain queued at .omo/ulw-loop/${sessionId}/${PENDING_SPAWNS_FILE} - re-issue any you have not already re-issued`,
```
code-yeongyu, line 117: [sisyphus-bot] The context only reports a count and a path, not the queued task identities or original inputs. After compaction the lead cannot reliably know which specific refused requests were already re-issued, despite the directive saying "re-issue any you have not already re-issued"; that forces another manual file read and still cannot distinguish duplicates. Rehydrate stable request IDs/task summaries (bounded for context) and test multiple distinct refusals.

packages/omo-codex/plugin/components/ulw-loop/hooks/hooks.json
```diff
@@ -36,6 +36,31 @@
...
+						"command": "node \"${PLUGIN_ROOT}/dist/cli.js\" hook post-tool-use-spawn",
+						"timeout": 5,
+						"statusMessage": "(OmO 5.0.0-beta.21) Recording Swarm Spawn Outcomes"
+					}
+				]
+			}
+		],
+		"PostCompact": [
+			{
+				"hooks": [
```
code-yeongyu, line 54: [sisyphus-bot] nit: the new hook status messages use beta.21 while this component is beta.22 and all existing neighboring ulw-loop messages use beta.22. This is user-visible shipped metadata and will make diagnostics misleading; use the package/version source of truth or update both new messages to beta.22.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST_CHANGES: this implementation does not yet provide durable queue semantics under the failure modes it claims to fix. The append-only files have no idempotency or lifecycle model, malformed state is silently converted into an empty state, and the active-plan gate accepts arbitrary JSON objects, so the feature can both duplicate work and lose the very truth it is meant to preserve. The added tests cover only happy-path single-process calls and do not exercise retries, concurrent hooks, corrupted/truncated JSONL, or real hook payloads. The focused 17-test file passes and the component build passes; the broader plugin run still has an unrelated generated-skill fixture failure and typecheck cannot run without installing missing dependencies.

- Duplicate PostToolUse delivery/retry appends the same refusal repeatedly, inflating the queue and causing repeated re-issuance.
- JSONL parse/read failures are swallowed as empty state, so PostCompact can confidently rehydrate incomplete or corrupt data.
- `goals.json` being any non-null object is treated as an active plan, including `{}`/corrupt-but-parseable state, causing unrelated spawn sessions to write durable files.
- No tests pin deduplication, concurrent writers, corruption handling, actual Codex response shapes, or aggregate hook registration for the new hooks.
- nit: the new hook status messages say beta.21 while the component and neighboring hooks are beta.22.

MoerAI (changes requested):
Thanks for targeting the real #6201 failure mode. This head records repeated hook events as permanent pending work and treats corrupted durable state as empty state.

Please rebase onto current `dev`, then replace the append-only pending list with request state keyed by `(sessionId, toolUseId)`. Duplicate delivery must be idempotent, a successful reissue must resolve the original request, and PostCompact must show only unresolved requests with bounded IDs and task summaries. Serialize updates or use a recovery-aware journal, preserve valid records before a malformed line, and validate the persisted plan shape.

Add regression coverage for duplicate delivery, concurrent or interrupted writes, malformed JSONL with a valid prefix, reissue-to-resolved lifecycle, aggregate hook registration, and captured real Codex hook payloads.

PR #7401, opened 2026-08-26
fix(delegate-task): truncate sync results to max_bytes to prevent context saturation (Fixes #6164)

### LINKED ISSUE #6164: tool_output.max_bytes does not apply to sync delegate_task results — subagent outputs bypass truncation
## Problem

OMO's sync `delegate_task` path (`fetchSyncResult` in `packages/omo-opencode/src/tools/delegate-task/sync-result-fetcher.ts`) returns the subagent's full final assistant message (all text+reasoning parts concatenated) WITHOUT any size truncation. The OpenCode `tool_output.max_bytes` setting (default 51200) does NOT apply to the task tool's own output — it only governs plugin tool outputs via OpenCode's `registry.ts`.

## Evidence

Observed incident on 2026-07-16: An Atlas orchestrator session (glm-5.2) dispatched two sync `delegate_task` calls whose results measured **529KB and 535KB** respectively. These were injected directly into the parent model's context, causing:

1. **Context saturation** — glm-5.2 received a massive payload and produced an empty response (0 output tokens, finish_reason "stop").
2. **Session halt** — the agent loop exited normally (no tool calls) but no work was done.
3. **Pruning cycle** — when the agent tried to re-collect via `background_output`, OpenCode's `compaction.prune` immediately ejected the results (788KB across 6 items pruned in one pass), and OMO's incremental cursor returned "No new output" on re-collection (#2915).

Source confirmation: `sync-result-fetcher.ts` lines 57-75 and 144-163 return raw `textContent` without applying a size limiter.

## Proposed Fix

Add a configurable maximum on sync result size before returning to the parent:

```typescript
const MAX_SYNC_RESULT_BYTES = config.delegate_task?.max_sync_result_bytes ?? 8192;
if (textContent.length > MAX_SYNC_RESULT_BYTES) {
  textContent = textContent.substring(0, MAX_SYNC_RESULT_BYTES) +
    "\n\n[... truncated. Use background_output with full_session=true to retrieve the complete output.]";
}
```

Could be exposed as config: `delegate_task.max_sync_result_bytes` (default 8192).

This would be the proper fix because:
- It's a hard ceiling the model can't forget to apply
- It protects all agents, not just those with prompt-level directives
- The full transcript remains accessible via `background_output`

## Related
- #1734 — Background Task Output Distillation + Non-Destructive Recovery
- PR #5458 — do not compact while results are gathered
- PR #5204 — persist background tasks across restarts

AceRothstein71:
Fixes #6164

---
## Summary by cubic
Fixes #6164 — large delegated-task sync results now truncate to the configured `tool_output.max_bytes` limit (default 8192 bytes) instead of saturating the context window. Truncated results append a notice pointing to `background_output` with `full_session=true` to retrieve the complete output.

- Truncation respects multibyte UTF-8 character boundaries and never splits a character mid-sequence.
- A missing or invalid config value falls back to the default limit without failing the request.
- The markdown link audit now excludes generated `.omo/` and `.opencode/` files.

<sup>Written for commit 29d00d297221fc6fa095f3c4db123ce46c05c528. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/omo-opencode/src/tools/delegate-task/sync-result-fetcher.ts
```diff
@@ -2,10 +2,54 @@ import type { OpencodeClient } from "./types"
...
+  return typeof value === "object" && value !== null && !Array.isArray(value)
+}
+
+function readConfiguredMaxSyncResultBytes(configResult: unknown): number | undefined {
+  if (!isRecord(configResult) || !isRecord(configResult.data) || !isRecord(configResult.data.tool_output)) {
+    return undefined
+  }
+
+  const maxBytes = configResult.data.tool_output.max_bytes
+  return typeof maxBytes === "number" && Number.isSafeInteger(maxBytes) && maxBytes > 0 ? maxBytes : undefined
```
code-yeongyu, line 23: [sisyphus-bot] This parser accepts any positive safe integer, including values smaller than the truncation notice itself. For example, max_bytes=1 is treated as valid, but truncateSyncResult("x", 1) necessarily returns the ~90-byte notice and violates the configured byte limit. Either enforce the host's actual minimum here or define a behavior for caps smaller than the notice; add a boundary test so this cannot regress.

packages/omo-opencode/src/tools/delegate-task/sync-result-fetcher.ts
```diff
@@ -2,10 +2,54 @@ import type { OpencodeClient } from "./types"
...
+}
+
+async function resolveMaxSyncResultBytes(client: OpencodeClient): Promise<number> {
+  const config = client.config
+  if (!config || typeof config.get !== "function") return DEFAULT_MAX_SYNC_RESULT_BYTES
+
+  try {
+    const configResult = await config.get()
```
code-yeongyu, line 31: [sisyphus-bot] This adds an unconditional config RPC to every fetchSyncResult call, including the hot recovery path, even when the result is already below the limit. A slow or stuck /config request now stalls completion of an otherwise completed delegated task, and repeated sync continuations pay the same round trip. Resolve/cache this setting at task invocation (or use a bounded, non-blocking fallback) rather than making result delivery depend on a second endpoint.

packages/omo-opencode/src/tools/delegate-task/sync-result-fetcher.ts
```diff
@@ -2,10 +2,54 @@ import type { OpencodeClient } from "./types"
...
+    return readConfiguredMaxSyncResultBytes(configResult) ?? DEFAULT_MAX_SYNC_RESULT_BYTES
+  } catch (error) {
+    if (error instanceof Error) return DEFAULT_MAX_SYNC_RESULT_BYTES
```
code-yeongyu, line 34: [sisyphus-bot] The catch only defaults for values that pass instanceof Error; a rejected SDK/mock/runtime value such as a plain object or string is rethrown. That contradicts the stated fallback behavior for config failures and turns an optional size lookup into a new failure mode. Config reads should be best-effort for all rejection values, with the fallback applied unconditionally.

packages/omo-opencode/src/tools/delegate-task/sync-result-fetcher.ts
```diff
@@ -2,10 +2,54 @@ import type { OpencodeClient } from "./types"
...
+    throw error
+  }
+}
+
+function truncateSyncResult(textContent: string, maxBytes: number): string {
+  if (Buffer.byteLength(textContent, "utf8") <= maxBytes) return textContent
+
+  const noticeBytes = Buffer.byteLength(SYNC_RESULT_TRUNCATION_NOTICE, "utf8")
+  const contentMaxBytes = Math.max(0, maxBytes - noticeBytes)
```
code-yeongyu, line 43: [sisyphus-bot] Math.max(0, maxBytes - noticeBytes) prevents negative slicing but does not preserve the max_bytes contract when maxBytes < noticeBytes: the notice is still appended in full. This is the actual point where the returned string becomes oversized; handle an undersized limit explicitly (for example, clamp/validate before reaching this helper) and test it.

packages/omo-opencode/src/tools/delegate-task/sync-result-fetcher.ts
```diff
@@ -97,6 +141,8 @@ export async function fetchSyncResult(
     return { ok: false, error: `No assistant response found.\n\nSession ID: ${sessionID}` }
   }
 
+  const maxSyncResultBytes = await resolveMaxSyncResultBytes(client)
```
code-yeongyu, line 144: [sisyphus-bot] Config is fetched only after messages have been loaded and an assistant response exists, so even the no-truncation case now incurs an extra await. More importantly, this ordering makes result extraction's success dependent on config availability after the session API already succeeded. The size setting should be obtained once outside this per-result path or supplied by the caller, not resolved here for every branch.

packages/omo-opencode/src/shared/markdown-link-audit.test.ts
```diff
@@ -13,7 +13,13 @@ const MAINTAINER_LOCAL_PATH_RE = /file:\/\/\/(?:Users|home)\/|(?:^|[\s(`'"])(?:\
 function collectMarkdownFiles(): string[] {
   const output = Bun.spawnSync(["git", "ls-files", "*.md"], { cwd: WORKSPACE_ROOT, stdout: "pipe" })
   expect(output.exitCode).toBe(0)
-  return output.stdout.toString("utf-8").trim().split("\n").filter(Boolean).map((filePath) => resolve(WORKSPACE_ROOT, filePath))
+  return output.stdout
+    .toString("utf-8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((filePath) => !filePath.startsWith(".omo/") && !filePath.startsWith(".opencode/"))
```
code-yeongyu, line 21: [sisyphus-bot] This silently removes every markdown file under .omo/ and .opencode/ from both audits based only on its path. That can hide a genuine broken link or maintainer-local path in a checked-in document, and there is no test pinning that these trees are generated-only artifacts. Prefer excluding the specific generated files/artifacts (or explicitly assert they are untracked/generated) and add a test showing the exclusion does not mask a tracked markdown offender.

packages/omo-opencode/src/tools/delegate-task/sync-result-truncation.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        ],
+      }),
+      status: async () => ({}),
+    },
+  }
+}
+
+describe("sync result truncation", () => {
+  test("uses the configured OpenCode tool output byte limit", async () => {
+    //#given - a sync result larger than the host's configured output limit
```
code-yeongyu, line 39: [sisyphus-bot] This test suite does not pin the most failure-prone contract: no test uses a max_bytes smaller than the notice, invalid values, or a config rejection. The implementation currently accepts max_bytes=1 and can return a payload far above the requested cap, so the happy-path assertions would not catch the central boundary bug.

packages/omo-opencode/src/tools/delegate-task/sync-result-truncation.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    //#when
+    const result = await fetchSyncResult(client, "ses_test")
+
+    //#then - the default limit bounds the result and retains the recovery hint
+    const expectedPrefix = "x".repeat(maxBytes - byteLength(TRUNCATION_NOTICE))
+    expect(result).toEqual({ ok: true, textContent: `${expectedPrefix}${TRUNCATION_NOTICE}` })
+  })
+
+  test("does not split a multibyte character when truncating by bytes", async () => {
+    //#given - a multibyte sync result with only a partial character before the limit
```
code-yeongyu, line 67: [sisyphus-bot] The multibyte test only proves the particular emoji/3-byte budget case; it does not assert the general byte bound for a cut through each UTF-8 width, nor a case where contentMaxBytes lands exactly on a leading byte. Add 2-, 3-, and 4-byte boundary cases and assert byteLength(output) <= maxBytes for each, especially alongside the notice-underflow case.

code-yeongyu (changes requested):
[sisyphus-bot] I am requesting changes: the truncation path is not actually guaranteed to honor max_bytes and it adds a new runtime dependency to every sync-result fetch. The current tests cover only the happy path with large ASCII limits and therefore miss the boundary and failure behavior.

- The helper accepts any positive safe integer, but the notice is longer than many accepted values; the returned result can exceed the configured limit.
- Every fetch now awaits a separate config RPC, including recovery paths, and a hanging/unavailable config endpoint can delay or break otherwise available results.
- The markdown audit change excludes entire hidden trees without a regression test proving those files are generated and safe to ignore.
- Added tests do not exercise invalid/absent config, config rejection, small max_bytes, or the strict-abort/tagged branches that were modified.

MoerAI (changes requested):
Please rebase onto current `dev`, move max-byte resolution out of per-result fetching, and thread one best-effort validated cap through normal and continuation or recovery handoffs.

Caps smaller than the truncation notice must fall back to the documented default. Every truncated return must satisfy the actual UTF-8 byte ceiling. Add coverage for malformed or rejected config, under-notice caps, multi-byte boundaries, tagged deliverables, strict abort recovery, normal recovery, and exactly-once cap resolution.

Restore the Markdown audit and attach isolated real OpenCode evidence showing bounded sync output plus full-session recovery of the complete result.

PR #7455, opened 2026-08-28
fix(omo-senpi): stop an unspawnable formatter from killing the agent on Windows

sanguneo:
## Problem

On Windows, editing any file in a project with a `.prettierrc` (or any other detected formatter marker) kills the whole agent:

```
Error: spawn C:\...\node_modules\.bin\prettier ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:287:19)
    at onErrorNT (node:internal/child_process:508:16)
  errno: -4058,
  syscall: 'spawn C:\...\node_modules\.bin\prettier',
  spawnargs: [ '--write', 'C:\...\src\pages\...\index.vue' ]
```

The process dies mid-edit, so a multi-edit turn lands only partially and the rest of the work is lost.

Two defects combine in `packages/omo-senpi/src/components/formatter/formatter.ts`:

1. **The CLI fallback can never run on Windows.** `formatOne` resolves the tool through `resolveServerBinary`, whose `executableSuffixes` probes the bare name first, so on Windows it returns `node_modules/.bin/prettier` — the extensionless POSIX shim npm/bun write next to `prettier.CMD`. Windows cannot execute that file, so `spawn` always fails. Naming the `.CMD` directly is not a fix either: Node has refused a bare `.cmd`/`.bat` since the CVE-2024-27980 mitigation and throws `EINVAL` synchronously.

2. **A failed spawn crashes the process.** `childExit` only listened for `"exit"`. A child that never starts emits `"error"` and never `"exit"`, and with no `"error"` listener Node re-throws it as an `uncaughtException`. So a formatter that merely isn't launchable takes the agent down, instead of degrading through the `missing` path that already exists for exactly this case.

Measured on Node 24 / Windows 11, spawning the real `node_modules/.bin/prettier`:

| invocation | result | injection-safe |
|---|---|---|
| `.bin/prettier`, `shell: false` (current) | `ENOENT` (async, uncaught) | — |
| `.bin/prettier.CMD`, `shell: false` | `EINVAL` (sync throw) | — |
| `shell: true` | works | **no** — args are concatenated unescaped (DEP0190); a path containing `&` executes |
| `cmd.exe /d /s /c` + separate argv (this PR) | works | yes |

## Fix

Both changes are contained in the formatter component; no new dependency and no new import path.

- Build the child with `createSpawnCommand` from `@oh-my-opencode/lsp-core` (already imported in this file for `resolveServerBinary`). It resolves the real Windows launcher and routes `.cmd`/`.bat` shims through `ComSpec` as separate argv entries, so the file path is never reparsed as shell syntax. This is the same path the LSP layer has been spawning through, and it is already covered by the `createSpawnCommand` win32 tests in `packages/lsp-tools-mcp/test/process.test.ts`. Non-Windows behaviour is unchanged: `createSpawnCommand` returns the command as-is.
- Resolve a `SPAWN_FAILED` sentinel from `childExit` on `"error"`, and report that as `status: "missing"` so the existing "install it with bun add -d" notice fires (and `required` mode surfaces its error) instead of the process exiting. The synchronous throw Bun raises for the same condition is caught too.

`resolveServerBinary` itself is left alone: `createSpawnCommand` re-resolves the extension, so the LSP resolution order does not need to change for this bug.

## Tests

Two regression tests in `formatter.test.ts` drive `createFormatterStep` with a binary path that cannot be spawned, and assert the step degrades instead of throwing:

- best-effort → resolves, no error, warns once that the formatter is unavailable
- required → resolves with `Formatter prettier is unavailable`

Negative control: with only the `"error"` listener removed, both fail with the exact production error (`ENOENT`, `errno: -4058`), so the tests pin the fix rather than passing vacuously.

```
bun test packages/omo-senpi/src/components/formatter packages/omo-senpi/src/components/post-mutation
  8 pass, 0 fail

bun run --cwd packages/omo-senpi typecheck
  exit 0
```

## Notes

`plugin/extensions/omo.js` is regenerated in the follow-up commit. It was rebuilt on Linux: `digestBuildSources` hashes repo-relative paths, so a Windows build emits backslash-separated paths and the freshness gate can never match it (same reason the gate itself is pinned to ubuntu-latest).

Ultraworked with [omo](https://github.com/code-yeongyu/oh-my-openagent)

---
## Summary by cubic
Fixes a Windows crash where an unspawnable formatter (e.g. the extensionless `node_modules/.bin/prettier` shim) terminated the agent mid-edit. Formatting now spawns via `createSpawnCommand` and reports a failed spawn as a missing formatter, so best-effort warns and required mode errors instead of crashing.

- Adds regression tests for the unspawnable-binary path in best-effort and required modes.
- Tracks the spawn failure with a distinct sentinel so it can't be confused with a real exit code.
- Regenerates `omo.js` on Linux so the bundle-freshness gate accepts the source changes.

<sup>Written for commit 7ad84e9b139ab8ace5638180672b6280be2f1875. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
Reviewed against the actual `lsp-core` source and the production failure mode — the diagnosis and the fix are both correct:

- Verified `createSpawnCommand` (`packages/lsp-core/src/lsp/process.ts`): on win32 it probes `PATHEXT` candidates before the bare POSIX shim, so `prettier.CMD` wins, and `.cmd`/`.bat` shims route through `ComSpec` as `/d /s /c` + separate argv with `shell: false` everywhere — the injection-safety claim in the PR table holds. Non-win32 returns the command untouched, so POSIX behaviour is unchanged.
- The `childExit` `"error"` listener is the real crash fix: an unspawnable child emits `"error"` and never `"exit"`, and without a listener Node re-throws as `uncaughtException`. The regression tests drive this exact path (async ENOENT) on every OS, and the negative control in the PR description shows they pin the fix.

Two things I pushed directly to the branch (maintainer edit):

1. `9a2269c06` — typed the spawn-failure sentinel as a `"spawn-failed"` variant instead of `-1`. A Windows child exiting `0xFFFFFFFF` surfaces as `-1`, which would misclassify a real failed format run as a missing formatter and skip the pre-format content restore. A non-numeric variant makes the collision impossible.
2. Regenerated `plugin/extensions/omo.js` in the same commit so the bundle freshness gate (`senpi-compatibility`) stays green on top of your `718e92f38`.

With those, this is in the state I'd ship. Merging once CI settles.

PR #7457, opened 2026-08-28
fix(release): tolerate npm registry propagation in the platform gate

code-yeongyu:
## Problem

`publish-main`'s `Verify platform packages are published` gate probed each of the 24 platform packages (2 name families x 12 platform triples) **exactly once**. The npm registry read path lags a successful write, so a just-published version can 404 for a short while — and a single probe reads that lag as a missing package and aborts the release.

Release run [33146967411](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/33146967411) hit exactly this:

- all 12 platform **build** jobs: success
- all 12 platform **publish** jobs: success
- the gate: `Missing platform package(s); refusing to publish wrappers` naming `oh-my-opencode-windows-x64-baseline`, `oh-my-opencode-windows-arm64`, `oh-my-openagent-windows-arm64`
- all three returned `HTTP 200` when queried manually moments later
- re-running the failed job alone completed the release with **no code change**

## Root cause, measured

| probe | result |
|---|---|
| existing version | `HTTP 200`, `cache-control: max-age=300`, `cf-cache-status: DYNAMIC` |
| absent version | `HTTP 404`, **no** `cache-control`, **no** `cf-cache-status` |
| absent version + `Cache-Control: no-cache` request header | `HTTP 404`, identical |

This is **read-replica propagation lag after write, not CDN negative caching**. Cache-busting headers would be a placebo, so none were added.

This workflow already acknowledges the same hazard: `post-publish-verify` wraps its `omo-ai` registry check in `for attempt in $(seq 1 5)` + `sleep 15` and logs "registry propagation". This gate simply never received that treatment.

## Fix

Same house idiom, applied to the gate:

- re-probe **only the still-pending** packages each round, so a settled package is not re-fetched
- break as soon as the pending set empties
- after the deadline, fail with the **same hard error** listing whatever is still missing — a genuinely unpublished package still blocks the release
- a `curl`-level failure resolves to `000` and counts as **pending, not absent**, so a transient network blip neither aborts the step nor is mistaken for a missing package
- deadline 20 attempts x 15s (~5 min), overridable via `PROPAGATION_ATTEMPTS` / `PROPAGATION_DELAY`

The gate is not weakened: the only behavior change is *when* it concludes a package is absent.

## QA

Evidence: `.omo/evidence/20260828-publish-platform-gate-propagation/`

The harness **extracts the real script text from `publish.yml`** and runs it against a mocked `curl`/`sleep`, so it cannot pass against a stale copy of the logic. Run on macOS **bash 3.2** — deliberately stricter than the runner's bash 5.x for empty-array-under-`set -u` semantics.

| Case | Scenario | Expected | Result |
|---|---|---|---|
| A | all 24 visible immediately | exit 0, single round | PASS — 24 OK, no wait |
| B | 3 lag, appear after round 2 | exit 0 after retries | PASS — the production failure, healed |
| C | 1 never appears | exit 1, names it | PASS — same hard error |
| D | `curl` exits nonzero | pending, no crash | PASS |

`actionlint .github/workflows/publish.yml` → exit 0.

## Residual risk / out of scope

- No live re-publish was performed; that would burn a real version number. The lag condition is reproduced by mocking the probe, with run 33146967411 cited as the real occurrence.
- The sibling `Check if already published` steps in the same job stay single-shot **by design**: a stale read there causes a duplicate-publish attempt that npm rejects outright — a loud, safe failure — rather than a wrapper referencing a missing dependency. Changing publish idempotency is deliberately out of scope.

---
## Summary by cubic
Fixes the release gate in `publish.yml` so npm registry read-path lag after a successful publish no longer aborts a release. The gate previously probed each of the 24 platform packages (`oh-my-opencode` / `oh-my-openagent` × 12 platform triples) exactly once and treated a brief 404 after a successful publish as missing; it now re-probes only the still-pending packages on a bounded schedule before failing.

**Bug Fixes**

- Re-probes only the still-pending packages each round, up to 20 attempts × 15s (overridable via `PROPAGATION_ATTEMPTS` / `PROPAGATION_DELAY`), and stops as soon as the pending set empties.
- A curl-level failure (HTTP `000`) counts as pending, not as evidence of absence, so a transient network blip neither aborts nor fakes a missing package.
- A still-missing package fails after the deadline with the same hard error naming it, keeping the `Missing platform package(s); refusing to publish wrappers.` contract string verbatim so the existing workflow contract test still passes.
- Keeps the sibling `Check if already published` steps single-shot by design; a stale read there triggers a duplicate-publish attempt that npm rejects outright, which is a loud, safe failure.
- Adds evidence under `.omo/evidence/20260828-publish-platform-gate-propagation/`, including a harness that extracts and tests the real script text from `publish.yml` and root-cause header captures showing the lag is read-path propagation, not CDN negative caching.

<sup>Written for commit 2d9cb32955d44178c5072dc22c1e7997f88d8e3e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7457?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


AceRothstein71:
@code-yeongyu please handle the tons of green other PR's aswell

PR #7466, opened 2026-08-28
feat(mcp): add optional Parallel search provider

georgeatparallel:
## Summary

Adds Parallel to the existing `websearch.provider` setting for web search and page extraction without a Parallel account or API key. Exa stays the default, Tavily keeps its current key handling, and existing MCP overrides and permissions still apply.

I work at Parallel, which operates this service.

## Changes

- Adds an explicit `"parallel"` selection using OpenCode's existing remote MCP client and `https://search.parallel.ai/mcp`. No new dependency, transport, fallback, or credential handling.
- Documents setup in `omo.jsonc`, disabling, rate limits, and data sent to Parallel. Once enabled, agents may send queries, URLs, and supplied context through the tools.
- Makes the Librarian's two search references use the configured provider's advertised tool schema, and recognizes Parallel tool names in agent reminders, JSON error recovery exclusions, and built-in tool guidance.
- Regenerates both configuration schemas. The unified schema also picks up existing `formatOnMutation` fields that were absent from the checked-in artifact. Regenerating the unmodified base reproduces those same 206 lines; the only additional schema changes are the Parallel enum entries.

## QA & Evidence

- **Real OpenCode 1.18.25 in Docker:** loaded the built plugin and documented configuration, connected anonymously, and exposed `websearch_web_search` and `websearch_web_fetch`. Search returned 10 results; fetch returned one page with no per-URL errors. A local response fixture drove tool selection; the MCP calls and results were live. The initial container had no host mounts; the follow-up containers mounted only task-owned code and evidence. All used isolated configuration and state with no real API credentials.
- **Configuration and runtime counterexamples:** 12 independent config cases pass, including project/profile precedence and credential isolation. Four native OpenCode cases verify `disabled_mcps`, `enabled:false`, custom URL precedence without fallback, and permission denial.
- **Regression checks:** 98 focused tests pass, covering provider behavior, credential isolation, configuration merging, schema freshness, and the helper paths that consume websearch tool names. New behavior was reproduced as failing before the corresponding fixes. The updated OpenCode suite passed with 8,434 tests and one skip; typecheck passed.
- Sanitized evidence is recorded locally under `.omo/evidence/20260828-parallel-search/`, including `live-qa.log`, `loader-qa.log`, `scoped-final.log`, `typecheck-final.log`, `build.log`, `root-tests-bun14.log`, and the baseline schema comparison.

## Risks & Residuals

- This is an explicitly selected anonymous service with rate limits. Paid authentication, quota exhaustion, and model-driven search quality were not tested.
- Before the tool-name follow-up, the full root suite on Bun 1.4.0 passed with 16,812 tests, 32 skips, and no failures across 2,163 files. Full build also passed. After the follow-up, the plugin was rebuilt and real search/fetch plus the reminder hook were reverified in OpenCode. The initial run on installed Bun 1.3.14 was interrupted after an unsupported binary-asset test; the completed run uses the required version.
- No Codex or Senpi runtime source was changed; unrelated generated distribution bundles are excluded.

## Automated Checks

```bash
bun test packages/omo-opencode/src/mcp packages/omo-opencode/src/plugin-handlers/mcp-config-handler.test.ts tests/omo-schema-freshness.test.ts
bun run build:schema
bun run typecheck
bun run build
bun test
git diff --check
```

### DIFF
--- a/assets/oh-my-opencode.schema.json
+++ b/assets/oh-my-opencode.schema.json
@@ -9775,7 +9775,8 @@
           "type": "string",
           "enum": [
             "exa",
-            "tavily"
+            "tavily",
+            "parallel"
           ]
         }
       },
--- a/assets/omo.schema.json
+++ b/assets/omo.schema.json
@@ -5,6 +5,47 @@
   "description": "Configuration schema for the omo.json / omo.jsonc harness-neutral config surface",
   "type": "object",
   "properties": {
+    "formatOnMutation": {
+      "type": "object",
+      "properties": {
+        "mode": {
+          "default": "best-effort",
+          "type": "string",
+          "enum": [
+            "off",
+            "best-effort",
+            "required"
+          ]
+        },
+        "languages": {
+          "type": "object",
+          "propertyNames": {
+            "type": "string"
+          },
+          "additionalProperties": {
+            "type": "boolean"
+          }
+        },
+        "maxFileBytes": {
+          "default": 1048576,
+          "type": "integer",
+          "exclusiveMinimum": 0,
+          "maximum": 9007199254740991
+        },
+        "timeoutMs": {
+          "default": 3000,
+          "type": "integer",
+          "exclusiveMinimum": 0,
+          "maximum": 9007199254740991
+        }
+      },
+      "required": [
+        "mode",
+        "maxFileBytes",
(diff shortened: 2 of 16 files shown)


georgeatparallel:

PR #7489, opened 2026-08-29
fix(omo-native): restore npm-launcher parity in the compiled entry

Keonho-Chu:
## Summary

The standalone binary entry (`packages/omo-native/compile-entry.ts`, added in d3dc741e5 as "compiled-binary entry with launcher parity") diverged from the npm launcher on three commands, and a fourth defect surfaced once the second one was fixed. In a `bun build --compile` binary, `omo ulw-loop` re-entered omo itself, `omo setup` skipped the credential import entirely, `omo doctor --reap <pid>` was silently ignored, and the setup plan crashed on a sidecar JSON the bundle cannot serve. After this PR the compiled binary answers these commands exactly like `bin/omo.js` does.

Correction from the first revision of this description: compiled binaries ARE a live user surface - the beta.26, beta.28-31 releases each carry 13 binary assets (the earlier "never shipped" claim was stale; it came from #7358's beta.18-21 window). That makes these dispatch defects reachable by real users today.

## Changes

- **`omo ulw-loop` (compile-entry.ts, `runCompiledLauncher`)**: under `bun build --compile`, `process.execPath` is the omo binary, so `spawn(process.execPath, [cli.js, ...])` started a second omo with the script path as engine argv. The command now runs through the staged `plugin/runtime/agent-toolkit/omo-agent-toolkit` shim, the same artifact the engine already invokes via `OMO_AGENT_TOOLKIT_BIN` (`.cmd` through `cmd.exe` on Windows, mirroring the engine's `toSpawnTarget`). It is awaited through `runChild` and its status propagated with `propagateResult`, instead of the previous fire-and-forget spawn that always exited 0. New export `toolkitSpawnTarget()`; `runCompiledLauncher` gains an `options` seam for tests.
- **`omo setup`**: dispatches to `runSetup(args.slice(1))` (report, model hints, plan, consent, `auth.json` write) instead of `printSetupReport` only, so `--yes` and `--dry-run` work as in the npm launcher.
- **`omo doctor --reap`**: handled first through `reapStaleEngines` (no credential-store scan needed); other doctor arguments are forwarded to `runDoctor`. The reap refusals and pid checks are unchanged.
- **`bin/lib/setup-import.js`**: `provider-map.json` is imported statically (`with { type: "json" }`) so the bundler inlines it. Reading it through the module URL resolved to `/$bunfs/root/provider-map.json` inside the binary and threw ENOENT. The npm path is unaffected (Node >= 24 already required).
- **Tests** (`compile-entry.test.ts`, given/when/then): shim target and never-`process.execPath`; child awaited and exit status propagated; Windows `.cmd` via `cmd.exe`; setup dispatch with `--yes` / `--dry-run`; `--reap` signals a stale pid and refuses a live one; `setup-import.js` resolves no sibling file through the module URL.

## QA & Evidence

Current head (rebased onto dev b5cbae3fb, 2026-09-01). Evidence lives in the task worktree at `.omo/evidence/20260901-compiled-entry-parity-rebase/` (gitignored per repo convention); key results inlined here.

- **What was tested:** `qa-driver.sh` built the darwin-arm64 binary twice from the same staged payload (omo-ai 5.0.0-0.beta.31) with `script/build-omo-binary.ts` under bun 1.4.0 (BEFORE = upstream/dev b5cbae3fb sources, AFTER = this head) and ran each command under a throwaway `HOME`/`USERPROFILE`, so self-provisioning into `~/.omo/binary-runtime/<version>/` and every state write stayed in the sandbox.
- **Observed result:**

  | command | BEFORE (upstream/dev binary) | AFTER (this branch binary) |
  |---|---|---|
  | `omo ulw-loop --help` | prints the **senpi engine** help (`OmO - AI coding assistant ...`): omo re-entered itself | prints the agent-toolkit `ulw-loop` usage, exit 0 |
  | `omo setup --dry-run` | HARNESS table only, `--dry-run` ignored, exit 0 | HARNESS table + MODEL AVAILABILITY + `DRY RUN: no files will be written` + planned/skipped lines, exit 0 |
  | `omo doctor --reap 1` | plain doctor output, exit 0, `--reap` dropped | `FAIL refusing pid 1: it is not an omo engine process`, exit 1 |

  Files under the real `~/.omo` modified during the run: 0 before, 0 after.
- **Why sufficient:** each defect is a dispatch decision in `runCompiledLauncher`; the unit tests pin the decision through injected seams, and the compiled binaries prove the decisions survive `bun build --compile` on the shipped artifact. The first AFTER run of `setup --dry-run` is what exposed the `provider-map.json` ENOENT (`/$bunfs/root/provider-map.json`), a failure only the real binary can show; the table above is the re-run after that fix.
- **Automated gates (current head):** `bun test packages/omo-native/test` on bun 1.4.0 (CI toolchain): 229 pass / 0 fail. `bun run typecheck` (root) green. `bun run build` and `bun run build:omo-native` green. On a local bun 1.3.9 the only two failures are the pre-existing `node:sqlite` resolution errors in `setup-detect.test.ts` / `setup-import.test.ts`, identical on unchanged dev.
- **Review changes applied:** rebased onto current dev (conflicts with the compile-runtime split, fast-path, and cursor OAuth changes resolved by re-applying each fix to the new structure; all newer provisioning/OAuth behavior preserved untouched). The awaited-child test no longer uses a timer: the fake child is a promise the test resolves explicitly, and a microtask drain proves the launcher stays unsettled until the child exits.

## Risks & Residuals

- The toolkit shim executes `node` from `PATH`; that is already the contract for every engine-initiated toolkit call in the compiled runtime (`OMO_AGENT_TOOLKIT_BIN` points at the same shim), so `omo ulw-loop` now shares one dependency instead of a broken one. Mitigated.
- Windows `.cmd` routing is pinned by a unit test only. A real Windows compiled-binary smoke cannot be produced from this darwin contributor machine; the `.cmd`-through-`cmd.exe` target mirrors the engine's own `toSpawnTarget` contract for the identical shim. If a maintainer-side Windows run is required before merge, the release lane's windows-x64 smoke covers the same entry.
- Out of scope, noted for a follow-up: the binary-runtime path uses `os.homedir()` while `agent-dir.js` resolves `runtimeHome(env)`; unrelated to command dispatch and left untouched here.

## Related Issues

Context: #7358 (release binary lane has never uploaded assets). No issue exists for these defects.


MoerAI (changes requested):
Thanks. The three dispatch corrections match the npm launcher's intended behavior, but this head cannot be merged or validated yet.

Please rebase onto current `dev` and resolve the compiled-entry conflicts while preserving the newer provisioning and OAuth changes. Replace the `setTimeout(20)` assertion with deterministic child-completion synchronization.

Please attach reviewable current-head evidence for native tests, typecheck, package build, and an isolated compiled-binary run covering `ulw-loop --help`, `setup --dry-run`, and `doctor --reap 1`. Include a real Windows compiled-binary smoke for the `.cmd` shim path. Also correct the claim that compiled binaries had not shipped; they were already a live user surface.

Keonho-Chu:
Thank you for the review - all points addressed at the new head (`1eaec0649`), point by point:

1. **Rebase / conflicts**: rebuilt all four commits on current `dev` (`b5cbae3fb`). The compile-runtime split, the provision-before-fast-path ordering, and the bundled cursor OAuth registration are preserved untouched; each fix was re-applied to the new structure rather than merged over it. The PR is MERGEABLE again.
2. **Deterministic test**: the `setTimeout(20)` assertion is gone. The fake child is now a promise the test resolves explicitly; a microtask drain proves the launcher stays unsettled while the child runs, then the test resolves the child with status 3 and asserts the propagated exit code. No timers involved.
3. **Current-head evidence** (details and artifact paths in the updated PR description): `bun test packages/omo-native/test` on bun 1.4.0 = 229 pass / 0 fail; root `bun run typecheck`, `bun run build`, `bun run build:omo-native` all green; darwin-arm64 binaries rebuilt from this head (BEFORE = dev `b5cbae3fb`, AFTER = head, same staged 5.0.0-0.beta.31 payload) re-confirm all three behaviors in isolated HOMEs: `ulw-loop --help` prints the toolkit usage instead of the engine help, `setup --dry-run` runs the full consent-gated plan, `doctor --reap 1` refuses with exit 1. Real `~/.omo` untouched (0 files modified).
4. **Windows smoke**: I cannot produce a real Windows compiled-binary run from this darwin contributor machine, and I would rather say that plainly than fabricate evidence. The `.cmd` path is pinned by a unit test (`cmd.exe /d /s /c <shim>.cmd`, mirroring the engine's own `toSpawnTarget` contract for the identical shim). If a maintainer-side Windows run is required before merge, the release lane's windows-x64 smoke exercises the same entry; I'm happy to iterate on any failure it surfaces.
5. **Shipped-binaries claim**: corrected in the PR description. You are right - beta.26 and beta.28-31 each carry 13 release assets, so the compiled entry is a live user surface; my original claim was stale (it came from #7358's beta.18-21 window). That makes these defects user-reachable, which strengthens the case for the fix.

MoerAI (approved):
Approved on current head. The requested source and test changes are resolved:

- the four fixes are rebuilt on current `dev` without discarding newer provisioning or OAuth behavior;
- `ulw-loop` uses the staged toolkit shim, awaits the child, and propagates its exit status;
- setup uses the full consent-gated import path and statically bundles `provider-map.json`;
- doctor forwards arguments and handles `--reap` before the credential scan;
- the child-completion test is event-driven and contains no timing delay;
- the PR now describes compiled binaries as a shipped user surface;
- the isolated darwin compiled-binary before/after results cover all three commands.

Merge remains deferred for two external gates:

1. The required current CI workflow is `action_required` and has produced no protected matrix:
   https://github.com/code-yeongyu/oh-my-openagent/actions/runs/33452396008

2. The requested real Windows compiled-binary smoke has not run. The unit test correctly constructs the `.cmd` target through `cmd.exe`, but that does not prove Windows process quoting and staged-shim execution in the shipped binary. Run the Windows release/native-binary smoke and attach its exact output before merge.

No further source redesign is requested. Do not merge until the current matrix and the real Windows compiled surface are green.

PR #7496, opened 2026-08-30
hashline-edit: clarify pos-only replace semantics + duplicate-definition guard

### LINKED ISSUE #7495: edit tool: silent duplicate class/module corruption on pos-only multi-line replace
### Summary

The hashline edit tool (`edit` with `LINE#ID` anchors) can silently produce a file with a **duplicated class/module definition** when a multi-line replacement is anchored with `pos` but no `end`. The docs state "replace with pos only -> replace one line at pos", and the tool does exactly that — but the resulting corruption is invisible at edit time and only surfaces later as runtime/test failures. Three occurrences in one working session today (two by an autonomous sub-agent, one by the orchestrator), each caught only downstream.

### Reproduction

1. File `trusted_proxy.rb` (27 lines) defines `class TrustedProxy` with a `valid?` method using exact-match source checks.
2. Agent intends a whole-file rewrite and calls:

```
edit(file, { op: "replace", pos: "1#YJ", lines: [<56-line complete replacement file>] })
```

(omitting `end` while supplying a full-file replacement — an easy misread of the `pos`-only semantics).

3. Tool reports success. Result on disk: lines 1–56 = new implementation, **lines 57–83 = the entire original file appended after it**. Ruby re-opens the class; the OLD `valid?` silently overwrites the NEW one.

Observed downstream symptom: tests asserting new CIDR behavior failed with results inverted from expectations (old exact-match `include?` was winning), and one test expecting `false` received `true` — a genuinely confusing debugging session until the file was re-read in full.

### Same failure shape seen twice more (same session, different agent)

- `src/zammad_ai_sidecar/ingestion/vector_import.py`: duplicated restore-barrier block left in the file after an edit pass.
- `src/zammad_ai_sidecar/reference/dispatch.py`: duplicated `return ingest` statement.

All three were silent successes; none produced an edit-tool error.

### Proposed guardrails

1. **Post-edit duplicate-definition heuristic**: after a `replace`, scan the result for top-level duplicate `class X` / `module X` definitions (same file, same name, non-trivial bodies) or a tail that duplicates the replaced range's original content, and emit a loud warning (or fail) instead of a clean success.
2. **File-complete replacement warning**: if the inserted block looks file-complete (e.g., starts with `# frozen_string_literal:` / shebang) but the file still contains trailing content beyond the replaced range, warn.
3. **Docs**: make the `pos`-only + multi-line `lines` semantics unmissable ("inserts lines *at* pos; everything after pos is preserved"), and consider a `wholeFile: true` mode for whole-file rewrites so agents don't hand-roll them via `pos`-only.

### Why it matters

The tool's value proposition is edit reliability. Today's three incidents were all *silent* — the worst failure mode for an agent that trusts its own tool results and moves on. A cheap post-edit lint would have caught all three at edit time.

Jrjy3:
## Summary

- Makes the hashline-edit `replace` semantics explicit: a `pos`-only call swaps **one** line and preserves everything after `pos`, so a whole-file rewrite **must** pass `end`. Adds a BAD example mirroring a real incident where a pos-only full-file insert left the original file appended after the inserted block - the old class definition silently overrode the replacement.
- Adds a post-edit guard that warns (in the tool result) when the edited file now contains a top-level definition name more times than the pre-edit content did - catching the duplication at the moment of damage instead of three test failures later.

## Changes

- `packages/omo-opencode/src/tools/hashline-edit/tool-description.ts`: OPERATION CHOICE bullets rewritten ("replace ONE line at pos ... NEVER consumes lines beyond pos"; "Pass `end` whenever the replacement should consume the rest of the region"), plus a new BAD example for whole-file rewrites via pos-only.
- `packages/omo-opencode/src/tools/hashline-edit/duplicate-definition-guard.ts` (new): `detectDuplicateDefinitions(old, new)` + result-string formatter. Top-level definitions only (column-0 `class|module|def|function|interface|struct|enum|trait|namespace`), language-agnostic, fail-safe.
- `packages/omo-opencode/src/tools/hashline-edit/hashline-edit-executor.ts`: computes the warning after `applyHashlineEditsWithReport` and appends it to both success return paths (formatted and non-formatted).
- `packages/omo-opencode/src/tools/hashline-edit/duplicate-definition-guard.test.ts` (new): 7 unit tests.

## QA & Evidence

- **What was tested:** A/B experiment with 18 agent runs (3 rounds x 6 agents; per round 3 agents received the OLD description, 3 the NEW one) performing a whole-file rewrite of a seeded Python file through a sandbox implementing the documented pos/pos+end semantics. Round 3 enforced a single edit call per agent with mechanical op journals.
  **Observed result:** 18/18 correct final files; no agent ever attempted the pos-only full-file insert (round-3 journals: OLD-arm agents used pos-only single-line + ranges or one region rewrite; NEW-arm agents used one whole-file pos+end rewrite). The clarified wording is therefore zero-risk and acts as insurance for attention-failure cases; the guard covers the residual (the motivating incident was a real corruption where the old class definition silently overrode the new one and three test failures surfaced it downstream).
  **Artifact:** journals and scored files reproduced locally; unit tests in `duplicate-definition-guard.test.ts`.
  **Why sufficient:** the guard is unit-tested against the exact incident shape (class duplicated 2x, pre-existing duplicate increased, single definitions not flagged, indented definitions ignored, fail-closed formatting).

## Risks & Residuals

- The definition regex is deliberately simple (column-0 keywords); exotic languages or non-top-level duplicates are out of scope. Consequence: warning-only, never blocks the write. Accepted.
- False-positive surface: legitimately re-opening a class/module now produces a warning. Frequency is low and the warning is advisory. Accepted.
- Repo-level `bun run typecheck` / full `bun test` could not run locally (workspace `prepare` build script fails on a clean clone); the new module is dependency-free and its standalone test suite passes 7/7. Flagged as residual for CI.

## Automated Checks

```bash
bun test ./duplicate-definition-guard.test.ts   # 7 pass, 0 fail (standalone)
```

## Related Issues

Closes #7495

---
## Summary by cubic
Clarifies `hashline-edit` pos-only `replace` semantics and adds a post-edit guard that detects duplicate top-level definitions. A pos-only whole-file insert previously left the original file appended after the inserted block, so the old definition silently overrode the replacement.

**Bug Fixes**
- pos-only `replace` now swaps ONE line and preserves everything after `pos`; whole-file rewrites must pass `end` so the old content is consumed.
- Added a BAD example to the tool description mirroring the incident case.

**New Features**
- A post-edit guard warns in the tool result when a top-level definition appears more times after an edit than before, and only top-level column-0 definitions are considered.
- The warning is advisory and never blocks the write; legitimately re-opening a class or module now also produces one.

`install-local.mjs` is a regenerated bundle with no behavior change. Closes #7495.

<sup>Written for commit b90cd0fb727e90c65e832c82ca25960d82a0f64f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7496?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-codex/scripts/install-dist/install-local.mjs
+++ b/packages/omo-codex/scripts/install-dist/install-local.mjs
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-// omo-codex-install:819e258f7485f9ccdcbb64a54ac0097bf81cd8e2980d7f907017f224b1f99b98:3a0f0cc479b76911cc517d857bad53dceaaf7421ac945a5476ac1a5112c0bcf5
+// omo-codex-install:819e258f7485f9ccdcbb64a54ac0097bf81cd8e2980d7f907017f224b1f99b98:20913738301914f89b9289f00d4112e98591dfe4483837153ce39cec0eabbd2f
 var __defProp = Object.defineProperty;
 var __returnValue = (v) => v;
 function __exportSetter(name, newValue) {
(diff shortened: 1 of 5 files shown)


Jrjy3:

PR #7518, opened 2026-08-30
fix(senpi): recover Claude SDK single-exchange overflow

### LINKED ISSUE #7497: [Bug]: Anthropic/Fable exceeds 1M after OMO system/tool injection and cannot compact single-exchange sessions
### Prerequisites

- [x] I searched existing issues to avoid duplicates
- [x] I am using a current OMO beta build
- [x] I reviewed the docs and related issues

### Bug Description

In a long-running OMO/Senpi thread, Anthropic-family **Fable 5** repeatedly becomes unrecoverable with `Prompt is too long`, while an OpenAI GPT model can continue the **same thread** successfully.

The most useful error explains that the conversation itself is about **535,486 tokens** (below the 1M limit), but the final provider request is about **1,166,728 tokens** after adding the system prompt, tool definitions, and attachment content:

```text
Prompt is too long · the request is ~1166728 tokens (limit 1000000) but this conversation is only ~535486 tokens — the rest is system prompt, tool definitions, and attachment content. A single-exchange conversation cannot be compacted; reduce attached files/tools or start with less context.
```

After that, even a one-word follow-up such as `continue` / `keep going` returns only:

```text
Prompt is too long
```

Switching the model/provider to GPT in the same OMO thread resumes normally. This has reproduced repeatedly with Fable/Anthropic and has not reproduced with the GPT-family model used in the same session.

The model-specific difference suggests the provider-aware request budget or compaction path is not accounting for the complete payload that OMO/Senpi injects (system instructions + tool schemas + attachment/context blocks) before sending the Anthropic request.

### Steps to Reproduce

1. Start a durable OMO/Senpi coding session with a large repository instruction file and the normal tool set enabled.
2. Continue a multi-stage goal for many turns until OMO compacts the history into a large synthetic summary/single exchange.
3. Use an Anthropic-family Fable model.
4. Send a short continuation message.
5. Observe the 1M context rejection above and the `single-exchange conversation cannot be compacted` recovery failure.
6. Send another very short message; the same error repeats.
7. Switch the same thread to a GPT/OpenAI model; the thread continues successfully.

### Expected Behavior

Before dispatching to Anthropic, OMO should budget against the **complete provider request**, not only conversation tokens.

If the final payload would exceed the model limit, OMO should recover without requiring a new thread, for example by one or more of:

- compacting before converting history into an indivisible synthetic single exchange;
- recursively summarizing an oversized synthetic summary;
- trimming/distilling injected tool schemas or attachment blocks that are not needed for the next turn;
- using provider/model-aware context limits and reserved system/tool overhead;
- providing a provider fallback without losing the active goal/thread state.

At minimum, a short follow-up should not enter a permanent error loop when the conversation token count itself is below the model limit.

### Actual Behavior

- OMO reports request ~1,166,728 tokens against a 1,000,000-token limit.
- It reports the conversation as only ~535,486 tokens and attributes the remainder to system prompt, tools, and attachments.
- Recovery refuses to compact because the session is represented as a single exchange.
- Every subsequent Fable turn fails immediately, even for a one-word message.
- The same thread remains usable after switching to GPT.

### Doctor Output

```shell
 oMoMoMoMo Doctor

 ⚠ 1 issue found:

1. Invalid configuration
   ../../../.omo/omo.jsonc: Invalid omo config at /home/cha/.omo/omo.jsonc: task
   Affects: plugin startup
```

The invalid `task` config warning may be relevant and is included for completeness. I am not claiming it is unrelated. However, the same loaded thread/config works when switched to GPT, so it does not explain the provider-specific behavior by itself.

### Error Logs

```text
Prompt is too long · the request is ~1166728 tokens (limit 1000000) but this conversation is only ~535486 tokens — the rest is system prompt, tool definitions, and attachment content. A single-exchange conversation cannot be compacted; reduce attached files/tools or start with less context.
```

### Environment

- OS: Linux
- OMO: `omo 5.0.0-0.beta.26`
- Engine: `senpi 2026.8.28-2`
- OpenCode: `1.18.21`
- Failing provider/model family: Anthropic / Fable 5 (exact provider model id was not exposed in the captured error)
- Working comparison in the same thread: OpenAI Codex / GPT family

### Related Issues

- #96 — older general report that compaction itself can exceed the limit; also worked around by switching to GPT. This issue adds a concrete modern payload breakdown and the unrecoverable `single-exchange` condition.
- #1815 — 200k context overflow caused by total prompt overhead; fixed for that case, but this report exceeds a 1M limit even though conversation tokens are below 1M.
- anomalyco/opencode#6068 and anomalyco/opencode#4845 — related upstream auto-compaction / unrecoverable prompt-overflow reports.

### Additional Context

This is particularly costly in autonomous deep-work threads: the active goal and tool state are durable, but one provider becomes unusable for all later turns. Starting a new thread loses the very context/goal continuity OMO is meant to preserve.

KNN-07:
## Summary

Recover Anthropic/Fable sessions that exceed the provider limit after Claude Agent SDK system, tool, and attachment injection. Senpi normally delegates compaction to the SDK, but the SDK cannot compact an oversized single flattened exchange; this patch permits Senpi's emergency overflow path to recursively regenerate the existing synthetic summary and retry the interrupted turn.

Normal threshold and speculative compaction remain delegated to the Claude Agent SDK. Only provider-reported overflow uses the Senpi fallback.

## Changes

- Extend the pinned `@code-yeongyu/senpi@2026.8.30-2` patch so overflow bypasses both sticky SDK-delegation guards and the builtin external-owner hook.
- Enable `allowSummaryOnly` for pre-prompt and automatic overflow preparation, allowing a prior synthetic summary to be summarized again when no older exchange remains.
- Add installed-runtime regression coverage for normal delegation, overflow admission, recursive summary regeneration, and one retry continuation.

## QA & Evidence

- **What was tested:** `bun test packages/omo-native/test/senpi-overflow-compaction-patch.test.ts`
  - **Observed result:** 3 passed, 0 failed, 9 assertions.
  - **Artifact:** `.omo/evidence/omo-senpi-adapter/20260830-anthropic-overflow-recovery/overflow-regression-test.log`
  - **Why sufficient:** Exercises the compiled patched Senpi runtime, including sticky delegation and summary-only auto-overflow.
- **What was tested:** isolated real Senpi driver with the OMO plugin and mock provider.
  - **Observed result:** PASS; ultrawork injection PASS; comment-checker PASS; `realSenpiUntouched: true`.
  - **Artifact:** `.omo/evidence/omo-senpi-adapter/20260830-anthropic-overflow-recovery/live-driver.json`
  - **Why sufficient:** Proves the patched CLI starts and the adapter loads without touching the real Senpi home.
- **What was tested:** `bunx tsgo --noEmit -p packages/omo-senpi/tsconfig.json && bun run test:senpi`
  - **Observed result:** 2,423 passed, 7 skipped, 0 failed; evidence resolver 10 passed, 0 failed.
- **What was tested:** `bun run typecheck` and `bun run build`
  - **Observed result:** both passed.
- **What was tested:** full `bun test`
  - **Observed result:** 16,862 passed, 44 skipped, 7 failed. Six unrelated concurrent-suite failures passed in isolation (22/22). The remaining embedded-manifest parity probe explicitly requires Bun >=1.4 directory-asset support; this workstation has Bun 1.3.14.
  - **Artifact:** `.omo/evidence/omo-senpi-adapter/20260830-anthropic-overflow-recovery/full-test.log`

## Risks & Residuals

- Emergency overflow now uses Senpi summarization even on the SDK-native lane. Scope is limited to `reason === "overflow"`; ordinary ownership is unchanged.
- A paid 1M-token Anthropic request was not run because no task-scoped credential is available. The regression drives the exact installed-runtime recovery state instead.
- The dependency patch should be removed when an equivalent Senpi release is pinned.

## Related Issues

Fixes https://github.com/code-yeongyu/oh-my-openagent/issues/7497

---
## Summary by cubic
Fixes Claude Agent SDK sessions that overflow after system, tool, and attachment injection. The SDK cannot compact an oversized single flattened exchange, so provider-reported overflow now routes through Senpi's emergency compact path to regenerate the existing synthetic summary and retry the interrupted turn. Fixes #7497.

**Bug Fixes**
- Overflow bypasses Senpi's sticky SDK-delegation guards and the builtin external-owner hook.
- Enables `allowSummaryOnly` for pre-prompt and automatic overflow, letting a prior synthetic summary be summarized again when no older exchange remains.
- Adds installed-runtime regression coverage for normal delegation, overflow admission, recursive summary regeneration, and one retry continuation.
- Emergency overflow now uses Senpi summarization on the SDK-native lane, limited to `reason === "overflow"`; remove the patch when an equivalent Senpi release is pinned.

<sup>Written for commit 6a0a2b3dd7e761f52ce2517c2cac7b271b8d95a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7518?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-native/test/senpi-overflow-compaction-patch.test.ts
+++ b/packages/omo-native/test/senpi-overflow-compaction-patch.test.ts
@@ -1,0 +1,193 @@
+import { afterEach, describe, expect, test } from "bun:test"
+import { mkdtempSync, rmSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { dirname, join } from "node:path"
+import { fileURLToPath, pathToFileURL } from "node:url"
+
+// Senpi's package exports hide these compiled internals, so the regression loads them from the resolved runtime path.
+const senpiDistDir = dirname(fileURLToPath(import.meta.resolve("@code-yeongyu/senpi")))
+const agentSessionModule = await import(pathToFileURL(join(senpiDistDir, "core", "agent-session.js")).href) as {
+  AgentSession: {
+    prototype: {
+      _runPrePromptCompaction: (...args: unknown[]) => Promise<boolean>
+      _runAutoCompaction: (...args: unknown[]) => Promise<boolean>
+    }
+  }
+}
+const compactionExtensionModule = await import(
+  pathToFileURL(join(senpiDistDir, "core", "extensions", "builtin", "compaction", "index.js")).href
+) as { default: (api: CompactionApi) => void }
+const compactionSettingsModule = await import(
+  pathToFileURL(join(senpiDistDir, "core", "compaction", "compaction-settings.js")).href
+) as { DEFAULT_COMPACTION_SETTINGS: Record<string, unknown> }
+
+const temporaryDirectories: string[] = []
+afterEach(() => {
+  for (const directory of temporaryDirectories.splice(0)) {
+    rmSync(directory, { recursive: true, force: true })
+  }
+})
+
+type CompactionRequest = {
+  allowSummaryOnly?: boolean
+  reason?: string
+}
+
+type CompactionApi = {
+  on: (event: string, handler: (event: Record<string, unknown>, context: Record<string, unknown>) => unknown) => void
+  appendEntry: () => void
+  getActiveTools: () => string[]
+  getAllTools: () => unknown[]
+  events: { emit: () => void }
+}
+
+function createCompactionHook() {
+  let handler: ((event: Record<string, unknown>, context: Record<string, unknown>) => unknown) | undefined
+  const api: CompactionApi = {
+    on: (event, candidate) => {
+      if (event === "session_before_compact") handler = candidate
+    },
+    appendEntry: () => {},
+    getActiveTools: () => [],
+    getAllTools: () => [],
+    events: { emit: () => {} },
+  }
+  compactionExtensionModule.default(api)
+  if (!handler) throw new Error("Senpi compaction extension did not register session_before_compact")
+  return handler
+}
+
+function createSdkLaneContext() {
+  const agentDir = mkdtempSync(join(tmpdir(), "omo-senpi-overflow-test-"))
+  temporaryDirectories.push(agentDir)
+  let modelReads = 0
+  return {
+    agentDir,
+    cwd: agentDir,
+    sessionManager: { getBranch: () => [], getEntries: () => [] },
+    get model() {
+      modelReads += 1
+      return modelReads <= 2
+        ? { provider: "claude-sdk-oauth", id: "fable-5", contextWindow: 1_000_000 }
+        : undefined
+    },
+  }
+}
+
+function compactionEvent(reason: "threshold" | "overflow") {
+  return {
+    type: "session_before_compact",
+    reason,
+    willRetry: reason === "overflow",
+    requestId: `request-${reason}`,
+    preparation: { messages: [], settings: compactionSettingsModule.DEFAULT_COMPACTION_SETTINGS },
+    branchEntries: [],
+    signal: new AbortController().signal,
+  }
+}
+
+describe("patched Senpi overflow compaction", () => {
+  test("keeps normal Claude SDK compaction delegated but admits emergency overflow recovery", async () => {
+    const hook = createCompactionHook()
+
+    const thresholdResult = await hook(compactionEvent("threshold"), createSdkLaneContext())
+    const overflowResult = await hook(compactionEvent("overflow"), createSdkLaneContext())
+
+    expect(thresholdResult).toMatchObject({ cancel: true, rejectionCause: "external-owner" })
+    expect(overflowResult).toBeUndefined()
+  })
+
+  test("bypasses sticky delegation and enables recursive summary regeneration for pre-prompt overflow", async () => {
+    const requests: CompactionRequest[] = []
+    const delegatedSession = {
+      _isCompactionDelegated: () => true,
+      _claimCompactionController: () => {},
+      _emit: () => {},
+      _executeCompaction: async (request: CompactionRequest) => {
+        requests.push(request)
+        return { accepted: true }
+      },
+      _compactionLifecycle: { state: { status: "succeeded" } },
+      _compactionAbortController: undefined,
+    }
+
+    const accepted = await agentSessionModule.AgentSession.prototype._runPrePromptCompaction.call(
+      delegatedSession,
+      undefined,
+      true,
+      "overflow",
+      true,
+    )
+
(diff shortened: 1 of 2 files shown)


MoerAI (changes requested):
Thanks for pursuing the overflow recovery. The scoped logic is promising, but the dependency patch is mechanically broken.

Its hunk offsets do not match the pinned Senpi tarball. Bun 1.4.0 applies them at the literal positions, inserting a method inside another block and causing the parse error that collapses the Ubuntu suite. Current `dev` has also moved to the 2026.8.31 patch file.

Please rebase, regenerate the patch from the real tarball rather than hand-editing offsets, verify the built module imports under CI's Bun version, and place the regression in a suite CI actually runs. Commit the cited evidence or remove those references.

MoerAI:
Drive-by diagnosis while auditing Windows CI. Not closing this - it is your PR and the call is yours. Sharing the root cause because all five failing checks trace to a single one.

**All five failures cascade from `build`.** `build`, `codex-compatibility (ubuntu-latest, full)`, `senpi-compatibility (ubuntu-latest)`, `test (ubuntu-latest, 1/2)` and `test (ubuntu-latest, 2/2)` are all downstream of one parse error, so there is likely only one thing to fix.

**Root cause: a truncated line in the patched Senpi runtime bundle.** From the `build` job on head `6a0a2b3dd`:

```
packages/omo-opencode/src/shared/dist-bundle-prompt-content.test.ts:
4843 |     async _runPrePromptCompaction(lastAssistantMessage, skipAbortedCheck, reason = "pre_prompt", willRetry = fals
                                         ^
error: Expected "=>" but found "("
error: Expected ";" but found ")"
error: Unexpected }
 2 errors
```

Note the line ends at `willRetry = fals` - it is cut off mid-token, so the bundle no longer parses. Since this PR patches the Senpi runtime for Claude SDK overflow recovery, the most likely cause is that the patch application truncated or mis-spliced that line rather than a logic error in your change.

**Worth confirming:** this CI result is against your current head `6a0a2b3dd`, not a stale commit, so it should still reproduce.

**Suggested next step:** regenerate the patched bundle rather than hand-editing it, then re-check that `_runPrePromptCompaction` retains its full signature. Once `build` parses, the other four checks should clear on their own.

For context, `dev` is currently green, so nothing here is blocking you.

Issue #7546, opened 2026-08-31
build: dist bundle emits a duplicate node:fs declaration - dev is red on Verify dist bundle tests


code-yeongyu:
## Problem

`dev` is red: the `build` job's "Verify dist bundle tests" step fails with bundler-emitted syntax errors, independent of any dependency change.

```
error: "writeFileSync" has already been declared
error: Expected ";" but found "{"
error: Unexpected }
8 errors
```

Command: `bun test packages/omo-opencode/src/shared/dist-bundle-bun-globals.test.ts packages/omo-opencode/src/shared/dist-bundle-prompt-content.test.ts`

## Why this is a product defect, not a test defect

The two test SOURCES on `dev` are clean (single `node:fs`/`node:fs/promises` imports each). The duplicate `writeFileSync` declaration therefore comes from the **generated dist bundle** these tests load and validate — i.e. the omo-opencode bundling step emits a module with a duplicated declaration, so the shipped bundle is syntactically invalid under Bun.

## Evidence it is pre-existing on dev (not PR-induced)

- dev merge-commit runs are already failing: run 33383686811 (head `c53faf72fd`, conclusion failure) and run 33381382489 (failure).
- Reproduced on an unrelated dependency-only PR (#7545, senpi pin bump): its diff touches only `package.json` x4, `bun.lock`, and a version-scoped patch rename — zero overlap with `packages/omo-opencode/**`, and a version string cannot create duplicate declarations.
- Likely regression window: `f7e8221a9 refactor(omo-opencode): move plugin src into packages/omo-opencode` (the last change to these dist-bundle tests) — the move plausibly left two module copies (old + new path) feeding the bundle.

## Ask

Fix the bundling so the emitted dist module declares each `node:fs` binding once (check for a duplicated/stale source copy left behind by the package move), and keep these dist-bundle tests as the gate. Until then every PR to `dev` inherits a red `build`.

code-yeongyu:
## Correction: root cause reattributed (by the reporter)

I filed this attributing the duplicate `node:fs` declaration to the omo-opencode **bundler**. That was wrong, and the evidence that corrected it:

The failing frames point inside the **patched senpi package**, not omo sources:

```
node_modules/.bun/@code-yeongyu+senpi@2026.8.31+d96a03985ed310ba/
  node_modules/@code-yeongyu/senpi/dist/core/extensions/builtin/hooks/trust-storage.js:2-3
```

[...]

PR #7548, opened 2026-08-31
ci(patches): verify version-scoped patch hunks against the published package (refs #7546)

robbe1912:
## Why

#7546: a version-scoped patch rename shipped hunks whose behavior had already landed upstream, and the invalid bundle reached `dev` before the dist bundle tests caught it. The follow-up ask on the issue: a guard that verifies each hunk still applies (`git apply --check`) against the newly published package, and reports hunks whose behavior landed upstream as obsolete (drop) instead of re-applying them.

## What

- `script/verify-patch-hunks.ts`: for every `patches/*.patch`, parse `<pkg>@<version>` from the file name, fetch the exact published tarball from the npm registry, and `git apply --check` it. When the whole patch fails, each hunk is checked standalone and classified:
  - `obsolete`: every added line already exists upstream (or a removal-only hunk has nothing left to remove), so the fix is to drop the hunk and regenerate;
  - `conflict`: the published source moved on, so the hunk needs a rebase onto the pinned version.
- Non-applying hunks fail the run with `::error` file annotations and a drop/rebase remediation per hunk.
- `.github/workflows/ci.yml`: new "Verify version-scoped patch hunks" step in the `build` job, placed before dependency install (the guard needs only Bun, git, and network access) and before `bun run build`, so a stale patch fails fastest; `run_heavy` gated like its neighbors.
- `script/verify-patch-hunks.test.ts`: 9 unit tests (file name parsing, hunk splitting including multi-file diffs, obsolete/conflict classification).
- `script/AGENTS.md`: script table and test count kept in sync.

## QA

- `bun test script/verify-patch-hunks.test.ts`: 9 pass, 0 fail.
- Positive: the committed `@code-yeongyu/senpi@2026.8.31` patch against the published tarball: all hunks apply, exit 0.
- Negative (repro of the #7546 defect): the pre-fix patch content from `e109689` against the same published version: hunks 1, 2, 5 apply; hunk 3 reported obsolete (its imports landed upstream); hunk 4 reported conflict (upstream rewrote the read/update path); exit 1. This matches the shipped fix `3fdd94f`, which dropped the trust-storage half.
- `bun run typecheck:script`: clean for the new files; one pre-existing dev error in `script/senpi-hooks-state.test.ts:52` (from 6139295) is unrelated to this PR.
- actionlint 1.7.7 on the modified workflow: clean.
- Evidence: `.omo/evidence/20260831-patch-hunk-guard/`

Portability note: patch text is normalized CRLF to LF before checking (Windows checkouts with `core.autocrlf=true` and no `*.patch` rule in `.gitattributes` otherwise produce false "corrupt patch" failures); CI (Linux) is unaffected either way.

---
## Summary by cubic
Adds a CI guard that verifies every `patches/*.patch` hunk still applies to the pinned published npm package, so stale version-scoped patches now fail the build instead of shipping invalid modules to `dev` (refs #7546).

- The guard fetches the exact published tarball per patch and runs `git apply --check`; failing hunks are classified as obsolete (drop and regenerate) or conflict (rebase onto the pinned version) with `::error` annotations.
- Runs in the `build` job before dependency install and is gated on `run_heavy`.
- Patch text is normalized CRLF to LF so Windows checkouts don't produce false failures.

<sup>Written for commit a770d7ff3707f16c6b58ceecd0c1c122def66c8f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


robbe1912:

PR #7561, opened 2026-08-31
fix(runtime-fallback): stop the cancel-poison re-dispatch loop (incident ses_fa750d527ffe)

privettoli:
## Summary

A fallback session in the incident loop was aborted and re-dispatched every 15.05 seconds, twelve times, with `attempt:1` forever: each cycle classified our own watchdog/quota abort as an external cancellation, rewound retry state to the primary model, re-armed nothing, and slipped the prompt-gate dedupe window edge-for-edge because the 15s hold exactly equaled the 15s retry period. This PR fixes all five defects with six atomic commits; the observable outcome is that a healthy-but-slow fallback is no longer aborted mid-stream, the max-attempts cap can actually trip, and an identical retry is refused twice over.

## Changes

**Abort classification and event routing (F1a, F1b)**
- `2600e765a`: watchdog and quota aborts are recorded in the internally-aborted allowlist, so the cancellation branch no longer treats our own aborts as user cancellations.
- `80bf3ef75`: `hook.ts` now forwards `message.updated` to the base event handler; previously the composed hook dropped it before the base handler, so progress events never reached fallback state tracking.

**Timeout re-arm on progress (F2)**
- `f3369b54c`: `rearmSessionFallbackTimeout` extends the armed window on assistant progress that has no visible output yet (armed-only; the agent captured at arm time is reused). Prolonged silence still aborts at a full window. 12 test helper literals widened with the new helper.

**Cancel rebuild seeds the live model (F3-i)**
- `fb0050dde`: `resetRetryState` now seeds from the event model, else `state.currentModel`, instead of rewinding to `originalModel`. Attempt accounting still zeroes, but the chain position is preserved, so a later error dispatches AFTER the live model and `max_fallback_attempts` becomes reachable.
- Three pre-existing cancellation tests asserted the rewind itself (currentModel back to the primary after cancel). Those tests encoded this bug: they are updated to expect the live seed, with rationale in the commit message. Every `attemptCount-0` reset pin is untouched.

**Duplicate dispatch guards (F3-ii, F3-iii)**
- `0d443d6c9`: the dispatcher records the accepted `(retryMessageID, model)` tuple per session and refuses the identical tuple while the fallback result is awaited. Timeout escalation to the next model is a different tuple and keeps flowing (pinned by a control test).
- `d88dd9cfc`: the dispatch call site passes `semanticDedupeHoldMs = effective timeout + 5s margin` (option `session_timeout_ms` wins over `timeout_seconds*1000`; disabled timeouts defer to the gate default), so a retry cycle equal to the timeout period can no longer edge the dedupe window.

## QA & Evidence

All artifacts are committed under `.omo/evidence/20260831-runtime-fallback-cancel-poison/` in this branch (WHAT-TESTED / OBSERVED / WHY-ENOUGH / OMITTED + raw logs).

- **What:** `bun test packages/omo-opencode/src/hooks/runtime-fallback/` (state machine: abort sources, routing, re-arm, cancel seeding, tuple refusal, dedupe hold).
  **Observed:** 305 pass / 0 fail, 33 files, exit 0. Every fix above has a test that failed pre-fix (RED transcripts captured during implementation; final GREEN in log).
  **Why sufficient:** tests drive the real composed handlers/dispatcher/gate module with only clock/client/provider-error injection.
- **What:** `bun test` over prompt-async-gate suites + `src/shared/prompt-async-route-audit.test.ts` (raw-prompt-route invariant).
  **Observed:** 103 pass / 0 fail, 13 files (`test-gate-audit.log`).
- **What:** `bun run typecheck` (tsgo root + script + all packages).
  **Observed:** exit 0 (`typecheck.log`).
- **What:** opencode-qa minimal viable on the real binary (v1.18.25), isolated XDG sandbox: `server-smoke.sh` + `sse-hook-probe.sh --self-test`.
  **Observed:** health + 162 documented paths + 401 unauthenticated; SSE delivered `server.connected`; real `opencode.db` session count unchanged 1475 before/after (`opencode-qa-isolation.txt`).
  **Why sufficient:** proves the plugin load path and the event-hook channel the fallback consumes are unharmed on the real harness, with host DB untouched.
- **What:** attribution of the pre-existing `bun test packages/omo-opencode/src/hooks/` directory-scope failures (46 on branch).
  **Observed:** the byte-identical fail-name set reproduces on an unmodified `origin/dev` detached checkout, and every affected file passes standalone (`runtime-fallback/index.test.ts`: 67/0 in isolation). Side-by-side logs + name sets in the evidence dir.
  **Why sufficient:** isolates a base test-ordering defect (cross-file mock/state leakage) that is out of scope for this bugfix; CI's serial-quarantine root suite is the authoritative full-suite gate.
- **Rebase:** branch replayed onto `origin/dev b5cbae3fb`; all logs above captured post-rebase.

## Risks & Residuals

- No live provider-failure loop reproduction (would burn paid quota and is non-deterministic). Mitigation: deterministic fake-clock integration tests pin the full abort/re-dispatch cadence behavior, and both new guards fail closed. Accepted.
- The pre-existing hooks-directory test-ordering leak is NOT fixed here (test-suite hygiene, out of scope); captured for follow-up. Accepted.
- `timeout_seconds` default (30s) and all user-facing config are unchanged.
- Remaining regression risk is bounded: the timeout abort now only fires after a full silent window, and a pathological chain exhausts within `max_fallback_attempts` instead of looping at attempt 1.

---
## Summary by cubic
Fixes the runtime-fallback cancel-poison loop where a fallback session was aborted and re-dispatched every 15 seconds at `attempt:1` forever. Now own watchdog/quota aborts are treated as internal, timeouts only fire after a full silent window, and identical retries are refused.

**Changes**
- Watchdog and quota aborts are classified as internal, so they no longer surface as user cancellations.
- `message.updated` is forwarded to the base event handler so fallback state sees progress.
- Assistant progress re-arms the fallback timeout; prolonged silence still aborts after a full window.
- Cancel rebuild seeds retry state from the live model instead of rewinding to the original, so `max_fallback_attempts` can trip.
- The dispatcher refuses re-dispatch of an already-accepted `(messageID, model)` tuple, and the semantic dedupe hold now exceeds the timeout period.
- Three cancellation tests were updated to expect the live-model seed; they previously encoded the rewind bug.

**Verification**
- 305 runtime-fallback tests, 103 gate/audit tests, and full typecheck pass; opencode-qa smoke and SSE probe pass on the real binary.
- A pre-existing test-ordering failure in the hooks directory is not fixed here; it reproduces on unmodified `dev`.

<sup>Written for commit 25572f05a374fd5958e63fb3f6aecb9387e8586f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


privettoli:

privettoli:
Sorry for the noise, for some reason the agent decided I wanted to raise a PR for this. 😁

PR #7601, opened 2026-09-01
test(omo-senpi): await DAG child disposal explicitly

code-yeongyu:
Flake fix for a test that only became visible once the permanent senpi-compat red was cleared by #7594.

`assembled DAG runtime > #given a live-shaped in-process child whose abort rejects #when the assembled runtime cancels #then no unhandled rejection escapes and a later run still completes` failed at 8359ms on senpi-compatibility (windows-latest), CI run 33508665211.

Root cause (`packages/omo-senpi/src/components/task/dag-runtime.test.ts:596-600`): the test used `setImmediate` to INFER that deferred child disposal had finished. That tick ordering is not guaranteed relative to the runtime's cancellation bookkeeping, so on Windows the test could observe the wrong state or stall for seconds.

Fix: subscribe to the exact transition instead of guessing a tick — `ScriptedRunner` now exposes an explicit `disposed` deferred, and the test awaits that through the existing bounded `within()` helper. Test-only; **no production change**.

Production behavior confirmed correct as-is: DAG cancellation deliberately calls `cancelTask(..., { abort: "skip" })` for in-process children, and lifecycle disposal is deferred until `waitForOutcome()` settles, which is what keeps the injected abort rejection from escaping.

Mutation proof: removing `{ abort: "skip" }` from production cancellation makes the test fail exactly as designed (`abortCalls` 0 -> 2, AbortError unhandled rejections captured); restoring it returns green.

Evidence: 30x loop green (~33-37ms per run), omo-senpi scope 2527 pass / 0 fail / 1 pre-existing platform skip, both tsgo typechecks pass, LSP clean.

---
## Summary by cubic
Fixes a flaky DAG runtime test by awaiting the child's explicit `disposed` deferred instead of inferring completion with `setImmediate`, so the test no longer stalls or fails on Windows. Test-only; no production behavior changes.

<sup>Written for commit 32aa4aca7924ad4ce29256cb29f2b0a0d333b18e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-senpi/src/components/task/dag-runtime.test.ts
+++ b/packages/omo-senpi/src/components/task/dag-runtime.test.ts
@@ -61,6 +61,7 @@ class ScriptedRunner implements ManagedRunner {
     readonly emit: (event: ManagedChildEvent) => void
     readonly listenerCount: () => number
     readonly settle: (output: string) => void
+    readonly disposed: Promise<void>
     readonly fail: (message: string) => void
   }> = []
   readonly #signals = new Map<number, ReturnType<typeof deferred<void>>>()
@@ -74,6 +75,7 @@ class ScriptedRunner implements ManagedRunner {
 
   start(spec: ManagedStartSpec): Promise<ManagedChildHandle> {
     const outcome = deferred<RunnerOutcome>()
+    const disposed = deferred<void>()
     const listeners = new Set<(event: ManagedChildEvent) => void>()
     const handle: ManagedChildHandle = {
       task_id: spec.taskId,
@@ -95,6 +97,7 @@ class ScriptedRunner implements ManagedRunner {
       lastAssistantText: () => undefined,
       dispose: () => {
         this.disposeCalls += 1
+        disposed.resolve()
         return Promise.resolve()
       },
     }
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Follow-up correction: this fix was **incomplete**, and CI proved it after merge.

On dev push run 33525950049 (head `dff4a156a`) the same test failed again on `senpi-compatibility (windows-latest)` at 7019ms, and it brought two siblings with it:
- `assembled DAG runtime > ... detach switches sessions before its task settles ...` (5953ms)
- `assembled DAG runtime configuration > subscriber_ring is one ... durable overflow` (7202ms)

Why this PR could not hold: it correctly replaced a `setImmediate` tick-guess with an explicit `disposed` deferred, but the wait itself remained `await within(disposed)` — and `within()` (`dag-runtime.test.ts:42`) imposed a hard **300ms real-clock** deadline, used ~50 times in that file, with a second copy in `dag-runtime-config.test.ts:26`. The enclosing tests allow `{ timeout: 15_000 }`, so the outer budget was generous while every inner await stayed bounded at 300ms; on a loaded Windows runner these ran 20x+ that budget.

So the exposure was a shared seam, not a per-test bug — fixing one test left its neighbours on the same racy bound. Superseded by #7607, which makes both `within()` implementations pass-throughs to already event-driven promises (no timeout inflation), adds occurrence-aware `whenAttached`, and is mutation-proven: forcing the pass-through to reject fails 13 of 17 runtime/config tests, confirming it was the common deadline seam.

PR #7621, opened 2026-09-02
fix(memory-core/locks): reclaim stale .lock.recovery before publishing new one

### LINKED ISSUE #7573: memory: reflection-scheduler lock recovery deadlocks when the recovery lock itself is stale (dead holders require manual rm)
## Summary

The memory subsystem's `reflection-scheduler.lock` recovery protocol deadlocks when BOTH the lock and its `.lock.recovery` companion are stale (holders dead): every subsequent session bind fails with `LockContentionError: Lock is held` and boot-time `memory bind-time reconcile failed`. Observed on mengmotaHost (omo v5.0.0-beta.31 runtime bundle, desktop shared host) with locks whose recorded holders died 4 days earlier.

## Evidence (2026-09-01, agent userdata-a676c9d8)

- `reflection-scheduler.lock`: {pid: 23997, process_start: Fri Aug 28 13:34:49} — pid dead (ps empty).
- `reflection-scheduler.lock.recovery`: {pid: 23813, Fri Aug 28 13:34:46} — dead.
- `reflection-scheduler.lock.recovery.candidate-8533c4ed...`: {pid: 42430, Tue Sep 1 15:32:00} — today's recovery ATTEMPT, wedged behind the stale recovery lock (candidate never promoted).
- Every session boot on the shared host then logs: `memory bind-time reconcile failed { LockContentionError: Lock is held: .../reflection-scheduler.lock }`.
- Lock records carry {pid, process_start} exactly for staleness detection, but recovery evidently cannot proceed when the recovery lock itself is stale — recovery-of-recovery is missing.
- Vicious cycle under host instability: sessions killed while holding the lock re-mint stale locks each round; manual `rm` of the stale trio unblocks binds immediately (verified: sessions reached running right after cleanup).

## Expected

Stale detection (dead pid / mismatched process_start) must apply to the recovery lock the same as the primary lock, and candidate promotion must not require a live recovery holder. A crashed holder should never require manual lock surgery.

## Context

Amplified by the shared-host kill-war of code-yeongyu/omo-desktop-app#273 (hosts dying repeatedly -> locks leaking every round).

MoerAI:
## What changed

When `recoverDeadOwner` finds a stale primary lock (dead owner), it tries to publish a recovery lock at `${lockPath}.recovery` before evicting the primary. If that recovery lock already exists from a previous crashed recovery attempt whose owner is also dead, `publishExclusive` returns `false` (EEXIST from `link`) and `recoverDeadOwner` bails out without reclaiming anything. The contender then hits `LockContentionError` on the primary lock — permanent deadlock requiring manual `rm` of the recovery file.

Fixes #7573

## Root cause

`publishExclusive` uses `link(candidate, target)` which is atomic but returns `false` on EEXIST. There was no path to reclaim a stale recovery lock, so a dead-owner recovery file permanently blocked all future recovery attempts on that lock.

## Fix

Added `reclaimStaleRecoveryLock(recoveryPath)` called at the top of `recoverDeadOwner`, before `publishExclusive`. It reads the existing recovery lock, runs `isProvenDead` on its owner (same staleness detection used for the primary lock), and unlinks it if dead. ENOENT is swallowed (already gone). Any other unlink error propagates.

The fix is POSIX-only: `isProvenDead` returns `false` on Windows (no `getProcessStartIdentity` probe), so the new helper is a no-op there — same fail-closed behaviour as the existing primary-lock recovery path.

## Test

New test file: `packages/memory-core/src/locks/stale-recovery-lock.test.ts`

- **RED before fix**: 1 fail (deadlock scenario), 2 pass (regression guards)
- **GREEN after fix**: 3 pass, 0 fail
- **Full locks suite**: 22 pass, 0 fail across 5 files
- **Typecheck**: clean (`tsgo --noEmit`)

### Test cases
1. Both primary and recovery locks held by dead processes → contender succeeds without manual file deletion (the bug)
2. Single stale primary lock → still recovers (existing behaviour preserved)
3. Live recovery holder → still blocks the contender (safety invariant preserved)

## Observed behaviour

Before fix:
```
(fail) stale recovery lock deadlock (issue #7573) > ... #then it succeeds without manual file deletion
```

After fix:
```
 3 pass
 0 fail
Ran 3 tests across 1 file. [682.00ms]
```

---
## Summary by cubic
Fixes #7573 so a stale `.recovery` lock left by a crashed recovery attempt no longer blocks lock recovery until manually deleted. `recoverDeadOwner` now reclaims stale recovery locks whose owner is proven dead before publishing a new one, with reclamation serialized so concurrent contenders cannot both proceed.

**Bug Fixes**
- Reclamation is serialized through an immutable hard-link claim chain outside the candidate namespace, so the candidate sweeper cannot remove an active claim and contenders cannot displace an elected claimant.
- Claims left by a dead reclaimer are advanced through a bounded successor chain; a chain that cannot be attributed fails closed with a non-retriable `LockRecoveryBlockedError` instead of deadlocking.
- The recovery lock is re-read before unlinking; if it changed, reclamation stops and the replacement owner is preserved.
- POSIX-only; Windows keeps the existing fail-closed behavior because `isProvenDead` returns `false` there.
- Adds regression tests for the original deadlock, orphaned claims, legacy-claim fail-closed behavior, replacement-owner preservation, and two contenders racing to reclaim.
- Recovery error messages use harness-neutral wording so the shared-core extraction guard accepts them.
- Refreshes the generated `omo-senpi` extension bundles and records Senpi adapter QA evidence for the race, claim-namespace, and orphan-claim scenarios.

<sup>Written for commit 697e881233cce482f57e467520a772a950c2bc2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
I have read the CLA Document and I hereby sign the CLA.

MoerAI:

MoerAI:
Closing because current dev removed packages/omo-senpi/plugin/extensions/omo-memory-mcp.js in db0feecd0, leaving this branch with a modify/delete conflict and stale generated runtime bundles. Issue #7573 remains open; this closure does not mean the lock-recovery bug is fixed. The change needs to be reworked and re-verified against the current adapter layout. Happy to reopen or rework if this fix is still wanted.

PR #7622, opened 2026-09-02
fix(omo-native): make close_session teardown idempotent under unresolved abort (fixes #7587)

### LINKED ISSUE #7587: rpc/session-registry: close_session teardown is unbounded and non-idempotent — a wedged abort permanently retains the session path reservation
## Summary

`close_session` teardown can hang forever, permanently retaining a session's path reservation. The fix belongs here (host), not in clients: a client cannot bound a host-side unbounded await.

## Root cause (session-registry.js)

- `closeMarked()` awaits `abort() -> waitForIdle() -> dispose()` with **no bound**. If the session's abort never resolves, teardown never completes and the path reservation is never released.
- `beginClose()` throws `unknown_session` when the entry is already `closing`, so a client's second/reclaim `close_session` is rejected rather than joining the in-flight teardown.
- Consequence for clients: after their own bounded close times out, they cannot reclaim the path. `open_session` on that path keeps failing in-use, and the only escape is restarting the whole shared host — which kills every unrelated session on it.

## Requested change

1. **Bounded teardown.** Race the graceful `abort -> waitForIdle -> dispose` chain against a host-side grace deadline (~10s). On expiry, escalate to hard disposal that cannot block on the stuck promise, release the reservation, and transition to `closed`. Guarantee: `close_session` always terminates and the reservation is always eventually released. (Forced disposal is strictly less dangerous than the status quo escape hatch of killing the entire host process — same durability requirements, scoped to one session.)
2. **Idempotent close.** A `close_session` for an entry already in `closing` should join the in-flight teardown (await the same lifecycle mutex) and return its outcome, instead of throwing `unknown_session`.
3. *(Optional, cheap)* a per-row monotonic `lastTransitionAt` (or teardown phase) on `list_sessions` rows, so clients can distinguish "draining slowly" from "wedged" honestly. Purely additive; old clients ignore it.

No new RPC verb is needed — a host that can force-close on request can do it on its own timer, so this needs no capability negotiation. If caller-controlled grace is ever wanted, an optional `graceMs` field on `close_session` degrades safely on old hosts.

## Why now

Surfaced while narrowing a desktop-side blast radius: code-yeongyu/omo-desktop-app#278 / #281 stopped one wedged turn from killing every session on the shared host. That change is a strict improvement, but it exposes this pre-existing host gap — the affected thread is now stranded visibly instead of the whole host being nuked. Clients can only mitigate (poll `list_sessions` until the row disappears, bounded, then surface a legible error); the durable fix is the bounded teardown above.

MoerAI:
## What changed

Patches two bugs in `@code-yeongyu/senpi`'s `dist/modes/rpc/session-registry.js` via `senpi-patch.mjs`, and fixes a syntax error in the patcher that prevented it from running.

### Bug 1 — bounded teardown (the wedge)

`closeMarked()` awaited `session.abort()` then `session.waitForIdle()` sequentially with no timeout. If `abort()` never resolved (e.g. a hung subprocess), the lifecycle mutex was held forever, blocking every subsequent `close_session` call on the same handle.

**Fix:** wrap `abort() + waitForIdle()` in a `Promise.race` against a 10 s deadline (`TEARDOWN_TIMEOUT_MS`). On expiry, `dispose()` and `scope.close()` still run inside the `finally` block so the registry entry is always cleaned up.

### Bug 2 — idempotent close

`beginClose()` threw `unknown_session` when called a second time on a handle already in `"closing"` state. A concurrent `close_session` arriving while teardown was in progress would surface as a spurious error to the caller.

**Fix:** return the existing entry immediately when `entry.state === "closing"`, so the second caller joins the in-flight teardown instead of erroring.

### Patcher syntax fix

The `transforms` array in `senpi-patch.mjs` was missing one closing `],` (the second entry tuple was not closed before the outer array's `]`), causing `SyntaxError: Unexpected token 'for'` at the `for...of` loop. Added the missing bracket.

## Files changed

- `packages/omo-native/bin/senpi-patch.mjs` — two new patch tuples for `dist/modes/rpc/session-registry.js` + bracket fix
- `packages/omo-native/test/senpi-patch-session-registry.test.ts` — new test verifying patch applicability and correctness against the installed senpi package

## Observed behavior

```
bun test packages/omo-native/test/senpi-patch-session-registry.test.ts
 5 pass
 0 fail
 14 expect() calls
Ran 5 tests across 1 file. [700.00ms]
```

Patcher syntax check: `node --input-type=module --check < packages/omo-native/bin/senpi-patch.mjs` exits 0.

## Residual risk

- The 10 s deadline is hardcoded. If a legitimate teardown takes longer than 10 s (e.g. very slow subprocess shutdown), `dispose()` will be called while `abort()` is still pending. This is safer than the current infinite hang but may produce a benign double-dispose warning in edge cases.
- The patch targets the exact string content of `session-registry.js` at `@code-yeongyu/senpi@2026.8.31`. If senpi ships a new version that changes the teardown block, the sentinel check (`TEARDOWN_TIMEOUT_MS`) will detect the already-patched state and skip gracefully; an unrecognized new shape will cause the patcher to exit non-zero with a clear error message.

---
## Summary by cubic
Fixes `close_session` teardown in `@code-yeongyu/senpi` (fixes #7587) so a hung abort no longer blocks the session registry. The teardown now has a 10s deadline; if it expires, cleanup still runs, which is safer than the previous infinite hang but may produce a benign double-dispose warning.

- `beginClose()` now returns the existing entry when state is `"closing"` instead of throwing `unknown_session`, so concurrent closes join the in-flight teardown.
- The patcher's `transforms` array was missing a closing bracket; added it so the patcher runs.
- The patch targets the exact string content of `session-registry.js` at `@code-yeongyu/senpi@2026.8.31`; if senpi changes the teardown block, the patcher skips if already patched or exits non-zero with a clear error.
- Adds a test that verifies the patch applies and is idempotent against the installed senpi package.

<sup>Written for commit 5d9a170d3ce341d1d91ffc25e02288bac5c059f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7622?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-native/bin/senpi-patch.mjs
+++ b/packages/omo-native/bin/senpi-patch.mjs
@@ -20,24 +20,96 @@ try {
   throw new Error("omo-ai: unable to resolve the installed @code-yeongyu/senpi package", { cause: error })
 }
 
-const transforms = {
-  "dist/core/extensions/builtin/claude-sdk-oauth/session-registry-pump.js": [
+// Each entry is [relative-path, sentinel, [...[from, to] pairs]].
+// The sentinel is a string that appears in the patched output but not the
+// original; if it is already present the file is skipped as already patched.
+const transforms = [
+  [
+    "dist/core/extensions/builtin/claude-sdk-oauth/session-registry-pump.js",
+    "describeUnclaimedResult",
     [
-      'throw new SessionTurnAttributionError("Claude SDK OAuth result arrived before replay claim");',
-      'throw new SessionTurnAttributionError(describeUnclaimedResult(message));',
-    ],
-    [
-      'function handleMessage(registry, entry, message) {',
-      'function describeUnclaimedResult(message) {\n    const errors = Array.isArray(message.errors) ? message.errors : [];\n    const detail = errors.length > 0 ? String(errors[0]) : typeof message.result === "string" ? message.result : typeof message.error === "string" ? message.error : typeof message.terminal_reason === "string" ? message.terminal_reason : undefined;\n    return `Claude SDK OAuth query result${typeof message.subtype === "string" ? ` (${message.subtype})` : ""}${detail ? `: ${detail}` : ""}`;\n}\nfunction handleMessage(registry, entry, message) {',
+      [
+        'throw new SessionTurnAttributionError("Claude SDK OAuth result arrived before replay claim");',
+        'throw new SessionTurnAttributionError(describeUnclaimedResult(message));',
+      ],
+      [
+        'function handleMessage(registry, entry, message) {',
+        'function describeUnclaimedResult(message) {\n    const errors = Array.isArray(message.errors) ? message.errors : [];\n    const detail = errors.length > 0 ? String(errors[0]) : typeof message.result === "string" ? message.result : typeof message.error === "string" ? message.error : typeof message.terminal_reason === "string" ? message.terminal_reason : undefined;\n    return `Claude SDK OAuth query result${typeof message.subtype === "string" ? ` (${message.subtype})` : ""}${detail ? `: ${detail}` : ""}`;\n}\nfunction handleMessage(registry, entry, message) {',
+      ],
     ],
   ],
-}
+  [
+    "dist/modes/rpc/session-registry.js",
+    "TEARDOWN_TIMEOUT_MS",
+    [
+    // Bug 2: make beginClose() idempotent — a second close_session while state is
+    // "closing" joins the in-flight teardown instead of throwing unknown_session.
+    [
+      `    beginClose(handle) {
+        const entry = this.entries.get(handle);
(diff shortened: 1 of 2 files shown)


MoerAI:

MoerAI:
Closing because the linked issue is already closed and the fix has been superseded on dev. Happy to rework if this behavior still needs a follow-up.

PR #7624, opened 2026-09-02
fix(model-core): cap grok output reserve when output limit equals context window (fixes #7541)

### LINKED ISSUE #7541: [Bug]: Models with no separate output limit (xai/grok-4.6) lose half their context — outputReserve treats maxTokens as a reservation
## Summary

`xai/grok-4.6` becomes unselectable once a session grows past ~154K tokens, despite the model having a **500K** context window. The usability projector reserves **250,000 tokens — half the window — for output**.

The same model via a different provider (`github-copilot/grok-4.6`) works fine on the identical session, which isolates the problem to how `maxTokens` is interpreted.

## Repro

```
omo -c --model xai/grok-4.6     # session with ~240K live context
```

```
ModelUsabilityBudgetError: Model "xai/grok-4.6" cannot switch: target context window 500000
tokens is 85889 tokens short of the 585889-token requirement (live context 240565,
system prompt 18693, active tool schemas 15671, output reserve 250000, compaction reserve 20000,
speculation lead 32768, safety margin 8192 [default]).
```

## Root cause

`dist/core/extensions/builtin/compaction/extension-wiring.js`:

```js
const MAX_OUTPUT_RESERVE_RATIO = 0.5;

export function getPromptContextWindow(contextWindow, maxTokens) {
    if (typeof maxTokens !== "number" || !Number.isFinite(maxTokens) || maxTokens <= 0 || contextWindow <= 0) {
        return contextWindow;
    }
    const outputReserve = Math.min(maxTokens, Math.floor(contextWindow * MAX_OUTPUT_RESERVE_RATIO));
    return contextWindow - outputReserve;
}
```

`model-usability-budget.js` then derives:

```js
const outputReserveTokens = input.model.contextWindow
    - getPromptContextWindow(input.model.contextWindow, input.model.maxTokens);
// => Math.min(maxTokens, contextWindow * 0.5)
```

The catalog entry is `contextWindow: 500000, maxTokens: 500000`, so `min(500000, 250000)` → **the 0.5 ratio cap becomes the value**, and half the window vanishes before any input is counted.

Effective input ceiling on this model:

```
500,000 − 250,000 (output) − 18,693 (system) − 15,671 (tools)
        −  20,000 (compaction) − 32,768 (speculation) − 8,192 (margin)
= 154,676 tokens
```

## Why this is an interpretation bug, not bad catalog data

I checked xAI's published spec before filing: **Grok 4.6 has no separate output limit — output shares the same 500K window as the prompt.** So `maxTokens: 500000` is *accurate*; it is the catalog's way of saying **"no distinct output ceiling"**.

`maxTokens` means *"the largest output this model can produce"* (a capability ceiling), not *"this much must be held back"* (a reservation). Treating the former as the latter inverts the meaning whenever a provider declines to publish a separate output cap.

Evidence that the model itself is fine — the same `grok-4.6` registered under GitHub Copilot with `maxTokens: 128000`:

```
required = 240,565 + 18,693 + 15,671 + 128,000 + 20,000 + 32,768 + 8,192
         = 463,889  <  500,000       ✅ switches without complaint
```

Affected `xai` catalog entries (3 of 4):

```
grok-4.3         ctx=1,000,000   max=    30,000
grok-4.5         ctx=  500,000   max=   500,000   <-- max == ctx
grok-4.6         ctx=  500,000   max=   500,000   <-- max == ctx
grok-build-0.1   ctx=  256,000   max=   256,000   <-- max == ctx
```

Notably `grok-4.3` — the *older* model — remains usable because it declares a modest 30K output cap. Users are pushed toward older models by this bug.

## Suggested fix

Treat `maxTokens >= contextWindow` as "unspecified" and fall back to a practical default instead of letting the 0.5 cap decide:

```js
const outputReserve = maxTokens >= contextWindow
    ? DEFAULT_OUTPUT_RESERVE   // e.g. 32_768
    : Math.min(maxTokens, Math.floor(contextWindow * MAX_OUTPUT_RESERVE_RATIO));
```

xAI's own guidance puts a realistic single-call output at **32K–64K** for this model, so the current 250K reservation over-provisions by roughly 4–8x.

Alternatively (or additionally), expose the output reserve as a user-overridable setting, since the right value is workload-dependent.

## Minor, related observation

`MODEL_SAFETY_MARGIN_PROFILES` in `model-usability-budget.js` has entries for `anthropic`, `openai`, `google`, and `deepseek`, but none for `xai` — grok therefore falls through to `default` (8,192), visible as `safetyMarginProfile: "default"` in the projection above. This is not the cause of this bug (a smaller margin only helps here), but a grok profile may be worth adding if its runway characteristics differ.

## Environment

- `omo-ai@5.0.0-0.beta.30`, bundled senpi
- Bun 1.4.1-canary.1+d393bf9e6, macOS arm64

MoerAI:
## What changed

Added supplemental model capability entries for xAI grok models where the bundled snapshot reports `output: contextWindow` (i.e. the output limit equals the full context window). The supplemental entries override those values with `output: 32768`, which is a realistic maximum generation limit for these models.

**Affected models:**
- `grok-4.5`, `xai/grok-4.5`, `x-ai/grok-4.5` (context: 500 000 → output capped at 32 768)
- `grok-4.6`, `xai/grok-4.6` (context: 500 000 → output capped at 32 768)
- `xai/grok-build-0.1`, `x-ai/grok-build-0.1` (context: 256 000 → output capped at 32 768)

## Why

OpenCode's `getPromptContextWindow` computes the usable prompt budget as `contextWindow - min(maxOutputTokens, contextWindow / 2)`. When `maxOutputTokens` equals `contextWindow` (as the bundled snapshot reports for these models), the formula reserves exactly half the context window for output — 250 000 tokens for a 500 000-token model — leaving only 250 000 tokens for the prompt. This matches the symptom reported in #7541.

`grok-4.3` is unaffected because its snapshot entry already has `maxTokens: 30000`.

The fix is entirely plugin-side: `SUPPLEMENTAL_MODEL_CAPABILITIES` entries win over the bundled snapshot via `getBundledModelCapabilitiesSnapshot()`, so no changes to the generated JSON or the OpenCode runtime are needed.

## Observed behavior

Before fix: `getModelCapabilities({ providerID: 'xai', modelID: 'xai/grok-4.6' }).maxOutputTokens === 500000`
After fix: `maxOutputTokens === 32768` (< 250 000, so the output reserve drops from 250 000 to 32 768 tokens)

## QA / evidence

- RED test added to `packages/model-core/src/model-capabilities-bundled-snapshot.test.ts` asserting `maxOutputTokens < contextWindow * 0.5` for all affected model IDs — confirmed failing before the fix.
- GREEN after adding supplemental entries: `2 pass, 0 fail`.
- Full `packages/model-core/` regression: `364 pass, 0 fail`.
- Typecheck (`bun run typecheck`): clean, no errors.

## Residual risk

The `output: 32768` value is based on xAI's documented generation limit for these models. If xAI raises the actual output limit in future, the supplemental entry can be updated or removed once the bundled snapshot reflects the correct value.

---
## Summary by cubic
Fixes #7541 by adding supplemental capability entries that cap the output limit for xAI `grok` models whose bundled snapshot reports `output` equal to the context window. Previously `getModelCapabilities` returned `maxOutputTokens` as the full context window, so `getPromptContextWindow` reserved half the context for output and left only half for the prompt.

**Affected models**

- `grok-4.5`, `xai/grok-4.5`, `x-ai/grok-4.5`, `grok-4.6`, `xai/grok-4.6`, `xai/grok-build-0.1`, and `x-ai/grok-build-0.1` now have `output` capped at 32,768.

<sup>Written for commit b5a757323b8287903c733d389c3ef1eb55ff06f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7624?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/model-core/src/model-capabilities-bundled-snapshot.test.ts
+++ b/packages/model-core/src/model-capabilities-bundled-snapshot.test.ts
@@ -4,6 +4,32 @@ import { getBundledModelCapabilitiesSnapshot, getModelCapabilities } from "./mod
 import bundledModelCapabilitiesSnapshotJson from "../../../packages/omo-opencode/src/generated/model-capabilities.generated.json"
 
 describe("bundled model capabilities snapshot", () => {
+  describe("#given xAI grok models where maxTokens equals contextWindow", () => {
+    test("#when resolving maxOutputTokens #then it is capped to a practical limit well below contextWindow", () => {
+      // given
+      const bundledSnapshot = getBundledModelCapabilitiesSnapshot(bundledModelCapabilitiesSnapshotJson)
+      const affectedModels = [
+        { providerID: "xai", modelID: "xai/grok-4.5", contextWindow: 500_000 },
+        { providerID: "xai", modelID: "xai/grok-4.6", contextWindow: 500_000 },
+        { providerID: "xai", modelID: "xai/grok-build-0.1", contextWindow: 256_000 },
+      ]
+
+      // when
+      const results = affectedModels.map(({ providerID, modelID, contextWindow }) => ({
+        modelID,
+        contextWindow,
+        caps: getModelCapabilities({ providerID, modelID, bundledSnapshot }),
+      }))
+
+      // then — maxOutputTokens must be strictly less than half the context window
+      // so OpenCode's getPromptContextWindow does not reserve half the window for output
+      for (const { modelID, contextWindow, caps } of results) {
+        expect(caps.maxOutputTokens).toBeDefined()
+        expect(caps.maxOutputTokens!).toBeLessThan(contextWindow * 0.5)
+      }
+    })
+  })
+
   test("keeps GPT-4.1 OpenAI variants marked as supporting tool calls", () => {
     // given
     const bundledSnapshot = getBundledModelCapabilitiesSnapshot(bundledModelCapabilitiesSnapshotJson)
--- a/packages/model-core/src/model-capabilities/supplemental-entries.ts
+++ b/packages/model-core/src/model-capabilities/supplemental-entries.ts
@@ -111,4 +111,109 @@ export const SUPPLEMENTAL_MODEL_CAPABILITIES: Record<string, ModelCapabilitiesSn
 			output: 128000,
 		},
 	},
+	"grok-4.5": {
+		id: "grok-4.5",
+		family: "grok",
+		reasoning: false,
+		temperature: true,
+		toolCall: true,
+		modalities: {
+			input: ["text", "image"],
+			output: ["text"],
+		},
+		limit: {
+			context: 500000,
+			output: 32768,
+		},
+	},
+	"xai/grok-4.5": {
+		id: "xai/grok-4.5",
+		family: "grok",
+		reasoning: false,
+		temperature: true,
+		toolCall: true,
+		modalities: {
+			input: ["text", "image"],
+			output: ["text"],
+		},
+		limit: {
+			context: 500000,
+			output: 32768,
+		},
+	},
+	"x-ai/grok-4.5": {
+		id: "x-ai/grok-4.5",
+		family: "grok",
+		reasoning: false,
+		temperature: true,
+		toolCall: true,
+		modalities: {
+			input: ["text", "image"],
+			output: ["text"],
+		},
+		limit: {
+			context: 500000,
+			output: 32768,
+		},
+	},
+	"grok-4.6": {
+		id: "grok-4.6",
+		family: "grok",
+		reasoning: false,
+		temperature: true,
+		toolCall: true,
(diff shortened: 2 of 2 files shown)


MoerAI:

PR #7638, opened 2026-09-02
perf(omo-senpi): keep session-start git probes off the startup stack

1Morganmore:
## What
`init-deep-advisor` preflight ran two synchronous `git` `execFileSync` probes on the `session_start` handler stack with no timeout. That blocks every interactive/print boot (and is unbounded if git wedges); on this Windows host each spawn also pays real-time AV scan cost.

## Change
- `runAdvisor` awaits an **async** preflight; git probes now use `child_process.execFile` (no longer sync, no longer on the handler stack — `session_start` stays fire-and-forget).
- Added an injectable `advisorPreflight` seam in `InitDeepAdvisorComponentDependencies` so the fake-timer unit test stays deterministic; production default is the git-probing preflight.
- Regenerated `plugin/extensions/omo.js` bundle.

## QA
- `bun test packages/omo-senpi/src/components/init-deep-advisor/component.test.ts` → **17/17 green**.
- Built bundle swap verified against the real harness (`script/qa/web-terminal-visual-qa.mjs`, real pty + xterm.js + Chrome, `OMO_TIMING=1 OMO_STARTUP_BENCHMARK=1`): boot completes with **no extension-load diagnostics**; `omo -p "Reply with exactly: OK"` returns `OK`.
- Startup timing before/after captured in `OMO_TIMING` benchmark evidence (baseline cold 12.2s → post-fix cold 3.5s on the same machine; environment factors documented in the PR thread if needed).

## Why it is enough
Behavior for git-repo users is unchanged (preflight detects the same states); only the blocking primitive became async. Unit coverage pins the component contract; live QA pins load/no-regression.

## STOP WHEN
CI green and one approval.

---
## Summary by cubic
Makes the init-deep-advisor preflight async so session-start no longer blocks on synchronous git probes, and hardens the probes so a wedged git can't stall startup.

- Replaces `execFileSync` with async `execFile` calls, each with a 5s timeout, keeping them off the startup stack.
- Adds an injectable `advisorPreflight` seam so the fake-timer unit test stays deterministic.
- The work-tree probe now requires git to output `true`, so bare repos and paths inside `.git` skip the advisor instead of showing a spurious warning; non-git-directory behavior is pinned by a new unit test.

<sup>Written for commit d45b4a7c4c122dc9535deb8d231062db167059c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7638?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Thanks for this. While triaging #8371 I checked this branch against current `dev`:

- `dev` already carries part of the intent: `gitIsRepo` is gone from the preflight, and the single remaining probe (`gitToplevel` → `git rev-parse --show-toplevel`) is bounded to 5 s with stderr contained (35cdb7f5d, 52df6c747). What is left of this PR's value is making that one probe async so it never sits on the `session_start` handler stack.
- The branch now conflicts with `dev` on all three files (`component.ts`, `component.test.ts`, the committed `omo.js` bundle).

If you want to carry the async version forward, a rebase onto `dev` keeping the injectable `advisorPreflight` seam and the `--is-inside-work-tree === "true"` check would still be welcome; the bundle should be regenerated on the rebased tree. #8371 itself (`omo --help` cost) is fixed on the engine side in code-yeongyu/senpi#1758 and does not depend on this PR.

code-yeongyu:

I'm closing it as **superseded**: the residual value the maintainer identified on 2026-09-16 — keeping the one remaining `git rev-parse --show-toplevel` probe off the `session_start` handler stack — landed the next day via PR #8421 (*perf(omo-senpi): move plugin startup work past the first paint*, merged 2026-09-17, commit `79d118c48`): on `dev`, `init-deep-advisor/component.ts` now runs the whole advisor, preflight and probe included, inside `deferUntilAfterFirstPaint(...)`, and the probe itself is bounded to 5 s with stderr contained (`git-helpers.ts`).

**How I checked:** `gh pr view 8421` → `MERGED 2026-09-17T10:53:52Z`; `component.ts:53` wraps `runAdvisor` (which calls `advisorPreflight` → `gitToplevel`) in `deferUntilAfterFirstPaint`; `git-helpers.ts:18` `timeout: 5000`.

PR #7641, opened 2026-09-02
test(omo-senpi): budget Windows memfs git cases

MoerAI:
## Summary

The memfs-extra.test.ts integration suite drives real git subprocesses but still used the Bun 5 second default on Windows. CI run 33600175911 timed out the interactive restore case after 5.697 seconds.

This mirrors the existing platform budget from the sibling memfs.test.ts suite: 30 seconds on Windows and the unchanged 5 seconds elsewhere. No restore operation, assertion, sleep, retry, polling loop, or platform skip changed.

## Verification

- Focused interactive restore: 1 pass, 0 fail, 4 expectations.
- Both memfs suites: 26 pass, 0 fail, 84 expectations.
- packages/omo-senpi typecheck: pass with no diagnostics.
- Real Senpi adapter driver: PASS; ultrawork injected, comment checker passed, real Senpi and OMO homes untouched, sandbox removed.

## Evidence

.omo/evidence/omo-senpi-adapter/20260902-senpi-memfs-windows-budget/README.md

## Risk

Test-only platform timeout configuration. Windows PR CI is the authoritative rerun of the original failure.


MoerAI:
Closing as part of a scoped review of the Windows CI flake work. `dev` is currently green and this PR is not blocking, so it is being closed rather than left to rot. The branch is preserved and the analysis below is written so the work can be resumed directly.

These PRs were assessed against a five-gate standard for distinguishing a real determinism fix from a mask: (1) name the production invariant; (2) prove it via a state transition, event, handle closure, PID death, or durable postcondition - not sleep duration, call count, or "no error before N ms"; (3) still fail under a negative control; (4) assert cleanup AND next-operation liveness; (5) survive repeated stress at unchanged semantics.

#7641 - "test(omo-senpi): budget Windows memfs git cases"
Verdict: classified as a mask, not a fix.
Gap: gates 3 through 5. `packages/omo-senpi/src/components/memory/commands/memfs-extra.test.ts:21-24` only raises the whole file's Windows timeout from 5 to 30 seconds. It adds no new state observation, no failure injection, and no liveness proof. This PR is also currently red on `senpi-compatibility (windows-latest)`.
To resume: replace the blanket timeout with observation of the actual restore phases - selected backup, exact restored bytes, safety backup, git child exit, handle closure, no temporary root left - plus a bounded negative control that holds the git child and fails at a named phase.

PR #7644, opened 2026-09-02
refactor: remove the CodeGraph integration

code-yeongyu:
## What changed

CodeGraph is removed from the product. It shipped as a bundled MCP server wired into all three harnesses, and this PR deletes the feature rather than deprecating it: no compatibility shim, no renamed remnant, and no test pinning its absence.

Removed surfaces:

- **Codex plugin** — the `components/codegraph` workspace (with its bundled Node runtime), the SessionStart bootstrap hook, the PostToolUse init-guidance hook, the `.mcp.json` server entry, the `omo-codegraph` bin, and the `migrate-omo-sot` migration that existed only to move `*_CODEGRAPH_*` env vars into config.
- **OpenCode plugin** — the built-in `codegraph` MCP, the `codegraph-bootstrap` hook, the config schema block, the tool-result init guidance, and the process-sweep family for its orphaned daemons.
- **Shared packages** — `packages/utils/src/codegraph/**`, the legacy `validateOmoConfig` codegraph validator, and the `codegraph` block in `omo-config-core`'s schema.
- **Installer** — the `codegraphMcpEnabled` gate, the MCP path stamping, and the Node-runtime resolution.
- Prompts, skills, docs, READMEs (5 languages), CI shards, third-party notices, and the tracked `.codegraph/.gitignore`.

Generated artifacts are **regenerated from source**, never hand-edited: the two JSON schemas, the bundled Codex installer, the plugin `package-lock.json`, the synced skill/directive copies, and the Senpi extension bundles.

## Why the loader change ships with this

`OmoConfigLayerSchema` is `.strict()`, and the loader dropped an **entire config layer** when it hit one unrecognized key. Removing the `codegraph` key would therefore have silently reset every category, agent, and task setting for anyone whose `omo.json` still carried it — which is the common case, since the scaffold used to write a `codegraph` block.

So `06417b2c`/`2d59325` makes the loader strip unrecognized keys, re-validate, and report one `unknown-keys` diagnostic naming each ignored dotted path. Malformed values still reject the layer, unchanged.

**Prototype-pollution input stays fail-closed.** This was a real regression I caught mid-work: tolerance initially made a layer containing `__proto__` load partially instead of being rejected. Note that a JSON `__proto__` member is written *through* the prototype rather than becoming an own property, so zod reports the injected payload's inner keys and never `__proto__` itself — a guard matching the key name never fires. The guard inspects the parsed value's prototype chain directly.

## Observed behavior / QA

Verified by me on `fc5f1a621`, not relayed from a subagent:

- **Sweep (the deliverable):** `rg -il codegraph` over the repo (excluding `.git`, `node_modules`, `CHANGELOG.md`, lockfiles, `local-ignore`, and `.omo/evidence`) → **no output**. Was 241 files.
- `bun test packages/omo-config-core` → **205 pass / 0 fail**
- `bun test packages/utils` → **398 pass / 0 fail**
- `bun test packages/omo-opencode` (scoped suites) → **2045 pass / 0 fail** after fixing the prototype-pollution regression
- `bun run --cwd packages/omo-codex/plugin test` → **323 pass / 0 fail**
- `bun test packages/omo-codex/src packages/omo-config-core` post-rebase → **538 pass / 1 skip / 0 fail**
- `bun run build:schema` → both assets **byte-identical** to the checked-in versions (proving regeneration, not hand-editing)
- **Config compatibility, real surface:** `loadOmoConfig` against a copy of a real `~/.omo/omo.jsonc` carrying `codegraph` at the root, inside a profile, and inside `[codex]` → layer **loads** with all 9 categories, 4 agents, and `task` settings intact, plus `[unknown-keys] Ignored unknown keys ...: [codex].codegraph, codegraph`.
- **Mutation-tested, because a green test proves nothing on its own:** neutering the `unrecognized_keys` branch flips 2 tests to failing; neutering the prototype-tamper detector, or passing the `toRecord` copy instead of the raw parsed value, each flip 2 to failing. Restoring returns all green.

Full gates (typecheck, root `bun test`, build, notices, plugin) are running on a separate machine against this exact head; I will post the result before merging.

## Residual risk

- `.omo/evidence/20260727-codegraph-*/` is **kept on purpose** — dated records of QA that actually happened, same class as `CHANGELOG.md`. Deleting them would falsify the audit trail.
- A pre-existing `omo-codegraph` bin shim from an older install is no longer in the cleanup list, so an upgrade will not reap it. Called out rather than reintroducing the name to clean it.
- The plugin `package-lock.json` was regenerated with npm, which is what `npm ci` in CI requires; everything else stays bun.

Companion PR: code-yeongyu/senpi#1281 removes the four incidental references there.

---
## Summary by cubic
Removes the CodeGraph integration from the Codex plugin, OpenCode plugin, and Senpi adapter, plus the shared config, installer, and docs. The feature is deleted outright—no deprecation shim or leftover references—and the config loader no longer drops an entire `omo.json` layer when it hits an unrecognized key.

- Existing configs that still carry the retired `codegraph` key keep loading; the loader reports it as an ignored unknown key.
- Unrecognized keys are stripped, the layer re-validated, and one `unknown-keys` diagnostic names each ignored dotted path. Malformed values still reject the layer.
- The merge rules drop the codegraph-only `excluded_roots` union special-case, so arrays follow the normal replace rule.
- Prototype-pollution input stays fail-closed: the guard inspects the parsed value's prototype chain directly, since JSON `__proto__` members don't become own properties.
- Generated artifacts (schemas, bundled installer, plugin lockfile, synced skill/directive copies, Senpi bundles) are regenerated from source, not hand-edited.

<sup>Written for commit b941e9f345aebb5df7a634e7c4fa92c685837eb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7644?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


louisremi:
I couldn't find any explanation why you decided to remove codegraph. I'm really curious to know why!

Issue #7650, opened 2026-09-02
[Bug] Claude Fable 5.1 fails because OMO advertises Claude Code 2.1.75


bokjk:
### Problem

`anthropic/claude-fable-5-1` fails in OMO Native and falls back to `anthropic/claude-opus-5`. The fallback also fails with:

```text
Claude Code 2.1.75 does not support this model; version 2.1.251 or newer is required.
error_code: claude_code_version_too_old
```

The standalone Claude Code installation is already current:

```text
2.1.258 (Claude Code)
```

### Root cause

The published `omo-ai@5.0.0-0.beta.34` package pins:

```json
"@code-yeongyu/senpi": "2026.9.2"
```

The published Senpi tarball still contains this in `package/node_modules/@earendil-works/pi-ai/dist/api/anthropic-messages.js`:

```js
const claudeCodeVersion = "2.1.75";
```

`omo-ai`'s `bin/senpi-patch.mjs` does not patch this value, so OAuth requests advertise `User-Agent: claude-cli/2.1.75` regardless of the locally installed Claude Code version.

Upstream pi already fixed this by changing `2.1.75` to `2.1.251`:

https://github.com/earendil-works/pi/commit/96317e50b8d6e7f6d0e47fd29122baf1461c00f5

The fix is not present in Senpi `2026.9.2`, and therefore is not included in OMO Native beta.34.

### Verification

After locally changing the packaged value to `2.1.258`, a fresh OMO process successfully completed the same model request:

```text
PATCHED_OK
exit code: 0
```

### Suggested fix

1. Sync or cherry-pick upstream commit `96317e50` into Senpi.
2. Publish a new `@code-yeongyu/senpi` version.
3. Update the pinned Senpi dependency in OMO Native.
4. Add a regression test for Claude Fable 5.1 / Opus 5 OAuth requests.

Longer term, deriving this value from the bundled SDK or a shared version source would avoid future breakage whenever Anthropic raises the minimum supported Claude Code version.

---

### Problem

When using `anthropic/claude-fable-5-1` in OMO Native, the request fails and falls back to `anthropic/claude-opus-5`. The fallback request also fails with the following error:

```text
Claude Code 2.1.75 does not support this model; version 2.1.251 or newer is required.
error_code: claude_code_version_too_old
```

The Claude Code installed on the system is already `2.1.258`.

### Root cause

[...]

Issue #1312, opened 2026-01-31
[Bug]: On macos, after installing opencode and oh-my-opencode, opencode cannot be started


fishANDdog:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-opencode
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme)

### Bug Description

On macos, after installing opencode and oh-my-opencode, opencode cannot be started

### Steps to Reproduce

On macos, after installing opencode and oh-my-opencode, opencode cannot be started

### Expected Behavior

On macos, after installing opencode and oh-my-opencode, opencode cannot be started

### Actual Behavior

On macos, after installing opencode and oh-my-opencode, opencode cannot be started

### Doctor Output

```shell
nothing output
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

On macos, after installing opencode and oh-my-opencode, opencode cannot be started

### Operating System

macOS

### OpenCode Version

1.1.47

EricSolshkov:
opencode log related to omo:

```plaintext
INFO  2026-01-31T03:43:42 +235ms service=plugin path=oh-my-opencode loading plugin
INFO  2026-01-31T03:43:42 +0ms service=bun pkg=oh-my-opencode version=latest installing package using Bun's default registry resolution
INFO  2026-01-31T03:43:42 +1ms service=bun cmd=["C:\\Path\\to\\OpenCode\\opencode-cli.exe","add","--force","--exact","--cwd","C:\\Path\\to\\.cache\\opencode","oh-my-opencode@latest"] cwd=C:\Path\\to\.cache\opencode running
```
the opencode crushes here, with no more log output.

code-yeongyu:
This might be related to the TUI freeze bug fixed in v3.1.10. Please try updating:

```bash
bunx oh-my-opencode@3.1.10 install
```

If the issue persists, please provide more details:
1. What exactly happens when you run `opencode`? (blank screen, error message, crash?)
2. Can you share the output of `opencode --print-logs`?
3. What's in your `~/.config/opencode/opencode.json`?

Related fix: PR #1304 (v3.1.10) resolved a deadlock during plugin initialization that caused blank screens on startup.

[...]

Issue #1310, opened 2026-01-31
[Feature]: Please provide support for OpenCode Desktop.


martjay:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-opencode/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues and discussions to avoid duplicates
- [x] This feature request is specific to oh-my-opencode (not OpenCode core)
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-opencode#readme)

### Problem Description

Yesterday, I had AI automatically modify my OpenCode config file for me, but since then, the OpenCode application has been stuck on the launch screen. I hope you can provide support for the OpenCode Desktop version—that would be the biggest benefit.

### Proposed Solution

Please provide support for OpenCode Desktop.

### Doctor Output (Optional)

```shell

```

### Feature Type

Configuration Option

### Contribution

- [ ] I'm willing to submit a PR for this feature
- [ ] I can help with testing
- [ ] I can help with documentation

PR #7669, opened 2026-09-03
Fix #7626: prefer a newer standalone Claude Code over a stale bundled claude-agent-sdk binary

### LINKED ISSUE #7626: [Bug]: claude-sdk-oauth silently answers from a fallback model when the bundled claude-agent-sdk binary does not know the requested model id (claude-fable-5-1)
### Summary

On `omo-ai 5.0.0-0.beta.31` (engine senpi 2026.8.31), selecting a Claude model that the **bundled** Claude Code native binary does not recognize (here `claude-sdk-oauth/claude-fable-5-1`) fails the query, and with the default retry-fallback enabled the run **silently answers from a different model** with no user-visible indication.

The visible transcript is indistinguishable from a successful run on the requested model. The substitution is only discoverable by reading the session JSONL.

### Environment

- omo `5.0.0-0.beta.31` (engine: senpi 2026.8.31)
- macOS darwin-arm64, node v26.7.0, global npm install
- Provider: `claude-sdk-oauth` (OAuth login, working — control run below)
- Bundled binary: `@anthropic-ai/claude-agent-sdk-darwin-arm64` **0.3.241**
- Standalone Claude Code also installed: `2.1.251`

### Reproduction

```bash
omo --model claude-sdk-oauth/claude-fable-5-1 --no-tools -p "Reply with exactly: FABLE_5_1_OK"
```

Output:

```
Warning: Model "claude-fable-5-1" not found for provider "claude-sdk-oauth". Using custom model id.
FABLE_5_1_OK
```

Looks like a success on `claude-fable-5-1`. It is not. The session JSONL records:

```json
{"type":"model_change","provider":"claude-sdk-oauth","modelId":"claude-opus-5",
 "reason":"fallback","originalProvider":"claude-sdk-oauth","originalModelId":"claude-fable-5-1"}
```

The reply came from `claude-opus-5`.

### What actually happens without fallback

```bash
omo --no-model-fallback --model claude-sdk-oauth/claude-fable-5-1 --no-tools --mode json -p "..."
```

```json
{"role":"assistant","provider":"claude-sdk-oauth","model":"claude-fable-5-1",
 "usage":{"totalTokens":0},"stopReason":"error","errorMessage":"unknown",
 "diagnostics":[{"type":"claude_sdk_oauth_session_continuity",
   "details":{"kind":"flatten","reason":"registry_miss"}},
  {"type":"claude_sdk_oauth_session_continuity",
   "details":{"kind":"flatten","reason":"query_failed"}}]}
```

Zero tokens, `stopReason: error`, `errorMessage: "unknown"`.

Control run on the same lane, same flags, proving auth/lane are healthy:

```bash
omo --no-model-fallback --model claude-sdk-oauth/claude-opus-5 --no-tools -p "Reply with exactly: CONTROL_OK"
# -> CONTROL_OK, exit 0
```

### Root cause

The `claude-sdk-oauth` lane spawns the native `claude` binary bundled as `@anthropic-ai/claude-agent-sdk-darwin-arm64` (0.3.241 here). That build does not know `claude-fable-5-1`, so the query dies upstream. Standalone Claude Code 2.1.251 reports the same id as unrecognized but is a newer build that does serve it:

```
[claude-code:unrecognized_model] {"model":"claude-fable-5-1","query_source":"sdk"}
```

Pointing the lane at the newer executable makes the model work, no code or config change:

```bash
CLAUDE_CODE_EXECUTABLE=/path/to/claude \
  omo --no-model-fallback --model claude-sdk-oauth/claude-fable-5-1 --no-tools -p "Reply with exactly: PROOF51"
```

Session record for that run:

```json
{"role":"assistant","provider":"claude-sdk-oauth","model":"claude-fable-5-1",
 "stopReason":"stop","usage":{"totalTokens":14724}}
```

`resolveClaudeCodeExecutable` reads `CLAUDE_CODE_EXECUTABLE` before the bundled candidates, so this is the intended escape hatch — it just is not discoverable from the error.

### Secondary issue: models.json cannot patch this provider

Adding the model under `~/.omo/agent/models.json` for `claude-sdk-oauth` has no effect. In `composeModelProvider`, `applyExtension` **replaces** the composed list with the extension's own `models` array whenever the extension declares one:

```js
function applyExtension(providerId, models, config) {
    if (!config) return [...models];
    if (!config.models) { ... }
    return config.models.map((definition) => { ... });   // replaces, not merges
}
```

`claude-sdk-oauth` declares `models: MODELS` in its registration, so user-defined models.json entries for that provider are discarded. The id never appears in `--list-models` or the Ctrl+P picker, and `modelOverrides` cannot help because it only patches models that already exist. So there is no config-only way for a user to keep up with a new Claude model — only explicit `--model` selection, which then hits the bundled-binary failure above.

### Impact

1. **Silent wrong-model answers.** A user who asks for model A and reads a normal-looking answer has no signal that model B produced it. Cost, capability, and reasoning-level assumptions are all wrong. This is the serious part.
2. `errorMessage: "unknown"` gives the user nothing to act on when fallback is off.
3. New Claude models are unusable until both the catalog and the bundled SDK binary are bumped, with no user-side workaround surfaced.

### Suggested fixes

1. Surface fallback substitution in the visible output, not just the session file — e.g. `Model claude-fable-5-1 failed (unknown); answered with claude-opus-5 instead`. At minimum warn once per run.
2. Map the SDK's `unrecognized_model` / `query_failed` case to an actionable message naming `CLAUDE_CODE_EXECUTABLE` and the bundled SDK version.
3. Let `models.json` **merge** custom models into extension-declared provider catalogs (upsert by id, as `applyModelsJson` already does for the builtin layer) instead of replacing them, so users can register a new Claude id without waiting for a release.

ashmoonori-afk:
Fixes the root cause of #7626 for the `omo-ai` distribution: a `claude-sdk-oauth` model id that the **bundled** `@anthropic-ai/claude-agent-sdk` native binary does not know fails the query with `stopReason: "error"` / `errorMessage: "unknown"` / zero tokens, and the retry fallback then answers from a *different* model with nothing on screen saying so.

## Why the launcher is the right seam

`omo-ai` pins `@code-yeongyu/senpi`, which pins `@anthropic-ai/claude-agent-sdk` **exactly** (`0.3.241`, with per-platform native binaries as pinned optional deps). A dependency-side `overrides` block is ignored by npm for a non-root package, so this package cannot pull a newer SDK build — but it *can* stop handing the engine a binary that is too old for the ids the model catalog advertises. `resolveClaudeCodeExecutable` already reads `CLAUDE_CODE_EXECUTABLE` before the bundled candidates; this change makes the launcher answer that variable when, and only when, it is safe to.

This is the same shape as the existing `claudeCodeVersionFloor` in `bin/senpi-patch.mjs` (the #7650 fix): a version floor that no-ops the moment the pinned engine ships a build at or above it.

## What changed

New `bin/lib/claude-code-executable.js`, wired into `senpiEnvironment()` in `bin/lib/launcher.js`. It returns a path only when **all** of these hold:

1. the user has not set `CLAUDE_CODE_EXECUTABLE` (an explicit value always wins);
2. the engine's bundled `@anthropic-ai/claude-agent-sdk` is **below `0.3.257`**;
3. a standalone `@anthropic-ai/claude-code` is installed beside this package at **`2.1.257` or newer**;
4. its `bin.claude` binary actually exists on disk.

Anything else — unreadable manifest, missing package, malformed JSON, absent binary — returns `undefined` and the engine spawns exactly as before. Resolution is fail-open by contract; it never gates the spawn.

## How the floor was pinned

Not guessed — string-scanned from the published win32-x64 native binaries:

```text
@anthropic-ai/claude-agent-sdk-win32-x64@0.3.252  ->  claude-fable-5-1 ABSENT
@anthropic-ai/claude-agent-sdk-win32-x64@0.3.257  ->  claude-fable-5-1 present
```

`0.3.253`–`0.3.256` were never published, so `0.3.257` is the exact first build carrying the id. Standalone Claude Code ships the same binary under a lockstep patch number (`claude-code 2.1.N` ↔ `claude-agent-sdk 0.3.N`; locally `claude-code@2.1.259/bin/claude.exe` is 219,715,232 bytes against `claude-agent-sdk-win32-x64@0.3.259`'s 219,715,808-byte unpacked size), which is why the standalone floor is `2.1.257`.

The bundled `0.3.241` binary does know `claude-opus-5`, `claude-sonnet-5` and `claude-mythos-5` — the gap is specific to ids published after that build.

## Verification

`bun test packages/omo-native/test/claude-code-executable.test.ts` — 8 pass, covering: stale bundled + newer standalone → path returned; explicit env value → untouched; bundled at floor → untouched; standalone missing → untouched; standalone below floor → untouched; standalone exactly at floor → returned; unreadable bundled manifest → fail-open; malformed standalone manifest → fail-open.

Real-surface check against a live global install (Windows 10.0.26200, node v24.13.0, `omo-ai@5.0.0-0.beta.35`, engine `senpi@2026.9.2-4`, bundled SDK `0.3.241`, standalone `claude-code@2.1.259`):

```text
claudeCodeExecutableOverride({ env: {}, senpiRoot, root })
  -> C:\Users\lg\AppData\Roaming\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe
claudeCodeExecutableOverride({ env: { CLAUDE_CODE_EXECUTABLE: "/x" }, ... })
  -> undefined
```

That resolved path is exactly the binary this behavior was measured against on the same box:

| model | executable | fallback | rc | secs | outcome |
|---|---|---|---|---|---|
| `claude-fable-5-1` | bundled 0.3.241 | off | 124 | 152 | `stopReason: error`, 0 tokens, then the process hangs past `agent_end` |
| `claude-fable-5-1` | standalone 2.1.259 | off | 0 | 116 | `stopReason: stop`, 19387 tokens, correct answer |
| `claude-fable-5-1` | bundled 0.3.241 | on | 0 | 118 | normal-looking answer — silently produced by another model |
| `claude-fable-5` | bundled 0.3.241 | on | 0 | 42 | ok (control) |
| `claude-opus-5` | bundled 0.3.241 | on | 0 | 45 | ok (control) |

Full diagnostic capture is in [this comment on #7626](https://github.com/code-yeongyu/oh-my-openagent/issues/7626#issuecomment-5519204488).

`bun test packages/omo-native/test` on this branch: 168 pass / 34 skip / 8 fail — the 8 failures are identical on pristine `dev` in the same worktree (they need an installed workspace and a staged `plugin/` payload, neither of which this change touches).

## Scope / what this does not fix

- Silent fallback substitution is still invisible when no standalone is installed. Suggestions 1 and 2 in the issue body (surface the substitution; map `query_failed` to an actionable message naming `CLAUDE_CODE_EXECUTABLE`) live in the engine and are untouched here.
- On Windows the failed query also leaves the process alive after `agent_end`; that hang is an engine-side issue and is not addressed by this PR.
- Users with a stale bundled SDK and no standalone Claude Code see no behavior change at all.

---
## Summary by cubic
Fixes #7626: queries for Claude model ids the bundled `@anthropic-ai/claude-agent-sdk` binary does not know fail with `errorMessage: "unknown"` and the retry fallback silently answers from a different model. The launcher now points `CLAUDE_CODE_EXECUTABLE` at a newer standalone `@anthropic-ai/claude-code` when one is installed, so the selected model actually answers.

**Behavior**
- Sets `CLAUDE_CODE_EXECUTABLE` only when the bundled SDK is below `0.3.257` and a standalone `@anthropic-ai/claude-code` at `2.1.257` or newer is installed with its `bin.claude` binary present.
- An explicit user-set value always wins, and every resolution failure is fail-open, so the engine spawns exactly as before.

<sup>Written for commit 3b03315fa0c43c92bb660c2592aa658d866617d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7669?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-native/AGENTS.md
+++ b/packages/omo-native/AGENTS.md
@@ -10,6 +10,13 @@ omo-senpi plugin payload produced by `bun run build:omo-native` (gitignored, nev
 - brand: the launcher injects a `SENPI_BRAND` profile (name, `~/.omo/agent` home, `OMO_*` env prefix, wire identity, omo-ai beta update channel) so the pinned engine presents as omo; `--version` and every self-update spelling are answered by the launcher. See `docs/reference/omo-ai-publishing.md`.
 - `bin/lib/` - launcher modules:
   - `launcher.js` — `runLauncher()` dispatch, senpi environment/brand/update routing
+  - `claude-code-executable.js` — `claudeCodeExecutableOverride()`: the engine's claude-sdk-oauth lane
+    runs the native binary bundled with its pinned `@anthropic-ai/claude-agent-sdk`, and a build below
+    `BUNDLED_SDK_FLOOR` does not know the newer Claude model ids the catalog advertises, so those queries
+    fail with `errorMessage: "unknown"` and the retry fallback silently answers from another model. When
+    the bundled build is stale AND a newer standalone Claude Code is installed beside this package, the
+    launcher points `CLAUDE_CODE_EXECUTABLE` at it. An explicit user value always wins, and every
+    resolution failure is fail-open — this never gates the engine spawn.
   - `agent-dir.js` — `canonicalAgentDir()`, `adoptLegacyFlatState()`, legacy flat-dir migration
   - `setup-detect.js` / `setup-import.js` / `setup-models.js` / `setup-report.js` — harness detection, SQLite read-only import, provider mapping, report rendering
   - `setup-detect-cache.js` / `setup-detect-refresh.js` — the interactive launch's setup-suggestion cache: a
--- a/packages/omo-native/bin/lib/claude-code-executable.js
+++ b/packages/omo-native/bin/lib/claude-code-executable.js
@@ -1,0 +1,98 @@
+import { existsSync } from "node:fs"
+import { basename, dirname, join } from "node:path"
+import { packageRoot, readJson } from "./package-paths.js"
+
+// The engine's claude-sdk-oauth lane runs the native `claude` binary bundled with the pinned
+// @anthropic-ai/claude-agent-sdk. Builds below this floor do not carry the newer Claude model ids:
+// 0.3.252 has no `claude-fable-5-1` string, 0.3.257 does. Asking a build for an id it does not know
+// fails the query with `stopReason: "error"`, `errorMessage: "unknown"` and zero tokens, which the
+// retry fallback then hides by answering from a different model - the user reads a normal-looking
+// answer produced by a model they did not select.
+export const BUNDLED_SDK_FLOOR = "0.3.257"
+// Standalone Claude Code ships the same native binary under a lockstep patch number
+// (claude-code 2.1.N carries the claude-agent-sdk 0.3.N build), so this is the same floor.
+export const STANDALONE_CLAUDE_CODE_FLOOR = "2.1.257"
+
+function parseVersion(value) {
+  const match = /^(\d+)\.(\d+)\.(\d+)/.exec(typeof value === "string" ? value : "")
+  return match === null ? undefined : [Number(match[1]), Number(match[2]), Number(match[3])]
+}
+
+function isBelow(version, floor) {
+  const left = parseVersion(version)
+  const right = parseVersion(floor)
+  if (left === undefined || right === undefined) return false
+  for (let index = 0; index < 3; index += 1) {
+    if (left[index] !== right[index]) return left[index] < right[index]
+  }
+  return false
+}
+
+function enclosingNodeModules(host) {
+  let current = dirname(host)
+  while (basename(current) !== "node_modules") {
+    const parent = dirname(current)
+    if (parent === current) return undefined
+    current = parent
+  }
+  return current
+}
+
+// Both the nested layout (dependency under the host package) and the hoisted one (host package and
+// dependency as siblings of the same node_modules), which is what a global install produces.
+function packageManifestPaths(host, name) {
+  const segments = [...name.split("/"), "package.json"]
+  const nodeModules = enclosingNodeModules(host)
+  const paths = [join(host, "node_modules", ...segments)]
+  if (nodeModules !== undefined) paths.push(join(nodeModules, ...segments))
+  return paths
+}
+
+function readVersion(paths, exists, readManifest) {
+  for (const path of paths) {
+    if (!exists(path)) continue
+    const version = readManifest(path).version
+    if (typeof version === "string") return { version, path }
+  }
+  return undefined
+}
+
+function binaryFromManifest(manifestPath, manifest, exists) {
+  const bin = manifest.bin
+  const relative = typeof bin === "string" ? bin : typeof bin?.claude === "string" ? bin.claude : undefined
+  const candidates = relative === undefined ? [] : [join(dirname(manifestPath), relative)]
+  for (const name of ["claude.exe", "claude"]) candidates.push(join(dirname(manifestPath), "bin", name))
+  return candidates.find((candidate) => exists(candidate))
+}
+
+/**
+ * The `CLAUDE_CODE_EXECUTABLE` value to hand the engine, or `undefined` to leave it alone.
+ *
+ * Only answers a path when the bundled SDK build is too old to know current Claude model ids AND a
+ * newer standalone Claude Code is already installed beside this package. An explicit user value
+ * always wins, a bundled build at or above the floor is left untouched, and every failure is
+ * fail-open: this is a launcher convenience, never a gate on spawning the engine.
+ */
+export function claudeCodeExecutableOverride(options = {}) {
+  const {
+    env = process.env,
+    senpiRoot,
+    root = packageRoot,
+    exists = existsSync,
+    readManifest = readJson,
+  } = options
+  try {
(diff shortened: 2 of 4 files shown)


code-yeongyu:

I'm closing it because the issue it fixes, #7626, was resolved on the engine side and closed by the maintainer on 2026-09-03: senpi PR #1310 (`7df28bad8`) pins `@anthropic-ai/claude-agent-sdk` 0.3.259 so `claude-fable-5-1` works without `CLAUDE_CODE_EXECUTABLE`, and senpi PR #1312 (`221e64067`) surfaces the SDK's API error text and makes `is_error` results trigger fallback — both shipped in senpi 2026.9.3-2. With the bundled SDK current, preferring a standalone Claude Code binary is no longer needed.

**How I checked:** `gh issue view 7626` → `CLOSED COMPLETED 2026-09-03T07:01:05Z`, closing comment by the maintainer cites both senpi PRs.

PR #7706, opened 2026-09-03
fix: hop runtime_fallback on OpenCode UnknownError

### LINKED ISSUE #7702: runtime_fallback never hops OpenCode UnknownError (HTTP-code-only retry_on_errors)
## Summary
runtime_fallback.retry_on_errors only matches numeric HTTP statuses. OpenCode often wraps gateway failures as UnknownError with message Unexpected server error or provider text like authentication expired. Those never walk agents.*.models[].

## Repro
1. Configure model_fallback true and runtime_fallback.enabled with retry_on_errors including 401,429,500,502,503.
2. Set agents.sisyphus.models to [gateway-a/model, gateway-b/model].
3. Make the first model fail as OpenCode UnknownError (no HTTP status on the session error object).

## Actual
The session dies on the first model. No hop.

## Expected
Classified provider failures (401/403/429/5xx, auth-expired, rate-limit, overloaded) hop to the next models[] entry even when the host wraps them as UnknownError.

## Suggested fix
In the runtime-fallback error classifier, also match error.name UnknownError plus message/status in error.data, and string patterns already used for RateLimit/Authentication in model-core RETRYABLE_ERROR_NAMES. Numeric retry_on_errors should still work when a status is present.

RaviTharuma:
Keep runtime_fallback hopping when OpenCode wraps provider failures as UnknownError.

Fixes #7702

## Why
`retry_on_errors` only matched numeric HTTP codes. OpenCode often wraps 401/expired-auth as UnknownError with no status, so models[] never hopped.

## Changes
- Treat UnknownError / authentication-expired strings as retryable in `isRuntimeFallbackRetryableError`
- Tests on the shipped classifier

## Test
`bun test` on the classifier tests (16 pass)


RaviTharuma:
@code-yeongyu you confirmed the root cause on #7702 (`retry_on_errors` only matches numeric HTTP codes while OpenCode wraps failures as `UnknownError`). This PR implements the classifier change you described and is ready for review.

Scoping note carried over from the issue: `UnknownError` is matched together with known provider failure patterns rather than unconditionally, so a genuine internal bug still fails fast instead of burning the whole chain. Say the word if you want it widened.

PR #7707, opened 2026-09-03
fix: throw HephaestusRequiresGptError for non-GPT primaries

### LINKED ISSUE #7704: Hephaestus non-GPT primary fails as opaque UnknownError
## Summary
The Hephaestus GPT-only constraint (no-hephaestus-non-gpt) is reasonable. When a user sets Hephaestus primary to a non-GPT catalog id, the run fails as OpenCode UnknownError / unexpected server error instead of a typed, actionable message.

## Repro
1. Set agents.hephaestus.models[0] to a non-GPT id (for example a Grok catalog id).
2. opencode run --agent hephaestus ping (using the remapped display name).

## Actual
Opaque UnknownError. No mention that Hephaestus requires a GPT-family model.

## Expected
A typed plugin error, e.g. Hephaestus requires a GPT-family model (got provider/model-id). Do not silently delete the GPT-only policy.

## Suggested fix
Fail in the hook with a stable error name/message before the request hits the provider. Keep the constraint.

RaviTharuma:
Fail Hephaestus non-GPT primaries with a typed error instead of an opaque host UnknownError.

Fixes #7704

## Why
The GPT-only constraint is intentional. Headless `opencode run` currently dies as UnknownError with no mention of Hephaestus or GPT.

## Changes
- Throw `HephaestusRequiresGptError` from the hook
- Tests for the typed error

## Test
`bun test` on the Hephaestus tests (5 pass)


RaviTharuma:
@code-yeongyu per your call on #7704 — typed, actionable error thrown from the hook before the request reaches the provider, GPT-only constraint itself unchanged. Ready for review.

PR #7708, opened 2026-09-03
fix: hidden config-key agent aliases and fail-closed category --agent

### LINKED ISSUE #7701: opencode run --agent config-key silently uses default after display-name remap
## Summary
`opencode run --agent <config-key>` does not resolve Oh My OpenAgent display-name remaps. After the plugin remaps `sisyphus` to `Sisyphus - ultraworker` (and similarly prometheus/atlas/hephaestus), the CLI looks up the original config key, does not find it, and silently runs `default_agent`.

## Repro
1. Enable oh-my-openagent so agent keys are remapped via AGENT_DISPLAY_NAMES.
2. Run: opencode run --agent sisyphus ping
3. Run: opencode run --agent prometheus ping

## Actual
Session banner/agent id is the default (often Sisyphus - ultraworker) even when --agent prometheus was passed. No hard error.

## Expected
--agent sisyphus and --agent "Sisyphus - ultraworker" both address the same agent. Unknown names fail loudly with the registered names listed.

## Suggested fix
resolveRunnableRunAgent / getAgentConfigKey should accept both config keys and display names before OpenCode exact-key lookup. Do not fall through to default_agent without a warning.

### LINKED ISSUE #7703: Categories are not addressable via opencode run --agent
## Summary
OMO categories (quick, deep, ultrabrain, unspecified-high, and others) are used by task(category=...) but are not first-class opencode run --agent targets. Passing a category name either falls through to default Sisyphus or looks like a successful category run.

## Repro
opencode run --agent quick ping

## Actual
Default Sisyphus handles the prompt. No indication that quick is not a CLI agent.

## Expected
Either document that categories are task()-only and reject --agent quick with a message that lists runnable names vs category names, or optionally register category names as runnable aliases.

## Suggested fix
Docs in the configuration reference plus a loud unknown-agent error listing runnable names vs category names.

RaviTharuma:
Keep config-key aliases after display-name remap and fail closed on category CLI agents.

Fixes #7701
Fixes #7703

## Why
Display-name remapping dropped `sisyphus`, so `opencode run --agent sisyphus` silently used the default agent. Categories (`quick`, `deep`, …) are not runnable agents.

## Changes
- Hidden aliases for original config keys (visible keys first, aliases after, so TUI does not duplicate rows)
- Unknown / category `--agent` values throw `UnknownRunAgentError`
- Tests on remapper + resolver

## Test
`bun test` on remapper+resolver (18 pass)


RaviTharuma:
@code-yeongyu this covers both #7701 and #7703, which you identified as the same root problem: hidden aliases for original config keys after the display-name remap, plus fail-closed `--agent` validation that lists runnable agents separately from category names. Ready for review.

PeterPonyu:
The hidden-alias approach is the right one (OpenCode's TUI filters `hidden: true` agents while `Agent.get` still resolves them, and OMO already relies on that for `build`/`plan`), but I think there is an ordering problem in where the aliases are built, found while triaging #7701 against `dev@4ec56b79d`:

- `remapAgentKeysToDisplayNames` runs inside `applyAgentConfig` (`agent-config-finalizer.ts:17`), which `config-handler.ts` awaits *before* `applyToolConfig({ config, pluginConfig, agentResult })` at line 143.
- `applyToolConfig` then mutates the agent objects in place — `sisyphus.permission = { ...sisyphus.permission, … }`, `atlas.permission = …`, `looker.permission = { …, task: "deny", look_at: "deny" }`, `denyTaskForAgent(...)` for `TASK_DENIED_SUBAGENT_KEYS`, and so on (`tool-config-handler.ts:45, 92-125`).
- This PR builds each alias as a shallow spread copy at remap time (`{ ...(value as Record<string, unknown>) }`), so the alias entry is a *different object* from the one `applyToolConfig` later edits. `opencode run --agent sisyphus` would resolve to a Sisyphus without its task/teammate/question permission changes, and `--agent prometheus` to a Prometheus without `bash: "deny"`, while the display-name entries are correct.

Two ways out: create the aliases in `config-handler.ts` after the `applyToolConfig` call, spreading the *final* objects (`{ ...config.agent[displayName], hidden: true }`), or keep them in the remapper but make them lazy (a getter/Proxy over the display-name entry) — the first is simpler and keeps the remapper a pure key rename. A test that asserts `config.agent["sisyphus"].permission` deep-equals `config.agent["Sisyphus - ultraworker"].permission` after the full `config` hook would pin it. Happy to review the update.

RaviTharuma:
@PeterPonyu thanks — the ordering catch is correct and I missed it.

`remapAgentKeysToDisplayNames` runs inside `applyAgentConfig` before `applyToolConfig` mutates permissions in place, and this PR’s aliases are shallow copies taken at remap time, so `--agent sisyphus` / `--agent prometheus` can resolve to pre-tool-config objects while the display-name entries are correct.

I will follow the simpler fix: build the hidden aliases in `config-handler.ts` **after** `applyToolConfig`, spreading the final display-name objects (`{ ...config.agent[displayName], hidden: true }`), and add a test that `config.agent["sisyphus"].permission` deep-equals the display-name entry after the full `config` hook. Will push that update on this PR.

RaviTharuma:
@PeterPonyu Cloud agent is implementing your suggested fix now: build hidden aliases in `config-handler.ts` after `applyToolConfig`, spreading the final display-name objects, plus a deep-equal permission test. Will ping when the tip lands.

RaviTharuma:
@PeterPonyu the ordering fix is on this PR now.
Hidden aliases are built in `config-handler.ts` after `applyToolConfig`, spreading the final display-name objects (`{ ...config.agent[displayName], hidden: true }`). The remapper is a pure key rename again and no longer creates shallow copies at remap time.
New full-hook test: `config.agent["sisyphus"].permission` deep-equals the display-name entry (and the same for prometheus, including `bash: "deny"`). Cache-hit path is covered too. Tip: `607654ed8`.

RaviTharuma:
@PeterPonyu friendly ping — the post-`applyToolConfig` hidden-alias ordering fix (plus the deep-equal permission test) has been on tip `607654ed8` since Sep 17. Ready for another look whenever you have a moment.

PR #7816, opened 2026-09-05
fix(senpi): recover from stale launcher overrides

### LINKED ISSUE #7815: fix: stale SENPI_BIN blocks current launcher fallback
## Summary

A long-running OMO/Senpi process can retain a `SENPI_BIN` path from an installation that was replaced or removed. Memory reflection/dream child resolution treats that missing explicit override as terminal and does not continue to a valid current Senpi on PATH, eventually reporting `Unable to resolve a runnable Senpi launcher`.

## Deterministic reproduction

- `SENPI_BIN` points to a deleted absolute path.
- `process.argv[1]` points to a deleted old CLI.
- package-local CLI lookup returns null to model the moved install.
- PATH contains a valid current Senpi executable.

`resolveSenpiExecutable()` currently returns null at the stale override instead of scanning PATH, and `resolveSenpiLaunch()` throws.

## Expected

A valid explicit override still wins. A missing override falls through to the existing sibling/PATH resolution with engine-version parity checks intact. Background memory launch failures remain isolated from the parent turn.

JeremyDev87:
## What changed

- A deleted `SENPI_BIN` no longer blocks discovery of a valid Senpi on the current PATH.
- Stale install paths, PATH version parity, and the memory child resolver are locked down with regression tests.
- The existing fail-open boundary — a background memory runner rejection must not propagate to the parent turn — is locked down with a test.
- Senpi plugin artifacts are regenerated from source.

## Why

When a long-running OMO session kept a removed npm install path in `SENPI_BIN` and argv, memory reflection/dream failed with `Unable to resolve a runnable Senpi launcher` even though a newly installed Senpi was on PATH.

## How

A valid explicit override still takes priority, as before. Only when the override does not resolve to a real file do we fall back to the existing Bun sibling/PATH search, and PATH candidates keep going through the existing engine-version parity check.

## Verification

- New test RED: with a stale override, the resolver returns null and the launcher throws
- Same test GREEN: 43 targeted tests PASS
- `tsgo --noEmit -p packages/senpi-task/tsconfig.json` PASS
- `bun run test:senpi` PASS: 2,688 passed, 32 skipped, 0 failed
- Evidence resolver 10/10 PASS
- Independent code review: APPROVE

## Out of scope

The project-trust writer lock issue in Senpi core is fixed separately in code-yeongyu/senpi#1394. No publish/release version changes are included.

Closes #7815

---
## Summary by cubic
Fixes two memory reflection failure modes: stale `SENPI_BIN` overrides no longer block Senpi launcher recovery, and detached reflection work no longer reaches stale live session APIs.

**Bug Fixes**
- Falls back to Bun sibling/PATH search only when the explicit override doesn't resolve to a real file; valid overrides keep priority and engine-version parity checks stay intact.
- Model registry and session context are captured in a synchronous snapshot at launch, so detached reflection can't go stale after shutdown.
- Renderer registration is static only; launch entry and completion publication failures are logged without escaping, and session shutdown clears the live session so detached work can't reach its API.
- Adds regression tests covering stale install paths, PATH parity, stale live context, session shutdown, and the fail-open boundary.
- Regenerates Senpi plugin artifacts in sync with the dev merge.

Closes #7815.

<sup>Written for commit 5bf7b1595467f79d6f9c2b1e7aa7bb14abe5dd91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7816?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


JeremyDev87:
The stale extension-context defect found during additional real-usage QA is fixed on the same branch.

- The async reflection path was using a synchronous snapshot instead of the live event context
- Renderer registration is kept in static wiring only
- Stale live API handles are cleared on session shutdown
- A launch-entry append failure is isolated so it no longer turns the reflection result into a failure
- Removing the launch-entry catch gives RED (`expected merged / received failed`); restored, 2/2 GREEN
- Removing the shutdown clear gives a stale-API-handle RED; restored, focused GREEN
- Final `bun run test:senpi`: 2,692 pass / 32 skip / 0 fail
- Evidence resolver: 10/10 pass
- Independent code re-review: APPROVE, no blockers
- The paired automatic-memory E2E was 17/17 pass before the commit, and the stale `SENPI_BIN` PATH fallback 7/7 pass

Current head: `c6a3f1174f20772de220f2436c2b35e07916cf55`

JeremyDev87:
Additionally fixed the generated-provenance issue found in the final gate and the deep stale-registry issue in shutdown dream.

- Detached reflection/dream now uses a snapshot of `getAll`/`getAvailable` values instead of the live `ModelRegistry` object
- The in-process facts child keeps using the concrete live registry, which needs auth/runtime
- Pre-fix fresh-plugin automatic shutdown dream: exact stale-extension-context error, no completion
- Post-fix automatic shutdown dream: `outcome=merged`, `model=omo-mock/mock-1`, `filesChanged=1`
- Full `test:senpi`: 2,692 pass / 32 skip / 0 fail
- After the generated-bundle stabilization commit, `build-extension.mjs --check` exits 0
- Paired memory E2E on the stabilized committed bundle: 17/17 pass
- Installed source launcher: `5.0.0-beta.43+source.5bf7b1595`
- Final concurrent sessions, the exact unresolvable-launcher fail-open plus next turn, automatic reflection, and automatic shutdown dream were all re-verified, and the sandbox was cleaned up

Final head: `5bf7b1595467f79d6f9c2b1e7aa7bb14abe5dd91`

PeterPonyu:
Independently confirmed the root cause against `dev@4ec56b79d` while triaging #7815: `resolveSenpiExecutable` (`packages/senpi-task/src/runners/rpc/spawn.ts:168-174`) returns `canonicalExecutable(override)` directly, so a `SENPI_BIN` that no longer resolves yields `null` and the sibling/PATH stages below it are never reached; `spawn.test.ts:75` ("no silent PATH fallthrough") currently pins that behaviour, and the omo-native launcher writes `SENPI_BIN` from the install tree, so any engine process outliving an upgrade carries the dead pointer. Your validate-and-fall-through shape (warn on the ignored override, keep a *valid* override unconditional and parity-free as `9cd6bce11` intended) is the right one.

Two things that I think are standing between this PR and a merge, offered in case they help:

1. **Scope.** The `spawn.ts` + `spawn.test.ts` (+ the `senpi-command.test.ts` scenario) change is ~8 lines and self-contained; the memory wiring / runner changes (`identity-runtime`, `wiring*`, `runner*`, `model-registry-resolver`) are a different concern and are what makes the diff hard to review as one unit. Splitting the launcher-resolution fix into its own PR would let it merge on its own evidence while the wiring work continues here.
2. **Bundle conflict.** `dev` re-stamps `plugin/extensions/omo*.js` several times a day, so committed bundles conflict almost immediately. What has worked for my PRs today: merge `upstream/dev` right before pushing, then regenerate with the CI-pinned Bun (`~/.bun-1.4.2`, `node packages/omo-senpi/plugin/scripts/build-extension.mjs`) so the `senpi --check` gate compares against a current build; regenerating with a different Bun version produces byte differences that fail that gate.

Happy to review the split PR, or to open it with you credited if you'd rather keep this one focused on the wiring side.

PR #7835, opened 2026-09-06
fix(publish): ship the prompts-core Codex prompt in the oh-my-openagent payload

LilMGenius:
## What breaks

`bunx oh-my-openagent@5.0.0-beta.43 install` (platform `codex` or `both`) cannot install the Codex
plugin. The OpenCode half succeeds, then the Codex half dies and the user is left with
`Codex install failed (OpenCode install remains successful)`. Every registry install of
`oh-my-openagent` / `oh-my-opencode` since the directive was single-sourced hits this;
`bunx lazycodex-ai@5.0.0-beta.43 install` is unaffected, which is why it has gone unnoticed.

## Failing input

```
$ npm pack oh-my-openagent@5.0.0-beta.43
$ tar -tzf oh-my-openagent-5.0.0-beta.43.tgz | grep prompts-core
(no output)

$ tar -tzf lazycodex-ai-5.0.0-beta.43.tgz | grep prompts-core
package/packages/prompts-core/prompts/ultrawork/codex.md
```

Installing from the first tarball:

```
npm notice run @sisyphuslabs/omo-codex-plugin@5.0.0-beta.43 sync:skills
Error: ENOENT: no such file or directory, open
'<CODEX_HOME>\plugins\cache\sisyphuslabs\omo\.tmp-5.0.0-beta.43-...\packages\prompts-core\prompts\ultrawork\codex.md'
    at async syncSkills (.../scripts/sync-skills.mjs:259:38)
```

## Root cause

`copyCanonicalPromptSources` in `packages/omo-codex/src/install/codex-cache-install.ts` materializes
`packages/prompts-core/prompts/ultrawork/codex.md` into the plugin cache so the
`resolveCanonicalUltraworkDirectivePath` fallback in `plugin/scripts/canonical-ultrawork-directive.mjs`
resolves, but it skips silently when the source file is absent. The root `package.json` `files`
allowlist never listed that path, so it is absent in the published tarball, nothing is materialized,
and `npm run sync:skills` reads a file that is not there.

#7630 fixed exactly this for `lazycodex-ai` by adding the path to the `.files` override inside the
publish workflow. The base package's own allowlist was not updated, so the two payloads diverged.

## RED / GREEN

Driven from the published tarballs against an isolated `CODEX_HOME` (`installMarketplaceLocally` from
the packaged `install-dist/install-local.mjs`); `~/.codex/config.toml` SHA256 unchanged across all runs.

| Input | Before | After |
| --- | --- | --- |
[...]

PR #7836, opened 2026-09-06
fix(publish): require the canonical Codex prompt in npm payload verification

LilMGenius:
## What breaks

`script/verify-npm-payload.mjs` is the gate the publish workflow runs before every `npm publish`.
It ran during the 5.0.0-beta.43 release and printed `npm payload containment OK`, and the payload it
approved cannot install: `bunx oh-my-openagent@5.0.0-beta.43 install` dies at `npm run sync:skills`
with `ENOENT ... packages/prompts-core/prompts/ultrawork/codex.md`. The gate has no opinion about the
one file the Codex cache install has to read, so nothing stopped the release.

## Failing input

dev's own `npm pack --dry-run --json` path list, fed to the two required-path sources the verifier
has today and to the one this PR adds:

```
packed paths: 1222
shipped gate (shared-skills only) missing: []
new gate required: ["packages/prompts-core/prompts/ultrawork/codex.md"]
new gate missing: ["packages/prompts-core/prompts/ultrawork/codex.md"]
```

## Root cause

`requiredSharedSkillsPaths()` was added (#7813) after `lazycodex-ai@5.0.0-beta.43` shipped without
`skill-source-filter.mjs`, and it covers exactly the shared-skills export targets. The canonical
ultrawork directive is a second runtime input of the same install step — `copyCanonicalPromptSources`
materializes it into the plugin cache and `sync-skills.mjs` reads it — but it is not derived from any
manifest, so no required-path rule covers it.

## RED / GREEN

| Input | Before | After |
| --- | --- | --- |
| dev's real packed path list | gate reports nothing missing (exit 0) | gate reports the canonical prompt missing (exit 1) |
| `bun test script/npm-payload-required-paths.test.ts` | file did not exist | 3 pass / 0 fail |
| mutation: `requiredCodexInstallPaths()` returns `[]` | - | 2 of 3 new tests fail, green again on revert |
| `bun test script/npm-payload-containment.test.ts` | 5 pass / 0 fail | 5 pass / 0 fail (control) |

Evidence: `.omo/evidence/20260906-npm-payload-codex-prompt-gate/`.

## The fix

The required path is read from `plugin/scripts/canonical-ultrawork-directive.mjs`, the module the
installer itself uses, so the gate cannot drift from the installer, and it is normalized to posix
separators because `npm pack` reports posix paths. The existing shared-skills requirement is
untouched; both feed the same `missing` report.

## Ordering

This gate is intentionally strict: on current `dev` it fails, because the base payload really is
missing that path. #7835 adds the path to the root `files` allowlist. Merging #7835 first (or both)
keeps `publish.yml` green; merging this one alone turns the next base publish into a loud, correct
failure instead of a broken tarball.

- No change to the forbidden-path rules, the shared-skills requirement, or the workflow.
- `packedPaths()` still assumes npm's array-shaped `npm pack --json`; that is correct for the pinned
  Node 24 (npm 11) used by CI and publish, and npm 12's object shape is out of scope here.

---
## Summary by cubic
Fixes the npm publish gate so it requires the canonical Codex prompt, which the Codex installer reads at `npm run sync:skills`. The gate previously approved payloads missing `packages/prompts-core/prompts/ultrawork/codex.md`, so 5.0.0-beta.43 published broken.

- The required path is read from the installer's own `canonical-ultrawork-directive.mjs` module and normalized to posix separators, so the gate and installer cannot drift.
- The gate now fails on current `dev` because the base payload is missing that path; merge the companion PR that adds the path to the `files` allowlist to keep `publish.yml` green.

<sup>Written for commit d510125f8defb070f35ffb728bcb1776dcf827f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7836?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260906-npm-payload-codex-prompt-gate/README.md
+++ b/.omo/evidence/20260906-npm-payload-codex-prompt-gate/README.md
@@ -1,0 +1,40 @@
+# The npm payload verifier does not require the prompt the Codex installer reads
+
+## What was tested
+
+`script/verify-npm-payload.mjs`, the gate the publish workflow runs before every `npm publish`
+(base package, `oh-my-openagent` alias, and `lazycodex-ai`), against the real `npm pack --dry-run`
+path list of this branch's base commit.
+
+- Probe: `gate-probe.mjs` computes what the shipped gate requires (shared-skills export targets only)
+  and what the added Codex-install requirement resolves to, both against the same packed path list.
+- Unit: `bun test script/npm-payload-required-paths.test.ts`.
+
+## What was observed
+
+```
+packed paths: 1222
+shipped gate (shared-skills only) missing: []
+new gate required: ["packages/prompts-core/prompts/ultrawork/codex.md"]
+new gate missing: ["packages/prompts-core/prompts/ultrawork/codex.md"]
+```
+
+| Input | Before | After |
+| --- | --- | --- |
+| real packed path list of `dev` (payload that shipped as 5.0.0-beta.43) | gate reports nothing missing | gate reports the canonical Codex prompt missing |
+| `bun test script/npm-payload-required-paths.test.ts` | file did not exist | 3 pass / 0 fail |
+| mutation: `requiredCodexInstallPaths()` returns `[]` | - | 2 of 3 tests fail, then pass again once reverted |
+
+## Why it is enough
+
+The gate ran in the 5.0.0-beta.43 publish and printed containment OK for a payload whose Codex
+install cannot run, which is the exact miss this change closes. The probe reproduces that verdict on
+the same packed path list, and the mutation shows the new assertions fail when the requirement is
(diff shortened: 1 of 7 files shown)


code-yeongyu:
Merged as 073f00928, on top of #7835 (b6d76f546) so the verifier's new requirement is satisfied by the payload it gates. Included in the 5.0.0-beta.44 hotfix.

PR #7839, opened 2026-09-06
ci(published-smoke): install the published payloads instead of only dry-running

LilMGenius:
## What breaks

`lazycodex-published-smoke` is the only job that looks at what was actually published, and it runs
two `--dry-run` commands and nothing else. A `--dry-run` prints the command it would run, so the job
is structurally incapable of failing on a payload that cannot install.

It was green through the entire 5.0.0-beta.43 release, and **both** payloads of that release die on
first install:

- `oh-my-openagent@5.0.0-beta.43` - missing `packages/prompts-core/prompts/ultrawork/codex.md` (#7835)
- `lazycodex-ai@5.0.0-beta.43` - missing `packages/shared-skills/skill-source-filter.mjs`, found after
  the fact and fixed by #7813

Both surface to the user as `npm run sync:skills failed ... exit code 1` during
`bunx <pkg> install`.

## Failing input

```
$ node script/published-install-smoke.mjs --package=oh-my-openagent@5.0.0-beta.43
published install smoke failed: npm run sync:skills failed in <sandbox>/plugins/cache/sisyphuslabs/omo/.tmp-5.0.0-beta.43-... with exit code 1
$ echo $?
1
```

## Root cause

The job never leaves `--dry-run`, so no payload is ever unpacked or installed. Nothing between
`npm publish` and a user's machine executes the published installer.

## RED / GREEN

| Input | Before (job as it stands) | After |
| --- | --- | --- |
| `oh-my-openagent@5.0.0-beta.43` (payload that cannot install) | job green, no install attempted | step exits 1 naming the `sync:skills` failure |
| the same tarball plus the missing directive | job green | step exits 0, `published install smoke OK (1 plugin(s), ...)` |
| `bun test script/published-install-smoke.test.ts` | file did not exist | 3 pass / 0 fail; drops to 1 fail under a mutation of the required-artifact list |
| `.github/workflows/ci.yml` parsed after the edit | - | `continue-on-error: true` and both dry-run assertions unchanged |

### Observed in CI on this branch

The job ran the new step for real (run 34023407423):

```
published install smoke passed for lazycodex-ai@latest
##[warning]published install smoke failed for lazycodex-ai@beta
published install smoke passed for oh-my-openagent@latest
##[warning]published install smoke failed for oh-my-openagent@beta
```

Both stable payloads install in about 6 seconds each; both `5.0.0-beta.43` payloads die at
`sync:skills` - one per known defect (#7835 for `oh-my-openagent`, the `skill-source-filter.mjs`
omission #7813 fixed after the fact for `lazycodex-ai`). That is the exact pair the dry-run-only job
shipped past.

Evidence: `.omo/evidence/20260906-published-install-smoke/`.

## The fix

A small script plus one step. Two properties worth checking in review:

- **The sandbox is created by the script, never taken from the environment.** `CODEX_HOME` and
  `CODEX_LOCAL_BIN_DIR` are always `mkdtemp` paths, and the whole sandbox is removed in a `finally`
  block unless `--keep` is passed, so a caller's real `~/.codex` cannot be touched.
- **The assertion is the composed artifact, not a file count.** It requires
  `skills/ultrawork/SKILL.md`, which only exists if `sync:skills` actually ran inside the cache, which
  is exactly what both broken payloads failed to do.

The step keeps the job's existing shape: the job stays `continue-on-error: true` and the step emits
`::warning::` per failing spec. The `@latest` tag - what `bunx` resolves for users - fails the step;
the `@beta` tag only annotates, because 5.0.0-beta.43 cannot install today and failing on it would
redden every unrelated PR until the next release. Flipping beta to blocking is one line
(`case "$spec"` in the step) once a good beta is published.

- The two existing dry-run assertions are unchanged.
- No change to `verify-npm-payload.mjs`, the `files` allowlists, or `publish.yml`.
- The `test (windows-latest, 2/2)` failure on this branch is `LspClient diagnostics freshness` in
  `packages/lsp-core`, which this diff never touches; it is not caused by this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -572,7 +572,9 @@ jobs:
               echo "published install smoke passed for $spec"
             else
               echo "::warning::published install smoke failed for $spec"
-              status=1
+              case "$spec" in
+                *@latest) status=1 ;;
+              esac
             fi
             echo "::endgroup::"
           done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -566,7 +566,7 @@ jobs:
         run: |
           set -uo pipefail
           status=0
-          for spec in lazycodex-ai@latest oh-my-openagent@latest; do
+          for spec in lazycodex-ai@latest lazycodex-ai@beta oh-my-openagent@latest oh-my-openagent@beta; do
             echo "::group::published install smoke: $spec"
             if node script/published-install-smoke.mjs --package="$spec"; then
               echo "published install smoke passed for $spec"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -558,6 +558,26 @@ jobs:
             echo "::warning::lazycodex-ai doctor dry-run output changed: $npx_doctor_output"
           fi
 
+      - uses: actions/checkout@v7
+        if: needs.ci-mode.outputs.run_heavy == 'true'
+
+      - name: Install published payloads into an isolated CODEX_HOME
+        if: needs.ci-mode.outputs.run_heavy == 'true'
+        run: |
+          set -uo pipefail
+          status=0
+          for spec in lazycodex-ai@latest oh-my-openagent@latest; do
+            echo "::group::published install smoke: $spec"
+            if node script/published-install-smoke.mjs --package="$spec"; then
+              echo "published install smoke passed for $spec"
+            else
+              echo "::warning::published install smoke failed for $spec"
+              status=1
+            fi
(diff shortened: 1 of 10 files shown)


code-yeongyu:
Merged (f4fdb53da) into the 5.0.0-beta.44 train. Reviewed: the smoke now drives `installMarketplaceLocally` for both published payloads — exactly the step where beta.43 died — while staying `continue-on-error` in ci.yml, so it can never block a publish. It would have flagged both beta.43 tarballs.

PR #7840, opened 2026-09-06
fix(publish): read both npm pack --json shapes in the payload verifier

LilMGenius:
## What breaks

`script/verify-npm-payload.mjs` is the last gate before `npm publish`. On npm 12 it dies before it
checks a single path:

```
TypeError: parsed is not iterable
    at packedPaths (script/verify-npm-payload.mjs:33)
```

npm 11 prints `npm pack --json` as an array of package results; **npm 12 prints an object keyed by
package name**, and npm 12 is the npm bundled with Node 26. `publish.yml` and `ci.yml` pin Node 24
today, so this is not breaking CI right now - it breaks the day that pin moves, and it already breaks
every local run of the script on a current Node.

## Failing input

Real output from this repository, `npm pack --dry-run --json --ignore-scripts` on npm 12.0.2:

```
old reader THROWS: TypeError: parsed is not iterable
new reader -> 1222 paths
new reader first path: .agents/command/get-unpublished-changes.md
```

## Root cause

`packedPaths()` destructures the array form directly:

```js
const [result] = JSON.parse(raw)
return result.files.map((file) => file.path)
```

Any non-array shape throws on the destructure, and a result without `files` throws a
read-of-undefined - neither names npm pack, so the failure reads as a broken script.

## RED / GREEN

| Input | Before | After |
| --- | --- | --- |
| real npm 12 `pack --json` output | `TypeError: parsed is not iterable` | 1222 paths |
| npm 11 array output | 2 paths | 2 paths (control, unchanged) |
| `[]`, `{}`, `null`, a result without `files` | read of undefined | `Error: npm pack --json returned no packed file list` |
| whole verifier run against the npm 12 payload | never reached the checks | `npm payload containment OK (1222 packed paths, 0 offenders)`, exit 0 |

Evidence: `.omo/evidence/20260906-npm-pack-json-shape/`.

## The fix

One reader, `script/npm-pack-paths.mjs`, that accepts both shapes and fails with a named error for
anything else, plus its unit test. The verifier's forbidden-rule and required-path logic is untouched.

- No Node or npm version pin is changed; this only removes the assumption.
- `execFileSync("npm", ...)` still cannot resolve `npm.cmd` on Windows without a shell
  (`spawnSync npm ENOENT`), so the script remains Linux/macOS-only to run locally. That is a separate
  defect and I left it out of this diff; happy to send it separately if you want it.
- No change to the forbidden rules, the shared-skills requirement, or `publish.yml`.

---
## Summary by cubic
Fixes the publish payload verifier so it reads both npm 11 and npm 12 `npm pack --json` output shapes. npm 12 (bundled with Node 26) prints an object keyed by package name; the old array-only reader threw `TypeError: parsed is not iterable` before checking a single path.

- Extracts the path reader into `script/npm-pack-paths.mjs` with unit tests; npm 11 array output behavior is unchanged.
- Malformed output, non-JSON output, and entries without a string path now fail with errors that name `npm pack` instead of a bare `SyntaxError` or read-of-undefined.
- Adds captured RED/GREEN evidence under `.omo/evidence/20260906-npm-pack-json-shape/` and `.omo/evidence/20260907-rebase-7840/`.
- No Node or npm version pins change; CI still runs Node 24 and is unaffected.

<sup>Written for commit 601880588564695b4a84c7b0cbf37d8fbf189764. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7840?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260906-npm-pack-json-shape/README.md
+++ b/.omo/evidence/20260906-npm-pack-json-shape/README.md
@@ -1,0 +1,40 @@
+# The payload verifier reads only npm 11's pack --json shape
+
+## What was tested
+
+`script/verify-npm-payload.mjs`, the gate `publish.yml` runs before every `npm publish`, against
+real `npm pack --dry-run --json` output produced by npm 12.0.2 on this machine (1222 packed paths).
+
+## What was observed
+
+`red-npm12.log`, both readers on the same real npm 12 payload:
+
+```
+old reader THROWS: TypeError: parsed is not iterable
+new reader -> 1222 paths
+new reader first path: .agents/command/get-unpublished-changes.md
+```
+
+| Input | Before | After |
+| --- | --- | --- |
+| real npm 12 `pack --json` output | `TypeError: parsed is not iterable` | 1222 paths, identical to the npm 11 shape |
+| npm 11 array output | 2 paths | 2 paths (unchanged) |
+| output of neither shape (`[]`, `{}`, `null`, a result without `files`) | read of undefined | `Error: npm pack --json returned no packed file list` |
+| whole verifier end to end on the npm 12 payload | could not reach the checks | `npm payload containment OK (1222 packed paths, 0 offenders)`, exit 0 |
+
+Unit run: `green-unit.log`, 3 pass / 0 fail. Verifier run: `green-verifier.log`.
+
+## Why it is enough
+
+The RED is real registry-tool output from this repository, not a synthetic fixture, and the GREEN
+run drives the actual script rather than the extracted helper. Severity is forward-looking: CI and
+`publish.yml` pin Node 24 (npm 11) today, so this is the gate breaking the day that pin moves, plus
+a named error instead of a read-of-undefined for any other shape.
+
+## What was omitted
+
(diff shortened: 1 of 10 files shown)


code-yeongyu:
Thanks for the rebase — taking your head (601880588) as-is; merging as soon as checks are green, then #7844 (which I stacked on top of it). Going into 5.0.0-beta.44.

code-yeongyu:
Merged as part of the 5.0.0-beta.44 hotfix train. Reviewed: the npm 11 array path returns the same file list as before, npm 12's object shape is handled via `Object.values`, and malformed output now fails with a named error instead of a bare TypeError — the forbidden-rule and required-path checks are untouched.

PR #7844, opened 2026-09-06
fix(publish): spawn npm through a shell on Windows in the payload verifier

LilMGenius:
## What breaks

A contributor on Windows cannot run the publish payload gate at all:

```
$ node script/verify-npm-payload.mjs
Error: spawnSync npm ENOENT
    at packedPaths (script/verify-npm-payload.mjs:28)
  syscall: 'spawnSync npm', path: 'npm',
  spawnargs: [ 'pack', '--dry-run', '--json', '--ignore-scripts' ]
```

It never reaches npm, so the whole containment check is unavailable locally on that platform. CI runs
it on Linux, so the gate itself is fine in the pipeline; this is purely the local surface.

## Root cause

npm is `npm.cmd` on Windows, and Node's `spawnSync` does not apply `PATHEXT` unless a shell is
requested, so `execFileSync("npm", ...)` cannot resolve it. Spawning `.cmd` directly is rejected by
Node since the argument-injection fix, so a shell is the remaining route.

## RED / GREEN

| Input | Before | After |
| --- | --- | --- |
| `node script/verify-npm-payload.mjs` on Windows | `spawnSync npm ENOENT`, exit 1 | reaches npm; stops at the npm 12 pack shape, which is #7840 |
| the same run with #7840's reader also applied | dies before npm | `npm payload containment OK (1259 packed paths, 0 offenders)`, exit 0 |
| `bun test script/npm-invocation.test.ts` | file did not exist | 2 pass / 0 fail; 1 fails under a mutation that drops the win32 branch |
| `bun test script/npm-invocation.test.ts script/npm-payload-containment.test.ts` | - | 7 pass / 0 fail |

Evidence: `.omo/evidence/20260906-verify-payload-windows-spawn/`.

## The fix

One exported helper, `npmSpawnOptions(platform)`, returning `{ shell: true }` on win32 and `{}`
elsewhere, spread into the existing `execFileSync` options. Non-Windows behavior is byte-identical.

Worth checking in review: the four npm arguments are fixed literals with no interpolation, so the
shell carries no caller-controlled input. Node still prints DEP0190 about argument escaping under
`shell: true`; the alternative would be resolving npm's JS entry by hand, which is more fragile than
the warning is costly.

- The npm 12 `pack --json` object shape is #7840; this PR only removes the spawn failure. They are
  independent and can merge in either order.
- No other npm spawn in the repo is changed; the helper exists so the next one can adopt it
  deliberately.

---
## Summary by cubic
Fixes both publish payload verifiers so they can run on Windows, and makes the npm payload verifier accept npm 12's pack output. They previously died before inspecting any packed paths: `spawnSync npm ENOENT` on Windows (npm is `npm.cmd` there, and Node's spawn doesn't apply `PATHEXT` without a shell) and `TypeError: parsed is not iterable` on npm 12, which prints `pack --json` as an object keyed by package name rather than an array.

**Details**
- Both verifiers spread `npmSpawnOptions()` into their `execFileSync` calls; the helper returns `{ shell: true }` on win32 and `{}` elsewhere, so non-Windows behavior is byte-identical.
- The npm verifier's pack reader `parseNpmPackPaths()` accepts both the npm 11 array shape and the npm 12 object shape, failing with a named error on non-JSON output, entries without a string path, or unrecognized shapes.
- Tests spawn real npm through those options, pack a temp package, and assert `stdout` parses as JSON; the omo-ai verifier fixture copies the helper so its import resolves there.
- Node prints a DEP0190 deprecation warning about arg escaping under `shell: true`; the npm arguments are fixed literals with no interpolation, so the warning is accepted rather than worked around.

<sup>Written for commit 8b6bd37391af4a0c197ce69d48420647e769b47a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7844?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260906-npm-pack-json-shape/README.md
+++ b/.omo/evidence/20260906-npm-pack-json-shape/README.md
@@ -1,0 +1,40 @@
+# The payload verifier reads only npm 11's pack --json shape
+
+## What was tested
+
+`script/verify-npm-payload.mjs`, the gate `publish.yml` runs before every `npm publish`, against
+real `npm pack --dry-run --json` output produced by npm 12.0.2 on this machine (1222 packed paths).
+
+## What was observed
+
+`red-npm12.log`, both readers on the same real npm 12 payload:
+
+```
+old reader THROWS: TypeError: parsed is not iterable
+new reader -> 1222 paths
+new reader first path: .agents/command/get-unpublished-changes.md
+```
+
+| Input | Before | After |
+| --- | --- | --- |
+| real npm 12 `pack --json` output | `TypeError: parsed is not iterable` | 1222 paths, identical to the npm 11 shape |
+| npm 11 array output | 2 paths | 2 paths (unchanged) |
+| output of neither shape (`[]`, `{}`, `null`, a result without `files`) | read of undefined | `Error: npm pack --json returned no packed file list` |
+| whole verifier end to end on the npm 12 payload | could not reach the checks | `npm payload containment OK (1222 packed paths, 0 offenders)`, exit 0 |
+
+Unit run: `green-unit.log`, 3 pass / 0 fail. Verifier run: `green-verifier.log`.
+
+## Why it is enough
+
+The RED is real registry-tool output from this repository, not a synthetic fixture, and the GREEN
+run drives the actual script rather than the extracted helper. Severity is forward-looking: CI and
+`publish.yml` pin Node 24 (npm 11) today, so this is the gate breaking the day that pin moves, plus
+a named error instead of a read-of-undefined for any other shape.
+
+## What was omitted
+
+The end-to-end verifier run substitutes the `npm pack` spawn with the saved npm 12 payload for the
+duration of that one run, because `execFileSync("npm", ...)` cannot resolve `npm.cmd` on Windows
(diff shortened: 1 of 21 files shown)


code-yeongyu:
Heads-up @LilMGenius: I rebased this branch (maintainer edit) onto your rebased #7840 head (601880588) so both verifier changes stack cleanly on the merged #7836 gate — the only conflict was the import block plus the `...npmSpawnOptions()` spread in `packedPaths()`. Please pull before pushing again. Plan: merge #7840 when its checks are green, then this one. Both are going into the 5.0.0-beta.44 hotfix.

code-yeongyu:
Merged (stacked on #7840). Included in the 5.0.0-beta.44 hotfix: the verifier's POSIX spawn options are byte-for-byte unchanged, `shell: true` only on win32, argv stays the four literals.

Issue #5971, opened 2026-07-08
[Bug]: background task accidentally interrupted without informing main agent, casuing eternal waiting


UCASRMTG5D:
### Prerequisites

- [x] I will write this issue in English (see our [Language Policy](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md#language-policy))
- [x] I have searched existing issues to avoid duplicates
- [x] I am using the latest version of oh-my-openagent
- [x] I have read the [documentation](https://github.com/code-yeongyu/oh-my-openagent#readme) or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer

### Bug Description

A background task accidentally stopped, maybe the provider suddenly failed. 
The background task is now dead, but the main agent doesn't know it, it is still waiting for an answer.
No answer will come, nothing will come, and the main agent waits forever, in dreadful silence.

### Steps to Reproduce

1. use ulw mode with a hard task, so the main agent distributes background tasks.
2. one of the background task failed as: no provider available.
3. main agent main session wait forever, it doesn't know one of the background task failed.

### Expected Behavior

1. after background task fail, main agent main session can know it and retry.
2. if retry fail, background task should fallback to other models as user configured.

### Actual Behavior

main session stuck and wait forever without knowing a background task is dead.

### Doctor Output

```shell
I currently don't have source code on my current computer but I use macbook air (m4 chip) and latest system.
```

### Error Logs

```shell

```

### Configuration

```json

```

### Additional Context

https://github.com/code-yeongyu/oh-my-openagent/issues/5804 I notice a similar issue but it's not the same behaviour.

### Operating System

macOS

### OpenCode Version

1.17.15

PR #7850, opened 2026-09-06
fix(background-agent): stop reading an unreachable server as a live session

LilMGenius:
## Summary

- When the OpenCode server is unreachable, running background tasks are never finalized. `session.get` rejects with `Unable to connect. Is the computer able to access the url?`, and both readers report that as "cannot tell", which the stale-task sweep reads as "keep waiting". A task with no activity for its entire timeout window stays `running` and the parent session is never told.
- Root cause: `checkSessionExistence` returns `"unknown"` for any error that is not a 404, and `getSessionActivityFromClient` returns `"unavailable"` for any thrown error. Both answers are indistinguishable from a healthy server that had a transient hiccup, so the sweep skips its own timeout.
- Fix: classify connection-level failures separately. An unreachable server is no longer evidence that the session is alive. A real API error still answers `"unknown"` / `"unavailable"` exactly as before.

## Changes

- `error-classifier.ts`: adds `isTransportUnreachableError`, matching connection-level failure text only (`unable to connect`, `ECONNREFUSED`, `ECONNRESET`, `ENOTFOUND`, `socket hang up`, `fetch failed`, `network error`, `not connected`, `connection closed/refused/reset`).
- `session-existence.ts`: a connection-level failure answers `"missing"` instead of `"unknown"`, for both a rejected promise and an error carried in the response.
- `session-activity.ts`: the same failure answers `{ type: "missing" }` instead of `{ type: "unavailable" }`, with its own log line so the reason is visible.
- `session-reads-unreachable.test.ts`: new tests, including two controls.

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/features/background-agent/session-reads-unreachable.test.ts`.
  **Observed result:** RED before the change, 3 fail / 2 pass. GREEN after, 5 pass.
  **Why sufficient:** the three failing cases are exactly the reads this PR changes, and the two that passed both before and after are the controls: a 500 API error must still answer `"unknown"` and `"unavailable"`, so the change is not a blanket loosening.
- **What was tested:** the error text was taken from a real client, not invented. A `createOpencodeClient({ baseUrl: "http://127.0.0.1:59117" })` pointed at a closed port rejects with `Error: Unable to connect. Is the computer able to access the url?`, which `dev` classified as `existence=unknown`, `activity=unavailable`.
  **Observed result:** recorded locally at `.omo/evidence/20260907-archive-branch-triage/live-transport-probe.test.ts.txt`.
  **Why sufficient:** it pins the patterns to what the SDK actually emits today rather than to a guess.
- **What was tested:** `bun test packages/omo-opencode/src/features/background-agent/` as the regression control.
  **Observed result:** 754 pass, 0 fail across 61 files.
  **Observed result:** exit 0.

## Risks & Residuals

- Reading an unreachable server as "not visible" makes a task eligible for the stale path it was already eligible for by time: the default windows are 45 minutes of no activity and 60 minutes of message staleness, so a brief network blip cannot finalize live work on its own. Accepted, and stated here rather than left for review to find.
- The classifier matches text, so a provider that words a connection failure differently is not covered. Narrow by design: adding a pattern is cheap, and a broad match would swallow real API errors, which the two control tests forbid.
- Open PR #7265 also edits `session-existence.ts` (session-id prefix stripping, issue #6263). Different root cause, adjacent lines; if it lands first I will rebase onto it.

## Automated Checks

```bash
bun run typecheck
```

## Related Issues

None open for this root cause. Searched `checkSessionExistence`, `getSessionActivityFromClient session activity unavailable`, `Unable to connect background task` and `stale task never finalized running forever` across open issues and PRs. The precedent is #5971, which landed terminate-and-notify for terminal session errors; this is the same shape for an endpoint that cannot be reached at all.

The companion case is #7851: when the abort of a confirmed-gone session fails, the sweep backs off forever. They are independent and can merge in either order; together they close the unreachable-endpoint path end to end.

---
## Summary by cubic
Fixes the stale-task sweep so background tasks no longer stay `running` forever when the OpenCode server is unreachable. Connection-level failures now read as "session missing" instead of "unknown"/"unavailable", letting the sweep finalize stale tasks, while real API errors still answer as before.

**Bug Fixes**
- Added `isTransportUnreachableError` matching connection failure patterns like `ECONNREFUSED`, `ENOTFOUND`, and "unable to connect".
- Connection failures in the activity read log a dedicated line so the reason stays visible.
- Added tests covering thrown and response-carried connection errors, with 500-response controls that must stay `unknown`/`unavailable`.

<sup>Written for commit 412c240a2a41f44306e23c4158f332734c2606f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7850?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/error-classifier.ts
+++ b/packages/omo-opencode/src/features/background-agent/error-classifier.ts
@@ -6,6 +6,24 @@ export function isAbortedSessionError(error: unknown): boolean {
   return message.toLowerCase().includes("aborted")
 }
 
+const TRANSPORT_UNREACHABLE_PATTERNS: readonly RegExp[] = [
+  /unable to connect/i,
+  /econnrefused/i,
+  /econnreset/i,
+  /enotfound/i,
+  /socket hang up/i,
+  /fetch failed/i,
+  /network(?:\s|_)?error/i,
+  /not connected/i,
+  /connection (?:closed|refused|reset)/i,
+]
+
+export function isTransportUnreachableError(error: unknown): boolean {
+  const text = extractErrorMessage(error) ?? getErrorText(error)
+  if (!text) return false
+  return TRANSPORT_UNREACHABLE_PATTERNS.some((pattern) => pattern.test(text))
+}
+
 export function getErrorText(error: unknown): string {
   if (!error) return ""
   if (typeof error === "string") return error
--- a/packages/omo-opencode/src/features/background-agent/session-activity.ts
+++ b/packages/omo-opencode/src/features/background-agent/session-activity.ts
@@ -1,4 +1,5 @@
 import { isRecord, log } from "../../shared"
+import { isTransportUnreachableError } from "./error-classifier"
 import type { OpencodeClient } from "./opencode-client"
 
 export type SessionActivityLookup =
@@ -39,6 +40,10 @@ export async function getSessionActivityFromClient(
       ...(directory ? { query: { directory } } : {}),
     })
     if (isRecord(response) && response.error !== undefined && response.error !== null) {
+      if (isTransportUnreachableError(response.error)) {
+        log("[background-agent] Session server is unreachable while reading activity:", { sessionID, error: response.error })
+        return { type: "missing" }
+      }
       log("[background-agent] Failed to read session activity:", { sessionID, error: response.error })
       return { type: "unavailable" }
     }
@@ -46,6 +51,13 @@ export async function getSessionActivityFromClient(
     const sessionInfo = isRecord(response) && "data" in response ? response.data : response
     return sessionActivityLookupFromInfo(sessionInfo)
   } catch (error) {
+    if (isTransportUnreachableError(error)) {
+      log("[background-agent] Session server is unreachable while reading activity:", {
+        sessionID,
+        error: error instanceof Error ? error.message : error,
+      })
+      return { type: "missing" }
+    }
     if (error instanceof Error) {
       log("[background-agent] Failed to read session activity:", { sessionID, error: error.message })
       return { type: "unavailable" }
--- a/packages/omo-opencode/src/features/background-agent/session-existence.ts
+++ b/packages/omo-opencode/src/features/background-agent/session-existence.ts
@@ -1,3 +1,4 @@
+import { isTransportUnreachableError } from "./error-classifier"
 import type { OpencodeClient } from "./opencode-client"
 
 export const MIN_SESSION_GONE_POLLS = 3
@@ -48,11 +49,13 @@ export async function checkSessionExistence(
     })
 
     if (response.error !== undefined && response.error !== null) {
+      if (isTransportUnreachableError(response.error)) return "missing"
       return isSessionNotFoundError(response.error) ? "missing" : "unknown"
     }
 
     return response.data != null ? "exists" : "missing"
   } catch (error) {
+    if (isTransportUnreachableError(error)) return "missing"
     return isSessionNotFoundError(error) ? "missing" : "unknown"
   }
 }
--- a/packages/omo-opencode/src/features/background-agent/session-reads-unreachable.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/session-reads-unreachable.test.ts
@@ -1,0 +1,53 @@
+import { describe, expect, mock, test } from "bun:test"
+import { unsafeTestValue } from "../../../../../test-support/unsafe-test-value"
+import { getSessionActivityFromClient } from "./session-activity"
+import { checkSessionExistence } from "./session-existence"
+import type { OpencodeClient } from "./opencode-client"
+
+function clientThatRejects(error: unknown): OpencodeClient {
+  return unsafeTestValue<OpencodeClient>({
+    session: {
+      abort: mock(() => Promise.resolve()),
+      get: mock(() => Promise.reject(error)),
+    },
+  })
+}
+
+function clientThatReturnsError(error: unknown): OpencodeClient {
+  return unsafeTestValue<OpencodeClient>({
+    session: {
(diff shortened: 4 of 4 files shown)


LilMGenius:
@code-yeongyu, when you have a moment, could you review #7850 and #7851 together? They close one user-visible path from two independent causes: a background task that never finishes because the OpenCode endpoint went away, leaving the parent session waiting on it forever.

I reproduced both against `dev` at `0c12a2157` with a real `@opencode-ai/sdk` client pointed at a closed port, not a hand-written mock, so the error text is what the SDK actually emits today:

| Observation on `dev` | Result |
| --- | --- |
| `session.get` against a closed port | rejects with `Unable to connect. Is the computer able to access the url?` |
| `checkSessionExistence` on that error | `"unknown"` — the stale sweep keeps waiting |
| `getSessionActivityFromClient` on that error | `"unavailable"` — the timeout is skipped |
| a task 6 hours stale against 60s timeouts | stays `running`, parent never notified |

The two fixes have distinct roles:

- **#7850** stops the two session readers from reporting an unreachable server as a live session. A real API error still answers `"unknown"` / `"unavailable"`, asserted by two control tests, so it is not a blanket loosening.
- **#7851** stops `interruptStaleTask` from backing off forever when the abort fails on a session that is already gone. It re-checks existence first, so a session that still exists keeps the current retry path byte-for-byte.

They are independent and can merge in either order — each one alone fixes its own case, and together they cover the unreachable-endpoint path end to end. Neither changes a timeout, a threshold, or the exit policy.

Both are MERGEABLE with all required checks green (23 success, 0 failing), and each carries its RED capture before the change and GREEN after, plus `bun test packages/omo-opencode/src/features/background-agent/` as the regression control (750 and 754 pass respectively).

The closest precedent is #5971, which landed terminate-and-notify for terminal session errors; this is the same shape for an endpoint that cannot be reached at all. I also checked #6263/#7265 (session-id prefix) and #5656 (stale-cleanup refactor): different root causes, and #5656 leaves the `if (!aborted) return` path untouched, so nothing here competes with work already in review.

code-yeongyu (approved):
Approving. I re-derived the path from `dev` rather than taking the description on trust, and it holds.

**What I checked**

- With the endpoint gone, `manager.ts` `session.status()` rejects, so `allStatuses` is `undefined` and the sweep enters `checkAndInterruptStaleTasks` with `sessionStatuses === undefined`. That means `sessionGone` is false and `shouldRefreshFromSessionActivity` is false, so the only thing standing between a 45-minute-stale task and finalization is `interruptStaleTask` -> `abortWithTimeout`, which also fails against a dead endpoint. So this PR by itself does not finalize the task; it is #7851's post-abort re-check that consumes the `"missing"` answer this PR produces. The two really are one fix split along a clean seam, which is fine, and the end-to-end path is exercised in a combined probe below.
- Combined tree (dev + #7850 + #7851) on an isolated Linux box: a scratch test drives `checkAndInterruptStaleTasks` with a client whose `session.get` / `session.abort` / `session.status` all reject with the real Bun message `Unable to connect. Is the computer able to access the url?`. A task 46 min stale goes `running -> cancelled` and `notifyParentSession` fires once; on plain `dev` the same task stays `running`. A task 5 min stale stays `running` on the combined tree, so a blip cannot finalize live work. `bun test packages/omo-opencode/src/features/background-agent/` on the combined tree: 757 pass, 0 fail across 63 files (bun 1.4.0, Linux box).

**Observations, none blocking**

1. `verifySessionExists` (`session-existence.ts:60`) is also consumed by the crash path in `manager.ts:3086` (`failCrashedTask`), which has no timeout of its own beyond `MIN_SESSION_GONE_POLLS`. After this change a transport error on that single `session.get` reads as "gone". The window is narrow — that branch only runs when `session.status()` succeeded milliseconds earlier — and a server that died between the two calls has taken the session with it, so I am fine with it; noting it so the next reader does not have to rediscover it.
2. `/not connected/i` is the broadest pattern in the list. For `session.get` it is safe because the server never words a per-session error that way, but it is the one I would drop first if a false positive ever shows up.
3. I will land a follow-up adding the combined end-to-end regression test to the repo (`task-poller-unreachable-server.test.ts`), so the interaction between the two PRs stays pinned rather than only each half.

Thanks for the real-client probe instead of a guessed error string — that is what made this quick to verify.

code-yeongyu:
Reviewed both together as asked — approved and merging in the order #7850 -> #7851. Two things beyond what the descriptions already say:

1. On the unreachable-endpoint path `session.status()` also rejects, so the sweep runs with no status registry at all; in that shape #7850 alone never finalizes (the abort fails, `interruptStaleTask` returns) and #7851 alone keeps waiting (`unknown`). They are one fix split along a clean seam, which is fine, but the combined behavior is what users get, so I verified that directly: on a clean Linux box, a scratch test driving `checkAndInterruptStaleTasks` with a client that rejects every call with the real Bun message shows `running -> cancelled` + one parent notification on the combined tree, `running` forever on `dev`, and a 5-minute-stale task untouched. I'll add that test to the repo as a follow-up so the interaction stays pinned.
2. `verifySessionExists` also feeds the crash path at `manager.ts:3086`, which has no timeout of its own. The window is narrow (status call succeeded ms earlier), so not blocking — noted in the review for the next reader.

Background-agent suite on the combined tree: 757 pass, 0 fail across 63 files (bun 1.4.0, Linux box). Thanks for the clean split and the real-client probe.

PR #7851, opened 2026-09-06
fix(background-agent): finalize a stale task whose session is gone and cannot be aborted

LilMGenius:
## Summary

- A background task whose child session is already gone is never finalized when the abort call cannot succeed, so the task stays `running` forever and the parent session is never notified.
- `interruptStaleTask` treats a failed abort as a reason to retry on the next sweep. When the session is gone, that retry can never succeed, so the sweep backs off forever.
- Fix: before backing off, confirm whether the session still exists. A session that exists keeps the retry path unchanged; a session confirmed missing is finalized without a clean abort.

## Changes

- `packages/omo-opencode/src/features/background-agent/task-poller.ts`: `interruptStaleTask` now takes the caller's `directory` and, when `abortWithTimeout` returns false, calls `checkSessionExistence` before returning. Only `"missing"` proceeds to finalization; `"exists"` and `"unknown"` keep the existing skip-and-retry behavior and the existing log line.
- `packages/omo-opencode/src/features/background-agent/task-poller-abort-failure.test.ts`: new regression test for that path.

## QA & Evidence

- **What was tested:** `bun test packages/omo-opencode/src/features/background-agent/task-poller-abort-failure.test.ts` on a running task whose session is absent from the status registry, whose `session.get` answers 404, and whose `session.abort` rejects.
  **Observed result:** RED before the change, `Expected: "cancelled" / Received: "running"`. GREEN after: 1 pass, and `notifyParentSession` called once.
  **Why sufficient:** the assertion fails on exactly the behavior this PR changes, and it observes the user-visible outcome (task finalized, parent told) rather than the internal call.
- **What was tested:** `bun test packages/omo-opencode/src/features/background-agent/` as the regression control.
  **Observed result:** 750 pass, 0 fail across 61 files.
  **Why sufficient:** the surrounding stale-task, parent-wake and session-status suites are unchanged, so the new path did not loosen the existing ones.
  **Observed result:** exit 0.

## Risks & Residuals

- A session that still exists, or whose existence cannot be determined, is unaffected: the early return, its log line and the retry path are byte-identical for those two cases. Mitigated by the control run above.
- `checkSessionExistence` adds one `session.get` call, and only on the path where an abort has already failed. Accepted.
- Open PR #5656 also touches `interruptStaleTask`, but refactors terminalization through a `completeTask` callback and leaves this `if (!aborted) return` early return untouched, so the two changes are independent. If #5656 lands first I will rebase onto it.

## Automated Checks

```bash
bun run typecheck
```

## Related Issues

None open for this root cause. Searched `interruptStaleTask`, `interruptStaleTask abortWithTimeout`, `abortWithTimeout` and `stale task never finalized running forever` across open issues and PRs; the nearest open items (#6263, #7265, #5656) are different root causes in the same files.

The companion case is #7850: an unreachable server currently answers `unknown` to the same existence check, so that PR is what makes this one fire when the endpoint is down rather than 404-ing. They are independent and can merge in either order; merging this one alone already fixes the confirmed-gone case.

---
## Summary by cubic
Fixes stale background tasks staying `running` forever when their child session is already gone and abort can never succeed. `interruptStaleTask` now confirms the session still exists before backing off; a confirmed-missing session is finalized and the parent is notified, while `exists` and `unknown` keep the existing retry behavior.

**Changes**
- Threads `directory` into `interruptStaleTask` and calls `checkSessionExistence` only when the abort fails.
- Adds a regression test asserting the task is cancelled and the parent notified when the session returns 404 and abort rejects.

<sup>Written for commit 98385e2509bfec0e98ba73a128e6a195a360f4ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/task-poller-abort-failure.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/task-poller-abort-failure.test.ts
@@ -1,0 +1,56 @@
+import { afterEach, describe, expect, mock, spyOn, test } from "bun:test"
+import { unsafeTestValue } from "../../../../../test-support/unsafe-test-value"
+import { checkAndInterruptStaleTasks } from "./task-poller"
+import type { BackgroundTask } from "./types"
+
+describe("checkAndInterruptStaleTasks abort failure on a confirmed-gone session", () => {
+  const originalDateNow = Date.now
+  const fixedTime = new Date("2026-09-07T03:00:00.000Z").getTime()
+
+  afterEach(() => {
+    Date.now = originalDateNow
+  })
+
+  test("finalizes a stale task whose session is gone even when abort fails", async () => {
+    //#given - the session is gone from the status registry and from storage, and abort can no longer reach it
+    spyOn(globalThis.Date, "now").mockReturnValue(fixedTime)
+    const staleActivity = new Date(fixedTime - 45 * 60 * 1000)
+    const task: BackgroundTask = {
+      id: "task-1",
+      sessionId: "ses-1",
+      parentSessionId: "parent-ses-1",
+      parentMessageId: "msg-1",
+      description: "test",
+      prompt: "test",
+      agent: "explore",
+      status: "running",
+      startedAt: staleActivity,
+      progress: { toolCalls: 2, lastUpdate: staleActivity },
+      consecutiveMissedPolls: 3,
+    }
+    const client = unsafeTestValue<Parameters<typeof checkAndInterruptStaleTasks>[0]["client"]>({
+      session: {
+        abort: mock(() => Promise.reject(new Error("session abort failed"))),
+        get: mock(() => Promise.resolve({ error: { status: 404, message: "session not found" } })),
+      },
+    })
+    const notify = mock(() => Promise.resolve())
+
+    //#when - the stale sweep runs well past the session-gone timeout
+    await checkAndInterruptStaleTasks({
+      tasks: [task],
+      client,
+      config: { staleTimeoutMs: 60_000, sessionGoneTimeoutMs: 60_000 },
+      concurrencyManager: unsafeTestValue<Parameters<typeof checkAndInterruptStaleTasks>[0]["concurrencyManager"]>({
+        release: mock(() => {}),
+      }),
+      notifyParentSession: notify,
+      sessionStatuses: {},
+    })
+
+    //#then - the task is cancelled and the parent is told, instead of waiting on a session that no longer exists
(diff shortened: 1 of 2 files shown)


LilMGenius:
@code-yeongyu, this one is the second half of the pair with #7850, where I left the full reproduction and the ordering note.

On its own it fixes a narrower case that does not need an unreachable server: when a session is already gone and `abortWithTimeout` therefore fails, `interruptStaleTask` returns early and the task stays `running` forever. It now confirms existence before backing off — `"exists"` and `"unknown"` keep the current retry path and its log line unchanged, only a confirmed-missing session is finalized and the parent notified.

RED before the change was `Expected: "cancelled" / Received: "running"`; GREEN after, with `bun test packages/omo-opencode/src/features/background-agent/` at 750 pass, 0 fail. Checks are green and it is MERGEABLE. Either order with #7850 works.

code-yeongyu (approved):
Approving. Verified together with #7850 on the combined tree; details in my review there.

Specific to this PR:

- The `exists` / `unknown` branches keep the original log line and early return byte-for-byte, so the retry behavior for a live session is unchanged — confirmed by reading the diff, not just the description.
- The second `checkSessionExistence` call is redundant on the `sessionGone` path (the caller just asked the same question a few lines earlier), but it costs one request only after an abort has already failed, and it is what makes the `sessionStatuses === undefined` path (status endpoint itself unreachable) work at all. Correct trade.
- Combined probe: stale task + fully unreachable client -> `cancelled`, parent notified once; fresh task -> stays `running`. Background-agent suite on the combined tree: 757 pass, 0 fail across 63 files (bun 1.4.0, Linux box).

Merging #7850 first, then this one.

code-yeongyu:
Reviewed and approved together with #7850 (verification details and the combined end-to-end probe are in my review there). Confirmed here that the `exists` / `unknown` branches keep the retry path and its log line byte-for-byte, and that this post-abort re-check is exactly what lets the unreachable-endpoint case finalize once #7850's readers answer `missing`. Merging #7850 first, then this one.

PR #7855, opened 2026-09-06
fix(lsp-daemon): clean up dead owner when the endpoint socket vanished (fixes #7827)

### LINKED ISSUE #7827: lsp-daemon: startup deadlocks on owner_changed_during_cleanup after reboot (dead owner never cleaned)
## Summary

After a machine reboot, the LSP daemon repeatedly fails to start with `DaemonStartupDeferredError: owner_changed_during_cleanup`, and the stale owner metadata of the dead daemon is **never** cleaned up. `lsp_diagnostics` and the other LSP tools return "LSP daemon unreachable" until one of the periodic sweeps (`lsp-proxy-sweep` / stale-version sweep) happens to wipe the stale metadata, after which the daemon starts again. The bug has reproduced on multiple machines across several reboots.

## Environment

- oh-my-openagent (package `oh-my-openagent` under opencode)
- Linux, `~/.omo/lsp-daemon/v0.1.0/`
- Node 24, two opencode sessions running concurrently (two `cli.js mcp` proxies)

## Observed state

After a reboot, `~/.omo/lsp-daemon/v0.1.0/` contains metadata of the pre-reboot (dead) daemon:

```
daemon.owner  -> {"pid":<dead-pid>, ...}
daemon.pid    -> <dead-pid>
daemon.endpoint, daemon.sock (stale)
```

Every `lsp_*` request then spawns a new daemon child, which dies with this stack (looping in `daemon.log` until a sweep intervenes):

```
DaemonStartupDeferredError: LSP daemon startup deferred: owner_changed_during_cleanup
    at validateExistingOwner (lsp-daemon/dist/cli.js)
    at async acquireStartupLease
    at async startDaemonServer
```

`ensureDaemonRunning` times out (`DaemonUnreachableError`) and the MCP tools report "LSP daemon did not become reachable".

## Root cause

`packages/lsp-daemon/src/ownership.ts`, `validateExistingOwner()`:

```ts
if (isProcessAlive(owner.pid)) throw new DaemonStartupDeferredError("owner_pid_live_unreachable");
const reread = readDaemonOwner(paths);
const endpoint = endpointIdentity(owner.endpoint.path);
if (!reread || reread.nonce !== owner.nonce || !sameEndpoint(endpoint, owner.endpoint)) {
  throw new DaemonStartupDeferredError("owner_changed_during_cleanup");   // <-- thrown BEFORE cleanup
}
cleanupDeadOwner(paths, owner);                                           // <-- never reached
```

Two compounding problems:

1. **Cleanup is skipped for the exact case it is meant to handle.** When the owner file changes between the initial read and the re-read, the error is thrown *before* `cleanupDeadOwner()`, so the stale metadata of the dead owner is never removed. Since a reboot always leaves a dead owner file behind, the next daemon startup is guaranteed to hit this path — and can never self-heal on its own.

2. **Two concurrent MCP proxies race the owner file.** With two opencode sessions, each `lsp_*` request can spawn a daemon child; the two children interleave their `readDaemonOwner()` / re-read / (would-be) cleanup, so the nonce check fails intermittently and the loop in `daemon.log` is sustained. The startup lease (`daemon.lock`) does not prevent this because the lock is released before the error propagates (`acquireStartupLease` catches and `lock.release()`s), and the ping re-read window is async.

Result: LSP tools stay broken after every reboot until the periodic `lsp-proxy-sweep`/stale-version sweep coincidentally removes the stale files — non-deterministic self-healing.

## Proposed fix

In `validateExistingOwner()`:

- On re-read mismatch where the *re-read* owner's pid is also dead (or the file is gone), run `cleanupDeadOwner()` for the verified-dead owner before deferring, so the next attempt (or the immediate retry) starts clean.
- Re-check under the startup lease and retry once instead of throwing after a single mismatch; the lock is already held during `validateExistingOwner()`, so the mismatch should be re-evaluated rather than escalated.
- Consider having the daemon remove its own metadata on `SIGTERM`/`SIGKILL`-recoverable shutdown paths, and treating "owner file exists but pid dead and socket refused" as an unconditional cleanup case.

## Workaround

```sh
D=~/.omo/lsp-daemon/v0.1.0
rm -f $D/daemon.owner $D/daemon.pid $D/daemon.endpoint $D/daemon.sock $D/daemon.lock
```

After that, the next `lsp_*` request starts a fresh daemon successfully.

MoerAI:
## Summary

After a reboot the LSP daemon could never start again: it failed on `DaemonStartupDeferredError: owner_changed_during_cleanup` in a loop, and the dead daemon's metadata was never cleaned. This makes the dead-owner cleanup actually run in the case it was written for.

## Root Cause

`writeDaemonOwner` persists the owner as `JSON.stringify(owner)`, where `endpoint` is a concrete `{ kind: "unix", path, dev, ino }`. `parseEndpoint` only re-derives the identity from disk when `kind` is **absent** (`ownership.ts:190`); when `kind === "unix"` it returns the persisted value verbatim (`ownership.ts:195`).

So after a reboot, inside `validateExistingOwner`:

```ts
if (isProcessAlive(owner.pid)) throw new DaemonStartupDeferredError("owner_pid_live_unreachable"); // owner is dead, falls through
const reread = readDaemonOwner(paths);
const endpoint = endpointIdentity(owner.endpoint.path);   // socket file is gone -> { kind: "missing" }
if (!reread || reread.nonce !== owner.nonce || !sameEndpoint(endpoint, owner.endpoint)) {
  throw new DaemonStartupDeferredError("owner_changed_during_cleanup");   // <- thrown here
}
cleanupDeadOwner(paths, owner);   // <- never reached
```

`sameEndpoint` returns `false` immediately on the `kind` mismatch (`missing` vs the persisted `unix`), so it throws before `cleanupDeadOwner` can remove `daemon.owner` / `daemon.pid` / `daemon.endpoint`. Every later startup repeats this, until an unrelated sweep happens to wipe the directory.

Note this branch is only reached once `isProcessAlive(owner.pid)` is already `false`, so the owner is definitively dead.

## Changes

| File | Change |
|------|--------|
| `packages/lsp-daemon/src/ownership.ts` | Defer only on a genuine rebind: treat a vanished (`missing`) endpoint as cleanup-eligible instead of as an owner-change race. The nonce re-read guard is unchanged. |
| `packages/lsp-daemon/test/auth-ownership.test.ts` | Regression test: dead pid + persisted `{ kind: "unix", ... }` endpoint + no socket file on disk. |

The endpoint comparison exists to catch the endpoint being rebound to a **different still-existing** endpoint. A vanished endpoint means nothing is bound, so the dead owner's metadata is safe to clean.

Windows is unaffected: `endpointIdentity` returns `{ kind: "windows" }` there and never `"missing"`, so the new guard is a no-op. The test is skipped on win32 for the same reason.

## Reproduction (before fix)

Reverting only the 2-line source change, keeping the new test:

```
FAIL  test/auth-ownership.test.ts > daemon IPC authentication and ownership >
      given a dead owner with a missing Unix socket when candidate starts then it rotates auth and takes ownership
DaemonStartupDeferredError: LSP daemon startup deferred: owner_changed_during_cleanup
    147|    throw new DaemonStartupDeferredError("owner_changed_during_cleanup"...

Tests  1 failed | 12 passed (13)
```

## Verification (after fix)

```
Test Files  1 passed (1)
     Tests  13 passed (13)
```

The pre-existing neighbours still pass, including `given live unreachable owner metadata when candidate starts then it defers without unlinking endpoint` (proves live owners still defer) and `given a dead unchanged owner when candidate starts then it rotates auth and takes ownership`.

## Test

- Regression test: `packages/lsp-daemon/test/auth-ownership.test.ts` (newly added, Unix-guarded)
- Package suite: `npm run test` in `packages/lsp-daemon` - 162 pass across 22 files
- Typecheck: `tsc --noEmit` clean
- Note: `npm run check` also runs `biome check .`, which reports 2 pre-existing errors on these files. I verified this is unrelated to this change: stashing the change and running biome on the pristine `dev` files reports the same 2 errors. I left them alone rather than reformatting unrelated code.

Fixes #7827

---
## Summary by cubic
Fixes the LSP daemon startup deadlock after a reboot where a dead owner's stale metadata was never cleaned, so the daemon now starts again instead of throwing `owner_changed_during_cleanup` in a loop.

- Defer only when the endpoint was rebound to a different still-existing socket; a vanished socket means the stale metadata is safe to remove.
- The nonce re-read race guard is unchanged.
- Adds a regression test for a dead pid plus a persisted Unix endpoint with no socket file on disk; Windows is unaffected.
- Adds sanitized replay evidence under `.omo/evidence/`.

<sup>Written for commit 5d397e798a095e35496a01d689ae9d110ba7a61f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7855?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260908-7855-lsp-restart/README.md
+++ b/.omo/evidence/20260908-7855-lsp-restart/README.md
@@ -1,0 +1,92 @@
+# PR 7855: real vanished-endpoint restart proof
+
+## Scope and result
+
+Addresses review [discussion_r3944865688](https://github.com/code-yeongyu/oh-my-openagent/pull/7855#discussion_r3944865688)
+on source head `da130a5550e78c2181f412fb7373e4c1ba2e8d4b`, for issue #7827.
+This remediation adds evidence only; it does not change ownership behavior.
+
+**PASS:** real OpenCode 1.18.4 invoked `lsp_status` through the PR-built daemon's
+stdio MCP proxy after a dead owner's persisted Unix endpoint had disappeared.
+The daemon replaced PID **692** / nonce `bcce7221-e59b-4fea-826f-91be257eab19`
+with live PID **743** / nonce `f605a639-eb96-40a3-9897-39ff96bee275`, recreated
+the socket, rotated authentication, answered an authenticated owner ping and
+tool call, and rejected the old token. No token is included in this evidence.
+
+## What was tested
+
+The matching OpenCode QA guidance selected an isolated server/tool surface,
+not a TUI smoke or an in-process daemon substitute. `replay.mjs` drives:
+
+1. A task-owned child that exits naturally; its PID is confirmed dead by ESRCH.
+2. A real temporary Unix socket whose device/inode identity is captured before
+   closing it. The socket is confirmed absent before stale owner, PID, endpoint
+   and auth metadata are written. An authenticated ping fails with ENOENT.
+3. Real `opencode serve`, configured with a local MCP command pointing directly
+   at this worktree's built `packages/lsp-daemon/dist/cli.js`, not an installed
+   or main-checkout runtime. Its bundle SHA-256 is in `result.json`.
+4. A synchronous HTTP prompt whose deterministic local provider selects the
+   actual LSP tool advertised by OpenCode. The completed tool part is read back
+   from OpenCode's session messages; the model and daemon are not mocked.
+5. Replacement-owner and authentication checks, followed by owned-process and
+   sandbox teardown. Event waits are registered before triggers and bounded;
+   the driver has no fixed sleeps or polling delays.
+
+Only the model responses are scripted. The test used Node 24.18.0 and Bun 1.4.0
+inside an ARM64 disposable `omo-qa` container. Docker inspection confirmed its
+only host mount was the task repository; no real-home configuration, credential,
+database, or daemon state was mounted. One isolated session was created and
+removed with the sandbox. The container was removed after capture.
+
+## Replay
+
+Use a disposable ARM64 container with OpenCode, Node, Bun 1.4.0, and sqlite3.
+Mount a checkout of the tested PR at `/repo`, with no host configuration mounts.
+Install workspace dependencies and build this checkout's daemon, then run:
+
+```sh
+npm --prefix packages/lsp-daemon ci
+npm --prefix packages/lsp-daemon run typecheck
+npm --prefix packages/lsp-daemon run build
+node .omo/evidence/20260908-7855-lsp-restart/replay.mjs /repo
+```
+
+Run the driver **inside the disposable container**, not against a real home.
+The command emits sanitized JSON and fails if the required assertions fail.
+The provider is bound to loopback; no live model account is needed.
+
+## Captures and validation
+
+- `result.json`: exact successful harness tool output, before/after ownership,
+  auth booleans, runtime digest, isolation and cleanup receipts.
+- `checks.txt`: successful typecheck/build, lead-provided 13/13 focused test
+  result, independent 162/162 package-suite result, and all observed setup,
+  driver and lint failures with their disposition.
+- `replay.mjs`: reviewer-repeatable driver used for the successful capture.
+
+The package lint command still reports **11 existing format/import-order
+errors** on unchanged PR source/test files. They are not hidden or fixed by this
+evidence-only patch. The root automatic prepare build hit a container Git-pointer
+limitation; only the relevant daemon build is claimed green.
+
+## Why this is enough, and limits
+
+The reported failure happens before daemon startup can reclaim metadata. The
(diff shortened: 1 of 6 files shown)


MoerAI:

MoerAI:
Closing in favor of #8014, which covers both missing sockets and socket-identity drift after reboot for #7827. This PR only handles the missing-socket case, so the broader fix supersedes it.

Issue #7666, opened 2026-09-03
Startup config migration: existing-target-wins merge silently discards model overrides from legacy config files


dougstanford:
## Summary

The `2026-07-opencode-config-unification` startup migration merges a legacy config file (e.g. `~/.config/opencode/oh-my-opencode.jsonc`) into `~/.omo/omo.jsonc` with **existing-target-wins** semantics, then deletes the legacy source. Any values that already exist in the target silently win — user overrides in the legacy file are discarded without warning.

## Environment

- oh-my-openagent v4.19.4 (also reproduced on the pre-update version)
- macOS, opencode 1.18.26

## Reproduction

1. Plugin startup migration creates `~/.omo/omo.jsonc` with default model mappings (in our case every agent/category mapped to `opencode/gpt-5-nano`, a model that does not exist in this environment).
2. User writes corrected model routing to the legacy path `~/.config/opencode/oh-my-opencode.jsonc` (a name still listed in `CONFIG_FILE_NAMES` in `dist/config-migration/discovery-paths.d.ts`).
3. On next opencode restart, the migration:
   - backs the legacy file up to `~/.omo/migration-backup-<timestamp>-opencode-config/`
   - merges it into `~/.omo/omo.jsonc` — but only keys **absent** from the target take effect; every key present in both keeps the target's (broken) value
   - deletes the legacy source file
4. Result: the user's fix silently vanishes, the broken defaults remain, and there is no log/diagnostic indicating overrides were dropped.

Observed in our case: of 20 model mappings in the legacy file, only the 2 keys absent from the target (`build`, `sisyphus`) survived the merge; the other 18 kept the dead `opencode/gpt-5-nano` values.

## Expected behavior

One or more of:
- Legacy-source values should win over existing target values (the user explicitly wrote them later), or at minimum conflicting keys should be reported
- A diagnostic/log line when the migration drops conflicting overrides
- The migration result summary should list which keys were skipped due to conflicts

## Workaround

Edit `~/.omo/omo.jsonc` directly. Note the legacy file cannot be used to override an existing omo.jsonc at all.

code-yeongyu:
Thanks for the clear reproduction steps. The existing-target-wins merge order is intentional to avoid clobbering manual edits in the target, but the silent discard of conflicting overrides is a real UX gap — especially when the legacy file has fresher, user-corrected values.

At minimum we should emit a diagnostic log when conflicting keys are dropped during migration. Labeling for triage.

PR #7856, opened 2026-09-06
fix(config-migration): log conflicting override paths

MoerAI:
## Summary

Log the key paths of migration conflicts that reach the no-clobber merge.
Existing target values still win. The current implementation logs paths only,
not the discarded or retained values.

## Changes

`create-plugin-module.ts` derives `skippedConflictPaths` from existing
`skipped:` diagnostics and adds the paths to the startup log alongside
`skippedConflictCount`. The factory tests cover path extraction and exclusion
of synthetic credential values. No shared migration type, merge policy or
toast behavior is changed.

## Verification

- Refreshed onto dev `776c4668d` with no new production logic.
- Bun 1.4.0 / Node 24: 75 focused tests, full typecheck and full build passed.
- Real isolated Docker OpenCode 1.18.30 emitted the actual default factory startup log with `skippedConflictCount: 2` and the two conflict paths, without their values. The target values remained unchanged and the migration toast was captured over SSE.
- Host database sessions remained 8104 before/after; isolated database contained zero sessions. No private config, skills or credentials were mounted.
- Replay and audited captures: `.omo/evidence/20260916-pr7856-refresh/`.
- Earlier unsuccessful startup probes remain failed historical evidence; the new receipt proves this scoped logging path, not the full agents/categories omission issue.

## Scope limitation

Since #7270, `transformLegacyKeys` in `config-migration/transform-opencode.ts`
deletes legacy `agents` and `categories` before the no-clobber merge. Those
omissions never reach this diagnostic and are not fixed here. A separate
transform-level omission diagnostic is needed for that path.

This PR therefore does not claim to resolve the full #7666 report. Its scope
is conflict paths that reach the merge, including safe handling of
credential-bearing configuration values.

Related to #7666.

### DIFF
--- a/packages/omo-opencode/src/testing/create-plugin-module.test.ts
+++ b/packages/omo-opencode/src/testing/create-plugin-module.test.ts
@@ -409,8 +409,40 @@ describe("createPluginModule()", () => {
       // then
       expect(log).toHaveBeenCalledWith("[config-migration] startup completed", expect.objectContaining({
         skippedConflictCount: 1,
-        skippedConflicts: [skippedConflict],
+        skippedConflictPaths: ["agents.sisyphus.model"],
       }))
+    })
+
+    it("#given a conflicting override holds a credential #then only the key path is logged", async () => {
+      // given
+      const log = mock(() => {})
+      const pluginModule = createTestPluginModule({
+        log,
+        runOpenCodeStartupMigration: () => ({
+          journalResumed: false,
+          migratedFrom: ["/home/alice/.config/opencode/oh-my-opencode.json"],
+          reloadRequired: true,
+          results: [{
+            diagnostics: ["skipped: openclaw.discordBotToken legacy=\"legacy-secret\" kept=\"kept-secret\""],
+            journalResumed: false,
+            status: "migrated",
+          }],
+          skippedConflictCount: 1,
+        }),
+      })
+      mockLoadPluginConfig.mockReturnValue({})
+
+      // when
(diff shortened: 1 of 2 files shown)


MoerAI:

PeterPonyu:
One caveat from re-checking #7666 against `dev@4ec56b79d` that may be worth a line in this PR or a follow-up: since #7270 (2026-08-25, after the reporter's 4.19.4) `packages/omo-opencode/src/config-migration/transform-opencode.ts:22-23` deletes legacy `agents` and `categories` outright before the merge runs, with no diagnostic. So a v5 user repeating the issue's exact steps (legacy `agents.*.model` overrides) never reaches `mergeWithoutClobber` and would still see silence even with the dropped-key paths logged here — the diagnostic that would cover them is an `omitted: agents, categories` line from the transform itself. The path-only logging here (and the credential test) looks right to me; just noting the gap so the issue isn't closed on a case it no longer covers.

MoerAI:
Thanks for re-checking. I confirmed that transformLegacyKeys drops agents/categories before the no-clobber merge. I corrected the title and body and removed the issue-closing claim. This PR is limited to key-path diagnostics for conflicts that reach the merge; transform-level omission diagnostics remain outside it. I also corrected the stale body text: the current implementation logs paths only, never conflicting values.

MoerAI:
Closing after more than 14 days without maintainer engagement. The verified fix and evidence remain available in this PR history. Happy to reopen or rework if this fix is still wanted.

PR #7860, opened 2026-09-06
fix(script): extract the published tarball with an archive path GNU tar can open

LilMGenius:
## What breaks

A contributor on Windows cannot run the published-payload smoke at all from Git Bash. It dies before it installs anything:

```
$ node script/published-install-smoke.mjs --package=oh-my-openagent@beta
oh-my-openagent-5.0.0-beta.45.tgz
tar (child): Cannot connect to C: resolve failed
published install smoke failed: Command failed: tar -xzf C:\Users\...\omo-published-smoke-ZscYNE\oh-my-openagent-5.0.0-beta.45.tgz -C C:\Users\...\extracted
```

The same shape fails on a GitHub `windows-latest` runner for all four specs the CI loop iterates. I observed that directly through a fork-only run, since `lazycodex-published-smoke` is `runs-on: ubuntu-latest` and never exercises this driver on Windows.

## Root cause

`extractTarball` in `script/published-install-smoke.mjs` passed the packed tarball to `tar` as an absolute path. GNU tar reads a colon before the first slash as `host:path`, and a Windows absolute path carries one at index 1 with no forward slash to end the host, so the archive name is taken for a remote machine. `tar` is GNU tar wherever Git Bash is on PATH, which includes the bash shell GitHub Actions runs on windows runners.

It stayed invisible on a Windows workstation because bare `tar` on a default PATH is `C:\Windows\system32\tar.exe`, a bsdtar that accepts drive letters.

## Why not `--force-local`

Measured over three identical rounds, spawning GNU tar 1.35 and bsdtar 3.8.4 from node with only the archive argument varying:

| tar | `-f` argument | result |
| --- | --- | --- |
| GNU | absolute | exit 2, `Cannot connect to C: resolve failed` |
| GNU | relative to cwd | exit 0 |
| GNU | absolute, plus `--force-local` | exit 0 |
| bsdtar | absolute, plus `--force-local` | exit 1, `Option --force-local is not supported` |

The flag repairs GNU tar and bsdtar rejects it. bsdtar is the `tar` on macOS and on a default Windows PATH, so `--force-local` would trade this failure for a different one. A relative archive argument is accepted by both.

## RED / GREEN

| Input | Before | After |
| --- | --- | --- |
| `--package=oh-my-openagent@beta`, GNU tar first on PATH | exit 1, `Cannot connect to C:` | exit 0, `published install smoke OK (1 plugin(s), ...)` |
| `--package=lazycodex-ai@latest`, GNU tar first on PATH | not run before the fix | exit 0 |
| `--package=oh-my-openagent@beta`, default PATH (bsdtar) | exit 0 | exit 0 |
| `bun test script/published-install-smoke.test.ts script/lazycodex-published-smoke-workflow.test.ts` | 6 tests | 8 pass / 0 fail |

The bsdtar row is the control: already passing, still passing, so this is not a loosening of the extraction.

`bun test script/` is not green on my machine either way. At upstream/dev with these three files reverted it is 449 pass / 7 fail; with the fix it is 451 pass / 7 fail, failing on the same six unrelated cases (generated Codex installer version, embedded manifest parity darwin-arm64, workflow job summaries, omo-ai payload verifier, two publish.yml checks). None touches the extraction path.

Evidence, including every log quoted here: `.omo/evidence/20260907-published-smoke-tar-windows/`.

## The fix

`extractTarball` runs `tar` from the working directory and addresses the archive relative to it. Non-Windows behavior is unchanged, since a relative archive name is what both tars want anyway.

The tests were rewritten to pack a real tarball and extract it through `extractTarball`, rather than asserting the argument array. Three mutations fail them: both paths back to absolute (2 fail under GNU tar), `cwd` dropped from the spawn (2 fail), archive reduced to a bare basename (1 fail).

## Worth checking in review

The destination is made relative in the same call. Several standalone probes show an absolute `-C` failing under GNU tar, but the same shape reached through `extractTarball` succeeds, so that half is not reproducible on demand and a mutation reverting only the destination is **not** caught by the suite. I kept it relative because nothing shows a cost, not because I measured a stable failure. The archive argument is the part that failed in every probe, in the driver, and on the runner.

Known limit: `relative()` returns an absolute path across drives and for UNC targets, so `--tarball=D:\...` while the temp directory is on `C:` would still carry a colon. Nothing in the repo passes `--tarball`.

- No CI change. I am not proposing a Windows leg for `lazycodex-published-smoke`; the fork run was only how I found this.
- The same absolute-path shape exists in shipped runtime code, including `binary-downloader.ts`, `zip-extractor.ts`, `tools/grep/downloader.ts` and the senpi ast-grep QA script. A Git Bash user hits the identical failure there. That is a larger change against user-facing code and belongs in its own PR with its own reproduction, so this one stays inside `script/`.

---
## Summary by cubic
Fixes the published-install smoke extraction so the tarball can be opened by GNU tar on Windows, where a drive letter was previously read as a remote host and the smoke died before installing anything. The extraction now runs tar from the working directory and passes archive and destination paths relative to it, which both GNU tar and bsdtar accept.

- Tests now pack a real tarball and extract it through the helper instead of asserting the argument array.
- Adds evidence logs for the Windows runner failure, isolation runs, and mutation checks.

<sup>Written for commit 3580eca30f90c3ed43f1b52952e3a79b35627bc0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:
@code-yeongyu ready for review when you have a slot. All checks are green on this one.

What it fixes: `node script/published-install-smoke.mjs --package=<spec>` cannot extract at all from Git Bash on Windows. `extractTarball` passed the packed tarball as an absolute path, and GNU tar reads the colon before the first slash as a remote host, so it dies at `tar (child): Cannot connect to C: resolve failed` before installing anything. The bash shell GitHub Actions runs on windows runners resolves the same GNU tar.

Verified against upstream/dev at `ad62603f0`:

| Input | Before | After |
| --- | --- | --- |
| `--package=oh-my-openagent@beta`, GNU tar first on PATH | exit 1, `Cannot connect to C:` | exit 0 |
| `--package=lazycodex-ai@latest`, GNU tar first on PATH | not run before the fix | exit 0 |
| `--package=oh-my-openagent@beta`, default PATH (bsdtar) | exit 0 | exit 0 |

The bsdtar row is the control, which is also why this never showed on a plain Windows shell: bare `tar` there is `C:\Windows\system32\tar.exe`.

Two things I would rather you hear from me than find: `--force-local` looks like the obvious fix and is not one, since bsdtar rejects it with `Option --force-local is not supported` and bsdtar is the `tar` on macOS. And the destination argument is made relative in the same call without a stable measurement behind it; a mutation reverting only the destination is not caught by the tests, and I said so in the body rather than papering over it.

No CI change here. I found this by running the driver on `windows-latest` in my own fork, and left `lazycodex-published-smoke` alone.

The same shape is in shipped runtime code, where a Git Bash user hits it for real rather than only in contributor tooling. That is #7861, filed separately with its own reproduction. No ordering between the two.

LilMGenius:
Closing this myself.

I opened far too many pull requests here in one day and buried the maintainer in notifications on top of it. Contributor tooling rather than anything a user hits. It added noise to a queue that was already too long because of me.

PR #7861, opened 2026-09-06
fix(zip): extract Windows zips with the system bsdtar instead of whatever tar PATH resolves

LilMGenius:
## What breaks

On Windows, launching OMO from Git Bash makes every zip download fail before it extracts anything:

```
zip entry listing failed (exit 128): tar: Cannot connect to C: resolve failed
```

That is the ripgrep auto-install for the grep tool and the comment-checker binary download, both of which take the win32 zip path. Both spawn sites are in the published `oh-my-openagent@5.0.0-beta.45` tarball at `dist/index.js` lines 7812 and 27659.

## Root cause

`getWindowsZipExtractor()` picks `"tar"` from the OS build number alone, and `listZipEntriesWithTar` and `extractZip` then spawn a bare `tar`. That resolves through PATH, and Git Bash puts Git for Windows' `usr/bin` first, so the process gets GNU tar 1.35 rather than the bsdtar Windows ships at `System32\tar.exe`. GNU tar reads the colon before the first slash as `host:path`.

Making the paths relative does not fix it. GNU tar has no zip support at all:

| tar | call | result |
| --- | --- | --- |
| bsdtar | `-xf <abs zip> -C <abs dir>` | exit 0 |
| GNU tar | `-xf <abs zip> -C <abs dir>` | exit 128, `Cannot connect to C: resolve failed` |
| GNU tar | `-tf <abs zip>` | exit 128, same |
| GNU tar | `-xf probe.zip` from the archive's own directory | exit 2, `This does not look like a tar archive` |

Only the Windows bsdtar can do this job, so the binary has to be resolved rather than looked up.

## RED / GREEN

Driving the real `extractZip` over a real zip, against upstream/dev at `ad62603f0`:

| Input | Before | After |
| --- | --- | --- |
| `extractZip`, Git Bash PATH (GNU tar first) | throws `zip entry listing failed (exit 128)` | resolves, payload extracted |
| `extractZip`, default PATH (bsdtar first) | resolves | resolves |
| `bun test packages/utils/src/zip-entry-listing/ packages/omo-opencode/src/shared/archive-entry-validator.test.ts` | 10 tests | 13 pass / 0 fail |

The default-PATH row is the control: already passing, still passing.

Evidence including every log quoted here: `.omo/evidence/20260907-windows-zip-extractor-tar/`.

## The fix

One resolver, `windowsSystemTarPath()`, returning `%SystemRoot%\System32\tar.exe` when it exists and `null` otherwise. The listing and the extraction both spawn it, and `getWindowsZipExtractor()` selects `"tar"` only when that binary is present, so a machine without it falls through to the PowerShell ladder that already exists. Non-Windows behavior is untouched: the resolver returns `null` off win32 and the bare name is used exactly as before.

Two mutations fail the new tests: returning the bare `"tar"` again gives `Executable not found in $PATH: "tar"` when PATH is emptied, and dropping the existence check breaks the SystemRoot fallback.

- `binary-downloader.ts` passes the same absolute shape for `.tar.gz`, but both callers select the zip asset on win32, so it is unreachable today and left alone.
- The senpi QA scripts carry it too and are contributor-only.
- No CI change.

---
## Summary by cubic
Fixes Windows zip extraction so it no longer fails when launched from Git Bash. The code now resolves the Windows System32 bsdtar directly instead of spawning a bare `tar` that PATH could redirect to GNU tar, which cannot read zips and rejects drive-letter paths with `Cannot connect to C: resolve failed`.

**Bug Fixes**
- Both affected call sites ship in the published `oh-my-openagent@5.0.0-beta.45` tarball at `dist/index.js` lines 7812 and 27659 (ripgrep auto-install and comment-checker download).
- `getWindowsZipExtractor()` now selects tar only when the System32 binary exists, otherwise falling back to the existing PowerShell ladder.
- Non-Windows behavior is unchanged; the resolver returns null off win32 and the bare name is used as before.
- New tests cover the resolution, PATH-independent execution, and the missing-System32-tar fallback.

<sup>Written for commit 95e23562ebf8d4c58b8eee6ca38d3d5bc1428cc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:

Every zip download fails on Windows when OMO runs from Git Bash:

```
zip entry listing failed (exit 128): tar: Cannot connect to C: resolve failed
```

That is the ripgrep auto-install behind the grep tool and the comment-checker binary download. Both spawn sites ship: published `5.0.0-beta.45` dist, `index.js` 7812 and 27659.

The win32 branch picks the extractor from the OS build number, then spawns a bare `tar` assuming System32 bsdtar. Git Bash puts `usr/bin` first, so it gets GNU tar, which cannot read a zip at all. Relative paths are not the repair: `tar -xf probe.zip` from the archive's own directory still answers `This does not look like a tar archive`.

| `extractZip` over a real zip | Before | After |
| --- | --- | --- |
| Git Bash PATH | throws exit 128 | resolves, payload extracted |
| default PATH | resolves | resolves |

The second row is the control: already passing, still passing.

The fix resolves `%SystemRoot%\System32\tar.exe` and selects `tar` only when that binary exists, so a machine without it falls through to the PowerShell ladder already there. `--force-local` is not an alternative: bsdtar rejects it with `Option --force-local is not supported`, and bsdtar is the `tar` on macOS.

I found this by running the existing smoke job on `windows-latest` inside my own fork and then throwing the workflow edit away. No CI change here.

#7860 fixes the same tar rule in `script/published-install-smoke.mjs`. Different tree, no ordering between them.

LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `getWindowsZipExtractor()` in `packages/omo-opencode/src/shared/zip-extractor.ts` still spawns a bare `tar`, so a zip download launched from Git Bash still resolves GNU tar and fails.

PR #7862, opened 2026-09-06
fix(codex-install): let the lazycodex update check reach npm on Windows

LilMGenius:
## What breaks

`lazycodex update` can never find a newer version on Windows. On the published artifact:

```
$ $env:LAZYCODEX_CURRENT_VERSION = '1.0.0'
$ npx -y lazycodex-ai@latest --dry-run update
npm notice run lazycodex --dry-run update
Unable to check lazycodex-ai updates (unknown-latest).
$ echo $LASTEXITCODE
1
```

Supplying the version through the environment, so the lookup returns before spawning anything, makes the same command exit 0 and print the update it would run. Everything after the lookup is healthy.

## Root cause

`resolveLatestVersion` in `lazycodex-manual-update.ts` spawns npm with no shell. npm is `npm.cmd` on Windows and Node's spawn does not apply `PATHEXT` without one, so it never reaches npm. The failure then hides inside the guard: `spawnSync` reports `status: null` rather than a non-zero exit, so `result.status !== 0` is true and the function returns `undefined` as though npm had answered with nothing.

| call, under node on Windows | status | error | `status !== 0` | returns |
| --- | --- | --- | --- | --- |
| as shipped | `null` | `ENOENT spawnSync npm` | `true` | `undefined` |
| with `shell: true` | `0` | none | `false` | `"4.19.4"` |

`resolveLazyCodexUpdatePlan` turns the `undefined` into `reason: "unknown-latest"`, which is the message above and the exit 1.

## RED / GREEN

| Input | Before | After |
| --- | --- | --- |
| `spawnSync("npm", ["view", ...])` under node, Windows | `status null`, `ENOENT` | `status 0`, `"4.19.4"` |
| `bun test packages/omo-codex/src/install/npm-spawn-options.test.ts` | file did not exist | 3 pass / 0 fail |

The published CLI cannot be shown green without a release, so that chain stands in for it: the artifact fails, the failure isolates to the lookup, and the lookup is repaired under the runtime that runs it.

Evidence, including every log quoted here: `.omo/evidence/20260907-windows-npm-view-enoent/`.

## The fix

`npmSpawnOptions()` beside the caller, returning `{ shell: true }` on win32 and `{}` elsewhere, spread into the existing `spawnSync` options. Non-Windows behavior is byte-identical.

`script/npm-invocation.mjs` already holds this branch and says the platform decision should live in one place, but the installer bundle cannot reach that module graph, so this is a deliberate second copy next to its consumer rather than a cross-tree import.

## Worth checking in review

The four npm arguments are fixed literals with no interpolation, so the shell carries no caller-controlled input. Node still prints DEP0190 about argument escaping under `shell: true`; #7844 accepted the same warning for the same reason.

A mutation removing `...npmSpawnOptions()` from the call site is **not** caught by the test. The test spawns npm through node, which reproduces the failing regime, but nothing binds that one call site to the helper. The suite runs under bun, and bun resolves `npm.cmd` on its own, so a test spawning npm directly would have passed with or without this change; that is why it shells out to node.

- `packages/omo-codex/scripts/install-dist/install-local.mjs` is tracked and embeds this code, but rebuilding it on a clean `dev` with no source change already moves the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`. The checked-in bundle is two releases stale and CI regenerates it before the Codex suite, so committing a rebuild here would bury this change under an unrelated version bump. Only the source changes.
- The `bun` spawns in the same file are left alone: `bun` is an `.exe` on Windows and resolves without a shell.

---
## Summary by cubic
Fixes `lazycodex update` on Windows so it can reach npm and detect newer versions instead of always failing with "unknown-latest".

**Notes**
- On Windows, npm is `npm.cmd` and Node's spawn doesn't apply `PATHEXT` without a shell, so the lookup silently died with ENOENT; `npmSpawnOptions()` adds `shell: true` on win32 and `{}` elsewhere, leaving non-Windows behavior unchanged.
- The installer bundle now ships with the fix because the freshness test reads it from the git index; the rebuild also carries an unrelated `5.0.0-beta.43` → `5.0.0-beta.45` version bump from upstream staleness.
- Node prints a DEP0190 deprecation warning under `shell: true`; #7844 already accepted the same warning.
- The new test shells out to node because bun resolves `npm.cmd` on its own; removing the spread at the call site is not caught by the test.

<sup>Written for commit 5e8109d6b620891d3e674b43b4fab22ac551003d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:

`lazycodex update` never finds a newer version on Windows:

```
$ npx -y lazycodex-ai@latest --dry-run update
Unable to check lazycodex-ai updates (unknown-latest).
```

Exit 1. The same run with `LAZYCODEX_LATEST_VERSION` supplied prints the update command and exits 0, so everything after the lookup is healthy and the lookup is the only broken step.

`resolveLatestVersion` spawns npm with no shell. npm is `npm.cmd` and Node does not apply PATHEXT without one, so it never reaches npm. The failure then hides: `spawnSync` reports `status: null` rather than a non-zero exit, so the `status !== 0` guard returns `undefined` as though npm had answered with nothing.

| `spawnSync("npm", ["view", ...])` under node on Windows | status | returns |
| --- | --- | --- |
| as shipped | `null`, `ENOENT` | `undefined` |
| with the shell option | `0` | `"4.19.4"` |

Two things worth your eye. The generated installer bundle is in the diff because `codex-install-bundle-freshness.test.ts` reads it from the git index, and regenerating it also moves the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`; that line is pre-existing staleness riding along, not part of the fix. And a mutation removing `...npmSpawnOptions()` at the call site is not caught by the test, which the body states rather than claiming coverage it does not have.

The four npm arguments are fixed literals with no interpolation, so the shell carries no caller-controlled input; #7844 accepted the same DEP0190 warning for the same reason.

LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `resolveLatestVersion` in `packages/omo-codex/src/install/lazycodex-manual-update.ts` still calls `spawnSync("npm", ...)` with no shell, so the update check still cannot reach `npm.cmd` on Windows.

Issue #7473, opened 2026-08-29
[Bug][Windows] Generated .cmd shims abort with "The syntax of the command is incorrect." when NODE_REPL_NODE_PATH is set in config.toml


fvegiard:
## Summary

Every generated Windows command shim (`~/.local/bin/*.cmd` — `ulw.cmd`, `omo-agent-toolkit.cmd`, `omo-codegraph.cmd`, …) dies instantly with `The syntax of the command is incorrect.` and exit code 255 **whenever `~/.codex/config.toml` contains a `NODE_REPL_NODE_PATH` key** — which Codex Desktop sets by default (its `cua_node` runtime). The shim never reaches the runtime exec line, so all omo CLI entry points are unusable on such machines.

## Root cause

`windowsNodeDiscoveryLines()` in `packages/omo-codex/src/install/codex-cache-command-shim.ts` emits two caret-escaped quote comparisons **inside a parenthesized block**:

```bat
if defined OMO_NODE_REPL_NODE_PATH (
  ...
  if "!OMO_NODE_BINARY:~0,1!"=="^"" set "OMO_NODE_BINARY=!OMO_NODE_BINARY:~1!"
  if "!OMO_NODE_BINARY:~-1!"=="^"" set "OMO_NODE_BINARY=!OMO_NODE_BINARY:~0,-1!"
  ...
)
```

That construct is valid at top level, but when cmd.exe reads a multi-line `(...)` block it runs a separate parse phase in which the `^` in `"^""` desynchronizes quote tracking; the whole block — and the script — aborts with the syntax error.

## Minimal repro (Windows 11, any cmd.exe)

`fail.cmd`:
```bat
@echo on
setlocal EnableExtensions EnableDelayedExpansion
set "OMO_NODE_REPL_NODE_PATH= 'C:\some path\node.exe'"
if defined OMO_NODE_REPL_NODE_PATH (
  if "!OMO_NODE_REPL_NODE_PATH:~0,1!"=="^"" set "X=1"
)
echo DONE
```
→ `The syntax of the command is incorrect.`, `DONE` never prints. Replacing the `"^""` line with a plain line makes the same block succeed.

## Verified bisect results

- Same block **without** the two `"^""` lines: works, strips quotes correctly.
- Block with **only** the `"^""` lines: fails — isolates them as the cause.
- Pinned by `packages/omo-codex/src/install/codex-cache-bins.test.ts:145-146`, so the bug ships in every generated shim.

[...]

PR #7867, opened 2026-09-06
fix(codex-install): stop generated .cmd shims aborting when NODE_REPL_NODE_PATH is set

LilMGenius:
## What breaks

Every generated `.cmd` entry point dies before reaching its target on any Windows machine where `NODE_REPL_NODE_PATH` is present in `config.toml`, which is what Codex Desktop writes by default:

```
exit 255
The syntax of the command is incorrect.
```

Reported as #7473 on 2026-08-29. No PR has fixed it.

## Root cause

`windowsNodeDiscoveryLines()` strips surrounding quotes with two caret-escaped comparisons:

```bat
if "!OMO_NODE_BINARY:~0,1!"=="^"" set "OMO_NODE_BINARY=!OMO_NODE_BINARY:~1!"
```

`^"` is only a valid quote escape at the cmd top level. Both lines sit inside the `if defined OMO_NODE_REPL_NODE_PATH (` block, and inside a parenthesized block the parser rejects the construct and aborts the script before the exec line. The block is only entered when `NODE_REPL_NODE_PATH` resolves to something, which is why a machine without it never sees this.

## RED / GREEN

Generating the shim, writing a `config.toml` with `NODE_REPL_NODE_PATH`, and running it under `cmd.exe`, against upstream/dev at `ad62603f0`:

| Input | Before | After |
| --- | --- | --- |
| shim run with `NODE_REPL_NODE_PATH` set | exit 255, `The syntax of the command is incorrect.` | exit 0, target prints `shim-ran hello` |
| `bun test` over the shim, cache-bins and cleanup suites | 1 failure, on the caret assertion | 40 tests, 0 fail |

Evidence, including every log quoted here: `.omo/evidence/20260907-windows-cmd-shim-caret/`.

## The fix

The trim line above the comparisons already runs the value through `for /f`, and `%%~N` strips surrounding quotes as part of the expansion. Changing `%%N` to `%%~N` removes the need for both comparisons, so they are deleted rather than escaped differently. The single-quote pair below is untouched: `"'"` needs no caret and is valid inside a block.

Restoring the pre-fix block fails three assertions: the shim execution under `cmd.exe`, a guard that no caret-escaped comparison is generated, and the existing runtime-wrapper parity test.

## Worth checking in review

`codex-cache-bins.test.ts` line 145 required the generated wrapper to contain the broken caret line verbatim, so the defect was pinned as intended behavior and any repair would have failed CI. I replaced that assertion with its negation and put the real coverage in a new file that runs the shim rather than reading it. If you would rather keep a text assertion there, say so and I will narrow to that.

`windowsNodeDiscoveryLines()` is shared with the runtime wrapper in `codex-cache-runtime-wrapper.ts`, so this repairs both call sites at once.

The generated installer bundle is in the diff because `codex-install-bundle-freshness.test.ts` reads it from the git index and it embeds this source. Regenerating it also moves the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`; the checked-in bundle is two releases stale, and a rebuild on a clean `dev` produces that line with no source change at all.

---
## Summary by cubic
Fixes generated `.cmd` shims aborting on Windows when `NODE_REPL_NODE_PATH` is set in `config.toml`, which is what Codex Desktop writes by default (issue #7473). Previously the shim exited 255 with "The syntax of the command is incorrect" before reaching its target; the caret-escaped quote comparisons are invalid inside a parenthesized block, and `%%~N` in the existing `for /f` expansion already strips quotes, so those comparisons are deleted.

**Bug Fixes**
- `windowsNodeDiscoveryLines()` is shared with the runtime wrapper, so both call sites are repaired at once.
- The `codex-cache-bins.test.ts` assertion that pinned the broken caret line verbatim is now its negation; new coverage in `codex-cache-command-shim.test.ts` runs the generated shim through `cmd.exe` with `NODE_REPL_NODE_PATH` set.
- The installer bundle is regenerated because it embeds this source, which also moves the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`; a clean rebuild already produces that line with no source change.

<sup>Written for commit 2dfd2d5d1730493ea632a4e371543f7b26748cc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7867?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:

This closes #7473, open since 2026-08-29 with nothing filed against it.

Every generated `.cmd` entry point dies before reaching its target when `NODE_REPL_NODE_PATH` is in `config.toml`, which is what Codex Desktop writes by default:

```
exit 255
The syntax of the command is incorrect.
```

`^"` is only a valid quote escape at the cmd top level, and the two comparisons using it sit inside the `if defined OMO_NODE_REPL_NODE_PATH (` block, where the parser rejects the construct and aborts before the exec line.

| shim run under cmd.exe, `NODE_REPL_NODE_PATH` set | Before | After |
| --- | --- | --- |
| exit code | 255 | 0 |
| stderr | `The syntax of the command is incorrect.` | empty |
| target output | none | `shim-ran hello` |

The repair is a deletion, not a different escape: the trim line above already runs the value through `for /f`, and `%%~N` strips surrounding quotes as part of the expansion, so both comparisons go away. The single-quote pair stays, since it needs no caret.

One thing worth your eye. `codex-cache-bins.test.ts` line 145 required the generated wrapper to contain the broken caret line verbatim, so the defect was pinned as intended behavior and any repair would have failed CI. I replaced that assertion with its negation and put the real coverage in a new file that runs the shim through `cmd.exe` instead of reading it. If you would rather keep a text assertion there, say so and I will narrow to that.

`windowsNodeDiscoveryLines()` is shared with the runtime wrapper, so this repairs both call sites. The generated installer bundle is in the diff because the freshness test reads it from the git index; that regeneration also carries the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`, which is pre-existing staleness rather than part of the fix.

LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `packages/omo-codex/src/install/codex-cache-command-shim.ts` still emits both `^"` comparisons inside the `if defined OMO_NODE_REPL_NODE_PATH (` block, which is the abort reported in #7473 on 2026-08-29 and still open.

Issue #7144, opened 2026-08-23
[Bug]: Hook dispatcher spawns visible console window and echoes hook JSON payload on Windows


heath1002:
## Summary
On Windows, every Claude-code-compatible hook dispatch (UserPromptSubmit, PostToolUse, etc.) spawns a **visible `cmd.exe`/conhost window** that flashes the Windows banner, the cwd prompt, and the full hook JSON payload echoed on screen. When OpenCode is launched headlessly by an IDE extension (no attached console), this produces a black popup window for every prompt submission and tool use.

## Environment
- oh-my-openagent version: 4.15.1 (installed as `"plugin": ["oh-my-openagent@latest"]`)
- Host: OpenCode running as background server (`opencode --port 20982`), launched by IDE extension (Antigravity), so the plugin process has **no attached console**
- OS: Windows 11 Pro, 10.0.22631.6199
- Install method: npm package resolved by OpenCode plugin loader (`~/.config/opencode/node_modules/oh-my-openagent`)
- Relevant config: default Claude-code-hooks bridge enabled (UserPromptSubmit / PostToolUse command hooks)

## Repository Decision
- Target repository: `code-yeongyu/oh-my-openagent`
- Why this belongs here: the defect is entirely inside this repo's own hook executor, `packages/utils/src/command-executor/execute-hook-command.ts`. It is not upstream Codex CLI behavior (this is the OpenCode plugin surface). Verified against latest `main` at commit `0902c4a` (v5.0.0-beta.17 source-state merge): the missing option is still present there.

## Reproduction
1. On Windows, run OpenCode with the oh-my-openagent plugin where the host process has no visible console (e.g. spawned by an IDE extension or service wrapper)
2. Type any prompt into the chat (triggers `UserPromptSubmit`) or let any tool run (triggers `PostToolUse`)
3. Observe a flashing `cmd.exe` console window showing:
   - Windows banner (`Microsoft Windows [Version 10.0.22631.x]`)
   - The cwd prompt (`D:\...\project>`)
   - The raw hook payload echoed after it, e.g.
     `{"session_id":"ses_...","cwd":"...","permission_mode":"bypassPermissions","hook_event_name":"UserPromptSubmit","prompt":"?","hook_source":"opencode-plugin"}`

## Expected Behavior
Hook commands run invisibly; no console window is allocated and no hook JSON is displayed to the user.

## Actual Behavior
[...]

PR #7870, opened 2026-09-06
fix(hooks): redirect the hook home directory on Windows, not just HOME

LilMGenius:
## What breaks

Every hook that resolves the user's home directory sees the real profile on Windows, not the one `executeHookCommand` redirected it to. For the `allowedEnvVars` branch, whose purpose is to hand a plugin-sourced hook a scrubbed environment, the real profile path arrives there too.

## Root cause

`executeHookCommand` builds the child env with `HOME: home`. Nothing on Windows reads `HOME`: `os.homedir()` reads `USERPROFILE`, and Windows supplies `USERPROFILE` to the child from the parent whatever env object `spawn` is given, including an empty one.

| env passed to spawn | `USERPROFILE` seen by the child |
| --- | --- |
| `{}` | the real profile |
| `{ PATH }` | the real profile |
| `{ PATH, HOME: sandbox }` | the real profile |
| `{ PATH, HOME: sandbox, USERPROFILE: sandbox }` | the sandbox |

## RED / GREEN

A hook that prints `os.homedir()`, with `HOME` redirected to a sandbox and `USERPROFILE` left pointing elsewhere:

| Input | Before | After |
| --- | --- | --- |
| hook resolving the user directory | the `USERPROFILE` path | the redirected home |
| `bun test packages/utils/src/command-executor/execute-hook-command.test.ts` | 6 tests | 7 pass / 0 fail |

Reverting the source to `dev` fails the new test with the decoy `USERPROFILE` path in place of the sandbox. Evidence, including both probes: `.omo/evidence/20260907-windows-hook-home-redirect/`.

## The fix

`env.USERPROFILE = home` on win32, behind the `isWin32` flag already computed above, applied after both env branches so it covers the allowlist path as well. `USERPROFILE` joins `PROTECTED_ENV_KEYS` so the `allowedEnvVars` merge cannot restore the ambient one. Non-Windows behavior is unchanged.

## Worth checking in review

On a platform where `os.homedir()` reads `HOME`, the new assertion holds with or without the fix, so the mutation is only meaningful on the Windows leg. The test sets `USERPROFILE` to its own decoy directory rather than relying on the ambient value, so it does not depend on what a sibling test left behind.

The same file's visible-console-window behavior is #7144 with open PR #7258, and `cmd.exe` resolution is open PR #7205. Neither is touched here.

---
## Summary by cubic
Fixes the hook home directory redirect on Windows, where a hook resolving the user directory saw the real profile instead of the redirected home because `os.homedir()` reads `USERPROFILE`, not `HOME`. Windows supplies its own `USERPROFILE` to the child whatever env object is passed to `spawn`, so setting `HOME` alone never took.

- Sets `USERPROFILE` to the redirected home on win32, behind the existing `isWin32` check; non-Windows behavior is unchanged.
- Adds `USERPROFILE` to `PROTECTED_ENV_KEYS` so the `allowedEnvVars` env scrub cannot restore the ambient profile path into the hook.
- Adds a regression test that runs a hook printing `os.homedir()` and asserts it sees the redirected home.

<sup>Written for commit 4bcebdad17e77352a9b9e03c186e16521f9a88c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:

Every hook that resolves the user's home directory sees the real profile on Windows, not the one `executeHookCommand` redirected it to. The `allowedEnvVars` branch is the sharper case, since its whole purpose is to hand a plugin-sourced hook a scrubbed environment and the real profile path arrives there anyway.

`HOME: home` is set, but nothing on Windows reads `HOME`: `os.homedir()` reads `USERPROFILE`, and Windows supplies `USERPROFILE` to the child whatever env object `spawn` is given.

| env passed to spawn | `USERPROFILE` seen by the child |
| --- | --- |
| `{}` | the real profile |
| `{ PATH, HOME: sandbox }` | the real profile |
| `{ PATH, HOME: sandbox, USERPROFILE: sandbox }` | the sandbox |

| a hook printing `os.homedir()` | Before | After |
| --- | --- | --- |
| `HOME` redirected, `USERPROFILE` pointing elsewhere | the `USERPROFILE` path | the redirected home |

The fix is `env.USERPROFILE = home` on win32, behind the `isWin32` flag already computed above and applied after both env branches so the allowlist path is covered too. `USERPROFILE` joins `PROTECTED_ENV_KEYS` so the merge cannot restore the ambient one. Non-Windows behavior is unchanged.

Worth your eye: on a platform where `os.homedir()` reads `HOME` the new assertion holds with or without the fix, so the mutation only bites on the Windows leg. The test sets `USERPROFILE` to its own decoy directory rather than relying on the ambient value, because my first version set both and passed with the fix reverted.

The same file's visible-console-window behavior is #7144 with open PR #7258, and `cmd.exe` resolution is open PR #7205; neither is touched here.

LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `packages/utils/src/command-executor/execute-hook-command.ts` still passes `HOME: home` only, and nothing on Windows reads `HOME`, so every hook still resolves the real profile through `USERPROFILE`.

PR #7871, opened 2026-09-06
fix(codex-install): wait out a live runtime when pruning the plugin cache on Windows

LilMGenius:
## What breaks

Running an install or update while anything from the previous cache is still alive leaves the stale cache entry in place. `pruneMarketplaceCache` removes entries with `rm(dir, { recursive: true, force: true })`, and Windows refuses to unlink a mapped image, so a leftover MCP server or daemon aborts the removal outright. `install-codex.ts:157` and `:163` call it on every install and update, which is exactly when that leftover is most likely to exist.

Driving the real function against a cache entry holding a running executable that exits 300 ms later, on `upstream/dev`:

```
bun 1.3.14   | pruneMarketplaceCache -> threw EACCES | removed=false | 2ms
node v26.1.0 | pruneMarketplaceCache -> threw EPERM  | removed=false | 4ms
```

## Why not `maxRetries`

Node's `rm` takes `maxRetries` and `retryDelay`, and this repo already uses them for cache removal in `packages/omo-codex/plugin/scripts/sync-skills.mjs:250`. That is not enough here, because the runtimes disagree:

| runtime | no retry | `maxRetries: 5, retryDelay: 100` |
| --- | --- | --- |
| node v26.1.0 | EPERM | ok in 644 ms |
| bun 1.3.14 | EACCES | EACCES, 1 ms |

bun ignores the option and reports `EACCES`, which is not in Node's retriable set either. An explicit loop covering `EPERM`, `EBUSY`, `EACCES` and `ENOTEMPTY` is what works under both.

## RED / GREEN

| Input | Before | After |
| --- | --- | --- |
| prune with a live runtime that exits after 300 ms, node | EPERM, entry survives | resolved, entry removed, 393 ms |
| the same under bun | EACCES, entry survives | resolved, entry removed, 405 ms |
| `bun test packages/omo-codex/src/install/codex-cache-prune.test.ts` | file did not exist | 1 pass / 0 fail |

Reverting the source fails the new test with `EACCES: permission denied, rm ...omo-old`. Evidence, including both runtime logs: `.omo/evidence/20260907-windows-cache-prune-ebusy/`.

## The fix

`removeWithRetry` mirrors `renameWithRetry` in `codex-config-atomic-write.ts`, one directory over: a delay ladder plus a set of retriable codes. The ladder is `[50, 100, 200, 400, 800]` rather than that file's `[10, 25, 50]`, because 85 ms is not enough for a process to finish exiting. Behavior on other platforms is unchanged, since none of those codes occur there for this call.

## Worth checking in review

On a platform that allows unlinking a running executable the new assertion holds with or without the fix, so the mutation only bites on Windows.

The retry is bounded at about 1.5 s. A runtime that never exits still fails, which is the correct outcome: the install should not hang waiting for a process the user has to close.

`codex-cache-install.ts` moves the live plugin directory aside with a bare `rename` during install, which is the same Windows rule and a different root cause. It is not in this PR.

The generated installer bundle is in the diff because `codex-install-bundle-freshness.test.ts` reads it from the git index. That regeneration also carries the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`, which a rebuild on a clean `dev` produces with no source change at all.

---
## Summary by cubic
Fixes Windows installs/updates leaving stale cache entries when a runtime from the previous cache is still alive. Windows refuses to unlink a running executable, so the prune now retries with a delay ladder until the process exits instead of failing immediately with `EPERM`/`EACCES`.

- Node's `maxRetries` option is insufficient: bun ignores it and reports `EACCES`, which isn't in Node's retriable set.
- Retries on `EPERM`, `EBUSY`, `EACCES`, `ENOTEMPTY` with delays `[50, 100, 200, 400, 800]` ms.
- Adds a test that prunes a cache entry whose runtime exits 300 ms into the prune.
- Regenerates the installer bundle; the embedded version moves `5.0.0-beta.43` → `5.0.0-beta.45`, which a clean rebuild also produces.
- Behavior on other platforms is unchanged; the retry is bounded at ~1.5 s and a process that never exits still fails rather than hanging the install.

<sup>Written for commit 56f820cbbdf7d4c1ca8cfbadf316d52c0e553441. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7871?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LilMGenius:

Running an install or update while anything from the previous cache is still alive leaves the stale entry in place. `pruneMarketplaceCache` removes entries with `rm(dir, { recursive: true, force: true })`, Windows refuses to unlink a mapped image, and `install-codex.ts:157` and `:163` call it on exactly the occasion when a leftover MCP server or daemon is most likely to still be running.

Driving the real function against an entry holding a running executable that exits 300 ms later:

| runtime | Before | After |
| --- | --- | --- |
| node v26.1.0 | EPERM in 4 ms, entry survives | resolved in 393 ms, entry removed |
| bun 1.3.14 | EACCES in 2 ms, entry survives | resolved in 405 ms, entry removed |

The obvious fix does not work, which is the part worth your eye. Node's `rm` takes `maxRetries`, and `plugin/scripts/sync-skills.mjs:250` already uses it for cache removal, but bun ignores the option and reports `EACCES`, which is not in Node's retriable set either. So the retry is explicit and covers `EPERM`, `EBUSY`, `EACCES` and `ENOTEMPTY`. `removeWithRetry` mirrors `renameWithRetry` in `codex-config-atomic-write.ts` one directory over; the ladder is longer because that file's 85 ms total is not enough for a process to finish exiting.

Two disclosures. The retry is bounded at about 1.5 s, so a runtime that never exits still fails, which I think is right: the install should not hang on a process the user has to close. And on a platform that allows unlinking a running executable the new test holds with or without the fix, so the mutation only bites on Windows.

`codex-cache-install.ts` moves the live plugin directory aside with a bare `rename` during the same install, which is the same Windows rule and a different root cause. Not in this PR.

The generated installer bundle is in the diff because the freshness test reads it from the git index; that regeneration also carries the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`, which a rebuild on a clean `dev` produces with no source change.

LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `pruneMarketplaceCache` in `packages/omo-codex/src/install/codex-cache-prune.ts` still removes with a bare recursive rm and no retry, so a leftover runtime still aborts the prune on Windows.

PR #7872, opened 2026-09-06
fix(codex-install): retry a cached-plugin promote whose directory is still busy

LilMGenius:
Installing or updating the Codex plugin fails on Windows whenever anything from the previous install is still exiting.

```
EPERM: operation not permitted, rename
  '<CODEX_HOME>/plugins/cache/<marketplace>/omo/<version>'
  -> '<CODEX_HOME>/plugins/cache/<marketplace>/omo/.backup-<version>-...'
```

`promoteDirectory` in `packages/omo-codex/src/install/codex-cache-install.ts` renames with a bare `rename`. Windows refuses a directory rename while any process holds a file inside it open or has it as its cwd, and an npm helper that has not finished exiting keeps its cwd in the directory the installer just built. The promote gives up in single-digit milliseconds over a lock that clears a few hundred milliseconds later, and the user is left on the old plugin version.

This retries the rename over 50/100/200/400/800ms on EPERM and EBUSY, the same shape `renameWithRetry` in `codex-config-atomic-write.ts` already uses for a busy file rename. The delays are longer here because the holder is an exiting child process rather than a file handle being closed on the same thread.

### Measured

A driver that starts a helper with its cwd inside the live cache directory, under **both runtimes the installer runs on**:

| | node v26.1.0 | bun 1.3.14 |
|---|---|---|
| helper exits at 300ms, as shipped | threw EPERM (4ms), plugin dir old | threw EPERM (11ms), plugin dir old |
| helper exits at 300ms, with retry | ok (377ms), plugin dir **new** | ok (377ms), plugin dir **new** |
| helper never exits, with retry | throws EPERM (1594ms) | throws EPERM (1616ms) |

The last row is the intended ceiling: a genuinely stuck holder still fails, it just fails after the window instead of inside it. The backup restore path holds in every case, so a failed promote always leaves the previous plugin dir intact.

### Test

`packages/omo-codex/src/install/codex-cache-promote.test.ts` drives `installCachedPlugin` against that holder. Reverting the default back to a bare `rename` fails it with the EPERM above; with the retry it passes in 848ms.

`bun test src/install` in `packages/omo-codex`: 267 pass / 12 fail on my Windows machine, against 266 pass / 12 fail on `dev` at `ad62603f0` with the same `node_modules`. Same twelve pre-existing local failures, one new pass. `bun run typecheck` exits 0.

### One line that is not mine

The bundle commit also moves the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`. The checked-in `install-local.mjs` was stale against the current `package.json`, and `bun run build:codex-install` picks that up on any regeneration. Happy to split it out if you would rather it landed on its own.

---
## Summary by cubic
Fixes Codex plugin installs on Windows failing with EPERM when a process from the previous install is still exiting. The cache directory promote now retries the rename on EPERM and EBUSY, so a busy hold clears and the new plugin version lands instead of the old staying in place.

- Retries with 50/100/200/400/800ms delays; a holder that never exits still fails after about 1.6s.
- The backup restore path still holds, so a failed promote leaves the previous plugin dir intact.
- Adds a test that drives `installCachedPlugin` against a busy directory and exits the holder during promotion.
- The checked-in `install-local.mjs` bundle is regenerated, which also moves the embedded package version from `5.0.0-beta.43` to `5.0.0-beta.45`.

<sup>Written for commit 0da7baaa4dc3bd0470a97a2d745d87e6cfba8669. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7872?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-codex/scripts/install-dist/install-local.mjs
+++ b/packages/omo-codex/scripts/install-dist/install-local.mjs
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-// omo-codex-install:81cf03496fa0db52b428d9e5f1d2a022f753fe60530d84f72d767af03333373f:cc1555990cc7467cc545d7f3fcb60ad36f9dc529737a841ed819c2d8ac9303be
+// omo-codex-install:e3b01c9c644553d3d24772808854b691edfe72f6e4a5520c62fb39a167063866:9a6a7b7fce7ae3051ac44306753da57f686bcebab247b8268978191e3beea728
 var __defProp = Object.defineProperty;
 var __returnValue = (v) => v;
 function __exportSetter(name, newValue) {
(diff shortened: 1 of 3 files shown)


LilMGenius:

Installing or updating the Codex plugin fails on Windows whenever anything from the previous install is still exiting:

```
EPERM: operation not permitted, rename
  '<CODEX_HOME>/plugins/cache/<marketplace>/omo/<version>'
  -> '.../omo/.backup-<version>-...'
```

`promoteDirectory` renames with a bare `rename`, and Windows refuses a directory rename while a process holds a file inside it open or has it as its cwd. An npm helper that has not finished exiting still has its cwd in the directory the installer just built, so the promote gives up in 4ms over a lock that clears 300ms later and the user stays on the old version.

Verified against a driver that starts a holder with its cwd inside the live cache directory, under both runtimes the installer runs on:

| | node v26.1.0 | bun 1.3.14 |
|---|---|---|
| holder exits at 300ms, as shipped | EPERM after 4ms, plugin dir old | EPERM after 11ms, plugin dir old |
| holder exits at 300ms, with the retry | ok after 377ms, plugin dir new | ok after 377ms, plugin dir new |
| holder never exits, with the retry | EPERM after 1594ms | EPERM after 1616ms |

The third row is the ceiling working as intended, and a failed promote always leaves the previous plugin dir in place. `codex-cache-promote.test.ts` fails with the EPERM above when the default is reverted to a bare `rename`.

One disclosure: the bundle commit also moves the embedded version from `5.0.0-beta.43` to `5.0.0-beta.45`, because the checked-in `install-local.mjs` was already stale against `package.json` and any regeneration picks that up. Happy to split it out.

This is the same family of Windows lock failure as #7871 but a different call and a different trigger, and there is no ordering between them.

LilMGenius:
Correction to my own note above. I called the `5.0.0-beta.43` to `5.0.0-beta.45` line "stale drift, not part of this fix" and that was wrong, so please ignore the offer to split it out.

`script/codex-installer-version.test.ts` asserts that the embedded version in `install-local.mjs` matches the root release version, and it is red on `dev` at `ad62603f0` right now:

```
$ bun test script/codex-installer-version.test.ts
Expected to contain: "version: \"5.0.0-beta.45\""
Received: ... version: "5.0.0-beta.43" ...
 0 pass, 1 fail
```

On this branch the same command is 1 pass / 0 fail. So the version line is the repair for a gate that is currently failing, not incidental churn, and pulling it out would leave that gate red.

Nothing else in the PR changes. Sorry for the noise on my own thread.

LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `promoteDirectory` in `packages/omo-codex/src/install/codex-cache-install.ts` still renames with a bare rename and no retry, so a Windows install still fails with EPERM while the previous install is exiting.

PR #7873, opened 2026-09-07
fix(schema): regenerate omo.schema.json after memory.tool_exposure was removed

LilMGenius:
`assets/omo.schema.json` still advertises `memory.tool_exposure`, and a config that uses it no longer parses:

```
OmoConfigSchema.safeParse({ memory: { enabled: true, agent: "auto", tool_exposure: "direct" } })
  -> success: false
  -> memory Unrecognized key: "tool_exposure"
```

That file is what `docs/reference/omo-json.md` tells users to point `$schema` at, so an editor completes the key with its `direct`/`search` enum and the completion is a config error.

`refactor: consolidate memory tool surface` (6fd53f2cf) removed the key from the Zod schema and the generated artifact was not regenerated with it. Nothing under `packages/`, `script/` or `docs/` mentions `tool_exposure` any more; only the published schema does.

### The gate for this already exists and is red

`tests/omo-schema-freshness.test.ts` compares the regenerated schema against the committed one. On `dev` at `ad62603f0`, with nothing applied:

```
$ bun test tests/omo-schema-freshness.test.ts
 2 pass
 1 fail
```

The reported diff is exactly the 44 `tool_exposure` lines. After `bun run build:omo-schema`, the same command is 3 pass / 0 fail.

### The change

Output of `bun run build:omo-schema`, deletions only, one file. Running the sibling generator `bun run build:schema` afterwards regenerates both artifacts and leaves the same single modified file, so the output is stable and `assets/oh-my-opencode.schema.json` was already fresh.

No source, test or workflow change. The existing gate goes from red to green on its own.

---
## Summary by cubic
Regenerates `assets/omo.schema.json` so it no longer advertises the removed `memory.tool_exposure` key. The stale schema made editor completions of that key produce configs that fail parsing; now the artifact matches the Zod schema.

- The change is deletions only, and the existing schema-freshness test goes from red to green.

<sup>Written for commit d33d45daff676a6d19429353735220d888d1ac4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7873?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/assets/omo.schema.json
+++ b/assets/omo.schema.json
@@ -1314,14 +1314,6 @@
           "type": "string",
           "minLength": 1
         },
-        "tool_exposure": {
-          "default": "direct",
-          "type": "string",
-          "enum": [
-            "direct",
-            "search"
-          ]
-        },
         "reflection": {
           "default": {
             "enabled": true,
@@ -1844,7 +1836,6 @@
       "required": [
         "enabled",
         "agent",
-        "tool_exposure",
         "reflection",
         "nudge",
         "facts",
@@ -12998,13 +12989,6 @@
               "type": "string",
               "minLength": 1
             },
-            "tool_exposure": {
-              "type": "string",
-              "enum": [
-                "direct",
-                "search"
-              ]
-            },
             "reflection": {
               "type": "object",
               "properties": {
@@ -14624,13 +14608,6 @@
             "agent": {
               "type": "string",
               "minLength": 1
-            },
-            "tool_exposure": {
-              "type": "string",
-              "enum": [
-                "direct",
-                "search"
-              ]
             },
             "reflection": {
               "type": "object",
@@ -16257,13 +16234,6 @@
               "agent": {
                 "type": "string",
                 "minLength": 1
-              },
-              "tool_exposure": {
-                "type": "string",
-                "enum": [
-                  "direct",
-                  "search"
-                ]
               },
               "reflection": {
                 "type": "object",
@@ -27791,13 +27761,6 @@
                     "type": "string",
                     "minLength": 1
                   },
-                  "tool_exposure": {
-                    "type": "string",
-                    "enum": [
-                      "direct",
-                      "search"
-                    ]
-                  },
                   "reflection": {
                     "type": "object",
                     "properties": {
@@ -29417,13 +29380,6 @@
                   "agent": {
                     "type": "string",
                     "minLength": 1
-                  },
-                  "tool_exposure": {
-                    "type": "string",
-                    "enum": [
-                      "direct",
-                      "search"
-                    ]
                   },
                   "reflection": {
                     "type": "object",


LilMGenius:

This one is a regeneration rather than a fix, and the gate that catches it is already yours and already red. On `dev` at `ad62603f0`, with nothing applied:

```
$ bun test tests/omo-schema-freshness.test.ts
 2 pass
 1 fail
```

The reported diff is exactly the 44 `memory.tool_exposure` lines. `refactor: consolidate memory tool surface` (6fd53f2cf) removed the key from the Zod schema and `assets/omo.schema.json` was not regenerated with it.

What that costs a user: `assets/omo.schema.json` is the file `docs/reference/omo-json.md` points `$schema` at, so an editor still completes `memory.tool_exposure` with its `direct`/`search` enum, and taking the completion fails the parse.

```
OmoConfigSchema.safeParse({ memory: { enabled: true, agent: "auto", tool_exposure: "direct" } })
  -> success: false
  -> memory Unrecognized key: "tool_exposure"
```

The diff is the output of `bun run build:omo-schema`, one file, deletions only, and the freshness test is 3/3 after it. Running `bun run build:schema` afterwards leaves the same single modified file, so the output is stable and `assets/oh-my-opencode.schema.json` was already fresh.

No source, test or workflow change, and no ordering against anything else I have open.

code-yeongyu (approved):
Approving. Verified rather than trusted:

- **Regeneration, locally (build surface):** on `dev` `ad62603f0`, `bun run build:omo-schema && bun run build:schema` followed by `git diff --exit-code -- assets/oh-my-opencode.schema.json assets/omo.schema.json` exits 1 with exactly the 44 `tool_exposure` deletions in `omo.schema.json`; on this branch (`d33d45daf`) the same sequence exits 0 and `git status --porcelain assets` is empty, so `oh-my-opencode.schema.json` is indeed already fresh and the output is stable.
- **The gate, on an isolated Linux box (bun 1.4.0, `bun install --ignore-scripts` so nothing regenerated the artifact first):** `bun test tests/omo-schema-freshness.test.ts` is 2 pass / 1 fail on the dev tree (`- Expected - 44 / + Received + 0`) and 3 pass / 0 fail on this branch.

**Why the red gate never blocked anything — a finding beyond this PR, not a request on it.** `script/build.ts` runs `build:schema` and `build:omo-schema`, and every CI job runs `bun install` (root `prepare` = `bun run build`) or `bun run build` before its tests. So by the time `tests/omo-schema-freshness.test.ts` runs in CI, the committed artifact has already been overwritten with the regenerated one and the comparison is a tautology; the test is only meaningful on a clean checkout, which is how you caught it. `auto-commit-schema` cannot backstop it either: it is gated on `refs/heads/master`, whose last commit is from 2026-07-13, while the default branch is `dev`. I am landing a follow-up that adds a `git diff --exit-code` step on the two artifacts to the required `build` job right after `bun run build`, so this class of drift fails a PR instead of waiting for someone to run the test locally.

Merging now.

code-yeongyu:
Reviewed and approved — regeneration verified locally on both trees and the freshness test run on a clean Linux box (dev 1 fail -> this branch 3 pass). You also surfaced something bigger: the gate is a tautology in CI because the root build regenerates the artifacts before the test runs, and the auto-commit job is pinned to a master branch nothing pushes to anymore. Details in the review; I am following up with a real CI drift gate on the required build job. Merging.

Issue #6738, opened 2026-08-11
[senpi-task][Windows] Runtime crashes with uncaughtException "spawn taskkill ENOENT" when PATH lacks System32


grim-susemi:
## Symptom

The pi/omo runtime exits during session cleanup with:

```
pi exiting due to uncaughtException:
Error: spawn taskkill ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:287:19)
    at onErrorNT (node:internal/child_process:525:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn taskkill',
  path: 'taskkill',
  spawnargs: [ '/F', '/T', '/PID', '40368' ]
}
```

Observed on Windows when the agent is launched from a parent that sanitizes the child environment (a terminal app), so `PATH` does not include `C:\Windows\System32`. Triggered from bash-tool cleanup / session teardown (process-tree kill). Crashed 2 sessions on 2026-08-11 during routine cleanup.

## Root cause

1. `killProcessTree()` on Windows — in `senpi/dist/utils/shell.js` and the identical function in `@earendil-works/pi-agent-core/dist/harness/env/nodejs.js` — spawns the bare name `"taskkill"`.
2. `child_process.spawn` does **not** throw synchronously when the executable cannot be resolved; the failure is emitted asynchronously via the child's `'error'` event.
3. Neither call site attaches an `'error'` listener, so `ENOENT` surfaces as an `uncaughtException` and the whole runtime process terminates. The surrounding `try/catch` cannot help because the error is asynchronous.

## Suggested fix

In both `killProcessTree` implementations:

```js
const taskkill = `${process.env.SystemRoot ?? "C:\\Windows"}\\System32\\taskkill.exe`;
spawn(taskkill, ["/F", "/T", "/PID", String(pid)], {
  stdio: "ignore",
  detached: true,
  windowsHide: true,
}).on("error", () => { /* never crash the runtime on spawn failure */ });
```

Also audit other bare-name system-tool spawns (e.g. `powershell.exe` for process listing) for missing error listeners. Note: `spawnSync` sites (e.g. lsp-daemon) are safe — `spawnSync` returns the error in the result object instead of emitting it.

## Local workaround (2026-08-11)

[...]

PR #7885, opened 2026-09-07
fix(openclaw): identify the reply-listener daemon on Windows

LilMGenius:
On Windows the reply-listener daemon can never be detected or stopped, and every attempt orphans it.

`isReplyListenerDaemonProcessWithDeps` reads `/proc/<pid>/cmdline` on linux and falls back to `ps -p <pid> -o args=` everywhere else. Windows has neither, so the spawn fails, `ignoreReplyListenerProcessProbeError` swallows it, and the function returns `false` for every pid including a genuine daemon.

That `false` decides three things:

- `isDaemonRunning` reports a live daemon as not running **and** calls `removeReplyListenerPid()`, deleting the only record of its pid.
- `stopReplyListener` returns `success: false` with *"Refusing to kill PID N: process identity does not match the reply listener daemon (stale or reused PID - removed PID file)"* and removes the PID file.
- `terminateReplyListenerProcess` returns without sending SIGTERM.

So a Windows user who starts the daemon cannot stop it through the product, and the first status check strands it.

### Measured against real processes

Driving the exported `isReplyListenerDaemonProcess` against processes spawned with and without the marker in argv:

| | dev at `b2f688f1c` | this branch |
|---|---|---|
| marker in argv, process alive | `false` | `true` (1021ms) |
| no marker, process alive | `false` | `false` |
| same pid after it exited | `false` | `false` |

The middle and bottom rows are the controls: the fix must not make it say yes to everything.

### The change

Read the command line from `Win32_Process`, which is the supported way to get it on Windows:

```
Get-CimInstance Win32_Process -Filter "ProcessId=<pid>" | Select-Object -ExpandProperty CommandLine
```

PowerShell is addressed through `SystemRoot` rather than PATH, because PATH can be missing System32, which is the failure in #6747. `pid` is typed `number`, so it cannot escape the filter string. A dead pid returns exit 0 with empty output, so the marker check is already the right answer there.

The branch goes through the existing `deps.spawn` seam, so the new test needs no real process. It models Windows by throwing `ENOENT` for `ps`, and asserts the outcome rather than the command shape.

`bun test packages/openclaw-core` is 69 pass / 0 fail; removing the win32 branch returns the file to 2 pass / 1 fail. `bun run typecheck` in `packages/openclaw-core` exits 0.

Found by scanning the published `oh-my-opencode@5.0.0-beta.45` bundle for calls that cannot work on Windows, then checking each for reachability. No ordering against anything else I have open.

---
## Summary by cubic
Fixes reply-listener daemon detection on Windows so the daemon can be stopped and status checks no longer orphan it.

- Reads the command line from `Win32_Process` via PowerShell instead of relying on `ps` or `/proc`.
- Locates PowerShell through `SystemRoot` to avoid missing PATH entries.
- Adds fake-spawn and real-process tests, and forces the full OS matrix when the probe file changes.

<sup>Written for commit 6ef4963db5b80d0b74644612260aa8070d2ef95e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/openclaw-core/src/__tests__/reply-listener-process.real-process.test.ts
+++ b/packages/openclaw-core/src/__tests__/reply-listener-process.real-process.test.ts
@@ -1,0 +1,65 @@
+import { afterAll, describe, expect, test } from "bun:test"
+import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
+import { isReplyListenerDaemonProcess, REPLY_LISTENER_DAEMON_IDENTITY_MARKER } from "../reply-listener-process"
+
+// The fake-spawn tests prove which branch runs; only a real child proves the platform
+// command line lookup works (#7885: the win32 branch shipped with zero Windows execution).
+// The child reports its own pid after exec: probing the spawn-returned pid immediately
+// can race the exec and read the pre-exec command line (observed on ubuntu CI).
+const scriptDir = mkdtempSync(join(tmpdir(), "openclaw-reply-listener-identity-"))
+const idleScript = join(scriptDir, "idle-daemon.ts")
+writeFileSync(
+  idleScript,
+  `const fs = require("node:fs")\nfs.writeFileSync(process.argv[2] + ".ready", String(process.pid))\nsetTimeout(() => {}, 60_000)\n`,
+)
+
+function spawnIdleChild(name: string, extraArgs: readonly string[]) {
+  return {
+    readyPath: join(scriptDir, `${name}.ready`),
+    child: Bun.spawn([process.execPath, "run", idleScript, join(scriptDir, name), ...extraArgs], {
+      stdin: "ignore",
+      stdout: "ignore",
+      stderr: "ignore",
+    }),
+  }
+}
+
+async function awaitReady(readyPath: string): Promise<number> {
+  const deadline = Date.now() + 20_000
+  while (Date.now() < deadline) {
+    if (existsSync(readyPath)) {
+      const pid = Number(readFileSync(readyPath, "utf-8").trim())
+      if (Number.isSafeInteger(pid) && pid > 0) return pid
+    }
+    await new Promise((resolve) => setTimeout(resolve, 25))
+  }
+  throw new Error(`waited 20s for the spawned child's ready marker, never appeared: ${readyPath}`)
+}
+
+afterAll(() => {
+  rmSync(scriptDir, { recursive: true, force: true })
+})
+
+describe("isReplyListenerDaemonProcess against real processes", () => {
+  test(
+    "#given live children with and without the daemon marker #when probing their self-reported pids #then only the marked child is the daemon and a dead pid is not",
+    async () => {
+      const marked = spawnIdleChild("marked", [REPLY_LISTENER_DAEMON_IDENTITY_MARKER])
+      const unmarked = spawnIdleChild("unmarked", [])
+      const markedPid = await awaitReady(marked.readyPath)
+      const unmarkedPid = await awaitReady(unmarked.readyPath)
+      try {
+        expect(await isReplyListenerDaemonProcess(markedPid)).toBe(true)
+        expect(await isReplyListenerDaemonProcess(unmarkedPid)).toBe(false)
+      } finally {
+        marked.child.kill()
+        unmarked.child.kill()
+        await Promise.all([marked.child.exited, unmarked.child.exited])
+      }
+      expect(await isReplyListenerDaemonProcess(markedPid)).toBe(false)
+    },
+    30_000,
+  )
+})
--- a/packages/openclaw-core/src/__tests__/reply-listener-process.test.ts
+++ b/packages/openclaw-core/src/__tests__/reply-listener-process.test.ts
@@ -69,4 +69,28 @@ describe("isReplyListenerDaemonProcess", () => {
     expect(await probe).toBe(true)
     expect(procReadCalls).toBe(0)
   })
+
+  test("#given windows where ps does not exist #when probing a daemon pid #then the command line still identifies it", async () => {
+    const spawnedCommands: string[][] = []
+
+    const result = await isReplyListenerDaemonProcessWithDeps(1234, {
+      platform: "win32",
+      spawn: (command) => {
+        spawnedCommands.push(command)
+        if (command[0] === "ps") {
+          throw Object.assign(new Error("spawn ps ENOENT"), { code: "ENOENT" })
+        }
+        return {
+          exitCode: 0,
+          exited: Promise.resolve(0),
+          stdout: createOutputStream(
+            '"C:\\\\Program Files\\\\nodejs\\\\node.exe" daemon.js --openclaw-reply-listener-daemon',
+          ),
+        }
+      },
+    })
+
+    expect(result).toBe(true)
+    expect(spawnedCommands.some((command) => command[0] === "ps")).toBe(false)
+  })
 })
--- a/packages/openclaw-core/src/reply-listener-process.ts
+++ b/packages/openclaw-core/src/reply-listener-process.ts
@@ -1,4 +1,5 @@
 import { readFileSync } from "node:fs"
+import { join } from "node:path"
 import { spawn } from "@oh-my-opencode/utils/runtime"
 
 export const REPLY_LISTENER_DAEMON_IDENTITY_MARKER = "--openclaw-reply-listener-daemon"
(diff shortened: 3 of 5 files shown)


LilMGenius:

On Windows the reply-listener daemon can never be identified, which means it can never be stopped, and every status check strands it.

`isReplyListenerDaemonProcessWithDeps` reads `/proc/<pid>/cmdline` on linux and falls back to `ps -p <pid> -o args=` everywhere else. Windows has neither, so the spawn fails, the probe swallows the error, and the function returns `false` for every pid including a real daemon. Three callers act on that:

- `isDaemonRunning` reports a live daemon as not running **and** calls `removeReplyListenerPid()`.
- `stopReplyListener` returns `success: false` with "Refusing to kill PID N: process identity does not match the reply listener daemon (stale or reused PID - removed PID file)", and removes the PID file.
- `terminateReplyListenerProcess` returns without sending SIGTERM.

Driving the exported function against real processes rather than a fake:

| | dev at `b2f688f1c` | this branch |
|---|---|---|
| marker in argv, alive | `false` | `true` (1021ms) |
| no marker, alive | `false` | `false` |
| same pid after it exited | `false` | `false` |

The bottom two rows are the controls, so the fix is not just answering yes.

It reads the command line from `Win32_Process`, which is the supported way on Windows. PowerShell is addressed through `SystemRoot` rather than PATH, because PATH can be missing System32, which is the failure in your open #6747. `pid` is typed `number` so it cannot escape the filter string, and a dead pid already returns exit 0 with empty output, so no extra liveness handling was needed.

The branch goes through the existing `deps.spawn` seam, so the test spawns nothing and asserts the outcome rather than the command shape. `bun test packages/openclaw-core` is 69 pass / 0 fail, removing the win32 branch returns the file to 2 pass / 1 fail, and `bun run typecheck` exits 0.

How I found it, since it may be worth reusing: I packed the published `oh-my-opencode@5.0.0-beta.45` and grepped the shipped `dist/index.js` for calls that cannot work on Windows, then checked each for reachability. Most were already answered. Three of the four `which` calls are correctly gated to `where`, the one `.join(":")` is percent-decoding inside a URI library rather than a PATH join, and bare `taskkill` and bare `powershell` are #6738 and #6747, which already have PRs from other contributors. This `ps` call was the one with nothing on it.

code-yeongyu:
> @code-yeongyu ready when you have a slot. All checks are green and it is mergeable.
> 
> On Windows the reply-listener daemon can never be identified, which means it can never be stopped, and every status check strands it.
> 
> `isReplyListenerDaemonProcessWithDeps` reads `/proc/<pid>/cmdline` on linux and falls back to `ps -p <pid> -o args=` everywhere else. Windows has neither, so the spawn fails, the probe swallows the error, and the function returns `false` for every pid including a real daemon. Three callers act on that:
> 
> - `isDaemonRunning` reports a live daemon as not running **and** calls `removeReplyListenerPid()`.
> - `stopReplyListener` returns `success: false` with "Refusing to kill PID N: process identity does not match the reply listener daemon (stale or reused PID - removed PID file)", and removes the PID file.
> - `terminateReplyListenerProcess` returns without sending SIGTERM.
> 
> Driving the exported function against real processes rather than a fake:
> 
> | | dev at `b2f688f1c` | this branch |
> |---|---|---|
> | marker in argv, alive | `false` | `true` (1021ms) |
> | no marker, alive | `false` | `false` |
> | same pid after it exited | `false` | `false` |
> 
> The bottom two rows are the controls, so the fix is not just answering yes.
> 
> It reads the command line from `Win32_Process`, which is the supported way on Windows. PowerShell is addressed through `SystemRoot` rather than PATH, because PATH can be missing System32, which is the failure in your open #6747. `pid` is typed `number` so it cannot escape the filter string, and a dead pid already returns exit 0 with empty output, so no extra liveness handling was needed.
> 
> The branch goes through the existing `deps.spawn` seam, so the test spawns nothing and asserts the outcome rather than the command shape. `bun test packages/openclaw-core` is 69 pass / 0 fail, removing the win32 branch returns the file to 2 pass / 1 fail, and `bun run typecheck` exits 0.
> 
> How I found it, since it may be worth reusing: I packed the published `oh-my-opencode@5.0.0-beta.45` and grepped the shipped `dist/index.js` for calls that cannot work on Windows, then checked each for reachability. Most were already answered. Three of the four `which` calls are correctly gated to `where`, the one `.join(":")` is percent-decoding inside a URI library rather than a PATH join, and bare `taskkill` and bare `powershell` are #6738 and #6747, which already have PRs from other contributors. This `ps` call was the one with nothing on it.
> 
> No ordering against anything else I have open.

Can you stop fucking tagging me?

LilMGenius:
> > @code-yeongyu ready when you have a slot. All checks are green and it is mergeable.
> > On Windows the reply-listener daemon can never be identified, which means it can never be stopped, and every status check strands it.
> > `isReplyListenerDaemonProcessWithDeps` reads `/proc/<pid>/cmdline` on linux and falls back to `ps -p <pid> -o args=` everywhere else. Windows has neither, so the spawn fails, the probe swallows the error, and the function returns `false` for every pid including a real daemon. Three callers act on that:
> > 
> > * `isDaemonRunning` reports a live daemon as not running **and** calls `removeReplyListenerPid()`.
> > * `stopReplyListener` returns `success: false` with "Refusing to kill PID N: process identity does not match the reply listener daemon (stale or reused PID - removed PID file)", and removes the PID file.
> > * `terminateReplyListenerProcess` returns without sending SIGTERM.
> > 
> > Driving the exported function against real processes rather than a fake:
> > 
> > 
> > 
> > 
> > dev at `b2f688f1c`
> > this branch
> > 
> > 
> > 
> > 
> > marker in argv, alive
> > `false`
> > `true` (1021ms)
> > 
> > 
> > no marker, alive
> > `false`
> > `false`
> > 
> > 
> > same pid after it exited
> > `false`
> > `false`
> > 
> > 
> > 
> > The bottom two rows are the controls, so the fix is not just answering yes.
> > It reads the command line from `Win32_Process`, which is the supported way on Windows. PowerShell is addressed through `SystemRoot` rather than PATH, because PATH can be missing System32, which is the failure in your open #6747. `pid` is typed `number` so it cannot escape the filter string, and a dead pid already returns exit 0 with empty output, so no extra liveness handling was needed.
> > The branch goes through the existing `deps.spawn` seam, so the test spawns nothing and asserts the outcome rather than the command shape. `bun test packages/openclaw-core` is 69 pass / 0 fail, removing the win32 branch returns the file to 2 pass / 1 fail, and `bun run typecheck` exits 0.
> > How I found it, since it may be worth reusing: I packed the published `oh-my-opencode@5.0.0-beta.45` and grepped the shipped `dist/index.js` for calls that cannot work on Windows, then checked each for reachability. Most were already answered. Three of the four `which` calls are correctly gated to `where`, the one `.join(":")` is percent-decoding inside a URI library rather than a PATH join, and bare `taskkill` and bare `powershell` are #6738 and #6747, which already have PRs from other contributors. This `ps` call was the one with nothing on it.
> > No ordering against anything else I have open.
> 
> Can you stop fucking tagging me?

Sry... 😔 and I will not tag you again.

I spent today on Windows bugs that had been open a long time, and I got too absorbed in them. I saw a tag on almost every merged pull request here and thought that was the convention, so I did the same on all of mine. That was my mistake, and it buried you in notifications.

There was no malice in it. I have stopped, and I closed the ones of mine that were noise.

code-yeongyu:
Verified the diagnosis independently and took the fix through real Windows CI. Two findings beyond the diff, then the evidence.

**The fix is correct.** Reading the command line from `Win32_Process` through the `SystemRoot`-addressed PowerShell is the supported path, `pid: number` keeps the filter string safe, a dead pid returns empty output so the marker check is already the right answer, and routing through the existing `deps.spawn` seam is the right shape.

**Finding 1 - the Windows checks on this PR were empty successes.** On the original run both `test (windows-latest, 1/2)` and `(2/2)` had every real step `skipped`; only checkout and the summary writer ran. `script/ci-fast-path.mjs` forces the full 3-OS matrix on push events, the `ci:full-matrix` label, basenames containing `windows`/`win32`, or four exact paths - and `packages/openclaw-core/src/reply-listener-process.ts` matches none of them, so a Windows-only fix never executed on Windows. Added that file to `platformSensitiveExactPaths` with a classifier test row (commit `6ef4963db`).

**Finding 2 - the fake-spawn test proves the branch is reached, not that the lookup works.** The stub manufactures the `CommandLine`. Added `reply-listener-process.real-process.test.ts` (commit `ef2dfcc66`): it spawns real children with and without the marker and probes them through the exported `isReplyListenerDaemonProcess` on every OS. Each child writes its own pid after exec, because probing the spawn-returned pid can race the exec and read the pre-exec command line - that raced on ubuntu before I fixed it, and it is worth knowing if you write similar tests.

**Evidence**

| | run | job | result |
|---|---|---|---|
| RED, dev code (probe #7896) | [34085548922](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/34085548922) | `test (windows-latest, 2/2)` | steps ran, **1 failing test in 5585**: the real-process test, marked child probed `false` |
| GREEN, this PR (`6ef4963db`) | [34086210908](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/34086210908) | `test (windows-latest, 2/2)` | steps ran, real-process test **pass in 3723 ms** (three CIM probes), 5666 tests |

Linux box (bun 1.4.0): `bun test packages/openclaw-core` 70 pass / 0 fail; classifier test 22 pass; removing the win32 branch fails exactly the fake win32 test (your claimed mutation, reproduced); removing the new classifier path fails exactly the new classifier row; `bun run typecheck` exits 0. Captured under `.omo/evidence/20260907-7885-windows-reply-listener-identity/` in this branch.

Merging as-is with those two commits on top.

code-yeongyu:
Merged into dev as 6d404434b with the two commits described above (real-process regression test + classifier exact path). Thanks for the fix and for the clear write-up of how you found it - the bundle-grep approach for calls that cannot work on Windows is worth reusing.

PR #7888, opened 2026-09-07
fix(monitor): kill the monitored process tree on Windows

LilMGenius:
On Windows the monitor never kills anything. `killProcessGroup` calls `process.kill(-pid, signal)`, Windows has no process groups, and the empty catch swallows the rejection. The child is spawned `detached: true`, so nothing else reaps it.

Measured on Windows against a real detached child:

```
child pid: 46476  alive before: true
process.kill(-pid, "SIGTERM")  -> THREW ESRCH
alive after negative-pid kill: true
taskkill /PID 46476 /T /F      -> exit 0
alive after taskkill: false
```

So a monitor whose `maxRuntimeMs` expires runs its watchdog, calls `kill("SIGTERM")`, waits the 5s grace, calls `kill("SIGKILL")`, and then resolves `{ code: null, signal: "SIGALRM" }` to the caller while the process it was watching is still running. Every timeout leaks one detached tree, and the caller is told it was terminated.

### The change

Use `taskkill /PID <pid> /T /F` on win32, which is what `execute-hook-command.ts` in this repo already does for the same reason. It is addressed through `SystemRoot` rather than PATH, because PATH can be missing System32, which is the failure in #6738. A bare liveness probe (`signal === 0`) still goes through `process.kill` with a positive pid, since that works on Windows.

`platform` and `killProcessTree` join `spawn` as optional deps, matching how this file already injects its seams, so the test spawns nothing real.

### About the two edited tests

They assert `killCalls` equals `[{pid: -4321, ...}]`, which is POSIX process-group semantics. They passed on Windows only because they stub `process.kill`, so the negative pid never reached the OS. They now say `platform: "linux"` outright. No assertion changed, and without that they would exercise the new branch on `windows-latest` and go red in your CI.

| | before | after |
|---|---|---|
| `process.test.ts` | 3 pass, 2 fail (on Windows, with the fix and unpinned platform) | 5 pass, 0 fail |
| `bun test .../features/monitor` | | 96 tests, 0 fail |

Removing the win32 branch fails the new test. `bun run typecheck` in `packages/omo-opencode` exits 0.

Found by grepping the published `oh-my-opencode@5.0.0-beta.45` bundle for POSIX-only constructs and checking each for a Windows guard. The other `process.kill(-pid)` sites are guarded; this one was not. No ordering against anything else I have open.

---
## Summary by cubic
Fixes Windows monitor kills so a watchdog timeout actually terminates the process tree, where before `process.kill(-pid)` silently failed and left the detached child running.

**Bug Fixes**
- On Windows, `killProcessGroup` now uses `taskkill /PID <pid> /T /F` via `SystemRoot` instead of a negative-pid kill.
- The liveness probe (`signal === 0`) still uses `process.kill` with a positive pid.
- Tests pin POSIX process-group semantics to Linux and add a Windows case asserting `taskkill` is called.

<sup>Written for commit d67eec270bd6b7973520fd97db1927c6ed9b338d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7888?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/monitor/process.test.ts
+++ b/packages/omo-opencode/src/features/monitor/process.test.ts
@@ -135,7 +135,7 @@ describe("spawnMonitoredProcess", () => {
 
       const monitored = spawnMonitoredProcess(
         { command: "bun test", maxRuntimeMs: 60_000 },
-        { spawn: createFakeSpawn([subprocess], spawnCalls), ...clock.deps },
+        { spawn: createFakeSpawn([subprocess], spawnCalls), platform: "linux", ...clock.deps },
       )
 
       // when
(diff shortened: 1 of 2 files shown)


LilMGenius:
Closing this myself to cut an oversized queue. The defect is not the reason - I re-checked it today against `dev` at `db06fbe7a` and it still reproduces: `killProcessGroup` in `packages/omo-opencode/src/features/monitor/process.ts` still calls `process.kill(-pid, signal)` with no Windows branch, so an expired `maxRuntimeMs` still leaves the detached child running.

PR #7892, opened 2026-09-07
test(background-agent): build stale-sweep fixtures relative to the wall clock

code-yeongyu:
## Why
`task-poller-unreachable-server.test.ts` (#7854) pinned `FIXED_NOW = 2026-09-07T03:00Z` and spied `Date.now`, but after `afterEach`'s `mockReset()` the second test's `mockReturnValue` no longer governed the clock the poller read (reproduced on the box: the second test observes the real `Date.now()`). That made the "recent activity" assertion depend on wall time: the fixture (`02:55Z`) stays "recent" only while `Date.now() - 02:55Z < 45 min`. At **03:40Z today** it crossed the default stale timeout and the sweep correctly cancelled it — every shard on every OS is red from then on (dev CI 34080004387 `test (windows-latest, 1/2)`, release-state PR #7887 `test (macos-latest, 1/2)`), which is what killed the beta.46 prepare (run 34080011948).

The poller is correct: a 5-minute-old task under a 45-minute timeout must keep running. This is a test defect only.

## What
- Fixtures are built with `minutesAgo(n)` against the same `Date.now()` the poller reads; the spy and the absolute date are gone, so no fixture can rot again.
- Comment names the failure mode for the next reader.

## Evidence (ascii box, bun 1.4.0, omo dev 50aa9a8bf)
- RED (original file at 03:5xZ): `1 fail` — `Expected: "running" / Received: "cancelled"`.
- GREEN (this change): `2 pass, 0 fail`; whole `background-agent/` dir: `757 pass, 0 fail` across 63 files.

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/task-poller-unreachable-server.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/task-poller-unreachable-server.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, describe, expect, mock, spyOn, test } from "bun:test"
+import { describe, expect, mock, test } from "bun:test"
 import { unsafeTestValue } from "../../../../../test-support/unsafe-test-value"
 import { checkAndInterruptStaleTasks } from "./task-poller"
 import type { BackgroundTask } from "./types"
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Superseded by #7891 (LilMGenius), which was opened first and implements the same wall-clock-relative fixture fix. My box RED/GREEN evidence is posted there as the approving review.

PR #7907, opened 2026-09-07
fix(omo-senpi): carry the quick category chain into the memorian judge child

### LINKED ISSUE #7904: memorian judge child is built without the in-category model fallback chain
## Summary

The memorian judge child is built WITHOUT the in-category model fallback chain, so when its primary model returns a transient overload it has no rung to fall back to. Other child paths (task subagents, the facts extractor) do pass the chain. This is a routing-path gap at the ChildSpec boundary, not a bug in the fallback engine.

## Mechanism

At `packages/omo-senpi/src/components/memory/memorian-judge-run.ts:101` the judge spec is built with the resolved model and thinking level, but **not the resolved fallbacks**:

```ts
runner.start(buildMemorianJudgeSpec({
  launch: input, runId, runDir, agentDir: ...,
  model: ... findModelReference(input.modelRegistry, resolution.model),
  ...(resolution.thinking === undefined ? {} : { thinkingLevel: resolution.thinking }),
  accepted,
}))
```

`resolution.fallbacks` — the in-category chain (e.g. the `quick` category's rungs 2-3) — is never passed. Consequently the in-process child is constructed with `retry.modelFallback: false` and the retry-fallback engine has no chain available mid-loop.

## Evidence of asymmetry

Files that reference `fallbackModels` include `packages/senpi-task/src/manager/runner.ts`, `packages/senpi-task/src/category/resolver.ts`, `packages/senpi-task/src/model-chain.ts` and `packages/model-core/src/fallback-chain-from-models.ts`.

`packages/omo-senpi/src/components/memory/memorian-judge-spec.ts` does **not** appear among them — the judge spec builder has no `fallbackModels` surface at all. So the chain is not dropped by a decision inside the judge; there is simply no parameter for it.

## Why this is a gap rather than a deliberate refusal

The code documents a deliberate refusal for the **beyond-category** ladder only — `registry_fallback` and `session_inherit`. Refusing to escape the category is a defensible policy choice. Refusing the **in-category** chain is not documented anywhere, and it is the chain the operator explicitly configured for that category.

## Observed impact

On a provider returning `server_is_overloaded`, the judge fails outright instead of trying the next configured rung. Related work already landed for the *classification* half of this: a companion fix makes the gate classify an upstream 503 immediately as `child_failed_upstream` rather than hanging to deadline exhaustion. That improves the failure *report*; this issue is about the judge not attempting recovery in the first place.

## Suggested scope

Give `buildMemorianJudgeSpec` a `fallbackModels` surface and pass `resolution.fallbacks` through it, so the judge child receives the same in-category chain the task and facts paths already get. Leave the beyond-category refusal exactly as documented.

## Provenance

Root-caused from source on omo `dev` (09434eb). The `memorian-judge-run.ts:101` call site and the `fallbackModels` reference set above were verified directly against `dev` when filing.

code-yeongyu:
## Summary

Fixes #7904. The memorian judge child never received the quick category's configured model chain: `resolveReflectionModel("quick")` returned `resolution.fallbacks`, but the judge spec carried only `model`, so the child session was created with `retry.modelFallback: false` and a mid-turn `503 server_is_overloaded` on the primary rung ended the run. After this PR the judge carries the in-category chain as `selectedModel` + `fallbackModels`, senpi's retry-fallback controller rotates rungs mid-turn exactly as it does for task subagents and the main session, and a run only fails when the whole chain is exhausted.

The beyond-category ladder (`registry_fallback` / `session_inherit`) is still refused: the judge stays pinned to the quick category. What changes is that the category's *own* rungs are no longer silently dropped.

## Why the #7899 event-stream fail-fast is replaced

`memorian-judge-run.ts` stated its own premise: *"A hard-down provider leaves the child inside its retry chain, so the turn never settles and the deadline is the only exit... Resolving early classifies the run without waiting out the retries."* With `fallbackModels` passed, the retries **are** the recovery, so classifying on the first upstream `message_end` would abort the rotation the chain configures. The listener, its `wrappedCreateSession` pre-turn subscription, and `classifyJudgeEvent` are deleted; `classifyJudgeTurn` derives `child_failed_upstream` from the settled failure message with the same `UPSTREAM_FAILURE_PATTERN`, so the two classifications cannot drift.

Latency-to-report does not regress: the engine settles the turn itself once the budget and chain are exhausted (measured below, ~2.4 s), so the gate still reports `failed/child_failed_upstream` with the sanitized reason instead of waiting for the deadline.

## Why a one-retry same-model budget is load-bearing

`tool_call`-origin judges run under a 90 s deadline (`TOOL_CALL_JUDGE_DEADLINE_MS`, now exported from `memorian-trigger.ts`). Senpi's default profile is `turn.maxRetries: 5` with a 2 s exponential base, i.e. ~62 s of same-model backoff before `tryFallback()` is even consulted, so a two-rung chain could never be exercised inside that deadline. The spec therefore sets `retry: { maxRetries: 1 }` **only when the category has fallbacks**; a single-model quick category keeps the engine default because same-model retries are its only recovery. This rides a new additive `ChildSpec.retry` seam in senpi-task (`createRuntimeFallbackSettings(selectedModel, fallbackModels, retry?)`); every existing caller is byte-identical.

## Changes

- **senpi-task**: `ChildSpec.retry?: { maxRetries?, baseDelayMs? }` merged into the child's in-memory retry settings (`runtime-fallback-settings.ts`, `child-options.ts`, `in-process.ts`).
- **memory / judge chain**: `memorian-judge-chain.ts` converts `resolution.fallbacks` into `ResolvedModelRecord`s keyed by the bare primary selector; `memorian-judge-spec.ts` sets `selectedModel`, `fallbackModels` and the retry budget; `memorian-judge-run.ts` passes the chain.
- **memory / classification**: `memorian-judge-outcome.ts` returns `child_failed_upstream` from the settled outcome; the event-stream listener and `classifyJudgeEvent` are removed from `memorian-judge-run.ts`.
- **Docs**: `memorian-runner.ts` header/launch comments no longer claim "no fallback ladder"; `AGENTS.md` rows for the runner, judge spec/chain and judge outcome updated.
- **Bundle**: `plugin/extensions/omo.js` and `omo-task.js` regenerated with bun 1.4.0 on Linux; `build-extension.mjs --check` reports current.

## QA & Evidence

All tests ran on a Linux x86_64 ascii box (`bx_vthb6wa4`, bun 1.4.0, `bun install --frozen-lockfile --ignore-scripts`), never on the local machine. RED was captured before the production change.

- **What was tested:** `memorian-runner-fallback.test.ts` — the REAL `InProcessRunner` + REAL senpi `AgentSession` against a fake provider whose primary rung answers `OpenAI API error (503): auth_unavailable (model gpt-5.6-luna, server_is_overloaded)`, run with `deadlineMs = TOOL_CALL_JUDGE_DEADLINE_MS` (90 s).
  **Observed result:** healthy fallback: provider calls `[dead-primary, dead-primary, healthy-fallback]`, a `retry_fallback_applied` event, verdict `empty`, 2.2 s. Every rung 503: calls `[dead-primary, dead-primary, healthy-fallback]` (a switched-to rung starts at attempt 1 of the same budget), the turn settles, gate `failed/child_failed_upstream` with the 503 reason, 2.1 s — never `dropped/deadline`. On dev (RED) both runs stopped after the first `dead-primary` call with `child_failed_upstream`.
  **Why sufficient:** proves the rotation on the engine, not just the wiring, on the strict 90 s origin.
- **What was tested:** `memorian-runner-chain.test.ts` — quick `{ model: mock-1, fallback_models: [mock-2, mock-3] }`.
  **Observed result:** child `settingsManager` has `modelFallback: true`, `chains["omo-mock/mock-1"] == ["omo-mock/mock-2", "omo-mock/mock-3"]`, `maxRetries: 1`; single-model quick keeps `modelFallback: false`. RED on dev: `modelFallback: false`, `maxRetries: 5`.
- **What was tested:** `memorian-runner-upstream.test.ts` (replaces the #7899 fail-fast trio) — settled upstream error, mid-turn upstream event that the engine recovers from, settled non-upstream error, silent child.
  **Observed result:** `child_failed_upstream` + reason + operator warn; `nudged` (RED on dev: `child_failed_upstream`); `child_failed`; `dropped/deadline`.
- **What was tested:** `memorian-judge-spec.test.ts`, `memorian-judge-chain.test.ts`, `memorian-judge-outcome.test.ts`, `runtime-fallback-settings.test.ts`.
  **Observed result:** focused 8-file run 35 pass / 0 fail after GREEN (9 fail + 1 unhandled at RED).
- **What was tested:** `bun test packages/omo-senpi/src/components/memory packages/senpi-task` on the box.
  **Observed result:** 3070 pass / 2 skip / 4 fail; the 4 failures are the bwrap sandbox tests (`sandbox.test.ts`, `sandbox-absent-paths.test.ts`), reproduced identically on a clean `dev` (`b210521`) clone in the same box, which has no `bwrap`. (A fifth failure in `run-terminal-precedence.integration.test.ts` was caused by my transfer omitting the shipped `memory-run-supervisor.mjs`; it passes with the file present.)
- **What was tested:** `node packages/omo-senpi/plugin/scripts/build-extension.mjs --check` after regeneration on the box (CI's exact gate).
  **Observed result:** "omo-senpi extension build is current", exit 0; committed bundles match the box checksums.
- **What was tested:** `bun run typecheck` (repo root, tsgo across all packages) locally.
  **Observed result:** exit 0.
- The full-repo `bun test` run on the box is in flight at PR-open time; its summary will be posted as a comment before merge.

## Risks & Residuals

- **Contract change for the gate ledger:** a provider failure on the primary rung is now recovered by the next rung when one is configured; only chain exhaustion reports `child_failed_upstream`. `scripts/qa/memorian-gate-e2e.mjs` S2 is unaffected: its mock is a 500 "mock judge outage" without an upstream marker on a single-model quick category, so its `cause === "child_failed"` expectation still holds.
- **Secret-bearing upstream messages** classify as `child_failed` (the reason is `redacted` before the pattern test). Acceptable: an upstream 503 line carries no secret, and the cause must not be derived from text the gate refuses to store.
- **Outcome records keep `model: resolution.model`** (the launch primary) even when a fallback rung produced the verdict; the run dir's child transcript shows the actual model. Left as-is to keep the record schema unchanged.
- The facts extractor already traverses `resolution.fallbacks` at launch time; giving it the same mid-turn seam is a separate change.

## Related Issues

Closes #7904. Supersedes the event-stream half of #7899.

---
## Summary by cubic
The memorian judge child now carries the quick category's configured model chain, so a mid-turn upstream failure rotates through the category's own rungs instead of ending the run. Previously only `model` was passed, leaving the child with `modelFallback: false`; now a run fails only when the whole chain is exhausted.

**Behavior**
- With fallbacks configured, the judge spec sets `retry.maxRetries: 1`; the engine's default ~62s same-model backoff would never reach the next rung inside the 90s `tool_call` deadline, while single-model categories keep the engine defaults.
- `child_failed_upstream` is now classified from the settled turn failure instead of the event-stream fail-fast listener, so the engine's mid-turn rung rotation isn't aborted.
- The beyond-category ladder (`registry_fallback` / `session_inherit`) is still refused; only the quick category's own rungs are carried into the child.
- `senpi-task` gains an additive `ChildSpec.retry` seam that child specs may use to override the same-model retry budget; existing task callers are unaffected.

**Validation**
- Live-engine tests with a fake provider whose primary rung returns 503 show rotation to the fallback rung and a settled verdict inside the deadline; when every rung fails the run settles quickly and reports `child_failed_upstream` rather than `dropped/deadline`.
- Unit tests cover the chain conversion, the retry-budget wiring, and settled-outcome classification; the extension bundle was regenerated and the build check passes.

<sup>Written for commit c10dcc4497d12af2f4b499bc4009256b285c5597. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Full-suite result (promised in the QA section)

Ran on the same Linux x86_64 ascii box (bun 1.4.0, `bun install --frozen-lockfile --ignore-scripts`).

- Full `bun test` on the transferred tree reported 46 failing test names, but that tree was shipped without `.git`, so every git-backed audit (DMCA provenance, markdown link audit, tracked-bundle markers, package layout, ...) failed for lack of git metadata.
- Control: the same 21 failing files were re-run from a real `git clone` of this branch (`c10dcc449`) and of clean `dev` (`b210521`) on the same box. Both produced the **identical 7-failure set**: the 4 bwrap-dependent reflection sandbox tests (no `bwrap` on the box), the DMCA submodule-pin check (shallow clone), and two Codex installer version/payload checks. Only-on-this-branch failures: none. Only-on-dev failures: none.
- Memory component + senpi-task suites: 3070 pass / 2 skip / 4 fail, the 4 being the same bwrap tests, reproduced identically on clean dev.

Together with CI green on this PR (23 success / 6 skipped), the change introduces no test regression relative to `dev`.

Issue #7912, opened 2026-09-07
reflection scheduler: dead-launcher reservation never reclaimed when a retired run dir shadows it (plus unbounded pending merge, temp-file leak, commit-path secret scan gap)


code-yeongyu:
Found during a research audit of the memory subsystem (read-only, no code change in this report). Three related defects in the reflection reservation lifecycle, measured on a live host at dev b210521:

## D7 — a dead-launcher active reservation is never reclaimed when a retired run dir shadows it

**Observed**: `runtime/reflection/active.lock` holds `reflection-run-1` (trigger dream / origin shutdown) reserved 2026-08-18T07:01:33Z by launcherPid 81327 (dead; `kill -0` fails). `runs/reflection-run-1/ledger.json` belongs to the *earlier* run with the same id (completed 2026-08-13T09:29, outcome merged). The reservation has not been reclaimed in 20 days; every subsequent dream/reflection request merged into `pending.json` (run-2) which now carries 280 conversation snapshots (165 MB) and nothing has launched — zero reflection/dream commits since 2026-08-13 while the memory tool made 1054 commits.

**Root cause**: `packages/omo-senpi/src/components/memory/worker/run-reconciliation.ts:88-89` — `reconcilePrelaunch` returns early when `runs/<runId>/ledger.json` exists, *before* the age/hostname/pid-liveness checks. The retired run dir (ledger finalizedAt 2026-08-13 < reservedAt 2026-08-18) is treated as "this reservation launched", so the dead launcher is never detected. The disk-scanned run-id factory (reflection-run-id.ts) prevents *new* id collisions but does not unwedge existing state.

**Minimal fix**: in reconcilePrelaunch, treat the run dir as belonging to this generation only when its ledger/final timestamps are >= active.reservedAt; otherwise complete the reservation as failed so pending promotes. Add a test seeding a merged final.json for run-1 plus a newer active.lock with a dead launcherPid.

## D5 — pending reservation merge is unbounded

`REFLECTION_SNAPSHOT_MAX_BYTES=131072` (journal/cursor.ts) caps a single capture window, but `mergeSnapshots` in reflection/machine.ts unions snapshots by conversationId with no aggregate cap. Measured: pending.json = 183,040,129 B carrying 280 snapshots, one 6.4 MB. Suggest a cap on total pending conversations/bytes with score-based eviction.

## D6 — writeJsonAtomic leaves pending.json.tmp-<uuid> files

[...]

PR #7924, opened 2026-09-07
fix(omo-senpi): reclaim a dead-launcher reservation shadowed by a retired run dir (#7912)

MoerAI:
## Summary

Implements the minimal fix you specified for **D7** of #7912: a dead launcher's reflection reservation was never reclaimed when a retired run dir shadowed it, so nothing ever launched again.

Scope is **D7 only**. D5 (unbounded pending merge), D6 (temp-file leak), and D8 (commit-path secret scan) are deliberately untouched.

## Root Cause

`reconcilePrelaunch` returned as soon as `runs/<runId>/ledger.json` existed:

```ts
const runDir = join(context.identity.paths.reflection, "runs", active.runId)
if (existsSync(join(runDir, "ledger.json"))) return undefined   // <- before every liveness check
const prelaunchPath = join(runDir, "prelaunch.json")
if (existsSync(runDir) && !existsSync(prelaunchPath)) return undefined
if (context.now() - Date.parse(active.reservedAt) <= 60_000 || ...) return undefined   // never reached
const liveness = (context.getPidLiveness ?? getPidLiveness)(active.launcherPid)        // never reached
```

That early return runs **before** the age, hostname, and pid-liveness checks. A retired run dir reusing the same run id therefore looked like proof that the *current* reservation had launched, so the dead launcher was never detected. As reported: the reservation sat held for 20 days, every later request merged into `pending.json` (280 snapshots / 165 MB), and zero reflection commits happened while the memory tool made 1054.

## Changes

| File | Change |
|------|--------|
| `packages/omo-senpi/src/components/memory/worker/run-reconciliation.ts` | Make the ledger-present early return generation-aware (9 added / 2 removed) |
| `packages/omo-senpi/src/components/memory/worker/run-reconciliation.test.ts` | Regression test seeding a retired merged run plus a newer reservation with a dead launcher |

The run dir counts as this generation only when its terminal timestamp (`final.json.finishedAt`, falling back to `ledger.finalizedAt`) is `>= active.reservedAt`. A demonstrably older terminal run falls through to the existing dead-launcher path and the existing `complete(active.runId, "failed", ...)`, so pending promotes.

Deliberate safety properties:

- **In-flight runs are unaffected.** A missing, unparseable, equal, or newer terminal timestamp keeps the original early return, so a live run is never reclaimed. The comparison is written so `NaN` from an unparseable timestamp also falls back to the early return.
- **The prelaunch guard is narrowed to ledger-less dirs only.** A retired run already cleaned up its own worktree and carries no `prelaunch.json`; without this it would still be blocked. Worktree cleanup remains gated on `prelaunch.json` existing, so nothing is removed blindly.

## Reproduction (before fix)

Reverting only the source change, keeping the new test:

```
(fail) reflection and dream run reconciliation > #given a retired merged run with a ledger and a newer
       reservation with a dead launcher #when reconciled #then the reservation is reclaimed

297 | expect(results).toEqual([{ runId: item.ledger.runId, outcome: "failed" }])
-     "outcome": "failed",

10 pass, 1 fail
```

Reconciliation returned `[]` and left the reservation held.

## Verification (after fix)

```
236 pass
  0 fail
1133 expect() calls
Ran 236 tests across 36 files.
```

## Test

- Regression test: `run-reconciliation.test.ts` (newly added), seeded exactly as you described - a merged terminal run plus a newer `active.lock` with a dead `launcherPid`, using the existing injectable `getPidLiveness` seam.
- Suite: `bun test packages/omo-senpi/src/components/memory/worker/` - 236 pass, 0 fail.
- Typecheck: `bun run typecheck` - exit 0.
- No generated bundle is committed, matching how recent `packages/omo-senpi/src/` changes land on `dev`.

Addresses D7 of #7912.

---
## Summary by cubic
Fixes a dead launcher's reflection reservation never being reclaimed when a retired run dir shadowed it, so pending promotions now go through again.

- `reconcilePrelaunch` returned early as soon as `ledger.json` existed, skipping the age, hostname, and pid-liveness checks; a retired run dir reusing the same run id looked like proof the current reservation had launched.
- The ledger-present early return now only counts when the run dir's terminal timestamp (`final.json.finishedAt`, falling back to `ledger.finalizedAt`) is at least the reservation's `reservedAt`; older terminal runs fall through to the existing dead-launcher path.
- In-flight runs are unaffected: missing, unparseable, equal, or newer terminal timestamps keep the original early return.
- The prelaunch guard is narrowed to ledger-less dirs only, since retired runs have no `prelaunch.json`.
- Scope is D7 of #7912 only; D5, D6, and D8 are untouched.
- Adds a regression test seeding a retired merged run plus a newer reservation with a dead launcher.

<sup>Written for commit a2a88c0b10a45978f1e00e492e61cf0ecdbcf146. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:

MoerAI:
Flagging a CI blocker I cannot resolve from a contributor machine, so you are not left guessing why this is red.

`senpi-compatibility (ubuntu-latest)` fails on `build-extension.mjs --check` with `extension build is not current: stale-output`. I tried both directions and both fail:

- **Source-only** (this is the current state of the PR, matching how the last 12 `packages/omo-senpi/src/` commits landed on `dev`, including your own e26fef656): the committed bundles are now stale against the changed source, so the digest check fails.
- **Source + regenerated bundles**: `--check` exits 0 on my machine, but CI still reports `stale-output` on the exact same commit. The artifacts are toolchain-dependent, and I am on Bun 1.3.14 / macOS arm64 against CI's Bun 1.4.0 / linux x64, so I cannot produce bytes that match.

I have left the PR **source-only**, which keeps the diff reviewable at 2 files / 34 lines and avoids committing artifacts built with the wrong toolchain. The bundles need a regeneration from your side or from CI.

The fix itself is verified independently of that gate: reverting only the source change makes the new regression test fail with the reclamation assertion, and with it `bun test packages/omo-senpi/src/components/memory/worker/` is 236 pass / 0 fail with `bun run typecheck` clean.

If there is a supported way for outside contributors to produce matching bundle bytes, I am happy to follow it. Otherwise this may be worth folding into a maintainer-side commit.

code-yeongyu:
Thanks for tracking this down — dev now carries the same repair via d69e67401 (retired-generation reclaim) and #7947 (missing-identity / older-ledger ghosts), so closing this as superseded. Closes #7912 if still open.

PR #7933, opened 2026-09-07
fix(omo-senpi): dispose idle-injection coordinator on session shutdown

### LINKED ISSUE #7932: [Bug]: Deferred idle-injection flush crashes OMO after reload with stale extension generation
> *This was generated by AI during triage.*

## Summary

OMO Native exits with an uncaught `stale extension generation after reload` error while sending an extension message. Investigation reproduced a lifecycle defect in the shared `IdleInjectionCoordinator`: a deferred flush can retain the previous generation's `pi.sendMessage` after reload invalidates that API.

The failure mechanism is reproduced against the installed coordinator and the real Senpi loader API. The original notification producer and the supplied minified bundle coordinate are not fully correlated; see the limitations below.

## Environment

- `omo-ai`: `5.0.0-0.beta.48`
- `@code-yeongyu/senpi`: `2026.9.7-2`
- macOS, Darwin `27.0.0`, arm64
- Installed through npm under Node.js `v24.19.0`
- Deterministic replays ran under Node.js `v24.19.0`. The original process's runtime was not independently recovered.
- The npm `omo-ai` beta tag matched the installed version when checked.

## Reported error

Personal installation paths have been replaced with `<OMO_PACKAGE>`.

```text
167 |     // can replay registrations in original call order (last-registration-wins).
168 |     let nextProviderRegistrationOrder = 0;
169 |     const eventBusUnsubscribers = new Set();
170 |     const assertActive = () => {
171 |         if (state.staleMessage) {
172 |             throw new Error(state.staleMessage);
                            ^
error: stale extension generation after reload
      at assertActive (<OMO_PACKAGE>/node_modules/@code-yeongyu/senpi/dist/core/extensions/loader.js:172:23)
      at assertActive (<OMO_PACKAGE>/node_modules/@code-yeongyu/senpi/dist/core/extensions/loader.js:279:17)
      at sendMessage (<OMO_PACKAGE>/node_modules/@code-yeongyu/senpi/dist/core/extensions/loader.js:398:13)
      at #s (<OMO_PACKAGE>/plugin/extensions/omo.js:2:21100)
      at <OMO_PACKAGE>/plugin/extensions/omo.js:2:21100
```

## RCA

1. [`composeOmoSenpiExtension`](https://github.com/code-yeongyu/oh-my-openagent/blob/a470d3d9ddd0893e1c2b56ff098ba8061358c924/packages/omo-senpi/src/extension/compose.ts) constructs one coordinator with a closure over the current `pi` and a 200 ms timer:

   ```ts
   const idleCoordinator = new IdleInjectionCoordinator(
     (message, options) =>
       pi.sendMessage(message, { triggerTurn: true, deliverAs: options.deliverAs }),
     { scheduleFlush: (flush) => void setTimeout(flush, 200) },
   )
   ```

2. [`IdleInjectionCoordinator.scheduleFlush()`](https://github.com/code-yeongyu/oh-my-openagent/blob/a470d3d9ddd0893e1c2b56ff098ba8061358c924/packages/omo-senpi/src/extension/idle-injection-coordinator.ts) schedules a detached callback. The coordinator has no disposal method, retained timer handle, or generation fence. Its private flush clears the queue before delivery. On a synchronous delivery error it calls `onDeliveryFailed`, then rethrows.

3. In the installed Senpi, `dist/core/agent-session.js:6115-6190`, `AgentSession.reload()` emits `session_shutdown`, reloads resources, builds a replacement runner, and calls:

   ```js
   oldExtensionRunner.invalidate("stale extension generation after reload");
   ```

4. `dist/core/extensions/runner.js:552-555` invalidates the shared runtime. `dist/core/extensions/loader.js:201-209` marks it stale and removes tracked event-bus subscriptions. That does not cancel the OMO coordinator's independently scheduled callback.

5. If an old pending flush runs after invalidation, the captured `pi.sendMessage` calls `assertActive()` synchronously at `loader.js:398`, then `:279`, then throws at `:172`. The host message-delivery action is never reached.

6. The coordinator rethrows from the detached callback. This is outside the runner's extension-event exception boundary, so it can become a process-level `uncaughtException` rather than a contained extension error.

The defect is ownership of deferred work across reload, not the stale-generation guard. The guard correctly prevents the retired extension from writing into a replacement session.

## Deterministic reproduction

The original interactive keystroke sequence was not recovered. The following interleaving was reproduced without changing installed files, making model calls, or relying on sleeps:

1. Extract the actual installed coordinator class `c1`, its source-order table `l1`, and its exact `sendMessage` closure from `plugin/extensions/omo.js`.
2. Evaluate that code in `node:vm`. Create a real extension API with the installed `createExtensionRuntime()`, `createEventBus()`, and `loadExtensionFromFactory()`.
3. Capture the coordinator's injected `scheduleFlush` callback instead of waiting for a timer. Replace only the host delivery action with an in-memory receipt.
4. Enqueue a non-passive `ulw-continuation` notification and call `scheduleFlush()`.
5. In the active control, invoke the captured flush. It delivers exactly one message.
6. In the failure case, first call `runtime.invalidate("stale extension generation after reload")`, then invoke the same captured flush. It throws the exact error before host delivery.
7. In a separate subprocess, subscribe `uncaughtExceptionMonitor` before invalidation and run the captured flush through `queueMicrotask`, with no catching `uncaughtException` handler. The monitor observes the error and Node exits with code 1.

The scheduler is controlled to force the lifecycle ordering. This reproduces the queue/API interaction and the uncaught escape; it does not claim an end-to-end replay of the original interactive session.

### Captured reproduction debug output

This is output from the constructed replay, not recovered production logs:

```text
ACTIVE_CONTROL: scheduled=1 pending=1
ACTIVE_CONTROL: delivered=1 failures=0 pending=0 thrown=none
RETIRED_GENERATION: scheduled=1 pending=1
RETIRED_GENERATION: delivered=0 failures=1 pending=0 thrown=stale extension generation after reload
Error: stale extension generation after reload
    at Object.assertActive (file://<OMO_PACKAGE>/node_modules/@code-yeongyu/senpi/dist/core/extensions/loader.js:172:19)
    at assertActive (file://<OMO_PACKAGE>/node_modules/@code-yeongyu/senpi/dist/core/extensions/loader.js:279:17)
    at Object.sendMessage (file://<OMO_PACKAGE>/node_modules/@code-yeongyu/senpi/dist/core/extensions/loader.js:398:13)
    at c1.<anonymous> (evalmachine.<anonymous>:1:1247)
    at #s (evalmachine.<anonymous>:1:760)
    at evalmachine.<anonymous>:1:350
PASS: installed coordinator and real loader reproduce stale callback; no timers or files created
```

The assertion replay exited 0. The separate uncaught-exception replay produced:

```text
UNCAUGHT_MONITOR: stale extension generation after reload
```

Its stderr contained the same `loader.js:172 -> :279 -> :398 -> c1 callback -> #s` stack and the process exited 1, as expected for this reproduction.

### Captured local config-reload logs

These are actual entries from `agent/logs/config-reload.log`, with private paths replaced:

```jsonl
{"ts":"2026-09-07T21:37:22.979Z","level":"info","event":"reload_requested","reason":"config changed","paths":["<OMO_CONFIG>/omo.jsonc"]}
{"ts":"2026-09-07T21:37:23.743Z","level":"info","event":"reload_completed","durationMs":764}
{"ts":"2026-09-07T21:37:23.744Z","level":"info","event":"change_detected","registrationId":"builtin","paths":["<OMO_CONFIG>/omo.jsonc"]}
{"ts":"2026-09-07T21:37:23.744Z","level":"info","event":"reload_requested","reason":"config changed","paths":["<OMO_CONFIG>/omo.jsonc"]}
{"ts":"2026-09-07T21:37:24.154Z","level":"info","event":"reload_completed","durationMs":410}
{"ts":"2026-09-07T21:42:40.752Z","level":"info","event":"reload_requested","reason":"config changed","paths":["<OMO_CONFIG>/omo.jsonc"]}
```

There was no subsequent `reload_completed` in the captured log through `21:44:08.202Z`. This shared log has no session identifier, so it establishes reload activity but cannot prove that the last request belongs to the reported crash. No matching original exception was recovered from `session.log`.

## Expected behavior and suggested fix

Reload should retire all deferred work owned by the previous coordinator without terminating OMO or silently losing required notifications.

- Dispose the coordinator on `session_shutdown`, before the old runner is invalidated.
- Retain and cancel the batch timer; fence already-queued microtasks and late producers from a retired generation.
- Explicitly fail or hand off pending deliveries according to their lifecycle contract.
- Keep Senpi's stale guard. Contain an expected retired-generation delivery failure at the detached callback boundary without broadly swallowing unexpected errors.

Suggested regression cases: a pending timer across reload, a queued `flushSoon` microtask across retirement, late enqueue after retirement, and an active-generation control that still delivers exactly once.

## Limitations and related reports

- The supplied `omo.js:2:21100` is not the raw callback location in the inspected beta.48 bundle. The installed coordinator's `#s` is near raw character offset `1034921`; its `sendMessage` call is near `1037394`. The raw file has no source map, and the inspected Jiti transformation did not resolve the supplied coordinate either. A different loaded artifact or transformation remains possible.
- The lifecycle defect and uncaught escape are reproduced. Attribution of the original crash to a specific producer or timer remains unverified.
- Open and closed issues were searched using the full error, `"stale extension"`, `"stale generation"`, `reload notification`, and `reload completion`. No matching report was found.
- #7114 concerns a stale command context during memory-reflection shutdown. It is related lifecycle behavior, but not this deferred `sendMessage` crash.
- #7419 concerns synchronous watcher recursion under duplicate extension loads, a different failure. No evidence here establishes duplicate loading.
- Merged #6401 and #7002 add reload vetoes for running resident children and live DAG runs. Those guards do not establish disposal of this shared notification queue.

code-yeongyu:
## Summary

The `IdleInjectionCoordinator`'s deferred flush (200ms `setTimeout`) could fire after `AgentSession.reload()` invalidated the old extension runtime, causing `pi.sendMessage` to throw `stale extension generation after reload` as an uncaught exception that killed the process.

## Changes

**`idle-injection-coordinator.ts`**
- Add `dispose()` method: clears pending queue, sets `#disposed` flag
- Guard all entry points (`enqueue`, `scheduleFlush`, `flushSoon`, `#flush`) and deferred callback interiors against disposed state
- Expose read-only `disposed` getter

**`compose.ts`**
- Hook `pi.on('session_shutdown', ...)` to call `idleCoordinator.dispose()` before the extension runner invalidates the old generation

**Tests (3 new)**
- Deferred flush after dispose → harmless no-op (not a stale-generation crash)
- enqueue/scheduleFlush/flushSoon after dispose → silently ignored
- flushOnIdle after dispose → returns 0

Closes #7932

---
## Summary by cubic
Prevents idle-injection notifications from being lost during session reloads. Previously, a queued completion could be dropped while `notified_epoch` remained persisted, so the next `session_start` reconcile skipped it; now retired coordinators report accepted deliveries as failed and redeliver them.

- Retiring the coordinator cancels its 200 ms timer, fails queued entries with `IdleInjectionRetiredError`, and refuses later enqueues.
- `senpi-task` rolls back failed notification epochs, records the failure, and re-enters the retry path.
- Producers handle refusals by throwing, retrying, releasing delivery reservations, or logging derived continuations.
- Adds coordinator, composition, and end-to-end reload recovery tests.

Closes #7932.

<sup>Written for commit 32f9d4fbd98281b257074669f09110c074d9bf6e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aslaii:
> *This was generated by AI during triage.*

### [P1] Preserve pending completion notifications when disposing the coordinator

Reviewed commit `cc8aeaae14f374a40b591cd210c5a3bf08a22ab2`. The disposal guard prevents the reproduced stale-`sendMessage` crash, but [`dispose()`](https://github.com/code-yeongyu/oh-my-openagent/blob/cc8aeaae14f374a40b591cd210c5a3bf08a22ab2/packages/omo-senpi/src/extension/idle-injection-coordinator.ts#L73-L82) clears pending notifications without delivering, failing, or handing them off. Enqueue after disposal also silently returns.

This matters for task completions because:

1. [`createParentNotifier.enqueue()`](https://github.com/code-yeongyu/oh-my-openagent/blob/cc8aeaae14f374a40b591cd210c5a3bf08a22ab2/packages/omo-senpi/src/components/task/parent-notifier.ts#L25-L45) returns successfully after adding the notification to the deferred coordinator queue, before actual delivery.
2. [`createCompletionNotifier.deliverRecord()`](https://github.com/code-yeongyu/oh-my-openagent/blob/cc8aeaae14f374a40b591cd210c5a3bf08a22ab2/packages/senpi-task/src/completion/notifier.ts#L94-L108) then immediately persists `notified_epoch`.
3. If shutdown disposes the coordinator before its flush runs, the notification disappears. [`reconcileUnnotifiedNotifications()`](https://github.com/code-yeongyu/oh-my-openagent/blob/cc8aeaae14f374a40b591cd210c5a3bf08a22ab2/packages/senpi-task/src/completion/notifier.ts#L155-L175) skips the already-notified epoch on resume, so the automatic completion notification is not recovered. The underlying task result remains stored.

The early acknowledgement is an existing adapter behavior; the new unconditional queue clearing does not account for it. Calling `onDeliveryFailed` alone would not fix ordinary task completions because their adapter currently supplies no receipt callback.

**Observed review evidence:** 15 standalone coordinator tests passed. Probes using the real installed Senpi loader API passed active delivery, a retired-generation timer, and a queued microtask after disposal. A separate probe enqueued a completion with both receipt callbacks, disposed the coordinator, and flushed:

```text
pending=0
delivered=0
onFlushed=0
onDeliveryFailed=0
```

That probe directly demonstrates silent queue loss. The no-redelivery consequence follows from the inspected notification-epoch and reconciliation code above; it was not a full interactive-session replay. The normal repository test invocation was blocked by missing workspace dependencies, so this is not a full build/CI approval.

Please preserve or hand off pending completions across reload, or change durable acknowledgement to occur only after actual delivery. Add a regression covering a background completion queued immediately before disposal, followed by same-session reload/reconciliation, and assert exactly one eventual notification without a stale-generation exception.

code-yeongyu (commented):
**REQUEST_CHANGES** (filed as a review comment: GitHub rejects the `REQUEST_CHANGES` event from the PR author — `POST /pulls/7933/reviews` → HTTP 422 `Review Can not request changes on your own pull request`. Treat this as blocking review feedback.)

**Verdict.** The seam is right and I verified it: senpi emits `session_shutdown` on the *old* runner (`coding-agent/src/core/agent-session.ts:6756`, reason `"reload"`) and only then calls `oldExtensionRunner.invalidate("stale extension generation after reload")` (`:6815`), so a flag set from that handler really does close the crash window in #7932. But the semantics chosen here convert a loud crash into silent, permanent notification loss, and the change is effectively unverified: I deleted the entire fix hunk (`compose.ts:104-109`) and **every** test that touches `composeOmoSenpiExtension` still passed (`bun test packages/omo-senpi/src/extension/` → 69 pass / 0 fail; `omo-native-component.test.ts` → 7 pass / 0 fail), and I removed 3 of the 5 new `#disposed` guards and all 15 coordinator tests still passed. Findings 1-3 are blocking on substance, 4 blocks merge mechanically, 5-6 are small. Two things I explicitly checked and did *not* find problems with, for the record: double-dispose is idempotent, and there is no cross-session refcount hazard, because each activation owns exactly one coordinator and senpi rebuilds the runner per session replacement (`agent-session-runtime.ts:202-208`, `:450`), so permanent disposal is safe.

**1. `dispose()` drops queued injections with no failure receipt, and post-dispose `enqueue` reports success — background task completions are then lost permanently, not just delayed. (blocking)**

`packages/omo-senpi/src/extension/idle-injection-coordinator.ts:75` clears `#pending` without calling `injection.onDeliveryFailed` for any dropped entry, and `:80-83` makes `enqueue()` after dispose a silent success (`void` return, no throw). Follow the real producer chain:

- `packages/omo-senpi/src/components/task/parent-notifier.ts:24-37` — a streaming parent's completion goes `coordinator.enqueue(...)` + `scheduleFlush()`, i.e. it sits in the 200 ms `setTimeout` window from `compose.ts:101`.
- `packages/senpi-task/src/completion/notifier.ts:243-253` — `tryEnqueue` treats "did not throw" as `{ ok: true }`.
- `packages/senpi-task/src/completion/notifier.ts:102-106` — on `ok`, `persistNotified()` writes `notified_epoch = run_epoch` to the store **immediately**, before the flush ever runs.
- `packages/senpi-task/src/completion/notifier.ts:165` — crash recovery (`reconcileUnnotifiedNotifications`, run on the post-reload `session_start` via `components/task/event-bridge.ts:68`) skips any record with `notified_epoch >= epoch`.

Consequence: a child that finishes inside the 200 ms batch window before a `/reload` (or a config-watch reload) is recorded as notified, its queued injection is discarded by `dispose()`, and no path can ever redeliver it. The parent silently never learns its background task completed. The same silence defeats `components/task/member-liveness.ts:101-108`, which wraps `enqueue` in `try/catch` precisely so a failure runs `failDelivery` → retry; after this PR the catch never fires and `delivered`/`pendingAcknowledgement` keep the key, so the liveness notice is dropped with no retry and no log. Note the codebase's existing contract for a deliberate drop is to *record* it (`notifier.ts:145-148` → `dropEntry` appends `notification_dropped`); this PR drops with no event, no receipt, and no log.

Fix: in `dispose()`, drain before clearing — `for (const injection of this.#pending.values()) injection.onDeliveryFailed?.(new Error("idle-injection coordinator disposed"))` — and make post-dispose `enqueue()` throw (or return `false` and update both call sites) instead of returning as if queued. A throw is already the handled path in both producers (`member-liveness.ts:106`, `notifier.ts:250-252`), and it routes the completion to `recordFailure` + `scheduleRetry`, which sets `notification_failed_epoch` and lets the post-reload reconcile redeliver. That turns the crash into a *recovered* notification rather than a lost one.

**2. The line that actually implements the fix has zero coverage. (blocking)**

`packages/omo-senpi/src/extension/compose.ts:107-109` is the whole behavioral change, and nothing exercises it. Evidence, not opinion: I removed `compose.ts:104-109` verbatim and re-ran everything that references `composeOmoSenpiExtension` — `packages/omo-senpi/src/extension/` (12 files) 69 pass / 0 fail, `packages/omo-senpi/src/components/telemetry/omo-native-component.test.ts` 7 pass / 0 fail. A reviewer or a future refactor can delete the fix and CI stays green. The harness for this already exists and is used in the same file's test (`compose.test.ts:26-66` composes real components against `FakeExtensionAPI` and calls `pi.dispatch("session_start", ...)`).

Fix: add one test in `compose.test.ts` — compose a stub component that captures `ctx.idleCoordinator`, enqueue an injection, `await pi.dispatch("session_shutdown", { type: "session_shutdown", reason: "reload" }, {})`, then drive the deferred flush and assert nothing was sent (`pi.messages`/`sendMessage` spy empty) and `coordinator.disposed === true`.

**3. Two of the three new tests cannot fail for the behavior their names claim. (blocking)**

Mutation-proved. Starting from this PR's head I removed `this.#disposed ||` from `idle-injection-coordinator.ts:128` and both deferred-interior guards (`:92`, `:105`): `bun test packages/omo-senpi/src/extension/idle-injection-coordinator.test.ts` → **15 pass / 0 fail**. Only when I additionally removed the head guards (`:88`, `:101`) did exactly one test fail — the one at `idle-injection-coordinator.test.ts:231`. Root cause: `dispose()` already clears `#pending` at `:75`, so `#flush` returns 0 through the pre-existing empty-queue branch whether or not the `#disposed` guard exists. Therefore:

- `idle-injection-coordinator.test.ts:210-229` ("a deferred flush fires → harmless no-op (no stale-generation crash)") passes with all three flush-path guards deleted. It also never reproduces the crash: the injected `deliver` at `:215` pushes to an array and cannot throw, so "not a throw" is asserted against a callback that could never throw.
- `idle-injection-coordinator.test.ts:249-256` ("flushOnIdle after dispose returns 0") is unfalsifiable for the same reason — an empty queue returns 0 anyway.

Fix: (a) make the crash test model the crash — construct the coordinator with `() => { throw new Error("stale extension generation after reload") }`, capture the deferred flush via the injectable scheduler, `dispose()`, then `expect(() => flush()).not.toThrow()`; that test *does* fail without the guard at `:92`/`:128`. (b) Decide which mechanism owns post-dispose safety and keep only that one: either dispose stops clearing the queue and the `#disposed` guards carry the behavior (reachable, testable, and compatible with finding 1's drain), or dispose keeps clearing and the now-redundant guard at `:128` goes away. Shipping both means the guard's contract is unpinned forever.

**4. The branch does not merge and was validated against a base 795 commits old. (blocking, mechanical)**

`git merge-tree --write-tree origin/dev cc8aeaa` → `CONFLICT (content): Merge conflict in packages/omo-senpi/src/extension/idle-injection-coordinator.ts`; `git rev-list --count cc8aeaa..origin/dev` → 795. The conflict is exactly at the line this PR edits: `dev` now has a second early return in `#flush` (`if ([...this.#pending.values()].every((injection) => injection.passive === true)) return 0`) directly below `if (this.#pending.size === 0) return 0`, which this PR rewrites at `:128`. So the resolution is semantic, not textual — the rebased `#flush` must keep the disposed guard *and* the passive-only guard — and the green CI on this branch says nothing about the tree that will actually land. Please rebase onto `dev` and re-run.

**5. `dispose()` cannot cancel the timer it exists to neutralize. (minor, real)**

`FlushScheduler` returns `void` (`idle-injection-coordinator.ts:33`) and production passes `(flush) => void setTimeout(flush, 200)` (`compose.ts:101`) with no `unref` and no handle, so `dispose()` can only mask the callback, never cancel it: after a `quit` shutdown the process holds a live 200 ms timer, and the fix depends on flag state surviving in a closure rather than on the timer being gone. Every sibling scheduler in this codebase does it the other way — `components/task/lead-poller-lifecycle.ts:239-243` and `packages/senpi-task/src/completion/notifier.ts:213-217` both `unref()` and return a cancel function.

Fix: change `FlushScheduler` to `(flush: () => void) => (() => void) | void`, store the returned canceller, call it in `dispose()`, and add `timer.unref?.()` in `compose.ts:101`.

**6. Docs and one comment now describe the old contract. (minor)**

`packages/omo-senpi/src/extension/AGENTS.md:10` (compose activation sequence) and `:14` (coordinator contract) are the canonical description of this module and neither mentions the new dispose-on-`session_shutdown` lifecycle — in this repo that file is what the next reader/agent trusts. Separately, inserting the getter at `idle-injection-coordinator.ts:119-121` put it *between* the comment `// Flush the whole queue as one idle-edge steer. Returns how many queued items were collapsed (0 = no-op).` (`:118`) and the `flushOnIdle()` it documents (`:123`), so the comment now reads as documentation for `disposed`. Move the getter above the comment (or the comment down onto `flushOnIdle`), and add one line to `AGENTS.md` stating that the coordinator is disposed on session shutdown and what happens to entries still queued (which depends on how you resolve finding 1).

code-yeongyu:
All four blocking findings are fixed. Branch head: `32f9d4fbd98281b257074669f09110c074d9bf6e` (rebased onto `origin/dev` @ `75c45a800`, so the two commits below are the whole delta against current dev).

| commit | contents |
|---|---|
| `124b193c7` | `fix(omo-senpi): hand back dropped idle injections instead of losing completions` — findings 1, 2, 3, 4 (source, tests, docs) |
| `32f9d4fbd` | `chore(omo-senpi): rebuild senpi plugin bundles on linux` — ubuntu-only bundle-freshness gate |

---

## Finding 1 (semantic): a dropped or post-retirement enqueue is never reported as delivered

You were right, and the loss was worse than the crash. The fix is a receipt contract on the coordinator plus an actual undo path in senpi-task, not a quieter drop.

**Contract now (`idle-injection-coordinator.ts`)**
- `enqueue` returns accept/refuse (`:104`). An **ACCEPTED** injection always gets exactly one receipt: `onFlushed`, or `onDeliveryFailed` on a failed flush **and on retirement**. A **REFUSED** one gets none, because ownership never transferred — the producer sees `false` synchronously and owns the failure. Nothing is dropped silently.
- `retire()` (`:155`) cancels the armed batch-window handle (`:159`), then hands every queued entry `onDeliveryFailed(new IdleInjectionRetiredError())` (`:162`) before clearing.

**Undo path (`packages/senpi-task/src/completion/notifier.ts`)**
- New `CompletionNotifier.recordDeliveryFailure({taskIds, error})` (`:135`): rolls `notified_epoch` back below the epoch (`unpersistNotified`, `:301`), stamps `notification_failed_epoch` + a `notification_failed` event, and re-enters the existing retry ladder. Without the rollback, `recordFailure` alone is a no-op (its mutate is gated on `notified_epoch < epoch`) and `reconcileUnnotifiedNotifications` still skips forever — that is exactly what the rollback exists for. Stale receipts are ignored: a record that already started a new run, a non-notifying terminal (cancel/interrupt), and a record already deleted by TTL cleanup.
- Wired in `engine.ts:165`, so the receipt reaches the store through the real composition, not just in tests.

**Producers act on the refusal**
- `parent-notifier.ts`: attaches the receipt (`:42`) and throws `IdleInjectionRetiredError` on a refusal (`:47`) — the engine's documented synchronous failure signal, so `tryEnqueue` → `recordFailure` + `scheduleRetry` runs and `notified_epoch` is never persisted.
- `member-liveness.ts:105`: a refusal runs `failDelivery` → retry (your point that the `try/catch` could no longer fire).
- `lead-poller-lifecycle.ts:220`: the sink throws so the lead poller releases its durable delivery reservation and redelivers after the reload.
- ulw-loop / ulw-execute-continuation: log the refusal (derived state, re-derived on the next edge) instead of vanishing.

**On your "which mechanism owns post-dispose safety" question (finding 3b):** the single mechanism is now *retirement empties the queue and `enqueue` refuses to refill it*. Because a retired coordinator's queue can never be non-empty, the retirement guards on the flush path were unpinnable-by-construction, so they are **gone**: `#flush` keeps only `size === 0` and dev's passive-only early return, and the `try { … } catch { if (!this.#retired) throw }` swallows in the two deferred interiors are removed. Also removed: `flushSoon`'s retirement guard (no observable effect) and the `disposed` getter (no production caller; the tests assert behavior instead). Everything that remains is pinned below.

---

## Finding 2 + 3a (verification): every assertion is mutation-proven

The three tests you called unfalsifiable are deleted (they asserted against `dispose()`, which dev's `retire()` superseded). The crash test now models the crash: the injected delivery **throws** `stale extension generation after reload`, so a post-retirement flush would really throw.

Method: remove exactly one production line/hunk, run the tests that claim to pin it, restore, re-run. Baseline and restored baseline are **53 pass / 0 fail** across the six files. Every mutation fails at least one test:

| # | production line removed | tests | result |
|---|---|---|---|
| M1 | `retire()` drain receipts (`coordinator:162`) | coordinator + compose + parent-notifier + e2e | 34 pass / **5 fail** |
| M2 | `retire()` `#pending.clear()` (`:161`) | coordinator | 18 pass / **3 fail** |
| M3 | `retire()` `cancelScheduledFlush?.()` (`:159`) | coordinator | 20 pass / **1 fail** |
| M4 | `enqueue` refusal `if (this.#retired) return false` (`:105`) | coordinator + compose + parent-notifier + e2e | 34 pass / **5 fail** |
| M5 | `scheduleFlush` retirement guard (`:114`) | coordinator | 20 pass / **1 fail** |
| M6 | `compose.ts:127` `pi.on("session_shutdown", …retire())` | compose | 6 pass / **2 fail** |
| M7 | `parent-notifier:47` throw on refusal | parent-notifier + e2e | 8 pass / **2 fail** |
| M8 | `parent-notifier:42` `onDeliveryFailed` wiring | parent-notifier + e2e | 7 pass / **3 fail** |
| M9 | `notifier:142` `unpersistNotified(...)` | senpi-task failure receipts + e2e | 4 pass / **3 fail** |
| M10 | `notifier:143` `recordFailure(...)` | senpi-task failure receipts | 4 pass / **1 fail** |
| M11 | `notifier:144` `scheduleRetry(...)` | senpi-task failure receipts | 4 pass / **1 fail** |
| M12 | `notifier:139-140` terminal/notifying guards | senpi-task failure receipts | 3 pass / **2 fail** |
| M13 | `engine.ts:165` receipt → `recordDeliveryFailure` | e2e reload recovery | 1 pass / **1 fail** |
| M14 | `member-liveness:105-108` refusal → `failDelivery` | member-liveness | 8 pass / **1 fail** |

Representative failures (full log kept from the run):

```
### RUN M1  (retire() no longer hands the queue back)
(fail) IdleInjectionCoordinator > … #when the coordinator retires #then every accepted entry is handed back as a delivery failure
(fail) createParentNotifier … #when the coordinator retires before the flush #then the engine is told delivery failed
(fail) background completion survival across a reload > … #then the rebuilt session redelivers the completion
(fail) composeOmoSenpiExtension > … #then its producer gets a failure receipt and later enqueues are refused
### RUN M2  (retire() no longer clears the queue -> the real #7932 crash)
(fail) IdleInjectionCoordinator > #given a deferred flush pending #when the coordinator is retired … #then the flush no-ops and the producer gets a failure receipt
### RUN M9  (notified_epoch is no longer rolled back)
(fail) createCompletionNotifier - async delivery failure receipts > … #then the epoch rolls back and the next session_start reconcile redelivers
(fail) background completion survival across a reload > … #then the rebuilt session redelivers the completion
### RUN M13 (engine no longer wires the receipt)
(fail) background completion survival across a reload > #given a child completing inside the batch window #when the reload retires the coordinator #then the rebuilt session redelivers the completion
```

New/changed coverage:
- `idle-injection-coordinator.test.ts`: drain receipts, refusal-with-no-receipt, canceller invoked on retirement, throwing-delivery no-op. The pre-existing "retired before the deferred pass" test now asserts the receipt (`failures` was previously asserted **empty**; that assertion was the bug).
- `compose.test.ts`: a completion queued in the real composed coordinator, `pi.dispatch("session_shutdown", {reason:"reload"})`, then the receipt fires and later enqueues are refused (pins the compose hook end-to-end; M6).
- `parent-notifier.test.ts`: refusal throws; accepted-then-retired reports task ids to the engine callback.
- `notifier-delivery-failure.test.ts` (new, senpi-task): rollback → reconcile redelivers; retry ladder re-entered (fenced on `getCurrentSessionId`); three stale-receipt cases.
- `completion-reload-recovery.test.ts` (new, omo-senpi): the whole loss window through the **real** `composeTaskEngine` + real record store — child completes inside the batch window, `notified_epoch` is persisted, retirement drops the window, epoch rolls back, and a freshly composed generation's `session_start` reconcile delivers `child finished its work`.
- `member-liveness.test.ts`: refusal fails the delivery and arms a retry.

---

## Finding 4 (mechanical): rebased, conflict resolved preserving both intents

Rebased twice (dev moved during the work): now onto `origin/dev` @ `75c45a800`; `gh pr view` reports `MERGEABLE`. The conflict was exactly where you said, and dev's own fix for #7932 had already landed there as `retire()` (`53ba00fef`), so the resolution is semantic:

- dev's passive-only early return in `#flush` — **kept** verbatim, plus its `size === 0` branch.
- dev's retirement intent — **kept and strengthened** (receipts + refusal + timer cancellation) instead of re-adding this PR's `dispose()`; the duplicate `pi.on("session_shutdown", …dispose())` hook that the textual merge produced in `compose.ts` is gone, and the flush-path guards that both versions carried are dropped as unpinnable (see finding 1 above).

---

## Finding 5/6 (minor)

- `compose.ts:114-117`: the batch window is now `const timer = setTimeout(flush, IDLE_FLUSH_BATCH_WINDOW_MS); timer.unref?.(); return () => clearTimeout(timer)`, and `FlushScheduler` is `(flush) => (() => void) | void` so `retire()` cancels the armed handle (M3). Consistent with `lead-poller-lifecycle`'s `scheduleInterval` and senpi-task's `defaultSchedule`. Test doubles that returned `Array.prototype.push`'s number were updated to return void.
- The orphaned comment is resolved by removing the `disposed` getter entirely, so `// Flush the whole queue as one idle-edge steer …` sits directly above `flushOnIdle()` again.
- Docs: `packages/omo-senpi/src/extension/AGENTS.md` (compose activation sequence now states the `session_shutdown` retirement; the coordinator row states the receipt contract, what retirement does to queued entries, and which producer recovers how), `packages/omo-senpi/src/components/task/AGENTS.md` (parent-notifier row), `packages/senpi-task/AGENTS.md` (completion recovery: "enqueue returned" ≠ delivered, and what `recordDeliveryFailure` does). Both `changes.md` files carry an entry.

---

## Test results (linux/x64, bun 1.4.2 — the CI pin; nothing was run on macOS)

```
$ bun test --timeout 20000 packages/omo-senpi
 3174 pass, 3 skip, 0 fail, 11189 expect() calls (408 files)

$ bun test --timeout 20000 packages/senpi-task
 2015 pass, 1 skip, 0 fail, 6661 expect() calls (271 files)

$ bun run typecheck            # root + script + all 30 package projects
 clean

$ node packages/omo-senpi/plugin/scripts/build-extension.mjs --check
 omo-senpi extension build is current
```

The committed bundles (`omo.js`, `omo-task.js`) are the artifacts of that Linux build, pulled back byte-identical (sha256 `d3d0ea42…` / `cc4a8e0d…`), which is what keeps the ubuntu-only freshness step green.

**Two assumptions worth flagging.** (1) `recordDeliveryFailure` resolves the epoch from the record's current `run_epoch` rather than taking one, because `CompletionDetails` carries no epoch; a receipt for a record that already re-ran is therefore ignored (M12 pins that). (2) DAG-run wakes and kibitzer nudges still ignore the refusal boolean by design — dag-wake already buffers while the parent state is `session_shutdown`, and kibitzer nudges are re-surfaced from their pending file — so only the durable producers were changed.

PR #7935, opened 2026-09-07
fix(senpi-task): preserve thinking level in live progress after model fallback

### LINKED ISSUE #7934: [Bug]: OMO/Senpi task status loses thinking level after runtime fallback
## Bug Description

OMO/Senpi's task status loses the thinking-level suffix after runtime model fallback. The fallback controller supplies a thinking level to the session, but its fallback event omits that information. The task widget then discards the resolved-model metadata that previously supplied the suffix and renders the bare fallback model string.

This establishes a display information-loss bug in the inspected release. It does **not** establish that thinking was disabled, nor that the configured fallback reasoning level was the level actually used by the original child.

## Prerequisites

- [x] Written in English.
- [x] Searched existing issues for fallback/thinking-level reports. Related #6295 is closed and concerns thinking-level application at child creation; this report concerns loss of display metadata on a runtime fallback event. No exact duplicate was identified in the searches.
- [x] Read the bug-report template and CONTRIBUTING.md, and investigated with an AI coding agent using this repository's URL.
- [x] Installed OMO matches the current **beta** npm dist-tag: 5.0.0-0.beta.48. The npm latest tag is 0.0.0-beta.0; this report explicitly targets beta rather than claiming latest-channel reproduction.

Guidelines consulted: [.github/ISSUE_TEMPLATE/bug_report.yml](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/.github/ISSUE_TEMPLATE/bug_report.yml) and [CONTRIBUTING.md](https://github.com/code-yeongyu/oh-my-openagent/blob/dev/CONTRIBUTING.md).

## Environment

- Operating system: macOS, arm64.
- OMO CLI: omo 5.0.0-0.beta.48 (engine: senpi 2026.9.7-2)
- OpenCode version: 1.18.29. The affected task UI is the **OMO/Senpi** runtime, not the OpenCode plugin UI.
- Evidence: installed package inspection and existing logs. No fresh provider-error reproduction, source changes, or regression tests were run for this report.

## Steps to Reproduce

The following describes the reported path and a maintainer reproduction scenario, not a newly executed reproduction:

1. Configure a category with a primary model and a fallback model, with reasoning levels as in the sanitized configuration below. Replace placeholders with valid available models.
2. Start a task in that category and observe the model/thinking suffix in its status.
3. When the child undergoes runtime fallback, for example following a retryable provider error, inspect its status again.
4. The model changes and the fallback counter increases, but the thinking suffix disappears.

A deterministic maintainer test can trigger the controller through injected dependencies rather than deliberately exhausting provider quota.

## Expected Behavior

Show the **effective applied** thinking level after every model transition, including clamping, explicit off, and fallback restoration. Do not infer it from the original category metadata.

If the effective level is high, the status should contain:

```text
category:deep(PROVIDER/FALLBACK_MODEL:high) · fallback:N · turn T (K tools)
```

## Actual Behavior

Sanitized observed status, with model identifiers and activity counts replaced:

```text
category:deep(PROVIDER/FALLBACK_MODEL) · fallback:N · turn T (K tools)
```

The fallback counter counts fallback events, not a model's position in the configured chain. The original child's complete event history was unavailable, so the individual transitions behind its counter cannot be reconstructed.

## Configuration

Relevant category configuration, with model selectors replaced consistently:

```json
{
  "[senpi]": {
    "categories": {
      "deep": {
        "models": [
          { "model": "PROVIDER/PRIMARY_MODEL", "reasoning": "xhigh" },
          { "model": "PROVIDER/FALLBACK_MODEL", "reasoning": "high" }
        ]
      }
    }
  }
}
```

Category fallback metadata alone is insufficient to prove the effective post-fallback level. The runtime controller uses its resolved fallback selector's thinking level, otherwise inherits the current level, then clamps to model support. The exact selector governing the original child was not recovered.

## Root cause analysis

References below are relative to the **installed published packages**, not claimed line numbers in the current dev branch. OMO's task extension is minified, so function names are given instead of invented source line numbers.

### 1. Runtime fallback computes thinking but emits only a bare model selector

Senpi `dist/core/retry-fallback/controller.js:131-149`:

```js
const thinking = this.selectThinking(candidate.selector, candidate.model, current.thinkingLevel);
await this.deps.switchModel(candidate.model, thinking, "fallback");
// ...
const to = formatSelector(candidate.model);
// State stores lastAppliedThinkingLevel: thinking.
this.deps.logger.info("fallback_applied", { from, to, chainKey: candidate.chainKey, reason });
this.deps.emit({ type: "retry_fallback_applied", from, to, chainKey: candidate.chainKey, reason });
```

`dist/core/retry-fallback/chains.js:78-81` appends a thinking suffix only when formatSelector receives a thinking argument. This call does not pass one. Neither the emitted event nor the log contains the applied level.

### 2. The task widget clears its only thinking metadata on this event

OMO `plugin/extensions/omo-task.js`, status-tracker function `$l`, excerpt with whitespace normalized:

```js
if ("retry_fallback_applied" === e.type && void 0 !== e.to)
  return d = void 0, l = e.to, u += 1, !0;
```

Here d is the resolved-model metadata, l is the displayed model string, and u is the fallback count. The ll/ul render helpers obtain the suffix from resolvedModel.reasoning, reasoning_effort, or variant. Once d is cleared, the renderer falls back to l, which is the suffix-free event.to.

**Causal chain:** fallback event omits thinking → tracker clears resolved-model metadata → bare event.to becomes the display model → thinking suffix disappears.

### 3. The session application path is separate from the broken display path

Senpi `dist/core/agent-session.js:3781-3829` applies the selected thinking level during _switchActiveModel and emits an authoritative model_changed event:

```js
if (opts.ephemeralThinkingLevel !== undefined) {
  this._applyEphemeralThinkingLevel(thinking.level);
} else {
  this._setThinkingLevel(thinking.level, false, thinking.selection);
}
// After admission hooks accept the candidate:
this._emit({
  type: "model_changed",
  model,
  thinkingLevel: this.thinkingLevel,
  source: opts.modelSelectSource,
});
```

Controller selectThinking at lines 204-210 uses selector.thinkingLevel ?? inherited ?? "off", then clampThinkingLevel. Consequently, the display must use effective session state, not assume that the category's high value was applied.

### 4. Trigger versus root cause

The existing logs show an HTTP 429 rate_limit_exceeded, followed by cooldown and a transient fallback. This explains a recorded fallback trigger, not the display bug itself. The metadata-loss path applies whenever the tracker receives retry_fallback_applied, regardless of why fallback happened.

The exact original child was not recovered. These nearby log entries cannot conclusively be assigned to that child. Its effective pre/post-fallback level and complete transition history remain unknown.

## Error Logs / Sanitized OMO debug evidence

Relevant excerpts from session.log and fallback.log, in recorded order. Only timestamps and model selectors are replaced. The error string's trailing ellipsis was already present in the stored log. No account names, session/task IDs, workspace paths, credentials, prompts, or full transcripts are included.

```jsonl
{"ts":"<REDACTED_TIMESTAMP>","level":"warn","event":"provider_error","kind":"error","error":"OpenAI API error (429): {\"param\":null,\"code\":\"rate_limit_exceeded\",\"type\":\"rate_limit_error\",\"message\":\"Error from provider (Console): Upstream request failed: [rate_limit_exceeded] Rate limit exce..."}
{"ts":"<REDACTED_TIMESTAMP>","level":"info","event":"cooldown_noted","selector":"PROVIDER/PRIMARY_MODEL","errorMessage":"OpenAI API error (429): {\"param\":null,\"code\":\"rate_limit_exceeded\",\"type\":\"rate_limit_error\",\"message\":\"Error from provider (Console): Upstream request failed: [rate_limit_exceeded] Rate limit exce..."}
{"ts":"<REDACTED_TIMESTAMP>","level":"info","event":"fallback_applied","from":"PROVIDER/PRIMARY_MODEL","to":"PROVIDER/FALLBACK_MODEL","chainKey":"PROVIDER/PRIMARY_MODEL","reason":"transient"}
```

The fallback_applied record lacks thinkingLevel, matching the emitter code. No additional relevant fallback/thinking entries were found in the inspected senpi-debug.log. The issue uses the operative fallback log rather than attaching unrelated debug output. There was no relevant crash stack trace to include.

## Doctor Output

Required template command `bunx oh-my-openagent doctor` completed with exit code 0. Full diagnostic output below, with the home directory redacted and model names replaced. This doctor checks OpenCode-oriented configuration and is supplemental to the OMO/Senpi source evidence.

```text
 oMoMoMoMo Doctor

 ⚠ 2 issues found:

1. Legacy OMO configuration remains
   Legacy configuration remains at <HOME>/.omo/config.jsonc. It is not part of the unified OMO config chain.
   Fix: Run `oh-my-openagent config migrate` to move it into ~/.omo/omo.jsonc.
   Affects: configuration migration

2. Configured models rely on compatibility fallback
   hephaestus=PROVIDER/PRIMARY_MODEL (unknown), multimodal-looker=PROVIDER/PRIMARY_MODEL (unknown), deep=PROVIDER/PRIMARY_MODEL (unknown)
   Affects: hephaestus, multimodal-looker, deep
```

These warnings are reported as environment findings, not established causes of the thinking-suffix loss. No migration or configuration changes were performed.

## Suggested correction and validation

1. Carry the effective applied thinking level through the task event/state contract. Use authoritative model_changed.thinkingLevel or an equivalent verified value on retry_fallback_applied. Include it in fallback diagnostics.
2. Update the task widget to retain the effective model/level after fallback, subsequent hops, and restoration. Do not retain a stale primary-model suffix or manufacture high from configuration.
3. Audit task-list and completion renderers for the same state-loss issue; their behavior has not been independently reproduced here.
4. Add deterministic coverage for explicit fallback thinking, inherited thinking, clamped thinking, off, multiple hops, and restoration. Assert the displayed level equals the applied session level and the fallback counter remains correct.
5. Exercise the real Senpi task UI under the project's senpi-qa workflow after implementing the fix, as required by CONTRIBUTING.md. Unit tests alone would not establish the visible behavior.

No implementation is included in this issue. The code-level cause is verified; original-child runtime details that cannot be recovered are explicitly left unknown.

code-yeongyu:
## Summary

The child progress tracker cleared `resolvedModel` on `retry_fallback_applied` events, falling back to the raw `event.to` string. When Senpi core emitted the selector without a `:thinking` suffix, the status line lost the applied reasoning level.

**Before:** `category:deep(openai-codex/gpt-5.6-sol) · fallback:2 · turn 14 (90 tools)`
**After:** `category:deep(openai-codex/gpt-5.6-sol:high) · fallback:2 · turn 14 (90 tools)`

## Changes

**`progress.ts`**
- Parse `event.to` into a proper `ResolvedModelRecord` via `parseModelSelector()` (mirrors the store's `runtime-fallback-event.ts` logic) so the live progress tracker retains provider/model_id structure and any selector-encoded thinking level
- Read an explicit `event.thinking` field when present, so the display reflects the actually-applied level even when the selector was normalized without the suffix
- Add `readEventString` helper for safe event field access

**Tests (2 new)**
- Fallback with bare selector + explicit `thinking` field → display includes applied level
- Fallback with `:thinking` in selector, no explicit field → selector suffix preserved

Note: Full fix requires Senpi core to include the applied thinking level in the `retry_fallback_applied` event payload. This PR handles the consumer side so progress is correct when the level is available via selector suffix or a future `thinking` event field.

Closes #7934

---
## Summary by cubic
Fixes the live progress status line losing the applied thinking level after a model fallback. Previously, `retry_fallback_applied` events replaced the resolved model with the raw selector string, so when Senpi core emitted the selector without a `:thinking` suffix the display dropped the reasoning level (e.g. `category:deep(openai-codex/gpt-5.6-sol)` instead of `category:deep(openai-codex/gpt-5.6-sol:high)`). Now the tracker parses `event.to` into a proper model record and prefers an explicit `thinking` event field when present.

Rebuilds the `omo-task.js` extension bundle to include the fix. Full fix requires Senpi core to include the applied thinking level in the `retry_fallback_applied` event payload; this PR handles the consumer side so progress is correct when the level is available via the selector suffix or the future `thinking` field. Closes #7934.

<sup>Written for commit fa6c935af309662993ce610c8f01ef90936fde24. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7935?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/senpi-task/src/progress.ts
```diff
@@ -69,8 +69,22 @@ export function createChildProgress(
...
       if (event.type === "retry_fallback_applied" && event.to !== undefined) {
-        resolvedModel = undefined
-        model = event.to
+        // Build a resolved model from the selector so the live status line retains the
+        // provider/model_id structure and any thinking level encoded in the selector suffix.
+        // When the Senpi core event carries an explicit thinking field, prefer that over the
+        // selector suffix so the display reflects the actually-applied level even when the
+        // selector was normalized without it.
+        const parsed = parseModelSelector(event.to, resolvedModel?.source ?? "category")
+        const eventThinking = readEventString(event, "thinking")
```
aslaii, line 78: P1: The production fallback path still loses the thinking level, so this does not yet close #7934.

At this commit, the handler only gets a level from a suffixed selector or the new hypothetical `event.thinking` field. The inspected Senpi 2026.9.7-2 controller emits neither: `retry_fallback_applied` contains a bare `to` plus `from, chainKey, reason`. Its preceding `model_changed` event carries the effective `thinkingLevel`, but this tracker ignores it.

I imported this PR's actual `createChildProgress` implementation with the real pi-tui renderer and replayed that runtime-shaped sequence, without model API calls:

```js
progress.accept({
  type: "model_changed",
  model: { provider: "provider", id: "fallback" },
  thinkingLevel: "high",
  source: "fallback",
});
progress.accept({
  type: "retry_fallback_applied",
  from: "provider/primary",
  to: "provider/fallback",
  chainKey: "provider/primary",
  reason: "transient",
});
```

Observed:
```text
before: category:deep(provider/primary:xhigh)
after:  category:deep(provider/fallback) · fallback:1
```

Please wire the authoritative applied level end to end, either by consuming `model_changed.thinkingLevel` and retaining it across the subsequent fallback event, or by landing and consuming the corresponding engine event-contract update. Add a regression using the actual emitted sequence; the new `thinking: "high" as never` fixture bypasses the event type and does not reproduce the reported failure. If this intentionally lands as preparatory consumer support only, remove `Closes #7934` and keep the issue open until the producer/UI integration is verified.

Verification limits: this was a direct module/event replay, not a live provider-driven TUI run. The targeted `bun test packages/senpi-task/src/progress.test.ts` attempt failed during repository preload because `@oh-my-opencode/rules-engine` was unavailable in the isolated checkout, before any test executed. GitHub's completed checks do not include a build/test gate. No source/configuration changes were made.

PeterPonyu:
While triaging #7934 against `dev@b5122f19d` I checked this change against the Senpi version dev pins (`@code-yeongyu/senpi` 2026.9.9) and I think it does not restore the level in production yet, for two reasons that are both on the emitter side:

1. `dist/core/retry-fallback/controller.js:131-135` computes the clamped level but formats the selector without it:
   ```js
   const thinking = this.selectThinking(candidate.selector, candidate.model, current.thinkingLevel);
   await this.deps.switchModel(candidate.model, thinking, "fallback");
   ...
   const to = formatSelector(candidate.model);   // bare provider/model_id, no ":level"
   ```
   so `parseModelSelector(event.to)` never sees a suffix (`formatSelector(model, level)` is only used for the chain lookup at line 167).
2. Line 149 emits `{ type: "retry_fallback_applied", from, to, chainKey, reason }` — there is no `thinking` field, so `readEventString(event, "thinking")` is always `undefined`.

The new tests pass because they construct events with a suffixed `to` / a `thinking` field that the pinned core never produces (same for the pre-existing fixtures in `progress.test.ts:115` and `manager-runtime-fallback.test.ts`).

What the stream does carry is the `model_changed` event that `switchModel(..., thinking, "fallback")` emits immediately before `retry_fallback_applied`, with `thinkingLevel` set to the *effective* (post-`clampThinkingLevel`) level and `source: "fallback"`. Remembering the last `model_changed` in the tracker and, on `retry_fallback_applied`, applying its `thinkingLevel` as `reasoning` when its model matches `event.to` (falling back to a `:suffix` if a future core adds one) restores the level without any Senpi change, and the same fold in `runtime-fallback-event.ts` would keep the persisted `resolved_model` in step so the task-list and completion rows agree with the status line. Happy to send that as a follow-up on top of this branch if useful.

Issue #7946, opened 2026-09-08
memory reflection launch fails once per RPC-host restart: stale extension ctx under --multi-session


code-yeongyu:
## Symptom
On mengmotaHost, running the packaged omo runtime under omo-desktop (`--mode rpc --multi-session` in the desktop's RPC host), `~/.omo/agent/rpc-desktop/daemon/stderr.log` records exactly one `memory reflection launch failed` per **web-service restart** (observed at 12:47:27 and 13:08:39 KST on 2026-09-08, runtime omo 31721f19c / 7c78c1464, senpi 4adba7afb):

```
memory reflection launch failed {
  error: "This extension ctx is stale after session replacement or reload. Do not use a captured pi or command ctx after ctx.newSession(), ctx.fork(), ctx.switchSession(), or ctx.reload(). ..."
}
```

The thrower is senpi's own stale-context guard (`packages/coding-agent/src/core/extensions/runner.ts:825`, guarded by `core/extensions/builtin/goal/stale-context.ts`). The omo-side log site is `components/memory/trigger-wiring.ts:94` (`launch()` catch) or `identity-runtime.ts:127` (runner.launch rejection).

## Trigger shape
The RPC host boots fresh (coldStart=transient), the first desktop thread session opens and its memory component binds (`session_start` -> `wiring.afterBind` -> `runtimeFor(identity).reconcile()` -> reserved/pending reflection run -> `runner.launch()`), and the in-process reflection/judge child opens a senpi session through a `pi`/extension ctx that is already stale under `--multi-session` (sessions are created/replaced inside one long-lived host). A fresh thread whose turn ends normally does NOT reproduce it, so it needs the pre-existing reservation/boot path.

## Impact
One reflection run is dropped per restart (a turn goes without its memorian reflection). The component recovers afterwards.

## Likely fix shape
Same stale-ctx discipline senpi itself applies to goal tickers (`goal/stale-context.ts`): the memory reflection runner must not capture a `pi`/`eventCtx` across session replacement; re-resolve the live session ctx per launch (or launch through `withSession`), so the reservation survives a host restart instead of throwing on the stale one.

code-yeongyu:
[...]

PR #7955, opened 2026-09-08
fix(omo-senpi): treat a bind superseded by session replacement as a skip

code-yeongyu:
## Summary
`memory bind-time reconcile failed` fires on omo-desktop whenever a session is replaced while its bind-time reconcile is still in flight:

```
memory bind-time reconcile failed {
  error: "Error: This extension ctx is stale after session replacement or reload. ..."
}
```

`index.ts` starts `wiring.afterBind(...)` as a floating promise on `session_start` (by design), and `afterBind` then awaits `reconcile()`, `reconcileSession(eventCtx)` and `reflectionLive.bind(pi, ...)`. If the host replaces the session in that window it retires the ctx, every one of those calls throws, and the operator sees a warn for something that is ordinary lifecycle — the replacement session runs its own bind and nothing is lost.

This is the same category `logBindReconcileFailure` already handles for reflection-lock contention, so it gets the same treatment: an `info` skip with `reason: "session replaced before the bind completed"`. Unexpected errors still warn.

Downstream symptom of the same stale ctx: `memory reflection launch failed` (`identity-runtime.ts`), which appears instead whenever a reflection reservation happens to be pending at bind time.

## Evidence
- Live on mengmotaHost (packaged runtime, `--mode rpc --multi-session`), 4 desktop restarts: 12:47 / 13:08 / 13:45 each resumed sessions and each logged it; the 15:41 boot with **0** resumed sessions was clean until the next session bound, then logged the bind-time variant. Tracked in oh-my-openagent#7946.
- RED: the new case failed with `"memory bind-time reconcile failed"` / `warn` instead of the skip.
- GREEN: `bind-reconcile-log.test.ts` 3/3; whole memory component **1265 pass / 0 fail**; `tsgo --noEmit` clean.

## Scope note
This makes the log honest about an expected lifecycle race. It deliberately does NOT change `afterBind`'s ctx handling — re-resolving the live ctx per use (or running under `withSession`) so a superseded bind can still complete its work is the larger follow-up tracked in #7946.

---
## Summary by cubic
Treats a bind superseded by session replacement as a recoverable skip instead of a failure, so `memory bind-time reconcile failed` and `memory reflection launch failed` no longer warn for ordinary lifecycle races. Unexpected errors still warn.

- Classifies the stale extension ctx error as an `info` skip with `reason: "session replaced before the bind completed"`, matching how reflection-lock contention is already handled.
- Adds a test covering the superseded-bind path and documents the behavior in `changes.md`.
- Does not change `afterBind`'s ctx handling; re-resolving the live ctx per use remains a follow-up.

<sup>Written for commit 2922b794de555f9e7e25ba075635143bf69e2215. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
**Status correction — converting this to draft, do not merge as written.**

I opened this to reclassify the superseded bind as a recoverable `info` skip, on the belief (stated in the body) that it fires once per restart. That belief was wrong, and I measured it properly on the live desktop afterwards:

**6+ occurrences within a SINGLE host lifetime** (host up 16:26 KST, one `rpc host ready` marker, 5 sessions in the projection). Details in oh-my-openagent#7946.

So the premise of this PR is inverted. `afterBind` throws at its first ctx touch, which drops the steps after it for that session — including ctx-free work like `factsWiringFor(identity).reconcileExtractor()` and the dream reconcile. Relabeling that as a benign `info` skip would delete the only signal for repeated, real dropped work. That is worse than the current noisy warn.

Leaving the test in place (it documents the classification question) but the fix needs to be the actual repair instead:
- run identity-scoped, ctx-free recovery unconditionally (`reconcile()` already is, ahead of the first ctx read),
- guard each ctx-dependent step (`reconcileSession(eventCtx)`, `reflectionLive.bind(..., eventCtx, ...)`) individually so a retired ctx skips only itself,
- and only then treat a genuinely superseded bind as an expected skip.

I'm not landing that restructure unilaterally — it is a design change inside the memory component's error boundaries and it belongs to whoever owns that lane. Handing the corrected diagnosis over rather than merging a fix I now know is wrong.

PR #7958, opened 2026-09-08
fix(team-core): retry contended Windows renames in team state atomicWrite (#7897)

### LINKED ISSUE #7897: [Bug] Windows: EPERM on rename() during team state.json atomic write - causes Team Mode failures
## Description

On Windows, Team Mode operations (create, update, cleanup) intermittently fail due to `EPERM` errors during the atomic state file write (`rename()` operation). This causes team messages to not be delivered, tasks to not start, and state to never persist to `~/.omo/runtime/`.

## Environment

- **OS**: Windows 11 (26100.x)
- **Node.js**: bundled with oh-my-openagent
- **oh-my-openagent version**: latest
- **Storage path**: `C:\Users\<user>\.omo\runtime\`

## Error

`
EPERM: operation not permitted, rename 'C:\Users\<user>\.omo\runtime\{teamRunId}\state.json.tmp.{uuid}' -> 'C:\Users\<user>\.omo\runtime\{teamRunId}\state.json'
`

**Errno**: `-4048`

## Occurrences

14 occurrences observed across 3 different hooks:

| Hook | Impact |
|------|--------|
| `team member error handler failed` | Member status not saved |
| `[background-agent] Error starting task` | Background tasks fail to start |
| `[team-mailbox-injector] Failed to inject team mailbox messages` | **Team messages not delivered** |

## Root Cause Analysis

The atomic write pattern (`write tmp → rename to target`) fails on Windows because:

1. **Concurrent access**: Multiple plugin hooks (`member-error-handler`, `background-agent`, `team-mailbox-injector`) write to the same `state.json` file simultaneously. The `rename()` syscall on Windows requires exclusive access — if another process/handle holds the tmp file, `rename` throws EPERM (not EBUSY as on POSIX).

2. **Windows Defender real-time scanning**: The newly created `.tmp` file may be briefly locked by Defender's file system filter driver during the write → rename window, causing the rename to fail.

## Impact

- `~/.omo/runtime/` directory remains **empty** — state never persists
- Team messages (`team_mailbox.json`) not delivered to members → members appear stuck
- Fallback to temp directory (`omo-runtime-` in `%TEMP%`) which is lost when process exits
- Lead agent falls back to direct execution instead of team delegation

## Additional: Toast Notification Failures (18 occurrences)

`
[session-notification] notification operation failed
FullyQualifiedErrorId: ArgumentException
`

`CreateToastNotifier('OpenCode')` fails because `'OpenCode'` is not registered as a valid AppUserModelID for Windows toast notifications.

## Suggested Fix

1. **Add retry logic** on EPERM for the `rename()` call (e.g., 3 retries with 50ms delay)
2. **Use `writeFileSync`** directly instead of tmp+rename pattern on Windows (trade atomicity for reliability)
3. **Add file locking** (e.g., `lockfile` or advisory lock) to prevent concurrent writes to the same state file
4. **Gracefully degrade**: If state write fails, log warning but continue team operation in memory-only mode

## Reproduction

`powershell
# 1. Start OpenCode Desktop
# 2. Create a team with 3+ members
# 3. Observe EPERM errors in oh-my-opencode.log
# 4. Check ~/.omo/runtime/ — directory will be empty
`

## Logs

Full log: `C:\Users\<user>\AppData\Local\Temp\oh-my-opencode.log` (5.2MB)

**Platform**: Windows only — POSIX `rename()` does not have this issue.

PeterPonyu:
## Summary

Fixes #7897. On Windows, Team Mode hooks intermittently lose their state write with
`EPERM: operation not permitted, rename '...state.json.tmp.<uuid>' -> '...state.json'` (errno `-4048`).
The reporter saw 14 occurrences across three hooks (`member-error-handler`, `background-agent`,
`team-mailbox-injector`), with `~/.omo/runtime/` left empty and team messages never delivered.

`atomicWrite` in `packages/team-core/src/team-state-store/locks.ts` writes a temp file, fsyncs it,
then renames it over the target. The write and the fsync are fine. The single `rename` attempt is
the problem: on Windows a handle briefly held by another writer or by a Defender scan makes libuv
return `EPERM` (`ERROR_ACCESS_DENIED`) or `EBUSY` (`ERROR_SHARING_VIOLATION`), and that update is
dropped permanently. `transitionRuntimeState` already holds `state.lock` when it calls
`saveRuntimeState`, so the contention is cross-process, not something a wider lock can fix.

## Changes

`packages/team-core/src/team-state-store/locks.ts`:

- New `renameWithContentionRetry`: up to 5 attempts, linear 50 ms backoff (50, 100, 150, 200), so a
  contended rename gets about half a second to clear before it fails.
- Gated on `win32` through the existing injectable-`platform` pattern already used by
  `assertRetryableLockOpenError`. On POSIX, `EPERM` from `rename` means a sticky bit, an immutable
  flag, or a cross-device move, and retrying only adds latency to a certain failure.
- `isRetryableLockReleaseError` renamed to `isTransientFileContentionError` and shared with the
  rename path, since it is now used by both `reapStaleLock` and the rename retry. Same predicate,
  no behavior change for the lock-release path.
- `AtomicWriteDeps` gains `delay` and `platform` so tests drive the retry without real timers.

Retries only re-issue the rename. The temp file is written and fsynced exactly once, so no attempt
can produce a torn `state.json`. The existing cleanup path still removes the temp file when the
retries are exhausted.

I deliberately did not take the issue's suggestion 2 (plain `writeFileSync` on Windows). Trading
atomicity away would let a concurrent reader observe a half-written `state.json`, which is a worse
failure than the one being fixed.

## Test plan

- [x] Four new cases in `locks.test.ts`, each pinning a state the code must distinguish: a transient
      `EPERM` that clears after 2 failures, an `EBUSY` that never clears, a POSIX `EPERM` that must
      not retry, and a Windows `ENOENT` that must not retry. The delay sequence is asserted, so a
      change to the attempt count or the backoff cannot pass silently.
- [x] Negative case: win32 EACCES is not retried (renameCalls 1, no delay) — pins that only EPERM/EBUSY are contention codes.
- [x] RED verified: both retry cases fail against the unpatched `locks.ts` from `dev`
      (12 pass / 2 fail), green with the fix (14 pass / 0 fail).
- [x] `bun test packages/team-core`: 163 pass / 1 skip / 0 fail.
- [x] `tsgo --noEmit` clean for `team-core`, `senpi-task`, and `omo-opencode`.
- [x] `bun test script/ci-fast-path.test.ts script/ci-fast-path.full-matrix.test.ts`: 36 pass,
      with the new row RED against the unchanged classifier.
- [x] Regenerated `packages/omo-senpi/plugin/extensions/{omo-task,omo-member}.js` with bun 1.4.0 after ubuntu senpi `--check` reported `stale-output` (team-core is in the task bundle).
- [x] Fault injection against the real filesystem (real `open`/`writeFile`/`fsync`/`rm`, only
      `rename` injected), before and after. The concurrent case is the one from the issue: 8 writers
      on one `state.json`, every third rename failing once with the reported errno shape.
      Before: 6 fulfilled / 2 rejected. After: 8 fulfilled / 0 rejected. No run left a
      `state.json.tmp.<uuid>` behind, and the final file parsed as valid JSON every time.

Evidence: `.omo/evidence/20260908-7897-atomic-write-rename-retry/` (README, `unit-gate.txt`,
`fault-injection.ts`, before/after transcripts).

## What this PR does not cover

I do not have a Windows host, so the Windows behavior is driven through the injectable seams using
the exact errno shape from the report rather than a real Defender filter driver. CI has to supply
the real Windows leg.

That leg would not have run by default. `locks.ts` has no `windows` or `win32` token in its
basename, so `classifyCiMode` treated this diff as platform-neutral and kept CI on ubuntu, which
would have shipped a Windows fix unproven on Windows. I cannot apply the `ci:full-matrix` label as
an outside contributor, so the third commit registers
`packages/team-core/src/team-state-store/locks.ts` in `platformSensitiveExactPaths` instead. That
matches the existing `packages/openclaw-core/src/reply-listener-process.ts` entry, which is there
for the same reason: a win32 branch behind a platform-neutral filename. `locks.ts` now has two
(the lock-open EPERM probe and the new rename retry). Verified: the classifier returns
`fullMatrix: true` for this PR's changed paths. Drop that commit and label the PR manually if you
would rather keep the list shorter.

The issue also reports toast-notification failures (`CreateToastNotifier('OpenCode')` /
`ArgumentException`, unregistered AppUserModelID). That is a separate subsystem and is not addressed
here; #7897 should stay open for it, or be split, whichever the maintainer prefers.

---
## Summary by cubic
Fixes #7897 so Team Mode state writes on Windows no longer drop when a rename fails with `EPERM` or `EBUSY` because another process or antivirus briefly holds the file. `atomicWrite` now retries those codes up to five times with a 50 ms linear backoff on Windows; POSIX behavior is unchanged.

- Writes and fsyncs the temp file once, then removes it if all rename attempts fail.
- Adds tests for retry timing, exhaustion, POSIX behavior, and non-contention errors like `EACCES` and `ENOENT`.
- Marks `locks.ts` as platform-sensitive so the Windows CI leg runs automatically.
- Regenerates the `omo-task.js` and `omo-member.js` bundles.
- Toast-notification failures from the issue remain outside this change.

<sup>Written for commit 002dcb3bc0b08577308ff333b898d0534d1f0671. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


PeterPonyu:
Status note on the remaining red check:

- All required checks of the `dev` ruleset are green here, including `test (windows-latest, 1/2)` and `(2/2)`, so the new `locks.test.ts` Windows rename-retry cases compiled and ran on windows-latest. `mergeStateStatus=UNSTABLE` comes only from the non-required `senpi-compatibility (windows-latest)` leg.
- That leg ([job 101995371039](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/34205881075/job/101995371039)) fails on three `live thread socket discovery` tests in `packages/omo-senpi/src/components/thread/live-surface.test.ts` and fails identically on `dev` ([run 34216613636 / job 102029836693](https://github.com/code-yeongyu/oh-my-openagent/actions/runs/34216613636/job/102029836693)). It is a base-branch defect (POSIX path literals in the tests versus `node:path` output on Windows), fixed test-only in #7972.
- The job log has no EPERM/EACCES/rename failures, so it is not evidence for widening the retry predicate.

PR #7986, opened 2026-09-08
fix(keyword-detector): preserve ULW on follow-up messages (fixes #5806)

### LINKED ISSUE #5806: ULW mode is edge-triggered instead of level-triggered - follow-up messages lose delegation

MoerAI:
## Summary

Keep explicit ULW activation for subsequent messages in the same session, rather than requiring the user to repeat the keyword. Use a compact synthetic marker on ordinary follow-ups, restoring full guidance through runtime system transforms after compaction until a real message persists it; model-family changes also refresh guidance. Retain the existing detector guards and clear activation when the session is deleted, the hook is disposed, or continuation is stopped.

## Root Cause

`createKeywordDetectorHook` previously scanned only the current message. Its default-mode cache did not retain keyword-triggered activation. A keyword-free follow-up therefore lost both the injected ULW directive and the later model override.

## Changes

| File | Change |
| --- | --- |
| `packages/omo-opencode/src/hooks/keyword-detector/hook.ts` | Track explicit activation and its prompt source in bounded 256-session state; inject compact follow-up markers without changing user text or repeating activation toasts. |
| `packages/omo-opencode/src/plugin/system-transform.ts` and `plugin-interface.ts` | Keep pending restoration in automatic runtime calls, using the actual runtime model and current agent guards. |
| `packages/omo-opencode/src/plugin/event.ts` | Route deletion and compaction to the keyword detector. |
| `packages/omo-opencode/src/plugin/stop-continuation.ts` | Clear activation through the shared command/tool stop owner. |
| `packages/omo-opencode/src/plugin/chat-message/types.ts` | Preserve the optional cleanup method in the owner contract. |
| `packages/omo-opencode/src/hooks/keyword-detector/ultrawork-followup.test.ts` | Cover retention, model resolution, isolation, guarded messages, lifecycle cleanup and eviction. |

The pure model resolver remains unchanged. The production message handler already invokes keyword injection before model override resolution.

## Reproduction (before fix)

The new follow-up regression returned `active: false` and `override: null` after an explicit ULW message.

```text
3 pass
3 fail
```

The common-stop regression also failed before that owner was wired. The final routing/agent-guard regressions failed at 20 pass / 4 fail, then passed at 24 / 0.

## Verification (after fix)

```text
Final scoped tests with Bun 1.4.0: 204 pass, 0 fail
Stop command/tool entrypoint tests: 28 pass, 0 fail
bun run typecheck: exit 0
bun run build: all steps completed
```

Real OpenCode 1.18.4 loaded the production system-transform registration through a PluginModule. After a deliberately lossy compaction summary (`Done.`), decoded provider request 4 contained the exact runtime-generated ULW guidance. The capture verifies byte equality, not merely that another model request occurred. Pending restoration remains available across runtime calls until a real message durably stores full guidance. Image-only restoration stores a synthetic full-guidance part; later ordinary image turns use a compact marker.

The replay also verified compact markers, stop/deletion cleanup, matching SSE events, and host DB counts of 8083 before and after. Only model responses were scripted; no host configuration or database was mounted.

Earlier CI failures matched dev run 34307575373 (missing build-omob exports and stale Senpi bundle); those unrelated files were not changed. See the PR checks for the current CI result. Local tests, full typecheck/build and real replay passed on Bun 1.4.0 and Node 24.18.0.

Evidence, including replay commands, the exact machine-readable capture, and isolation counts:
`.omo/evidence/20260908-5806-ultrawork-followups/README.md`.

## Scope and limits

The state uses bounded FIFO eviction, matching the existing 256-session cache policy. This verifies the hook and model-selection contract, not probabilistic model compliance. No prompt prose, global model policy, or model-resolver side effects were added.

Fixes #5806

---
## Summary by cubic
Keeps explicit ULW activation active for follow-up messages in the same session instead of dropping it after the first message. Users no longer need to repeat the keyword; activation clears when the session is deleted, the hook is disposed, or continuation is stopped.

- Follow-ups get one synthetic marker instead of the full directive; compaction or model-family changes restore full guidance once.
- After compaction, system transforms keep full guidance in the runtime prompt until a real message appends it; image-only turns get a durable synthetic part instead.
- Tracks activation per hook with a 256-session cap, evicting the oldest when full.
- `session.deleted` and `session.compacted` events route through the production event dispatcher; `/stop-continuation` clears via the shared stop owner.
- Existing guards for synthetic, planner, and subagent messages still apply, and routing honors the fallback-selected output model.
- Added tests and a replayable real-OpenCode QA harness cover retention, isolation, cleanup, compaction, and eviction.

Fixes #5806.

<sup>Written for commit a60181c2bea594b398ecd8c3016f00d5aa60fec7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7986?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:

code-yeongyu (approved):
Verified locally — keyword-detector suite 100/100 pass, stop-continuation entrypoints 4/4 pass, new follow-up tests 7/7 pass. tsc errors are pre-existing on dev (omo-senpi facts-child-launch).

Clean implementation: bounded FIFO set with proper lifecycle cleanup through all four paths (command, owner, session.deleted, dispose). Guard preservation for synthetic/planner/subagent is correct. Toast suppression on replay-only activation is a nice touch.

LGTM.

PR #8012, opened 2026-09-09
fix(omo-senpi): retire the idle-injection coordinator on session_shutdown

### LINKED ISSUE #7932: [Bug]: Deferred idle-injection flush crashes OMO after reload with stale extension generation

PeterPonyu:
## Summary

Fixes #7932. The shared `IdleInjectionCoordinator` defers its batched flush through `setTimeout(flush, 200)` while closing over the per-generation extension API. If senpi reloads inside that window, the timer fires after the old generation has been invalidated, `pi.sendMessage` throws `stale extension generation after reload`, the coordinator rethrows out of the timer callback, and the host's `uncaughtException` handler exits the process with code 1.

## Changes

`packages/omo-senpi/src/extension/idle-injection-coordinator.ts`
- New `retire()`: sets a fence, drops the pending queue, and turns `enqueue` / `scheduleFlush` / `flushSoon` / `#flush` into no-ops. The two deferred callbacks contain a delivery error only when already retired, so a flush racing the retirement cannot escape the timer queue. The synchronous `flushOnIdle` contract and the `FlushScheduler` seam are unchanged.

`packages/omo-senpi/src/extension/compose.ts`
- Registers `pi.on("session_shutdown", () => idleCoordinator.retire())`. senpi emits `session_shutdown` on the old runner before invalidating that generation, so every flush armed before a reload becomes a no-op. Pending injections are dropped on reload rather than carried over: the queue is a batching window, not a durable store, and durable producers (completions, liveness, ulw continuation) redeliver via the `session_start` recovery chain.

`packages/omo-senpi/plugin/extensions/omo.js`
- Regenerated with bun 1.4.2; it is the only committed bundle that includes `compose.ts`. The other bundles were byte-identical after the rebuild.

## Test plan

- [x] Three new coordinator tests (retire before the deferred flush, retire before the `flushSoon` microtask, enqueue after retire) — RED on unpatched `dev` (`coordinator.retire is not a function`), GREEN after.
- [x] New compose-level test with fake timers reproduces the crash on `dev` (`advanceTimersByTime(200)` throws `stale extension generation after reload` out of the timer) and passes after.
- [x] `bun test packages/omo-senpi/src/extension/`: 78 pass, 0 fail.
- [x] `bun run --cwd packages/omo-senpi typecheck` (tsgo): clean.
- [x] `build-extension.mjs` with bun 1.4.2; extension check reports current; only `omo.js` changed.

## Not covered

`retire()` also runs on `session_shutdown` with `reason: "quit"`, which is harmless. Whether `newSession` / `fork` / `switchSession` emit `session_shutdown` first was not verified against the host; those paths are covered only by the contained-rethrow hardening. A user who reloads mid-`ulw` gets the continuation re-enqueued on the next `agent_end`, matching pre-coordinator behaviour.

---
## Summary by cubic
Fixes #7932: retiring the idle-injection coordinator on `session_shutdown` so a reload inside the 200ms flush window no longer throws `stale extension generation after reload` out of the timer and crashes the host.

- `retire()` stops all enqueues and armed flushes, drops pending injections, and contains delivery errors racing the shutdown.
- Registers `pi.on("session_shutdown", ...)` in `compose.ts` so the fence is set before the old generation is invalidated.
- Pending injections are dropped on reload; durable producers redeliver via the `session_start` recovery chain.

<sup>Written for commit 61b496a0be5a78a15ecc0391c355ff7d4c6864d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8012?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
69/69 pass. Clean retire() pattern — fences deferred flushes, drops pending queue, contains stale-generation errors. Compose-level crash reproduction test is a great addition. LGTM.

PR #8013, opened 2026-09-09
fix(background-agent): keep a task while a synchronous continuation is attached

### LINKED ISSUE #7929: [Bug]: background cleanup deletes an active synchronous continuation
## Description

Background task cleanup deletes an active synchronous continuation (run_in_background=false session poll), while the old background timer still holds a reference to it. This causes the synchronous poll to miss and subsequent task resume attempts to fail with "Task not found".

## Root Cause

1. Task launched with `run_in_background=true` (async/fire-and-forget)
2. Old BackgroundManager completion timer fires at T0
3. Sync continuation called immediately after: `run_in_background=false` poll to same `ses_ID`
4. Client session lifecycle deletes active child at ~T+10m (cleanup timeout)
5. Sync poll at T+2m12s succeeds, but busy-wait at T+9m57s fails when delete occurs
6. Subsequent resume attempt on T+15m timeout fails with "Task not found"

Expected: Cancel old background timer on active sync continuation; reactivate shared lifecycle.
Observed: Active session deleted while still referenced by old background manager.

## Timeline

- T0: bg task completion
- T+2m12s: sync poll (success)
- T+9m57s: busy poll (fails mid-wait, session deleted)
- T+10m: cleanup delete fires
- T+15m: task resume timeout (not found)

## Impact

Mixed-mode workflows (background task launch → sync continuation resume) hang indefinitely or fail with "Task not found" when sync poll extends past cleanup deadline.

## Related Issues

- #6502: background_output "Task not found" (addresses completed result retrieval, not live SDK deletion)
- PR #7225: keep completed background tasks queryable (not applicable to active sync continuations)
- PR #7526: merged 10m cleanup timeout

## Ownership

Owning files:
- `sync-continuation.ts`: cancelSyncSessionDeletion only + direct prompt
- `manager.ts`: scheduleTaskRemoval SDK delete

Should distinguish cleanup for completed vs. active tasks.

PeterPonyu:
## Summary

Fixes #7929. A background task that finished and was later continued with `run_in_background=false` had two deletion timers on one child session that did not know about each other. `BackgroundManager.scheduleTaskRemoval` (10 minutes after completion) fired in the middle of the synchronous poll, dropped the task from `tasks` and called `session.delete`; the sync poller then 404-looped until its inactivity deadline, and a later `run_in_background=true` resume failed with `Task not found for session`. Only `resume()` disarmed the manager's timer; the sync continuation path never did, it only re-armed its own `pendingDeletionTimers`.

## Changes

`packages/omo-opencode/src/features/background-agent/manager.ts`
- `attachSyncContinuation(sessionID)` registers a session as being polled by a synchronous waiter. The removal timer callback returns early for attached sessions. The returned detach removes the registration and, if the task is still terminal with no live cleanup timer, re-arms `scheduleTaskRemoval`, so the 10-minute grace window restarts once the waiter leaves instead of being skipped. The set is cleared on shutdown.

`packages/omo-opencode/src/tools/delegate-task/sync-continuation.ts`
- `executeSyncContinuation` attaches on entry and detaches on every terminal path (prompt-failure return and the final `finally`). Optional-chained, so callers without a manager keep today's behaviour.

No leak is introduced: detach is unconditional in `finally`, and the poll itself is bounded by the parent abort signal and the poller's 30-minute inactivity deadline.

## Test plan

- [x] New `task-removal-sync-attach.test.ts` (fake timers, mocked client): an attached session survives the removal timer with no `session.delete`; detach re-arms a `TASK_CLEANUP_DELAY_MS` timer that then removes the task and deletes the session exactly once; detach before the timer fires leaves the pending timer alone; an untracked session creates no timer.
- [x] `sync-continuation.test.ts`: attach is called with the continuation session id at entry and detach runs on success, on a `pollSyncSession` throw, and on prompt failure.
- [x] RED on unpatched `dev`: 7 fail (`manager.attachSyncContinuation is not a function`, attach never called); GREEN after: 23 pass in the two files.
- [x] `bun test packages/omo-opencode/src/features/background-agent packages/omo-opencode/src/tools/delegate-task`: 1262 pass, 0 fail.
- [x] `bun run --cwd packages/omo-opencode typecheck` (tsgo): clean. Senpi extension bundles unchanged after a rebuild.

## Not covered

The poller's 404 loop and `pruneStaleTasksAndNotifications` (30 minutes after original completion, when other tasks are running) do not consult the attach set; both are pre-existing and left as follow-ups.

---
## Summary by cubic
Fixes a race condition where a completed background task resumed with `run_in_background=false` could be dropped by the manager's 10-minute cleanup timer mid-poll, causing a 404 loop and "Task not found" errors on later resumes. The sync continuation now registers its session with the background manager for the poll window so cleanup waits, and restarts the cleanup timer when the waiter leaves.

- `BackgroundManager.attachSyncContinuation` marks a session as being polled synchronously; the removal timer skips attached sessions, and detaching re-arms the 10-minute removal window.
- `executeSyncContinuation` attaches on entry and detaches on every terminal path, so callers without a manager keep existing behavior.

<sup>Written for commit 2bc89b6320e43a973915e4f640134e0d6a947568. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8013?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/manager.ts
+++ b/packages/omo-opencode/src/features/background-agent/manager.ts
@@ -267,6 +267,7 @@ export class BackgroundManager {
   private queuesByKey: Map<string, QueueItem[]> = new Map()
   private processingKeys: Set<string> = new Set()
   private completionTimers: Map<string, ReturnType<typeof setTimeout>> = new Map()
+  private readonly syncAttachedSessions = new Set<string>()
   private completedTaskArchive: Map<string, BackgroundTask> = new Map()
   private completedTaskSummaries: Map<string, BackgroundTaskNotificationTask[]> = new Map()
   private idleDeferralTimers: Map<string, ReturnType<typeof setTimeout>> = new Map()
@@ -1167,6 +1168,22 @@ The fallback retry session is now created and can be inspected directly.
       }
     }
     return undefined
+  }
+
+  /**
+   * Marks a session as being polled by a run_in_background=false continuation so the
+   * completion cleanup timer neither drops the task nor deletes the session underneath it.
+   * The returned detach restarts the cleanup grace window once the waiter leaves.
+   */
+  attachSyncContinuation(sessionID: string): () => void {
+    this.syncAttachedSessions.add(sessionID)
+    return () => {
+      if (!this.syncAttachedSessions.delete(sessionID)) return
+      const task = this.findBySession(sessionID)
+      if (!task || !TERMINAL_BACKGROUND_TASK_STATUSES.has(task.status)) return
+      if (this.completionTimers.has(task.id)) return
+      this.scheduleTaskRemoval(task.id)
+    }
   }
 
   private resolveTaskAttemptBySession(sessionID: string): { task: BackgroundTask; attemptID?: string; isCurrent: boolean } | undefined {
@@ -2341,6 +2358,9 @@ The task was re-queued on a fallback model after a retryable failure.
       const task = this.tasks.get(taskId)
       if (!task) return
 
+      // A sync waiter is still polling this session; its detach re-arms removal.
+      if (task.sessionId && this.syncAttachedSessions.has(task.sessionId)) return
+
       if (task.parentSessionId) {
         const siblings = this.getTasksByParentSession(task.parentSessionId)
         const runningOrPendingSiblings = siblings.filter(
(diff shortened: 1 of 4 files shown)


code-yeongyu (approved):
757/757 background-agent suite pass, 16/16 new tests pass. Solid race condition fix — attach/detach pattern cleanly prevents the 10-min cleanup timer from killing a session mid-sync-poll. Re-arming on detach is the right call. LGTM.

Issue #7788, opened 2026-09-05
[Bug]: OpenCode Team Mode does not propagate member worktree cwd to child sessions, breaking LSP


llc1123:
## Bug Description

OpenCode Team Mode creates a member's sibling Git worktree and injects its path into the member prompt, but does not propagate that worktree path as the child OpenCode session's working directory. As a result, LSP tools run with the lead session's request CWD and reject valid files in the member worktree:

```text
LSP file path must be inside request cwd: /root/project-1-fix-1/...
```

This affects Team Mode members working in a sibling worktree, for example:

```text
lead workspace: /root/project-1
member worktree: /root/project-1-fix-1
```

## Steps to Reproduce

1. Start OpenCode/OmO from `/root/project-1`.
2. Start an OmO Team Mode run with a member assigned to a sibling worktree `/root/project-1-fix-1`.
3. Let the member edit or inspect a file in that worktree.
4. Run `lsp_diagnostics` (or another `lsp_*` tool) for the worktree file.
5. Observe that the request is rejected because the request CWD remains `/root/project-1`.

## Actual Behavior

The worktree path is placed in the member prompt, but the child session is created using the parent session directory.

Relevant source paths on `dev`:

- `packages/omo-opencode/src/features/team-mode/team-runtime/create.ts:183-207`
  - creates/resolves `resource.worktreePath`
  - passes it only to `buildMemberPrompt(...)` and the prompt text
  - calls `bgMgr.launch(...)` without a `cwd`/`worktreePath` field
- `packages/omo-opencode/src/features/background-agent/types.ts:112-134`
  - `LaunchInput` has no `cwd` or worktree path field
- `packages/omo-opencode/src/features/background-agent/spawner.ts:47-65`
  - reads the parent session directory
  - calls `client.session.create({ query: { directory: parentDirectory } })`
- `packages/omo-opencode/src/features/background-agent/spawner.ts:119-122`
  - sends the child prompt using the same `parentDirectory`

The member therefore receives instructions like:

```text
Worktree: /root/project-1-fix-1
```

but its OpenCode/LSP request context remains rooted at:

```text
/root/project-1
```

The LSP guard then correctly rejects the sibling path in `packages/lsp-core/src/lsp/client-wrapper.ts:88-95`, where it checks `isPathInside(contextCwd(), canonicalFilePath)`.

## Expected Behavior

[...]

PR #8015, opened 2026-09-09
fix(team-mode): spawn members in their worktree directory

PeterPonyu:
## Summary

Addresses #7788. Team Mode resolves a member's sibling worktree path (`team-runtime/create.ts`) but only ever injected it into the member's prompt text. The child session itself was created and prompted in the lead session's directory: `LaunchInput` had no cwd field and `BackgroundManager.startTask` hard-coded `query.directory` to the parent directory for both `session.create` and the prompt. Team messaging was already the other half of the story and routes member prompts with `directory: recipientMember.worktreePath ?? directory`; spawning was the inconsistent half.

## Changes

`features/background-agent/types.ts`
- Optional `cwd` on `LaunchInput` and `BackgroundTask`: the directory the child session is created and prompted in; defaults to the parent session directory.

`features/background-agent/manager.ts`, `spawner.ts`, `spawner/task-record.ts`
- `childDirectory = input.cwd ?? parentDirectory` is used for `session.create`, `buildLocalSessionUrl` and both `promptWithRetryInDirectory` calls; `resume()` prompts with `existingTask.cwd ?? this.directory`. The record keeps `cwd` so retries inherit it. The spawner twin mirrors the same logic so it does not drift from the live path.

`features/background-agent/fallback-retry-handler.ts`
- The fallback retry input carries `cwd: task.cwd`.

`features/team-mode/team-runtime/create.ts`
- Passes `cwd: resource.worktreePath` into `launch` when the member has a worktree; members without one are unchanged.

Every other `launch` caller passes no `cwd`, so their behaviour is identical to before.

## Test plan

- [x] 8 new tests, RED on unpatched `dev` (6 fail: `session.create` received `{ directory: "/parent" }` instead of the worktree; `launchInput.cwd` undefined) → GREEN after: `manager-session-permission.test.ts` (child create + prompt use `input.cwd`; fallback to parent when absent; `resume()` uses the task cwd), `spawner.test.ts`, `fallback-retry-handler.test.ts` (retry input preserves `task.cwd`), `team-runtime/create.test.ts` (worktree path passed as `cwd`; omitted without a worktree).
- [x] `bun test packages/omo-opencode/src/features/background-agent packages/omo-opencode/src/features/team-mode`: 1060 pass, 1 pre-existing skip, 0 fail.
- [x] `bun run --cwd packages/omo-opencode typecheck` (tsgo): clean. Senpi extension bundles unchanged after a rebuild.

## Not covered

This is the OmO-side plumbing only. Whether the LSP error in the issue disappears depends on OpenCode creating a per-directory plugin instance (and therefore a worktree-rooted `lsp` MCP) for a child session created with `directory=<worktree>`; that cannot be verified in this repository, which is why this says "addresses" rather than "fixes". Poller, abort and session-existence checks still address the child through the plugin directory and are left as they were.

---
## Summary by cubic
Fixes team-mode spawning so member sessions run in their own worktree directory, addressing #7788. Previously sessions were created and prompted in the lead directory; now team members with worktrees pass their worktree path as `cwd`.

**Bug Fixes**

- Adds an optional `cwd` to `LaunchInput` and `BackgroundTask`, defaulting to the parent directory, and threads it through session creation, prompting, fallback retries, and resumes.
- Only team members with worktrees set it; other launch callers keep the parent directory default.
- Whether the LSP error disappears depends on OpenCode creating a per-directory plugin instance, so this addresses rather than fixes the issue.

<sup>Written for commit d32c3e5ba6654a4dbba24f7eccf1b2c04b9b3bf8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8015?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/fallback-retry-handler.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/fallback-retry-handler.test.ts
@@ -299,6 +299,20 @@ describe("tryFallbackRetry", () => {
       expect(retryInput?.sessionPermission).toEqual(QUESTION_DENIED_SESSION_PERMISSION)
     })
 
+    test("retry input preserves task.cwd", async () => {
+      const args = createDefaultArgs({
+        teamRunId: "team-run-1",
+        onSessionCreated: mock(async () => {}),
+        cwd: "/lead-fix-1",
+      })
+
+      await tryFallbackRetry(args)
+
+      const key = `${args.task.model!.providerID}/${args.task.model!.modelID}`
+      const retryInput = args.queuesByKey.get(key)?.[0]?.input
+      expect(retryInput?.cwd).toBe("/lead-fix-1")
+    })
+
     test("finalizes the failed attempt, creates a new pending attempt, and enqueues its explicit attemptID", async () => {
       const args = createDefaultArgs({
         status: "running",
--- a/packages/omo-opencode/src/features/background-agent/fallback-retry-handler.ts
+++ b/packages/omo-opencode/src/features/background-agent/fallback-retry-handler.ts
@@ -224,6 +224,7 @@ export async function tryFallbackRetry(args: {
     sessionPermission: task.sessionPermission,
     category: task.category,
     isUnstableAgent: task.isUnstableAgent,
+    cwd: task.cwd,
     onSessionCreated: task.onSessionCreated,
   }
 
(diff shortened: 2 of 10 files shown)


code-yeongyu (approved):
1052/1052 pass (1 pre-existing skip). Clean plumbing fix — cwd flows through LaunchInput to session.create and prompt, with proper fallback to parent dir. Team create passes worktree path, all other callers unchanged. LGTM.

PR #8017, opened 2026-09-09
fix(omo-senpi): rebind memory identity on before_agent_start after a host restart

### LINKED ISSUE #7764: [Bug]: Memory tools permanently fail with "no memory identity bound" after a host restart resumes an open session
### Prerequisites

### Bug Description

When the omo host process restarts **while a conversation is open** (in my case: `bun i -g omo-ai@beta` run mid-session — the normal way to track the beta channel) and the conversation is then resumed, every `memory` / `memory_apply_patch` call fails permanently for the remainder of that session:

> no memory identity bound to this session; enable omo memory and restart the session so the memory tools can initialize

…even though memory is enabled and worked correctly earlier in the same conversation (commits succeeded). A **fresh** session heals immediately; resuming the same conversation never does.

### Steps to Reproduce

1. Ensure `memory.enabled: true` in `~/.omo/omo.jsonc`. Start a new session and confirm the memory tool works (e.g. a `create` commits — the session jsonl gets exactly one `senpi-memory.session-binding` custom entry).
2. With that conversation still open, update omo in another terminal: `bun i -g omo-ai@beta` (host process restarts; the conversation resumes).
3. Call the memory tool again → the error above, permanently for that session. Recall notices also stop appearing for it.

### Expected Behavior

The resumed session re-establishes the memory binding (the recorded binding entry is present and still matches the config-resolved identity), so memory tools keep working across a host restart — or at minimum the failure is surfaced visibly with an accurate message.

### Actual Behavior

All memory tool calls are rejected, with advice ("enable omo memory") that is wrong for this state since memory is enabled. No warning or log is emitted anywhere else.

### Environment

- omo 5.0.0-0.beta.40 (engine: senpi 2026.9.4) — current `beta` dist-tag
- macOS 26.5 (darwin 25.5.0, arm64), bun global install

### Doctor Output

```
PASS senpi CLI: ~/.bun/install/global/node_modules/@code-yeongyu/senpi/dist/cli.js
PASS senpi version 2026.9.4
INFO omo 5.0.0-0.beta.40 (engine: senpi 2026.9.4)
```

### Error Logs

The tool error quoted above is the only observable — no surrounding log line appears (the in-code conflict/binding-failure signal is UI-notify only; see root cause). The memory repo itself was clean with its git identity intact, and commits made earlier in the same session were preserved.

### Configuration

```jsonc
"memory": {
  "enabled": true,
  "agent": "auto",
  "tool_exposure": "direct",
  // reflection/nudge blocks left at defaults
}
```
No `omo-senpi-memory-disabled` / `omo-senpi-disabled` flags set; no `SENPI_MEMORY_*` env vars set.

### Root cause (verified against the installed bundle)

Reading the installed `plugin/extensions/omo.js`:

- The memory identity context is bound **only** in the `session_start` handler: resolve identity from config → conflict-check against the newest `senpi-memory.session-binding` session entry → store the per-session context → append a fresh binding entry.
- A resumed conversation does **not** re-fire `session_start`, so the restarted process's session→context map stays empty; the tool guard (`identity() === undefined → error`) then rejects every call.
- Evidence from the affected session's jsonl: exactly **one** binding entry, from the original start; none appended after the resume.

Ruled out by direct inspection:

- **Disabled memory** — config enabled, no flags, no env overrides.
- **Identity conflict** — the recorded binding identity equals the identity re-resolved from the same config after the update; no new agent dir appeared under `~/.omo/memory/agents/`.
- **Memory repo corruption** — repo clean, identity configured, prior commits intact.

### Suggested fix

Re-bind on session resume/continuation (or lazily on the first memory tool call): re-validate the recorded binding entry against the currently resolved identity and restore the context when it matches — the validation logic already exists in the `session_start` path and can be reused verbatim. Also consider surfacing the conflict/binding-failure signal outside the UI (`ui.notify` is the only channel today, so headless/resumed sessions get no warning at all), and making the error message not claim memory is disabled when the cause is a missing binding.

### Additional Context

Severity is low (no data loss; a new session heals), but memory recording is silently disabled for the rest of a long resumed session — for long-running autonomous sessions that can be hours of unrecorded context. Happy to test a fix build.

PeterPonyu:
## Summary

Fixes #7764. The memory identity context was created in exactly one place, the `session_start` handler in `packages/omo-senpi/src/components/memory/index.ts`, and nothing else could ever (re)create it. When the host process restarts (for example `bun i -g omo-ai@beta` mid-session) and the open conversation is resumed without `session_start` reaching the new runner generation, `resolveContext` returns nothing for that session and every memory tool call fails for the rest of the session with `no memory identity bound to this session; enable omo memory and restart the session`, even though memory is enabled and the identity is fully derivable from config, the process cwd and the env. The reporter's session file shows exactly one binding entry and none after the restart, i.e. the handler never ran again.

## Changes

`packages/omo-senpi/src/components/memory/index.ts`
- The bind body moves into `bindSession(surface, eventCtx, { existing, verifyRepository })`, called from `session_start` exactly as before (fresh state, `releaseSession` first, identity-only conflict check, same conflict message).
- A new `before_agent_start` handler, registered after the static handlers so their projection-first result order is unchanged, binds a session that has no context yet: it skips events without a session id, sessions that already have a context (idempotent: no second binding entry, no second supervisor reference) and sessions whose enablement was latched `false` at `session_start`. The identity is derived from the same inputs as the original bind (config agent, process cwd captured at `register()`, env) and the session id comes from the same event context; nothing model-controlled feeds it.
- Because a rebind has no `session_start` behind it, it validates the session's recorded binding on both `identity` and `repoPathHash`; on a mismatch it fails closed, notifies once per session (`conflictNotified`) and logs one warning, so a handler that fires every turn cannot spam.
- On the resumed turn the memory tool is live immediately (`afterBind` marks the session active synchronously); the memory prompt block follows on the next turn.

`packages/omo-senpi/src/components/memory/tools.ts`
- The unbound guard message no longer blames disabled memory: `no memory identity bound to this session yet; the binding is re-established on the next user turn, or start a new session if this persists`.

`packages/omo-senpi/plugin/extensions/omo.js`, `omo-task.js`: regenerated with bun 1.4.2 (the memory component is bundled into both); the other bundles were byte-identical.

## Test plan

- [x] New `index.test.ts` cases: a session whose `session_start` never reached the runner is bound exactly once on `before_agent_start` from its recorded binding (two dispatches, one entry, one supervisor reference, no notification); a recorded binding with a different repository hash fails closed with one notification and one warning after two dispatches; a session bound at `session_start` takes no second entry; a latched-disabled session and an event without a session id bind nothing.
- [x] `tools.test.ts`: the guard message names the next-turn recovery and no longer says "enable omo memory".
- [x] RED on unpatched `dev`: 3 fail (rebind, fail-closed, message), the two regression guards already green; GREEN after: 20/20 in the two files.
- [x] `bun test packages/omo-senpi/src/components/memory/`: 1312 pass, 1 pre-existing skip, 0 fail (including `compaction-survival.test.ts`, which pins the projection-first `before_agent_start` result order). `bun test packages/omo-senpi/src/extension/`: 74 pass.
- [x] `bun run --cwd packages/omo-senpi typecheck` (tsgo): clean. Extension check reports current.

## Not covered

The exact senpi code path that withheld `session_start` on the reporter's host (senpi 2026.9.4) is not in this tree; the vendored 2026.9.9 emits it on every attach path I could find, so this fix is deliberately trigger-agnostic. The `session_start` path keeps its identity-only conflict check; the stricter repository check applies to the rebind only, where the recorded entry is the sole evidence.

---
## Summary by cubic
Fixes #7764. Memory identity was bound only in `session_start`, so after a host restart a resumed session could reach a turn without a binding and every memory tool call failed with "no memory identity bound". A new `before_agent_start` handler rebinds such sessions from their recorded binding.

- Rebind is idempotent: already-bound sessions, latched-disabled sessions, and events without a session id skip it.
- Rebind validates identity and repository hash; on mismatch it fails closed and notifies once per session.
- On the resumed turn the memory tool is live; the memory prompt block follows on the next turn.
- The unbound guard message now points to next-turn recovery instead of asking to enable memory.
- `session_start` behavior is unchanged apart from sharing the bind code; only the rebind checks the repository hash.
- The exact senpi path that skipped `session_start` is not in this tree, so the fix is trigger-agnostic.
- Extension bundles (`omo.js`, `omo-task.js`) were regenerated after merging upstream `dev`.

<sup>Written for commit af65785adc4335a370f3d57acf041159cc495f5e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
15/15 pass (index + tools). Solid fix for host-restart rebind gap — before_agent_start handler with idempotent binding, fail-closed on repo hash mismatch, conflict spam guard. LGTM.

Issue #7670, opened 2026-09-03
Correct keys with malformed values still discard the whole config layer


mooire733:
This is a follow-up to #7254, which was rightly closed for committing the generated assets/omo.schema.json artifact. Heads-up plus a question, not a demand.

a66022557 solved the bigger half: unrecognized keys are now stripped per dotted path with an `unknown-keys` diagnostic, and surviving entries keep working. That covers most of what #7254 was after, including its `bogus_key` test fixtures.

The remaining gap is the one your commit message calls out: malformed values still reject the layer exactly as before. A known key with a wrong-typed value, for example:

```jsonc
"task": { "default_concurrency": "not-a-number" }
"agents": { "sisyphus": { "model": 123 } }
```

throws away the entire config layer, so every category, agent, and task setting in that file silently reverts to defaults.

Proposal: bounded per-leaf handling for invalid agents/categories leaves (or a partial-accept analog consistent with the stripUnrecognizedKeys design), each removal reported as its own diagnostic, with the prototype-pollution fail-closed guard from a66022557 running first.

We have a source-only patch whose schema-keys part already passes against current dev (agent-schema-keys tests 3/3). The loader part needs a manual rebase onto the new control flow, which is fine. Would you like that PR, or do you plan to handle this case yourselves?

code-yeongyu:
Good catch on the remaining gap. You're right that a66022557 only handles unknown keys — malformed values on known keys still reject the entire layer, which is a rough failure mode when one typo nukes all your agent/category/task config silently.

Per-leaf validation with individual diagnostics (while keeping the prototype-pollution guard as the first gate) sounds like the right approach. Happy to take a PR for this if you've got one ready to rebase onto current dev.

PR #8019, opened 2026-09-09
fix(omo-config-core): run the prototype-tamper guard on every layer, not only on unknown-key failures

PeterPonyu:
## Summary

Defense-in-depth fix in `packages/omo-config-core/src/loader/loader.ts`, found while checking the loader's fail-closed prototype policy against the per-leaf salvage work in #7676 / #7670. It does not change what either of those proposes; it closes a gap that exists on `dev` today and would also exist under any salvage scheme that keeps the guard where it is.

The prototype-tamper guard (`hasTamperedPrototype`) only ran when `OmoConfigLayerSchema.safeParse` failed *and* reported at least one `unrecognized_keys` issue. A JSON `"__proto__"` member is written through the prototype rather than as an own property, and `OmoAgentDefSchema` runs a `z.preprocess` that rebuilds the agent object before strict validation, so this layer validates cleanly and loads:

```jsonc
{"agents":{"evil":{"__proto__":{"polluted":true},"model":"ok"}}}
```

`toRecord` rebuilds only the root from its own enumerable properties; nested objects keep their prototype, so `result.layers[].config.agents.evil` is handed to consumers with `Object.getPrototypeOf(evil)` set to the injected payload. Today the same layer is rejected only when a sibling value happens to be malformed (`"model": 123`), because that produces a validation failure — i.e. by accident. Reproduced on `dev@b5122f19d` with a throwaway probe (four payloads: nested + malformed sibling → rejected; nested + valid sibling → **loaded, tampered**; root-level → rejected; malformed only → rejected).

## Changes

- Run `hasTamperedPrototype(parsed.data)` once, right after the JSONC parse and before validation, for every layer; a hit rejects the layer with a `validation` diagnostic (`… "__proto__" member is not allowed`).
- `hasUnsafeUnrecognizedKey` keeps only the own-key check for `prototype` / `constructor` (those do surface as unrecognized keys); the redundant prototype walk is removed and the comments now describe the actual order.
- Hard-fail set is unchanged otherwise: unreadable file, unparsable JSONC, non-object root, tampered prototype at any depth, and layers that still fail after unknown-key stripping.
- `omo.js` / `omo-task.js` regenerated with bun 1.4.2 (the loader is bundled into both).

## Test plan

- [x] New `unknown-keys.test.ts` case: nested `__proto__` under an agent whose other values are valid → layer not loaded, one `validation` diagnostic, no loaded layer carries a tampered sub-object, global `Object.prototype` untouched. RED on `dev` (layer loads), GREEN after.
- [x] Existing root-level `__proto__` test and the `{"categories":"nope"}` rejection test unchanged and green.
- [x] `bun test packages/omo-config-core`: 206 pass, 0 fail. `bun run --cwd packages/omo-config-core typecheck` (tsgo): clean. Extension check reports current.

## Notes

Threat model is local: the payload has to be in the user's own config file, so this is robustness against a documented fail-closed policy rather than a remotely reachable issue. Related: #7670, #7676 (salvaging malformed values makes running this guard first mandatory, as the maintainer noted on the issue; with this change the order no longer depends on the shape of the salvage).

---
## Summary by cubic
Runs the config loader's prototype-tamper guard on every parsed layer instead of only after unknown-key validation failures. Nested `__proto__` payloads that previously loaded silently are now rejected with a `validation` diagnostic; valid layers are unaffected.

- Limits the unknown-key helper to own `prototype` and `constructor` keys.
- Adds a regression test for a valid agent containing nested `__proto__` data and verifies `Object.prototype` remains unchanged.
- Regenerates the bundled loader in `omo.js` and `omo-task.js`.

<sup>Written for commit 0ccdadd9b8929f6522ff58131a5d8b2538c0293c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
205/205 config-core pass. Good catch — prototype tamper guard moved before validation so nested __proto__ under valid siblings can't bypass it. Defense-in-depth, local threat model only, but closes a real gap. LGTM.

PR #8021, opened 2026-09-09
fix(memory-core): reclaim a proven-dead recovery lock instead of deadlocking

### LINKED ISSUE #7573: memory: reflection-scheduler lock recovery deadlocks when the recovery lock itself is stale (dead holders require manual rm)

PeterPonyu:
## Summary

Fixes #7573. `recoverDeadOwner` in `packages/memory-core/src/locks/acquire.ts` applies the full proven-dead test (same-host pid liveness plus the `/proc` start-identity PID-reuse check) to the primary `reflection-scheduler.lock`, but when publishing its short-lived `.lock.recovery` mutex fails with `EEXIST` it returned `false` without ever reading or liveness-checking the existing recovery file. That file's only remover is the nonce-matched `releaseLock` in its creator's `finally`, so a recoverer killed inside the millisecond critical section leaves a permanently un-evictable recovery lock, and every later contender hits `LockContentionError` on the primary until someone deletes the file by hand. The function has not changed since the original port, and the bug reproduces on `dev@b5122f19d`.

This reworks the closed #7621 (thanks @MoerAI — that branch was closed over a bundle modify/delete conflict, not on design) against the current adapter layout, with a smaller protocol: no second lock level and no claim chain.

## Changes (`acquire.ts` only, +52/−1)

- `reclaimDeadRecoveryLock(recoveryPath)`: read the recovery record; return `false` unless it parses **and** `isProvenDead` holds for its owner (same predicate the primary already uses: cross-host never reaped, `EPERM` = alive, PID reuse only reaped when `/proc` start identity is readable and mismatched, Windows fail-closed). Then `rename` it to a private `.reaping-<uuid>` tombstone — atomic, so exactly one reaper obtains the inode — re-read, and `unlink` only if the moved bytes equal what was seen. Otherwise a fresh live holder was grabbed and the file is handed back with `link` (`EEXIST` means yet another contender republished first; nothing is lost). `ENOENT` and Windows sharing violations return `false`; other fs errors propagate (no "just unlink" fallback). The tombstone name deliberately does not match `LEAKED_CANDIDATE_NAME` so a concurrent candidate sweep cannot delete it mid-inspection.
- The single `publishExclusive` early return becomes a bounded loop of at most one reclaim plus one re-publish, so a bind-time `acquireLock(..., { waitTimeoutMs: 0 })` recovers a doubly-stale lock in one pass instead of failing once and succeeding on the next bind.
- A nonce fence before `unlink(lockPath)`: the primary is evicted only while this contender still owns the recovery lock.

Unchanged on purpose: `isProvenDead` semantics, no age/mtime-based reaping of either file (`lock-record.test.ts` pins this), nonce-matched release, the raw-equality re-read before unlinking the primary.

`omo.js`, `omo-task.js`, `memory-run-supervisor.mjs` regenerated with bun 1.4.2 (the lock protocol is bundled into all three).

## Test plan

- [x] New `lock-record.test.ts` cases: dead primary owner **and** dead recovery-lock owner → a single `waitTimeoutMs: 0` acquire reclaims both and holds the primary (RED on `dev`: `LockContentionError: Lock is held`); dead primary + **live** recovery holder → contention raised, neither file touched; dead recovery holder on another host → not reclaimed. The last two pass before and after and pin the safety invariants.
- [x] `bun test packages/memory-core/src/locks/`: 22 pass, 0 fail. Whole `packages/memory-core`: 747 pass, 0 fail (85 files). `bun run --cwd packages/memory-core typecheck` (tsgo): clean. Extension check reports current.

## Not covered

The rename-then-inspect "hand-back" branch (recovery file replaced by a live record between the reaper's read and its `rename`) is reasoned from POSIX `rename`/`link` atomicity and not exercised deterministically: the window contains no I/O boundary a test can hook without adding a production seam only for tests. A reaper killed between `rename` and `unlink` leaves a `.recovery.reaping-<uuid>` tombstone that blocks nothing; extending `LEAKED_CANDIDATE_NAME` to collect it after the sweep interval is a possible hygiene follow-up.

---
## Summary by cubic
Fixes #7573 by reclaiming a stale `.lock.recovery` file so a recoverer killed inside the recovery critical section no longer blocks all future lock acquisitions.

- `reclaimDeadRecoveryLock` applies the same proven-dead test as the primary lock and removes the file atomically via rename-then-inspect.
- A `waitTimeoutMs: 0` acquire recovers a doubly-stale lock in a single pass; a nonce fence prevents unlocking the primary unless this contender still owns the recovery lock.
- New tests pin that live and foreign-host recovery holders are never displaced.
- Bundles (`omo.js`, `omo-task.js`, `memory-run-supervisor.mjs`) regenerated after merging `upstream/dev`.

<sup>Written for commit 3a727460ea8c26a30ae53996048a65c9f1f74f04. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8021?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
19/19 locks suite pass. Excellent fix — dead recovery lock reclaim with atomic rename-then-inspect, proper nonce fence, bounded retry loop for doubly-stale case. Safety invariants pinned (live holder untouched, cross-host fail-closed). LGTM.

PR #8022, opened 2026-09-09
fix(omo-senpi): stop failing reflection closed when a fresh catalog omits every candidate

### LINKED ISSUE #7923: [Bug]: Memory reflection dies with spawn_failed when the model-catalog preflight succeeds but omits the only candidate
### Prerequisites

### Bug Description

Memory reflection children are launched with model discovery disabled, so the parent first runs a **model catalog probe** (`senpi --no-extensions --no-skills --no-prompt-templates --no-context-files --list-models`) and filters the candidate list against the parsed catalog.

The failure handling of that preflight is asymmetric:

| preflight outcome | result |
| --- | --- |
| probe **errors** (timeout / non-zero exit / no parseable ids) | `kind:"unavailable"` → proceed with reactive model retries |
| probe **succeeds**, catalog omits the candidate, **but a <120s cached probe exists** | `kind:"unavailable"` → proceed |
| probe **succeeds**, catalog omits the candidate, **no fresh cache** | `kind:"none_visible"` → **throw** → run dies as `spawn_failed` |

So a catalog that parsed fine but happens to be *incomplete* is the only case treated as authoritative, and it is fatal. The reflection lane passes exactly **one** candidate, so there is no second rung to fall back to.

On this host that fires intermittently against a model id that works everywhere else: **21 of 245** reflection completion records are `spawn_failed`, all naming the same id (`openai-codex/gpt-5.6-luna-fast`), spread over **11 agent identities** between `2026-08-21T12:55:37.144Z` and `2026-09-07T14:53:45.927Z`. The same id succeeded in the other 213 runs, often minutes before and after a failure. Each occurrence surfaces a user-visible `✗ Memory reflection failed` warning and the transcript cursor is not advanced.

### Steps to Reproduce

1. Run OmO Native with memory reflection enabled and **no** reflection/memory model override, so the reflection lane resolves the built-in `quick` requirement to a single candidate (here: `openai-codex/gpt-5.6-luna-fast`).
2. Let reflection run normally for a few days across several agent identities.
3. Inspect `~/.omo/memory/agents/*/runtime/reflection/completions/*.json` and filter `"reason": "spawn_failed"`.
4. Observe failures interleaved with successes for the *same* model id, each finishing in ~1 second (`982–1830 ms`), i.e. far below the 10s probe timeout.

Deterministic variant of the same code path: point the reflection candidate at any id that the probe's `--list-models` table does not print. Every reflection run then dies immediately with `none_visible`, even though the parent session runs happily on its own model.

Manually running the exact probe argv on this host parses **1481** model ids including 12 `openai-codex/*` entries, and `openai-codex/gpt-5.6-luna-fast` **is** present — so the omission that kills these runs is transient, not a static configuration problem.

### Expected Behavior

A parseable-but-incomplete catalog should be treated as *inconclusive*, exactly like a probe error: degrade to `kind:"unavailable"` and let the reactive retry path decide. In particular, when the rejected id is the model the parent session is currently running, "not in this catalog snapshot" is far more likely to be a transient listing gap than a real absence.

Additionally, the reflection lane should receive the in-category fallback chain (the same gap that was fixed for the memorian judge child in #7904), so a single visibility miss has a second rung instead of killing the run.

### Actual Behavior

```
No reflection model candidate is visible to the discovery-disabled child: openai-codex/gpt-5.6-luna-fast (model_not_visible)
```

The run is recorded as `outcome: "failed", reason: "spawn_failed"` after ~1s, the owner gets a `✗ Memory reflection failed` warning, and the transcript cursor stays where it was.

### Doctor Output

```shell
PASS senpi CLI: /home/<user>/.bun/install/global/node_modules/@code-yeongyu/senpi/dist/cli.js
PASS senpi version 2026.9.7
INFO omo 5.0.0-0.beta.47 (engine: senpi 2026.9.7)
WARN stale engine pid 1501854 (age 9-02:28:42, tty ?) has no launcher; it was orphaned by a signaled launcher
INFO reap them explicitly by pid: omo doctor --reap 1501854
```

### Error Logs

```shell
# ~/.omo/memory/agents/<agent-K>/runtime/reflection/completions/reflection-run-29.json  (identities redacted)
{
  "schemaVersion": 1,
  "runId": "reflection-run-29",
  "identity": "<agent-K>",
  "category": "quick",
  "model": "openai-codex/gpt-5.6-luna-fast",
  "thinking": "low",
  "conversationIds": [
    "<uuid>"
  ],
  "trigger": "step-count",
  "outcome": "failed",
  "reason": "spawn_failed",
  "detail": "No reflection model candidate is visible to the discovery-disabled child: openai-codex/gpt-5.6-luna-fast (model_not_visible)",
  "startedAt": "2026-09-07T14:53:23.504Z",
  "finishedAt": "2026-09-07T14:53:24.566Z",
  "durationMs": 1062,
  "consecutiveFailures": 1,
  "delivery": {
    "status": "consumed",
    "sessionId": "<uuid>",
    "consumedAt": "2026-09-07T14:53:24.567Z"
  }
}

# aggregate over ~/.omo/memory/agents/*/runtime/reflection/completions/*.json
records=245  merged=174  no_changes=39  spawn_failed=21  child_exit=8  merge_conflict=3
spawn_failed detail  : all 21 are "No reflection model candidate is visible to the discovery-disabled child: openai-codex/gpt-5.6-luna-fast (model_not_visible)"
affected identities  : 11 (agent-A: 1, agent-B: 1, agent-C: 1, agent-D: 1, agent-E: 2, agent-F: 1, agent-G: 1, agent-H: 1, agent-I: 3, agent-J: 1, agent-K: 8)
window               : 2026-08-21T12:55:37.144Z .. 2026-09-07T14:53:45.927Z
duration             : 982-1830 ms (probe timeout is 10000 ms)

# manual run of the preflight argv on the same host
$ senpi --no-extensions --no-skills --no-prompt-templates --no-context-files --list-models
parsed model ids = 1481, openai-codex/* = 12, openai-codex/gpt-5.6-luna-fast present = true
```

### Configuration

```json
{
  "telemetry": { "enabled": false },
  "memory": { "sync": { "remote": "<redacted private repo>" } },
  "//": "no memory/reflection model override; the reflection candidate openai-codex/gpt-5.6-luna-fast comes from the built-in requirement chain for category 'quick'"
}
```

### Additional Context

Mechanism as published in `plugin/extensions/omo.js` of `omo-ai@5.0.0-0.beta.47` (minified; de-obfuscated below, grep the string literals to locate the source):

```js
const PROBE_ARGS = ["--no-extensions","--no-skills","--no-prompt-templates","--no-context-files","--list-models"];
const cache = new Map();

async function preflight(input) {
  const key   = cacheKey(input.launch, input.configSources);       // [command, prefixArgs, configSource mtimes]
  const now   = (input.now ?? Date.now)();
  const entry = cache.get(key);
  const fresh = entry !== undefined && now - entry.probedAt < 120_000 ? entry : undefined;
  let visible = fresh?.visible;

  if (visible === undefined) {
    try {
      visible = await probeCatalog(input.launch, input.env, input.timeoutMs ?? 10_000);
      cache.set(key, { visible, probedAt: now });
    } catch (error) {
      input.warn?.("memory child model preflight failed; falling back to reactive model retries", { error });
      return { kind: "unavailable", candidates: input.candidates };   // <- probe failure is tolerated
    }
  }

  const ok       = input.candidates.filter(c =>  visible.has(c.model));
  const rejected = input.candidates.filter(c => !visible.has(c.model))
                                   .map(c => ({ model: c.model, cause: "model_not_visible" }));

  return ok.length === 0
    ? (fresh === undefined
        ? { kind: "none_visible", rejected }                          // <- fatal: throws below
        : { kind: "unavailable", candidates: input.candidates })
    : { kind: "filtered", candidates: ok, rejected };
}

// caller
if (result.kind === "none_visible")
  throw new Error(`No ${surfaceName} model candidate is visible to the discovery-disabled child: ` +
                  result.rejected.map(r => `${r.model} (${r.cause})`).join(", "));
```

Suggested fix, smallest first:

1. Treat "candidate missing from a freshly probed catalog" as inconclusive rather than authoritative — return `unavailable` instead of `none_visible` (at minimum when the rejected id equals the parent session's current model), and let the reactive `model_not_visible` detector (the `Error: Model "X" not found` parser) make the final call after an actual spawn attempt.
2. Give the reflection lane the in-category fallback chain, as done for the memorian judge in #7904, so one miss is not fatal.
3. Optionally re-probe once before declaring `none_visible`, since the catalog child is cheap relative to losing a whole reflection run.

What I could not determine from the host side: *why* a successful probe occasionally omits the `openai-codex` (subscription OAuth) rows. Whatever that cause is, the handling above turns a transient listing gap into a dropped reflection run.

Related (not duplicates): #6808 (reflection stuck on a single quick-route model with no recovery), #7904 (closed — judge child built without the in-category fallback chain), #7829, #7912.

### Operating System

Linux

### OpenCode Version

n/a — OmO Native edition: `omo 5.0.0-0.beta.47 (engine: senpi 2026.9.7)`, node v24.19.0, bun 1.3.14, Linux 7.0.0-28-generic

PeterPonyu:
## Summary

Fixes #7923. `preflightMemoryModels` (`packages/omo-senpi/src/components/memory/worker/model-preflight.ts`) treats a *freshly probed* `--list-models` catalog as authoritative: when it lists none of the candidates the preflight returns `none_visible`, `resolveAndPreflightMemoryLaunch` turns that into a throw before any child is spawned, and `finalizeExecutionFailure` labels the run `spawn_failed`. The identical catalog served from the 120 s cache, or any probe error, is treated as inconclusive (`unavailable`) and proceeds to the reactive attempts. Nothing validates that the catalog is complete, the discovery-disabled child intermittently omits whole providers (the reporter's `openai-codex` rows), and on a single-provider host `resolution.fallbacks` is empty, so `filtered` can never rescue the lone candidate. Result: a model that the parent session is actively running on is declared invisible and reflection dies without a spawn.

## Changes (+8/−10 across two files)

- `model-preflight.ts`: drop the `none_visible` member of `ModelPreflightResult`; a parseable catalog that omits every candidate now logs one warning through `input.warn` and returns `{ kind: "unavailable", candidates }` — the same verdict the cached-negative and probe-error branches already give for the same snapshot. `filtered` is unchanged, so candidates the child provably lists as absent are still skipped whenever at least one is visible.
- `memory-launch-preflight.ts`: delete the `none_visible` throw; both remaining kinds flow into `runMemoryModelAttempts`.

Why this shape rather than the alternatives: more catalog fallbacks do not exist on a single-provider host; a `category_unavailable`-style notice is for "no model exists at all", whereas here the model exists and the parent runs on it. The pre-existing reactive path already classifies Senpi's `Error: Model "X" not found` into `MemoryModelExhaustedError` → `spawn_failed` with a `model_not_visible:<id>` fingerprint and the same remediation hint, so a *genuine* misconfiguration keeps its diagnosis; it now costs one fast-failing child spawn instead of one probe.

`omo.js` regenerated with bun 1.4.2 (the worker preflight is bundled into it; the other bundles were byte-identical).

## Test plan

- [x] New: `model-preflight.test.ts` — a fresh catalog that omits every candidate degrades to `unavailable` and warns (RED on `dev`: `none_visible`, no warning); `memory-launch-preflight.test.ts` — the lone candidate still reaches the child instead of failing closed (RED on `dev`: rejected with `No reflection model candidate is visible …`).
- [x] Retargeted the four tests that pinned the old verdict: the two `model-preflight` fresh/cached cases, the "genuinely cannot see" launch case (now asserts the reactive `MemoryModelExhaustedError` with `model_not_visible:<id>` and the config-pointing remediation), and the runner integration fresh-then-cached case (both runs now proceed reactively; the second run was given its own backlog and diff so the harness's `commit` child has something to commit — test-only change).
- [x] The two slash-id launch tests additionally assert that no preflight warning fired, since they previously relied on `none_visible` to keep a slash-rejecting parser from passing via the degraded path.
- [x] RED 6 fail → GREEN 24/24 on the three suites; `bun test packages/omo-senpi/src/components/memory/worker/`: 267 pass, 0 fail; whole memory directory 1312 pass, 1 pre-existing skip, 0 fail; `bun run --cwd packages/omo-senpi typecheck` (tsgo): clean. `none_visible` has no remaining consumer in `packages/**`.

## Not covered

The root cause of the incomplete `--list-models` output in the discovery-disabled child (provider registration timing for `openai-codex` OAuth) is Senpi-side and not addressed; this change stops the preflight from turning that gap into a fatal verdict. If the spawned child suffers the same listing gap it fails with `Model "X" not found` and the run is still `spawn_failed` — no worse than today.

---
## Summary by cubic
Fixes #7923: reflection preflight no longer fails closed when a fresh `--list-models` catalog omits every candidate. Previously a parseable catalog listing none of the candidates threw `none_visible` before any child spawned; it now warns and returns `unavailable`, so reactive attempts decide after a real spawn.

**Bug Fixes**
- Removed the `none_visible` verdict and throw; a genuinely absent model is still diagnosed as `spawn_failed` with `model_not_visible:<id>` and the same memory.reflection remediation after one child attempt.
- `filtered` is unchanged: candidates the child provably lists as absent are still skipped when at least one is visible.
- `omo.js` regenerated after merging `upstream/dev`.

The root cause of the incomplete `--list-models` output (provider registration timing for `openai-codex` OAuth) is Senpi-side and not addressed. If the spawned child suffers the same gap it fails with `Model "X" not found` and the run is still `spawn_failed` — no worse than today.

<sup>Written for commit 246cfeb2aeed81c3191435b567941177a3dd5d72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8022?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
13/13 pass (preflight + launch-preflight). Good fix — none_visible was too aggressive for incomplete catalogs. Degrading to unavailable lets the reactive path handle it properly while genuine misconfig still gets diagnosed via spawn_failed. LGTM.

PR #8025, opened 2026-09-09
fix(omo-senpi): scope thread_list's default listing to the caller's workspace

### LINKED ISSUE #7632: [Bug]: omo_sessions.send can silently steer an unrelated cross-project session
## Bug description

An `omo_sessions.send` call from **Project A** targeted a short session-ID prefix that resolved to an idle session belonging to unrelated **Project B**. The send was accepted without exposing or confirming the target working directory. The Project A task then appeared in Project B as a normal `role:user` message and was executed there.

This is not an inferred background-notification mix-up: the source transcript contains the explicit send, and the target transcript contains the same payload as the next user message.

## Environment

- OMO: `omo 5.0.0-0.beta.31 (engine: senpi 2026.8.31)`
- Senpi: `2026.8.10`
- OS: Windows 11
- Source session (Project A): `01a05694-cc46-7644-82a9-00ff128418d1`
- Target session (Project B): `01a05a9e-de12-7d64-90c2-42f174a42aff`

Project names and local paths are intentionally anonymized.

## Reproduction evidence

From the Project A session:

```json
{"action":"send","session":"01a05a9e","message":"<Project A task>"}
```

The tool resolved the short prefix to Project B and accepted it:

```json
{"action":"send","sessionId":"01a05a9e-de12-7d64-90c2-42f174a42aff","accepted":true}
```

Seconds later, the Project B transcript recorded the Project A payload as a regular user message:

```json
{"role":"user","content":[{"type":"text","text":"<Project A task>"}]}
```

The Project B assistant then announced that it would perform the Project A correction and started reading and modifying Project A files.

## Actual behavior

- A short session prefix can resolve to a session in a different project/cwd.
- The cross-project send is accepted without a cwd/ownership warning or explicit opt-in.
- The foreign payload is persisted as a normal user turn, so the target agent treats it as authoritative work.

## Expected behavior

Cross-project steering should be fail-safe. At least one of the following should be required:

1. Reject cross-cwd sends by default unless the caller passes an explicit cross-project opt-in.
2. Require the full session ID for a cross-cwd send.
3. Return the resolved full session ID and cwd and require confirmation before delivery.
4. Surface a prominent warning when source and target cwd differ.

## Impact

An agent trying to reuse an idle tool-capable session can silently repurpose an unrelated project's conversation. The target may read, edit, test, or commit against the foreign project while its transcript remains stored under the original project.

## Related

- #6828 concerns automatic ULW continuation hijacking another session in the same cwd. This report is different: it is an explicit `omo_sessions.send` crossing project/cwd boundaries without confirmation.

PeterPonyu:
## Summary

Related to #7632 (not a fix for it — see Notes). `thread_list`'s default scope is documented as "the caller's workspace" (`contracts.ts` parameter description, `thread/AGENTS.md`), and the result is labelled `scope: "workspace"`, but the filter behind it was a tautology: it kept every session whose durable id appears in the address book, and the address book is assembled from every host session with no workspace predicate. So a caller in project A listing threads got project B's threads too, labelled as in-workspace.

The other five thread tools already refuse cross-workspace targets with `scope_denied` unless `all_scope: true` is passed, and that error's `next_action` tells the model to retry with `all_scope`. An unscoped list is therefore the shortest path from "discovery" to a cross-project send that looks opted-in but was not deliberate.

Reproduced on `dev@b5122f19d` with two live sessions in two temp directories and the real `createThreadTools`: default scope returned both, `scope: "workspace"`.

## Changes

- `addressing.ts`: export `workspaceEntries(entries, callerWorkspaceRoot)`, judging entries exactly as the scoped name ladder in `resolveTarget` already does (git toplevel equality when both sides are in git, otherwise realpath equality; no caller root means nothing is in scope, since listing every workspace under a "workspace" label is what `all_scope` exists to make explicit). The helpers stay module-private; only the predicate is exposed.
- `tools.ts`: `thread_list` filters the default listing through it. `all_scope: true` behaviour is unchanged (every live session, `scope: "all"`).
- `omo.js` regenerated with bun 1.4.2 (the thread tools are bundled into it).

## Test plan

- [x] New `tools.test.ts` case: two live threads in two workspaces, caller rooted in the first → default scope lists only the caller's thread under `scope: "workspace"`, `all_scope` lists both under `scope: "all"`. RED on `dev` (both listed), GREEN after.
- [x] `bun test packages/omo-senpi/src/components/thread/ packages/omo-senpi/src/extension/thread-policy.test.ts`: 131 pass, 0 fail. `bun run --cwd packages/omo-senpi typecheck` (tsgo): clean. Extension check reports current.

## Notes

The tool named in #7632, `omo_sessions` (an `action`-dispatching tool with a `session` prefix argument and a `{ sessionId, accepted }` reply), does not exist in this repository at any commit, nor in the pinned Senpi engine; the in-repo `thread_send` resolves by exact id, then exact name within the caller's workspace, and returns `scope_denied` for a full foreign id, so the reported prefix-matched cross-project send cannot come from it. This PR closes the one in-repo scoping gap the investigation found; I have left a note on the issue asking which client shipped `omo_sessions`.

---
## Summary by cubic
Fixes `thread_list`'s default scope so it lists only threads from the caller's workspace. Previously the address book included all sessions without a workspace predicate, so a caller in one project could see another project's threads labeled as "workspace", creating a path to accidental cross-project sends; the other thread tools already refuse cross-workspace targets with `scope_denied`.

- Adds `workspaceEntries` to `addressing.ts`, judging entries with the same logic as the scoped name ladder in `resolveTarget`.
- Filters the default listing through it; `all_scope: true` behavior is unchanged.
- Adds a test case covering two workspaces and regenerates the extension bundles to include the thread tool changes.

This does not fix #7632; the tool named there (`omo_sessions`) does not exist in this repository.

<sup>Written for commit 7b0e52e9419dffc3160335270fd6a97528be433a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8025?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
125/125 thread suite pass. Clean workspace scoping — default thread_list now filters to caller's workspace via sameWorkspace, all_scope explicitly widens. New test covers both paths. LGTM.

PR #8027, opened 2026-09-09
fix(omo-senpi): gate plan continuation on successful turns

rlaope:
## Summary

Prevent OMO's `ulw-loop` and `ulw-execute` hooks from resubmitting failed or refused turns. These hooks ignored the `agent_end` payload and could independently restart a request even after the host stopped retrying it.

This is the OMO companion to [Senpi #1520](https://github.com/code-yeongyu/senpi/issues/1520) and [Senpi #1521](https://github.com/code-yeongyu/senpi/pull/1521). The host fix blocks builtin goal recovery; this change closes the separate OMO plan-continuation path.

## Changes

- Gate both producers on the last assistant outcome before changing continuation counts or status/signature deduplication.
- Do not continue errors, aborts, retry-owned turns, missing outcomes, or structured refusal/sensitive stops. Refusal details remain authoritative when Senpi normalizes an empty tool-use outcome to `stop`.
- Preserve successful continuation, existing caps, and deduplication. Ordinary assistant prose discussing a safety error is not classified as a failure.
- Add direct/coordinator regression coverage and an isolated real-Senpi QA driver. Existing synthetic clean-event fixtures now supply the successful assistant outcome required by the host contract.

## QA & Evidence

- **Failing first:** 44 cases failed before the initial guard. Real CLI QA then exposed normalized refusal, and eight additional cases failed before that correction.
- **Scoped regression:** 163 tests passed across fourteen files, including both hooks and idle-injection wiring.
- **Lead-run real CLI:** four scenarios passed: loop/boulder plans with the exact Codex policy diagnostic or structured refusal. Each made one failed provider call and **zero OMO sends**. Subsequent explicit clean input made one automatic continuation and then deduplicated.
- **Isolation:** all CLI children terminated and all scenario sandboxes were removed. Real OMO/Senpi credential digests were unchanged.

Reproduce from a built checkout:

```sh
BUN_BIN="$(command -v bun)" \
SENPI_BIN="$PWD/node_modules/@code-yeongyu/senpi/dist/cli.js" \
node packages/omo-senpi/scripts/qa/policy-continuation-e2e.mjs
```

Evidence is in `.omo/evidence/omo-senpi-adapter/20260909-policy-continuation-hooks/` and the independent lead run in `20260909-policy-continuation-lead/live-report.json`. The committed driver resolves the canonical evidence directory and captures outcome/ownership/cleanup receipts.

## Risks & Residuals

Error recovery remains Senpi-owned; this does not remove ordinary host retries or infrastructure goal recovery. The original backend classifier reason is unknown and unchanged.

The live fixture loads both changed production components and the real idle coordinator, using its microtask scheduler rather than the full plugin's batching timer. It creates real active toolkit/boulder state, so rejection scenarios cannot pass merely because a plan is missing. The full adapter gate separately covers plugin construction and wiring.

No dependency or lockfile changes. No user-session state was modified. Windows was not exercised. LSP is unavailable locally; strict source and QA-fixture typechecking passed.

## Automated Checks

`bun run test:senpi` passed under CI-pinned Bun 1.4.2: plugin build and typecheck,
3,110 adapter tests passed with 32 existing platform skips and zero failures;
all ten evidence-resolver tests passed. Tracked bundles were regenerated with
that same runtime. One additional legacy startup fixture was corrected to
include its intended clean assistant outcome after the first full gate exposed it.

## Related Issues

Companion to code-yeongyu/senpi#1520; does not replace the builtin goal fix in code-yeongyu/senpi#1521.

---
## Summary by cubic
Stops `omo-senpi` plan continuation from resubmitting failed or refused turns. Both `ulw-loop` and `ulw-execute-continuation` now record the run at `agent_end` and decide on `agent_settled`, the host edge that guarantees no retry, compaction, or queued continuation will run.

**Behavior**
- Blocks missing outcomes, errors, unfinished tool turns, aborted tool results, host-owned retries, and refusals without consuming continuation slots or dedupe state; every skip is logged with its reason.
- Uses the single `fallback-architect` refusal predicate, extended to admit structured refusals the host demotes to `stop` via its empty-tool-use diagnostic; ordinary assistant text stays continuable.
- Re-reads the recorded event at settle time, so late user aborts during the awaited status probe are honored.
- Keeps successful continuation, caps, deduplication, and footer cleanup unchanged; the `ulw-loop` footer still syncs on blocked outcomes so no spinner lingers.

**Validation**
- Adds regression coverage for both producers across direct and coordinator delivery paths, including settled-edge, late-abort, retry, and compaction cases.
- Adds a live CLI QA driver covering loop and boulder lanes against policy, refusal, retry, and compaction failures; full `omo-senpi` suite passes at 3,280 tests with 3 skips, and resolver tests, strict typechecking, and bundle regeneration pass.

<sup>Written for commit 78010859261c0d25fd02f43532269eba1d2eec91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8027?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (commented):
82/83 local — `session isolation > child-process probe` failed once:

```
(fail) omo-senpi ulw-loop continuation session isolation > #given two independent sessions share one cwd #when the child-process probe runs #then only the owner continues [740.22ms]
```

The other 82 tests across ulw-loop + ulw-execute-continuation pass clean. Could be a timing flake with the child-process probe — worth checking if it repros consistently on CI. The continuation gating logic itself looks correct: error/abort/refusal/retry-owned turns properly blocked, successful turns flow through.

rlaope:

rlaope:
Checked the child-process probe: it does not reproduce. Ran `session-scope.test.ts` five consecutive times on this branch — 7 passed / 0 failed each run, including the `session isolation > child-process probe` case.

Full `test:senpi` on the same tree is also clean: 3,110 passed, 32 pre-existing platform skips, 0 failed (CI-pinned Bun 1.4.2).

Agreed it looks like child-process timing under load rather than the gating logic. I did not touch the probe, since changing a test that passes deterministically here would be guessing at a failure I cannot observe; if CI reproduces it, the timing is worth attacking at the probe's synchronization rather than with a retry.

The branch has since been rebased onto current `dev` (the `omo.js` bundle conflict was resolved by regenerating from source, not by hand-merging the artifact), and the adapter gate was re-run green after the rebase.

code-yeongyu (changes requested):
**Verdict: request changes.** The core idea is right and the direction matches the host, but the gate is placed and defined wrongly in three provable ways: it is hoisted above the `ulw-loop` footer upkeep (a regression I reproduced), it is evaluated once at the top of a handler that then awaits a two-process toolkit spawn while Senpi is free to mutate that very event into an abort (also reproduced), and its terminal-outcome predicate silently disagrees with the two gates the host already ships (`_agentEndAllowsQueuedContinuation`, builtin goal `didAgentEndCleanly`) and with this package's own refusal predicate. The test surface hides this: one of the four new non-table cases still passes with the gate stubbed to `return true`, and the live QA driver disables `retry` and `compaction`, so `willRetry` and every compaction interaction — the two edges the gate exists to respect — were never exercised against the real CLI. All findings below were checked against the source at `750668d`, the pinned host in `node_modules/@code-yeongyu/senpi` (2026.9.9-2), and two throwaway tests I ran in a worktree of this branch (Bun 1.4.2).

1. **`packages/omo-senpi/src/components/ulw-loop/index.ts:90` — the gate skips footer upkeep, not just continuation.** The status probe and `footerStatus.sync(eventCtx, status?.active ?? false)` live at `:106-107`, below the new early return. On any blocked outcome the probe never runs, so the `⚡ ultraworking` footer keeps its 320 ms interval and stays on screen for a run that has already ended (e.g. Esc'd right after the last goal was completed; nothing clears it until the next user input, since `tool_result` also stops arriving). Reproduced: with a fake UI + fake timers, a clean `agent_end` then an `{aborted: true, stopReason: "aborted"}` `agent_end` leaves `probes === 2` and `timers.activeCount() === 1`; deleting line 90 makes that fail, because pre-PR the third probe returned `completeStatus()` and cleared the footer. Fix: keep the gate for delivery only — move `if (!canContinueAfterAgentEnd(payload)) return` to immediately after line 107. State mutation is at `:125-126`, so the PR's stated invariant ("no count/signature change on failure") is preserved, and the footer stays truthful. If one extra toolkit spawn per failed turn is unacceptable, sync the footer from the last known status before returning instead.

2. **`packages/omo-senpi/src/components/ulw-execute-continuation/agent-end-eligibility.ts:4` — `aborted` is read once, before an await window the host can invalidate.** `ulw-loop` awaits `readActiveStatus` at `index.ts:106` (two node startups, per the comment in that file) and only then delivers at `:127`. Senpi awaits `agent_end` handlers (`dist/core/agent-session.js:1980`) and, on a late user abort inside that open boundary, mutates the *same* event object it already dispatched: `event.aborted = true; event.abortSource = "user"` (`dist/core/agent-abort-provenance.js:22-33`). Reproduced: setting `aborted` on the payload from inside `runCommand` still yields one `omo-senpi:ulw-continuation` message. Consequence: pressing Esc while the probe runs still resubmits the plan — the exact bug class this PR closes. Fix: re-evaluate `canContinueAfterAgentEnd(payload)` immediately before `deliverContinuation` in `ulw-loop`; better, record the outcome at `agent_end` and act on `agent_settled`, which is exactly what `src/components/memory/trigger-wiring.ts:100-111` already does in this package and what the host defines `agent_settled` to mean ("no automatic retry, compaction, or queued continuation will run").

3. **`agent-end-eligibility.ts:7,17` — trailing aborted tool results are ignored and bare `toolUse` is accepted.** Both host gates — `_agentEndAllowsQueuedContinuation` (`dist/core/agent-session.js:1497-1526`) and builtin goal `didAgentEndCleanly` (`dist/core/extensions/builtin/goal/continuation.js:20-33`, `isContinuableStopReason` at `:45`) — accept only `stop`/`length` (plus a *malformed* `toolUse`, i.e. one with no `toolCall` content) **and** reject when any `toolResult` after the last assistant is an error matching `/\babort(?:ed)?\b/i`. This gate stops at the last assistant and accepts any `toolUse`. Consequence: a run stopped mid-tool-execution (Esc during a long bash call) ends with a `toolUse` assistant followed by aborted error tool results, passes the gate, and the plan is resubmitted. No test covers a trailing `toolResult`: `terminal-outcome.test.ts:5-19` only pairs `toolUse` with refusal/sensitive `stopDetails`. Fix: mirror the host — scan messages after the last assistant for aborted error tool results, and accept `toolUse` only in the malformed shape (`content` contains no `toolCall`).

4. **`agent-end-eligibility.ts:14-18` — required-compaction turns pass the gate, and the live QA cannot see it.** The host rejects a queued continuation whenever `_getRequiredAutoCompactionReason(lastAssistant)` is set (`dist/core/agent-session.js:1515`), but `_isRequiredCompactionError` requires `stopReason === "error"` (`dist:1574-1578`), so a clean `stop` turn that still needs auto-compaction produces `willRetry: false` and passes this gate. Consequence: at the context threshold — where long ULW runs live — OMO injects a follow-up into precisely the edge the host is holding for compaction. There is no evidence either way in this PR, because `scripts/qa/policy-continuation-e2e.mjs:45-46` writes `retry: { enabled: false }, compaction: { enabled: false }`: neither `willRetry: true` nor any compaction interaction was ever driven through the real CLI, so the "real-Senpi QA" does not cover two of the four blocked classes it claims. Fix: reject on a required-compaction/context-overflow signal too (or move to `agent_settled` per finding 2), and add a live scenario with retry and compaction enabled.

5. **`agent-end-eligibility.ts:11-12` — a second, conflicting definition of "refusal" in the same package.** `src/components/fallback-architect/detection.ts:47-63` mirrors the host's `isClassifierRefusal` (`@earendil-works/pi-ai/dist/utils/stop-details.js`) and carries an explicit comment that the stop reason is checked **first** so stale `stopDetails` can never be read as a refusal; it only admits `error`/`toolUse`. The new gate reads `stopDetails` regardless of stop reason. So for the exact case the PR says it observed live (normalized to `stop`, refusal details retained) the two components disagree: the plan halts while the fallback-architect nudge stays silent, and the user sees a silently stopped plan with no refusal or fallback notice. Note also that the host's real normalization is `demoteToolUseWithoutToolCalls` (`@earendil-works/pi-agent-core/dist/assistant-terminal-state.js:22-34`), which rewrites the stop reason to `stop` and documents the `empty_tool_use_terminal_state` diagnostic as "the only surviving evidence" — the host treats that shape as continuable, not as a refusal. Fix: one predicate. Export the refusal check from `detection.ts` (extending it with the normalized case, keyed on the diagnostic) and import it here, so the adapter has a single refusal definition.

6. **`agent-end-eligibility.ts:3-21` — the only silent skip in either hook.** Every other skip logs its reason (`ulw-loop/index.ts:92,102,112,117,121`; `ulw-execute-continuation/index.ts:68,78,85,92`). A blocked outcome logs nothing, so "my plan stopped continuing" is indistinguishable from "the hook never fired" — for the one skip reason that fires on transient provider failures. Fix: `ctx.logger.info("… skipped", { reason: "terminal-outcome", stopReason, aborted, willRetry })` before returning in both hooks.

7. **`packages/omo-senpi/src/components/ulw-execute-continuation/terminal-outcome.test.ts:115-121` — this case cannot fail.** Mutation check: stubbing `canContinueAfterAgentEnd` to `return true` fails 52 of 64 cases in this file, but "#given an already continued signature #when a failure intervenes #then it does not reset dedupe" still passes in all four variants, because the intervening failure carries the same status/signature and is dropped by `stale-status` / `stale-signature` with or without the gate. Fix: `scenario.advance(1)` before the failure event so an ungated handler would deliver, assert no delivery at that point, then assert the following clean end at that same revision still delivers.

8. **`packages/omo-senpi/scripts/qa/policy-continuation-e2e.mjs:143-159` — the PR's strongest evidence is unrunnable in CI and undiscoverable.** The scenarios execute at module top level with no `import.meta.url` entry guard and no `--self-test`, against the directory's own documented convention (`scripts/qa/AGENTS.md:32`), so merely importing the module spawns real Senpi children and binds sockets. It is referenced by nothing: `test:senpi` runs the plugin build, typecheck, `bun test packages/omo-senpi`, and the evidence-resolver test only, and the driver is absent from the driver table at `scripts/qa/AGENTS.md:9-15`. The "one failed provider call, zero OMO sends" result therefore cannot detect a regression. Fix: add the entry guard and a `--self-test` path, register the driver in that table, and wire the deterministic portion into the senpi QA lane.

9. **`packages/omo-senpi/AGENTS.md:32-33` — the documented contract still describes the old behavior.** Both bullets spell out the continuation preconditions (signature dedupe, cap of 8, reset on user input) and still say the hooks fire on `agent_end` with no outcome precondition. Fix: add the terminal-outcome gate to both bullets so the doc surface matches the shipped behavior.

code-yeongyu:
All nine findings are resolved in `5c76a7c4f6c7013bddc131faa1e5d0ccaf80a853` (rebased onto current `dev`, force-pushed). Both continuation producers now **record** the `agent_end` payload and **decide on `agent_settled`**, which is the fix shape finding 2 called "better" and the only way to settle finding 4 without duplicating host-internal math (see 4). Per finding:

1. **Gate above the footer upkeep** — `ulw-loop`'s `agent_end` handler is now footer upkeep plus outcome recording only; the terminal-outcome gate runs at settle, so a blocked outcome still probes and syncs the footer. New test `footer-status.test.ts` "blocked outcome still syncs the footer" proves it end to end (aborted `agent_end` → footer cleared, timers 0). Mutation: re-inserting the early return above the probe → that test fails (1 fail), restore → 6 pass.

2. **`aborted` read once, before the await window** — the recorded payload is re-read inside the `agent_settled` handler, after the host's mutation window closes (`agent-abort-provenance.js` keeps mutating that exact object until `closeAgentEndBoundary`). Two new tests: a late abort mutating the event between `agent_end` and settle, and a mid-probe abort (mutation fired inside `runCommand`) — no delivery in either. Mutation: moving the decision handler back to `agent_end` → 12 fail including both abort cases; restore → pass.

3. **Contract divergences** — the gate is now the intersection of `_agentEndAllowsQueuedContinuation` and `didAgentEndCleanly` as shipped in pinned `2026.9.10` (verified against the installed module, not the review-era copy): trailing `toolResult`s are scanned after the last assistant and an error result matching `/\babort(?:ed)?\b/i` blocks; `toolUse` is admitted only in the malformed tool-call-less shape; `stop`/`length` stay continuable. Mutations: deleting the trailing scan → 4 fail (the two new trailing-abort cases × both components); deleting the malformed-shape check → 8 fail.

4. **Required-compaction turns pass with `willRetry: false`** — this is why the gate moved to `agent_settled`, whose documented meaning is "no automatic retry, compaction, or queued continuation will run". The QA driver no longer disables anything: `retry: {enabled, maxRetries: 1}` and `compaction: {enabled}`. Two new live scenarios (both lanes, 8 scenarios total, all PASS against the real CLI):
   - **retry**: one retryable 503 → the host retries; recorded trace shows an `agent_end` with `willRetry: true` that never settles (hostOwnedEnds 1, retryOwnedEnds 1), zero sends on that edge, then exactly one continuation after the settle that follows recovery.
   - **compaction**: a 60k window with 95%-of-window reported usage → the host emits `compaction_start {reason:"threshold"}` on its own event stream; nothing rides an unsettled edge, and the one settled continuation the host could not admit was refused by the host itself (`extension_error send_message: "Context remains above the compaction threshold because compaction did not complete"`). Honest residual: an extension cannot compute `_getRequiredAutoCompactionReason` (provider usage vs adaptive reserve), so the host's own admission guard remains the last line for the threshold case — recorded in the driver report as `hostRefusedSends`.

5. **Second refusal definition** — one predicate now: `agent-end-eligibility.ts` imports `isRefusalLikeMessage` from `fallback-architect/detection.ts`, which gained exactly one extension, the `empty_tool_use_terminal_state` demotion diagnostic (mirroring the host's `isMalformedToolUseTurn` reading of `demoteToolUseWithoutToolCalls`). Stop reason is still checked first: a plain `stop` with stale `stopDetails` is continuable (new test) and never a refusal for the nudge either. The live refusal lane asserts the real host leaves the diagnostic on the demoted message and that it is what the gate keys on. Mutation: deleting the clause → 9 fail (detection + normalized-refusal cases).

6. **Silent skip** — both hooks log `skipped {reason: "terminal-outcome", blockedBy, stopReason, aborted, willRetry}` before returning. Mutation: renaming the reason → 69 fail across the blocked-outcome table and footer test.

7. **The case that could not fail** — rewritten per your prescription: `advance(1)` before the failing event, assert no delivery, then the clean end at the same revision still delivers. Stubbing the gate open (`return {blockedBy: null}`) now fails **73 of 109** cases in `terminal-outcome.test.ts` including `#given an already continued signature #when a failure intervenes #then it does not reset dedupe` (verified in the mutation log).

8. **Undiscoverable, unrunnable driver** — `policy-continuation-e2e.mjs` now has an `import.meta.url` entry guard, a `--self-test` (pure: scenario matrix, audit invariant, and the pinned host classifiers — the retryable text is pinned against `pi-ai`'s own `isRetryableErrorMessage`), per-phase trace slicing, and `--lane/--failure` filters. The deterministic half is `policy-continuation-e2e.test.mjs`, which `bun test packages/omo-senpi` — i.e. `test:senpi` and the CI step — already runs; the driver is registered in the `scripts/qa/AGENTS.md` lane table and commands. Mutation: deleting the unsettled-edge audit condition → 2 fail / restore 3 pass.

9. **`AGENTS.md:32-33`** — both bullets now describe the settled-edge contract: what `agent_end` does (footer + recording), where the decision runs, the host-aligned continuable shapes, the single refusal predicate, and the observable skip.

Verification (Bun 1.4.2, Linux): `bun test packages/omo-senpi` → **3280 pass / 0 fail / 3 skip**; resolver suite 10 pass; `tsgo --noEmit` clean; `build-extension.mjs --check` clean with the bundle rebuilt on Linux and committed; live QA 8/8 scenarios PASS with sandbox isolation and credential digests unchanged. Ten targeted mutations (delete/stub the line under test, run, restore, re-run) all produced red-then-green; summary above per finding.

code-yeongyu (approved):
All findings from review 5167578502 resolved and mutation-proved (10/10 RED then GREEN), omo-senpi 3280 pass/0 fail, tsgo clean, live QA 8/8 against the real CLI. Rebased onto current dev; the FlushScheduler fixture updated for the canceller-returning signature #7933 introduced. Approving.

PR #7988, opened 2026-09-08
fix(git-bash): explain deferred Codex tool discovery (fixes #6077)

MoerAI:
## Summary

Extend the Windows Git Bash reminder with conditional code-mode discovery guidance: inspect `ALL_TOOLS` through `exec`, discover the actual Git Bash tool names, and invoke those names through the `tools` object inside `exec` before treating the MCP as unavailable.

## Root Cause

The reminder previously stated a preference and fallback but did not explain deferred discovery. An abbreviated initial tool list could therefore be interpreted as tool unavailability, as reported in #6077.

## Changes

| File | Change |
| --- | --- |
| `packages/omo-codex/plugin/components/git-bash/src/codex-hook.ts` | Add the missing conditional discovery guidance; keep gating and marker behavior unchanged. |
| `packages/omo-codex/plugin/components/git-bash/test/codex-hook.test.ts` | Keep all seven behavioral cases, replacing authored-prose assertions with parsed hook fields. |

This is a prose-only reminder change. It adds no new wording tests and does not alter MCP registration or tool dispatch.

## Reproduction and verification

Before editing, the actual Windows-guarded hook emitted only the existing preference and fallback, with no deferred-discovery guidance.

- Component typecheck and build passed; component tests: **7 pass, 0 fail**.
- With Node 24.18.0 and Bun 1.4.0, `bun run test:codex` passed, including the LSP, ULW-loop, installer, notice, and plugin gates.
- Real Codex 0.144.6 executed `exec_command`, generated matching first-party `PreToolUse` start/completion notifications, and delivered the built component's context into the following local-model request.
- A subsequent code-mode probe discovered the actual `which_bash` name in `ALL_TOOLS` and invoked it through `tools[name]` inside `exec`. It used the production MCP handler with a Linux launcher adapter.
- The command succeeded and one reminder marker was created. The disposable container had networking disabled, isolated HOME/CODEX_HOME, and no host credentials/configuration mounts.

Evidence and captured machine fields:
`.omo/evidence/20260908-6077-deferred-git-bash/README.md`.

## Limits

The live test selected the existing Windows guard with synthetic `OS=Windows_NT`; it does not claim native Windows testing. It proves component reminder delivery and deferred invocation in that fixture, not probabilistic model choice or the native Windows installed-hook path. The component cache in this earlier probe was manually assembled.

[...]

PR #8028, opened 2026-09-09
fix(omo-senpi): suspend native workflows across session switches

### LINKED ISSUE #8020: Session switching cancels active workflows and same-host recovery rejects its own lease
## Bug Description

Switching away from a session with an active native OMO workflow can permanently cancel its DAG instead of suspending it for recovery. Returning in the same host process also encounters a lease check that rejects its own previous PID.

This is about **selecting another session through `/resume`**, not `/session` (information only) or opening and canceling the selector. Those display-only paths do not call runtime teardown.

## Steps to Reproduce

1. Start a workflow with a completed node, a running node, and a pending dependent.
2. Select another session through `/resume`, then return to the original session.
3. `wireDagLifecycle()` invokes `DagRuntime.detach()` on `session_before_switch`.
4. `detach()` calls `scheduler.cancel(runId, "runtime detached")`, making the unfinished nodes and run `cancelled` before shutdown recovery can pause them.

An isolated probe against production scheduler functions confirmed that completed work remains completed while running/pending nodes and the run become cancelled. A separate probe against current and installed recovery functions confirmed that a paused own-session run with `previousLeaseHolderPid === hostPid` is skipped as `live_lease`, while a dead predecessor can be claimed. Full live selector reproduction is still being prepared; these probes are not presented as TUI proof.

## Expected Behavior

Viewing/canceling session selection leaves work alone. An actual switch safely suspends resumable workflow ownership, and returning restores progress without duplicate node execution. Explicit workflow cancellation must remain terminal, and a different live process's lease must remain protected.

## Actual Behavior

The detach path cancels the workflow, and own-session recovery rejects even its own live predecessor PID. Simply removing cancellation would be unsafe without quiescing and fencing the outgoing scheduler.

## Doctor Output / Environment

Native `omo-ai 5.0.0-0.beta.49`, Senpi `2026.9.8`, macOS arm64. OpenCode doctor/version do not describe this native surface and were not run. Latest OMO source inspected: `b5122f19d`.

## Additional Context

Relevant files:
- `packages/omo-senpi/src/components/task/index.ts` (`wireDagLifecycle`)
- `packages/omo-senpi/src/components/task/dag-runtime.ts` (`detach`, `pauseForShutdown`)
- `packages/senpi-task/src/dag/recovery.ts` (`claimPausedRun`)

I am preparing a failing-first fix and isolated live lifecycle proof. Ordinary ephemeral terminal jobs and persistent-monitor respawn have different contracts; this report does not claim arbitrary external processes survive a session replacement.

rlaope:
## Summary

Preserve native workflow progress across an actual session switch and safely recover it when returning in the same host process.

Previously, `session_before_switch` cancelled active DAGs before shutdown could pause them. Recovery then rejected even an explicitly released lease held by the same process. Opening `/session` information or opening/canceling `/resume` does not cause teardown; those display-only production paths remain unchanged.

Closes #8020.

## Changes

- Move DAG retirement from vetoable pre-switch notification to committed shutdown. Await scheduler quiescence before persisting pause and suspending child sessions.
- Add single-shot scheduler suspension that stops admission/settlement/promotion writes, drains in-flight admission and journal delivery, and retires subscriptions without cancelling child records.
- Reclaim an explicitly released own lease while continuing to reject active self claims and live foreign owners.
- Reuse recovery-created controllers for subsequent control operations. Explicit workflow cancellation remains destructive, including cancellation of a recovered child.
- Add lifecycle/lease regressions and reusable real Senpi RPC/PTY QA drivers.

## QA & Evidence

- **Failing first:** returning through a released same-process lease was skipped instead of resumed; committed shutdown cancelled instead of paused. A second RED case caught cancellation missing the recovered child.
- **Focused regressions:** 141 tests passed, including admission quiescence, late completion fencing, output reuse, one-time dependent admission, active-self/foreign leases, switch veto, and explicit cancellation.
- **Lead real RPC A -> B -> A:** the run paused without cancellation, reused completed output, kept the running child's task/session identity, advanced its resume epoch, and admitted the pending dependent exactly once. Deliberate workflow cancel still cancelled both child and run.
- **Lead real PTY:** `/session` rendered information; `/resume` opened the selector; Escape restored the editor. Workflow snapshots stayed identical and no before-switch/shutdown event occurred until final cleanup.
- **Screen:** the actual PTY stream was replayed through xterm.js in Chromium and visually inspected. The selector screenshot captures its initial loading state; selecting an existing row is covered through the separate real runtime RPC replacement path, not claimed from the empty PTY list.
- **Isolation/cleanup:** test children exited, sandboxes were removed, and real credential digests remained unchanged.

Reproduce:

```sh
node packages/omo-senpi/scripts/qa/session-dag-resume-qa.mjs
node packages/omo-senpi/scripts/qa/session-dag-resume-qa.mjs --pty
```

Source/QA evidence: `.omo/evidence/omo-senpi-adapter/20260909-session-dag-suspend-resume/`.
Independent lead receipts: `.omo/evidence/omo-senpi-adapter/20260909-session-dag-lead-qa/`, including `live-report.json`, `pty-report.json`, and `selector-screen/terminal.png`. Both adapter and engine typechecks passed.

## Risks & Residuals

This is **suspend/resume**, not continuous execution of the outgoing runtime. It covers native DAG/task ownership with child resumption enabled. Arbitrary shell/PTY jobs and externally launched controller stacks have separate lifecycle contracts and are not made durable by this patch.

The existing recovery attachment waits for recovered scheduling to finish; startup responsiveness during that barrier is unchanged. Explicit `task.resume_children=false` behavior is also unchanged.

The live test loads the real source task component and real in-process child runner against a loopback provider, not every unrelated OMO component. Windows and real model services were not exercised. LSP tooling is unavailable locally; strict compiler checks were used.

## Automated Checks

Full `test:senpi`: 3,048 adapter tests passed, 32 existing platform skips, zero
failures; ten evidence-resolver tests passed. All 313 DAG-engine tests passed.
Final bundles were regenerated with CI-pinned Bun 1.4.2, followed by all 141
focused lifecycle tests passing. No assertion was skipped or weakened.

## Related Issues

Closes #8020. Separate from terminal policy-rejection recovery in code-yeongyu/senpi#1520.

---
## Summary by cubic
Preserves native DAG workflow progress across a real session switch and resumes it safely in the same host process. Previously, the vetoable pre-switch hook cancelled active DAGs and recovery rejected even released same-process leases; now runs pause on committed shutdown, and released own leases can be reclaimed. `/session` info and `/resume` selector cancellation are unchanged.

**Changes**
- DAG retirement moves from `session_before_switch` to committed shutdown; `pauseForShutdown` awaits scheduler suspension and deregisters controllers before persisting the pause, so `detach()` finds nothing to cancel.
- Single-shot scheduler suspension drains in-flight admission and settlement without cancelling child records; released own leases are reclaimable while active self claims and live foreign holders stay protected.
- Recovery reuses its created controllers for later control operations; explicit workflow cancellation remains destructive.
- Adds failing-first lifecycle tests and real Senpi RPC/PTY QA drivers covering same-host resume and `/session` plus `/resume` cancel.

Closes #8020.

<sup>Written for commit bda423127da06cab8144a725eceb8b94d560ff14. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8028?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
839/839 pass (task + DAG engine). Comprehensive suspend/resume implementation — scheduler quiescence, explicit lease release/reclaim, recovered controller reuse. Clean separation of display-only paths vs committed shutdown. Closes #8020. LGTM.

rlaope:

code-yeongyu (changes requested):
Requesting changes: this PR now conflicts with `dev` in a way that cannot be resolved mechanically, and part of it has been superseded.

**Superseded — please drop from this PR**
- `packages/senpi-task/src/dag/recovery.ts`: your `releasedBySelf` check (`fresh.leaseHolderPid === undefined && fresh.previousLeaseHolderPid === context.hostPid`) targets the same bug as `dev` commit 9fbe6cad1 (#8029, merged 2026-09-10), which landed a more general fix: `holderStillLive(context, runId, holderPid)` — for our own pid it consults `isRunHeldInProcess(runId)` instead of a signal-0 probe, so a fresh runtime reopening the session in the same process claims, while the cross-host liveness fence is kept for every other pid.
- `packages/omo-senpi/src/components/task/dag-runtime-lease-recovery.test.ts`: `dev` rewrote this around `FOREIGN_HOST_PID` and a `predecessor === "foreign-host"` branch; your `process.pid + 1` variant conflicts head-on.

**Not superseded — this is the valuable part, please keep it**
- Moving DAG retirement out of the vetoable `session_before_switch` notification into committed shutdown, with scheduler quiescence awaited before teardown (closes #8020). Nothing in `dev` covers that.

**Suggested path**
1. Rebase onto current `dev`.
2. Take `dev`'s side wholesale for `recovery.ts` and `dag-runtime-lease-recovery.test.ts`.
3. Keep only the switch/shutdown lifecycle change plus its own tests.
4. Regenerate the committed plugin bundles after the rebase — CI's `Verify committed Senpi plugin bundle is current` step runs `node packages/omo-senpi/plugin/scripts/build-extension.mjs --check` and fails with `stale-output` otherwise. The bundle conflict in `packages/omo-senpi/plugin/extensions/omo-task.js` should be resolved by rebuilding, never by hand-merging.

I merged the rest of the approved queue today (#7986 #7988 #8012 #8013 #8014 #8015 #8017 #8021 #8029 #8035 #8036 #8037 #8038) by merging `dev` in and regenerating bundles, but I am not resolving this one for you: choosing between two different implementations of the same lease-recovery invariant is an author decision, not a mechanical merge.

Altairpaca (commented):
Reviewed the session-switch lifecycle seam, specifically `wireDagLifecycle`, `DagRuntime.pauseForShutdown`, scheduler suspension fencing, and same-host lease reclaim. I don't see a blocker in those paths.

One ordering point I checked explicitly: `wireDagLifecycle` now has an async `session_shutdown` handler (`await pauseForShutdown(); detach()`) followed by the existing `dispose()` handler. Senpi's `ExtensionRunner.emit()` iterates handlers serially and `await`s each one, so `dispose()` cannot race ahead of the suspend/drain step within the same extension generation.

The suspension shape also looks internally consistent: it flips `suspended` before draining admission, wakes the frontier through the cancellation-request promise without marking the run cancelled, fences post-admission settlement/promotion, waits for journal delivery, and only then releases/persists the paused ownership. The recovery exception is limited to an explicitly released self lease (`leaseHolderPid === undefined && previousLeaseHolderPid === hostPid`), so a live foreign owner and an unreleased active self lease remain protected.

I only reviewed this lifecycle/concurrency slice, not claiming a full 17-file re-review.

rlaope:
Requested changes applied. Rebased onto current `dev` (`826d81902`) and took `dev`'s side wholesale for the superseded work.

**Dropped (superseded by `dev`):**
- `packages/senpi-task/src/dag/recovery.ts` and `recovery.test.ts`
- `packages/omo-senpi/src/components/task/dag-runtime-lease-recovery.test.ts`
- `packages/omo-senpi/src/components/task/dag-runtime.ts` and `dag-runtime.test.ts` — these ended up superseded too. My earlier version replaced `dev`'s admission-latch mechanism, and once rebased that broke four contracts `dev` already guarantees (the same-process retry un-latch, the detach/session-switch ownership case, the shutdown-pause resume, and the live-foreign-pid fence). Rather than reshape `dev`'s tests around my approach, I took `dev`'s runtime as-is. The lease/`isRunHeldInProcess` semantics documented in the component AGENTS.md are now untouched by this PR.

**Kept — the valuable part (closes #8020):**
- `packages/omo-senpi/src/components/task/index.ts`: DAG retirement moves out of the vetoable `session_before_switch` into committed `session_shutdown`, awaiting scheduler quiescence before teardown.
- `packages/senpi-task/src/dag/scheduler.ts`: the new `suspend()` that retires a controller without cancelling its children, so no lease is released while an admission batch or queued journal delivery can still write, plus its tests in `scheduler.test.ts`.

**Bundles:** regenerated via `plugin/scripts/build-extension.mjs`, never hand-merged. `build-extension.mjs --check` reports current.

**Verification:** `bun test packages/omo-senpi/src/components/task packages/senpi-task/src/dag` — 872 pass / 0 fail. Full `bun run test:senpi` gate green.

code-yeongyu:
Rebased onto current `dev` (`b03c5796b`) and addressed every finding. Branch head is now `bda423127`.

| # | Finding | Resolution | Commit |
|---|---|---|---|
| 1 | **Superseded — drop** `packages/senpi-task/src/dag/recovery.ts` `releasedBySelf` | Gone. `dev`'s `holderStillLive(context, runId, holderPid)` from 9fbe6cad1 (#8029) is the only lease-recovery implementation on the branch. Verified: `git diff origin/dev -- packages/senpi-task/src/dag/recovery.ts` is **empty**, and `git grep releasedBySelf` matches nothing. | rebase |
| 2 | **Superseded — drop** `dag-runtime-lease-recovery.test.ts` `process.pid + 1` variant | `dev`'s side taken. The file keeps `FOREIGN_HOST_PID = 2_147_483_647` and the `Predecessor = "foreign-host" \| "same-host"` branch; no `process.pid + 1` remains. The only deltas vs `dev` are the `await`s that the now-async `pauseForShutdown` requires, plus the two same-runtime tests updated for the new retire semantics (see #4). | rebase + `2518df235` |
| 3 | **Keep** the switch/shutdown lifecycle change | Kept, and **made real** — see below. | `0c04724be` + `2518df235` |
| 4 | **Regenerate the committed bundles**; never hand-merge `omo-task.js` | The bundle conflict was resolved by taking one side and rebuilding on **Linux** with CI-pinned **Bun 1.4.2** — never hand-merged. `node packages/omo-senpi/plugin/scripts/build-extension.mjs --check` reports **"extension build is current"** on the pushed tree. | `bda423127` |

### The lifecycle change did not actually work as submitted

Reviewing finding #3 turned up a blocking defect in the part you asked me to keep. `DagScheduler.suspend()` was added and unit-tested but **never called from production**: `DagRuntime.pauseForShutdown` was still synchronous, so `await runtime.pauseForShutdown()` in `wireDagLifecycle` awaited nothing, and the `runtime.detach()` that runs immediately after still called `scheduler.cancel(runId, "runtime detached")` on every live scheduler. `cancelRun` only early-returns for `completed|failed|cancelled` — **`paused` is not terminal** — so the committed shutdown persisted the pause and then cancelled straight over it. That is the exact #8020 teardown the PR set out to remove.

`2518df235` wires it up: the committed shutdown now retires the live schedulers, awaits both the suspension and the frontier loop's exit before persisting the pause, and deregisters them so `detach()` finds nothing to cancel. Deregistration also makes the #8029 in-process probe (`isRunHeldInProcess`) report the run unheld, so recovery in the same process reclaims it instead of skipping on a stale live lease.

### Mutation proof

Every assertion added or changed was proven by deleting the production line, observing the failure, restoring, and observing the pass.

| Mutation | Result |
|---|---|
| Delete `await Promise.all(retiring.map(([, owned]) => owned.scheduler.suspend()))` | **3 fail** — both same-runtime lease-recovery tests and `retry-after-pause`, all timing out because the run is never retired |
| Delete the deregistration loop | **1 fail** — `#given a control verb re-registered a controller after the shutdown pause ...`, because `detach()` cancels the paused run |
| Restore both | **17 pass / 0 fail** |

`await Promise.allSettled(... owned.running)` is retained but **no assertion isolates it**; it drains `ensureScheduled`'s unguarded `.finally(delete)` so it cannot evict a controller a later control verb re-registers. That is stated honestly in a code comment rather than claimed as covered.

Three `dev` tests encoded the old soft-pause semantics (`pauseForShutdown` left the scheduler live, so re-attach was a fenced no-op). Under real retirement the reopen reclaims the run, so they now await the reclaim instead. A probe confirmed the reclaim **reattaches to the durable child** — `handles=1`, no second admission, no lease watch armed, no own-pid deferral warning — so no assertion was weakened; `#then no second scheduler admits the node` still asserts `toHaveLength(1)`.

### Test results (all on Linux/x86_64, Bun 1.4.2, on the pushed tree `bda423127`)

```
bunx tsgo --noEmit -p packages/omo-senpi/tsconfig.json     -> clean
bunx tsgo --noEmit -p packages/senpi-task/tsconfig.json    -> clean
    -> "omo-senpi extension build is current"
bun test --timeout 20000 packages/omo-senpi packages/senpi-task
    -> 5367 pass, 4 skip, 3 fail   (5374 tests, 683 files)
```

The 3 failures are **pre-existing on `dev` and unrelated to this PR** — all in `packages/omo-senpi/src/skills-sync.test.ts`, from the #8085 role rename leaving synced skill files stale. Pristine `origin/dev` `b03c5796b` on the same machine: **5364 pass, 4 skip, 3 fail** — the identical three. This branch therefore adds 3 passing tests and introduces zero failures.

The contributor's commit `0c04724be` is preserved as authored; my changes are added on top.

code-yeongyu (approved):
Review findings addressed and verified; rebased onto current dev with the generated extension bundles regenerated on Linux. Approving.

PR #8029, opened 2026-09-09
fix(senpi-task): claim a paused DAG run whose previous host pid is our own instead of waiting on it

### LINKED ISSUE #8006: [Bug]: Same-PID session reopening leaves paused DAGs waiting for their current host to exit
### Prerequisites
- [x] I have searched existing issues to avoid duplicates.
- [x] I am using the latest version of oh-my-openagent.
- [x] I have read the documentation or asked an AI coding agent with this project's GitHub URL loaded and couldn't find the answer.

**Version qualification:** Native beta.49 (Senpi 2026.9.8) is currently installed and matches the published beta channel. The incident below was observed on beta.48 / Senpi 2026.9.7-2; it has **not** been reproduced on beta.49. This is a standalone OmO Native report, not an OpenCode-plugin report.

### Bug Description
Reopening a saved parent session inside the same multi-session RPC host left two DAGs paused, waiting for the previous host PID to exit even though that PID was the current recovering host. Child tasks reattached and completed, but their DAG nodes remained running.

### Steps to Reproduce
Observed incident sequence (not a deterministic reproducer):
1. Run a saved parent with two active DAGs in a multi-session RPC host over a Unix socket.
2. The logical parent session shuts down while the shared host process remains alive.
3. Both DAGs persist dag.run.paused with reason session_shutdown and previousLeaseHolderPid 80553.
4. Reopen the same saved parent in the same host, PID 80553.
5. Child verifier/reviewer tasks reattach and complete, but both DAGs remain paused waiting for PID 80553 to exit.
6. Workflow attach and idempotent start do not resolve the state.

Suggested regression (not executed): gate child completion, dispose only the parent runtime, await ownership release, reopen the saved parent in the same host, and release the child gate. Verify reconciliation/resumption, with a competing-owner control to retain duplicate-scheduler fencing.

### Expected Behavior
After the previous session runtime releases ownership, the same parent should safely reacquire paused DAGs within the surviving host, reconcile terminal child outcomes, and resume normal scheduling. A still-active owner must remain fenced. This does not request deleting leases, blindly ignoring live owners, or rerunning completed children.

### Actual Behavior
| Observation | UTC, 2026 |
|---|---|
| Host PID 80553 started | September 8, 12:32:13 |
| First DAG paused | September 8, 19:03:24.999 |
| Second DAG paused | September 8, 19:03:25.142 |
| Verifier completed | September 8, 23:59:57.880 |
| Reviewer completed | September 9, 00:02:59.571 |

Both checkpoints were paused, lacked a current leaseHolderPid, and retained previousLeaseHolderPid 80553. The same host remained alive. Relevant children were completed while their DAG nodes remained running; neither journal contained a subsequent resume event.

Process liveness alone does not distinguish a released session-runtime owner from its replacement inside the same shared process. A web client can reconnect and reopen the session but cannot finish DAG recovery when recovery waits for its own host to die. Restarting the whole shared host would affect unrelated sessions.

### Doctor Output
Current **Native-harness** doctor output, collected after the incident; only the executable location is redacted:

```text
PASS senpi CLI: <Senpi CLI>
PASS senpi version 2026.9.8
INFO omo 5.0.0-0.beta.49 (engine: senpi 2026.9.8)
```

Template diagnostic applicability: the separately installed oh-my-openagent CLI resolves to 4.16.3 (OpenCode harness), not the affected Native harness. Its doctor was attempted through the existing installation under enforced read-only execution and exited 1 with EPERM during directory initialization. The Native output above is **not** claimed to be output from bunx oh-my-openagent doctor. No package upgrade/download or runtime restart was performed for these diagnostics.

### Error Logs
Sanitized state summary, not a fabricated verbatim log:

```text
dag.run.paused
reason: session_shutdown
previousLeaseHolderPid: 80553
current recovering host PID: 80553
DAG status: paused
relevant node state: running
corresponding child task status: completed
```

Runtime recovery messages explicitly said the runs would remain paused while previous host PID 80553 remained alive.

### Configuration
Standalone OmO Native / Senpi; multi-session RPC over a Unix socket; two DAGs of one saved parent reopened within the same host. No lease deletion, journal editing, or ownership bypass. A minimal configuration reproducer is not yet established.

### Additional Context
Incident environment: OmO 5.0.0-0.beta.48, Senpi 2026.9.7-2, Bun 1.4.2, macOS arm64.

The initiating shutdown cause is unknown. A provider-connection-lost message was observed, but transport cleanup is only a candidate. This report concerns recovery after logical session shutdown, independently of its trigger.

Related issues checked: #7859 (startup waiting after recovery), #7866 (foreign-session adoption), #7902 (never-started tasks after process death). None appears to cover waiting for the recovering host's own PID. Beta.49 notes do not establish whether this particular case is fixed; latest-version reproduction remains unverified.

### Operating System
macOS

### OpenCode Version
N/A — standalone OmO Native on Senpi, not OpenCode.

PeterPonyu:
## Summary

Fixes #8006. The DAG resume lease identifies its holder by OS pid only, and `claimPausedRun` (`packages/senpi-task/src/dag/recovery.ts`) decides "holder still live" with a signal-0 probe. Reopening a saved session in the same multi-session host creates a fresh runtime with the same `parentSessionId` in the same process, so the probe on our own pid is vacuously true, the run is skipped as `live_lease`, and dag-runtime arms a lease watch that can only fire when the host observes its own death — the paused DAG waits forever for "the current host to exit". Reproduced on `dev@b5122f19d`: own-session claim → `skipped/live_lease/holderPid=<own pid>`; the same record through the fork-source adoption path → `adopted`; a watch on our own pid → 50 polls, 0 fires.

#7905 introduced the pid fence and deliberately left the own-session path without a self-exemption to protect the same-runtime pause + re-attach case. This keeps that protection but implements it as "own pid ⇒ ask whether *this runtime* still schedules the run" rather than "own pid ⇒ live". The fork-source path and the control verbs already exempt `holder === hostPid`.

## Changes (+27/−3 production, `recovery.ts` + `dag-runtime.ts` + two `AGENTS.md` notes)

- `recovery.ts`: `DagRecoveryOptions.isRunHeldInProcess?: (runId) => boolean`, threaded through `RecoveryContext` like `stopAdmission`/`reattach`. `claimPausedRun` judges a holder pid equal to `hostPid` by that predicate (default `false`, matching `claimOrphanedRun` and `node-control-context.ts`) and every other pid by `isProcessAlive` exactly as before.
- `dag-runtime.ts`: passes `isRunHeldInProcess: (runId) => schedulers.has(runId)` — the map is per-runtime, so a disposed predecessor runtime in the same process never fences its successor, while a runtime that paused and re-attaches itself still refuses to build a second scheduler. `watchLiveLease` no longer arms a watch when the holder is our own pid (it could never fire, and its "resuming once that pid is gone" warning is the message the reporter quoted).
- Unchanged on purpose: the foreign-pid `isProcessAlive` fence and its lease watch (two live host processes must never both schedule a run), `pauseRunsForShutdown` writing `previousLeaseHolderPid`, `resumeClaimedRun` / `releaseLease`, the checkpoint schema.
- `omo-task.js` regenerated with bun 1.4.2 (the recovery protocol is bundled into it).

## Test plan

- [x] RED on `dev` → GREEN: `recovery.test.ts` own-pid holder with an always-alive probe and no in-process holder → `resumed` (was `skipped/live_lease`); `dag-runtime-lease-recovery.test.ts` same-host reopen after `pauseForShutdown()` + `dispose()` → attach claims and completes, `dag.run.resumed` journaled, no lease watch armed (`timers.pending() === 0`), no self-pid warning.
- [x] Fences pinned, green before and after: own-pid holder with `isRunHeldInProcess → true` stays `live_lease`; same runtime pause + re-attach keeps exactly one scheduler; foreign live pid never claims.
- [x] Two fixtures that pinned the bug were made honest: the `pausedHandoffFixture` handoff tests now stamp a genuinely foreign predecessor pid (`2_147_483_647`, the suite's existing convention) instead of relying on our own pid reading as "alive", and the `dag-runtime.test.ts` "another adapter never claims" case does the same, so all keep their cross-host meaning.
- [x] `bun test packages/senpi-task/src/dag`: 310 pass / 0 fail. `packages/omo-senpi/src/components/task/`: 560 pass / 0 fail. Whole `packages/senpi-task`: 1978 pass / 1 pre-existing skip. `tsgo --noEmit` clean for both packages. The live handoff driver `packages/omo-senpi/scripts/qa/dag-lease-handoff-qa.ts` (real spawned predecessor, default probe) reports PASS, exercising the untouched foreign-pid branch.

## Not covered / follow-ups

- `manager.ts` `liveLeaseHolder` still treats our own pid as "another live lease" for `amend` on a reopened session; `amend` already replaces the scheduler, so the one-line self-exemption there is a candidate follow-up kept out of this PR to stay scoped to resume.
- `schedulers.has` is per-runtime-closure, not per-process: if a host ever kept two *live* runtimes for the same session, the second would claim while the first schedules. Today the host shuts a session down before reopening it, and a disposed runtime's scheduler is admission-latched with idempotent terminal folds. A process-global `runId → runtime nonce` registry would harden this and can land separately.
- A very fast close → reopen where the predecessor's task lifecycle is still draining may fold a `task_interrupted` instead of re-attaching; user-driven reopens are seconds to hours apart (the report's was ~5 h).

---
## Summary by cubic
Fixes #8006: reopening a saved session in the same host process no longer leaves its paused DAG runs waiting forever on our own pid. The lease holder was judged live via signal-0, which is always true for our own pid, so the run was skipped as `live_lease` and a watch armed on our own pid that could only fire once the process exited. Now an own-pid holder is judged by whether this runtime still schedules the run, so the reopen claims and resumes it.

**Bug Fixes**
- `recovery.ts` takes `isRunHeldInProcess(runId)` and consults it only when the holder pid is this host's own pid (default: not held); foreign pids keep the signal-0 liveness fence.
- `dag-runtime.ts` wires it to `schedulers.has(runId)`: a disposed predecessor runtime in the same process never fences its successor, while a same-runtime pause + re-attach still refuses a second scheduler.
- `watchLiveLease` no longer arms a watch or logs the deferral warning for a holder pid equal to our own.
- The shipped `omo-task.js` bundle was regenerated with bun 1.4.2 so the recovery protocol change ships.

<sup>Written for commit 0c4afb94e2db40aafd7cefc2d344b4556ec3c053. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8029?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
284/284 DAG engine pass. Direct fix for #8006 — same-PID paused run now claimed instead of waiting for self to exit. Clean complement to #8028's suspend/resume. LGTM.

PR #7463, opened 2026-08-28
fix: preserve assistant model variants for continuations

AceRothstein71:
## Summary
- Preserve flat assistant-message variants when todo continuation rebuilds model metadata.
- Preserve the same variant through SDK message lookup.
- Add direct regressions for both message shapes.

## Validation
- `bun test packages/omo-opencode/src/hooks/todo-continuation-enforcer/resolve-message-info.test.ts packages/omo-opencode/src/features/hook-message-injector/injector.test.ts packages/omo-opencode/src/features/hook-message-injector/sdk-message-lookup-error-envelope.test.ts`
- `bun run typecheck`
- `bun run build`
- Isolated Docker OpenCode live-plugin QA recorded under `.omo/evidence/20260828-fix-7462-variant-cache/`

Fixes #7462

Ultraworked with [Sisyphus](https://github.com/code-yeongyu/oh-my-openagent)

---
## Summary by cubic
Stops todo continuation from dropping the model variant on flat assistant messages, so resumed tasks keep the same model variant.

- Preserves flat `variant` when rebuilding model metadata during continuation.
- Keeps the variant through SDK message lookup when the nested variant is absent.
- Adds regression tests for both message shapes.

Fixes #7462.

<sup>Written for commit 9cc761be84b3c9966f69901dfe486f367bbbc340. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/7463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #8035, opened 2026-09-09
fix(omo-opencode): preserve the flat assistant variant in the sibling continuation injectors

### LINKED ISSUE #7462: [Bug]: todo-continuation-enforcer drops model variant, breaking prompt cache (injected turns miss ~137K cached tokens, 99%→2%)
## Prerequisites

- [x] I will write this issue in English (see the Language Policy)
- [x] I have searched existing issues to avoid duplicates (searched `variant`, `reasoning_effort`, `cache hit`, `cache miss`, `todo continuation`; closest related are #6981 headless `-p -c` full cache-miss via `registry_miss`/flatten, and #6881 volatile memory metadata in system prompt breaking prefix cache — same family, different seam)
- [x] I am using the latest version of oh-my-openagent (4.19.4 confirmed)

## Bug Description

`todo-continuation-enforcer` injections lose the model **variant**, so any auto-continuation turn (e.g. a TODO CONTINUATION directive injected when incomplete todos remain) is sent to the provider **without `reasoning_effort` / `verbosity`**. Since provider prompt caching keys on the full effective prefix, one missing variant field invalidates the entire ~137K-token cached prefix: the continuation turn drops from ~99% cache hit to ~2%, paying full price for the whole conversation again.

### Observed data — same session, consecutive requests, same model, byte-identical message prefix

| request | source | prompt | cached | hit % | `reasoning_effort` | `verbosity` |
|---|---|---|---|---|---|---|
| 2940–2945 | user-initiated | ~137–140K | ~137–139K | **98–99%** | `max` | `medium` |
| **2946** | **TODO-continuation injection** | 140,331 | **3,072** | **2.2%** | **absent** | **absent** |
| 2947–2952 | continuations | 140–153K | — | 95–98% (new warm cache) | absent | absent |
| 2954 | user-initiated again | 154,404 | 139,264 | 90.2% | `max` (restored) | `medium` (restored) |

Decisive facts:

1. **The 131-message prefix of request 2946 is byte-identical to request 2945** (verified including every `reasoning_content` and `tool_calls`). The only body-level difference is the top-level `reasoning_effort` / `verbosity` fields being absent.
2. Request 2946 was constructed by `todo-continuation-enforcer` (its last message is `[SYSTEM DIRECTIVE: OH-MY-OPENCODE - TODO CONTINUATION]`).
3. When 2954 returned to `reasoning_effort=max, verbosity=medium`, the cache instantly re-matched the **old** prefix at exactly 139,264 tokens — the same value 2945 had cached. The provider cache never expired; the injected turn just asked with a different effective key.

## Root cause

The injection path resolves the model from the last real message via `resolveLatestMessageInfo` (`packages/omo-opencode/src/hooks/todo-continuation-enforcer/resolve-message-info.ts`):

```ts
model: info.model ?? (info.providerID && info.modelID ? { providerID: info.providerID, modelID: info.modelID } : undefined),
```

opencode 1.18.25 stores **user and assistant messages asymmetrically**:

- user message: `"model": {"providerID","modelID","variant":"max"}` — nested, variant inside `model`
- assistant message: `"model"` **absent**, flat fields `modelID` / `providerID` / `variant` at top level

When the loop scans back from the tail, the last real message is an **assistant** message (`agent` is truthy, so it returns immediately), `info.model` is null, and the fallback constructs `{providerID, modelID}` — **variant is dropped**. `launchVariant = model?.variant` in the injector then evaluates to `undefined`, the `...launchVariant ? { variant: launchVariant } : {}` spread is omitted, and the upstream request carries no variant → `reasoning_effort`/`verbosity` never materialize.

Note the top-level `info.variant = "max"` is present on that very assistant message — the code just never reads it.

This is the same family as #6981 (continuation path reconstructs an effectively different request → full cache miss) and #6881 (prefix instability breaks provider caching), but the seam here is different: variant loss during `todo-continuation-enforcer` model resolution, not registry_miss or volatile system-prompt metadata.

## Steps to Reproduce

1. Configure any model with multiple variants where the variant carries `reasoningEffort`/`textVerbosity` (e.g. `axonhub` provider, `deepseek-v4-pro` with `max` variant `{ reasoningEffort: "max", textVerbosity: "medium" }`).
2. Run a session that ends with incomplete todos so `todo-continuation-enforcer` injects a continuation (2s countdown → `[SYSTEM DIRECTIVE: OH-MY-OPENCODE - TODO CONTINUATION]`).
3. Compare the injected request body with the preceding user-initiated request: the continuation is missing `reasoning_effort` and `verbosity` even though the message array is byte-identical.
4. Observe provider usage on the continuation turn: `cached_tokens` collapses to a tiny fraction (e.g. 3,072 / 140,331 = 2.2% vs 99.4% on the previous turn).

## Expected Behavior

The continuation request should carry the same `variant` (and thus the same `reasoning_effort` / `verbosity`) as the previous turn, preserving the provider cache prefix. The injected turn should show ~99% cache hit, not ~2%.

## Actual Behavior

The injected continuation drops the variant, the effective prompt prefix diverges, and the provider re-bills ~137K tokens as a cache miss.

## Suggested fix location

`resolveLatestMessageInfo` (and the sibling `convertSDKMessageToStoredMessage` in `sdk-message-lookup.ts`) must read the variant from the **flat** assistant message shape, not only from a nested `info.model`:

```ts
// resolve-message-info.ts — read variant from both shapes
const variant = info.model?.variant ?? info.variant;          // assistant messages keep variant flat at top level
model: info.model ?? (info.providerID && info.modelID
  ? { providerID: info.providerID, modelID: info.modelID, ...(variant ? { variant } : {}) }
  : undefined),
```

`packages/omo-opencode/src/features/hook-message-injector/sdk-message-lookup.ts` (`convertSDKMessageToStoredMessage`) has the identical flat-shape blind spot (`const variant = info.model?.variant;` with no `?? info.variant` fallback) and should be fixed the same way.

## Operating System

macOS (Darwin)

## OpenCode Version

opencode 1.18.25 (sqlite backend), oh-my-openagent 4.19.4, openai-compatible provider (AxonHub → ark / DeepSeek-compatible endpoint)

PeterPonyu:
## Summary

Fixes #7462 — the remaining part. The reported path (the todo-continuation enforcer's `resolve-message-info.ts`) was already fixed in `ca87be2f1`, but the same hand-rolled normalisation exists in four sibling injectors and each still drops the variant.

opencode stores user messages with a nested `info.model = { providerID, modelID, variant }` but assistant messages with flat `info.providerID / modelID / variant`. Every auto-continuation injector rebuilds the model as `info.model ?? { providerID, modelID }`, forgetting the third flat field on the assistant message that the backwards scan almost always lands on. Because `variant` is then sent as a separate top-level `promptAsync` body field derived from `model.variant`, the injected turn goes out without `reasoning_effort` / `verbosity`, which changes the provider's cache key and invalidates the whole cached prefix (the report's 99% → 2% cache hit). The SDK types do not declare `variant` on messages, so the type checker cannot catch it. All four sites confirmed RED on `dev@4ec56b79d`.

## Changes (+28/−4 production)

Same idiom the maintainer accepted in `ca87be2f1` and #7698 — read the variant from either message shape and spread it only when present, so output is byte-identical when no variant exists:

- `features/hook-message-injector/sdk-message-lookup.ts` — the enforcer's own fallback lookup (also used by atlas): `info.model?.variant ?? info.variant`.
- `hooks/ralph-loop/continuation-prompt-injector.ts` — flat rebuild spreads `variant`.
- `hooks/atlas/recent-model-resolver.ts` — returned model carries `variant`.
- `hooks/unstable-agent-babysitter/unstable-agent-babysitter-hook.ts` + `task-message-analyzer.ts` — the narrowed `getMessageInfo` never copied a flat `variant`, so the hook's spread alone would still yield `undefined`; both are covered.

No shared helper (a `resolveMessageModel(info)` across five call sites with different local types is a follow-up refactor, not a bug fix) and `json-message-lookup.ts` is untouched.

## Test plan

- [x] One `#given … #when … #then …` case per site, RED on `dev` → GREEN: `injector.test.ts` (flat assistant variant preserved by `findNearestMessageWithFieldsFromSDK`, plus a nested-vs-flat control pinning that the nested value wins), `recent-model-resolver.test.ts`, `continuation-prompt-injector.test.ts` and `index.test.ts` for the babysitter (both assert `promptAsync` receives `variant` as a top-level field).
- [x] Four target directories 420 → 425 pass / 0 fail; `hooks/todo-continuation-enforcer/` unchanged at 146 pass; combined 571 pass / 0 fail. `bun run --cwd packages/omo-opencode typecheck` (tsgo): clean.
- [x] Existing `toEqual({ providerID, modelID })` assertions stay green because the spread adds nothing when the variant is absent.

## Relationship to #7463

#7463 (@AceRothstein71) identified the flat-assistant shape first; its `resolve-message-info.ts` / `types.ts` hunks have since landed via `ca87be2f1`, and its remaining `sdk-message-lookup.ts` hunk is the same one-line change included here. It has been waiting on a rebase since Aug 31 with its author having paused work; if it is rebased first I will drop the overlapping hunk from this PR.

---
## Summary by cubic
Fixes the last #7462 gap: four auto-continuation injectors rebuild the model from assistant-shaped messages without carrying the flat `variant` field over, so the injected turn goes out without `reasoning_effort`/`verbosity` and the provider cache key changes (cache hits dropped from 99% to 2%). Each injector now reads `variant` from either the nested or flat message shape and only adds it when present, so output is byte-identical when there is no variant.

**Changes**
- `hook-message-injector/sdk-message-lookup.ts` falls back to `info.variant` when the nested `model.variant` is absent.
- The ralph-loop continuation injector, atlas recent-model resolver, and babysitter hook spread the flat `variant` when rebuilding `{ providerID, modelID }`.
- `task-message-analyzer.ts` copies the flat `variant` through its narrowed `getMessageInfo` so the babysitter hook can use it.
- New tests cover the flat-assistant case at each site; the nested `model.variant` still wins when both are present.

<sup>Written for commit cf45e33555d6eb46011125c87dc32375061551b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8035?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/features/hook-message-injector/injector.test.ts
+++ b/packages/omo-opencode/src/features/hook-message-injector/injector.test.ts
@@ -53,6 +53,7 @@ function createMockClient(messages: Array<{
     model?: { providerID?: string; modelID?: string; variant?: string }
     providerID?: string
     modelID?: string
+    variant?: string
     tools?: Record<string, boolean>
     time?: { created?: number }
   }
@@ -98,6 +99,38 @@ describe("findNearestMessageWithFieldsFromSDK", () => {
       model: { providerID: "openai", modelID: "gpt-5" },
       tools: undefined,
     })
+  })
+
+  it("#given an assistant-shaped message with a flat variant #when resolving the nearest message #then the stored model keeps the variant", async () => {
+    // given
+    const mockClient = createMockClient([
+      { info: { agent: "sisyphus", providerID: "openai", modelID: "gpt-5.5", variant: "max" } },
+    ])
+
+    // when
+    const result = await findNearestMessageWithFieldsFromSDK(unsafeTestValue(mockClient), "ses_flat_variant")
+
+    // then
+    expect(result?.model).toEqual({ providerID: "openai", modelID: "gpt-5.5", variant: "max" })
+  })
+
+  it("#given a user-shaped message with a nested variant and a differing flat variant #when resolving the nearest message #then the nested variant wins", async () => {
+    // given
+    const mockClient = createMockClient([
+      {
+        info: {
+          agent: "sisyphus",
+          model: { providerID: "openai", modelID: "gpt-5.5", variant: "max" },
+          variant: "low",
+        },
+      },
+    ])
+
+    // when
+    const result = await findNearestMessageWithFieldsFromSDK(unsafeTestValue(mockClient), "ses_nested_variant")
+
+    // then
+    expect(result?.model).toEqual({ providerID: "openai", modelID: "gpt-5.5", variant: "max" })
   })
 
   it("returns nearest (most recent) message with all fields", async () => {
--- a/packages/omo-opencode/src/features/hook-message-injector/sdk-message-lookup.ts
+++ b/packages/omo-opencode/src/features/hook-message-injector/sdk-message-lookup.ts
@@ -20,6 +20,7 @@ export interface SDKMessage {
     }
     readonly providerID?: string
     readonly modelID?: string
+    readonly variant?: string
     readonly tools?: Record<string, ToolPermission>
     readonly time?: {
       readonly created?: number
@@ -38,7 +39,7 @@ function convertSDKMessageToStoredMessage(msg: SDKMessage): StoredMessage | null
 
   const providerID = info.model?.providerID ?? info.providerID
   const modelID = info.model?.modelID ?? info.modelID
-  const variant = info.model?.variant
+  const variant = info.model?.variant ?? info.variant
 
   if (!info.agent && !providerID && !modelID) {
     return null
--- a/packages/omo-opencode/src/hooks/atlas/recent-model-resolver.test.ts
+++ b/packages/omo-opencode/src/hooks/atlas/recent-model-resolver.test.ts
@@ -42,4 +42,33 @@ describe("resolveRecentPromptContextForSession", () => {
     expect(result.model).toEqual({ providerID: "openai", modelID: "gpt-5.4" })
     expect(result.tools).toEqual({ edit: true })
   })
+
+  test("#given a flat assistant message with a variant #when resolving recent prompt context #then the model keeps the variant", async () => {
+    // given
+    const ctx = unsafeTestValue<PluginInput>({
+      client: {
+        session: {
+          messages: mock(async () => ({
+            data: [
+              {
+                id: "m2",
+                info: {
+                  providerID: "openai",
+                  modelID: "gpt-5.5",
+                  variant: "max",
+                  time: { created: 2 },
+                },
+              },
+            ],
+          })),
+        },
+      },
+    })
+
+    // when
+    const result = await resolveRecentPromptContextForSession(ctx, "ses_atlas_flat")
+
+    // then
+    expect(result.model).toEqual({ providerID: "openai", modelID: "gpt-5.5", variant: "max" })
+  })
 })
--- a/packages/omo-opencode/src/hooks/atlas/recent-model-resolver.ts
+++ b/packages/omo-opencode/src/hooks/atlas/recent-model-resolver.ts
@@ -38,6 +38,7 @@ export async function resolveRecentPromptContextForSession(
         model?: ModelInfo
         modelID?: string
         providerID?: string
+        variant?: string
         tools?: Record<string, boolean | "allow" | "deny" | "ask">
         time?: { created?: number }
       }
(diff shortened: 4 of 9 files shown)


code-yeongyu (approved):
420/420 pass. Clean fix for the remaining #7462 sibling sites — flat assistant variant preserved across all four continuation injectors, matching the pattern from ca87be2f1. LGTM.

Issue #7521, opened 2026-08-31
Doctor deprecation hint contradicts plugin schema (fallback_models vs models) — following it breaks plugin startup


johndfowler:
### Version

4.19.4 (identical behavior from both `oh-my-openagent` and `oh-my-opencode`)

### Summary

The doctor CLI's deprecation fix hint contradicts the plugin's own config schema, and following it breaks plugin startup. `config migrate` does not handle the case.

### Repro

`~/.omo/omo.jsonc`:

```jsonc
{
  "[opencode]": {
    "agents": {
      "sisyphus": {
        "models": [ { "model": "kimi-for-coding/k3", "reasoning": "max" } ]
      }
    }
  }
}
```

1. `bunx oh-my-openagent doctor` → 11 × `Unknown config key: agents.<name>.models` ("Affects: plugin startup"). So far so good — `models` is not in the plugin schema.
2. Doctor's own fix hint for a related finding says: `Replace fallback_models with models`. Following it (renaming the corrected `fallback_models` array to `models`) brings back issue 1 — plugin startup breaks again.
3. Meanwhile, with `fallback_models` in place, doctor emits `Deprecated reasoning config key … Fix: Replace fallback_models with models` — a warning whose fix reverts you to the startup-breaking state.
4. `bunx oh-my-openagent config migrate` reports `status: skipped — Nothing to migrate` for all of the above.

### Evidence (from the installed dist, 4.19.4)

Plugin schema accepts per-agent keys:

```js
recordFields(definition, ["description", "model", "fallback_models", "variant", "temperature", "top_p", "maxTokens", "thinking", "reasoningEffort", "textVerbosity", "tools", "prompt_append", ...])
```

— `fallback_models` valid, `models` rejected.

Doctor CLI deprecation table:

```js
CANONICAL_REPLACEMENT = new Map([ ..., ["fallback_models", "models"] ]);
```

— treats `fallback_models` as deprecated in favor of `models`.

### Impact

- With the invalid `models` key present, every agent model override is silently ignored at plugin startup — the plugin runs on built-in defaults with no error surfaced beyond `doctor`.
- Users who follow the doctor's fix hint loop between "startup broken" and "deprecated warning" with no valid target state.
- Additionally, `models` entries using the deprecated `reasoning` field are surfaced as "Deprecated reasoning config key" even when the path is the *valid* `fallback_models` shape, which is how this confusion compounds.

### Expected

[...]

PR #8036, opened 2026-09-09
fix(omo-opencode): make the doctor fallback_models hint truthful

PeterPonyu:
## Summary

Addresses the remaining part of #7521. The 4.19.4 contradiction the issue describes (doctor tells you to rename `fallback_models` → `models`, the plugin schema rejects `models`) is already closed on `dev` — `989636bc5` made the plugin's agent overrides accept `models`, and `ac85d1f3c` taught the doctor to skip the plugin-owned `[opencode]` block — but only in `5.0.0-beta.x`; `v4.19.4` (2026-08-01) was cut inside the window where the two sides disagreed and is still npm `latest`. I replayed the reporter's config against `validatePluginConfig` and the doctor check on `dev@4ec56b79d`: accepted and materialised.

What is still wrong on `dev` is the hint text itself:

1. It says "Replace fallback_models with models". `models` is an ordered chain whose head is the *primary*, so a literal rename of a `fallback_models` list silently promotes the first fallback to primary.
2. It unconditionally adds "or run: oh-my-openagent config migrate". That migration is deliberately one-shot (journaled via `hasMigrationMarker`) and is auto-applied at first plugin startup, so for anything edited afterwards it answers "Nothing to migrate" — the reporter's step 4.

## Changes (`packages/omo-opencode/src/cli/doctor/checks/deprecated-reasoning-keys.ts`, +19/−6)

- The `fallback_models` replacement reads `models: [<primary>, ...fallbacks] (the first entry becomes the primary model)`.
- The `config migrate` suggestion is appended only when the scanned file lacks the `2026-08-reasoning-unification` marker, using the existing `hasMigrationMarker` / `REASONING_UNIFICATION_MIGRATION_ID`. No schema or migration-engine change; the one-shot policy is untouched.

## Test plan

- [x] `deprecated-reasoning-keys.test.ts`: marker present → fix text explains that the first entry becomes the primary and does not mention `config migrate` (RED on `dev`: old string); no marker → still suggests `config migrate` (pins the scoped behaviour).
- [x] File 3 → 5 pass / 0 fail; `bun test packages/omo-opencode/src/cli/doctor/`: 176 pass / 0 fail (28 files). `bun run --cwd packages/omo-opencode typecheck` (tsgo): clean.

## Notes

Not a 4.x backport; the reporter's loop is resolved by upgrading to `5.0.0-beta`. Answered on the issue.

---
## Summary by cubic
Makes the doctor's `fallback_models` fix hint truthful so users don't lose their primary model or get sent in a loop.

- Updates the replacement text to `models: [<primary>, ...fallbacks]` and explains the first entry becomes the primary.
- Only suggests `config migrate` when the reasoning-unification migration marker is absent from the scanned file.

<sup>Written for commit 9b094392a26e123f39a36a1990a9f5227dd56576. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/omo-opencode/src/cli/doctor/checks/deprecated-reasoning-keys.test.ts
+++ b/packages/omo-opencode/src/cli/doctor/checks/deprecated-reasoning-keys.test.ts
@@ -3,6 +3,23 @@ import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
 
+import { REASONING_UNIFICATION_MIGRATION_ID } from "../../../config-migration"
+
+const FALLBACK_MODELS_FIXTURE = {
+  categories: {
+    deep: {
+      model: "openai/gpt-5.6-sol",
+      fallback_models: ["openai/gpt-5.6-terra"],
+    },
+  },
+  agents: {
+    explore: {
+      model: "openai/gpt-5.6-luna",
+      fallback_models: ["openai/gpt-5.6-terra"],
+    },
+  },
+}
+
 describe("deprecated reasoning keys check", () => {
   it("reports deprecated reasoning keys with exact file and key path", async () => {
     //#given a unified config containing deprecated reasoning keys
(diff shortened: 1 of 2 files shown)


code-yeongyu (approved):
3/3 pass. Doctor fallback_models hint now reflects actual config state. LGTM.

PR #8038, opened 2026-09-09
test(memory-core): end lock-holder fixtures on parent stdin EOF and await teardown

PeterPonyu:
## Summary

Addresses the remaining part of #7335. The Senpi `hold-lock` fixture the issue names was already fixed on `dev` by #7404 and #7479 (its lifecycle regression is green on HEAD and leaves no orphans). The identical pre-fix shape survives in memory-core's test fixtures: `locks/subprocess-worker.ts` `hold` mode and `concurrency/writer-child.ts` `kill-holder` mode both parked on `await new Promise<never>(() => {})` and relied entirely on the parent test's `afterEach` SIGKILL, which never runs if the `bun test` worker is itself killed.

Why that spins a core rather than idling: Bun, unlike Node (which exits with code 13), keeps a process with an unsettled top-level await alive and, once the event loop has no live handle, busy-polls. Measured with Bun 1.4.2 while investigating: ~265 CPU ticks per 3 s for the pre-fix shape (and for memory-core's `hold` mode after stdin EOF), 0 ticks for a timer keep-alive, immediate exit for the current event-driven Senpi fixture.

## Changes (test fixtures and tests only; no lock semantics change, no production code)

- `subprocess-worker.ts` `hold` mode and `writer-child.ts` `kill-holder` mode park on the parent-owned stdin pipe (`end` / `close` / `error`, destroying stdin in the resolver) — the same wait shape as the merged Senpi fixture — so a holder idles on a live handle while held and exits on its own when the parent dies, on Linux, macOS and Windows. No PPID polling, no wall-clock timeout (both add nondeterminism the merged fix deliberately removed). `releaseLock` is intentionally *not* called on this path: the "SIGKILLed holder is recovered" tests depend on the marker being left behind like a crashed owner; `writer-child` exits explicitly for the same reason instead of falling into its `finally`.
- `subprocess-contention.test.ts` and `two-process.test.ts`: ordered, bounded teardown (SIGTERM → await ≤2 s → SIGKILL → await ≤2 s → throw if still alive → then `rm`), replacing fire-and-forget kills.
- New `locks/process-liveness.test-support.ts`: `exitedWithin` and identity-guarded pid probes ported from the Senpi support file (memory-core cannot import from omo-senpi).

## Test plan

- [x] New lifecycle tests in `subprocess-contention.test.ts`, RED on `dev` → GREEN: a hold-mode holder exits by itself within 5 s after the parent ends its stdin (RED: `timed out waiting for child exit`); a wrapper that spawns the holder and SIGKILLs itself leaves the *original* worker terminal within 5 s (RED: orphan survives with ppid 1); after teardown every tracked child has `exitCode` or `signalCode` set (RED: old `afterEach` never awaited). Termination is asserted on every platform; POSIX signal details only off-Windows. Every test kills and awaits its children in `finally`, and the wrapper kills itself on its own stdin EOF so a dead test runner still takes the tree down.
- [x] `bun test packages/memory-core/src/locks/ packages/memory-core/src/concurrency/`: 26 pass / 0 fail; the lifecycle file run 5× consecutively 7/7 each time; `pgrep` for the fixture scripts empty after every run. `bun run --cwd packages/memory-core typecheck` (tsgo): clean.

## Housekeeping (not in this PR)

#7335 itself can be closed once this lands (the Senpi part is #7404 / #7479); #7342 is superseded by those; #7883 (a six-line await for the Windows-only control-socket race in the merged lifecycle test) was closed unmerged on 2026-09-07 and may be worth reopening.

---
## Summary by cubic
Fixes memory-core's lock-holder test fixtures so an orphaned holder no longer busy-polls a core forever under Bun (#7335). The fixtures used to park on an unsettled top-level await and rely on the parent's SIGKILL; they now wait on the parent-owned stdin pipe and exit on their own when the parent dies.

- Both fixture modes wait on stdin `end`/`close`/`error` and exit without releasing the lock, leaving the marker behind like a crashed owner.
- Teardown in `subprocess-contention.test.ts` and `two-process.test.ts` now SIGTERMs, waits up to 2s, SIGKILLs, waits up to 2s, then throws, before removing temp dirs.
- New `locks/process-liveness.test-support.ts` adds bounded, zombie-aware pid probes ported from `omo-senpi`.
- New lifecycle tests in `subprocess-contention.test.ts` fail on `dev` (orphan survives) and pass here.
- Callers must spawn these fixtures with a piped stdin they keep open; all current call sites already do.

<sup>Written for commit b3f417157359a9c2b7e402c4ab2939d6d6e0ef15. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/memory-core/src/concurrency/two-process.test.ts
+++ b/packages/memory-core/src/concurrency/two-process.test.ts
@@ -7,6 +7,7 @@ import { join } from "node:path"
 import { fileURLToPath } from "node:url"
 
 import { buildIdentityPaths, type MemoryIdentity } from "../identity"
+import { exitedWithin } from "../locks/process-liveness.test-support"
 import { ReflectionReservationStore, type ReservationResult } from "../reflection/reservation"
 import { realpathSync } from "node:fs"
 
@@ -14,6 +15,8 @@ const writerChildPath = fileURLToPath(new URL("./writer-child.ts", import.meta.u
 const reflectionChildPath = fileURLToPath(new URL("./reflection-child.ts", import.meta.url))
 const temporaryDirectories: string[] = []
 const liveChildren = new Set<ChildProcessWithoutNullStreams>()
+
+const TEARDOWN_GRACE_MS = 2_000
 
 type Exit = { readonly code: number | null; readonly signal: NodeJS.Signals | null }
 
@@ -144,7 +147,20 @@ async function assertTwentyLinearCommits(identity: MemoryIdentity): Promise<void
 }
 
 afterEach(async () => {
-  for (const child of liveChildren) child.kill("SIGKILL")
+  // Temp roots die only after every tracked child is confirmed exited: rm under a live lock
+  // holder strands it, and a fire-and-forget SIGKILL is not a confirmation.
+  const tracked = [...liveChildren]
+  liveChildren.clear()
+  for (const child of tracked) {
+    if (child.exitCode !== null || child.signalCode !== null) continue
+    child.kill("SIGTERM")
+    if (!(await exitedWithin(child, TEARDOWN_GRACE_MS))) {
+      child.kill("SIGKILL")
+      if (!(await exitedWithin(child, TEARDOWN_GRACE_MS))) {
+        throw new Error(`tracked child pid ${String(child.pid)} survived SIGTERM and SIGKILL teardown`)
+      }
+    }
+  }
   await Promise.all(temporaryDirectories.splice(0).map((directory) => rm(directory, { recursive: true, force: true, maxRetries: 10, retryDelay: 200 })))
 })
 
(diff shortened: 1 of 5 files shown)


code-yeongyu (approved):
19/19 locks pass. Good test hygiene — stdin EOF teardown prevents orphan fixture processes. LGTM.

PR #8091, opened 2026-09-10
fix(senpi-task): keep the compiled engine when its binary has no .exe suffix

code-yeongyu:
## Problem

`senpi-compatibility (windows-latest)` is red on `dev`. Four tests fail, in two families:

```
(fail) resolveSenpiLaunch > #given the running engine is a compiled omo binary and PATH carries a senpi ...
(fail) resolveSenpiLaunch > #given a compiled omo binary with no senpi on PATH, no installed CLI and no entry script ...
(fail) resolveSenpiLaunch > #given a script-hosted engine and a senpi on PATH #then the PATH senpi is still chosen
```
```
  "command": "C:\Users\RUNNER~1\...\omo",          <- received
  "prefixArgs": ["D:\a\...\bundle-lazy-barrel.test.ts"]
```

The received `prefixArgs` is the **test file the runner happened to be executing**, which is the signature of resolution falling all the way through to the `argv[1]` entry-script guess.

## Root cause

`resolveSenpiExecutable` returns the running executable when the process embeds the engine (`isCompiledEngine`, added in 95ddd4be5). `resolveSenpiLauncher` then passed that answer to `normalizeSenpiLauncher`, which on `win32` reads **any** candidate whose name does not end in `.exe` as an npm shim and looks for an adjacent `dist/cli.js`.

A compiled single-file executable may be named anything — `omo`, `omo-dev` — so the shim reader found no `cli.js`, returned `null`, and **discarded the one candidate guaranteed to match the running version and its embedded assets**. Resolution then continued into the PATH shim scans and finally `argv[1]`.

That heuristic is right for a PATH or sibling hit, where a bare `senpi` on Windows genuinely is a shim. It is wrong for the compiled engine, which *is* the child engine by construction.

## Fix

`resolveSenpiLauncher` returns the running compiled executable directly and never re-interprets it. The shim reader keeps handling PATH and sibling candidates unchanged.

## Why no test caught it

The existing compiled-engine spawn test pins `platform: "linux"` (`spawn.test.ts`), so nothing covered the win32 branch. A win32 mirror is added.

Separately, the two memory-launcher PATH fixtures created an extension-less `senpi` while Windows resolution scans for `senpi.exe` — so those cases never exercised the PATH branch they assert. They now use the platform binary name, which makes the "compiled engine beats a PATH senpi" case actually meaningful on Windows.

## Verification

RED/GREEN on Linux (`gorky`, bun 1.4.2) with the win32-pinned test, plus a Linux bundle regeneration. Evidence in a follow-up comment.

Windows is only reachable through CI, so `ci:full-matrix` is required on this PR.

## Note (not fixed here)

`packages/senpi-task/src/runners/rpc/spawn.ts` is 272 pure LOC, already past the repo's 250 ceiling before this change. The cohesive split is launcher resolution (`detectBunBinary`/`detectCompiledEngine`/`resolveSenpiExecutable`/`normalizeSenpiLauncher`/`resolveSenpiLauncher` + version parity) into its own `senpi-launcher.ts`, leaving `spawn.ts` owning only the spawn descriptor. That move is deliberately **not** bundled into a red-`dev` unblocking fix on a file with heavy concurrent churn; it wants its own PR.

The `task-rpc-e2e.windows` `"Child prompt failed to start."` failure in the same job is a **separate** defect and is not addressed here.

---
## Summary by cubic
Fixes Windows resolution for compiled engine binaries without a `.exe` suffix so the child process launches itself instead of falling back to an entry-script guess.

**Bug Fixes**
- `resolveSenpiLauncher` now returns the running compiled executable directly, bypassing the shim reader that misreads non-`.exe` names as npm shims.
- Adds a Windows-pinned test mirror for the compiled-engine spawn case and fixes PATH fixtures to use platform-specific binary names so they actually exercise the PATH branch.
- Regenerates the `omo-senpi` extension bundles to carry the fix.

<sup>Written for commit 6855e9cf4fddd2fc55b9bd65f197b7351c23b77c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8091?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
### Verification evidence

**The job is deterministically red on `dev`, not flaky.** The Windows `senpi-compatibility` job only executes its test step under `ci:full-matrix`, so most `dev` runs report `success` with the step **skipped**. Every run that actually executed it failed:

| `dev` commit | CI run | job | test step |
|---|---|---|---|
| `f7d55806e` | 34472338671 | success | **skipped** |
| `88a5b8bc2` | 34471094511 | failure | **failure** |
| `826d81902` | 34461139555 | failure | **failure** |

**Failing-first proof** (`gorky`, Linux x86_64, bun 1.4.2). The win32-pinned test was added, then only the `spawn.ts` fix was reverted (`git show HEAD~1:...spawn.ts`, confirmed the guard is absent):

```
=== RED: revert ONLY the spawn.ts fix, keep new tests ===
fix absent: OK
(fail) buildRpcSpawn spawn strategy > #given a compiled engine on Windows whose binary has no .exe suffix
       #when building an RPC child #then it still launches itself [0.94ms]
 35 pass
 1 fail
Ran 36 tests across 1 file.
```

Exactly the new case fails, and only it — so the assertion is genuinely coupled to the fix rather than passing for an unrelated reason.

```
=== GREEN: restore fix ===
 44 pass
 0 fail
Ran 44 tests across 2 files.
```
(`packages/senpi-task/src/runners/rpc/spawn.test.ts` + `packages/omo-senpi/src/components/memory/worker/senpi-command.test.ts`)

Extension bundles were regenerated on Linux from the rebased head (`build-extension.mjs`, exit 0; `omo-task.js` and `omo.js` changed) because the fix is bundled source.

Windows itself is only reachable through CI, so `ci:full-matrix` is applied here and the Windows shards on this head are the authoritative check.

code-yeongyu:
### The fix turns the job green

`senpi-compatibility (windows-latest)` — the job that fails deterministically on `dev` whenever it actually runs — **passed on this PR's head `159d5394d`**, together with the other 9 required checks.

That closes the loop on the diagnosis:

| where | `senpi-compatibility (windows-latest)` |
|---|---|
| `dev` `826d81902` (run 34461139555) | failure — 3 × `resolveSenpiLaunch` |
| `dev` `88a5b8bc2` (run 34471094511) | failure — same 3 |
| this PR `159d5394d` | **success** |

`dev` has since moved, so the branch was rebased and the Linux bundle regenerated on top (`build-extension.mjs`, exit 0). Re-running the full matrix on `6855e9cf4` before merge.

PR #8092, opened 2026-09-10
feat(omo-senpi,omo-config-core): opt-in side panel for the Senpi TUI with usage bars and clickable rows

IvoryKir:
## What this is

An opt-in right-hand column in the Senpi TUI carrying what the session is actually doing. It is
**off by default** (`side_panel.enabled`), because it rearranges the whole screen.

```
SESSION  $0.14 · 826
model   claude-opus-5
elapsed 43s
tokens  in 4 · out 822
cache   97% hit
agents  $0.00
CONTEXT  32.9K/1M
used    █▎░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 3%
USAGE  1m14 ago
claude
5h      ███┊░░░░░░░░░░░░░░░░░░░░░░░░  11%  4h19
7d      ██████████▍░░░░░┊░░░░░░░░░░░  37%  Wed 08:00
Fable   ░░░░░░░░░░░░░░░░┊░░░░░░░░░░░   0%  Wed 08:00
account work
codex
5h      ▎░░┊░░░░░░░░░░░░░░░░░░░░░░░░   1%  4h25
7d      ┊░░░░░░░░░░░░░░░░░░░░░░░░░░░   0%  Sun 09:16
AGENTS  1 done
✓ Read src/pricing.ts and report the VAT value  5s
TOOLS  1
task
FILES  6 changed
D  cart.ts  +0/-1
M  pricing.ts  +1/-0
 M server.ts  +2/-1
??
??
?? checkout.ts
~/code/checkout-service · main
```

*A real frame, not a mock: a delegated child that has finished, the working copy as git sees it, and
the plan's rolling windows.*

## Why

Mid-run you cannot see what the session is costing, how much context is left, which subagents are
still alive, what changed on disk, or how much of your subscription is gone. omo knows every one of
those; none of them is on screen.

## What it shows

| Section | Contents |
|---|---|
| `session` | model, elapsed, cost and tokens, cache hit rate, and child spend called out separately because senpi's totals do not include it |
| `context` | context-window usage as a bar |
| `usage` | subscription windows for the serving account (Anthropic + Codex), with a marker where an evenly paced burn would be |
| `agents` | delegated children with live status and a ticking clock |
| `tools` | what the current exchange has been doing |
| `files` | `git status` as git sees it, both status columns, per-file `+/-` |
| location | cwd and branch, closing the column |

## Interaction

Rows are clickable: a file opens its diff in a framed scrollable viewer, a subagent opens its card
followed by its whole recorded transcript (rendered by the task engine itself, the same text
`task_output` returns). The wheel scrolls the viewer. `/side-panel-diff` reaches the file viewer by
name, for keyboards and for hosts that expose no URL hook. **No keyboard chord is registered by
default** - taking a global key in someone else's editor invites a collision.

Clicking works because the alt-screen renderer already captures the mouse (`options.mouse ?? true`,
and senpi passes none), and OSC 8 hyperlink activation is the one path it dispatches to the
application: on a press and release inside one cell with no drag it resolves a link from its own
rendered buffer and calls `openUrl`. Rows are therefore painted as hyperlinks to a private
`omo-panel:` scheme, and the panel claims that callback while it is mounted, handing every foreign
URL straight back. Text selection is untouched: dragging still selects, double and triple clicks
still take a word and a line.

## Host coupling

All of it is isolated in `components/side-panel/host-surface.ts`:

- The live renderer arrives through a zero-height anchor widget; the current layout root is wrapped
  in a pi-tui `hstack` so the transcript reflows beside the column instead of being covered, and the
  original root is restored on `session_shutdown` / `session_before_switch` - only if it is still
  ours.
- The layout contract is reached through `Symbol.for` registry symbols, so a host bundling its own
  pi-tui copy still interoperates. No renderer seam means a widget block above the editor; a
  headless run renders nothing.
- The host's `openUrl` is parked on the renderer in a plain object rather than kept in a closure:
  the renderer reaches an extension through a proxy that wraps every function it hands out, so a
  naive save-and-restore gains a wrapper per mount cycle.
- The wheel is read through `ui.onTerminalInput` and claimed with `consume`, because the host routes
  a wheel event to the layout under the pointer and an overlay is composited outside that layout.
- Rows come from pure builders in `sections/` that emit `{text, color}` and never touch the host, so
  the theme is applied once and every section is testable without a renderer.
- `files` reads `git status --porcelain=v1 -z` plus both `--numstat` sides through the host's
  `exec` (feature-detected; absent means the section stays empty), on a 2s floor rather than a
  watcher - a recursive fs watch burns the machine's scarce inotify instances for a section that
  only needs to be right within a second. The branch comes from `.git/HEAD` with no subprocess.

## Subscription usage

The only part of the panel that leaves the machine, and it is switchable off (`sections.usage`);
nothing is constructed while it is off - no credential read, no timer, no request.

- A quota belongs to an account, not to a session, so the numbers live in one cache file per machine
  (`$XDG_CACHE_HOME/omo-senpi/side-panel-usage.json`). Freshness, a shared backoff and a 30s
  "fetch announced" claim are decisions over that file, which is what keeps N sessions from making
  N requests. It is validated on read, because sibling sessions and older versions write it too.
- The token, the name printed under the bars and the cache key come from **one** decision. Deriving
  them separately lets the panel print one account's name over another account's quota.
- A rotation is honoured from both places senpi records it - the pool slot and the account object -
  so a rate-limited account stops being credited within a refresh.
- A short internal marker is never sent as a bearer: Anthropic answers that with a 429 carrying a
  48-minute retry-after. A stale token **is** sent, because the 401 is what lets the column say
  "auth stale - run /login".
- A failed poll keeps the previous bars and adds the reason with its retry deadline.

## Configuration

```jsonc
{
  "side_panel": {
    "enabled": true,
    "width": "26%",          // or a column count; clamped to 32-80 and to a 60-column transcript
    "min_columns": 120,      // narrower terminals keep the classic single-column layout
    "clickable": true,
    "usage_poll_seconds": 150,
    "sections": { "session": true, "context": true, "usage": true, "agents": true, "tools": true, "files": true }
  }
}
```

`--omo-side-panel` forces it on for one run; senpi coerces a boolean extension flag to `true` and
rejects `--no-` forms, so switching it off is an `omo.json` edit. `assets/omo.schema.json` is
regenerated and `docs/reference/omo-json.md` documents the block.

## Tests and gates

- `bun run test:senpi` - exit 0 in one run: **Ran 3672 tests across 445 files**, 0 failures. 290 of them are this component's.
- Root `bun run typecheck` - clean.
- Extension bundle **1180100 bytes** against the budget this PR raises to 1,220,000, with the measurement and the headroom documented in `bundle-size.test.ts` the way every previous raise is.
- Hostless tests never reach the network: the component test factory turns `sections.usage` off, and
  the tests that exercise it inject their own fetch, credential reader and cache path.

## Manual QA

Driven against the real `senpi` binary in a pty, not simulated: mount and teardown in fullscreen and
regular mode, the git section across all five states including an untracked directory, a real SGR
mouse click opening a file's diff, wheel scrolling measured by the viewer's own position indicator
(`1-32/405` → three notches → `10-41/405` → two back → `4-35/405`, with the transcript behind
byte-identical), and the usage bars against live accounts. Evidence lives under
`.omo/evidence/omo-senpi-adapter/`.

Three defects were found by that manual QA and no test: a cache-hit rate rendered as `9811%`
(`latestCacheHitRate` is already a percentage), usage bars whose width drifted per row, and an
untracked **directory** rendering as a status with no name. All three are fixed here, each with a
regression test.

## Caveats

- The reflowing column needs the fullscreen TUI; regular mode gets the same rows as a block above
  the editor, and a headless run renders nothing.
- Clicking a subagent was not driven live end to end - a live child costs a real model turn - but
  the transcript path is covered by an integration test that writes a child's event log into a real
  state dir and reads it back through the engine's own reader and renderer.
- `components/memory/facts-payload-cap.test.ts` is order-dependent and fails roughly one run in
  three on its own; it is untouched by this PR and mentioned only so a red CI run on it is not
  mistaken for this change.


IvoryKir:

PR #8097, opened 2026-09-10
fix(memory-core): read darwin process start time in-process instead of forking /bin/ps

### LINKED ISSUE #8096: [Bug]: memory lock identity probe forks /bin/ps on every check; shared RPC hosts never reap them (process-table exhaustion)
## Summary

`getProcessStartIdentity()` (`packages/memory-core/src/locks/process-identity.ts`) shells out to `/bin/ps -o lstart= -p <pid>` on macOS for every memory-lock ownership check. Inside a long-lived shared RPC host those children are never reaped, so `<defunct>` entries accumulate until the macOS process table is full and every `posix_spawn` on the machine fails with `EAGAIN`.

## Reproduction

1. Run a shared RPC host (`--mode rpc --multi`, or the desktop's `--mode rpc --multi-session --listen`) with a few live sessions for several hours, so memory locks / reflection reservations keep running.
2. `ps -Ao stat,ppid | awk '$1 ~ /^Z/ && $2==<host pid>' | wc -l`

Zombie children accumulate monotonically and are not reclaimed while the host lives.

## Expected (ideal state)

The memory-lock identity probe never forks a child on macOS. Process start time is read in-process, so there is no child to leak regardless of how the embedding host handles `SIGCHLD`.

## Actual

One `/bin/ps` fork per identity check, continuously (memory-write locks, reflection/facts reservations, lock recovery), each of which can become a permanent zombie inside a shared host.

## Evidence

- 2026-09-09, a macOS workstation: process table reached 10,981 entries with 9,750 zombies; **9,386 of them had `p_comm = ps`**, all parented to shared RPC hosts. Every fork on the box failed with `EAGAIN` — shells, `gh`, and `ps` itself. Upstream tracking issue: code-yeongyu/senpi#1507.
- 2026-09-10, new datapoint: the desktop socket hosts (`omo --mode rpc --multi-session --listen`) **also** leak. Measured 14 / 3 / 1 stuck `<defunct>` children on three hosts (uptimes 5h32m / 4h50m / 4h47m, 18m / 3m51s / 3m36s CPU). This refutes senpi#1507's scoping note that only `--mode rpc --multi` hosts are affected.
- `kill -CHLD <host pid>` did not clear them, i.e. the parent never `waitpid`s. `SIGTERM` on the three hosts reclaimed all 18 at once (machine-wide zombies 31 -> 13) and the hosts removed their own socket directories, so the shutdown path itself is clean.
- The probe is not intrinsically broken: 60 sequential `execFile("/bin/ps", ["-o","lstart=","-p",pid])` calls from a normal engine process finish in 125 ms and leave no accumulation (2 -> 0 -> 1 transient). The unreaped state is specific to the shared-host context, which is why removing the fork is the durable fix rather than chasing the host's signal handling.

## Root cause

`packages/memory-core/src/locks/process-identity.ts` resolves the darwin/freebsd identity by spawning `ps`:

```ts
const value = await execFileText("/bin/ps", ["-o", "lstart=", "-p", String(pid)])
return value === null ? null : `ps-lstart:${value.replace(/\s+/g, " ")}`
```

macOS exposes the same value in-process through `libproc`: `proc_pidinfo(pid, PROC_PIDTBSDINFO, 0, &info, sizeof info)` fills `struct proc_bsdinfo`, whose `pbi_start_tvsec` / `pbi_start_tvusec` are the process start time.

Verified locally for the current pid: libproc reports `sec = 1789041850` -> `Thu Sep 10 21:04:10 2026`, byte-identical to `/bin/ps -o lstart=`. `proc_pidinfo` returns `0` both for a nonexistent pid and for a pid owned by another uid, so the existing `/bin/ps` path has to stay as a fallback.

## Scope and acceptance criteria

- [ ] darwin: `getProcessStartIdentity()` reads the start time in-process (libproc) and spawns no child in the same-uid case.
- [ ] The identity string stays byte-identical to today's `ps-lstart:<lstart>` form, so lock files written by older builds still compare equal (no migration, no spurious stale-lock recovery).
- [ ] The `/bin/ps` fallback is retained for what libproc cannot answer (cross-uid pids, any failure, non-Bun runtime).
- [ ] The linux `/proc/<pid>/stat` path is unchanged.
- [ ] Regression test: the spawn-free path returns a value equal to `/bin/ps -o lstart=` for the current pid, and the test fails if the in-process path is removed or regresses to spawning.

## Related issues and PRs

- code-yeongyu/senpi#1507 — tracking issue for the host-side reaping failure. The host-level `waitpid` sweep proposed there is deliberately **not** taken here: a blind `waitpid(-1, WNOHANG)` inside the host would consume exit statuses that the runtime's own reaper is waiting for, hanging `await proc.exited` and never firing `execFile` callbacks. Removing the fork is safe by construction.
- #1306 — historical "prevent zombie processes with proper process lifecycle management" fix, against the pre-monorepo layout.
- #7579 — terminal resident children never reclaimed in-session (different subsystem, same family of defect).

code-yeongyu:
## What

The memory lock protocol fingerprints a process start time to detect PID reuse. On darwin it resolved that by spawning `/bin/ps -o lstart= -p <pid>` on **every** check - and every lock record creation, reflection/facts reservation and stale-lock recovery does one.

A long-lived shared RPC host never reaps those children. They accumulate as `<defunct>` entries until the macOS process table is full, at which point every `posix_spawn` on the machine fails with `EAGAIN`: shells, `gh`, `ps` itself, and the agent's own tooling.

This reads the same value in-process through libproc instead, so there is no child to leak regardless of how the embedding host handles `SIGCHLD`.

## Root cause

`packages/memory-core/src/locks/process-identity.ts` used `execFile("/bin/ps", ["-o","lstart=","-p",pid])`. macOS exposes the identical value in-process via `proc_pidinfo(pid, PROC_PIDTBSDINFO, ...)` -> `struct proc_bsdinfo.pbi_start_tvsec`.

## Evidence

- 2026-09-09: a macOS workstation reached 10,981 process-table entries with 9,750 zombies; **9,386 had `p_comm = ps`**, all parented to shared RPC hosts, and every fork on the box failed with `EAGAIN`. Tracked upstream as code-yeongyu/senpi#1507.
- 2026-09-10: the desktop socket hosts (`--mode rpc --multi-session --listen`) leak too - 14 / 3 / 1 stuck `<defunct>` children over 5h32m / 4h50m / 4h47m uptime. `kill -CHLD` did not clear them (the parent never `waitpid`s); `SIGTERM` on the hosts reclaimed all 18 at once.
- The probe is not intrinsically broken: 60 sequential `execFile("/bin/ps", ...)` calls from a normal engine process finish in 125 ms with no accumulation. The unreaped state is specific to the shared-host context, which is why this removes the fork rather than trying to fix the host's signal handling.
- libproc parity verified for the live pid: `pbi_start_tvsec = 1789041850` -> `Thu Sep 10 21:04:10 2026`, byte-identical to `/bin/ps -o lstart=`.

## Why not a `waitpid` sweep in the host

senpi#1507 proposes a host-level `SIGCHLD`/`waitpid` sweep as defence in depth. That is deliberately **not** done here: a blind `waitpid(-1, WNOHANG)` inside the host consumes exit statuses the runtime's own reaper is waiting for, which would hang `await proc.exited` forever and silently drop `execFile` callbacks. Deleting the fork is safe by construction.

## Compatibility

- The identity string keeps its exact `ps-lstart:<lstart>` shape, formatted with C-locale abbreviations and a space-padded day, then whitespace-collapsed by the same code path as before. Lock files written by older builds still compare equal, so no migration and no spurious stale-lock recovery.
- `/bin/ps` remains the fallback for pids libproc cannot inspect (another uid, any failure, non-Bun runtime). `proc_pidinfo` returns 0 for both a dead pid and a foreign-uid pid.
- Dead pids are now rejected up front by the existing fork-free `getPidLiveness()` probe, so the stale-lock recovery path no longer forks either. Same result as before (`null` -> lock treated as stale), one less spawn.
- The linux `/proc/<pid>/stat` path is untouched. Windows still returns `null` as before.

## Tests

`packages/memory-core/src/locks/process-start-time.test.ts`:
- the in-process probe for the current pid equals `/bin/ps -o lstart=` output;
- unreachable / zero / negative / non-integer pids report `null` rather than a guess;
- `getProcessStartIdentity()` returns the spawn-free probe value verbatim (this fails if the in-process path is removed, since the fallback would leave the probe `null`);
- a dead pid resolves to `null` so the lock still counts as stale.

Fixes #8096
Refs code-yeongyu/senpi#1507

---
## Summary by cubic
Fixes #8096 by replacing macOS's per-check `/bin/ps` fork with an in-process libproc read, so shared RPC hosts no longer leak zombie `ps` children until every `posix_spawn` fails with `EAGAIN`. Identities are now published as `proc-start-epoch:<seconds>` instead of `ps-lstart:` local time, which also avoids timezone-dependent mismatches.

**Compatibility**
- Legacy `ps-lstart:` records are treated as no conflict, so a live owner keeps its lock during upgrade.
- `/bin/ps` stays as fallback for pids libproc cannot inspect; Linux and Windows paths are unchanged.
- Dead pids are rejected by the existing fork-free liveness probe before any spawn, so stale-lock recovery no longer forks.
- The committed Senpi plugin bundles were rebuilt with the CI-pinned bun so the bundle-freshness build gate stays green.

<sup>Written for commit feb54dad31f1c10116dced7453c3da988317101b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8097?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## QA evidence (measured, not inferred)

Runs on a macOS arm64 workstation, `bun test v1.4.0`, worktree at `origin/fix/8096-spawn-free-process-identity` @ `b8cfa7e76`. Not run on the session machine.

**GREEN - lock suite**
```
bun test packages/memory-core/src/locks/
 35 pass, 0 fail, 92 expect() calls   (35 tests across 5 files)
```

**GREEN - whole package**
```
bun test packages/memory-core/src/
 760 pass, 0 fail, 5305 expect() calls   (760 tests across 86 files, 73.14s)
```

**Regression that the first cut introduced, now fixed.** `subprocess-contention.test.ts > #given a live Bun lock holder #when a contender tries immediately #then the live owner is never stolen` failed on the first commit and passes now. Baseline on `origin/dev` was 7 pass / 0 fail, so the failure was mine, not pre-existing.

**RED proof - the new tests can fail for the right reason.** Disabling the in-process probe (early `return null` in `readDarwinProcessStartSeconds`) flips the suite to `6 pass, 3 fail`, including:
```
(fail) darwin process start time > #given the lock identity API #when it resolves the current pid #then it publishes the spawn-free epoch identity
```
The mutation was reverted afterwards (empty `git diff`).

**libproc parity, measured outside the test runner** (where the JS timezone is not pinned to UTC):
```
{"pid":38185,"ps":"Thu Sep 10 22:26:18 2026","mine":"Thu Sep 10 22:26:18 2026","equal":true}
{"pid":1,   "ps":"Thu Aug 20 19:31:20 2026","mine":"null","equal":false}   <- root-owned, falls back by design
```

**Typecheck**: `bun run --cwd packages/memory-core typecheck` clean on both machines.

**Pure LOC** of touched files: `process-start-time.ts` 41, `process-identity.ts` 63, `process-start-time.test.ts` 89, `lock-record.test.ts` 163. `acquire.ts` is 284 and already over the 250 ceiling on `dev`; this change replaces one expression there and adds no lines, so the split is left as a separate concern.

PR #8144, opened 2026-09-11
fix(background-agent): fan out poll completion probes across sessions

AceRothstein71:
## Problem

With many background tasks alive, poll ticks stop keeping up and completion detection lags badly. The more idle sessions there are, the worse it gets, which looks from the outside like a global rate limit or a stalled event loop.

## Root cause

`pollRunningTasks()` walked running tasks in a sequential `for` loop and awaited the completion probes inline, one task at a time. Each idle session could cost two serial round-trips (`validateSessionHasOutput`, then `checkSessionTodos` or `verifySessionExists`), so probe depth was O(idle tasks) per tick. With 40 idle tasks that is up to 80 serial round-trips inside a single 3 s tick, while `pollingInFlight` holds off the next one.

This is provider agnostic. It is not a rate limit and not a model problem, it is serial fan-out in the poll path.

## Fix

Split the poll tick into three phases:

1. `collectPollCompletionCandidates` walks tasks synchronously and classifies each one as terminal or probe. No awaits, so no session can block classification of another.
2. `startPollSessionProbes` (new `poll-session-probes.ts`) runs the probes as a bounded worker pool across sessions. Each session keeps its own sequential chain, so probe ordering per session is unchanged, but distinct sessions now overlap.
3. `resolvePollCompletionCandidates` consumes the per-session results and applies completion.

Probe concurrency is capped at `MAX_POLL_PROBE_CONCURRENCY = 8` and every individual probe is bounded by `POLL_PROBE_TIMEOUT_MS = 2_500`. A probe that fails or times out degrades to a partial result, which surfaces as "unresolved" and defers to the next tick. A failed probe can never falsely complete or fail a task.

Behavior that is deliberately preserved: per-session probe ordering, the `pollingInFlight` re-entry guard, terminal candidates short-circuiting without any probe, and the existing crash-detection rule that requires `hasOutput === false` and the session-gone threshold and `sessionExists === false` together.

## Tests

`manager.polling.parallel-probes.test.ts` pins the actual regression by recording `maxInFlight` across a tick. Against `dev` it fails with `Expected: 5, Received: 1`, which is the serial behavior. Supporting suites cover the worker pool directly (`poll-session-probes.test.ts`, `poll-session-probes.concurrent.test.ts`) and the manager guards, fan-out and todo paths (`manager.polling.parallel-probes.guards.test.ts`, `.fanout.test.ts`, `.todos.test.ts`, `.order.test.ts`). All deferred gates in these tests are deterministic, with no `setTimeout` sleeps.

## Verification

- `bunx tsgo --noEmit -p tsconfig.json` clean
- `bun test src/features/background-agent/` 784 pass, 0 fail, 71 files

## Review

The first draft of this change was put through two independent adversarial reviews before this revision. They raised a stale-attempt race on the completion path, todo-cache staleness, unbounded probe fan-out, missing per-probe timeouts, shutdown not being honored mid-resolution, and oversized test files. Every one of those is addressed here: the completion path re-checks `isCurrentPollCompletionCandidate` after each await, shutdown is checked before probes start and again inside the resolution loop, fan-out is capped, every probe carries a timeout, and the integration suite is split so each file stays under the 200 LOC guideline.

### DIFF
--- a/packages/omo-opencode/src/features/background-agent/manager.polling.parallel-probes.fanout.test.ts
+++ b/packages/omo-opencode/src/features/background-agent/manager.polling.parallel-probes.fanout.test.ts
@@ -1,0 +1,107 @@
+/// <reference types="bun-types" />
+
+import { describe, expect, test } from "bun:test"
+import { tmpdir } from "node:os"
+import type { PluginInput } from "@opencode-ai/plugin"
+import { unsafeTestValue } from "../../../../../test-support/unsafe-test-value"
+import { BackgroundManager } from "./manager"
+import { MAX_POLL_PROBE_CONCURRENCY } from "./poll-session-probes"
+import type { BackgroundTask } from "./types"
+
+function createDeferred(): { readonly promise: Promise<void>; readonly resolve: () => void } {
+  let resolve: (() => void) | undefined
+  const promise = new Promise<void>((settle) => {
+    resolve = settle
+  })
+  return { promise, resolve: () => resolve?.() }
+}
+
+function createTask(sessionId: string): BackgroundTask {
+  return {
+    id: `bg_test_${sessionId}`,
+    sessionId,
+    parentSessionId: "parent-session",
+    parentMessageId: "parent-message",
+    description: "test task",
+    prompt: "test",
+    agent: "explore",
+    status: "running",
+    startedAt: new Date(),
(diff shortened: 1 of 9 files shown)


AceRothstein71:
Please intergate this PR thanks !

Issue #8186, opened 2026-09-12
doctor: fallback_models→models fix suggestion fails schema — agent config rejects 'models' key (schema/doctor contradiction)


rohitvipin-inapp:
### Bug description

`oh-my-openagent doctor` (v4.19.4) flags every `agents.*.fallback_models` key as deprecated:

```
15. Deprecated reasoning config key
    /Users/.../.omo/omo.jsonc: [opencode].agents.multimodal-looker.fallback_models
    Fix: Replace fallback_models with models, or run: oh-my-openagent config migrate
```

The suggested replacement is rejected by doctor's own schema validation. Converting agent entries to use a `models` array (exactly as the fix message instructs) produces:

```
Unknown config key: [opencode].agents.sisyphus.models
```

for every converted agent — because `AgentOverrideConfigSchema` (`packages/omo-opencode/src/config/schema/...` compiled to `dist/cli/index.js:104417`) has no `models` field:

```js
var AgentOverrideConfigSchema = exports_external.object({
  model: exports_external.string().optional(),
  fallback_models: FallbackModelsSchema.optional(),
  ...
```

`fallback_models` is a **declared** key on `AgentOverrideConfigSchema` — it is valid, not deprecated by schema. The doctor checker `CANONICAL_REPLACEMENT` map (`deprecated-reasoning-keys.ts`, compiled to `dist/cli/index.js:108270`) unconditionally maps `fallback_models → models` for **all** containers:

```js
var CANONICAL_REPLACEMENT = new Map([
  ["variant", "reasoning"],
  ["reasoningEffort", "reasoning"],
  ["thinking", "reasoning"],
  ["textVerbosity", "provider_options.textVerbosity"],
  ["fallback_models", "models"]
]);
var TUNING_CONTAINERS = new Set(["agents", "categories", "models"]);
```

but only **categories** accept `models` (CategoryConfigSchema has a `models` array field). Agents never do — in any version.

### Expected behavior

The checker should either:

1. Only warn about `fallback_models → models` inside `categories` (where `models` is actually accepted), and not flag `agents.*.fallback_models` (which is the only valid spelling for agents), **or**
2. Add `models: z.array(z.string()).optional()` to `AgentOverrideConfigSchema` so the replacement actually validates.

Additionally, `config migrate` reports `Nothing to migrate` even when agent entries still use `fallback_models`, so the suggested remediation is a dead end.

### Reproduction

1. `~/.omo/omo.jsonc`:
[...]

PR #8192, opened 2026-09-12
fix(config-migration): keep a category models chain when merging fallback_models

### LINKED ISSUE #8191: [Bug]: config migrate drops a category's models chain when fallback_models is also present
### Prerequisites

- [x] I am using the latest version of oh-my-openagent (reproduced on `dev` @ `e0746bc`)

### Bug Description

The `2026-08-reasoning-unification` migration folds a definition's legacy `fallback_models` into the canonical `models` chain. For agents it keeps the chain that is already in the file; for categories it throws it away. `normalizeDefinition()` in `packages/omo-opencode/src/config-migration/reasoning-unification.ts` reads the existing chain only for agents:

```ts
const primary = primaryModelRef(value, opencode)
const existing = kind === "agent" ? normalizedList(value["models"]) : []
const fallbacks = normalizedList(fallbackModels)
normalized["models"] = [...(primary === undefined ? [] : [primary]), ...existing, ...fallbacks]
```

`CategoryConfigSchema` accepts `models` ("Ordered model chain; the first entry is the primary model and the rest are fallbacks", `packages/omo-opencode/src/config/schema/categories.ts`), so a category that already uses the canonical chain and still carries a `fallback_models` key loses the chain. The reasoning plan runs in `replace-target` mode and its `sources` list is empty, so no backup of the target is taken and the previous content is not recoverable from `~/.omo`.

Two shapes, both silent:

- `models: [A, B]` + `fallback_models: [C]` becomes `models: [C]`, so the primary model is replaced by what used to be a fallback.
- `models: [A]` + `fallback_models: []` becomes `models: []`, so the category is left with no model at all and falls back to defaults.

### Steps to Reproduce

Against `dev` @ `e0746bc`, transform only (no config on disk needed):

```ts
import { transformReasoningUnification } from "./packages/omo-opencode/src/config-migration/reasoning-unification"

transformReasoningUnification({
  categories: { quick: { models: ["primary/model", "second/model"], fallback_models: ["legacy/fallback"] } },
})
// => { categories: { quick: { models: ["legacy/fallback"] } } }
```

End to end:

1. Write that `categories` block into `~/.omo/omo.jsonc` on a config that has no `2026-08-reasoning-unification` entry in `_migrations`.
2. Run `oh-my-openagent config migrate`.
3. Read `~/.omo/omo.jsonc` back: `categories.quick.models` is now the single legacy fallback.

The same input under `agents` is migrated correctly, which is what makes the category behaviour look unintended.

### Expected Behavior

Categories behave like agents: the declared chain keeps its order and its head, and legacy fallbacks are appended behind it.

```
models: ["primary/model", "second/model", "legacy/fallback"]
```

### Actual Behavior

The declared chain is dropped and only the legacy fallbacks survive. With an empty `fallback_models` the chain becomes an empty array.

### Environment

- oh-my-openagent: `dev` @ `e0746bc` (5.0.0-beta.56)
- Bun: 1.4.2
- macOS 26.6 (Darwin 25.6.0, arm64)

### Additional Context

The existing fixture (`packages/omo-opencode/src/config-migration/2026-08-reasoning-unification/fixture-input.json`) covers `model` + `fallback_models` for a category and `model` + `models` for an agent, but never `models` + `fallback_models` on a category, which is why this slipped through.

I have a one-line fix plus two regression tests ready and will open a PR against `dev`.

L4XB:
## Summary

- `config migrate` destroyed a category's model chain when the category carried both the canonical `models` array and a leftover `fallback_models` key: `models: [A, B]` plus `fallback_models: [C]` was rewritten to `models: [C]`, so the former fallback became the primary model. With an empty `fallback_models` the category was left with `models: []` and no model at all.
- Agents were already merged correctly. Categories now follow the same rule, so the migration is `[primary?, ...declared chain, ...legacy fallbacks]` for both kinds.
- Fixes #8191. Context for anyone converting agents or categories by hand: #8186.

## Changes

- `packages/omo-opencode/src/config-migration/reasoning-unification.ts`: `normalizeDefinition()` reads the existing `models` chain for categories too. The `kind === "agent"` guard predates `CategoryConfigSchema` accepting `models`, and with categories now declaring an ordered chain the guard silently dropped it. One line plus a comment saying why it matters.
- `packages/omo-opencode/src/config-migration/reasoning-unification.test.ts`: two regression tests, one for a chain plus legacy fallbacks and one for a chain plus an empty fallback list.
- `.omo/evidence/20260912-fix-8191/`: driver, before and after logs, red/green mutation output, related suites, typecheck.

## QA & Evidence

- **What was tested:** the real CLI entry point `runConfigMigrate` against a throwaway `HOME` holding a category with both keys and no `_migrations` marker, so the plan, journal, commit and target rewrite all ran.
  **Observed result:** on `origin/dev` the file on disk became `models: ["legacy/fallback"]`; on this branch it became `models: ["primary/model", "second/model", "legacy/fallback"]`. Both runs exit 0 and add the `2026-08-reasoning-unification` marker, so only the migrated document differs.
  **Artifact:** `.omo/evidence/20260912-fix-8191/driver.ts`, `before-fix.txt`, `after-fix.txt`
  **Why sufficient:** it is the same function `oh-my-openagent config migrate` calls, and it proves the loss and the fix on the file the user actually keeps.

- **What was tested:** the two new tests with the changed line reverted to `origin/dev`, then restored.
  **Observed result:** `4 pass / 2 fail` reverted, `6 pass / 0 fail` restored. The committed 2026-08 fixture test passes in both, because that fixture has no category carrying both keys, which is how this slipped through.
  **Artifact:** `.omo/evidence/20260912-fix-8191/mutation-red.txt`, `mutation-green.txt`
  **Why sufficient:** the tests fail for the intended reason and are not tautological.

- **What was tested:** `bun test` over `packages/omo-opencode/src/config-migration`, `packages/omo-config-core/src/migration`, `startup-migration.test.ts`, `packages/omo-opencode/src/config`, `packages/omo-config-core/src/schema`, and the `config-migrate` CLI test.
  **Observed result:** `317 pass / 0 fail`.
  **Artifact:** `.omo/evidence/20260912-fix-8191/related-suites.txt`
  **Why sufficient:** covers both callers of the transform and the engine that commits it.

- **What was tested:** `tsgo --noEmit -p packages/omo-opencode/tsconfig.json`.
  **Observed result:** exit 0, no output.
  **Artifact:** `.omo/evidence/20260912-fix-8191/typecheck.txt`

## Risks & Residuals

- **No live harness QA.** I cannot drive a real OpenCode session in this environment, so this is not an `opencode-qa` run. Accepted, with the CLI-level driver above as the substitute: the change is confined to the pure document transform shared by `config migrate` and the plugin startup migration.
- **Behaviour change is narrow.** A category with only `model` plus `fallback_models` migrates byte for byte as before, which the unchanged 2026-08 fixture pins. Only categories that already use `models` see a different result, and for them the old result was the data loss.
- **Duplicate heads are possible** when a category sets `model` and repeats the same entry as the head of `models`. That matches the existing agent behaviour and is left as is rather than adding dedupe in a fix PR.
- **Root `bun test` and `test:codex`** were not part of this evidence; the targeted suites and the typecheck above are. CI runs the full matrix.

## Automated Checks

```bash
bun test packages/omo-opencode/src/config-migration packages/omo-config-core/src/migration \
  packages/omo-opencode/src/startup-migration.test.ts packages/omo-opencode/src/config \
  packages/omo-config-core/src/schema packages/omo-opencode/src/cli/config-migrate.test.ts
# 317 pass, 0 fail

./node_modules/.bin/tsgo --noEmit -p packages/omo-opencode/tsconfig.json
# exit 0
```

---
## Summary by cubic
Fixes `config migrate` dropping a category's declared `models` chain whenever a leftover `fallback_models` key is present. Previously `models: [A, B]` plus `fallback_models: [C]` was rewritten to `models: [C]`, promoting a fallback to primary, and an empty `fallback_models` left the category with no model at all; categories now merge the same way agents do, preserving `[primary?, ...declared chain, ...legacy fallbacks]`. Fixes #8191.

**Bug Fixes**
- `normalizeDefinition()` now reads the existing `models` chain for categories, matching the agent path.
- Adds two regression tests: a chain with legacy fallbacks and a chain with an empty fallback list.
- Categories that don't declare `models` already migrate byte-for-byte as before, so only the data-loss case changes.
- A category setting both `model` and the same entry at the head of `models` can produce duplicate heads; that matches existing agent behavior and is left as-is.

<sup>Written for commit 162eedd7c25edfa19391e001da1473437c65de9a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/evidence/20260912-fix-8191/after-fix.txt
+++ b/.omo/evidence/20260912-fix-8191/after-fix.txt
@@ -1,0 +1,46 @@
+# bun run driver.ts on this branch (fix applied)
+
+--- ~/.omo/omo.jsonc before ---
+{
+  "categories": {
+    "quick": {
+      "models": [
+        "primary/model",
+        "second/model"
+      ],
+      "fallback_models": [
+        "legacy/fallback"
+      ]
+    }
+  }
+}
+--- oh-my-openagent config migrate ---
+status: migrated
+target: /private/var/folders/7_/jgfykr3d5b12qrh10lvz_l000000gn/T/omo-e2e-GG9L2T/.omo/omo.jsonc
+transform: {
+  "categories": {
+    "quick": {
+      "models": [
+        "primary/model",
+        "second/model",
+        "legacy/fallback"
+      ]
+    }
+  }
+}
+--- exit code 0 ---
+--- ~/.omo/omo.jsonc after ---
+{
+  "categories": {
+    "quick": {
+      "models": [
+        "primary/model",
+        "second/model",
+        "legacy/fallback"
+      ]
+    }
+  },
+  "_migrations": [
+    "2026-08-reasoning-unification"
+  ]
+}
--- a/.omo/evidence/20260912-fix-8191/before-fix.txt
+++ b/.omo/evidence/20260912-fix-8191/before-fix.txt
@@ -1,0 +1,42 @@
+# bun run driver.ts on origin/dev (unpatched)
+
+--- ~/.omo/omo.jsonc before ---
+{
+  "categories": {
+    "quick": {
+      "models": [
+        "primary/model",
+        "second/model"
+      ],
+      "fallback_models": [
+        "legacy/fallback"
+      ]
+    }
+  }
+}
+--- oh-my-openagent config migrate ---
+status: migrated
+target: /private/var/folders/7_/jgfykr3d5b12qrh10lvz_l000000gn/T/omo-e2e-LVO6Dx/.omo/omo.jsonc
+transform: {
+  "categories": {
+    "quick": {
+      "models": [
+        "legacy/fallback"
+      ]
+    }
+  }
+}
+--- exit code 0 ---
+--- ~/.omo/omo.jsonc after ---
+{
+  "categories": {
+    "quick": {
+      "models": [
+        "legacy/fallback"
+      ]
+    }
+  },
+  "_migrations": [
+    "2026-08-reasoning-unification"
+  ]
+}
--- a/.omo/evidence/20260912-fix-8191/driver.ts
+++ b/.omo/evidence/20260912-fix-8191/driver.ts
@@ -1,0 +1,22 @@
+import { mkdirSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
+
+import { runConfigMigrate } from "./packages/omo-opencode/src/cli/config-migrate"
+
+const home = mkdtempSync(join(tmpdir(), "omo-e2e-"))
+mkdirSync(join(home, ".omo"), { recursive: true })
+const target = join(home, ".omo", "omo.jsonc")
+writeFileSync(target, `${JSON.stringify({
+  categories: {
+    quick: { models: ["primary/model", "second/model"], fallback_models: ["legacy/fallback"] },
+  },
+}, null, 2)}\n`)
+
+console.log("--- ~/.omo/omo.jsonc before ---")
+console.log(readFileSync(target, "utf-8").trim())
+console.log("--- oh-my-openagent config migrate ---")
+const code = runConfigMigrate({ cwd: home, environment: { HOME: home }, output: (line) => console.log(line) })
+console.log(`--- exit code ${code} ---`)
+console.log("--- ~/.omo/omo.jsonc after ---")
+console.log(readFileSync(target, "utf-8").trim())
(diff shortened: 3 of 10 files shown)


L4XB:

Issue #8166, opened 2026-09-12
Published JSON schema declares top-level 'codegraph' but the Zod schema rejects it; error message names the wrong key


RaviTharuma:
### Prerequisites

- [x] I am writing in English
- [x] I have searched existing issues for duplicates
- [x] I am using the latest version
- [x] I have checked the documentation

### Bug Description

`oh-my-openagent doctor` reports my `~/.omo/omo.jsonc` as **invalid**, pointing at the `[senpi]` key, and states the failure affects **plugin startup**:

```
2. Invalid configuration
   Invalid omo config at /Users/<user>/.omo/omo.jsonc: [senpi],
   Affects: plugin startup
```

The offending block is:

```jsonc
"[senpi]": {
  "telemetry": {
    "enabled": false
  }
}
```

This block is **valid according to the schema the config file itself references**. From `assets/omo.schema.json` at `v5.0.0-beta.34`:

- `[senpi]` is a declared top-level property.
- `[senpi].properties` includes `telemetry`.
- `[senpi].properties.telemetry` is exactly `{"type":"object","properties":{"enabled":{"type":"boolean"}},"additionalProperties":false}`.

My value is `{"enabled": false}` — a boolean, the only key allowed, matching the declared type.

Verified mechanically against the published schema:

```
$ curl -s https://raw.githubusercontent.com/code-yeongyu/oh-my-openagent/v5.0.0-beta.34/assets/omo.schema.json -o omo.schema.json
$ node -e '
const fs=require("fs");
const sc=JSON.parse(fs.readFileSync("omo.schema.json","utf8"));
const sp=sc.properties["[senpi]"];
console.log("declared:", JSON.stringify(sp.properties.telemetry));
'
declared: {"type":"object","properties":{"enabled":{"type":"boolean"}},"additionalProperties":false}
```

So the runtime validator and the published schema disagree. Either the validator is wrong, or the schema advertises support that was removed — either way a config written against the documented schema is rejected.

The error message also ends with a trailing comma and no reason (`... : [senpi], `), so it names the key but never says *what* about it is invalid. Even when the rejection is correct, this message cannot be acted upon.

### Steps to Reproduce

1. Put this in `~/.omo/omo.jsonc`, alongside a `$schema` pointing at `v5.0.0-beta.34`:
   ```jsonc
   {
[...]

PR #8206, opened 2026-09-12
fix(config): name unrecognized keys in validation diagnostics

RaviTharuma:
## Summary

`validationDiagnostic` builds its message from each issue's `path`. zod reports `unrecognized_keys` with the offending names in a separate `keys` field and leaves `path` pointing at the *enclosing* object — which is empty at the root.

The unknown key therefore contributes an empty string to the joined list. With another violation present, the message renders as:

```
Invalid omo config at ~/.omo/omo.jsonc: agents.x.models,
```

A trailing separator with nothing after it. The unknown key is never named anywhere in the output.

## How this surfaced

On my machine `doctor` reported:

```
2. Invalid configuration
   Invalid omo config at ~/.omo/omo.jsonc: [senpi],
   Affects: plugin startup
```

`[senpi]` is valid — `OmoConfigSchema.safeParse({ "[senpi]": … })` passes on its own. It appeared in the message only because the empty slot from the unrecognized-key issue landed next to it, and `doctor` passes `validation.errors` straight through to the user.

I spent a while checking `[senpi].telemetry` against the published schema before realising the named key was not the failing one. The actual unknown key never appeared in any output.

## Change

Render `unrecognized_keys` issues from their `keys` field:

```ts
if (issue.code === "unrecognized_keys" && issue.keys !== undefined && issue.keys.length > 0) {
  const keys = issue.keys.map((key) => String(key)).join(", ")
  return location.length > 0 ? `${location}: unrecognized key(s): ${keys}` : `unrecognized key(s): ${keys}`
}
```

Before and after, same input:

```
ALT: "agents.x.models, "
NEU: "agents.x.models, unrecognized key(s): codegraph"
```

`issuePaths` keeps the same shape, so consumers that read the structured field are unaffected. Only the rendered message changes.

## Note on the trigger

The unknown key in my config was `codegraph`, which the **published** JSON schema still declares as a top-level property:

```
$ curl -s .../v5.0.0-beta.34/assets/omo.schema.json | jq '.properties | keys'
[…, "codegraph", …]
```

but which no longer exists anywhere in the source. That is a separate documentation-drift issue and this PR does not touch it — a config written against the published schema is still rejected, just with an accurate message now.

Worth noting that the standalone path already handles this well: an unknown key *alone* produces a clean `unknown-keys` diagnostic naming `codegraph`. Only the `validation` path, reached when a real violation is also present, loses the name.

## QA & Evidence

```
$ bun test packages/omo-config-core/src/loader/unknown-keys.test.ts
 6 pass  0 fail

$ bun test packages/omo-config-core/src/
 223 pass  0 fail
Ran 223 tests across 34 files.
```

Reverting only `loader.ts` and re-running the file gives `5 pass, 1 fail`, so the new test covers the change rather than passing incidentally.

```
$ bun run typecheck
errors without patch: 4
errors with patch:    4
errors in loader.ts:  0
```

The 4 are pre-existing on `dev` (`timestamp-output.ts`, `process-cleanup.test-helpers.ts`, `utils/src/runtime/file.ts`) and untouched here.

## Risks & Residuals

- Message-only change; no behavioural or schema change.
- `issuePaths` is unchanged, so structured consumers see the same data.
- The new test uses a deliberate violation (`models: "not-an-array"`) to force the validation path, since an unknown key alone does not reach it.

## Screenshots

n/a — CLI diagnostic text, covered by the before/after above.

## Automated Checks

- [x] `bun test packages/omo-config-core/src/` — 223 pass
- [x] `bun run typecheck` — no new errors
- [ ] `bun run build` — not run; `bun install` fails on this machine in the `lsp-daemon` prepare script (`build:declarations` exit 2), unrelated to this change. Flagging rather than claiming a check I did not complete.

## Related Issues

Refines #8166. That issue reports `doctor` rejecting `[senpi].telemetry`; investigating it showed `[senpi]` was never the failing key — the message was pointing at the wrong thing. I have added a comment there with the corrected analysis.

---
## Summary by cubic
Validation diagnostics now name unrecognized keys from Zod's `keys` field instead of rendering an empty path, so messages no longer trail off with a separator or point at the wrong key. Refines #8166.

- Rebased onto `dev` and regenerated the committed Senpi extension bundles on Node 24 so the CI `--check` freshness gate passes; structured issue paths are unchanged.
- Adds coverage for an unknown `codegraph` key alongside invalid `agents.x.models` data.
- Records the rebase, bundle, and Node 24 `--check` QA under `.omo/evidence/omo-senpi-adapter/`.

<sup>Written for commit eaebbb764d8da8d9e484df13238a8adb5ffd7c87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


RaviTharuma:
Friendly ping — `fix(config): name unrecognized keys in validation diagnostics` is ready for review when you have a moment. Happy to adjust if anything looks off.

Issue #8235, opened 2026-09-13
Release binary diet: --splitting + full minify, embedded payload slim, provider registration adoption, retired browser CLIs, dependency cleanup


code-yeongyu:
## Summary
Track one coordinated dependency-diet release effort for the compiled native artifact, embedded payload, provider surface, browser tooling, workspaces, and QA dependencies.

## Reproduction
Build the darwin-arm64 release binary and inspect its compiled entries and embedded assets. Compare the current compile configuration with a probe using `--splitting --minify --keep-names`; inspect shipped skills, prompts, workspace consumers, package dependencies, and QA scripts.

## Expected
The release artifact contains only runtime-required code and assets, provider modules load in compiled binaries, shipped guidance uses the maintained in-process browser path, and dependency/workspace surfaces are minimal and deduplicated.

## Actual
The darwin-arm64 release binary is 130,850,802 B: Bun runtime 61,884,464 B + main JS bundle ~21.2 MiB + a second entry (RPC session worker) that duplicates ~99% of the same modules (~21.1 MiB) + ~25 MiB embedded sidecars. Compile uses `--minify-whitespace` only; compile probes with `--splitting --minify --keep-names` cut the JS payload from 42.9 MiB to 16.2 MiB. Bedrock/Cursor/Devin providers cannot load inside the compiled binary because the launcher only registers OAuth flows while the engine's lazy provider modules need static registration. `docs/images` screenshots (2.2 MiB) and plugin README/NOTICE/LICENSE are embedded without a runtime reader (docs/, examples/, CHANGELOGs are read at runtime and stay). Shipped skills/prompts still instruct agents to use `agent-browser` and `npx playwright` (retired; replacement is the in-process Bun.WebView / playwright-core script path). `pi-goal`/`pi-webfetch` workspaces have no consumers. `@code-yeongyu/comment-checker` npm dep installs 261 MB while the hook downloads the binary from GitHub anyway. Shipped QA scripts require devDependencies (puppeteer-core, node-pty, @xterm).

## Evidence
- Release binary: 130,850,802 B total; Bun runtime 61,884,464 B; main JS ~21.2 MiB; RPC session worker ~21.1 MiB; embedded sidecars ~25 MiB.
- Probe compile: `--splitting --minify --keep-names` reduces JS payload from 42.9 MiB to 16.2 MiB.
- Current compile flag: `--minify-whitespace` only.
- The worker entry duplicates ~99% of the main entry's modules.

[...]

PR #8248, opened 2026-09-13
fix(native): package comment-checker and handle Bun missing-module errors

### LINKED ISSUE #8247: [Bug]: omo-ai omits comment-checker and leaks Bun ResolveMessage
## Summary

The native `omo-ai` package does not declare `@code-yeongyu/comment-checker`,
although its bundled Senpi extension requires that package after successful
write-like tool results. On Bun, the missing package also escapes the intended
fallback because `ResolveMessage` is not an `Error` instance.

The visible error is:

```text
Extension "/opt/homebrew/lib/node_modules/omo-ai/plugin/extensions/omo.js" error:
ResolveMessage: Cannot find module '@code-yeongyu/comment-checker' from
'/opt/homebrew/lib/node_modules/omo-ai/plugin/extensions/omo.js'
```

## Environment

- `omo-ai`: `5.0.0-0.beta.62`
- Senpi: `2026.9.13`
- Bun: `1.4.0`, reporting `1.4.0-canary.1+45ee9556a` in CLI error output
- Node: `v24.20.0`
- macOS, Apple Silicon
- Installed globally using npm under `/opt/homebrew/lib/node_modules`

## Reproduction

1. Install `omo-ai@5.0.0-0.beta.62` in a clean prefix or consumer, without a
   separately installed `@code-yeongyu/comment-checker`.
2. Run OmO under Bun and perform a successful `write`, `edit`, or `apply_patch`.
3. The comment-checker tool-result hook attempts to load the missing package,
   then propagates the `ResolveMessage` instead of checking PATH or disabling
   the unavailable checker.

I also reproduced the missing dependency by packing `packages/omo-native`,
installing the tarball into a fresh consumer, and resolving
`@code-yeongyu/comment-checker` with a `createRequire` anchored at the installed
`omo-ai/package.json`. The resolution fails without relying on the interactive
session.

## Expected behavior

- A normal native installation includes the checker required by its extension.
- If the package is unavailable, a checker on PATH can still be used.
- If neither source is available, the component disables itself once without
  reporting an extension exception or breaking successful tool results.

## Cause

`packages/omo-native/package.json` declares only `@babel/parser` and
`@code-yeongyu/senpi`. The root workspace does declare comment-checker, so
workspace hoisting masks the published-package omission.

In `packages/omo-senpi/src/components/comment-checker/resolver.ts`,
`resolvePackageApiBinary()` returns `null` only when the caught value is an
`Error`. A real Bun missing-module exception has:

```json
{
  "name": "ResolveMessage",
  "code": "MODULE_NOT_FOUND",
  "importKind": "require-call",
  "isError": false
}
```

It is rethrown by both the package resolver and the component boundary.

## Workaround and proposed correction

Installing `@code-yeongyu/comment-checker@0.8.0` globally restored package
resolution, extension loading, and actual checker execution in a fresh Bun
process.

The proposed fix declares the native runtime dependency and recognizes the
structured `MODULE_NOT_FOUND` condition without swallowing unrelated thrown
values. Regression coverage exercises a real missing package, PATH fallback,
one-time component disabling, and installation of the packed native package.

This differs from #4162, which concerned an unpublished dependency version,
and #6396, which concerns child-process stdin EPIPE.

gunggme:
## Summary

Ship `@code-yeongyu/comment-checker` with native `omo-ai` and handle Bun's
non-Error `MODULE_NOT_FOUND` values without losing the existing PATH fallback.
The checker becomes unavailable once, rather than failing successful
write-like tool results, when neither the package nor a PATH binary exists.

## Conflict-resolution update

Merge commit `bace8d10b` incorporates current `dev` (`db06fbe7a`) while preserving
the original PR history. It does not rewrite history or merge this PR upstream.

- Rebuilt the four conflicting generated extensions from both source histories
  with the actual CI toolchain: Bun 1.4.2, Node 24, Linux/amd64.
- Regenerated the lockfile after upstream removed the root checker dependency.
  The native `0.8.0` dependency and its package-resolution record remain.
- The original resolver and regression tests are unchanged from the reviewed
  PR. All helper bundles now match `dev`; only `omo.js` differs.
- Relative to current `dev`, the compiled main body adds only the missing-module
  predicate. The main bundle is 1,161,447 bytes versus upstream's 1,161,412,
  below the unchanged 1,180,000-byte budget.

The PR now has seven changed paths: native manifest and lockfile, two native
tests, checker resolver and missing-package test, and generated `omo.js`.

## QA & Evidence

| Verification | Result on the resolved tree |
| --- | --- |
| Checker and native-consumer regressions, bundle purity and size | 38 pass, 0 fail across 8 files |
| Complete repository typecheck | Exit 0 |
| Native payload build | Exit 0 |
| Complete native package suite | 302 pass, 0 fail across 33 files |
| Canonical generated-artifact freshness check | Exit 0 |
| Sanctioned QA driver self-test | Exit 0 |
| Real Senpi write without package or PATH checker | Success; one warning; no extension exception |
| Real Senpi write with only a PATH checker | Success; actual checker feedback |
| Full extension driver | PASS; checker PASS; both isolation certificates true |

Real QA used offline scripted providers and actual file writes. The missing
scenarios used an isolated component bundle with an empty node_modules
directory and did not mock module resolution. The real PATH binary was
resolved from the native package that now declares it. The generic full-driver
fixture used that native dependency directory for its Node-side lookup; fresh
packed-consumer tests separately cover default installation resolution.

Native tests and live QA ran as a non-root user with Python 3 and pgrep
available. Initial minimal-image/tooling failures were recorded, then the
environment was corrected; no product test was changed or weakened.
No host home or credentials were mounted. Driver subprocesses and sandboxes
were cleaned, and the providers made no live external-model calls.

Raw command receipts and full stdout/stderr, the merge plan, and actual-harness
JSON are retained locally under
`.omo/evidence/omo-senpi-adapter/20260914-comment-checker-conflicts/`.
The summaries here are public because those local evidence files are not in
the source diff.

## Remaining known failure

The full `bun run test:senpi` build and typecheck succeeded, but the test stage
reported two existing formatter CLI fallback failures among 3,468 tests:

- Best-effort unavailable formatter: expected one warning, received none.
- Required unavailable formatter: expected an unavailable error string.

An independent, unchanged `db06fbe7a` worktree reproduced both failures in a
single-file run: 3 pass, 2 fail. Formatter implementation, tests, immediate
helpers, preloads, and root manifest are byte-identical to this resolution;
the baseline used its own frozen dependencies. This is a narrow baseline
reproduction, not a claim that the entire baseline suite was rerun. Formatter
code is outside this conflict-resolution change and was left untouched.

The earlier bundle-size failure was measured with the older Bun 1.4.0
toolchain. Current CI-pinned generation and its size test pass; the size
ceiling was not raised.

## Commands run

```sh
bun install --frozen-lockfile --ignore-scripts
bun test packages/omo-senpi/src/components/comment-checker packages/omo-native/test/package-shape.test.ts packages/omo-native/test/packed-install.test.ts packages/omo-senpi/src/bundle-size.test.ts packages/omo-senpi/src/bundle-purity.test.ts
bun run typecheck
bun run build:omo-native
bun test packages/omo-native/test
node packages/omo-senpi/scripts/qa/drive.mjs --self-test
bun run test:senpi
```

The final command's known formatter failures are disclosed above, not presented
as a green suite or merge approval.

## Related Issues

Closes #8247.

---
## Summary by cubic
Ships `@code-yeongyu/comment-checker` with the native `omo-ai` package and fixes comment-checker resolution under Bun. Native installs previously omitted the runtime dependency, and Bun's `ResolveMessage` (a non-`Error` value) was rethrown instead of falling back to PATH or disabling the checker.

**Notes**
- The checker dependency is pinned to `0.8.0` in the native manifest and lockfile.
- The resolver treats the structured `MODULE_NOT_FOUND` condition as a missing package while still propagating unrelated thrown values.
- Tests cover PATH fallback, no-binary handling, one-time disabling, unrelated thrown values, and a packed consumer resolving the checker API and executable.
- Extension artifacts were regenerated with the CI-pinned Bun 1.4.2 builder after syncing with `dev`.
- The full Senpi gate still fails only on the pre-existing bundle-size ceiling; the unmodified base tree fails the same check.

Closes #8247.

<sup>Written for commit bace8d10b2044ded0d5f0978669ca79163c56053. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gunggme:

code-yeongyu (changes requested):
Thank you for this, and sorry it sat without a review for so long. The diagnosis is exactly right on both counts, and I reproduced it against the published `omo-ai@5.0.0-0.beta.86`: installed into a clean npm prefix with no `@code-yeongyu/comment-checker` and no checker on `PATH`, one scripted `write` under Bun 1.3.14 ends in

```
Extension error (.../omo-ai/plugin/extensions/omo.js): ResolveMessage: Cannot find module '@code-yeongyu/comment-checker' from '.../omo-ai/plugin/extensions/omo.js'
```

while the same install under Bun 1.4.0/1.4.1/1.4.2 or Node degrades to the intended `comment-checker binary unavailable` warning. Measured directly: `ResolveMessage instanceof Error` is `false` on Bun 1.3.14 and `true` from 1.4.0 on. So the leak is real for every `bun add -g` install that re-executes under a 1.3.x bun, and the missing checker is real on every native install.

I have to request changes on the dependency half, though.

## Must change

1. `packages/omo-native/package.json` — `"@code-yeongyu/comment-checker": "0.8.0"` as a runtime dependency.
   `npm view @code-yeongyu/comment-checker@0.8.0 dist.unpackedSize` is **267,670,796 bytes** (9 files: every platform's binary in one tarball). Declaring it makes every `omo-ai` install about 255 MiB larger. #8256 (merged 2026-09-14, one day after this PR opened) removed that same dependency from the OpenCode plugin for exactly this reason and replaced it with the pinned GitHub-release downloader in `packages/omo-opencode/src/hooks/comment-checker/downloader.ts` (`COMMENT_CHECKER_VERSION = "0.8.0"`), and #8235 lists "remove the comment-checker npm dependency while retaining the hook's binary download behavior" as the release direction. The native edition needs the checker the same way: cached binary, then a lazy download of the pinned release for the host platform, then `PATH` — never the npm payload.

2. `packages/omo-native/test/package-shape.test.ts` (lines 48-57) and `packages/omo-native/test/packed-install.test.ts` (lines 32-37) pin the inverse of the contract we want: they assert the dependency is present and resolvable from the packed consumer. The guard should assert that the manifest stays free of the dependency, that the built extension carries the pinned-release downloader, and that the resolver on a packed consumer returns a cached binary without throwing when the npm package is absent.

## Keep

- `packages/omo-senpi/src/components/comment-checker/resolver.ts` line 68 — the `MODULE_NOT_FOUND` predicate is the right fix and matches what `comment-checker-core/src/runner.ts` already does for the OpenCode edition. A short comment that this covers Bun 1.3.x, where `ResolveMessage` is not an `Error`, would help the next reader.
- `comment-checker.missing-package.test.ts` — good coverage (real module resolution, PATH fallback, one-time disable, unrelated values still propagate).

## What happens next

The design moved under this PR after it opened, which is on us, not you. Rather than ask you to rewrite it, I am carrying the fix in a follow-up PR that keeps your resolver predicate and tests with `Co-authored-by` credit, adds the native download path shared with the OpenCode edition, and adds the inverse package-shape guards. I will link it here and close this PR once it merges; a review from you on that PR would be very welcome.

code-yeongyu:
Thank you for finding this and for the CLA — the diagnosis was right, and the dependency half of your fix is what shipped.

#8745 carries it: `@code-yeongyu/comment-checker` declared as an exactly pinned runtime dependency in `omo-native`, plus a manifest audit so it cannot disappear again. It was opened separately because this branch had drifted ~975 commits behind `dev` (4 conflicting files) and because the resolver change was left out: on Bun 1.4.0 and 1.4.2 a missing module raises `ResolveMessage`, which IS an `Error` instance, so the existing `error instanceof Error` branch already catches it — the `code === "MODULE_NOT_FOUND"` test would be dead code.

Closing this as superseded; #8247 closes with #8745.

code-yeongyu:
Superseded by #8748, which carries your resolver predicate and four of your regression tests with `Co-authored-by` credit, and closes the dependency gap by downloading the pinned release (as the OpenCode edition does since #8256) instead of declaring the ~255 MiB npm package. Thank you again for the report and the fix - the Bun 1.3.x measurement in that PR confirms your `MODULE_NOT_FOUND` predicate was required. Closing this one in favor of #8748; a review there is welcome.

code-yeongyu:
#8748 is merged as `01c1d49469bd95310fb8583a61000b577367ddee` with your resolver predicate and four of your regression tests, credited via `Co-authored-by`. The Bun 1.3.14 measurement in that PR (`ResolveMessage instanceof Error === false`) confirmed the predicate was required. The dependency half went the other way for the size reason described above: the native edition now downloads the pinned release the way the OpenCode edition does. Thank you for the report and the fix.

PR #8277, opened 2026-09-14
chore(deps): align the TUI fork and dedupe dependencies

code-yeongyu:
## What and why

- Resolve the existing pi-tui imports to the published fork matching the exact engine pin; remove the three dead Pi-family overrides while preserving unrelated overrides.
- Apply the requested dependency versions and regenerate the lockfile. `bun dedupe` removes 17 duplicate versions.
- Remove the separate `yaml` development dependency without dropping its three test consumers: reuse `js-yaml` with YAML 1.2 core semantics, including dedicated octal/NaN regression tests.
- Update the isolated package-resolution fixture for npm aliases, and keep the extension inside its unchanged byte budget by taking the extra compression pass only where it is needed.

## Current head

Rebased onto current `origin/dev` (`59123ea17`) and force-pushed. Head `e98a75fa7` rebuilds Senpi extension fingerprints after that rebase. GitHub Actions at this head: 23 pass / 7 skip, including typecheck, build, Ubuntu/macOS/Windows test shards, and Senpi compatibility.

Dedicated-host re-run of fresh install, `test:fast`, `test:senpi`, adapter `tsgo`, `bun outdated`, and compiled QA is still pending a reachable test runner.

## Evidence

- Fork-alignment RED: 4 expected failures; GREEN covered by 111 passing focused tests with 0 failures.
- YAML-core correction: 2 expected RED failures, then 67 passing tests with 0 failures.
- Fresh and frozen installation, explicit engine postinstall, both adapter typechecks, root typecheck, root build, extension build and the fingerprint check all pass on the prior dedicated-host capture.
- Senpi suite (prior dedicated-host head): 3,437 pass / 32 skip / 0 fail. OpenCode edition directory: 8,424 pass / 0 fail.
- Main extension: 1,133,751 bytes against the unchanged 1,180,000-byte ceiling.
- Compiled-binary QA (prior head) passes the dependency-audit RPC and extension cases, and a real session performs a memory write, renders the notice with its path and commit, and rejects a traversal path.
- Upstream 0.85.1 mutation keeps both adapter typechecks and 63 renderer tests green, so the exercised API is currently upstream-compatible; the fork alias is retained for engine alignment.
- All requested bumped packages disappear from `bun outdated --filter '*'` on the prior capture.
- Published `@code-yeongyu/senpi-tui@2026.9.13` and `@code-yeongyu/senpi-ai@2026.9.13` still exist; `dist-tags.latest` is now `2026.9.15`. The pin stays on `2026.9.13` to match the engine.

## Compression scope

The bumps pushed the main bundle past its budget, so a second Terser pass was applied everywhere. That doubled work on five sidecars that already fit and moved the artifact-freshness round-trip from about 40s to 52s against its 60s bound. The extra pass is now taken only for the main entry, with a regression pinning the forwarded per-bundle budget; the round-trip returns to about 105s for the full 13-test file and passes three consecutive runs.

## Known gaps

- The broad fast suite still reports unrelated timeout and cleanup-cascade failures on a loaded machine, plus one pinned structural-search binary that only exists once provisioned into the isolated home. These are reported, not claimed green, and are not caused by these dependency changes.
- The prescribed OpenTUI 0.5.11 package pins `marked@17.0.1` while the pinned engine uses `18.0.11`. The requested changes reduce three versions to two, not one. A forced transitive major override is intentionally not introduced.

Refs #8235

Plan: .omo/plans/omo-dependency-diet.md


code-yeongyu:
Updated to 41ab507af. Two generated-only reconciliations with dev landed without rewriting history, then the extra Terser pass was scoped to the main bundle so the artifact-freshness round-trip keeps its headroom. Senpi suite: 3,437 pass / 32 skip / 0 fail. OpenCode directory: 8,424 pass / 0 fail. Typecheck, root build, Senpi build and fingerprint check all pass, and the main bundle stays at 1,133,751 bytes against the unchanged 1,180,000-byte cap. The broad fast suite still shows unrelated timeout and cleanup failures; they are reported, not claimed green.

PR #8298, opened 2026-09-14
fix(rules): name agent_type in the bundled Hephaestus spawn examples

LilMGenius:
A Codex session whose only spawn guidance is the bundled Hephaestus rule loses every subagent spawn. The rule is injected at SessionStart, the session copies its `spawn_agent` example, and the spawn guard shipped in 5.0.0-beta.61 denies it. One long orchestration run that had been spawning workers that way for two days stopped spawning mid-lap the moment beta.62 loaded.

**Failing input** - the rule's own example, filled in:

```json
{"message":"TASK: act as a explorer. ...","fork_context":false}
```

```
Tool call blocked by PreToolUse hook: LazyCodex requires an explicit registered agent_type:
explorer, ... plan. Received no agent_type. ... Describing a role in message does not select its TOML.
```

**Root cause.** `spawnRoleDenial` in `components/ulw-loop/src/spawn-role-guard.ts` admits a spawn only when `tool_input.agent_type` is in `LAZYCODEX_SPAWN_ROLES`, and all six examples across `components/rules/bundled-rules/hephaestus/{gpt-5.5,gpt-5.6,gpt-6}.md` spell the call without `agent_type`. `ultrawork` and `ulw-plan` already teach the admitted shape; the bundled rule is the one document that still teaches the denied one, and it is the one every session receives without asking for a skill.

**RED / GREEN** - every example in the shipped rule files, run through the real guard via `applySpawnGuards`:

| example | dev `db06fbe7a` | this branch |
| --- | --- | --- |
| gpt-5.5 v1 | denied | admitted |
| gpt-5.5 v2 | denied | admitted |
| gpt-5.6 v1 | denied | admitted |
| gpt-5.6 v2 | denied | admitted |
| gpt-6 v1 | denied | admitted |
| gpt-6 v2 | denied | admitted |

RED on dev stops at the first row with `expected '{"hookSpecificOutput":{"hookEventName...' to be ''`. The 45 existing rows of that matrix pass on both sides, so the new assertion is not loosening the guard - it is the guard, read against the documents that teach it.

**The fix** adds `"agent_type":"<role>"` to the six examples and nothing else. The new case lives in the existing `spawn-role-matrix.test.ts` beside `pins explicit selectors to actual bundled TOML names`, which already reads a sibling component's shipped files, so it fails here rather than at someone's runtime if a rule drifts again.

**Scope deliberately untouched.** The guard itself is correct and unchanged. `components/ulw-loop/skills/ulw-loop/SKILL.md` has a `Codex Tool Mapping` row that says `do NOT set agent_type/model/reasoning_effort`, which contradicts the same file's own routing section - a separate case with a separate reproduction, held until this shape is reviewed rather than filed alongside it.

**Disclosure.** `components/rules` is red on `dev` already: 13 failing tests across 6 files, an identical failure set before and after this branch (`diff` of the two `FAIL` lists is empty). `components/ulw-loop` is 630 pass with one pre-existing failure, `codex-goal-snapshot.test.ts > reads from sample fixture path`, also red on `dev`. `tsc --noEmit` and `biome check` are clean on the changed file.

Reported first as an issue on the distribution mirror since a PR cannot land there; this is the same case against the source tree, so that issue can be closed once this is decided.

---
## Summary by cubic
Previously, bundled Hephaestus `multi_agent_v1.spawn_agent` examples omitted the required registered `agent_type` and were rejected by the spawn guard. They now include the selector; flat `multi_agent_v2` examples remain unchanged because that surface may not support it.

- Adds regression coverage that reads all three bundled variants and runs each documented v1 example through the real guard.

<sup>Written for commit ab0f945d90a2f69376bb6c99bdf3785571853f87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8298?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (commented):
CI failure root cause: `aggregate.test.mjs` line 70 test (`#given shipped generated skills and Hephaestus rules #when spawn parameters are scanned #then unsupported role and model arguments are absent`) conflicts with the changes here.

**The issue:** `findSpawnAgentCallsWithUnsupportedParameters` in `aggregate-plugin-fixture.mjs` (line 237-238) treats `agent_type` as unsupported when the call uses bare `spawn_agent()` (no `multi_agent_v1.` receiver). The Hephaestus rules use bare `spawn_agent()`, so the newly-added `agent_type` is flagged as a violation across all 3 files (gpt-5.5, gpt-5.6, gpt-6).

**The contradiction:** `spawn-role-guard.ts` *requires* `agent_type` to admit a spawn, but the aggregate fixture *rejects* `agent_type` in non-`multi_agent_v1` calls. The guard and the test disagree on whether `agent_type` is valid in bare `spawn_agent()`.

**To fix:** The `findSpawnAgentCallsWithUnsupportedParameters` filter logic needs updating to allow `agent_type` in bare `spawn_agent()` calls as well (since the spawn guard actually requires it). Specifically, the condition at line 237-238:
```js
(name === "agent_type" && (direct || !allowsObjectAgentType))
```
should probably drop the bare-call restriction for `agent_type`, or the Hephaestus rules should use `multi_agent_v1.spawn_agent()` syntax. The former seems more correct given the guard behavior.

LilMGenius:
Thank you - the failure is real and it was mine. I verified the change by running the two component suites that own the files I touched and never asked which suites *read* them, so `plugin/test/aggregate.test.mjs` never entered the loop. The same miss cost a second round one directory over: my formatting pass resolved biome from the plugin root instead of from `components/ulw-loop`, so that component's own `npm run check` still rejected the file. Green at `ab0f945`, though not by the repair the review proposes, and the reason is in your own tree.

**Checked the proposed repair before taking it.** Dropping the bare-call restriction on `agent_type` would have to delete an assertion that is pinned deliberately:

```js
// plugin/test/aggregate.test.mjs:40
spawn_agent({"message":"invalid flat role","agent_type":"explorer","fork_turns":"none"})
```

and `plugin/test/sync-skills-test-support.mjs:61` records why it is invalid rather than merely unsupported:

> On the v2 surface `agent_type` may be ABSENT from the spawn schema (verified 2026-07-11: only `fork_turns`/`message`/`task_name`) - when absent, omit it and describe the role inside `message`

So the fixture is not disagreeing with the guard by accident. It encodes that the v2 schema may not carry the field at all, and `spawnRoleDenial` already has the matching answer for that case in its own denial text: stop and report incompatible role routing rather than spawn.

**What I did instead** - the second option you named. `agent_type` stays on the three `multi_agent_v1.spawn_agent` examples and is reverted on the three flat ones, and the new matrix assertion is scoped to `multi_agent_v1` calls. That is also where the denial I reproduced came from; the log in the report is a `multi_agent_v1` session. No fixture, no assertion and no guard logic is touched.

**Measured**, `packages/omo-codex/plugin`, `node --test test/*.test.mjs`:

| | tests | pass | fail | `aggregate.test.mjs:70` |
| --- | --- | --- | --- | --- |
| `341e2b7`, what you reviewed | 340 | 338 | 2 | fail, 3 violations |
| `ab0f945` | 340 | 339 | 1 | pass |

The one remaining failure is `subagent-limit-migration.test.mjs:266`, which is red on `dev` at `db06fbe7a` on its own (26 pass / 1 fail) and is not from this branch. `npm run check` in `components/ulw-loop` exits 0 and its suite is 46/46. All checks on the pull request are green, `codex-compatibility` included on ubuntu, macos and windows.

**Left open rather than patched.** A flat v2 spawn with no `agent_type` is still denied by the guard, so on a v2 session the rule's remaining example teaches a shape the guard refuses - and the `Codex Tool Mapping` row in `components/ulw-loop/skills/ulw-loop/SKILL.md` still says `do NOT set agent_type/model/reasoning_effort`, which contradicts the same file's routing section. That is one question about what v2 should do, not a rule-file defect, so it is not in this diff. Happy to file it separately if it is worth fixing.

PR #8397, opened 2026-09-16
fix(compaction): keep the post-compaction autocontinue on the session working model

mradwankhalil:
## Summary

- Keeps the post-compaction autocontinue on the session's **working model** when the agent's `compaction.model` differs from it.
- OpenCode's synthetic continue turn inherits the summarize request's model (`session/compaction.ts`: `model: userMessage.model`), so with a per-agent `compaction.model` configured the *compaction* model silently ran the continuation turns. The existing guard only matches `agent === "compaction"` and can never catch a model-only mismatch.
- On mismatch the handler now disables OpenCode's synthetic continue and re-issues the same continue prompt through `session.promptAsync` with the working model, the working agent, and the `compaction_continue` metadata marker.
- New helper module `plugin/compaction-autocontinue-reroute.ts` + three regression tests.

## Problem

With `agents.<agent>.compaction.model` configured (preemptive compaction enabled), a live session was observed continuing **on the compaction model** after auto-compaction. Message rows from that session's `opencode.db`:

| time (UTC) | role | model | notes |
| --- | --- | --- | --- |
| 17:23:18 | user | `zai-coding-plan/glm-5.3-flash` | synthetic compaction trigger (carries the `compaction` part, `auto: true`) |
| 17:24:20 | assistant | `zai-coding-plan/glm-5.3-flash` | summary (`agent: "compaction"`) — expected |
| 17:31:38 | user | `zai-coding-plan/glm-5.3-flash` | synthetic continue: "Continue if you have next steps…" |
| 17:31:38 → 17:33:25 | assistant ×5 | `zai-coding-plan/glm-5.3-flash` | **task continuation ran on the compaction model** |
| 17:36:11 | user | `opencode-go/deepseek-v4.1-flash` | next real prompt returns the session to its working model |

### Root cause chain

1. Preemptive compaction calls `client.session.summarize({ providerID, modelID: <compaction.model>, auto: true })` (`hooks/preemptive-compaction-trigger.ts`).
2. OpenCode stores the compaction trigger user message with that model (`SessionCompaction.create`).
3. OpenCode's autocompaction creates the synthetic continue message with `model: userMessage.model` — i.e. the summarize model — and gates it only on the `experimental.compaction.autocontinue` plugin hook (`packages/opencode/src/session/compaction.ts`, around lines 497–548 in v1.18.31).
4. `createCompactionAutocontinueHandler` disables the continue only when `isCompactionAgent(input.agent)`. When the trigger's agent is the working agent (normal case) and only the **model** differs, the guard cannot match, so the continue runs on the compaction model.

## Fix

`plugin/compaction-autocontinue-reroute.ts`

- Resolves the **trigger model** (newest message carrying a `compaction` part) and the **working model** (nearest non-compaction message, via the existing `findNearestMessageWithFieldsFromMessages`, which already skips compaction messages).
- `createCompactionAutocontinueHandler` now, after the existing checkpoint restores:
  - sets `output.enabled = false` to drop OpenCode's poisoned synthetic continue, then
  - re-issues the identical continue prompt with `session.promptAsync({ body: { agent, model: <working>, parts: [{ type: "text", text: <same continue text>, metadata: { compaction_continue: true } }] } })`.

Behavior matrix:

| condition | behavior |
| --- | --- |
| models equal | unchanged |
| message lookup fails / no working model resolvable | unchanged (fail-open, never breaks the flow) |
| models differ | rerouted to the working model; if the re-issue call itself fails, no continue is issued (preferable to continuing on the wrong model) and the failure is logged |

Wiring: `testing/create-plugin-module.ts` passes `client` and `directory` into the handler options. Both are optional — without a client the handler keeps the previous behavior, so existing unit tests and non-client call sites are unaffected.

## Notes / limitations

- The SDK `session.promptAsync` body has no `variant` field, so the re-issued continue runs at the model's default effort; model selection itself is exact.
- Scope is limited to the post-compaction autocontinue; summaries still run on the configured `compaction.model`.
- Applies to every agent with a `compaction.model`, not just one.

## Verification

- `bun test src/index.compacting.test.ts src/index.compaction-model-agnostic.static.test.ts` → **16 pass / 0 fail**, 52 expect() calls (3 new cases: reroute on mismatch; unchanged on equal models; unchanged on lookup failure)
- `bunx tsgo --noEmit -p packages/omo-opencode/tsconfig.json` → passed
- `bun run build` → passed
- Live reproduction recorded before the fix (table above); the guard logic is additionally covered by unit tests.

Only the intended files are included (4 files, +308/−1); unrelated local worktree edits were not staged.

---
## Summary by cubic
Keeps post-compaction autocontinue on the session's working model when the agent's `compaction.model` differs.

- Compares the compaction trigger's model with the working model; on mismatch, disables OpenCode's synthetic continue and reissues the same prompt via `session.promptAsync` on the working model.
- Resolves the working model from the newest real assistant turn, skipping compaction summaries and internal continue prompts so rerouting can't cascade.
- Equal models and message lookup failures keep the existing flow; if reissuing fails, no continue is sent instead of continuing on the wrong model.
- Summaries still run on the configured `compaction.model`; only the continue prompt is rerouted.

<sup>Written for commit fc569ee60ea4d6e18534f45517cefe29f016e843. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Rebased onto `v5.0.0-beta.72` (2026-09-18)

The fix has been ported forward from `beta.62` to the current release so it can merge cleanly:

- 8 local fix commits cherry-picked onto `v5.0.0-beta.72`; the only conflict was `changes.md` (changelog) and the base version was kept there.
- The generated-artifacts commit (lockfile + extension bundles) was dropped and regenerated instead (`bun install` + `bun run build`) — replaying it collides with extension files this release removed.
- `packages/omo-opencode/src/index.compacting.test.ts`: **14 pass / 0 fail**; typecheck clean on `omo-opencode`, `omo-senpi`, `omo-codex`.
- Verified this fix is **not** present upstream at `v5.0.0-beta.72` (`compaction-autocontinue-reroute` absent from the tree), so it is still required.

Reproduction this addresses: after auto-compaction, the synthetic continuation ran on the **compaction** model instead of the session working model (`session/compaction.ts` stamps the trigger message with the per-agent `compaction.model`, and the continuation inherits `userMessage.model`). The added `compaction-autocontinue-reroute` module disables core's synthetic continue and re-issues the prompt via `session.promptAsync` with the working model.


mradwankhalil:
Root cause (the synthetic continue inheriting the summarize model) is now tracked upstream at https://github.com/anomalyco/opencode/issues/49386 — note the repo moved from sst/opencode to anomalyco/opencode. This PR is the plugin-side mitigation until core can resolve the continue model from the session instead.

mradwankhalil:
Follow-up commit pushed: `a98b94dd8` — expands the working-model resolution, because the current one is still reachable by two message classes that make the reroute sticky:

1. **Compaction summaries** (`agent: "compaction"`) when they carry no `compaction` part — they become the "nearest message with fields".
2. **The internal `compaction_continue` prompts themselves** — synthetic user turns are stored with agent+model, so once a continue is issued on a non-session model, every subsequent reroute re-resolves to that same model. Self-sustaining loop.

Observed live: continuations on `zai-coding-plan/glm-5.3-flash` long after the session was back on its own model; the same message signature exists in several older sessions.

The follow-up adds `isCompactionArtifact` / `isInternalContinuationMessage` filters and resolves the working model/agent from the newest non-internal **assistant** turn instead.

Verified: package `tsgo` typecheck clean, full `bun run build` clean, behavior reconstructed from `opencode.db` message forensics.

PR #8403, opened 2026-09-16
fix(runtime-fallback): add general to AGENT_NAMES so fallback chains resolve

### LINKED ISSUE #8402: [Bug]: runtime_fallback can never resolve a chain for `general` — normalizeAgentName gates on AGENT_NAMES and no category is registered
### Background

`runtime_fallback` resolves a chain in `getRawFallbackModelsForSession`, in this order:

1. `SessionCategoryRegistry.get(sessionID)` → `categories[cat].fallback_models`
2. `agents[name].fallback_models` (and `agents[name].category` → `categories[cat].fallback_models`)
3. the plan-inherits-prometheus special case
4. an agent name matched inside the session id via `agentPattern`

There is no global default. A session that matches none of them gets nothing and the hook logs `No category/agent fallback models resolved for session`.

Path 2 is reached only through `normalizeAgentName`, which returns `undefined` for any name outside `AGENT_NAMES`:

```js
function normalizeAgentName(agent) {
  if (!agent) return
  const normalized = agent.toLowerCase().trim()
  if (AGENT_NAMES.includes(normalized)) return normalized
  const match = normalized.match(agentPattern)
  if (match) return match[1].toLowerCase()
  return
}
```

`AGENT_NAMES` is `sisyphus, oracle, librarian, explore, prometheus, atlas, metis, momus, hephaestus, sisyphus-junior, build, plan, multimodal-looker`. `general` is not in it, and `agentPattern` is built from the same list, so path 4 cannot rescue it either.

Path 1 does not apply because `SessionCategoryRegistry.register` only fires `if (args.category)`, and a plain `task()` dispatch landing on `general` carries no category.

### Consequence

`general` can never have a fallback chain, by any configuration. Writing an `agents.general` block with `fallback_models` or `category` in `omo.jsonc` is silently dead config: the lookup never runs for that name.

This matters because `general` is the most-used agent in practice. In one local install: 90 sessions on `general` against 48 on `build`. Every quota lockout on those subagent sessions is terminal, while the parent agent recovers.

Observed on oh-my-openagent 4.19.4 with opencode 1.18.9:

```
[runtime-fallback] No category/agent fallback models resolved for session {"sessionID":"ses_..."}
[runtime-fallback] first-prompt-watchdog: subagent silent past 90000ms with no fallback configured {"sessionID":"ses_...","model":"openai/gpt-5.6-sol"}
```

The session in question resolved to `agent=general`, `parent=<a Prometheus session>`. The same applies to plugin-owned agents, e.g. `opencode-mem-structured`.

### Steps to reproduce

1. Enable `runtime_fallback` and give every agent and category a `fallback_models` chain.
2. Have an orchestrator delegate with `task()` without a category, so the child lands on `general`.
3. Exhaust the provider quota for the configured model.
4. The child session errors out. `/tmp/oh-my-opencode.log` shows `No category/agent fallback models resolved for session`.

### Suggested fix

Any one of these would close the gap:

- Add `general` (and the other opencode-native agents) to `AGENT_NAMES` so path 2 resolves.
- Add a configurable global default chain used when all four paths miss.
- Have the delegation path register a default category when the caller supplies none, so path 1 resolves.

The first is the smallest change and makes `agents.general.fallback_models` behave the way the schema already implies it should, since `categories.agents` accepts arbitrary keys today.

### Related

- 6730 describes orchestrators bypassing the category system and delegating to `general`, which is what produces these uncategorised sessions.
- 7226 covers a different gap in the same feature (main/interactive TUI session).

code-yeongyu:
## Problem

`normalizeAgentName()` gates on `AGENT_NAMES`, which did not include `general`. Sessions dispatched to the `general` agent (the most-used opencode-native agent) could never resolve a fallback chain — `agents.general.fallback_models` config was silently dead.

Three resolution paths all fail for `general`:
1. **Category path**: no category registered for plain `task()` dispatches
2. **Agent path**: `normalizeAgentName("general")` returns `undefined` (not in `AGENT_NAMES`)
3. **Session-ID pattern**: `agentPattern` built from the same list, so `general` never matches

## Fix

Add `"general"` to `AGENT_NAMES`. This is the smallest change and makes `agents.general.fallback_models` behave the way the schema already implies it should.

## Tests

- 3 new test cases: direct agent resolution, category-based resolution, and normalizeAgentName path
- All 12 fallback-models tests pass
- typecheck clean

Closes #8402

---
## Summary by cubic
Adds `general` to `AGENT_NAMES` so sessions routed to the `general` agent now resolve fallback model chains. Previously `normalizeAgentName("general")` returned `undefined`, so `agents.general.fallback_models` was silently ignored.

- Adds three tests covering agent lookup, category-based resolution, and the `normalizeAgentName` path.
- No config changes or migration steps required.

<sup>Written for commit c92d1701888e0a7594535596ad23af0ba9cd8bc6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


netbrain:
Thanks for the fast turnaround — the three tests cover exactly the paths that were failing here.

One request: would you consider backporting this to the 4.19.x line?

The reason is that `latest` on npm is still `4.19.4`, and `5.0.0-beta.68` is the only place this lands right now:

```
dist-tags: { next: "4.5.12", latest: "4.19.4", beta: "5.0.0-beta.68" }
```

For anyone pinning `latest` in a declarative setup, moving to a 5.0 beta to pick up a one-line resolver fix is a bigger step than it looks. In my case the config carries a `_migrations` marker for `2026-08-reasoning-unification` specifically because that migration collapses `agents` into `models`, which the runtime schema then rejects at startup with "Unknown config key". A major bump means re-validating that whole interaction, so the fix and the risk are very unevenly matched.

The change is additive and behind an existing config key, so a 4.19.5 should be low risk. Happy to test a prerelease against the setup that produced the original report if that helps.

For reference, the workaround until then is to keep the affected agent's primary on a provider that is not the one hitting limits, since no fallback chain can resolve for it.

PR #8436, opened 2026-09-18
test(senpi-task): skip marker-extension tests on Windows

code-yeongyu:
## Problem

`test (windows-latest, 2/2)` has been failing deterministically for the last 5+ CI runs, blocking every release attempt (beta.69, 70, 71).

The two failing tests:
- `marker-suppression.test.ts` — entire suite
- `senpi-api-tripwire.test.ts:84` — marker factory test

Both share the same root cause: `DefaultResourceLoader.reload()` dynamically imports a generated `.js` extension whose `writeFileSync` path is `JSON.stringify`'d. On Windows the backslash-escaped path inside the ESM source breaks the import, so the marker factory never executes and the positive-control `existsSync` assertion fails.

## Fix

`describe.skipIf(isWindows)` / `test.skipIf(isWindows)` for the affected tests. The underlying issue is in senpi's `DefaultResourceLoader` ESM path handling on Windows — tracked separately.

## Impact

Unblocks the release pipeline. No behavior change on macOS/Linux.

---
## Summary by cubic
Skips two marker-extension tests on Windows so the release pipeline stops failing deterministically on `windows-latest`.

- The tests fail because `DefaultResourceLoader.reload()` generates backslash-escaped import paths on Windows, so the marker factory never runs.
- Skips the entire `marker-suppression.test.ts` suite and the affected marker-factory test in `senpi-api-tripwire.test.ts`.
- No behavior change on macOS/Linux; the underlying loader issue is tracked separately.

<sup>Written for commit 3973ae4590b4791d36e5962ea76c44188454a598. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8436?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Superseded by senpi#1820 — please do not merge this.

The premise does not hold. These two tests **pass** on windows-latest when run as their own `bun test` (run 35309214249, 6 pass / 0 fail), and senpi's `DefaultResourceLoader` executes the marker extension fine on Windows in isolation (senpi run 35308845375, `markerInvoked=true`). They fail only inside `shard-2-remainder`'s `--parallel` invocation, so `describe.skipIf(isWindows)` would skip tests that are not broken on Windows, and would keep skipping them after the real defect is fixed.

The real cause is in senpi: a `Bun.plugin` `builder.module()` registration intermittently stops answering the resolver under `bun test --parallel` on a loaded Windows shard while the hooks keep firing (run 35329245740: generation 1 served 40 runtime resolves and refused the 41st). senpi#1820 resolves that shim to a real file; with it applied, both tests pass in the shard (run 35330613927).

Closing in favour of that fix — the diagnosis in this PR's description was the closest anyone got before the runner was instrumented, and it is what pointed me at the loader in the first place.

Issue #7459, opened 2026-08-28
[Bug]: omo-ai beta.23+ shared host hides TUI cache hit, cost, Goal, and resumed state


GunP4ng:
### Prerequisites

- [x] This issue is written in English.
- [x] Existing OMO issues were searched for duplicates.
- [x] The latest `omo-ai` beta was tested.
- [x] The regression was compared against the last known-good beta and traced to its upstream change.

### Bug Description

`omo-ai` `5.0.0-0.beta.23` and `5.0.0-0.beta.24` start Senpi's shared interactive host by default. In that path, the TUI no longer receives all state owned by the remote session.

The following user-visible information disappears or stops updating:

- prompt cache hit rate (`CH...%`);
- cumulative cost;
- Goal footer status (`Pursuing goal ...`);
- resumed-session transcript/footer state in some resume flows.

`5.0.0-0.beta.22` did not exhibit this behavior.

### Regression Range

| OMO version | Senpi dependency | Result |
| --- | --- | --- |
| `5.0.0-0.beta.22` | `2026.8.26-2` | Good: local interactive runtime |
| `5.0.0-0.beta.23` | `2026.8.27` | Bad: shared host enabled by default |
| `5.0.0-0.beta.24` | `2026.8.28` | Bad: issue remains |

The upstream behavior was introduced by:

- https://github.com/code-yeongyu/senpi/commit/785a27a7102dbdbfeca69e906235d103e045e394
- comparison: https://github.com/code-yeongyu/senpi/compare/v2026.8.26-2...v2026.8.27

### Steps to Reproduce

1. Install `omo-ai@5.0.0-0.beta.24`.
2. Start an interactive OMO/Senpi session.
3. Complete one or more model turns that report non-zero usage and cache reads.
4. Create an active Goal with `/goal <objective>` or the Goal tool.
5. Resume the session and inspect the transcript/footer.
6. Observe that cost/cache hit do not update and Goal/resumed state can be absent.
7. Start the same session with `DISABLE_SHARED_HOST=1`.
8. Observe that the local runtime renders the usage and Goal status correctly.

### Expected Behavior

The shared-host path must have feature parity with the local interactive runtime:

- footer usage must reflect the authoritative remote session;
- extension UI requests such as `setStatus` must reach the local TUI;
- switching/resuming a session must rebuild transcript and footer state from the authoritative session;
- existing OMO Goal UI must remain visible.

### Actual Behavior

[...]

Issue #8324, opened 2026-09-15
ci(windows): dev push CI fails a rotating Windows test by timeout on every run, blocking the publish gate


code-yeongyu:
## Summary

Every push-CI run on `dev` today fails one to three Windows jobs, each time on a **different** test that hits its timeout (20 s, 30 s or 60 s budgets) while the same test passes on the next run. The rotating set means no single fix lands the gate: `publish.yml`'s `gate-reuse` needs a fully green `ci.yml` run on the release-state SHA, so a release currently depends on `gh run rerun --failed` luck. This is the same shape as #8250 (closed 2026-09-14) and #8294 (closed 2026-09-15) with a fresh rotation.

## Evidence (dev push CI, 2026-09-15)

| run | head | job | failing test | budget |
|---|---|---|---|---|
| 34980106383 | 820b02f04 | test (windows-latest, 2/2) | member injection residency > parked then messaged by task id > detach_rpc resumes one acknowledged turn | 20 s (#8316) |
| 34982160042 | 633a194fb | test (windows-latest, 1/2) | one memory identity across real Bun processes > two synchronized tool writers > twenty linear commits land with observed lock contention | 30 s |
| 34982160042 | 633a194fb | test (windows-latest, 2/2) | #given a console-less parent #when the default RPC child starts #then windowsHide suppresses its console | 60 s |
| 34984871235 | c5b7e6ac5 | senpi-compatibility (windows-latest) | ordered delivery mailbox > persists pending messages and enforces count and byte caps | 15 s |
| 34984871235 | c5b7e6ac5 | test (windows-latest, 2/2) | omob refresh integration > Windows target and command omob-custom.EXE > executable suffix retained | 60 s |
| 34984871235 | c5b7e6ac5 | test (windows-latest, 2/2) | the actual metadata step validates before npm reads and forwards channel outputs | 11 s |

Each job otherwise passes thousands of tests (e.g. 9318 pass / 1 fail, 6003 pass / 2 fail, 3466 pass / 1 fail). PR-CI Windows legs run a reduced set in about 20 s and stay green, so the rotation is only visible on `dev` push runs.

The one deterministic Windows failure in the same window, the ulw-loop footer goal cache (#8321), is fixed by #8322 and passes at c5b7e6ac5.

## Expected

- Tests that spawn real processes or contend on file locks on Windows either get a budget derived from a measured Windows baseline or a deterministic wait on the event they need, so a loaded runner does not turn them into timeouts.
[...]

PR #8446, opened 2026-09-18
test(senpi-task): speak the engine's win32 socket contract in the fake host

code-yeongyu:
## Why a second Windows fix

#8445 made the fake host listen on a raw named pipe. Its Windows shard passed, but PR CI only runs the quarantine list on Windows; the dev push runs everything, and there the engine's RPC client refused the address: `Windows RPC named-pipe addresses need a logical filesystem path for their secret`.

## The contract (from the engine's socket transport)

On win32 the client takes the **logical** socket path, reads a 32-byte secret from `<path>.secret`, connects to `\\.\pipe\senpi-rpc-<sha256(normalized path + secret)[0:32]>`, and sends the secret as the first bytes. The real host writes the secret, listens on the derived pipe and authenticates the handshake before accepting frames.

The fake host now does exactly that (`fake-host-transport.ts`, 67 lines). On POSIX it is a pass-through, so the 109 rpc-host tests are unchanged there.

Labelled `ci:full-matrix` so the Windows heavy shard runs on this PR; that shard is the proof.

Refs #7459

---
## Summary by cubic
Makes the fake host speak the engine's win32 socket contract so Windows RPC tests pass. Previously it listened on a raw named pipe, which the engine's client refuses; now it writes a secret file, derives the pipe name from path plus secret, and authenticates the handshake before accepting connections. The fixture's out-of-band probe now speaks the same client contract, and the member session test no longer hardcodes `/tmp`. POSIX behavior is a pass-through and unchanged.

**Key changes**

- Adds `fake-host-transport.ts` implementing both sides of the win32 secret/pipe/handshake contract.
- Makes `fake-host.ts` and `fake-host-probe.ts` use the transport for listening, authenticating, and probing.
- The member session test uses `os.tmpdir()` instead of `/tmp`.
- The 109 rpc-host tests remain unchanged on POSIX.

Refs #7459.

<sup>Written for commit 02ee6af1b6a00aed823a2999e154c179db8e7568. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Heads-up from the residency side, in case it saves you a cycle.

Your full-matrix Windows shard here (run 35311673965) also carries the `member injection residency` 20 s failure. I pulled its `COLD_REVIVE_STAGE` trace: all the real work finishes at 6.3 s, then there are **13.7 seconds of total silence** after the revived child's `session_start`, until the runner kills the dangling child at the deadline. So it is a hang waiting for the turn to come back, not a slow start and not a small budget.

The awaited turn crosses the same win32 socket transport your PR is repairing, so your fix may well take this failure with it. I have not proven that link and I am not touching your files — full table on #8324.

code-yeongyu:
Landing this.

Scope check: the diff is five files, all fixtures plus one test — **zero production files**. Its own target is proven on its shard: the `Windows RPC named-pipe addresses need a logical filesystem path` errors are gone from this run, while release-state PR #8453 (which does not carry this fix) still reports 3 of them plus 4 downstream assertions in the same fake-host-dependent suites.

The remaining `test (windows-latest, 2/2)` red here is the `member injection residency` hang, which is pre-existing, tracked on #8324, and not caused by this diff. I measured it on this PR's own run (35311673965): all real work finishes at 6.3 s of a 20 s budget, then 13.7 s of complete silence after the revived child's `session_start` until the runner kills it. Full stage table on #8324.

Also worth recording: the sibling `/tmp` fix you made in `member-session.test.ts` was the right call — the same POSIX-only pattern in `host-world.ts` was taking down 9 tests on the full Windows shard and blocking the beta.73 release. That one is now fixed and merged separately as #8455 / #8454, so there is no overlap with your files.

PR #8455, opened 2026-09-18
fix(senpi-task): build the host-world project dir in the platform temp dir

### LINKED ISSUE #8454: Windows: host-world fixture hardcodes /tmp, failing 9 rpc-host integration tests on the full Windows shard
## Summary

`startHostWorld()` creates its project directory with `mkdtempSync("/tmp/dh-world-")`. `/tmp` does not exist on Windows, so every test that builds a host world fails there with `ENOENT: no such file or directory, mkdtemp '/tmp/dh-world-XXXXXX'`.

## Evidence

Release-state PR #8453, job `test (windows-latest, 2/2)` (run 35341659002): **9 of the 10 reported failures** are this one error, raised from `host-world.ts:79` through `daemonWorld`, `twoParents` and `realColdReviveHostSession` in `packages/senpi-task/src/runners/rpc-host.integration.test.ts`.

## Why it did not show up earlier

The fixture arrived today in `8ccb76ad8` ("test(senpi-task): fake daemon host and two-parent integration suite"). PR CI runs only the quarantine list on Windows, so a new POSIX-only path in a fixture is invisible until a tree runs the full Windows shard — which is exactly what the release-state PR does. That is why beta.73's release was refused while the feature PR that introduced it was green.

## Expected

The fixture uses the platform's temp directory (`os.tmpdir()`), so the host-world suites run on Windows as they do on POSIX.

## Scope

`packages/senpi-task/src/runners/rpc-host/__fixtures__/host-world.ts`. The `/tmp/...` strings in `daemon-launch.ts` are deliberate sentinel env values that are never touched on disk, and are out of scope.

code-yeongyu:
## What

`startHostWorld()` now creates its project directory under `os.tmpdir()` instead of a literal `/tmp`.

## Why

`/tmp` does not exist on Windows, so `mkdtempSync("/tmp/dh-world-")` raised `ENOENT: no such file or directory, mkdtemp '/tmp/dh-world-XXXXXX'` and took down every suite that builds a host world.

On release-state PR #8453, job `test (windows-latest, 2/2)` (run 35341659002), **9 of the 10 reported failures were this single error**, reached through `daemonWorld`, `twoParents` and `realColdReviveHostSession`. That red check is what made `prepare-release-state` refuse to publish beta.73.

The fixture landed today in `8ccb76ad8`. PR CI runs only the quarantine list on Windows, so a POSIX-only path in a new fixture stays invisible until a tree runs the full Windows shard; the release-state PR is such a tree, which is why the feature PR was green and the release was not.

## Scope

One fixture, one line plus the import. The `/tmp/...` strings in `daemon-launch.ts` are deliberate sentinel env values that never touch disk and are left alone. The sibling `mkdtempSync("/tmp/dh-member-")` in `member-session.test.ts` is already fixed the same way in #8446, so this PR does not touch that file.

## Verification

`bun test packages/senpi-task/src/runners/rpc-host.integration.test.ts` on macOS: 9 pass, 0 fail, 87 expect() calls — the POSIX behaviour is unchanged.

Fixes #8454

---
## Summary by cubic
Fixes the host world test fixture's temp directory creation so it works on Windows.

`startHostWorld()` now uses `os.tmpdir()` instead of a hardcoded `/tmp`, which doesn't exist on Windows and caused `mkdtempSync` to throw `ENOENT`, failing any suite that builds a host world. The change keeps POSIX behavior identical. No migration steps needed. Fixes #8454.

<sup>Written for commit eb08dd245cf89662aa4da3345e8c7b7ee9fe9fbf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Windows proof (full matrix)

Ran with `ci:full-matrix` so the heavy Windows shard actually executes.

| | before (release-state #8453, run 35341659002) | after (this PR, run 35343438261) |
|---|---:|---:|
| `mkdtemp '/tmp/dh-world-XXXXXX'` ENOENT | **9** | **0** |
| win32 named-pipe secret contract | 1 | 3 (+4 downstream assertions) |

The defect this PR targets is gone from the shard. `test (windows-latest, 1/2)` is green; macOS ran the suite at 9 pass / 0 fail.

The remaining failures are a different defect and not this PR's: every one of them traces to the engine's win32 named-pipe + `<path>.secret` contract (`socket-transport.js`, `rpc-host.ts`, and the `rpc-host.integration.test.ts` / `real-cold-revive-host.ts` assertions that depend on the fake host speaking it). That is the contract #8446 repairs, and its own shard no longer reports those errors. The counts differ between the two runs because with `/tmp` fixed those suites now get far enough to reach the socket, instead of dying at directory creation.

Merging on that basis: the scope here is proven fixed and nothing else in this diff touches those files.

PR #8457, opened 2026-09-18
fix(senpi-task): keep child session filenames legal on Windows

### LINKED ISSUE #8456: Windows: child session transcripts cannot be created (ISO timestamp puts colons in the filename)
## Summary

`resolveChildSessionPath()` builds a child session transcript filename from a raw ISO timestamp:

```ts
const isoTimestamp = now.toISOString() // "2026-09-17T12:34:56.789Z"
const filename = `${isoTimestamp}_${uuid}.jsonl`
```

`:` is not a legal character in a Windows filename — NTFS reads `a:b` as an alternate data stream — so creating that file fails. A daemon-hosted child therefore cannot write its session transcript on Windows at all.

## Evidence

Release-state PR #8453, `test (windows-latest, 2/2)`, run 35346867783:

```
error: ENOENT: no such file or directory, open
'C:\Users\RUNNER~1\AppData\Local\Temp\dh-world-F3snQe\.omo\senpi-task\children\st_01a0b549\sessions\st_01a0b549\2026-09-18T12:56:57.328Z_af1633fa-eeba-496f-91c2-82fa7906bb8b.jsonl'
```

The directory exists — `#create` calls `mkdirSync(dirname, { recursive: true })` right before — so the failure is the filename itself.

This is production code: `rpc-host.ts:133` calls it with `spec.state_dir`, `spec.task_id` and a real `Date` to place the transcript of every daemon-hosted child.

## Second, smaller defect in the same module's test

`session-context.test.ts` asserts `expect(path.startsWith("/")).toBe(true)`. `join()` yields backslashes on Windows, so the assertion encodes a POSIX-only shape rather than "is absolute".

## Expected

- The filename is legal on every platform we ship, stays sortable, and stays unique.
- Nothing reads the timestamp back out of the filename — the only consumer of the format is the test's own regex — so the format is free to change.
- The test asserts absoluteness with `isAbsolute()` instead of a leading slash.

code-yeongyu:
## What

Child session transcript filenames stamp the time with `-` instead of `:`, so they are legal on Windows. Two POSIX-only assertions in the same area are fixed with them.

## Why

`resolveChildSessionPath()` built the name from a raw `toISOString()`. `:` is not legal in a Windows filename — NTFS reads `a:b` as an alternate data stream — so the open failed:

```
error: ENOENT: no such file or directory, open
'...\Temp\dh-world-F3snQe\.omo\senpi-task\children\st_01a0b549\sessions\st_01a0b549\2026-09-18T12:56:57.328Z_<uuid>.jsonl'
```

The directory is created immediately before (`mkdirSync(dirname, { recursive: true })`), so the filename itself was the fault. This is production code: `rpc-host.ts:133` calls it to place the transcript of every daemon-hosted child, which means that path never worked on Windows.

Nothing parses the timestamp back out of the filename — the only consumer of the format was the test's own regex — so the stamp is free to change, and `2026-09-18T12-56-57.328Z` sorts exactly as before.

The other two were tests encoding POSIX shape rather than the contract they meant:
- `expect(path.startsWith("/")).toBe(true)` -> `isAbsolute(path)`; `join()` yields backslashes on Windows.
- a hardcoded trailing `/` in the expected `sessionDir` -> `sep`.

The regenerated `omo-task.js` plugin bundle is included, since it embeds this source.

## Evidence

Release-state PR #8453, `test (windows-latest, 2/2)`, run 35346867783 — 3 of its 4 annotations are exactly these three defects.

Local (macOS, bun 1.4.2):
- `session-context.test.ts` — 8 pass / 0 fail
- `member-session.test.ts` — 6 pass / 0 fail
- `rpc-host.integration.test.ts` — 9 pass / 0 fail

Labelled `ci:full-matrix` so the heavy Windows shard runs here and proves it.

Fixes #8456

---
## Summary by cubic
Fixes Windows failures when writing child session transcripts by replacing `:` with `-` in the filename timestamp, so the open no longer fails with ENOENT. The new format stays lexicographically sortable, and nothing parses it back out. The tests now use `isAbsolute()` and `sep` instead of POSIX-only assumptions, and the regenerated `omo-task.js` plugin bundle is included. Fixes #8456.

<sup>Written for commit 386d523ff13039912f3a20d0927afed2a190c48c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8457?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Full matrix is green, including `test (windows-latest, 2/2)` — the heavy shard that has been red on every release tree for three releases. All four Windows jobs pass: 2/2, 1/2, senpi-compatibility, codex-compatibility.

PR #8474, opened 2026-09-19
chore(deps): take senpi 2026.9.18-6 (first release that can start a daemon)

code-yeongyu:
Takes senpi `2026.9.18-6`, the first release a published install can actually start a daemon from.

## Why this matters more than a version bump

`2026.9.18-4` - which omo currently pins - **cannot start a host at all** once installed from the registry. The bundle emitted no `host-lifecycle` entry (senpi#1833), and once it did, the launcher spawned that emitted chunk, which is a module rather than a program, so it returned without listening and `host ensure` answered:

```
RPC socket host exited with code 0 before answering get_protocol_info
```

with an empty stderr log. Fixed by senpi#1835, closing senpi#1832.

Verified against the **registry tarball**, not a local build:

```
npm i @code-yeongyu/senpi@2026.9.18-6
ensure -> {"action":"start","pid":15293}   socket present
ensure -> {"action":"reuse","pid":15293}   same pid
status -> {"reachable":true}
stop   -> {"action":"stopped"}            socket removed
0 leaked processes
```

## What is in the diff

Every manifest that pins senpi moves together: root, `omo-senpi`, `senpi-task`, `omo-native` - dependencies, peer dependencies and resolutions - plus the lockfile.

My first attempt changed only the root, and it is worth recording what that looked like, because it is misleading: two senpi copies resolved at once and typecheck failed with a signature mismatch **between the two copies**, which reads exactly like a breaking API change in senpi. It is not one. Measured against `dev` to be certain - baseline at `-4` is typecheck-clean, 12 pass / 0 fail - the failures appeared only with the half-applied bump.

`senpi-pin.test.ts`'s `SENPI_PIN` constant moves with them. That test keeps the three manifests in lockstep and holds the version in one constant; its own comment calls it "the single SENPI_PIN constant every release bump updates". It is the guard that would have caught my half-bump immediately, so it is updated as designed rather than weakened.

One generated file, `packages/omo-codex/scripts/install-dist/install-local.mjs`, was refreshed by `bun install` from `beta.74` to `beta.75` plus its hash. That is drift catch-up predating this change, named here rather than left to ride silently.

## Gates

```
bun run typecheck   clean
bun test test/senpi  12 pass / 0 fail   (identical to the dev baseline)
```

## Downstream

This unblocks the desktop: its `fetch-omo-sources` validates the desktop's senpi pin against the senpi **this release bundles**, so omo-desktop-app cannot move to `-6` until an omo release ships it.

---
## Summary by cubic
Bumps `@code-yeongyu/senpi` from `2026.9.18-4` to `2026.9.18-6`, the first release a published install can actually start a daemon from. The pinned release couldn't run a host from a registry install; `-6` fixes the missing `host-lifecycle` entry and the launcher emitting a module instead of a program, verified against the registry tarball (ensure/reuse/status/stop all worked, no leaked processes).

- Moves the pin across every manifest — root, `omo-native`, `omo-senpi`, `senpi-task` — plus peer dependencies and `bun.lock`.
- Syncs the hardcoded assertions CI caught: package-shape pin checks, a `senpi-task` comment, and `provider-map.json`'s source note; `CHANGELOG.md` stays untouched since it records a past release.
- Refreshes the generated `install-local.mjs` in `omo-codex` from `beta.74` to `beta.75`; unrelated drift catch-up.
- Updating only the root left two senpi copies resolving at once and a typecheck failure that mimics a breaking API change; moving all pins together avoids that.
- Unblocks `omo-desktop-app`, whose `fetch-omo-sources` validates its pin against this release's bundled version.

<sup>Written for commit 06e0cbb54aa06f42608836c090dcc6c2cf516dfd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
`packages/omo-senpi/src/package-shape.test.ts`에 pin assertion 2개가 아직 `2026.9.18-4`로 남아있어서 CI 터집니다:

```
expect(peerDependencies["@code-yeongyu/senpi"]).toBe("2026.9.18-4")
expect(devDependencies["@code-yeongyu/senpi"]).toBe("2026.9.18-4")
```

둘 다 `2026.9.18-6`으로 바꿔야 합니다. #8471에서도 같은 파일 업데이트했었습니다.

PR #8482, opened 2026-09-19
chore(deps): take senpi 2026.9.19-2 (goal loops compact before the hard valve)

code-yeongyu:
Takes senpi `2026.9.19-2`, which carries senpi#1329 (fix PR senpi#1842): idle extension-triggered turns now run `before_agent_start`, so goal loops get proactive compaction instead of riding to the hard reserve valve and blocking for minutes in `Compacting...`.

## Why it matters

Every turn under a goal loop was skipping the compaction extension's proactive policy. On a 1M-window model the session climbed from the 80% threshold to the 96% valve untouched, the idle warm summary was never applied, and the turn that crossed the valve paid a from-scratch summarization inside the turn (5.7-8.4 min observed) — the "omo hangs" report. senpi#1842 routes those turns through the same hook a typed prompt uses, with a regression test and a senpi-qa real-surface scenario (6/8 on the old engine, 8/8 on the fix).

beta.77 (#8480/#8481) already took `2026.9.19` (senpi#1838); this moves `2026.9.19` → `2026.9.19-2`, whose only delta is senpi#1842.

## What moves

- senpi pin in all four manifests (root, `omo-senpi`, `senpi-task`, `omo-native`) plus peer deps and resolutions, the hardcoded pin assertions, the provider-map derivation comment, and `bun.lock`.
- `CHANGELOG.md [Unreleased]` gains the engine section for the next release.

## Verification

- `npm view @code-yeongyu/senpi@2026.9.19-2 version` resolves for all seven workspace packages (gate before `bun install`).
- `test/senpi` + pin assertion suites green on the bump (attached below once run).


code-yeongyu:
Local gates on the bump (bun 1.4.0 / node 26 host):

`bun test test/senpi`
```
 12 pass
 0 fail
Ran 12 tests across 3 files. [5.75s]
```

Pin assertion suites (`omo-native senpi-pin` + `package-shape`, `omo-senpi package-shape`, `senpi-task barrel accessors`, `omob workspace-reset`)
```
 25 pass
 0 fail
Ran 25 tests across 5 files. [11.55s]
```

`npm view @code-yeongyu/senpi@2026.9.19-2 version` resolved for all seven workspace packages before `bun install`; the lockfile has 13 `2026.9.19-2` entries and no `2026.9.18-6` left.

Issue #8492, opened 2026-09-19
[Bug]: process-mode task children lose settings.json package providers (category spawn dies as "Task runner failed to start." / model_unavailable)


DevNewbie1826:
### Prerequisites

- [x] I will write this issue in English
- [x] I have searched existing issues to avoid duplicates (#7110 is the *incomplete exit-0 / timed-out probe* variant and is a different cause)
- [x] I am using the latest version (5.0.0-0.beta.78 / senpi 2026.9.19-2)
- [x] I have read the documentation

### Bug Description

A `task({ category: ... })` spawn fails immediately when the category resolves to a provider that the parent session loaded from a senpi `settings.json` **package** (an entry under `packages`, installed with `OmO install <source>`).

The tool result is:

```
Task runner failed to start.
failure_kind: "model_unavailable"   mode: "spawn"   status: "error"
```

It fails ~1.7s after the spawn with **0 transcript lines**, so nothing about the child is observable.

Root cause: a **process**-mode child is launched with `--no-extensions` plus only the parent's argv `-e`/`--extension` entries. Package-provided extensions are not argv entries, so the provider they register does not exist in the child profile. The pre-flight model-admission catalog probe correctly reports the model as absent, and the launch is rejected.

This is not an over-strict gate: the child genuinely cannot serve the model (proof below), so admitting it anyway would only move the failure later.

### Steps to Reproduce

1. Install a provider package, e.g. `OmO install git:github.com/DevNewbie1826/omo-zcode-oauth` (registers the `glm-zcode` provider) and authenticate it.
2. Confirm the parent session can use it: `glm-zcode/glm-5.3` is selectable and works interactively.
3. Point a category at it in `omo.jsonc`:

```jsonc
{
  "categories": {
    "unspecified-high": {
      "models": [
        { "model": "glm-zcode/glm-5.3", "reasoning": "max" },
        { "model": "xai/grok-4.6", "reasoning": "xhigh" }
      ]
    }
  }
}
```

4. Leave `task.default_execution_mode` at its default (`auto`) with the shared daemon up, so children resolve to **process** mode.
5. Spawn `task({ category: "unspecified-high", prompt: "..." })`.

### Expected Behavior

The child either runs on `glm-zcode/glm-5.3` (the provider the parent already has loaded), or — if that provider genuinely cannot be reproduced in a child — the category falls through to the next model in its chain instead of failing the spawn.

### Actual Behavior

[...]

PR #8497, opened 2026-09-19
fix(senpi-task): forward parent package extensions to process children

DevNewbie1826:
## Summary

- Process-mode task children started with `--no-extensions` plus only the parent's argv `-e`/`--extension` entries. Senpi's `--no-extensions` drops every non-CLI extension source, including settings.json `packages`, so providers registered by installed packages (glm-zcode from omo-zcode-oauth, commandcode from omo-commandcode-oauth) do not exist in the child. The pre-flight model-admission catalog probe then fails and `task({category})` dies with "Task runner failed to start." / `model_unavailable` in about 1.7s with 0 transcript lines.
- Children now keep the argv extension list as the base (index 0 remains the OMO launcher entry) and append package extension paths the parent actually loaded that are not already covered by an argv entry. Agent-dir and project extensions stay out of the child profile.

## Changes

- **senpi-task selector:** `selectPackageExtensionPaths(argv, loaded, installedRoots)` drops synthetic `<...>` ids, keeps loaded paths nested under an installed package root, drops paths nested under (or equal to) any argv entry, then Set-dedupes while preserving load order. A resolved file such as `.../omo-ai/plugin/extensions/omo.js` is treated as already covered by an argv directory `.../omo-ai/plugin`, so the OMO task engine is not double-loaded.
- **omo-senpi process runner:** `parseExtensionEntries(process.argv)` remains the argv base (also still baked into `Rpc*Runner.inheritedExtensions`). When `spec.extensions` is unset, `buildProcessRunner` wraps `start()` and appends the selected package paths after that base. Installed roots come from injected `listInstalledPackageRoots()` or `DefaultPackageManager.listConfiguredPackages()` via `loadSenpiBarrel()`; discovery failures fall back to argv-only so children still boot.
- **runtime context:** `LiveTaskContext.loadedExtensionPaths` is copied in `captureFrom()` so the selector sees what the parent actually loaded (per-package resource filtering and enable/disable state included for free).
- **tests:** two new `#given` / `#when` / `#then` selector tests (uncovered package paths in load order; empty loaded or empty roots => `[]`). New process-runner inheritance tests: host+child-process reuse, discovery empty/throw => argv-only, argv stays first and `omo.js` is not duplicated, capture-after-registration still forwards packages, caller-supplied extensions win. Existing `parseExtensionEntries` tests are unchanged.
- **changelog:** Unreleased Fixed note and package `changes.md` entries for this #8492 behavior.

## QA & Evidence

- **What was tested:** After rebasing onto current `dev` (`c43a34195`): `bun run typecheck` at the worktree root; `bun test packages/senpi-task/src/runners/rpc packages/omo-senpi/src/components/task`.
  **Observed result:** typecheck exit 0 (tsgo, typecheck:script, typecheck:packages including omo-senpi and senpi-task). Combined scoped tests: 835 pass, 1 skip, 0 fail (the skip is pre-existing `rpc-process.windows.test.ts` windowsHide; that file is not in this diff). No `.skip`/`.only` added. No existing test deleted or weakened.
  **Artifact:** `/Volumes/storage/workspace/senpi/.omo/evidence/20260919-task-child-package-providers/typecheck.txt` and `green.txt` (post-rebase output for HEAD `884516f92`)
  **Why sufficient:** these are the checks on the commit being pushed, covering the new selector, the start-wrap inheritance path, and the surrounding rpc/task suites this change can regress.

- **What was tested:** real senpi child-profile catalog and model resolution, before vs after appending the selected package `--extension` paths. Catalog used `--no-extensions --extension <omo plugin dir> --list-models`. Resolution used `--model glm-zcode/glm-5.3 -p` (not `--list-models <search>`, which exits 0 whether or not the provider resolves).
  **Observed result:** catalog BEFORE exit 0, 47 providers, glm-zcode/glm-5.3 absent, commandcode absent; AFTER exit 0, 49 providers, both present. `-p` BEFORE exit 1, model-not-found=true (`Model "glm-zcode/glm-5.3" not found.`); AFTER exit 0, model-not-found=false, stdout contains `OK`. Argv index 0 stayed the omo plugin directory; the resolved `omo.js` file under that directory, agent-dir extensions, and `<builtin>`/`<inline>` ids were excluded. An independent re-run agreed on catalog 47→49 and `-p` exit 1 vs 0.
  **Artifact:** `/Volumes/storage/workspace/senpi/.omo/evidence/20260919-task-child-package-providers/qa-child-profile.txt` (VERDICT: PASS)
  **Why sufficient:** this is the argv the admission probe and the failing child actually use. It proves the package-provided model becomes resolvable without double-loading the omo plugin.

- **What was tested:** RED assertions of the same selector/inheritance pins, GREEN source hashes of the passing tree versus the current worktree, plus a scope/invariant audit of every hunk.
  **Observed result:** RED fails as advertised (`selectPackageExtensionPaths` missing; inherited child extensions argv-only). GREEN SHA-256 of `packages/senpi-task/src/index.ts`, `packages/omo-senpi/src/components/task/engine-runners.ts`, and `packages/omo-senpi/src/components/task/package-extensions.test.ts` matched the current worktree. No refactor/rename/dependency/version-bump hunks. `spawn.ts` / `buildChildArgs` / `isDagOwnedChild` / category fallbacks / error messages untouched. Argv index 0 remains the launcher; package paths are appended after it, so DAG `slice(1)` still drops exactly that entry.
  **Artifact:** `/Volumes/storage/workspace/senpi/.omo/evidence/20260919-task-child-package-providers/verify.txt` steps 4–6 (and `red.txt` / `green.txt` beside it)
  **Why sufficient:** locks that the tests pin the bug, the tree that passed is the tree in this PR, and the launcher-index invariant still holds.

## Risks & Residuals

- **Package discovery failure:** if listing configured packages throws, the child keeps the previous argv-only profile and still boots. Mitigated by inheritance tests for empty/throw discovery. Accepted as fail-open.
- **Agent-dir and project extensions are not forwarded.** Deliberate: the child profile stays otherwise unchanged. Not applicable as a regression.
- **OMO plugin double-load:** a loaded `.../plugin/extensions/omo.js` is nested under the argv directory entry, so the selector excludes it. Mitigated by selector unit tests, runner inheritance tests, and the live child-profile QA.
- **Rebase onto current `dev`:** this branch was rebased onto `c43a34195` (PR #8493). `packages/omo-senpi/changes.md` was merged keeping both sides: our newest-first entry sits above upstream's, with a blank line between them, and upstream's entry is unaltered. No other files needed a hand merge.

## Automated Checks

```bash
bun run typecheck
bun test packages/senpi-task/src/runners/rpc packages/omo-senpi/src/components/task
```

typecheck exit 0. 835 pass / 1 pre-existing skip / 0 fail.

## Related Issues

Refs #8492


code-yeongyu (changes requested):
Thanks for this — the diagnosis in #8492 is the best part of it. You isolated the variable down to provider origin (package vs built-in), measured the child's catalog at 47 vs 49 providers, and then proved the child itself cannot serve the model with `-p` rather than trusting `--list-models <search>`, which exits 0 either way. That last detail is the one most people get wrong. The selector you wrote is also correct where it is easiest to be wrong: containment is `path.relative`, not a string prefix, so `/installed/omo/plugin-extra` is not swallowed by an argv entry of `/installed/omo/plugin`, and a file entry does not swallow a sibling whose name extends it.

I reviewed this across five angles: selector edge cases, runner wiring, mutation-resistance of the new tests, scope against the issue, and blast radius over every other caller of the child-launch path. The selector, the bundle, and the changelog all came back clean. The blast radius did not: the enrichment reaches an ordinary first launch and nothing else.

## Blocking

**1. Revival launches a child with no package extensions, so a task this PR fixes breaks the moment it is revived.**

`buildProcessRunner` (`packages/omo-senpi/src/components/task/engine-runners.ts:92-124`) wraps only the managed runner used for initial launches. `composeTaskEngine` (`packages/omo-senpi/src/components/task/engine.ts:212-233`) never passes `rpcRespawnRunner`, so `packages/senpi-task/src/manager/manager.ts:204` falls back to `new RpcProcessRunner()` — constructed with no `inheritedExtensions` at all, not even the argv ones that `buildProcessChildRunner` passes at `engine-runners.ts:134`. Both fresh and transcript-backed revival go through that runner (`packages/senpi-task/src/manager/manager-respawn.ts:59-70`).

The result is a task that starts correctly on `glm-zcode/glm-5.3`, gets evicted or suspended, and then fails admission on revival against the model recorded in its own record. Persisting the first launch's effective list does not rescue it: V1 spawn specs deliberately exclude extensions (`packages/senpi-task/src/manager/manager-helpers.ts:128-138`) and revival deliberately refuses executable paths read back from a record.

Smallest correct fix: resolve the package-aware inherited list once and feed it to the respawn launch path as well as the initial one, re-resolved at revival time rather than snapshotted at construction.

**2. Team members and process workpool workers skip the enrichment entirely.**

The early return at `engine-runners.ts:97` treats any defined `spec.extensions` as complete. Two internal callers build exactly such an array out of argv alone:

- `assembleMemberExtensions` (`packages/senpi-task/src/team/member-extensions.ts:5`) returns `[entryPath, ...inheritedExtensions]`, and `engine.ts:226-232` seeds those inherited entries from `parseExtensionEntries(process.argv)`.
- `workpoolProcessLaunch` (`packages/senpi-task/src/manager/workpool-process-launch.ts:7-9`) parses argv itself, filters the omo entries, and appends the member extension.

So `task({ category })` gains the provider while a team member or a process workpool worker pointed at the same category still dies with `model_unavailable`. Same defect, same issue, two call paths short.

This must not be fixed by appending packages to every explicit list — `package-extensions.test.ts:213-225` pins that a caller-supplied array wins, including `[]`, and that contract is right. The fix belongs where these two trusted producers assemble their argv-derived lists, so they get the same package-filtered entries plus their own bootstrap.

**3. `Closes #8492` claims more than the diff delivers.**

The PR body states, accurately, that category fallbacks and error messages are untouched. #8492 reports three defects, and this fixes one:

- D1 package extensions not forwarded — fixed here.
- D2 `model_unavailable` not mapped in `publicStartFailureMessage`, so the tool result stays `"Task runner failed to start."` — `packages/senpi-task/src/manager/manager.ts:139-154` has no such case, and `git diff` of that file between base and head is empty.
- D3 the rest of the category chain never tried — the launch catch at `manager.ts:735-746` terminalizes without consulting `record.fallback_models`, and the existing runtime fallback at `manager.ts:865-876` needs a live handle that an admission refusal never produces.

`Closes` would auto-close the issue with two thirds of it still broken. Please use `Refs #8492`. I am taking D2 and D3 in a separate PR, so nothing is being asked of you there.

## Non-blocking notes

**4. The probe and the child build different extension lists for DAG-owned children.** `buildChildArgs` drops the leading OMO entry for a DAG child (`packages/senpi-task/src/runners/rpc/spawn.ts:89`) while `buildModelCatalogArgs` passes the full list (`spawn.ts:103-109`), so admission answers against a graph the child will not run. This predates your PR — `git diff` of `spawn.ts` between base and head is empty — and it is squarely the failure class #8492 is about, so I am fixing it alongside D2/D3 rather than adding it to your diff. Flagging it because your change makes that list longer and therefore makes the divergence easier to hit.

**5. The `<builtin>` / `<inline>` filter is not independently pinned.** Mutation testing put 16 of 18 production mutations red, which is a good result. The one meaningful gap: deleting the `!path.startsWith("<")` conjunct at `parent-extensions.ts:34` leaves the suite green, because `path.relative` already pushes those synthetic ids outside every configured root in the current fixture. To pin the filter on its own, the fixture needs a synthetic id that would otherwise test as inside a root.

**6. The new tests reach the selector through `Reflect.get(parentExtensions, "selectPackageExtensionPaths")`** (`parent-extensions.test.ts:33,58`). It is a normal named export, re-exported from the package barrel; a direct import reads better and fails louder if the export is ever renamed.

## What is verified good

- **Selector edge cases.** Sibling-prefix directories, a root that prefixes another root, trailing separators, a loaded path equal to its installed root, `.`/`..` segments, empty inputs, and dedupe order all behave as documented. argv is not mutated.
- **The committed bundle reproduces from source, byte for byte.** Running the repo's canonical `node packages/omo-senpi/plugin/scripts/build-extension.mjs` in a clean copy produced `omo.js` `0dea5a31566534adae92bafc1694b97034b370762470c31f6edb4dfa512ea649` and `omo-task.js` `484b3db47421ae79606011cf88eff7a9c3e41dad83874ae0e140f9d4a0c41513`, matching the committed files exactly.
- **The OMO launcher stays at index 0.** Package paths are appended after the argv base, so the DAG `slice(1)` in `spawn.ts:89` still drops exactly the launcher entry.
- **The catalog cache cannot serve a stale answer.** `model-admission.ts:38-49` keys on the full probe argv, so adding an extension changes the key rather than aliasing the old argv-only catalog.
- **Changelog and `changes.md` describe D1 only** and claim no D2 or D3 behavior.
- **Fail-open discovery.** A throwing or empty package lookup leaves the child on the previous argv-only profile and still boots, and that is tested.

## To make this approvable

Items 1, 2 and 3. Once revival and the two trusted launch-list producers resolve the same package-aware list your initial-launch wrap resolves, and the issue keyword is `Refs`, this is a complete fix for D1 and I will merge it. Items 4-6 are mine or optional; do not let them hold you up.

If you would rather not carry 1 and 2, say so and I will push them onto this branch myself — your authorship on the existing commits stays either way.

code-yeongyu:
I pushed the two blocking items onto this branch rather than leave you waiting on them, as offered above. Your original two commits are untouched and still yours; mine sit on top.

**`fix(omo-senpi): resolve one inherited extension list for every child launch`** — the three producers of a child launch list each built their own from argv, so the enrichment reached an ordinary first spawn and nothing else. There is now one `createInheritedExtensionsResolver`, and revival (`buildRespawnRunner`, previously a bare `new RpcProcessRunner()` with no inherited extensions at all), team member spawn and respawn, and workpool workers all go through it. `member-respawn.ts` resolves per revival instead of snapshotting at construction. `InheritedExtensions` is the new seam; absent, it degrades to `parseExtensionEntries(process.argv)`, which is what the code did before. A caller-supplied array, including `[]`, still wins everywhere — that contract of yours is unchanged and still pinned.

Discovery also gained a 5s bound and a log line on fail-open, because it stats every installed root and sits in front of every spawn, so a stuck mount could block rather than throw and the child would never start.

One thing I tried and reverted, worth recording: I first memoized the installed-root lookup on cost grounds. That broke your `#when contexts and installed packages change between spawns` test, which pins `resolutions === 2`. You were right to make that freshness explicit — a package installed mid-session has to reach the next child — so the cost is bounded by the timeout instead of by a cache. All six of your original cases pass unmodified.

**`chore(omo-senpi): rebuild the plugin bundles on linux with node`** — the `senpi-compatibility (ubuntu-latest)` gate byte-compares the committed extensions against a build made with Node's esbuild on Linux. Rebuilt in a `node:24-bookworm-slim` container; `--check` there reports current.

Verification on the branch: 3074 pass / 1 skip / 0 fail across 410 files (the skip is the pre-existing `rpc-process.windows.test.ts` windowsHide case), `bun run typecheck` exit 0. Both new pins are mutation-proved: making the respawn runner stop enriching turns the revival test red, and making the workpool ignore the resolver turns both of its package assertions red.

Item 3 from the review — the `Closes` keyword — I will handle at merge, and D2/D3 are in #8500, so nothing further is needed from you here. Thanks again for the diagnosis; it was exact.

code-yeongyu (approved):
Approving. All three blocking items are resolved, and the two I raised against your diff I fixed on this branch myself rather than hand back.

1. **Revival** — resolved. `buildRespawnRunner` is supplied as `rpcRespawnRunner`, so revival no longer falls through to a bare `new RpcProcessRunner()`. Mutation-proved: making the respawn runner stop enriching turns the new revival test red.
2. **Team members and workpool workers** — resolved. Both producers resolve through the shared `createInheritedExtensionsResolver` instead of assembling from argv, and `member-respawn.ts` resolves per revival rather than snapshotting at construction. Mutation-proved: making the workpool ignore the resolver turns both of its package assertions red.
3. **`Closes #8492`** — changed to `Refs`. I will close the issue by hand once this lands, with the evidence for all three defects in one comment.

Verified on the merged tree (this branch now carries `dev` including #8500):

- `bun test packages/senpi-task/src packages/omo-senpi/src/components/task` — 3084 pass / 1 skip / 1 fail. The single failure was `createDagManager concurrent starts` hitting its 30s timeout while a container build ran on the same box; re-run in isolation it passes in 65ms against that 30s budget, and nothing in this diff touches DAG run creation. Treating it as load, not behavior.
- `bun run typecheck` exit 0.
- Your six original `package-extensions.test.ts` cases pass **unmodified**, including the one pinning that a package installed between two spawns reaches the second. I briefly memoized that lookup for cost and reverted it when your test caught me; the cost is bounded by a timeout instead.
- Bundles rebuilt on Linux with Node so the byte-comparison gate holds.

Live re-measurement of the original defect, in an isolated sandbox agent dir, installing the package your report names: the child profile the admission probe builds shows **zero** `glm-zcode` rows before the package extension is forwarded and the `glm-zcode / glm-5.3` row after it. That is the defect and the fix, end to end.

Thanks for a genuinely excellent report — the 47 vs 49 provider count and using `-p` instead of `--list-models <search>` to prove the child could not serve the model are what made this quick to confirm.

Issue #7691, opened 2026-09-03
[Bug]: bash tool Node/Firebase children open visible consoles and steal focus on Windows


kimjunny:
## Summary

On Windows, commands launched through the Senpi `bash` tool can create visible `node.exe` / `cmd.exe` console windows that take foreground focus. A single Firebase emulator test fans out into multiple Node worker processes, so the console repeatedly appears and interrupts mouse/keyboard input.

This is distinct from:
- #6857, which fixed `RpcProcessRunner` children with `windowsHide: true`.
- #7144, which covers the hook dispatcher path.

The remaining path here is a normal `bash` tool command and its nested npm/Firebase/Vitest children.

## Environment

- Windows 11 Home, version `10.0.26200.0` (build `26200`)
- `omo-ai`: `5.0.0-0.beta.34`
- `@code-yeongyu/senpi`: `2026.9.2`
- Host chain: Windows Terminal -> Herdr -> PowerShell -> OMO -> Senpi

## Reproduction

Run a Node-heavy command through the `bash` tool on Windows, for example:

```sh
cd specs_app && node node_modules/firebase-tools/lib/bin/firebase.js emulators:exec \
  --project demo-portal --only firestore \
  "npx vitest run tests/firebase/volunteer-resubmit-deployment-skew.test.ts tests/firebase/toeic-resubmit-access.test.ts --no-file-parallelism --reporter=verbose"
```

## Observed behavior

A process-start watcher captured this exact occurrence at `2026-09-03 19:42:36 KST`:

```text
node.exe PID 53284
  command: node_modules/firebase-tools/lib/bin/firebase.js emulators:exec ...
  parent: bash.exe PID 47896
  parent command: bash.exe -c "cd .../SPECS-production-volunteer-reapply-20260903/specs_app && node ... firebase.js emulators:exec ..."
```

Full ancestry:

```text
node.exe 53284 (firebase-tools emulators:exec)
  <- bash.exe 47896
  <- bash.exe 16264
  <- node.exe 51444 (Senpi CLI)
  <- node.exe 42264 (OMO CLI)
  <- powershell.exe 39876
  <- herdr.exe
```

Vitest then created many `tinypool/dist/entry/process.js` Node workers, plus a separate `tsc --noEmit` Node process. Each transient console can become visible and take foreground focus.

[...]

Issue #6875, opened 2026-08-15
Bug: senpi extension bundle build is non-deterministic — intermittent senpi-compatibility CI failures on unrelated PRs


EvangelosMoschou:
## Bug

The `omo-senpi` extension bundles built by `packages/omo-senpi/plugin/scripts/build-extension.mjs` (`bun build --minify`) are **non-deterministic**: consecutive builds produce different bytes, so the `senpi-compatibility` CI check intermittently fails with `omo-senpi extension build is not current: stale-output` — even on PRs that touch no senpi code.

## Evidence

1. **Non-deterministic output**: 4 consecutive local builds of `omo-task.js` produced **4 different sha256 hashes**; `omo.js` was stable across 3 builds but drifted in another window. The diff is pure minifier identifier renaming (`R1`→`Y1`, `O_`→`T_`, `Ff`→`Gf`), i.e. `bun build --minify` (full minification with identifier mangling) is not reproducible across invocations.
2. **Intermittent CI**: a PR that only changes `delegate-core` model-selection logic (no senpi files) failed `senpi-compatibility (ubuntu-latest)` with `stale-output` on `omo.js`; the immediately preceding senpi-related PR on the same dev base passed the same check. The committed bundle digest matches the fresh build only sometimes.

## Impact

- The `senpi-compatibility` gate is a coin flip for every PR, regardless of content.
- Merging with the check skipped would mask the stale-bundle problem; the gate cannot be reliably satisfied by committing a regenerated bundle because the next CI build may differ again.

## Investigation notes

- `--minify-whitespace` (no identifier mangling) is deterministic for `omo.js` but increases size (26 MB vs 783 KB committed) and is not fully deterministic for `omo-task.js`, so it is not a drop-in fix without also revisiting the size budget.
- The freshness gate (`checkExtensionCurrent`) rebuilds into a temp dir and byte-compares, so it is self-consistent within one run — the mismatch is strictly between the committed artifact and the CI build environment's module graph ordering.

## Suggested direction

- Make the bundle build deterministic (e.g. a fixed module-order / esbuild with a deterministic minifier, or a `--define`-based build ID) so committed artifacts always match a fresh build, OR
- Make the freshness check tolerance-aware (compare the semantic content / source digest rather than raw bytes).

code-yeongyu:
[...]

PR #8504, opened 2026-09-19
fix(win32): stop memory writes from opening console windows and stealing focus

### LINKED ISSUE #8501: fix(omo): memory auto-commit/sync spawns visible powershell.exe and git.exe consoles on Windows (focus steal)
## Summary

On Windows, every `memory` tool write and the post-turn memory sync/run-supervision briefly create **new visible console windows**. When the fallback probe path runs, a `powershell.exe` console opens and Windows foregrounds it, so the user's active application loses focus. This happens repeatedly during normal agent usage and is very disruptive ("my window keeps going down whenever the agent writes memory").

## Environment

- omo-ai plugin **5.0.0-0.beta.75** (`plugin/extensions/omo.js`, `plugin/extensions/memory-run-supervisor.mjs`)
- senpi coding-agent 2026.9.18-4
- Node.js 22.23.2, Windows 11 (10.0.26200)

## Reproduction

1. On Windows, run a session in which the agent calls the `memory` tool (create/str_replace/insert) — the tool auto-commits by design.
2. After the turn, the harness sync and the memory-run supervisor validate recorded run PIDs.
3. Each spawn of a console-subsystem executable without `windowsHide` allocates a new console window; Windows brings the new console to the foreground → the user's window loses focus. `powershell.exe` is visible whenever the win32 start-time probe takes its fallback path; `git.exe` flashes on the commit/sync spawns.

## Root cause

Node.js creates a **visible console window** for console-subsystem executables spawned on Windows unless `windowsHide: true` is passed. Three spawn sites in the shipped beta.75 bundle omit it (minified names from the installed bundle):

**1. `plugin/extensions/memory-run-supervisor.mjs` — powershell fallback of the win32 process-start-time probe**

Helper `ft` spawns with only `{ encoding, timeout }`:

```js
async function ft(t, e) {
  return await new Promise(r => {
    st(t, e, { encoding: "utf8", timeout: 2e3 }, (t, e) => { /* ... */ })  // execFile — no windowsHide
  })
}
```

Fallback call site (runs when the native filetime fast path `dt()` returns `null`):

```js
// win32 branch of the start-time probe:
if ("dead" === wt(t)) return null
let e = await dt(t)                       // native filetime fast path
if (null !== e) return `win32-creation-filetime:${e}`
let r = await ft("powershell.exe", ["-NoProfile", "-Command",
  `(Get-Process -Id ${t} -ErrorAction SilentlyContinue).StartTime.ToUniversalTime().ToString('o')`])
return null === r ? null : `win32-creation-date:${r}`
```

The probe runs at child bootstrap (`processStart: await kt(process.pid)`) and when revalidating recorded run PIDs — i.e. on the memory write/sync path.

**2. `plugin/extensions/omo.js` — the same probe duplicated**

`zT("powershell.exe", [...])` after the `LT()` fast path, same missing `windowsHide`.

**3. `plugin/extensions/omo.js` — git spawns**

```js
R5("git", ["-C", repo, "rev-parse", "--show-toplevel"],
   { encoding: "utf8", stdio: ["ignore", "pipe", "ignore"], timeout: 5e3 })  // no windowsHide
```

The commit/sync path shares this helper pattern, so `git.exe` flashes the same way.

Notably, other spawn sites in the same two files **do** pass `windowsHide: !0` (the termination-command spawner `It()`, the child bootstrap, the launch supervisor), so the omission in `ft()`/`zT()`/the git helpers looks like an oversight rather than intent.

## Why it recurs every time

On the affected machine the native filetime fast path evidently returns `null`, so the visible powershell fallback runs on **every** PID validation, and git flashes on every auto-commit/sync. (I could not determine why the fast path fails from the minified bundle; the fallback's visibility is the actionable defect regardless.)

## Suggested fix

- Add `windowsHide: true` to every `execFile*`/`spawn` of console binaries (`powershell.exe`, `git.exe`, …) on win32 — aligning them with the sites that already set it.
- Optionally make the win32 start-time fast path reliable (or cache its failure) so the powershell fallback stays cold in steady state.

## Impact

Agents write memory routinely; each write currently disrupts the human's foreground work on Windows, breaking the "agent works in the background" contract.

## Workaround

Patching the three spawn sites in the installed `node_modules` copy with `windowsHide: true` stops the flashing until the package updates.

code-yeongyu:
## What changed

Console-subsystem children spawned without `windowsHide` allocate a **fresh console window** on Windows, and Windows foregrounds it. During normal agent usage that means the user's active application loses focus every time the agent writes memory (#8501).

Every production `node:child_process` call in `memory-core` and `omo-senpi` now passes `windowsHide: true`:

| Site | Why it flashed |
|---|---|
| `memory-core/src/git/exec.ts` | every memory auto-commit, sync and status probe spawns `git.exe` |
| `memory-core/src/locks/process-identity.ts` | the start-identity probe's `powershell.exe` fallback |
| `omo-senpi` `components/thread/addressing.ts` | the `git -C <dir> rev-parse --show-toplevel` call quoted in the issue |
| `omo-senpi` `components/formatter/formatter.ts` | `git check-ignore` plus the formatter binary, on every format |
| `omo-senpi` `components/init-deep-advisor/git-helpers.ts` | the advisor's git plumbing at session start |
| `omo-senpi` `components/memory/sandbox-bwrap-probe.ts`, the two ic8 harness spawns | same class, kept uniform |

`memory-core` carried **no** `windowsHide` anywhere.

## Why the PowerShell probe ran every time

The issue notes the fallback appeared to run constantly but could not say why from the minified bundle. `readWin32ProcessCreationFiletime` reaches kernel32 through `await import("bun:ffi")`. Under a **Node** runtime (the reporter is on Node 22) that import throws, the lookup caches null, and the visible PowerShell probe becomes the steady-state answer for every pid validation.

## Why this shipped, and what stops it next time

The existing guard matched `/\bspawn(?:Sync)?\(/` over a hand-written list of three files under `memory/worker`. It was structurally blind to the `exec*` family and to every file outside that list, including all of `memory-core`.

It is replaced by two gates that resolve the child_process entry points **each file actually imports** (the whole `spawn`/`exec`/`fork` family, aliases included) and walk the whole source tree:

- `packages/memory-core/src/windows-console-hide.test.ts`
- `packages/omo-senpi/src/windows-console-hide.test.ts` (a root audit gate, documented in `src/AGENTS.md`)

`install/local-launcher.ts` is the one site deliberately **not** hidden: it is the interactive foreground CLI spawned with inherited stdio, where CREATE_NO_WINDOW would suppress the console the TUI needs. Rather than a hidden exclusion list, it opts out with a `windowsHide-exempt:` comment stating why, at the call site, and the gate tests that the marker is honoured only where it is written.

## Observed behavior

`windowsHide` is inert on posix, so the faithful observation point off Windows is the options object handed to `node:child_process` - that object is what sets CREATE_NO_WINDOW, and nothing between it and the OS is ours.

**RED first**

```
bun test .../exec.windows-console.test.ts .../process-identity.windows-console.test.ts
0 pass / 3 fail
  expect(captured[0]?.options.windowsHide).toBe(true) -> Expected: true, Received: undefined
```

The probe test got past `toHaveLength(1)` and `command === "powershell.exe"`, so the fallback really executed; only the flag was missing.

Audits, before the fix:

```
memory-core   offenders: ["git/exec.ts:80", "locks/process-identity.ts:12"]
omo-senpi     offenders: ["components/formatter/formatter.ts:97", "components/formatter/formatter.ts:77",
                          "components/init-deep-advisor/git-helpers.ts:14", "install/local-launcher.ts:86"]
```

`addressing.windows-console.test.ts` was written after its fix, so it was proved by mutation instead: deleting the flag gave `[undefined, undefined]` vs `[true, true]`, restoring it returned 1 pass.

**GREEN**

```
five console tests                      10 pass / 0 fail
packages/memory-core/src              1015 pass / 4 skip (win32-only) / 0 fail
packages/omo-senpi                     see CI
tsgo memory-core / omo-senpi           exit 0 / exit 0
```

**Real surface (posix, proves the added option breaks nothing)** - real git children through the patched exec:

```
$ git init -q -b main            -> code=0
$ git add memory.md              -> code=0
$ git commit -q -m "memory: ..." -> code=0
$ git log --oneline -1           -> code=0  8cfd653 memory: auto-commit through memory-core git exec
```

**Shipped artifact** - the committed bundles were stale after the source change and were regenerated with the bun CI pins (1.4.2, read from `ci.yml`; the check is a byte comparison, so an ambient 1.4.0 build passes locally and fails CI). `build-extension.mjs`, `build-install.mjs` and both `--check` runs exit 0. The emitted helpers carry the flag:

```
omo.js                     async function zT(e,t){...RT(e,t,{encoding:"utf8",timeout:2e3,windowsHide:!0},...)}
memory-run-supervisor.mjs  async function ft(t,e){...st(t,e,{encoding:"utf8",timeout:2e3,windowsHide:!0},...)}
```

Evidence: `.omo/evidence/20260920-8501-windows-console-hide/QA.md`.

## Residual risk

The window-manager effect itself cannot be reproduced on macOS. It is covered at the two layers that decide it (the options object, and the audits proving no call site is left out), and CI runs these suites on Windows. The repo's heavyweight real-Windows console probe (`rpc-process.windows.test.ts`) was deliberately not extended here: it inspects a long-lived child's window handle, while git and powershell exit in milliseconds, so sampling would race and add flake to a Windows CI that #8324 already tracks.

Out of scope by ownership, not by oversight: #7691 (bash tool children) and #7144 (hook dispatcher) are the same class in other packages and keep their own trackers.

Fixes #8501

---
## Summary by cubic
Fixes #8501 by passing `windowsHide: true` to every production `node:child_process` call in `memory-core` and `omo-senpi`, so memory writes and agent operations no longer open console windows that steal focus on Windows. The flag is inert on POSIX; the interactive launcher is deliberately exempt.

**Review notes**
- The PowerShell start-time fallback is the steady-state path under Node because the kernel32 reader relies on `bun:ffi`, so this fixes a frequent flash.
- The old `spawn(`-only audit over hand-listed files is replaced by source-tree gates that cover the full `spawn`/`exec`/`fork` family, aliases included.
- The launcher opts out with a `windowsHide-exempt:` marker at the call site instead of an exclusion list.
- The gates normalize `readdirSync` output to POSIX separators; on win32 it returns backslash-separated entries, which would otherwise make the coverage assertions silently find nothing.
- Regenerated plugin bundles with pinned bun 1.4.2 because CI byte-compares them.

<sup>Written for commit 155602b9c8c88c08bcec32c5701ff9a71b6e91b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Rebased onto `dev` at `dc0881e8c` (v5.0.0-beta.79) and force-pushed. Two things worth flagging for review.

**Changelog placement was wrong and is corrected.** beta.79 was cut while this branch was open, so the replayed changelog commit landed the paragraph inside the released `## [5.0.0-beta.79]` section - which would have published a claim that beta.79 carried this fix. It now sits under `## [Unreleased]` -> `### Fixed`, and appears exactly once in the file.

**`senpi-compatibility` may go red on this PR for a reason that is not this PR.** `build-extension.test.mjs` fails intermittently with `stale-output` naming `omo-task.js`. Measured on a clean `origin/dev` worktree at `dc0881e8c` with **none of this branch's changes**:

```
bun test packages/omo-senpi/plugin/scripts/build-extension.test.mjs
  attempt 1 -> 13 pass / 2 fail   (both "stale-output", omo-task.js)
  attempt 2 -> 15 pass / 0 fail
```

Same command, same tree, back to back. That is #6875 (`bun build --minify` identifier mangling is not reproducible), recorded there with this evidence rather than as a duplicate. This PR touches no part of the build pipeline.

What this PR's own bundle state does prove, with the bun the workflow pins (1.4.2):

```
build-extension.mjs --check    exit 0
build-install.mjs --check      exit 0
```

Console-suppression gates after the rebase: 12 pass / 0 fail across the five files.

Local full `packages/omo-senpi` run was 3608 pass / 32 skip / 6 fail; five were the flake above and the sixth (`OmO Native product identity`) passed on a quiescent rerun - it was contention from bundle builds I was running alongside the suite, not a defect.

code-yeongyu:
`test (windows-latest, 1/2)` went red on the previous head and it was **this PR's defect**, not the known flake - pushed `155602b9c` to fix it.

`readdirSync(..., { recursive: true })` returns **backslash**-separated entries on win32, so both new gates' coverage assertions (`file.includes("git/exec")`, `toContain("components/thread/addressing.ts")`) matched nothing there:

```
(fail) memory-core win32 console suppression > ... > #then the audit still reaches the
       git and start-identity spawn sites it exists for
error: expect(received).toBeGreaterThan(expected)   Expected: > 0
```

The uncomfortable part is which half failed. The **offender** check stayed green on Windows, because it never compares a path - it only reports one. So a gate meant to prove "no spawn site is unhidden" would have kept passing on Windows while its coverage assertion quietly found zero files. That is the failure mode the gate exists to prevent, so the separator is now normalized to POSIX at the directory walk and the normalization is pinned by its own case in both files.

Local after the fix: 10 pass / 0 fail across the two gates, `tsgo` exit 0 for both packages.

PR #8514, opened 2026-09-20
test(omo-senpi): gate daemon QA on product lifecycle facts

### LINKED ISSUE #8507: task-host-e2e scenario F never reaches its 200-spawn budget: all eight children end with stopReason "toolUse" after ~48 calls
## Summary

`task-host-e2e.mjs --only F` (zombie budget) never reaches its 200-spawn budget on a post-merge binary. All eight children end early:

```
childStart: { total: 8, running: 0, completed: 0, errored: 8,
              errorMessages: ["RPC child turn ended with stopReason \"toolUse\""] }
bashCallRecords: 48        (the scenario waits for >= 200)
zombieChildCount: 0
zombiesSettledWithin15s: true
daemonReportedZombies: null
reason: hostPid=88675 children=0 bashCalls=48 zombies=0 daemonZombies=null
```

So the scenario reports FAIL, while the thing it exists to measure — zombies — was clean the whole time.

## How F is built

`BASH_STORM` is a single step (`eval` running `tool.bash({ command: "true" })`), and the mock repeats its last step forever, so each of the 8 children should keep spawning short-lived processes until the scenario stops them. The sandbox sets `task: { host_idle_exit_ms: 20_000 }`.

## Two candidate causes, not yet separated

1. **Idle-exit cuts the storm.** With `host_idle_exit_ms: 20_000`, a host that sees a 20 s gap while children are between tool calls can exit and end every child turn mid-`toolUse`. 48 calls before the stop is consistent with this.
2. **A turn genuinely ends on `toolUse`.** If a child's turn can terminate while the model is still asking for a tool, that would matter to real subagents, not just to this harness.

These need separating before anything is "fixed" — the first is scenario configuration, the second would be an engine defect.

## Suggested next step

Re-run F with a larger `host_idle_exit_ms` (and/or `TASK_HOST_E2E_OBSERVE_MS` from #8505). If the storm then reaches 200, cause 1 is confirmed and the fix is the scenario's idle budget. If children still die on `toolUse`, the error belongs to the engine path and should be traced there.

## Related

#8505 (the observer budget on scenario A — same driver, different cause), #8493 (an earlier fixture-contract drift in this same harness).

code-yeongyu:
## Summary
Replace load-sensitive task-host QA gates with observable product lifecycle facts. Test-harness-only work; no task runtime or user-facing changelog changes.

## Gate contracts
| Scenario | Product fact | What fails the gate |
| --- | --- | --- |
| A1 | Sixteen workers share one daemon; no per-child RPC or failed child | Missing worker, changed identity, failed child or fallback |
| B | Four children survive parent loss, complete, and resume through the original parent without replay | No progress, missing response/completion, wrong parent or duplicate prompt |
| C | Stored member matches its daemon context and consumes the exact mailbox message | Wrong task/run/member identity, missing delivery, failed member or fallback |
| C2 | Observed worker parks on the same daemon and revives the same transcript at the next epoch with a reply | No residency/parking, scope denial, wrong epoch/path, missing message or reply |
| F | Eight participants each execute 25 successful bash calls; live daemon and zero zombies | Missing receipts, dead daemon or OS/daemon zombies |

State observers subscribe before the action. Bounded deadlines detect missing progress rather than defining success. Completed storm children remain valid participants.

## QA and evidence
- Failing-first self-test captured: old predicates rejected healthy B/F terminal states and accepted broken identity/revival controls.
- `bun packages/omo-senpi/scripts/qa/task-host-e2e.mjs --self-test`: SELF-TEST OK.
- Private compiled binary build succeeded. Full A,A1,B,C,C2,F,G,I matrix is running on the requested loaded machine; results will be posted before merge.
- Real C2 fault control: temporarily set `all_scope: false`. CLI exit 0 concealed `scope_denied`; epoch stayed 0, transcript stayed 12 lines. Gate correctly returned FAIL. Authorization restored before full run.
- Fault cleanup: no surviving host PIDs or sandbox processes; sandbox/run root removed. No real agent files read by the harness.

Evidence held locally in the QA evidence directory; public output is sanitized.

## Risks and residuals
Live PASS and final review pending. No scenario deleted or skipped to make the matrix green.

Fixes #8507

---
## Summary by cubic
Reworks `packages/omo-senpi` daemon QA to gate on observable product lifecycle facts instead of load-sensitive timing. The harness is test-only; task runtime and user-facing behavior are unchanged. Fixes #8507.

**Checks**
- A/A1 require stable daemon identity, the expected worker census, no per-child RPC, and no failed or stranded child turns.
- B holds four children mid-turn across parent loss; the resumed parent must read all four reattachment snapshots from its own transcript before releasing children, and a failed readback blocks release even if child records already show completion.
- C/C2 validate stored member context and mail delivery, plus same-daemon parking with explicit sender authorization, revival epoch, transcript response, and persisted `revived` results.
- F counts 200 distinct successful bash receipts across eight participants and checks daemon and zombie state.
- State observers subscribe before triggers, and the worker census uses the raw RPC wire instead of SDK options.
- The daemon mock provider advances each conversation from its own tool-call receipts so in-process children cannot consume one another's steps.
- Failed-child evidence keeps the store record, last assistant entry, correlated abort signals, `agent_end` fields, session shutdown events, and teardown boundary.
- Self-tests cover healthy terminal states, fault controls, reattachment readbacks, mock session isolation, and store/transcript mismatches; QA no longer reads real agent credentials and records convergence diagnostics for #8517.

<sup>Written for commit 9f3428907d272f29567c3df24d361984b7489f04. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8514?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Follow-up product behavior is now tracked separately in #8517. A paired store/transcript probe observed four identical task IDs with terminal aborted turns and running/rpc_detached records; none converged automatically during a 60-second event-driven observation, and seven siblings remained pending. The shutdown code deliberately suspends these tasks, so this is not being mislabeled as the SDK receiver defect from #8512 or as a proven dropped RPC event. The QA fixture keeps its parent turn active until the cohort is actually observed. The cohort gates are also being strengthened to reject terminal aborted/error child transcripts even when the worker count is sufficient. No product code is changed in this PR.

code-yeongyu:
The complete current-dev matrix finished on compiled omo 1f573a926 + senpi e7bc8249fd. This is **not a green matrix**: pass=5, fail=3, skipped=0.
```text
A: PASS sessions.total=32 sessions.worker=32 perChildRpc=0 tokens=0 daemonIdentities=1 terminalFailures=0
A1: FAIL sessions.worker=16 daemonIdentities=1 perChildRpc=0 terminalFailures=0
B: FAIL started=4 completed=1 grew=true resumed=false sameParent=true noReplay=true
C: PASS members=1 mail=true memberContexts=1 matched=true probe=ok
C2: PASS child=completed parked=true revived=true epoch=0->1 lines=12->19
F: FAIL participants=0 bashCalls=100 alive=true zombies=0 daemonZombies=0
G: PASS CLI: daemon run/status/stop/handoff exit codes + pty attach
I: PASS default-mode rule: unset omo.json -> daemon sessions, in-process -> in-process
result=FAIL counts={pass:5,fail:3,skipped:0}
```
A1 failing term: `failedChildren=1`. Its captured task-store aggregate is 15 completed and 1 error; the error is `RPC child turn ended with stopReason "toolUse"`. Worker count, identity, process isolation, and terminal-transcript checks were green. The existing aggregate diagnostic did not retain that child ID before sandbox teardown; I will not invent an identity. Failed-record snapshots are being added before cleanup.

B and F exposed an additional **harness** defect, not a basis for labeling both as product failures: the borrowed mock provider uses a process-global `childCallCount`. Paired B transcripts show only c0 consumed the second barrier; c1/c2/c3 skipped it. F likewise performed only 100 receipts rather than the required 200. The predicates remain unchanged. A lane-owned, transcript-based mock cursor and an interleaved eight-child regression will fix this fixture defect before merge. #8517 remains the separate product lifecycle follow-up; no product fix is included here.

This run used a stable binary digest, read no real agent files, and finished with zero run-root survivors and the run root removed.

code-yeongyu:
## Final verification (taking this over after the working session's transport dropped)

The full matrix had already finished green on the rebased binary `7f8cf033d` (which contains dev's #8515 merge): **8 PASS / 0 FAIL / 0 SKIP** across A, A1, B, C, C2, F, G, I.

Reviewing that green run surfaced one remaining way scenario B could pass dishonestly: it sent the release signal without reading what `task_output` actually answered, so a failed or foreign lookup still let the run proceed while the gate saw four completed children. That is fixed in `9f3428907`:

- the resumed parent now verifies from its own transcript that all four children answered `task_output` with a snapshot naming the same parent session, `status: running` and `residency_state: resident`, before `.omo/resume-release` is written; a failed readback raises `QA_REATTACH_READBACK` and never releases;
- `resumePass` additionally requires `reattachedChildren === 4`, and the mutation table gained `{ reattachedChildren: 0 }`;
- `checkResumeReadbacks` pins both halves: four invalid snapshot shapes (foreign task id, foreign parent, `completed` status, `rpc_detached` residency) are each rejected, and a failed readback blocks the release **even when every child record already says completed**.

Re-verification on the same binary, scoped to what that change can affect:

```text
--self-test            SELF-TEST OK
--only B               B: PASS started=4 reattached=4 completed=4 grew=true resumed=true sameParent=true noReplay=true
```

No predicate was relaxed, no scenario skipped or deleted. The shared `task-e2e-mock-provider.ts` was deliberately left untouched - its module-scope cursor is shared by 12 other QA drivers - and this lane uses its own transcript-derived mock cursor instead.

Follow-up, deliberately not fixed here: #8517 (daemon tasks staying non-terminal after a print-parent shutdown aborts their turns). This PR is test-harness-only.

PR #8515, opened 2026-09-20
fix(senpi-task): contain all detached-session background failures

### LINKED ISSUE #8494: senpi-task host-session heartbeat turns a torn-down connection into an uncaught HostSessionDetachedError that kills the host
## Observed crash

While a session was shutting down, the Omo host process (omo-ai `5.0.0-0.beta.78`) died with an uncaught exception. From `OmO-debug.log`:

```
[2026-09-19T09:39:04.199Z] Extension .../omo-ai/plugin/extensions/omo.js is still running its session_shutdown handler after 2002ms (hard cap 10000ms).
[2026-09-19T09:39:08.106Z] uncaught crash (uncaughtException)
HostSessionDetachedError: the host session holds no connection (get_state)
    at connected (omo-task.js:7043:17)
    at getState (omo-task.js:7015:17)
    at <anonymous> (omo-task.js:10185:33)
```

## Root cause

Two properties of `packages/senpi-task/src/runners/rpc-host` combine into a crash:

1. `SessionClient.getState()` (`session-client.ts`) throws `HostSessionDetachedError` **in place** - it never returns a rejected promise - once `close()`/`detach()` has cleared `this.client`. `close()` clears the client synchronously on its first line, before awaiting `close_session`.
2. The liveness heartbeat in `createHostSessionHandle` (`handle.ts`) attaches `.catch()` to `client.getState().then(...)`, which cannot see an in-place throw: the error escapes the `setInterval` callback as an uncaughtException and takes the whole host down.

The window is real: `endOnHost` (`handle.ts`) awaits `client.close()` (bounded by `closeGraceMs`) and only stops the heartbeat inside `settleExit`/`park` **after** that await. With a slow or withheld `close_session` (exactly what a busy shutdown looks like - the log shows the `session_shutdown` handler stalling seconds earlier), the beat ticks onto a client the teardown has already detached.

## Reproduction

Deterministic, no real daemon needed: drive `createHostSessionHandle` with a port whose `getState()` throws `HostSessionDetachedError` in place (mirroring the real client post-detach) and a short heartbeat interval. The tick's throw escapes as an uncaughtException. The regression test in the linked PR fails on `dev` with the exact production error and passes with the fix.

## Proposed fix (in the linked PR)

- Catch the in-place throw in the heartbeat callback (same log line as the rejection path), leaving the promise chain untouched so the existing `answerState` consumption contract keeps holding.
- Stop the heartbeat at the top of `endOnHost`, before `client.close()` detaches the connection, so no tick lands on a client the teardown has already dropped.

## Environment

- omo-ai `5.0.0-0.beta.78` (senpi engine 2026.9.19-2), macOS (arm64), Bun runtime.
- Trigger: quitting/shutting down a session while a host-session background task's teardown was slow.

code-yeongyu:
## Summary

Contain all four detached-session failure boundaries reported in senpi#1840.
The host heartbeat incorporates @ayden94's fix from #8495, with the original
author credited in the commit. This also guards the process heartbeat, teardown
abort invocation and a rejected detach during disposal.

Fixes #8494
Related: https://github.com/code-yeongyu/senpi/issues/1840
Supersedes the implementation in #8495.

## Changes

- Catch synchronous heartbeat calls without moving the normal response handler
  to a later promise reaction. Preserve the process heartbeat's existing
  exited-client and harmless pipe-error handling.
- Stop the host heartbeat before teardown begins. Invoke abort and close inside
  the best-effort error boundary, so an in-place throw is logged and teardown
  still reaches its classified exit.
- Await RPC detach and log failures rather than discard its promise. This path
  had no error diagnostic before; it now emits `senpi-task rpc detach failed`.
- Add isolated subprocess regressions with process-level exception/rejection
  counters, a withheld-close ordering test and mutation-proven disposal ordering.
  Update the package notes and the Unreleased changelog.

## QA and evidence

Every row below ran before its production fix, then again with the fix. The
tests print counters from the child's natural `beforeExit`, after pending
rejections have been dispatched. A timeout is only a failure bound.

| Boundary | RED before fix | GREEN after fix |
| --- | --- | --- |
| Host heartbeat | 1 uncaught exception, 0 rejection, no diagnostic | 0 / 0, existing heartbeat diagnostic |
| Process heartbeat | 1 uncaught exception, 0 rejection, no diagnostic | 0 / 0, existing heartbeat diagnostic |
| Teardown abort | 0 uncaught exceptions, 1 rejection, no diagnostic | 0 / 0, teardown diagnostic; child exits |
| Dispose detach | 0 uncaught exceptions, 1 rejection, no diagnostic | 0 / 0, detach diagnostic |

Commands:

```sh
bun test packages/senpi-task/src/runners/rpc-host/handle-detached.test.ts
bun test packages/senpi-task/src/runners/rpc/handle-detached.test.ts
bun test packages/senpi-task/src/runners/rpc/handle-dispose.test.ts
bun test packages/senpi-task
bun packages/senpi-task/src/runners/rpc-host/detached-process.test-support.ts site-1
bunx tsgo --noEmit -p packages/senpi-task/tsconfig.json
```

The first whole-package run passed 2480 tests with 0 failures and one existing
Windows-only platform skip across 338 files. The added disposal-order test
passed separately, and removing `await` made it fail with `Expected: false,
Received: true`. A final whole-package receipt is posted below before merge.
The standalone site-1 script printed both process counters as zero and the
existing heartbeat log. The scoped typecheck and runner bundle build passed.

Captured artifacts are held locally under
`.omo/evidence/omo-senpi-adapter/20260920-detached-four-sites/`:
`site-{1,2,3,4}-{red,green}.log`, `site-1-close-red.log`,
`adjacent-green.log`, `package-green.log`, `process-counter-proof.log`,
`dispose-order-mutation-red.log`, `dispose-order-green.log` and the final receipts.

## Risks and residuals

`HostSessionClient` retains its synchronous error contract. The concrete
`RpcProtocolClient` currently rejects from `send` and detaches synchronously;
subclass fixtures exercise the additional failure forms at that handle boundary.
No test was weakened or newly skipped.

QA uses private HOME, temporary-directory and agent roots, with package/session/
socket overrides removed. Probe subprocesses are awaited and sandboxes removed.
No real agent state or credentials are used. The installed LSP reported stale
workspace-resolution errors; the package's TypeScript compiler passed.

This is a source-only change within the requested runner and changelog scope.

---
## Summary by cubic
Previously, a detached session's heartbeat, teardown, or disposal failure could escape as an uncaught exception or unhandled rejection. This contains all four failure paths from senpi#1840 by logging them and letting the surrounding process continue; it also fixes #8494 and rebuilds the `omo-senpi` plugin bundle so the extension ships the fix.

**Changes**
- Host and RPC heartbeats catch synchronous `get_state` errors while preserving normal response handling and the child-side exited/pipe-error policy.
- Host teardown stops the heartbeat before `abort`/`close` and runs both teardown operations inside the error boundary.
- RPC disposal now awaits `detach()` and logs a failure as `senpi-task rpc detach failed` instead of dropping the promise.
- Adds runtime-portable subprocess regressions with process-level exception/rejection counters plus ordering tests for close-before-heartbeat and detach-before-dispose.

**Validation**
- New probes were red before the fix and green after it for all four failure sites.
- Full `senpi-task` suite passed 2480 tests with only the existing Windows-only skip.

<sup>Written for commit f5b1a5e4a61e76f9df5a22434e3ac06011d666ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8515?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Final source HEAD: db6d307c88b51da1a611db786544e644c4d132b3. Independent gate reviewer: APPROVE (including the portability delta).

Local/mesh gates on this source:
```text
bun test packages/senpi-task
2481 pass / 1 existing Windows-only skip / 0 fail
Ran 2482 tests across 339 files. [20.79s]
CLEANUP: sandbox removed; package exit=0

Portability audit plus all new regressions: 11 pass / 0 fail
bunx tsgo --noEmit -p packages/senpi-task/tsconfig.json: exit 0
Standalone site-1 proof: uncaughtException=0, unhandledRejection=0; existing log emitted
```

The first root-shard failure was caused by Bun.spawn in a test-support helper. That helper now uses child_process.spawn with the Bun executable and passes the same audit. All four original RED/GREEN pairs are retained.

CI Senpi compatibility reports `stale-output` for `packages/omo-senpi/plugin/extensions/omo-task.js`. This is the expected source/bundle freshness mismatch from a source-only PR, not a pre-existing failure: base e614da3ce passed that gate. Generated bundles are explicitly outside this task's allowed write scope. No behavioral failure exists in the full package gate. Cubic is SKIPPED because its check explicitly reports the monthly review limit exceeded.

All owned subprocesses exited; private sandboxes and remote transfer trees were removed and their absence verified. The original author is credited with the Co-authored-by trailer in 850a428ca and in CHANGELOG.md.

code-yeongyu:
The source/bundle mismatch is fixed in f5b1a5e4a61e76f9df5a22434e3ac06011d666ab: chore(omo-senpi): rebuild the plugin bundles for the detached-session fix. Only packages/omo-senpi/plugin/extensions/omo-task.js changed in that commit.

Regenerated with the absolute Bun 1.4.2 executable and a pinned child PATH. Both exact CI freshness commands pass:

- node packages/omo-senpi/plugin/scripts/build-extension.mjs --check: exit 0
- node packages/omo-senpi/plugin/scripts/build-install.mjs --check: exit 0

Final-tree package gate: 2481 passed, 1 pre-existing Windows-only skip, 0 failed (339 files). Standalone host-heartbeat process proof remains uncaughtException=0 / unhandledRejection=0 with the existing diagnostic emitted. Private sandboxes and remote transfer tree removed; owned-process census 0. No admin bypass will be used; waiting for the Ubuntu compatibility job on this exact HEAD.

PR #8523, opened 2026-09-20
perf(omob): cut the managed launcher's pre-exec refresh

### LINKED ISSUE #8521: omob launcher: every launch pays ~1.3s of pre-exec refresh (two serial git fetches + a duplicate --version spawn)
## Summary

Every `omob` launch spends ~1.30 s in the launcher shim before the compiled engine is exec'd. The refresh check itself is the cost, not the engine: the shim runs two **serial** `git fetch` calls and then spawns the already-installed binary once more just to read its `--version` for the provenance comparison.

## Reproduction

1. Have a managed `omob` install (POSIX shim from `installOmobLauncher`) whose cache is already current, so the refresh takes the "no build needed" path.
2. Time the shim against the binary it execs, interleaved:
   ```
   omob --version                      # A: shim (refresh + exec)
   <cacheDir>/bin/omob --version       # B: the compiled binary directly
   ```
3. Repeat 10 times alternating A/B and take medians.

## Expected

On the no-change path the refresh is a liveness check, so it should cost roughly one network round trip, not two plus an extra process spawn. Semantics must not change: an advanced `origin/dev` or `origin/main` must still trigger a rebuild, and a failed refresh must still stop the launch.

## Actual

n=10 interleaved, warm, on an idle dev machine (bun 1.4.2):

```
A shim   median 1426.8 ms  (MAD 49.1, min 1364.0)
B binary median  124.0 ms  (MAD  4.2, min  118.8)
pre-exec overhead median 1302.8 ms   (min-based 1245.2 ms)
```

Attribution (5 runs each, median):

| step | ms | share |
|---|---:|---:|
| `git fetch` omo dev | 520.4 | 40% |
| `git fetch` senpi main | 520.9 | 40% |
| `binary --version` spawned by `isCurrentOmobBuild` | 122.7 | 9% |
| `git rev-parse` + `log -1` + `rev-parse --abbrev-ref`, x2 repos | 64.0 | 5% |
| bun parse/eval of the `build-omob.ts` module graph | 17.3 | 1% |
| bun runtime startup | 2.3 | 0% |

Modelled sum 1245.3 ms vs measured 1302.8 ms median — within 4.4%, and equal to the min-based figure.

`git ls-remote origin refs/heads/dev` measures 549.5 ms, i.e. not cheaper than the narrow `git fetch`. The cost is one round trip per repo, so the lever is concurrency rather than a lighter command.

## Root cause

- `script/build-omob.ts` `runBuild()` resolves the two clones one after the other:
  `ensureCacheClone(senpiUrl, …)` then `ensureCacheClone(omoUrl, …)`, each doing its own synchronous `git fetch`. The two repositories are independent, so the second round trip is pure latency.
- `script/omob-launcher.ts` `isCurrentOmobBuild()` reads provenance by `spawnSync(binary, ["--version"])`. That is deliberate (a sidecar must not outlive a failed install), but it re-pays the binary's full startup on every launch.
- `script/build-omob.ts` imports `RELEASE_BINARY_TARGETS` from `./build-omo-binary` at module scope, pulling zod and the sidecar-source resolver into every launch even though the value is only needed once a build is actually required.

## Scope + acceptance criteria

- Fetch the two clones concurrently on the refresh path.
- Avoid the extra `--version` spawn on the no-change path without letting provenance outlive a failed install (validate any cached provenance against the binary's own identity/stat, fall back to spawning when it does not match).
- Defer the build-only module graph so the fast path does not parse it.
- Behaviour preserved, proven by tests: an advanced `origin/dev` or `origin/main` still rebuilds; a failing fetch still exits non-zero and does not exec the binary.
- Target: no-change pre-exec overhead median cut by >= 50% on the same host, measured with the same interleaved A/B harness.

## Related

Prior omob launcher/perf work: #7991 (managed launch refreshes mainline), #7499 (compiled fast paths before the provisioning re-exec), #8046 (senpi artifact reuse by commit), #8479 (publish-staging residue).

code-yeongyu:
## What changed

The managed `omob` launcher shim runs a refresh check before it execs the engine. That check was costing more than the engine's own startup, so this makes it stop waiting on things it does not need to wait on. No behaviour changes: the same refresh decisions, the same failure handling.

1. **The two cache clones are fetched together.** They are independent repositories, so fetching `senpi` and then `omo` added a whole network round trip to every launch.
2. **The refresh check no longer spawns the installed binary to read its `--version`.** `installBinary` now records that text next to the executable together with the executable's identity (`dev`, `ino`, `size`, `mtimeMs`). The marker is believed only while it still describes the file on disk. `installBinary` publishes through `renameSync`, so every install yields a fresh inode and a marker left behind by a failed install can never be mistaken for the binary that is actually there; a missing, stale or corrupt marker falls back to spawning the executable exactly as before.
3. **Four git processes per repository became two.** `ensureCacheClone` verified the remote URL and ran `rev-parse`, then `readCommitInfo` ran `rev-parse` again plus a separate `git log`. The `rev-parse` result was never read (`runBuild` only used `.directory` and `readCommitInfo`'s commit), and one `git log` answers both the commit and its date.
4. **`RELEASE_BINARY_TARGETS` is imported on the build path**, not at module scope, so the fast path stops parsing zod and the sidecar fixture.

## Why

Measured on an idle machine (bun 1.4.2), 10 interleaved A/B pairs of the installed shim against the compiled binary it execs:

| | median | MAD | min |
|---|---:|---:|---:|
| shim `--version` | 1426.8 ms | 49.1 | 1364.0 |
| binary `--version` | 124.0 ms | 4.2 | 118.8 |
| **pre-exec overhead** | **1302.8 ms** | | 1245.2 |

Attribution (5 runs each, median): `git fetch` omo 520.4 ms, `git fetch` senpi 520.9 ms, `binary --version` 122.7 ms, `rev-parse`/`log` 64.0 ms, module graph 17.3 ms, bun startup 2.3 ms. Modelled sum 1245.3 ms vs measured 1302.8 ms, within 4.4%.

`git ls-remote` measures 549.5 ms, i.e. no cheaper than the narrow `git fetch` — the cost is one round trip per repository, so the lever is concurrency, not a lighter command.

## Observed behaviour

Fetch concurrency, production `fetchCacheClones` against the real remotes, 5 interleaved pairs:

```
serial     median 1073.8 ms (min 1016.8)
concurrent median  514.3 ms (min  510.6)
saved 559.5 ms median (52.1%)
```

End-to-end, old vs new script pair swapped between every sample so both meet the same network, n=10 interleaved: see the measurement comment below.

## QA / evidence

- `bun test script/omob-provenance.test.ts script/omob-launcher.test.ts script/omob-refresh.test.ts script/build-omob.test.ts script/omob-runtime-prune.test.ts script/omob-senpi-workspace-reset.test.ts` -> **62 pass, 0 fail**. That includes the `omob refresh integration` matrix (`same`, `changed`, `failure`, `network`, `feature`, `locked`), which is the semantics guard: an advanced `origin/dev`/`origin/main` still rebuilds, and a failing refresh still exits non-zero without launching the engine.
- New `script/omob-provenance.test.ts` covers marker hit (no spawn), a marker recording a different build (no spawn, reports not-current), no marker (spawns), a marker that outlived the executable it described (rejected, spawns), a corrupt marker (ignored, spawns), and a missing executable.
- New `fetchCacheClones` tests pin that the second fetch starts before the first finishes, that a failing fetch is still reported, and that `fetchRefArgs` narrows a plain branch and falls back for a SHA or a revision expression.
- `bun run typecheck:script` clean.

One pre-existing failure in this environment, `build-omo-binary.test.ts > sidecar parity set`, reproduces identically on the base commit with the same `bun install --ignore-scripts` tree (it needs the staged plugin payload), so it is not from this change.

Fixes #8521

---
## Summary by cubic
Fixes #8521 by cutting the managed `omob` launcher's pre-exec refresh roughly in half; the refresh was costing more than the engine's own startup. Refresh decisions and failure handling are unchanged.

- The `senpi` and `omo` cache clones are fetched together instead of serially, removing a network round trip from every launch.
- `installBinary` writes a provenance marker next to the executable binding its `--version` output to the file's identity; the marker is trusted only while it still matches, and a missing, stale, or corrupt marker falls back to spawning the executable.
- Each clone's commit and date now come from a single `git log`; the unused `rev-parse` and separate log were removed.
- `RELEASE_BINARY_TARGETS` is imported only on the build path, so the fast path no longer parses `zod` and the sidecar fixture.

<sup>Written for commit ccae8754a06e080d0594b801a4ec9a9b2019bd32. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Measured end to end

Old vs new script pair swapped between **every** sample so both meet the same network, n=10 interleaved, repeated 3 times on an idle machine:

| run | old pre-exec | new pre-exec | reduction |
|---|---:|---:|---:|
| 1 | 1287.7 ms | 678.8 ms | 47.3% |
| 2 | 1272.5 ms | 753.1 ms | 40.8% |
| 3 | 1290.5 ms | 696.1 ms | 46.1% |
| **median** | **1287.7 ms** | **696.1 ms** | **46.1%** |

Launch wall time (run 3): 1407.9 -> 813.6 ms, **42.2% faster**, and much steadier (MAD 29.7 -> 5.2). Every launch saves about 0.59 s.

Before measuring I confirmed the new path really is skipping the spawn rather than silently falling back: `readProvenanceMarker` returns the recorded text and the on-disk identity matches the marker exactly (`{"dev":16777234,"ino":1067169851,"size":109946610,"mtimeMs":1789888119608.7312}`).

### What is left, and why it stays

| remaining cost | ms | reducible? |
|---|---:|---|
| one concurrent `git fetch` round trip (max of the two repos) | ~510-600 | not without changing the freshness contract |
| `git log` + `rev-parse --abbrev-ref`, two repos | ~40 | already halved from ~88 ms |
| bun startup + `build-omob.ts` graph | ~10 | already cut by deferring the release-target import |
| `sh` shim + `exec` | ~3 | no |

`git ls-remote` measures 549.5 ms, so a lighter remote query is not cheaper — the floor is one round trip per repository. Going below it would mean answering from a cached ref (a TTL or a background refresh), i.e. a launch could run a stale engine. That is a behaviour change, so it is deliberately not done here.

PR #8524, opened 2026-09-20
perf(omob): build only what the dev binary embeds

code-yeongyu:
## What changed

`bun install` in the omo cache clone runs the root `prepare`, which is `bun run build` — the whole product graph. The compiled dev binary embeds none of it. It embeds the Senpi plugin payload, and `build-omo-native.ts` builds that payload itself, from the same script chain minus materialization, into a temp directory. So every `omob` build produced the same extension bundles twice:

```
build:senpi-plugin         -> packages/omo-senpi/plugin/extensions/...
  omo.js 1.78 MB (1009 modules), omo-task.js 1.22 MB (768), omo-member.js 204.95 KB (151),
  memory-run-supervisor.mjs 42.40 KB (151), omo-init-deep-advisor.js 86.66 KB (161), ...
build:senpi-plugin:native  -> <tmp>/plugin/extensions/...
  omo.js 1.78 MB (1009 modules), omo-task.js 1.22 MB (768), omo-member.js 204.95 KB (151),
  memory-run-supervisor.mjs 42.40 KB (151), omo-init-deep-advisor.js 86.66 KB (161), ...
```

On an omo-only rebuild that pair was 5.88 s + 6.82 s of a 21.29 s build.

The build graph and its edges move to `script/build-nodes.ts`. A **profile** names the nodes a consumer needs and the dependency closure is added automatically; `build-omob` asks for the `omo-native` profile. **With no profile the graph is unchanged**, so `bun run build`, `prepare`, publish and CI keep building exactly what they build today.

`omo-native` selects zero graph nodes on purpose: `materializeOnce()` still runs (the plugin's skills are materialized from the shared-skills submodules), and the two MCP runtime dists the payload needs are built on demand by `ensurePrebuiltNativeInputs` — which skips them entirely when their `dist` is already present.

## Why this is safe

Everything `build-omo-native.ts` reads out of the source plugin tree is tracked, so the earlier stage build was never a prerequisite:

| path | state |
|---|---|
| `packages/omo-senpi/plugin/runtime/dag/{library.js,sdk.js}` | tracked |
| `packages/omo-senpi/plugin/package.json` | tracked |
| `packages/omo-senpi/plugin/daemon-launch-spec.json` | tracked |
| `packages/omo-senpi/plugin/{README.md,NOTICE,LICENSE}` | tracked |
| `plugin/CHANGELOG.md` | copied from the repo root `CHANGELOG.md` |

## Observed behaviour

[...]

Issue #8522, opened 2026-09-20
omob build: the Senpi plugin payload is built twice per build, and the prepare DAG builds the whole product the binary never embeds


code-yeongyu:
## Summary

Every `omob` build builds the Senpi plugin payload twice. `bun install` in the omo cache clone runs `prepare` -> `bun run build`, whose DAG contains `build:senpi-plugin` (= `build:senpi-plugin:stage`); then `build-omo-binary.ts` -> `stagePluginPayload` -> `build-omo-native.ts` runs `build:senpi-plugin:native`, which is the same script chain minus `build:materialize-frontend`, redirected to a temp directory. Both runs emit byte-identical bundles.

On top of that, the prepare DAG builds the whole product (OpenCode plugin bundle, Codex Light plugin and its ~15 components, CLI, TUI, schemas, declarations, `cli-node`, ...) even though the compiled binary embeds none of it.

## Reproduction

```
bun script/build-omob.ts --cache-dir <scratch> --skip-install --name omob-prof --target darwin-arm64 --skip-fetch
```

with the senpi artifact already cached, so only the omo half runs. Timestamp each output line.

## Expected

One plugin payload build per `omob` build, and the prepare chain limited to what the compiled binary actually consumes.

## Actual

Both chains run, in one build, emitting identical bundles:

```
build:senpi-plugin      -> packages/omo-senpi/plugin/extensions/...
  omo.js 1.78 MB (1009 modules), omo-task.js 1.22 MB (768), omo-member.js 204.95 KB (151),
  memory-run-supervisor.mjs 42.40 KB (151), omo-init-deep-advisor.js 86.66 KB (161),
  sdk.js 112.97 KB (47), install.mjs 20.95 KB (7)
build:senpi-plugin:native (inside build-omo-native) -> <tmp>/plugin/extensions/...
  omo.js 1.78 MB (1009 modules), omo-task.js 1.22 MB (768), omo-member.js 204.95 KB (151),
  memory-run-supervisor.mjs 42.40 KB (151), omo-init-deep-advisor.js 86.66 KB (161),
  sdk.js 112.97 KB (47), install.mjs 20.95 KB (7)
```

Measured phases on an idle dev machine (bun 1.4.2), omo-only rebuild, TOTAL 21.29 s:

| phase | wall |
|---|---:|
| git (`--skip-fetch`) | 0.13 s |
| senpi artifact reuse + omo `bun install` | 0.71 s |
[...]

PR #8530, opened 2026-09-20
fix(omob): close the gate blockers in the dev-binary build and refresh

code-yeongyu:
## What changed

A gate review of the two merged performance PRs (#8523, #8524) found four defects. Each is fixed here with a test that fails first against the merged code.

### 1. Stale MCP runtime dists could ship under a new commit's provenance

`BUILD_PROFILES["omo-native"]` selected **no** graph nodes, which left `lsp-daemon` and `ast-grep-mcp` to `ensurePrebuiltNativeInputs`. That helper builds an artifact only when it is **absent**, and the cache checkout's `git clean -ffd` deliberately preserves ignored build output — so a dist produced by an older commit survived the checkout and was embedded unchanged into a binary stamped with the new commit's provenance.

Before #8524 the full graph rebuilt both unconditionally on every build. The profile now names both nodes, restoring exactly that. This is a correctness fix, and it gives back some of the measured build win — which is the right trade.

### 2. A malformed provenance marker crashed the launcher instead of falling back

`readProvenanceMarker` parsed the marker and then reached property access **outside** the `try`:

```
JSON.parse("null")                          -> parsed.identity           TypeError
JSON.parse('{"identity":null,...}')         -> typeof recorded.dev       TypeError
```

Both take down every `omob` launch, which is strictly worse than the behaviour being optimised away. The previous test only covered `"{not json"`, which `JSON.parse` rejects and the `catch` absorbs. The shape is now validated before use, and the new test drives eight malformed shapes through `readProvenanceMarker` **and** `isCurrentOmobBuild`.

### 3. A marker survived a mode change

The identity carried `dev`/`ino`/`size`/`mtimeMs`, so `chmod` on the executable left the marker authoritative even though the previous implementation — which actually executed the file — would have behaved differently. `mode` joins the identity.

### 4. A failed fetch abandoned its sibling

`fetchCacheClones` used `Promise.all`, which reports the first failure while the other `git fetch` is still running. The caller then releases the cache lock and exits, leaving a git process writing to a cache the next launch may lock. It now waits for every fetch to settle before propagating the first failure.

## Observed behaviour

All four captured RED first against the merged code:

```
(fail) omob provenance marker > #given a marker whose JSON is not an object ...
(fail) omob provenance marker > #given the executable's mode changes ...
(fail) fetchCacheClones > #given one fetch fails while its sibling is still running ...
(fail) selectBuildNodes > #given the omo-native profile ...
 29 pass, 4 fail
```

GREEN after: `33 pass / 0 fail` on those three files; `bun run typecheck:script` clean.

The sibling-fetch test asserts the contract without a timer: it checks the refresh has **not** reported failure while the sibling is still pending (12 microtask ticks), then releases it and asserts the rejection and that the sibling ran to completion.

## QA

Full native QA (`--version`, `doctor`, interactive boot smoke, and a complete embedded-payload digest comparison) is running against a binary built from this branch; results posted as a follow-up comment.

Full-suite note: `omob-runtime-prune.test.ts > listProcessCommands` timed out once when ten files ran together on a machine at load 107-127; it passes 12/12 when run alone (4.08s, then 0.82s). It is unrelated to this change.

Refs #8521, #8522

---
## Summary by cubic
Fixes the four gate blockers in the dev-binary build and refresh: the build now always rebuilds the MCP dists it embeds, malformed provenance markers fall back to spawning the executable, a `chmod` invalidates the marker, and a failed refresh no longer abandons a sibling fetch. This costs some of the earlier build-time win, but it is required for correctness.

**Bug Fixes**

- The `omo-native` profile now includes `lsp-daemon` and `ast-grep-mcp`, so a stale dist left in the cache clone can't be embedded under a new commit's provenance.
- `readProvenanceMarker` validates the marker's shape before use, so malformed markers are ignored instead of throwing on the launch path.
- `mode` joins the binary identity, so a permission change alone makes the marker stale.
- `fetchCacheClones` now waits for every fetch to settle before reporting a failure, so no fetch is left writing to the cache after the lock is released.

<sup>Written for commit 14324a988a3d53c155dbaf4055f98640eafae633. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Native QA at the fixed code

Built from `origin/dev` (base, `f7df95ba4`) and `origin/fix/omob-gate-blockers` (fix, `14324a988`)
in the same scratch cache, same senpi commit `9c1d8bf1c`, on the idle measurement host.

## The profile now rebuilds the MCP runtimes (blocker 1)

```
base  exit=0  nodes=1   build: profile omo-native selected 0 of 16 nodes
fix   exit=0  nodes=3   build: profile omo-native selected 2 of 16 nodes
```

`lsp-daemon` and `ast-grep-mcp` are back in the graph, so a dist left in the cache clone by an
older commit can no longer be embedded under a new commit's provenance.

## Binary parity

```
base: built darwin-arm64 (108278898 bytes, 884 embedded sidecar files)
fix : built darwin-arm64 (108278898 bytes, 884 embedded sidecar files)
```

## Full embedded payload parity (blocker 5)

Both binaries were provisioned by running `doctor`, then every file in each provisioned runtime
tree was digested and the two lists diffed — not size and count, the actual relPath + sha256 set.

```
base runtime: ~/.omo/binary-runtime/0.0.0-omob.f7df95b.9c1d8bf   886 entries
fix  runtime: ~/.omo/binary-runtime/0.0.0-omob.14324a9.9c1d8bf   886 entries
differing entries: 4 of 886   (882 byte-identical)
```

Every difference is accounted for and none is payload content:

| path | why it differs |
|---|---|
| `./omo` | the executable itself, stamped with its own commit sha |
| `./package.json` | carries `omoAiVersion`, derived from the commit pair |
| `./.provisioned` | provisioning marker written at first run |
| `./plugin/runtime/ast-grep-mcp/manifest.json` | `stagedAtUtc`; its recorded `sha256` matches |

## `--version` and `doctor` (blocker 5)

```
base --version -> omo f7df95ba4b56e78154618aace7f3cc1b0d9005ee (dev)          exit 0
fix  --version -> omo 14324a988a3d53c155dbaf4055f98640eafae633 (fix/...)      exit 0
base doctor    -> exit=0, 8 lines
fix  doctor    -> exit=0, 8 lines
```

## Interactive boot smoke (blocker 5)

Driven through a real pty (`script -q /dev/null`), not piped stdio — an earlier attempt with
`stdio: "pipe"` produced 0 bytes because the TUI needs a terminal.

```
omob-fix dev build
omo   14324a988a3d53c155dbaf4055f98640eafae633 2026-09-20T17:31:54+09:00 (fix/omob-gate-blockers)
senpi 9c1d8bf1c0c7c369ae66ad9e704f432c834fb01f 2026-09-15T22:07:36+09:00 (main)
engine-build +1789477656.9c1d8bf (scheme epoch)
⠋ Loading OmO… opening session
Optimized system prompt applied: claude-opus-5
 Tip: Use /session to see the session file, id, message count, tokens, and cost.
────────────────────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────────────────────
~ • 0/1M (0.0%) (auto)                                         claude-opus-5:off
```

Markers found: `dev build`, `omo `, `senpi `, `Tip:`. The engine loaded a session, applied its
system prompt, and rendered the composer. Teardown left `leftover procs: 0`.

## Gates
- `bun test` on the three changed-behaviour files: 33 pass / 0 fail (4 captured RED first)
- `bun run typecheck:script`: clean
- CI on #8530: 23 checks SUCCESS, 0 FAILURE

code-yeongyu:
## Correction to the build-time figure

Two corrections, both downward, both now measured.

**1.** The originally posted **-50.0%** came from n=4 on one side and n=3 on the other. Across all five timed builds per side it is **15.269878 -> 7.839220 s = -48.66%**.

**2.** More importantly, #8530 restores `lsp-daemon` and `ast-grep-mcp` to the `omo-native` profile, because leaving them to `ensurePrebuiltNativeInputs` (which builds only an *absent* artifact) let a dist from an older commit ship under a new commit's provenance. Rebuilding them unconditionally costs real time:

```
dev, empty profile (could ship a stale dist)   7.79s median   nodes=1
fix, correct profile                           9.53s median   nodes=3
pre-#8524 baseline, full prepare              15.27s median   nodes=17
```

The correctness fix costs **+1.74 s (+22.3%)**.

**The figure that stands is 15.27 s -> 9.53 s = -37.6%**, with the two MCP runtimes rebuilt on every build exactly as before. Any earlier -50.0% / -48.66% number in this issue describes a build that could embed a stale MCP runtime and should not be cited.

PR #8531, opened 2026-09-20
fix(omob): verify the executable's bytes, not just its stat metadata

code-yeongyu:
## What changed

The gate review's remaining blocker: the provenance marker compared `dev`/`ino`/`size`/`mtimeMs`/`mode`, so rewriting the executable in place while restoring those values left the marker authoritative. The refresh would then report a version the bytes on disk no longer produce. The implementation this replaced **executed** the file, which cannot miss a content swap.

The marker now records a **sha256 of the executable** and verifies it. The stat identity stays as a free pre-filter, so the digest is computed only when the cheap check already matches.

## Why this is faster than what it replaced, not a tax

Measured on the current 110 MB binary, 5 runs each:

```
sha256 of the binary   :  76  81  97 134 171 ms   median  97
spawn it for --version : 171 181 213 415 482 ms   median 213
```

Hashing is roughly half the cost of the spawn it replaces, **and** it is a stronger claim: the marker attests the exact installed bytes rather than what the executable reports about itself.

The launch figure moves accordingly — the marker no longer removes the whole verification cost, it replaces it with a cheaper one. I will re-measure and restate the launch number rather than leave the earlier one standing.

## Observed behaviour

RED first against the merged code:

```
(fail) omob provenance marker > #given the executable's bytes change while the recorded stat identity still matches ...
 8 pass, 1 fail
```

GREEN after: **18 pass / 0 fail** across `omob-provenance.test.ts` and `omob-launcher.test.ts`; `bun run typecheck:script` clean; `script/omob-provenance.ts` is 86 pure LOC.

The test re-stamps the marker's identity to the rewritten file's current stat while keeping the old digest, because `utimesSync` cannot restore sub-millisecond `mtimeMs` — my first attempt asserted an identical-metadata precondition that the filesystem would not grant. Constructing the identity match directly isolates the content check instead of approximating it.

Refs #8521

---
## Summary by cubic
Fixes the remaining blocker from #8521 by making the `omob` provenance marker bind to the executable’s bytes instead of its stat metadata. Previously the marker only compared dev/inode/size/mtime/mode, so rewriting the file in place while restoring those values left it authoritative; it now records a SHA-256 of the executable and verifies that, with stat identity kept as a cheap pre-filter.

- Adds a regression test that restores the stat identity after rewriting the binary, confirming only the content digest rejects the stale marker.
- Older markers without the digest are ignored and regenerated on the next refresh, so no migration is needed.
- Hashing the 110 MB binary is faster than spawning it for `--version` (~97 ms vs ~213 ms median), so the stronger guarantee does not add launch cost.

<sup>Written for commit c4af9eef5a1fc2e3b6089e9e0905b395c95d1234. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/oh-my-openagent/pull/8531?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Launch re-measured with the digest active

The marker file now carries `contentSha256` and the refresh verifies it, so this is the shipping configuration:

```
shim ms: 825.5 779.1 865.2 743.3 726.2 746.8 913.5 782.1 836.4 772.7
bin  ms: 124.3 151.1 131.7 129.9 126.9 125.1 127.3 141.4 142.9 146.3
shim median 780.6 +/-41.1
bin  median 130.8 +/-6.1
pre-exec overhead median 649.8 ms
```

Against the 1287.7 ms baseline that is **-49.5%**.

**The digest did not cost measurable time.** Pre-digest runs of this same path measured 678.8 / 753.1 / 696.1 ms; 649.8 ms sits inside that spread, so the added sha256 is not distinguishable from network variance at n=10. That matches the component measurement: hashing the 110 MB binary is 97 ms median against 213 ms to spawn it for `--version`, so this replaces the verification with a cheaper one rather than adding a tax.

So the launcher ends up both faster than the original **and** carrying a stronger guarantee than the code it replaced: the marker attests the exact installed bytes, where the original only asked the executable what version it thought it was.

